PR #22623, opened 2021-07-08
[NEW][ENTERPRISE] Canned responses Overview Page

MartinSchoeler:

Screen Shot 2021-07-08 at 4 51 00 PM


tiagoevanp (changes requested):
There is no necessity for the file: definition/IOmnichannelCannedResponse.ts

Need. to remove the empty files in commit.

tiagoevanp (approved):

PR #22657, opened 2021-07-13
[IMPROVE][APPS] New storage strategy for Apps-Engine file packages

d-gubert:

This is an enabler for our initiative to support NPM packages in the Apps-Engine. 

Currently, the packages (zip files) for Rocket.Chat Apps are stored as a base64 encoded string in a document in the database, which constrains us due to the size limit of a document in MongoDB (16Mb).

When we allow apps to include NPM packages, the size of the App package itself will be potentially _very large_ (I'm looking at you `node_modules`). Thus we'll be changing the strategy to store apps either with GridFS or the host's File System itself.

## Further comments

Implementation on the Apps-Engine side: https://github.com/RocketChat/Rocket.Chat.Apps-engine/pull/427


app/apps/server/storage/AppGridFSSourceStorage.ts
```diff
@@ -0,0 +1,75 @@
...
+			writeStream.write(zip);
+			writeStream.end();
+		});
+	}
+
+	public async fetch(item: IAppStorageItem): Promise<Buffer> {
+		return streamToBuffer(this.bucket.openDownloadStream(this.itemToObjectId(item)));
+	}
+
+	public async update(item: IAppStorageItem, zip: Buffer): Promise<string> {
```
d-gubert, line 40: As it is here, the `update` function will create a new file in GridFS with the new version of the app, while the previous version will be left "orphaned". This is not how it works in the FS counterpart.

What do you think should be the standardized behavior? @thassiov

thassiov: I think storing the new one, changing the reference in the db and then deleting the old one afterwards would be the way to go because we could for instance revert to the old one if there's some scenario we couldn't solve, like if the directory/file gets its permissions changed, or the server shuts down in the middle of it and we end up with some messy data

thassiov (approved):

PR #22661, opened 2021-07-13
[NEW] Logout other user endpoint

ostjen:


pierre-lehnen-rc (changes requested):
Rename the endpoint to just `users.logout` and the param to `userId`.
Make userId optional and when it's missing or the same as the currently logged user, then logout without checking the new permission.

app/api/server/v1/users.js
```diff
@@ -933,3 +933,22 @@ API.v1.addRoute('users.listTeams', { authRequired: true }, {
 		});
 	},
 });
+
+API.v1.addRoute('users.logout', { authRequired: true }, {
+	post() {
+		const userId = this.bodyParams.userId || this.userId;
+
+		if (!hasPermission(this.userId, 'logout-other-user') && userId !== this.userId) {
```
pierre-lehnen-rc, line 941:
```suggestion
		if (userId !== this.userId && !hasPermission(this.userId, 'logout-other-user')) {
```
If we compare the IDs first, then the hasPermission method will only be called when it is necessary.

tests/end-to-end/api/01-users.js
```diff
@@ -3106,4 +3106,57 @@ describe('[Users]', function() {
...
+						.expect(200)
+						.end(done);
+				});
+		},
+		);
+
+		it('should logout the requester', (done) => {
+			updatePermission('logout-other-user', []).then(
+				() => {
+					console.log(user._id);
```
pierre-lehnen-rc, line 3152:
```suggestion
```

ostjen: oops, mb

app/api/server/v1/users.js
ostjen: good call

pierre-lehnen-rc (approved):

pbenkoe:
Will this endpoint log the (other) user out of all active sessions, on all devices?

Thank you!

sampaiodiego:
yes it will @pbenkoe

PR #22667, opened 2021-07-14
[FIX] crush in the getChannelHistory method

MaestroArt:
There is a typo in the 56 line. I've added 'hiddenMessageTypes'.
How to reproduce:
- call getChannelHistory method with 'oldest' param.


johncrisp:
Thanks you for this.

The Product Team are looking at this. Can you update your branch please?

pierre-lehnen-rc (approved):

Issue #20823, opened 2021-02-19
Invite link not working using SAML authentication


leManu:
### Description:

When following an invite link on an instance using only SAML authentication, a previously logged-off user arrives on home page instead of the target channel. When the user is previously connected he comes directly to the target channel.

### Steps to reproduce:

1. Generate an invite link for a specific channel and copy it to clipboard. The URL should be like "https://go.rocket.chat/invite?host=my.rocket-chat.host&path=invite%2FxyzXYZ"
2. Open a new private browsing window
3. Paste the invite link in the address bar
4. If the intermediate page "Where do you want to open this link?" appears, click on "Web application"
5. On the instance sign-in page, click on the SAML login button
6.  Sign in using SAML

### Expected behavior:

After sign-in process the user should be redirected to the target channel.

### Actual behavior:

After sign-in process you are redirected to the instance home page, not the target channel. The invite link seems to be lost in the sign-in process.

### Server Setup Information:

- Version of Rocket.Chat Server: 3.11.0
- Operating System: Debian GNU/Linux
- Deployment Method: tar
- Number of Running Instances: 1
- DB Replicaset Oplog: yes
- NodeJS Version: v12.18.4
- MongoDB Version: 4.0.19

### Client Setup Information

- Desktop App or Browser Version: Mozilla Firefox, Google Chrome, Chromium, Safari
- Operating System: Mac OS X, Debian GNU/Linux

### Additional context

When setting `Accounts_ForgetUserSessionOnWindowClose` to on, invite links hardly work in most situations. When this setting is off, invite links work only when the user has previously opened a Rocket.Chat session in the browser.

### Relevant logs:

N/A

leManu:
Same problem with v3.11.2

leManu:
Same problem when providing a direct link to a channel or private group, after SAML authentication process the user is redirected to home page.

AvivBenShoham:
I would love to do this! Thanks for the guide, I'll follow it and let you know my results. :smile:

AvivBenShoham:
@leManu feel free to review my pull request

PR #22681, opened 2021-07-15
[FIX]  redirect doesn't work after SAML login

ostjen:
Co-authored-by: @AvivBenShoham

 

When following an invite link on an instance using only SAML authentication, a previously logged-off user is not added to the desired channel and is redirected to the home page instead of the target channel. 

When the user is previously connected he comes directly to the target channel.

## Issue(s)

https://github.com/RocketChat/Rocket.Chat/issues/20823


sampaiodiego:
closing as this was not a good solution for the problem.

PR #22684, opened 2021-07-16
[IMPROVE] Refactor `livechat.registerGuest` function

KevLehman:

### DIFF
--- a/app/livechat/server/lib/Livechat.js
+++ b/app/livechat/server/lib/Livechat.js
@@ -224,17 +224,30 @@ export const Livechat = {
 	registerGuest({ token, name, email, department, phone, username, connectionData } = {}) {
 		check(token, String);
 
-		if (email) {
-			email = email.trim();
-			validateEmailDomain(email);
-		}
-
 		let userId;
 		const updateUser = {
 			$set: {
 				token,
+				...phone?.number ? { phone: [{ phoneNumber: phone.number }] } : {},
+				...name ? { name } : {},
 			},
 		};
+
+		if (email) {
+			email = email.trim();
+			validateEmailDomain(email);
+			updateUser.$set.visitorEmails = [
+				{ address: email },
+			];
+		}
+
+		if (department) {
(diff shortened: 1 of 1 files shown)


app/livechat/server/lib/Livechat.js
```diff
@@ -282,6 +285,9 @@ export const Livechat = {
 			Object.assign(updateUser, { $unset: { department: 1 } });
```
renatobecker, line 285: We can't remove the department from the visitor object anymore. If the department is not provided, nothing will happend.

app/livechat/server/lib/Livechat.js
```diff
@@ -282,6 +285,9 @@ export const Livechat = {
...
 		} else {
 			const dep = LivechatDepartment.findOneByIdOrName(department);
+			if (!dep) {
```
renatobecker, line 288: We should validate the department first, otherwise the visitor will be created but an error will be returned.

renatobecker (changes requested):

app/livechat/server/lib/Livechat.js
```diff
@@ -268,7 +271,7 @@ export const Livechat = {
 			];
 		}
 
-		if (email && email.trim() !== '') {
+		if (email) {
```
renatobecker, line 274: What about we benefit from [Object Spread Syntax](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax) and all values into `updateUser` object once? 
IMO this improvement will make it easier to understand the code.

renatobecker (dismissed):

app/livechat/server/lib/Livechat.js
KevLehman: I left the 2 special cases (email, department) out of the spread, since they include more code than just checking for its existance.

MartinSchoeler (approved):
Looks good to me!

ostjen (approved):

PR #22686, opened 2021-07-16
[FIX] Livechat config endpoint is not returning all settings

renatobecker:


KevLehman (approved):

KevLehman:
Is this tied to any task?

PR #22689, opened 2021-07-18
[NEW] Join call action button

djcruz93:

Allows the agent to join and end the ongoing call from the omnichannel dashboard chat room if the agent has closed the call window. 

https://user-images.githubusercontent.com/51796498/126079526-8203ff54-86bd-4104-b78d-d84d9a6246e6.mp4


definition/IRoom.ts
```diff
@@ -32,6 +32,7 @@ export interface IRoom extends IRocketChatRecord {
 	lm?: Date;
 	usersCount: number;
 	jitsiTimeout: Date;
+	callStatus: string;
```
murtaza98, line 35:
```suggestion
	callStatus?: string;
```
This doesn't need to be required.

client/components/Message/Actions/Action.tsx
```diff
@@ -27,8 +27,19 @@ const Action: FC<ActionOptions> = ({ id, icon, i18nLabel, label, mid, runAction
 
 	const resolvedIcon = resolveLegacyIcon(icon);
 
+	const dangerButton = label === t('End_call');
```
murtaza98, line 30: I don't think its a good practice to resolve the value of this property here since this is a general component. Ideally it should be passed within the props just like other properties like `icon`, `label`, etc are being passed.

@gabriellsh You have more experience with UI components, am I correct in the above?

client/components/Message/Actions/Actions.tsx
```diff
@@ -19,14 +19,20 @@ type ActionOptions = {
...
-			{actions.map((action) => (
-				<Action runAction={runAction} key={action.id} {...action} />
-			))}
-		</ButtonGroup>
-	</Content>
-);
+}) => {
+	const t = useTranslation();
+
+	const alignment = actions[0]?.label === t('Join_call') ? 'flex-start' : 'center';
```
murtaza98, line 25: Same comment goes for here
https://github.com/RocketChat/Rocket.Chat/pull/22689/files#r672064888

definition/IRoom.ts
murtaza98: Also since the callStatus can only have finite states, why not create an explicit type for it rather that just assigning a general type of `string`

app/livechat/server/api/lib/livechat.js
```diff
@@ -116,10 +117,16 @@ export function settings() {
 			color: initSettings.Livechat_title_color,
 			offlineTitle: initSettings.Livechat_offline_title,
 			offlineColor: initSettings.Livechat_offline_title_color,
-			actionLinks: [
-				{ icon: 'icon-videocam', i18nLabel: 'Accept', method_id: 'createLivechatCall', params: '' },
-				{ icon: 'icon-cancel', i18nLabel: 'Decline', method_id: 'denyLivechatCall', params: '' },
-			],
+			actionLinks: {
+				webrtc: [
+					{ icon: '', i18nLabel: 'Join_call', label: TAPi18n.__('Join_call'), method_id: 'joinLivechatWebRTCCall', params: '' },
```
murtaza98, line 122: `icon` and `params` property are empty. So I don't think we should keep them unless the component isn't working without it

app/livechat/server/api/lib/livechat.js
```diff
@@ -116,10 +117,16 @@ export function settings() {
...
+					{ icon: '', i18nLabel: 'End_call', label: TAPi18n.__('End_call'), method_id: 'endLivechatWebRTCCall', params: '' },
```
murtaza98, line 123: `icon` and `params` property are empty. So I don't think we should keep them unless the component isn't working without it

app/livechat/server/api/lib/livechat.js
```diff
@@ -116,10 +117,16 @@ export function settings() {
...
+				],
+				jitsi: [
+					{ icon: 'icon-videocam', i18nLabel: 'Accept', method_id: 'createLivechatCall', params: '' },
```
murtaza98, line 126: `params` property is empty. So I don't think we should keep them unless the component isn't working without it

app/livechat/server/api/lib/livechat.js
```diff
@@ -116,10 +117,16 @@ export function settings() {
...
+					{ icon: 'icon-cancel', i18nLabel: 'Decline', method_id: 'denyLivechatCall', params: '' },
```
murtaza98, line 127: `params` property is empty. So I don't think we should keep them unless the component isn't working without it

app/livechat/server/api/v1/videoCall.js
```diff
@@ -56,3 +59,102 @@ API.v1.addRoute('livechat/video.call/:token', {
...
+			};
+			return API.v1.success({ videoCall });
+		} catch (e) {
+			SystemLogger.error('Error starting webRTC video call:', e);
+			return API.v1.failure(e);
+		}
+	},
+});
+
+API.v1.addRoute('livechat/webrtc.call/:callId', { authRequired: true }, {
```
murtaza98, line 120: Documentation missing for this endpoint

app/livechat/server/api/v1/videoCall.js
```diff
@@ -56,3 +59,102 @@ API.v1.addRoute('livechat/video.call/:token', {
...
+			if (!config.theme || !config.theme.actionLinks || !config.theme.actionLinks.webrtc) {
+				throw new Meteor.Error('invalid-livechat-config');
+			}
+
+			if (!room.callStatus || room.callStatus === 'ended' || room.callStatus === 'declined') {
+				Rooms.setCallStatus(room._id, 'ringing');
+				Messages.createWithTypeRoomIdMessageAndUser(
+					'livechat_webrtc_video_call',
+					room._id,
+					'Join my room to start the video call',
```
murtaza98, line 98: Translation missing for this statement

app/livechat/server/api/v1/videoCall.js
```diff
@@ -56,3 +59,102 @@ API.v1.addRoute('livechat/video.call/:token', {
...
+					this.user,
+					{
+						actionLinks: config.theme.actionLinks.webrtc,
+						callStatus: 'ringing',
+
```
murtaza98, line 103: Extra line

app/livechat/server/api/v1/videoCall.js
```diff
@@ -56,3 +59,102 @@ API.v1.addRoute('livechat/video.call/:token', {
...
+					},
+				);
+			}
+			const videoCall = {
+				rid: room._id,
+				provider: 'webrtc',
+				callStatus: room.callStatus,
```
murtaza98, line 110: I don't think the `room` object here has the current data. Incase the following condition is TRUE - `!room.callStatus || room.callStatus === 'ended' || room.callStatus === 'declined'` - you're updating the callStatus to `ringing` but the original object still contains stale data

app/livechat/server/api/v1/videoCall.js
```diff
@@ -56,3 +59,102 @@ API.v1.addRoute('livechat/video.call/:token', {
...
+			};
+			return API.v1.success({ videoCall });
+		} catch (e) {
+			SystemLogger.error('Error starting webRTC video call:', e);
```
murtaza98, line 114: Do we need this additional logger here. I think `API.v1.failure(e)` already logs all failed requests

app/webrtc/client/actionLink.js
```diff
@@ -0,0 +1,29 @@
+import { Session } from 'meteor/session';
+import { TAPi18n } from 'meteor/rocketchat:tap-i18n';
+import toastr from 'toastr';
+
+import { actionLinks } from '../../action-links/client';
+import { APIClient } from '../../utils/client';
+import { Messages } from '../../models';
+
+actionLinks.register('joinLivechatWebRTCCall', function(message) {
```
murtaza98, line 9: Why not use arrow functions here. The code looks much elegent that way

app/webrtc/client/actionLink.js
```diff
@@ -0,0 +1,29 @@
+import { Session } from 'meteor/session';
```
murtaza98, line 1: Lets convert this file to typescript

app/livechat/server/lib/Livechat.js
```diff
@@ -1200,6 +1200,10 @@ export const Livechat = {
 		};
 		LivechatVisitors.updateById(contactId, updateUser);
 	},
+	updateCallStatus(callId, rid, status) {
+		Rooms.setCallStatus(rid, status);
```
murtaza98, line 1204: I didn't understand why we need to maintain the same status on both `room` and `message` object. Doesn't it seem redundant?

murtaza98: or is there any difference between the statuses stored in each object? If so could you please explain?

app/webrtc/client/actionLink.js
```diff
@@ -0,0 +1,29 @@
...
+import { TAPi18n } from 'meteor/rocketchat:tap-i18n';
+import toastr from 'toastr';
+
+import { actionLinks } from '../../action-links/client';
+import { APIClient } from '../../utils/client';
+import { Messages } from '../../models';
+
+actionLinks.register('joinLivechatWebRTCCall', function(message) {
+	if (Session.get('openedRoom')) {
```
murtaza98, line 10: I didn't understand the use of this statement, could you please explain what its doing and why is it required?

app/webrtc/client/actionLink.js
```diff
@@ -0,0 +1,29 @@
...
+		const updatedMessage = Messages.findOne({ _id: message._id });
+		if (updatedMessage.callStatus === 'declined' || updatedMessage.callStatus === 'ended') {
+			toastr.info(TAPi18n.__('Call Already Ended', ''));
```
murtaza98, line 13: What does the second parameter doing? If its redundant, please remove it

app/webrtc/client/actionLink.js
```diff
@@ -0,0 +1,29 @@
...
+		}
+	}
+});
+
+actionLinks.register('endLivechatWebRTCCall', async function(message) {
+	if (Session.get('openedRoom')) {
+		const updatedMessage = Messages.findOne({ _id: message._id });
+		if (updatedMessage.callStatus === 'declined' || updatedMessage.callStatus === 'ended') {
+			toastr.info(TAPi18n.__('Call Already Ended', ''));
+		} else {
```
murtaza98, line 25: Why not simply add a return here and avoid the else statement like this.

```
               const updatedMessage = Messages.findOne({ _id: message._id });
		if (updatedMessage.callStatus === 'declined' || updatedMessage.callStatus === 'ended') {
			toastr.info(TAPi18n.__('Call Already Ended', ''));
			return;
		}
		await APIClient.v1.put(`livechat/webrtc.call/${ message._id }`, {}, { rid: updatedMessage.rid, status: 'ended' });
```

The later looks much elegant, don't you think?

app/webrtc/client/actionLink.js
```diff
@@ -0,0 +1,29 @@
...
+import { actionLinks } from '../../action-links/client';
+import { APIClient } from '../../utils/client';
+import { Messages } from '../../models';
+
+actionLinks.register('joinLivechatWebRTCCall', function(message) {
+	if (Session.get('openedRoom')) {
+		const updatedMessage = Messages.findOne({ _id: message._id });
+		if (updatedMessage.callStatus === 'declined' || updatedMessage.callStatus === 'ended') {
+			toastr.info(TAPi18n.__('Call Already Ended', ''));
+		} else {
```
murtaza98, line 14: Same goes here. Try to avoid unnecessary else statements
https://github.com/RocketChat/Rocket.Chat/pull/22689/files#r672102416

murtaza98 (changes requested):
@djcruz93 I've added my initial review to this PR. Please take a look at it. Thanks

app/livechat/server/lib/Livechat.js
djcruz93: Here I followed the same way Jitsi is done. Instead of `callStatus` we have `jitsiTimeout` stored in both the `room` and `message` object to know the status of the call. If the timeout expires then it means the call ended otherwise the call hasn't yet ended. 
We need to make sure that one call is initiated at a time in one room. In order to know whether the room has an ongoing call or not we need the call status in the `room` object. And to maintain the call status initiated by the message action buttons, we need the call status in the `message` object.

app/livechat/server/api/v1/videoCall.js
djcruz93: @murtaza98 we have a dedicated week for the documentation. Shall I do it at that time and it as a ticket in the tracking dashboard or shall I do it now only?

app/webrtc/client/actionLink.js
djcruz93: I think I can remove it because, in Jitsi, it was used to get the room id, and here I am using the room id from the message.rid instead of Session.get('openedRoom').

app/webrtc/client/actionLink.js
djcruz93: And should this string also exist in the i18n file?

app/livechat/server/api/v1/videoCall.js
murtaza98: Cool, that works

app/livechat/server/lib/Livechat.js
murtaza98: > We need to make sure that one call is initiated at a time in one room. In order to know whether the room has an ongoing call or not we need the call status in the room object.

:top: This totally makes sense. I also think that the room object should be the right place to store the call status since we have a rule that only one call can be initiated at a time in one room.

> And to maintain the call status initiated by the message action buttons, we need the call status in the message object.

:top: This doesn't makes sense, why can't we check the room's `callStatus`  property when any action button is clicked? All the checks you're performing [here](https://github.com/RocketChat/Rocket.Chat/pull/22689/files#diff-0e697798268df9d4a2d39482864eb4cf479141030a7287dd1208d51ff8c07885R12) and [here](https://github.com/RocketChat/Rocket.Chat/pull/22689/files#diff-0e697798268df9d4a2d39482864eb4cf479141030a7287dd1208d51ff8c07885R23) on the message `callStatus` property, why not perform the same check on the room `callStatus` property?

app/livechat/lib/messageTypes.js
```diff
@@ -81,6 +82,27 @@ MessageTypes.registerType({
...
+			const ss = parseInt(Math.abs(message.endTs - message.ts) / 1000) % 60;
+			let callDuration = 'Lasted';
+			if (hh > 0) {
+				callDuration += ` ${ hh } hours ${ mm } minutes ${ ss } seconds.`;
+			} else if (mm > 0) {
+				callDuration += ` ${ mm } minutes ${ ss } seconds.`;
+			} else {
+				callDuration += ` ${ ss } seconds.`;
+			}
+			return `<i class="icon-phone"></i> Call ended at ${ moment(message.endTs).format('h:mm A') } - ${ callDuration }`;
```
murtaza98, line 100: Translation missing for this message

client/components/Message/Actions/Actions.tsx
murtaza98: @djcruz93 You're right earlier, we definitely need to add some logic here to control the alignment but we need to make sure that its applied in a general way, so here's what I'm proposing here.

1. Introduce a new property to the actionLinks object, lets call it `actionLinksAlignment` for now. This property only needs to be set on the first action in the `actions` array and there we can specify the alignment value.
So [this](https://github.com/RocketChat/Rocket.Chat/pull/22689/files#diff-33cc1585dcb8424c6800cd440c0c3c8249a58b8ecafd74e38b39a408ac57f2c6R120-R129) would become
```
			actionLinks: {
				webrtc: [
					{ actionLinksAlignment: 'flex-start',  i18nLabel: 'Join_call', label: TAPi18n.__('Join_call'), method_id: 'joinLivechatWebRTCCall' },
					{ i18nLabel: 'End_call', label: TAPi18n.__('End_call'), method_id: 'endLivechatWebRTCCall' },
				],
				jitsi: [
					{ icon: 'icon-videocam', i18nLabel: 'Accept', method_id: 'createLivechatCall' },
					{ icon: 'icon-cancel', i18nLabel: 'Decline', method_id: 'denyLivechatCall' },
				],
			},
```
2. Then in this file, parse the value of this property and use it for alignment. Also we need to make sure that the default value is still `center`

djcruz93: @murtaza98 Do you think storing frontend attributes in the database is a good idea? Also, we need to anyway resolve the actionLinksAlignment in the general component.
```javascript
const Actions: FC<{ actions: Array<ActionOptions>; runAction: RunAction; mid: string }> = ({
	actions,
	runAction,
}) => {
	const t = useTranslation();

	const alignment = actions[0]?.actionLinksAlignment || 'center';

	return (
		<Content width='full' justifyContent={alignment}>
```
Please do confirm once.

murtaza98: Yups, that shouldn't be an issue.

murtaza98 (approved):

PR #22700, opened 2021-07-19
[IMPROVE] Make `shortcut` field of canned responses unique

KevLehman:


ee/app/models/server/models/CannedResponse.js
```diff
@@ -10,7 +10,7 @@ export class CannedResponse extends Base {
 
 		this.tryEnsureIndex({
 			shortcut: 1,
-		});
+		}, { unique: true });
```
renatobecker, line 13: What runs first? This command or migration files?
If this command is called before migrations, it might fail if there are duplicate shortcuts.

KevLehman: I thought I've removed this 🤔 weird. 
```suggestion
		});
```

murtaza98: Don't we still need this here? Based on my understanding, the index which is being added [here](https://github.com/RocketChat/Rocket.Chat/pull/22700/files#diff-83e1b8ab5461082ac521dde1cc7a1c61c732b893d9ae6266244e07abdd556f13R47) will only be applied when the server is updated. But what about someone who directly installs the latest version, we want this `unique` index to be applied over there as well right? And I think this can only be done by passing it in this constructor here.

Regarding Renato's comment, this command in constructor indeed runs first b4 migrations, however when I tested it locally by duplicating shortcuts on DB, this command didn't fail and the migrations ran properly and fixed the duplicate shortcuts. So based on that testing, I don't see any harm re-adding this statement here

KevLehman: The migration will create the index too on new installations, since all migrations will run on the server, even if it's a clean installation.

`ensureIndex` should "silently" fail if it cannot create the index, but that "silently" showed some errors on my local, and I'm not sure if users would also see them 🤔 that's why I preferred to remove. wdyt?

murtaza98: Ohk I wasn't aware of this concept, but if that's the case, then I think we're good

murtaza98 (approved):

Issue #21735, opened 2021-04-22
3.11.1: createDirectMessage & getRoomByTypeAndName Exception: Invalid User


apaarfus:
### Description:

Following a _major_ upgrade from 2.4 to 3.11.1, we started seeing several errors crop up in the log, listed as "`[error-invalid-user]`".  While there is some semblance of a stack trace, there's little to no additional information to explain _which_ user is generating this issue; or from what channel this issue is occurring in...just that there's a problem.

I've looked through our users both in RC's administration panel, as well as directly in MongoDB; and nothing is jumping out at me as out of place or misconfigured.

### Steps to reproduce:

1. Upgrade from (manual) 2.4 to (manual) 3.11.1
2. Check Rocket.Chat logs
  - The errors do not always appear immediately

### Expected behavior:

No exceptions -- or at the very least, a little more detail in the exception log.

### Actual behavior:

I've redacted the stack trace in this section for brevity.  Actual logs will be listed in full in the 'Relevant Logs' section.

```
I20210421-14:09:58.172(-4) Exception while invoking method getRoomByTypeAndName Error: Invalid user [error-invalid-user]
I20210421-14:09:58.431(-4) Exception while invoking method createDirectMessage Error: Invalid user [error-invalid-user]
```

While I do not have the exception to share, I have also seen similar messages for the method "`loadMissedMessages`" in the recent past as well.

### Server Setup Information:

- Version of Rocket.Chat Server: 3.11.1
- Operating System: CentOS 7
- Deployment Method: tar/manual
- Number of Running Instances: 1
- DB Replicaset Oplog: Yes (rs01)
- NodeJS Version: v12.18.4
- MongoDB Version: 4.0.24 / mmapv1 (oplog Enabled)

### Client Setup Information

N/A

### Additional context

We were not seeing these exceptions prior to our upgrade.  I'm _assuming_ something with these methods/functions changed during the _year_ we remained on the old build.  Additionally, here are some related issues I was able to locate in the tracker:

- #19703 
- #18917

We have not yet tried on the latest release to see if the issue still arises, though I fear it may still occur.

#### Rocket.Chat.Electron > Administration > View Logs

Exception @ Method: getRoomByTypeAndName

```
I20210421-14:09:58.172(-4) Exception while invoking method getRoomByTypeAndName Error: Invalid user [error-invalid-user]
	at MethodInvocation.getRoomByTypeAndName (server/publications/room/index.js:42:10)
[...]

PR #22709, opened 2021-07-20
[FIX] Avoid Invalid User Exception when entering direct chat while not authenticated #21735

vachigaggl:
Solves errors being shown to user when creating direct chat by directly navigating to URL #21735

Avoid "Invalid User Exception" and "No user with username..." being displayed to user as shown in this embedded gif when entering direct chat with no session present: 
image

## Issue(s)
#21735

## Steps to test or reproduce fix
1. Use iframe authentication (not sure this is a must)
2. Directly navigate to a URL like /direct/W5xxxxxxxxQD?layout=embedded (W5xxxxxxxxQD is the room id of an existing room) -OR- (Variant B) Directly navigate to a URL /direct/Christof_Teng?layout=embedded where Christof_Teng is a username
2. Red error notification "Invalid User" is no longer displayed
3.  "No user with username "W5xxxxxxxxQD" was found! is no longer briefly shown" -OR- (Variant B) "No user with username "Christof_Teng" was found!" is no longer briefly shown
4. The chatroom loads
5. search for getRoomByTypeAndName in javascript console. You will find that there are 2 calls, the first one fails, but the error caused by the first call is no longer shown to user.

## Further comments
The video above shows the myriad of errors that were displayed to the user when loading a direct chat when using iframe authentication. This fix shows the loading screen until successful render of the direct chat.

### DIFF
--- a/app/ui-utils/client/lib/openRoom.js
+++ b/app/ui-utils/client/lib/openRoom.js
@@ -101,14 +101,14 @@ export const openRoom = async function(type, name, render = true) {
 			return callbacks.run('enter-room', sub);
 		} catch (error) {
 			c.stop();
-			if (type === 'd') {
+			if (type === 'd' && error.error !== 'error-invalid-user') {
 				const result = await call('createDirectMessage', ...name.split(', '));
 				if (result) {
 					return FlowRouter.go('direct', { rid: result.rid }, FlowRouter.current().queryParams);
 				}
 			}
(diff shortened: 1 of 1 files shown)


nickerber:
Is there anything one can do to speed up this pull request? 
The problem is terrible for new users, especially when using the electron app. One has to reload the server after the error occurs. 
This fix perfectly solves the issue...

PR #22715, opened 2021-07-20
[NEW] NEW and EDIT pages for canned responses + Canned Responses Contextual Bar

MartinSchoeler:


KevLehman:
@MartinSchoeler , there's a failing test, can you fix what happened there? 😬

PR #22717, opened 2021-07-20
[NEW][APPS] Enable the user to choose a storage for the apps' source code

thassiov:

Adds new settings for the user to choose where the source code of the apps will be stored

2021-07-20-163450_588x513_scrot


d-gubert:
I'll suggest a different approach, that for me seems to be simpler to maintain in the long-term and requires no changes on the Apps-Engine side.

Consider we introduced a new class, say `ConfigurableAppSourceStorage`, that also extends the `AppSourceStorage` just like the FS and GridFS ones. Then, this class would maintain an internal reference to either one of the underlying storage strategies, and simply redirect method calls from the Apps-Engine to the correct object.

Since it's easier to visualize this with code, here goes a skeleton of what I imagined:
```typescript
class ConfigurableAppSourceStorage AppSourceStorage {
    private filesystem: AppFileSystemSourceStorage;
    private gridfs: AppGridFSSourceStorage;
    private storage: AppSourceStorage;
    
    constructor() {
        this.filesystem = new AppFileSystemSourceStorage();
        this.gridfs = new AppGridFSSourceStorage();

        this.store = this.gridfs;
    }

    public switchStorage(type: string): void {
        switch (type) {
            case 'filesystem':
                this.store = this.filesystem;
                break;
            case 'gridfs':
                this.store = this.gridfs;
                break;
        }
    }

    public async store(item: IAppStorageItem, zip: Buffer): Promise<string> {
        return this.storage.store(item, zip);
    }
    public async fetch(item: IAppStorageItem): Promise<Buffer> {
        return this.storage.fetch(item);
    }
    public async update(item: IAppStorageItem, zip: Buffer): Promise<string> {
        return this.storage.update(item, zip);
    }
    public async remove(item: IAppStorageItem): Promise<void> {
        return this.storage.remove(item);
    }
}
```

Then, in the orchestrator you'd simply need to assign a new instance of this class to the `appSourceStorage` property, and pass it along to the `AppManager`. Whenever the value of the setting changes, you switch the internal storage reference with a method call to `orchestrator.appSourceStorage.switchStorage(newSettingValue)`. This way, the Apps-Engine doesn't need to concern itself with what storage is the correct one.

What do you think? @thassiov

thassiov:
> I'll suggest a different approach, that for me seems to be simpler to maintain in the long-term and requires no changes on the Apps-Engine side.
> 
> Consider we introduced a new class, say `ConfigurableAppSourceStorage`, that also extends the `AppSourceStorage` just like the FS and GridFS ones. Then, this class would maintain an internal reference to either one of the underlying storage strategies, and simply redirect method calls from the Apps-Engine to the correct object.
> 
> Since it's easier to visualize this with code, here goes a skeleton of what I imagined:
> 
> ```ts
> class ConfigurableAppSourceStorage extends AppSourceStorage {
>     private filesystem: AppFileSystemSourceStorage;
>     private gridfs: AppGridFSSourceStorage;
>     private storage: AppSourceStorage;
>     
>     constructor() {
>         this.filesystem = new AppFileSystemSourceStorage();
>         this.gridfs = new AppGridFSSourceStorage();
> 
>         this.store = this.gridfs;
>     }
> 
>     public switchStorage(type: string): void {
>         switch (type) {
>             case 'filesystem':
>                 this.store = this.filesystem;
>                 break;
>             case 'gridfs':
>                 this.store = this.gridfs;
>                 break;
>         }
>     }
> 
>     public async store(item: IAppStorageItem, zip: Buffer): Promise<string> {
>         return this.storage.store(item, zip);
>     }
>     public async fetch(item: IAppStorageItem): Promise<Buffer> {
>         return this.storage.fetch(item);
>     }
>     public async update(item: IAppStorageItem, zip: Buffer): Promise<string> {
>         return this.storage.update(item, zip);
>     }
>     public async remove(item: IAppStorageItem): Promise<void> {
>         return this.storage.remove(item);
>     }
> }
> ```
> 
> Then, in the orchestrator you'd simply need to assign a new instance of this class to the `appSourceStorage` property, and pass it along to the `AppManager`. Whenever the value of the setting changes, you switch the internal storage reference with a method call to `orchestrator.appSourceStorage.switchStorage(newSettingValue)`. This way, the Apps-Engine doesn't need to concern itself with what storage is the correct one.
> 
> What do you think? @thassiov

Yes. Way better

app/apps/server/storage/AppFileSystemSourceStorage.ts
```diff
@@ -2,26 +2,28 @@ import { promises as fs } from 'fs';
 
 import { AppSourceStorage, IAppStorageItem } from '@rocket.chat/apps-engine/server/storage';
 
-const tempPath = '/Users/douglasgubert/dev/rocket.chat/Rocket.Chat/apps-engine-files/';
-
 export class AppFileSystemSourceStorage extends AppSourceStorage {
+	constructor(private storagePath: string) {
```
d-gubert, line 6: We still need a method in this class to change this `storagePath` property according to the value of the setting

d-gubert (changes requested):

app/apps/server/storage/AppFileSystemSourceStorage.ts
thassiov: Yes! It is done

app/apps/server/storage/ConfigurableAppSourceStorage.ts
```diff
@@ -0,0 +1,54 @@
...
+		this.filesystem = new AppFileSystemSourceStorage();
+		this.gridfs = new AppGridFSSourceStorage();
+
+		this.setStorage(storageType, filesystemStoragePath);
+	}
+
+	public setStorage(type: string, filesystemStoragePath?: string): void {
+		switch (type) {
+			case 'filesystem':
+				this.filesystem.setPath(filesystemStoragePath as string);
```
d-gubert, line 25: I think we should split this into two methods; one for controlling the storage and another for controlling the `fileSystemStoragePath` - this makes it easier for us to just change one setting at a time, without needing the other.

Maybe we could even make the `this.filesystem` property publicly accessible, then change the path configuration directly on the setting callback

d-gubert (changes requested):

app/apps/server/storage/ConfigurableAppSourceStorage.ts
thassiov: Done! It works well

PR #22719, opened 2021-07-20
[NEW] Collect data about LDAP, SAML, CAS and OAuth usage.

pierre-lehnen-rc:

### DIFF
--- a/app/models/server/models/Users.js
+++ b/app/models/server/models/Users.js
@@ -1560,6 +1560,18 @@ Find users to send a message by email if:
 		return this.find(query, options);
 	}
 
+	countActiveUsersByService(serviceName, options) {
+		const query = {
+			active: true,
+			type: { $nin: ['app'] },
+			roles: { $ne: ['guest'] },
+		};
+
+		query[`services.${ serviceName }`] = { $exists: true };
+
+		return this.find(query, options).count();
+	}
+
 	getActiveLocalUserCount() {
 		return this.findActive().count() - this.findActiveRemote().count();
 	}
--- a/app/models/server/models/Users.js
+++ b/app/models/server/models/Users.js
@@ -1565,7 +1565,7 @@ Find users to send a message by email if:
 			active: true,
 			type: { $nin: ['app'] },
 			roles: { $ne: ['guest'] },
-			[`services.${ serviceName }`]:  { $exists: true }
+			[`services.${ serviceName }`]: { $exists: true },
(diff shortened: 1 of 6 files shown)


app/models/server/models/Users.js
```diff
@@ -1560,6 +1560,18 @@ Find users to send a message by email if:
...
 	}
 
+	countActiveUsersByService(serviceName, options) {
+		const query = {
+			active: true,
+			type: { $nin: ['app'] },
+			roles: { $ne: ['guest'] },
+		};
+
+		query[`services.${ serviceName }`] = { $exists: true };
```
ggazzo, line 1570:
```suggestion
			roles: { $ne: ['guest'] },
			[`services.${ serviceName }`]:  { $exists: true }
		};
```

app/statistics/server/lib/getServicesStatistics.ts
```diff
@@ -0,0 +1,120 @@
...
+			enabled: value,
+			mergeRoles: settings.get(`Accounts_OAuth_Custom-${ name }-merge_roles`),
+			users: Users.countActiveUsersByService(name),
+		};
+	});
+
+	return data;
+}
+
+export function getServicesStatistics(): Record<string, any> {
```
ggazzo, line 20: provide something more specific than `any`, if you really dont know, use `unknown` but try to provide some kind of structure

ggazzo (changes requested):
almost there ;)

ggazzo (approved):

app/statistics/server/lib/getServicesStatistics.ts
```diff
@@ -0,0 +1,117 @@
+import { settings } from '../../../settings/server';
+import { Users } from '../../../models/server';
+
+function getCustomOAuthServices(): Record<string, {
+	enabled: boolean;
+	mergeRoles: boolean;
+	users: number;
+}> {
+	const customOatuh = settings.get(/Accounts_OAuth_Custom-[^-]+$/mi);
+	return Object.fromEntries(customOatuh.map(({ key, value }) => [key, {
```
geekgonecrazy, line 10: is this not a typo that will prevent from working @pierre-lehnen-rc ?

geekgonecrazy:
At some point we really should consider making that entire payload a type.  It would make it much easier to work with.  Right now to determine what type the value is supposed to be is a really involved exercise

PR #22742, opened 2021-07-22
[IMPROVE] Return open room if available for visitors

KevLehman:


app/livechat/server/api/v1/room.js
```diff
@@ -6,7 +6,7 @@ import { TAPi18n } from 'meteor/rocketchat:tap-i18n';
 import { settings as rcSettings } from '../../../../settings';
 import { Messages, LivechatRooms } from '../../../../models';
 import { API } from '../../../../api/server';
-import { findGuest, findRoom, getRoom, settings, findAgent, onCheckRoomParams } from '../lib/livechat';
+import { findGuest, findRoom, getRoom, settings, findAgent, onCheckRoomParams, findOpenRoom, findDepartment, getRoomById, updateGuest } from '../lib/livechat';
```
renatobecker, line 9: IMO the logic introduced here in the endpoint scope should be placed inside `getRoom` method ather than here. `getRoom` method is imported in other places too, so logic must be reused. 
Therefore, the new methods stated here would not be needed.

app/livechat/server/api/v1/room.js
```diff
@@ -40,7 +41,22 @@ API.v1.addRoute('livechat/room', {
 			}
 
 			const rid = roomId || Random.id();
-			const room = Promise.await(getRoom({ guest, rid, agent, extraParams }));
```
renatobecker, line 43: IMO the logic introduced here in the endpoint scope should be placed inside **getRoom** method rather than here. **getRoom** method is imported in other places too, so logic must be reused.
So, the only change needed would be to pass `departmentId` parameter via **getRoom** method.

renatobecker (changes requested):

app/livechat/server/api/v1/room.js
KevLehman: That could work 🤔 . I'm a bit reluctant to modify the behavior of a function that other places are using, since the new behavior could be not what the callers expect, so I prefered to do that on the endpoint.

Do you have in mind any place that could "break" with the modification? (some place that it's expecting a new room on every call, etc)

renatobecker: The `getRoom` method should have the same behaviour everywhere. If the same method is called, the result should be will be same in places.

renatobecker (approved):

PR #22750, opened 2021-07-23
[IMPROVE] Rewrite File Upload Modal

gabriellsh:

Image preview:
image

Video preview:
image

Files larger than 10mb:
image


client/components/modals/FileUploadModal/FileUploadModal.tsx
```diff
@@ -0,0 +1,138 @@
...
+			dispatchToastMessage({
+				type: 'error',
+				message: t('FileUpload_File_Empty'),
+			});
+			onClose();
+		}
+	}, [file, dispatchToastMessage, isValidContentType, t, onClose]);
+
+	return (
+		<Modal>
```
dougfabris, line 83: Why not the `GenericModal`?

client/components/modals/FileUploadModal/ImagePreview.tsx
```diff
@@ -0,0 +1,41 @@
...
+	};
+
+	if (error) {
+		return <GenericPreview file={file} />;
+	}
+
+	return (
+		<>
+			{loading && <PreviewSkeleton />}
+			<Box
```
dougfabris, line 29: In my opinion, we should place the image as background to prevent distortion!
**Example:**
image

dougfabris (changes requested):

client/components/modals/FileUploadModal/FileUploadModal.tsx
gabriellsh: In order to enable confirming upload with the `Enter` key, I had to place a form in the component a modify a button to be of `type='submit'`.

dougfabris: In your opinion, shouldn't we have this behavior in the `GenericModal`?

dougfabris (approved):

client/components/modals/FileUploadModal/ImagePreview.tsx
dougfabris:
```suggestion
			{!loading && <Box
```

dougfabris (changes requested):

dougfabris (dismissed):

dougfabris (approved):

PR #22751, opened 2021-07-23
Regression: Data in the "New Users" section is delayed in 1 day

matheusbsilva137:

 - Update nivo version (which was causing errors in the bar chart);
 - Fix 1 day delay in '7 days' and '30 days' periods;
 - Update tooltip theme.

## Steps to test or reproduce

Access **Administration > Engagement Dashboard**.
**Expected behavior:**
new-users-engagement-dashboard


client/components/GenericTable/index.ts
```diff
@@ -0,0 +1,6 @@
+import GenericTable from './GenericTable';
+import HeaderCell from './HeaderCell';
+
+export default Object.assign(GenericTable, {
```
d-gubert, line 4: Curious as to why you used `Object.assign` here 🤓 Is it just a shortcut to writing `GenericTable.HeaderCell = HeaderCell;` in the line above? Or is there another reason?

tassoevan: Oh boy, you are the right guy to ask me this one. `GenericTable` has a function type (kind of `(props: any) => ReactElement`) defined at `'./GenericTable'`. It doesn't provide any mechanism to extend it while changing the concrete value: `GenericTable.HeaderCell = HeaderCell ;` would trigger an error for me because `GenericTable`'s type is not an extension of `{ HeaderCell: (props: any) => ReactElement; }`. I've noticed that `Object.assign()` intersects the types of its arguments, so we get `{ (props: any): ReactElement; HeaderCell: (props: any) => ReactElement; }` at the return type.

tassoevan (approved):

client/components/GenericTable/index.ts
d-gubert: Got it! 

Does this generate a side-effect? I mean, if someone imported from the `./GenericTable` file directly, the `GenericTable` exported there would also have the `HeaderCell` object due to `Object.assign`?

Of course, that wouldn't be a problem; TS wouldn't consider the original object to contain `HeaderCell` and I guess even the usual `console.log` would not print it there, it could only be confusing if you tried to see `GenericTable`'s keys with `Object.keys` or other similar functions. Just asking, again, out of curiosity 😛

PR #22752, opened 2021-07-23
[IMPROVE] Rewrite Reactions Modal

dougfabris:
### before
image

### after
image


client/components/modals/ReactionList/ReactionList.tsx
```diff
@@ -0,0 +1,48 @@
...
+		}
+
+		openUserCard({
+			username,
+			rid,
+			target: e.currentTarget,
+			open: (e: React.MouseEvent<HTMLElement>) => {
+				e.preventDefault();
+				onClose();
+				tabBar.openUserInfo(username);
```
gabriellsh, line 36: This tabBar is received as a prop of room. Shouldn't we update this and use "useTabBarOpenUserInfo" in place of this?

client/components/modals/ReactionList/Reactions.tsx
```diff
@@ -0,0 +1,42 @@
...
+
+type ReactionsProps = {
+	reactions: ReactElement;
+	onClick: (e: React.MouseEvent<HTMLElement>) => void;
+};
+
+const Reactions = ({ reactions, onClick }: ReactionsProps): ReactElement => {
+	const useRealName = useSetting('UI_Use_Real_Name');
+	return (
+		<Box display='flex' flexDirection='column' minHeight='25vh'>
```
gabriellsh, line 18: I don't like this minHeight that much. If there are only a few reactions, it looks weird:
image

gabriellsh (changes requested):

client/components/modals/ReactionList/Reactions.tsx
dougfabris: changed to 15vh

client/components/modals/ReactionList/ReactionList.tsx
dougfabris: Already discussed

gabriellsh (approved):

PR #22772, opened 2021-07-27
[FIX] Support ID param on createVisitor method

murtaza98:

### DIFF
--- a/app/apps/server/bridges/livechat.ts
+++ b/app/apps/server/bridges/livechat.ts
@@ -127,6 +127,7 @@ export class AppLivechatBridge extends LivechatBridge {
 			email: '',
 			connectionData: undefined,
 			phone: {},
+			id: visitor.id,
 		};
 
 		if (visitor.visitorEmails && visitor.visitorEmails.length) {
--- a/app/livechat/server/lib/Livechat.js
+++ b/app/livechat/server/lib/Livechat.js
@@ -223,6 +223,7 @@ export const Livechat = {
 
 	registerGuest({ id, token, name, email, department, phone, username, connectionData } = {}) {
 		check(token, String);
+		check(id, Match.Maybe(String));
 
 		let userId;
 		const updateUser = {
@@ -264,7 +265,7 @@ export const Livechat = {
 			const userData = {
 				username,
 				ts: new Date(),
-				...id && { id },
+				...id && { _id: id },
 			};
 
 			if (settings.get('Livechat_Allow_collect_and_store_HTTP_header_informations')) {
--- a/app/livechat/server/lib/Livechat.js
+++ b/app/livechat/server/lib/Livechat.js
@@ -221,7 +221,7 @@ export const Livechat = {
 		return true;
 	},
 
-	registerGuest({ token, name, email, department, phone, username, connectionData } = {}) {
+	registerGuest({ id, token, name, email, department, phone, username, connectionData } = {}) {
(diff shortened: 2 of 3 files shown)


app/livechat/server/lib/Livechat.js
```diff
@@ -264,6 +264,7 @@ export const Livechat = {
 			const userData = {
 				username,
 				ts: new Date(),
+				...id && { id },
```
KevLehman, line 267: Mongo's id is `_id`.

app/livechat/server/lib/Livechat.js
```diff
@@ -221,7 +221,7 @@ export const Livechat = {
 		return true;
 	},
 
-	registerGuest({ token, name, email, department, phone, username, connectionData } = {}) {
+	registerGuest({ id, token, name, email, department, phone, username, connectionData } = {}) {
 		check(token, String);
```
KevLehman, line 226: Can you check if id is a String? Since this is a public facing endpoint, i don't want ppl to be passing objects as `id`.
`check(id, Match.Maybe(String));`

KevLehman (changes requested):

app/livechat/server/lib/Livechat.js
murtaza98: Ooops :see_no_evil: Glad you caught this

app/livechat/server/lib/Livechat.js
murtaza98: Makes sense, I've added it

app/apps/server/bridges/livechat.ts
```diff
@@ -127,6 +127,7 @@ export class AppLivechatBridge extends LivechatBridge {
 			email: '',
 			connectionData: undefined,
 			phone: {},
+			id: visitor.id,
```
renatobecker, line 130: The **username** needs to be optional too. When it comes to Omnichannel, it doesn't make much sense to request this information, the serverside handles this following the format `@guest-XXX` .

Currently, every App is using a different approach to create usernames and this is not a good UX.

murtaza98: This would require a new release on the apps-engine side to make the username property in `IVisitor` object optional, so we can take this up as a separate task in the next release

d-gubert: Changing this would break the compatibility of the API, so we'll most definitely want to avoid that kind of change ☹️ But by checking [the source code of the function](https://github.com/RocketChat/Rocket.Chat/blob/apps/change-storage-strategy/app/livechat/server/lib/Livechat.js#L260) it looks like just setting `visitor.username = ''` would be enough to trigger the behavior

d-gubert:
LGTM! But since @KevLehman left some comments I'll leave the approval to him 😉

server/features/EmailInbox/EmailInbox_Incoming.ts
```diff
@@ -43,6 +43,7 @@ function getGuestByEmail(email: string, name: string, department?: string): any
 		phone: undefined,
 		username: undefined,
 		connectionData: undefined,
+		id: undefined,
```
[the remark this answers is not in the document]
KevLehman: Just checked why this is needed 😅

KevLehman (approved):

d-gubert (approved):

PR #22798, opened 2021-07-28
Regression: Text wrap in MarkdownTextEditor and PreviewText

tiagoevanp:

### DIFF
--- a/client/components/MarkdownText.tsx
+++ b/client/components/MarkdownText.tsx
@@ -1,7 +1,7 @@
 import { Box } from '@rocket.chat/fuselage';
 import dompurify from 'dompurify';
 import marked from 'marked';
-import React, { FC, useMemo } from 'react';
+import React, { ComponentProps, FC, useMemo } from 'react';
 
 import { renderMessageEmoji } from '../lib/renderMessageEmoji';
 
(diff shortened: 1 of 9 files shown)


client/components/MarkdownText.tsx
```diff
@@ -107,7 +107,8 @@ const MarkdownText: FC<Partial<MarkdownTextParams>> = ({
 
 		return preserveHtml ? html : html && sanitizer(html, { ADD_ATTR: ['target'] });
 	}, [content, preserveHtml, sanitizer, markedOptions, parseEmoji]);
-
+	console.log(content);
```
KevLehman, line 110: Remove console logs pls :)

ee/client/omnichannel/components/CannedResponse/modals/CreateCannedResponse/PreviewText.tsx
```diff
@@ -5,6 +5,7 @@ import MarkdownText from '../../../../../../../client/components/MarkdownText';
 
 const PreviewText: FC<{ text: string }> = ({ text }) => (
 	<Box
+		style={{ wordBreak: 'break-word' }}
```
KevLehman, line 8: Some sites suggest break-word is not standard for Webkit and that it may not work there. MDN also says this is a deprecated property. Can we confirm if this change will work in all browsers or if we should change properties?

KevLehman (changes requested):

client/components/MarkdownText.tsx
tiagoevanp: ops

KevLehman (approved):

PR #22611, opened 2021-07-07
[FIX][ENTERPRISE] Make AutoSelect algo take current agent load in consideration

KevLehman:

PR #22838, opened 2021-08-03
[FIX] Chat Transfer Failure due to Unrecognized stage name departmentId

Shailesh351:
This PR fixes typo in the collection name and modifies aggregator query to pass the customFilter properly

## Issue(s)
https://open.rocket.chat/channel/omnichannel/thread/sxL6dbZBDoAoYJDqN

Because of extraQuery(customFilter) is being passed as a stage in aggregator [here](https://github.com/RocketChat/Rocket.Chat/pull/22611/files#diff-3fcd471e2731d9972594322e99cac33b2f78e05ef84e78bac061503d73180296R276), It is giving `MongoError: Unrecognized pipeline stage name: 'departmentId'`

## Steps to test or reproduce
Call `modify.getUpdater().getLivechatUpdater().transferVisitor` from RC App

### DIFF
--- a/app/models/server/models/Users.js
+++ b/app/models/server/models/Users.js
@@ -219,7 +219,7 @@ export class Users extends Base {
 		// if department is provided, remove the agents that are not from the selected department
 		const departmentFilter = departmentId ? [{
 			$lookup: {
-				from: 'rocketchat_livechat_department_agent',
+				from: 'rocketchat_livechat_department_agents',
 				let: { departmentId: '$departmentId', agentId: '$agentId' },
 				pipeline: [{
 					$match: { $expr: { $eq: ['$$agentId', '$_id'] } },
@@ -273,7 +273,7 @@ export class Users extends Base {
 					},
 				},
 			},
-			...customFilter ? [customFilter] : [],
+			...customFilter ? [{ $match: customFilter }] : [],
 			{
 				$sort: {
 					'queueInfo.chats': 1,


app/models/server/models/Users.js
```diff
@@ -273,7 +273,7 @@ export class Users extends Base {
 					},
 				},
 			},
-			...customFilter ? [customFilter] : [],
+			...customFilter ? [{ $match: customFilter }] : [],
```
murtaza98, line 276: Are you sure this was required? TBH, I don't think this is required because `customFilters` are getting defined by [this](https://github.com/RocketChat/Rocket.Chat/blob/607d42db9bf42a7e25f3fa520d989efdbe5d8eca/ee/app/livechat-enterprise/server/hooks/applySimultaneousChatsRestrictions.ts#L5) callback and as you can see [here](https://github.com/RocketChat/Rocket.Chat/blob/607d42db9bf42a7e25f3fa520d989efdbe5d8eca/ee/app/livechat-enterprise/server/hooks/applySimultaneousChatsRestrictions.ts#L9) and [here](https://github.com/RocketChat/Rocket.Chat/blob/607d42db9bf42a7e25f3fa520d989efdbe5d8eca/ee/app/livechat-enterprise/server/hooks/applySimultaneousChatsRestrictions.ts#L32) both these return values apply a $match filter

Shailesh351: When I logged what we are receiving in this function, It was showing `{departmentId: 'string'}`.

FYI I don't have enterprise edition, so `/ee/app/livechat-enterprise/server/hooks/applySimultaneousChatsRestrictions.ts`  hook won't be called.

Maybe the issue is only with a non-enterprise server?

murtaza98: Okay got it. I found out that this was happening because `livechat.applySimultaneousChatRestrictions` callback was simply returning the first param passed to it incase the callback was not found, which happens in the community edition. 

So I think the correct way to use this callback was to pass the params/constants in the second param and keep the first param empty/default. 

I took the liberty of raising a PR for the same https://github.com/RocketChat/Rocket.Chat/pull/22839

Shailesh351: Thanks @murtaza98 🙃

Shailesh351:
Closing in favor of https://github.com/RocketChat/Rocket.Chat/pull/22839

PR #22846, opened 2021-08-03
Call status update

Deepak-learner:

### DIFF
--- a/app/theme/client/imports/general/base.css
+++ b/app/theme/client/imports/general/base.css
@@ -239,7 +239,7 @@ button {
 	.rc-old .main-content-flex {
 		flex: 1 0 auto !important; /* 1 */
 
-		height: auto !important; /* 1 */ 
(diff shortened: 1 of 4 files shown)


Deepak-learner:
I create function updateCallStatus at sdk to communicate with rc-core. PR => https://github.com/RocketChat/Rocket.Chat.js.SDK/pull/142

 updateCallStatus is used at livechat to update the call status PR => https://github.com/RocketChat/Rocket.Chat.Livechat/pull/624

app/theme/client/imports/general/base.css
```diff
@@ -239,7 +239,7 @@ button {
 	.rc-old .main-content-flex {
```
murtaza98, line 239: Are this changes required?

Deepak-learner: No, it's not needed.

PR #22847, opened 2021-08-03
[NEW] P2P WebRTC Connection Establishment

djcruz93:

Allowing a visitor and agent to establish P2P webrtc connection and call.

Note: The user needs to increase the limit by a connection from 10 to 15 in the DDP rate limiter section.

## Testing Instructions
1. Enable `webrtc` settings in the `Administration > WebRTC`.
2. Set STUN/TURN server config. Please make sure those credentials are active and work fine.
3. Select `WebRTC` as the call provider from the `Video and Audio Call` section in the `Administration > Omnichannel`.
4. Agent should see a call button above in the action bar with the label as `WebRTC Call`.
5. Set the value of `Limit by Connection per Method: requests allowed` to `15` in the `Administration > Rate Limiter`.
6. Start a new LiveChat conversation.
7. Click on the call button to initiate a call with the visitor.
8. You will be redirected to a new tab with instructions as waiting for visitors to join.
9. Now extract the visitor token and id from the LiveChat widget.
10. Open a new tab with the URL as `http://localhost:3000/meet/<LivechatRoomID>?token=<VisitorToken>`
11. Accept the prompt to give mic and camera access and the call should start.

### DIFF
--- a/app/models/server/raw/Subscriptions.ts
+++ b/app/models/server/raw/Subscriptions.ts
@@ -25,7 +25,7 @@ export class SubscriptionsRaw extends BaseRaw<T> {
 		return this.find(query, options);
 	}
 
-	findByRoomIdAndNotUserId(roomId: string, userId: string | undefined, options: FindOneOptions<T> = {}): Cursor<T> {
+	findByRoomIdAndNotUserId(roomId: string, userId: string, options: FindOneOptions<T> = {}): Cursor<T> {
 		const query = {
 			rid: roomId,
 			'u._id': {
@@ -36,7 +36,7 @@ export class SubscriptionsRaw extends BaseRaw<T> {
 		return this.find(query, options);
 	}
 
-	findByLivechatRoomIdAndNotUserId(roomId: string, userId: string | undefined, options: FindOneOptions<T> = {}): Cursor<T> {
+	findByLivechatRoomIdAndNotUserId(roomId: string, userId: string, options: FindOneOptions<T> = {}): Cursor<T> {
 		const query = {
 			rid: roomId,
 			'servedBy._id': {
--- a/app/models/server/raw/Subscriptions.ts
+++ b/app/models/server/raw/Subscriptions.ts
@@ -25,7 +25,7 @@ export class SubscriptionsRaw extends BaseRaw<T> {
 		return this.find(query, options);
 	}
 
-	findByRoomIdAndNotUserId(roomId: string, userId: string, options: FindOneOptions<T> = {}): Cursor<T> {
+	findByRoomIdAndNotUserId(roomId: string, userId: string | undefined, options: FindOneOptions<T> = {}): Cursor<T> {
 		const query = {
 			rid: roomId,
 			'u._id': {
@@ -36,7 +36,18 @@ export class SubscriptionsRaw extends BaseRaw<T> {
 		return this.find(query, options);
 	}
 
-	countByRoomIdAndUserId(rid: string, uid: string): Promise<number> {
+	findByLivechatRoomIdAndNotUserId(roomId: string, userId: string | undefined, options: FindOneOptions<T> = {}): Cursor<T> {
+		const query = {
+			rid: roomId,
+			'servedBy._id': {
+				$ne: userId,
+			},
+		};
+
+		return this.find(query, options);
+	}
+
+	countByRoomIdAndUserId(rid: string, uid: string | undefined): Promise<number> {
 		const query = {
 			rid,
 			'u._id': uid,
--- a/client/startup/routes.ts
+++ b/client/startup/routes.ts
@@ -1,5 +1,6 @@
 import { FlowRouter } from 'meteor/kadira:flow-router';
 import { Meteor } from 'meteor/meteor';
+import { TAPi18n } from 'meteor/rocketchat:tap-i18n';
 import { Session } from 'meteor/session';
 import { Tracker } from 'meteor/tracker';
 import { lazy } from 'react';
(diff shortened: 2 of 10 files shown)


client/views/meet/MeetPage.js
```diff
@@ -0,0 +1,59 @@
...
+import NotFoundPage from '../notFound/NotFoundPage';
+import PageLoading from '../root/PageLoading';
+import CallPage from './CallPage';
+
+function MeetPage() {
+	const [isRoomMember, setIsRoomMember] = useState(false);
+	const [status, setStatus] = useState(null);
+	const roomId = useRouteParameter('rid');
+	const visitorToken = useQueryStringParameter('token');
+	const visitorId = useQueryStringParameter('id');
```
murtaza98, line 15: Passing ID in query param is redundant since we already have the token in the query param and we can get the id using token

app/models/server/raw/Subscriptions.ts
```diff
@@ -25,7 +25,7 @@ export class SubscriptionsRaw extends BaseRaw<T> {
 		return this.find(query, options);
 	}
 
-	findByRoomIdAndNotUserId(roomId: string, userId: string, options: FindOneOptions<T> = {}): Cursor<T> {
+	findByRoomIdAndNotUserId(roomId: string, userId: string | undefined, options: FindOneOptions<T> = {}): Cursor<T> {
```
murtaza98, line 28: I didn't understand why the second param is `undefined`. If the userID is not present, then I think we're only searching using `roomId`, so perhaps we should be using `findByRoomId` instead

app/models/server/raw/Subscriptions.ts
```diff
@@ -36,7 +36,18 @@ export class SubscriptionsRaw extends BaseRaw<T> {
 		return this.find(query, options);
 	}
 
-	countByRoomIdAndUserId(rid: string, uid: string): Promise<number> {
+	findByLivechatRoomIdAndNotUserId(roomId: string, userId: string | undefined, options: FindOneOptions<T> = {}): Cursor<T> {
```
murtaza98, line 39: https://github.com/RocketChat/Rocket.Chat/pull/22847/files#r683467955
Same goes here

app/models/server/raw/Subscriptions.ts
djcruz93: While I was in this file I saw this line as red underlined. I'll just try to revert back this line, push and see it it passes all the type check.

client/startup/routes.ts
```diff
@@ -50,6 +51,25 @@ FlowRouter.route('/login', {
...
+	name: 'meet',
+
+	async action(_params, queryParams) {
+		if (queryParams?.token !== undefined) {
+			// visitor login
+			const visitor = await APIClient.v1.get(`/livechat/visitor/${queryParams?.token}`);
+			if (visitor?.visitor) {
+				return appLayout.render({ component: MeetPage });
+			}
+			return toastr.error("Visitor doesn't exist!");
```
murtaza98, line 64: Translations missing for this error

definition/ISubscription.ts
```diff
@@ -7,6 +7,7 @@ type RoomID = string;
 
 export interface ISubscription extends IRocketChatRecord {
 	u: Pick<IUser, '_id' | 'username' | 'name'>;
+	v: Pick<IUser, '_id' | 'username' | 'name'>;
```
murtaza98, line 10:
```suggestion
	v?: Pick<IUser, '_id' | 'username' | 'name'>;
```

server/modules/notifications/notifications.module.ts
```diff
@@ -223,21 +227,42 @@ export class NotificationsModule {
...
+				if (room.t === 'l' && e === 'webrtc') {
+					notifyUser(room.servedBy._id, e, ...args);
+					return false;
+				}
 			}
-
-			const [roomId, e] = eventName.split('/');
 			if (await Subscriptions.countByRoomIdAndUserId(roomId, this.userId) > 0) {
+				const livechatSubscriptions: ISubscription[] = await Subscriptions.findByLivechatRoomIdAndNotUserId(roomId, this.userId, { projection: { 'v._id': 1, _id: 0 } }).toArray();
+				if (livechatSubscriptions) {
```
murtaza98, line 240: I think this also needs an event check - `e === 'webrtc'`

app/models/server/raw/Subscriptions.ts
djcruz93: @murtaza98 if I remove | undefined I get this warning where I call this function. Please refer the image below.
image

murtaza98:
Hi @djcruz93 

I found these 2 issues in this PR while testing
- [Critical] Audio not working. I tested this across multiple devices but couldn't get the audio to work
- The agent is not notified in any way that if the call is Declined by visitor. The agent is stuck on **Waiting for visitor to join...** screen.

Also I wanted to understand your plan regarding selecting Audio and Video sources so could you please share your plans about this?

djcruz93:
@murtaza98 
> * [Critical] Audio not working. I tested this across multiple devices but couldn't get the audio to work
The audio is working but I had added the muted attribute due to which you were not able to hear.
> * The agent is not notified in any way that if the call is Declined by visitor. The agent is stuck on **Waiting for visitor to join...** screen.
This case needs to be handled on the meet screen. I'll just add a commit.
> * After the agent closes the call via the end call button, I got the called Ended screen on both agent and visitor, but my camera was still on and I the browser shows that the tab is still using camera for both visitor and agent
This will be handled with the control buttons PR.
> Also I wanted to understand your plan regarding selecting Audio and Video sources so could you please share your plans about this?
The audio/video source selection is in the roadmap for @Deepak-learner. I think he'll be able to answer this question better.

murtaza98:
Update: Based on discussion with Dhruv, the call decline event will be part of the next PR

murtaza98 (approved):

PR #22848, opened 2021-08-03
[FIX] Broken download link on uploaded files

ostjen:

Uploaded files had wrong download links when the deploy had a sub directory. This misbehavior was caused by the wrong usage of the rtrim method, the 2nd parameter is a list of chars, [not a string](https://www.php.net/manual/pt_BR/function.rtrim.php) (this method was inspired by php)


app/utils/lib/getURL.js
```diff
@@ -54,7 +55,7 @@ export const _getURL = (path, { cdn, full, cloud, cloud_route, cloud_params, _cd
 	}
 
 	if (full) {
-		return s.rtrim(siteUrl, pathPrefix) + url;
+		return siteUrl.replace(new RegExp(`${ escapeRegExp(/pathPrefix$/) }`, '')) + url;
```
sampaiodiego, line 58: that's not how `escapeRegExp` works.. what you have to pass to it is the string you want to make sure doesn't behave like a regex, so you can use it's return on your own regex.. looks like the param for the `replace` is misplaced.. I think this fixes both issues:
```suggestion
		return siteUrl.replace(new RegExp(`${ escapeRegExp(pathPrefix) }$`), '') + url;
```

sampaiodiego (changes requested):

app/utils/lib/getURL.js
ostjen: I tried this approach and it worked :smiley ..  it's fixed, thanks

sampaiodiego (approved):

PR #22850, opened 2021-08-03
[FIX] Return transcript/dashboards based on timezone settings

KevLehman:

- Added new setting to manage timezones
- Applied new setting to omnichannel dashboards (realtime, analytics) [NOTE: Other dashboards aren't using this setting actually)
- Change getAnalyticsBetweenDate query to filter out system messages instead of substracting them


app/lib/server/startup/settings.js
```diff
@@ -1071,6 +1071,28 @@ settings.addGroup('General', function() {
...
+				i18nLabel: 'Default_User_Timezone',
+			}],
+		});
+		this.add('Default_Custom_Timezone', '', {
+			type: 'timezone',
+			enableQuery: {
+				_id: 'Default_Timezone_For_Reporting',
+				value: 'custom',
+			},
+		});
```
renatobecker, line 1094: The setting translations are missing(EN at least)
Screen Shot 2021-08-10 at 1 34 04 PM

renatobecker (changes requested):

app/lib/server/startup/settings.js
KevLehman: I added translations for english & spanish

sampaiodiego (commented):
since you changed a lot of `console.log`, may I ask you to use `Logger` instead? :hugs:

KevLehman:
> since you changed a lot of `console.log`, may I ask you to use `Logger` instead? 🤗

Done! 🚀

app/livechat/server/lib/Analytics.js
```diff
@@ -122,10 +155,14 @@ export const Analytics = {
 		Total_messages(date, departmentId) {
 			let total = 0;
 
-			LivechatRooms.getAnalyticsMetricsBetweenDate('l', date, { departmentId }).forEach(({ msgs }) => {
+			// we don't want to count visitor messages
+			const extraFilter = { $lte: ['$token', null] };
```
murtaza98, line 159: Hi @KevLehman Just wanted to confirm about the logic we're adding here. Based on what I see in our docs, the **Total messages** count includes messages from visitor as well
image

KevLehman: That's a great observation, let's get confirmation from @renatobecker about the expected behavior here, and if we should update code/docs. 

(IMO, It makes sense to show only user's messages on total messages, since that's an agent metric)

renatobecker: Let's render only agent metrics and wait for the refactor of Analytics dashboards to improve it.

client/views/omnichannel/analytics/InterchangeableChart.js
```diff
@@ -20,6 +43,8 @@ const InterchangeableChart = ({ departmentId, dateRange, chartName, ...props })
 
 	const draw = useMutableCallback(async (params) => {
 		try {
+			console.log(chartName);
```
murtaza98, line 46:
```suggestion
```

renatobecker (approved):
LGTM.

PR #22854, opened 2021-08-04
[NEW] update call status

Deepak-learner:
At livechat widget side, when visitor click on accept or decline button, with the updateCallStatus function, I am making a post request with updated call status value and update the status of call at rc-core. Below I attach sdk and livechat widget pr, which all together will be needed to make this work.

I create function updateCallStatus at sdk to communicate with rc-core. PR => RocketChat/Rocket.Chat.js.SDK#142

Livechat.updateCallStatus is used at livechat to update the call status PR => RocketChat/Rocket.Chat.Livechat#624

### DIFF
--- a/app/livechat/server/api/v1/visitor.js
+++ b/app/livechat/server/api/v1/visitor.js
@@ -128,6 +128,29 @@ API.v1.addRoute('livechat/visitor/:token/room', { authRequired: true }, {
 	},
 });
 
+API.v1.addRoute('livechat/visitor.callStatus', {
+	post() {
+		try {
+			check(this.bodyParams, {
+				token: String,
+				callStatus: String,
+				rid: String,
+			});
+
+			const { token, callStatus, rid } = this.bodyParams;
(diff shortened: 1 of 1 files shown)


app/livechat/server/api/v1/visitor.js
```diff
@@ -128,6 +128,29 @@ API.v1.addRoute('livechat/visitor/:token/room', { authRequired: true }, {
...
+				rid: String,
+			});
+
+			const { token, callStatus, rid } = this.bodyParams;
+			const guest = findGuest(token);
+			if (!guest) {
+				throw new Meteor.Error('invalid-token');
+			}
+			const status = callStatus;
+			Livechat.updateCallStatus('', rid, status, '');
```
murtaza98, line 146:
```suggestion
			Livechat.updateCallStatus('', rid, status);
```

The forth param is a user object, so a empty string won't be compatible

murtaza98 (approved):

PR #22856, opened 2021-08-04
[FIX] Message update not working in some cases

murtaza98:
## Issue(s)

The recent changes we made for canned responses is breaking message update feature on the app side. When an app tries to update a message, it gets the following error
```
 "{\"stack\":\"TypeError: Cannot read property 't' of undefined\\n    at isOmnichannelRoom (definition/IRoom.ts:96:90)\\n    at ee/app/canned-responses/server/hooks/onMessageSentParsePlaceholder.ts:32:7\\n    at callbacks.runItem (app/callbacks/lib/callbacks.js:111:70)\\n    at Object.callbacks.runItem (app/metrics/server/callbacksMetrics.js:24:20)\\n    at app/callbacks/lib/callbacks.js:39:35\\n    at app/callbacks/lib/callbacks.js:45:45\\n    at app/callbacks/lib/callbacks.js:45:47\\n    at callbacks.run (app/callbacks/lib/callbacks.js:127:9)\\n    at Object.callbacks.run (app/metrics/server/callbacksMetrics.js:14:17)\\n    at updateMessage (app/lib/server/functions/updateMessage.js:48:22)\\n    at app/apps/server/bridges/messages.ts:50:3\\n    at /app/bundle/programs/server/npm/node_modules/meteor/promise/node_modules/meteor-promise/fiber_pool.js:43:40\",\"message\":\"Cannot read property 't' of undefined\"}"
```


ee/app/canned-responses/server/hooks/onMessageSentParsePlaceholder.ts
```diff
@@ -28,7 +28,12 @@ const placeholderFields = {
 	},
 };
 
-callbacks.add('beforeSaveMessage', (message: IMessage, room: IOmnichannelRoom): any => {
+callbacks.add('beforeSaveMessage', (message: IMessage): any => {
+	if (!message.msg || message.msg === '') {
+		return message;
+	}
+
+	const room: IOmnichannelRoom = Rooms.findOneById(message.rid);
```
renatobecker, line 36: If a room object is expected but not provided, we should return IMO.
```suggestion
	if (!room || !isOmnichannelRoom(room)) {
```

murtaza98: Actually I went with this approach because I thought it would be the safest considering I found some inconsistencies in the codebase. When I checked all the places throughout the codebase which is running `beforeSaveMessage` hook, all of them expect one passes the second room param. For reference please see the image below.
image

Hence due to this inconsistency, I though it would be best if we used the first param to resolve the room object as it would work for all the hooks. 

Let me know if you still think its better to go with the approach you suggested. Thanks :smile:

murtaza98: Implemented the following logic
> try to use the room param when available and let's try to find the room if the param is not there or is invalid

Issue #22556, opened 2021-07-02
[IMPROVE] Quote Padding


TNF13:
### Description:

Reduce padding on quotes to be more user-friendly.

### Steps to reproduce:

1. Quote someone else's post or quote 
2. Send message

Lower quote padding is unnecessary and bulks up the chat.
![image](https://user-images.githubusercontent.com/86852550/124318782-b71aca00-db68-11eb-8acd-6f85821f5e78.png)

### Server Setup Information:

Unsure if relevant. Will update if it is.

- Version of Rocket.Chat Server: 13.13.1
- Operating System: 
- Deployment Method: MongoDB
- Number of Running Instances: 1

### Client Setup Information

- Desktop App or Browser Version: Both & mobile
- Operating System: Windows/Android/iOS

### Additional context

Shows in multiple versions of Rocket.

### Relevant logs:

NA

johncrisp:
Hi.

Can you please first test this on latest - currently 3.16.3?

Any idea on the CSS that might require looking at?

juniopereirab:
I can help with it. On current version there is no more cascading quote like the example, but the padding is still bigger than related above.

ggazzo:
Look at our design: 
![image](https://user-images.githubusercontent.com/5263975/142029315-4378bc20-783a-4d67-a2c7-8514afc8c44a.png)
🤷  
maybe in a future we can review our decisions about paddings and margins, but I dont think this is going to be the only place affected, until then I will consider this no longer an issue thanks for your feedback

PR #22872, opened 2021-08-05
[IMPROVE] Smaller quote padding

juniopereirab:

image
image

Change the padding size from x16 to x4.

## Issue(s)

#22556 
## Steps to test or reproduce

Quote someone message and then quote again to group quotes.


ggazzo:
this is a matter of design, I dont think we gonna change that based on a single feedback =/, but I will send to designers to check that

image

PR #22879, opened 2021-08-06
[FIX] Threads List being requested more than expected

ggazzo:


sampaiodiego (changes requested):
I'm still seeing lots of requests to get the threads list when a message is sent on a thread a have opened.

sampaiodiego (approved):

PR #22399, opened 2021-06-17
Chore: Meteor 2.2 and bump dependencies

ggazzo:

PR #22882, opened 2021-08-06
[FIX] Apps-Engine's scheduler failing to update run tasks

thassiov:

[Agenda](https://github.com/agenda/agenda), the library that manages scheduling, depended on setting a job property named `nextRunAt` as `undefined` to signal whether it should be run on schedule or not. [Rocket.Chat's current Mongo driver](https://github.com/RocketChat/Rocket.Chat/pull/22399) ignores `undefined` values when updating documents and this was causing jobs to never stop running as Agenda couldn't clear that property (set them as `undefined`). 
This updates Rocket.Chat's dependency on Agenda.js to point to [a fork that fixes the problem](https://github.com/RocketChat/agenda/releases/tag/3.1.2).

## Steps to test or reproduce

Starting from Rocket.Chat@3.17.0:

Install this example app. At startup, it creates a job that should run only once 10 seconds after it is registered.
[schedule-once_0.0.1.zip](https://github.com/RocketChat/Rocket.Chat/files/6948229/schedule-once_0.0.1.zip)

It runs every 5 seconds (minimum amount of time allowed by Agenda) indefinitely.

##### First iteration:
Make mongo driver's configuration about bson types explicit so `undefined` fields are recognized and not dropped by the query parser. The issue started happening after [Meteor's 2.2 update](https://github.com/RocketChat/Rocket.Chat/pull/22399).

[Agenda](https://github.com/agenda/agenda), the Apps Engine's dependency that manages apps scheduling, needs this feature as it depends on setting certain values as `undefined` when updating documents. For the case of tasks that run only once, the property `nextRunAt`, that determines if a job must be run in the future, is never set to undefined, thus making the scheduler run it indefinitely.

Somewhere down the dependency tree some function is ignoring or changing this value. As of right now we couldn't identify where, but it prevented updates with fields having `undefined` as their values. The problem is solved when we set this mongodb property explicitly.

Related documentation:

Node.js MongoDB Driver API: https://mongodb.github.io/node-mongodb-native/3.6/api/Db.html

##### Second iteration (current):

Fix agenda.js itself and update Rocket.Chat's dependency to the fixed version.


sampaiodiego (dismissed):
this change is breaking some endpoints, can you please take a look?

sampaiodiego (approved):

Issue #22558, opened 2021-07-03
No way to create Livechat room for visitor via REST API


mosskin-woast:
### Description:

When using the realtime livechat API, sending a message using the `sendMessageLivechat` method will create a livechat room if the given ID does not correspond to an existing one.

When using the REST API to send livechat messages, the livechat room pointed to by `rid` must already exist. However, there is no endpoint for creating livechat rooms via the REST API.

### Steps to reproduce:

1. Create a livechat visitor via the `livechat/visitor` POST endpoint
2. Attempt to send a message, passing a random room ID
3. Recieve HTTP status 400 error with code `[invalid-room]`

### Expected behavior:

Sending a message via the REST API and via the Realtime API should behave identically - OR, if there should be a difference in the behavior between the two, it should be documented, and an endpoint _must_ be added to allow creation of livechat rooms.

### Actual behavior:

Recieve HTTP status 400 error with code `[invalid-room]`

### Server Setup Information:

- Version of Rocket.Chat Server: 3.15.0
- Operating System: Docker container
- Deployment Method: Docker
- Number of Running Instances: 1 
- DB Replicaset Oplog: 
- NodeJS Version: 12.22.1
- MongoDB Version: 4.0.25 / mmapv1 (oplog Enabled)

johncrisp:
Hi,

I've raised this as a query with the OmniChannel team.

cuonghuunguyen:
https://developer.rocket.chat/api/rest-api/endpoints/omnichannel-endpoints/omnichannel-livechat-endpoints/livechat-room/livechat-room-info
is this what u meant?

PR #22890, opened 2021-08-09
[FIX] Can't create livechat room with custom room id

damianjakubiuk:

- revert back ability to use own roomId during creating livechat room
`"_currentUrl":"https://tuatarachat.eu-de.mybluemix.net/api/v1/livechat/room?token=myToken&rid=myCustomRoomId"}},"response":{"status":400,"statusText":"Bad Request","data":{"success":false,"error":"[invalid-room]","errorType":"invalid-room"}}}"`

## Issue(s)

I think this one is related to this issue #22558
## Further comments

Sorry but I cant sign in CLA assistant due to blank screen :/ 
image

### DIFF
--- a/app/livechat/server/api/v1/room.js
+++ b/app/livechat/server/api/v1/room.js
@@ -31,18 +31,19 @@ API.v1.addRoute('livechat/room', {
 			throw new Meteor.Error('invalid-token');
 		}
 
-		let room;
+		let room;		
+		let agent;
+		let newRoom = false;
+		const agentObj = agentId && findAgent(agentId);
+		if (agentObj) {
+			const { username } = agentObj;
+			agent = { agentId, username };
+		}
+
 		if (!roomId) {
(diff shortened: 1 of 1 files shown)


tomkos:
It seems to be a breaking and not compatible change made in 3.15 version, that prevents creating a room with custom rid, that was possible before this change:
https://github.com/RocketChat/Rocket.Chat/pull/22742/files
Can we get review and approval for this?

tomkos:
Fixes this issue: https://github.com/RocketChat/Rocket.Chat/issues/22892

renatobecker:
> It seems to be a breaking and not compatible change made in 3.15 version, that prevents creating a room with custom rid, that was possible before this change:
> https://github.com/RocketChat/Rocket.Chat/pull/22742/files
> Can we get review and approval for this?

Hi @tomkos 

Unfortunately, we can't revert the changes because this is by design.
We can't create rooms with external ids, instead, you can rely on custom fields in case you need to tie an Omnichannel-related room and external data.

PR #22893, opened 2021-08-10
[NEW] Webrtc-call Meet Page Design 

Deepak-learner:
I develop the webrtc-call meet page as given in figma Design. When the visitor ends the call, call ended text will appear along with a cross button to close the respective meet Page tab.

Screenshot 2021-08-10 at 9 07 38 PM
Screenshot 2021-08-10 at 10 20 14 PM

### DIFF
--- a/client/views/meet/CallPage.js
+++ b/client/views/meet/CallPage.js
@@ -95,7 +95,6 @@ function CallPage({ roomId, visitorToken, visitorId, status, setStatus, visitorN
 							</Box>
 							<p style={{ color: 'white', fontSize: 15, textAlign: 'center', margin: 15 }}>
 								{'Calling...'}
-								
 							</p>
 							<p style={{ color: 'white', fontSize: 35, textAlign: 'center', margin: 15 }}>
 								{visitorName}
--- a/client/views/meet/CallPage.js
+++ b/client/views/meet/CallPage.js
@@ -4,9 +4,10 @@ import React, { useEffect, useState } from 'react';
 import { Notifications } from '../../../app/notifications/client';
 import { WebRTC } from '../../../app/webrtc/client';
 import { WEB_RTC_EVENTS } from '../../../app/webrtc/index';
+import UserAvatar from '../../components/avatar/UserAvatar';
 import { useTranslation } from '../../contexts/TranslationContext';
 
-function CallPage({ roomId, visitorToken, visitorId, status, setStatus }) {
+function CallPage({ roomId, visitorToken, visitorId, status, setStatus, visitorName, agentName }) {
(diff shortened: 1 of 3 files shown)


djcruz93:
@Deepak-learner timer component is missing i.e. once the call is inProgress the timer should be working. Also, self-video div background color and the close window button color are different. Lastly, `close window` is a label should appear on hover with the same style as it is in the designs.

client/views/meet/CallPage.js
```diff
@@ -64,11 +65,42 @@ function CallPage({ roomId, visitorToken, visitorId, status, setStatus }) {
...
+						<Box
+							position='absolute'
+							zIndex='1'
+							style={{
+								top: '5%',
+								right: '2%',
+							}}
+							w='x200'
+						>
+							<div style={{ border: '1px solid black', padding: '30px' }}>
```
murtaza98, line 87: Why are you using <div> here when we have <Box> component?

client/views/meet/CallPage.js
```diff
@@ -64,11 +65,42 @@ function CallPage({ roomId, visitorToken, visitorId, status, setStatus }) {
...
+								<UserAvatar username={agentName} className='rcx-message__avatar' size='x32' />
+							</div>
+						</Box>
+						<div style={{ marginTop: -180, padding: 20 }}>
+							<Box display='flex' align='center' justifyContent='center'>
+								<UserAvatar username={visitorName} className='rcx-message__avatar' size='x124' />
+							</Box>
+							<p style={{ color: 'white', fontSize: 15, textAlign: 'center', margin: 15 }}>
```
murtaza98, line 95: I think we have a component named `Paragraph` for this in our codebase, so lets try to use it instead

client/views/meet/CallPage.js
```diff
@@ -64,11 +65,42 @@ function CallPage({ roomId, visitorToken, visitorId, status, setStatus }) {
...
+								{'Calling...'}
+							</p>
+							<p style={{ color: 'white', fontSize: 35, textAlign: 'center', margin: 15 }}>
```
murtaza98, line 98: Same goes here

client/views/meet/MeetPage.js
```diff
@@ -1,7 +1,9 @@
+import { Button, Box, Icon, Flex } from '@rocket.chat/fuselage';
```
murtaza98, line 1: Same comment regarding `<div>` and `<p>` applies to this file as well

murtaza98:
@Deepak-learner Could you please take a look at Dhruv and my comments on this PR?

PR #22907, opened 2021-08-11
[BREAK] Removed support of MongoDB 3.4; Deprecated MongoDB 3.6 and 4.0

ostjen:


ankar84:
Nice!
It is high time!

rodrigok (approved):

rockneverdies55:
I think `docker-compose.yml` file should be updated accordingly if mongo dependency is updated.

PR #22914, opened 2021-08-12
[FIX] Threads and discussions searches don't display proper results

matheusbsilva137:

 - _Fix_ issue in discussions search (which wasn't working after a search with no results was made);
 - _Improve_ discussions and threads searches: both searches (`chat.getDiscussions` and `chat.getThreadsList`) are now case insensitive (do NOT differ capital from lower letters) and match incomplete words or terms.

## Issue(s)

[Task - ClickUp](https://app.clickup.com/t/6ewnp5)

## Steps to test or reproduce

**FIX**
Enter a room and open the "Discussions" side bar (using the tab bar icon). Insert a text in order to get _no results_ and then erase the text and make a new search (which should return results).
**Expected behavior:** all the results should be displayed in the second search (after the search with no results).

**IMPROVE**
1. Create a new discussion named "Testing discussions searches" in a new channel;
2. Then, enter the channel, click the "Discussions" button and perform the following searches:
 * 'Z' or 'z' (or any other letter which "Testing discussions searches" doesn't contain): No messages found;
 * 'TESTING', 'testing' or 'Testing': should display the discussion in the list;
 * 'disc', 'searc' or 'sti' (or any other segment of "Testing discussions searches"): should display the discussion in the list;
Similar tests can be applied to threads search.


dougfabris (changes requested):
Could you take a look at the `chat.getDiscussions` endpoint? It seems the text parameter is not working properly.

g-thome (commented):
Backend seems fine to me. But it would be nice if you could add tests for chat.getThreadsList to guarantee that the text argument works as expected

dougfabris (approved):

g-thome (approved):

PR #22922, opened 2021-08-13
Chore: Environmental variable for marketplace url

graywolf336:


sampaiodiego (approved):
does it need to be documented somewhere? otherwise lgtm

graywolf336:
@sampaiodiego it doesn't. This is primarily for internal usage and specifically for Marketplace's staging environment so that we can test everything before we release to production.

PR #22930, opened 2021-08-13
[FIX] Bad words falling if message is empty

ggazzo:
## Issue(s)

In some places like Ominichannel channels creation was failing if the bad words is activated.
`sendMessage(guest, { t: 'livechat-started', msg: '', groupable: false }, room);`


app/lib/server/methods/filterBadWords.ts
```diff
@@ -0,0 +1,47 @@
...
+import Filter from 'bad-words';
+
+import { settings } from '../../../settings/server';
+import { callbacks } from '../../../callbacks/server';
+import { IMessage } from '../../../../definition/IMessage';
+
+
+Meteor.startup(() => {
+	Tracker.autorun(() => {
+		const allowBadWordsFilter = settings.get('AllowBadWordsFilter');
```
sampaiodiego, line 12:
```suggestion
		const allowBadWordsFilter = settings.get('Message_AllowBadWordsFilter');
```

sampaiodiego (changes requested):

Shailesh351:
Reverting back to `bad-words` version `3.0.3` worked for me. 

In the `bad-words` repo some users have reported similar issue for other cases as well. Like adding ":)" or "[" also causes the similar issue.

https://github.com/web-mech/badwords/issues/103 https://github.com/web-mech/badwords/pull/95

So, IMO it's better to use current stable version `"bad-words": "3.0.3",`

sampaiodiego (approved):

PR #22932, opened 2021-08-14
Webrtc meet page layout

Deepak-learner:


murtaza98:
@Deepak-learner There are some conflicts in this file + 2 alerts by LGTM bot. Can you please fix them so we can proceed with the review

murtaza98:
@djcruz93 Could you please help review this PR?

client/views/meet/MeetPage.js
```diff
@@ -14,20 +16,36 @@ function MeetPage() {
 	const roomId = useRouteParameter('rid');
 	const visitorToken = useQueryStringParameter('token');
 	const layout = useQueryStringParameter('layout');
+	const [visitorName, setVisitorName] = useState('');
+	const [agentName, setAgentName] = useState('');
+	const [isAgentRemote, setIsAgentRemote] = useState('false');
```
murtaza98, line 21: Based on my understanding these 2 variables i.e. `isAgentRemote` and `isAgentLocal` seems to be redundant. I think you're using this to check if the user is a visitor or an agent. If so, then I'd recommend either
1. using `visitorToken` token param (it will be set for visitor and empty for agent) (recommended IMO) or 
2. Define only one variable like `isAnAgent`

client/views/meet/OngoingCallDuration.js
```diff
@@ -0,0 +1,14 @@
+import React, { useEffect, useState } from 'react';
+
+export const OngoingCallDuration = () => {
+	const [counter, setCounter] = useState(0);
+	useEffect(() => {
+		setTimeout(() => setCounter(counter + 1), 1000);
+	}, [counter]);
+
+	return (
+		<div style={{ color: '#E4E7EA', fontSize: 15, textAlign: 'center', margin: 15 }}>
```
murtaza98, line 10: Let's use <Box> component instead of vanilla div

client/views/meet/OngoingCallDuration.js
```diff
@@ -0,0 +1,14 @@
+import React, { useEffect, useState } from 'react';
```
murtaza98, line 1: This counter component is not displaying a proper call duration if the page was refreshed by either agent or visitor after joining the call

murtaza98:
Please find the issues I found while testing this PR over [here](https://open.rocket.chat/group/gsoc-2021-livechat-webrtc-integration?msg=yQKWuDou8co37pcrR)

client/views/meet/OngoingCallDuration.js
murtaza98: This has been fixed in recent commits

PR #22934, opened 2021-08-14
[FIX] UI responsiveness

djcruz93:

Adjusts the size of control buttons and the self video component in iframe mode.
image

Adjusts the size of self video container and aspect ratio of the video in mobile view.
image

Adjusts the size of the remote video container when the visitor is on mobile.
image


client/views/meet/CallPage.js
```diff
@@ -81,6 +105,13 @@ function CallPage({ roomId, visitorToken, visitorId, status, setStatus, layout }
 		return window.close();
 	};
 
+	let iconSize = 'x21';
```
murtaza98, line 108: I think this can be a constant

murtaza98:
```suggestion
	const iconSize = layout === 'embedded' ? 'x19' : 'x21';;
```

client/views/meet/CallPage.js
```diff
@@ -81,6 +105,13 @@ function CallPage({ roomId, visitorToken, visitorId, status, setStatus, layout }
...
+	let buttonSize = 'x40';
```
murtaza98, line 109: Same as above

murtaza98 (approved):

PR #22943, opened 2021-08-16
[FIX] Call button visible even when chat is queued

djcruz93:

Call button is not rendered when the chat is queued. It will only show up after an agent takes it up.
image


murtaza98 (approved):

murtaza98:
LGTM!

PR #22949, opened 2021-08-17
[FIX] Avoid /last admin deactivate itself

ostjen:

### DIFF
--- a/server/methods/setUserActiveStatus.js
+++ b/server/methods/setUserActiveStatus.js
@@ -1,6 +1,7 @@
 import { Meteor } from 'meteor/meteor';
 import { check } from 'meteor/check';
 
+import { Users } from '../../app/models/server';
 import { hasPermission } from '../../app/authorization';
 import { setUserActiveStatus } from '../../app/lib/server/functions/setUserActiveStatus';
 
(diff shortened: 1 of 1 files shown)


server/methods/setUserActiveStatus.js
```diff
@@ -1,9 +1,19 @@
 import { Meteor } from 'meteor/meteor';
 import { check } from 'meteor/check';
 
+import { Users } from '../../app/models/server';
 import { hasPermission } from '../../app/authorization';
 import { setUserActiveStatus } from '../../app/lib/server/functions/setUserActiveStatus';
 
+const getAdminCount = () => {
+	Meteor.users.find({
```
KevLehman, line 9: Can we use the Users collection instead of the Meteor.users collection? 👀

server/methods/setUserActiveStatus.js
```diff
@@ -21,8 +31,16 @@ Meteor.methods({
 			});
 		}
 
-		setUserActiveStatus(userId, active, confirmRelenquish);
+		const userAdmin = Users.findOneAdmin(userId.count);
```
KevLehman, line 34: Can this logic be moved to the setUserActiveStatus function? 👀

KevLehman (changes requested):

app/lib/server/functions/setUserActiveStatus.js
```diff
@@ -39,6 +39,15 @@ export function setUserActiveStatus(userId, active, confirmRelinquish = false) {
 		return false;
 	}
 
+	const userAdmin = Users.findOneAdmin(userId.count);
```
KevLehman, line 42: This should only be called if you're actually trying to disable the user, like inside the following if statement

app/lib/server/functions/setUserActiveStatus.js
```diff
@@ -40,8 +40,18 @@ export function setUserActiveStatus(userId, active, confirmRelinquish = false) {
...
 	// Users without username can't do anything, so there is no need to check for owned rooms
 	if (user.username != null && !active) {
+		const userAdmin = Users.findOneAdmin(userId.count);
+		const adminsCount = Users.findActiveUsersInRoles(['admin']).count();
+		if (userAdmin && adminsCount === 1) {
+			throw new Meteor.Error('error-action-not-allowed', 'Leaving the app without an active admin is not allowed', {
+				method: 'removeUserFromRole',
+				action: 'Remove_last_admin',
+			});
+		}
```
geekgonecrazy, line 53: We should add this check to removing role if role is admin.  Some people seem to keep removing the last admin.  🙈

PR #22955, opened 2021-08-17
[NEW][ENTERPRISE] Maximum waiting time for chats in Omnichannel queue

KevLehman:

- Add new settings to support closing chats that have been too long on waiting queue
- Moved old settings to new "Queue Management" section
- Fix issue when closing a livechat room that caused client to not to know if room was open or not


ee/app/livechat-enterprise/server/settings.js
```diff
@@ -115,6 +63,93 @@ export const createSettings = () => {
 				],
 			});
 		});
+
+		this.section('Queue_management', function() {
```
renatobecker, line 67:
```suggestion
		this.section('Queue_Management', function() {
```

Otherwise, translation added [here](https://github.com/RocketChat/Rocket.Chat/pull/22955/files#diff-df6fd124535e1dc52dbd93a30d84f6be0b9b65083f81233b80f30878cfbf9d89R3356) won't be applied

ee/app/livechat-enterprise/server/settings.js
```diff
@@ -2,57 +2,6 @@ import { settings } from '../../../../app/settings';
 import { Settings } from '../../../../app/models/server';
 
 export const createSettings = () => {
-	settings.add('Livechat_waiting_queue', false, {
```
renatobecker, line 5: Do these changes demand a migration or just changing the setting section is enough?

ee/app/livechat-enterprise/server/settings.js
```diff
@@ -115,6 +63,93 @@ export const createSettings = () => {
...
+					i18nLabel: 'Livechat_close_chat',
+				}],
+			});
+
+			this.add('Livechat_max_queue_wait_time', 60, {
+				type: 'int',
+				group: 'Omnichannel',
+				section: 'Queue_management',
+				i18nLabel: 'Livechat_maximum_queue_wait_time',
+				i18nDescription: 'Livechat_maximum_queue_wait_time_description',
```
renatobecker, line 144: Where is the setting description located? I didn't find any translation key for this setting description.
Also, when the `i18nDescription` value ends with `_description` you don't need to implement it, just add the translation key and that's it :)

ee/app/livechat-enterprise/server/settings.js
```diff
@@ -115,6 +63,93 @@ export const createSettings = () => {
...
+					'livechat-enterprise',
+				],
+			});
+
+			this.add('Livechat_max_queue_wait_time_action', 'Nothing', {
+				type: 'select',
+				group: 'Omnichannel',
+				section: 'Queue_management',
+				i18nLabel: 'Livechat_max_queue_wait_time_action',
+				i18nDescription: 'Livechat_max_queue_wait_time_action',
```
renatobecker, line 124: You are displaying the setting label as setting description, is that on purpose?

app/livechat/server/lib/Helper.js
```diff
@@ -86,6 +86,7 @@ export const createLivechatInquiry = ({ rid, name, guest, message, initialStatus
 	const { _id, username, token, department, status = 'online' } = guest;
 	const { msg } = message;
 	const ts = new Date();
+	const maxQueueWaitTimeMinutes = settings.get('Livechat_max_queue_wait_time');
```
renatobecker, line 89: I don't think we should do this.
First, the new setting is available on EE only and here the system is reading it whenever an inquiry is created.
There are callbacks(EE) that can/should handle it properly.

ee/app/livechat-enterprise/server/lib/Helper.js
```diff
@@ -151,6 +154,21 @@ export const updatePredictedVisitorAbandonment = () => {
...
+export const updateQueueInactivityTimeout = () => {
+	const queueAction = settings.get('Livechat_max_queue_wait_time_action');
+	const queueTimeout = settings.get('Livechat_max_queue_wait_time');
+	if (!queueAction || queueAction === 'Nothing') {
+		logger.debug('QueueInactivityTimer: No action performed (disabled by setting)');
+		return LivechatInquiry.unsetEstimatedInactivityCloseTime();
+	}
+
+	logger.debug('QueueInactivityTimer: Updating estimated inactivity time for queued items');
+	LivechatInquiry.getQueuedInquiries().forEach((inq) => {
```
renatobecker, line 166: As far as I know, the [collection.update](https://docs.mongodb.com/v5.0/reference/method/db.collection.update/#db.collection.update) supports aggregation pipeline so you could update all queued inquiries at once, have you thought about it? In that case, you wouldn't need to loop all inquired queues and would execute just one query, something like this:
```
// { "date" : ISODate("2020-04-05T07:14:17.802Z"), "x" : "y" }
db.collection.updateMany(
  { date : { $exists : true } },
  [{ $set: { date: { $add: ["$date", 24*60*60000] } } }]
)
// { "date" : ISODate("2020-04-06T07:14:17.802Z"), "x" : "y" }
```

Think about it.

ee/app/livechat-enterprise/server/startup.js
```diff
@@ -32,5 +34,17 @@ Meteor.startup(async function() {
 			businessHourManager.startManager();
 		}
 	});
+	settings.onload('Livechat_max_queue_wait_time_action', function(_, value) {
+		updateQueueInactivityTimeout();
+		if (!value || value === 'Nothing') {
```
renatobecker, line 39: What if `Livechat_max_queue_wait_time` === 0 ? Shouldn't this be checked too?

ee/app/livechat-enterprise/server/startup.js
```diff
@@ -32,5 +34,17 @@ Meteor.startup(async function() {
...
+			return queueInactivityMonitor.stop();
+		}
+		return queueInactivityMonitor.start();
+	});
+
+	settings.onload('Livechat_max_queue_wait_time', function() {
```
renatobecker, line 45: What if `Livechat_max_queue_wait_time_action` === 0? Should `updateQueueInactivityTimeout` be called?

ee/app/livechat-enterprise/server/lib/QueueInactivityMonitor.js
```diff
@@ -0,0 +1,89 @@
+import { SyncedCron } from 'meteor/littledata:synced-cron';
+import { TAPi18n } from 'meteor/rocketchat:tap-i18n';
+import moment from 'moment';
+
+import { settings } from '../../../../../app/settings/server';
+import { Logger } from '../../../../../app/logger';
+import { LivechatRooms, Users, LivechatInquiry } from '../../../../../app/models/server';
+import { Livechat } from '../../../../../app/livechat/server/lib/Livechat';
+
+export class QueueInactivityMonitor {
```
renatobecker, line 10: Have you talked about using `Agenda` package? 
We introduced something similar [here](https://github.com/RocketChat/Rocket.Chat/blob/develop/ee/app/livechat-enterprise/server/lib/AutoCloseOnHoldScheduler.ts).
The solution you implemented works, I just want to make sure you can rely on other alternatives too(Perhaps with less complexity).

renatobecker (changes requested):

ee/app/livechat-enterprise/server/settings.js
KevLehman: Fixed in next commit

ee/app/livechat-enterprise/server/settings.js
KevLehman: I tested locally, believing a migration would be needed, but seems like thoseind of updates are done on startup, so no need to migrate

ee/app/livechat-enterprise/server/settings.js
KevLehman: Thanks! Yeah, this was something i forgot to remove while copying setting definition. I removed it.

ee/app/livechat-enterprise/server/settings.js
KevLehman: Nope, I removed it 😬

ee/app/livechat-enterprise/server/lib/Helper.js
KevLehman: Thanks for reminding me of this! Modified to use update pipes

ee/app/livechat-enterprise/server/hooks/afterReturnRoomAsInquiry.ts
```diff
@@ -1,19 +1,44 @@
...
+	callbacks.add('livechat:afterReturnRoomAsInquiry', unsetPredictedVisitorAbandonment, callbacks.priority.HIGH, 'livechat-after-return-room-as-inquiry');
+});
+
+settings.get('Livechat_max_queue_wait_time_action', (_, value) => {
+	if (!value || value === 'Nothing') {
+		callbacks.remove('livechat:afterReturnRoomAsInquiry', 'livechat-after-return-room-as-inquiry-set-queue-timer');
+		return;
+	}
+	const timer = settings.get('Livechat_max_queue_wait_time');
+	callbacks.add('livechat:afterReturnRoomAsInquiry', setQueueTimer(timer), callbacks.priority.HIGH, 'livechat-after-return-room-as-inquiry-set-queue-timer');
```
renatobecker, line 43: We can't rely only on this callback to update the `estimated Inactivity close time` because this callback is only called when the routing system algorithm is `Manual Selection`, but Inquiries are queued in all algorithms, so we need to rely on a more consist solution that will cover all use cases.

So, we need to track this method -> `LivechatInquiry.queueInquiry`. As far as I know, the method is called twice in the code, so rather than calling `LivechatInquiry.queueInquiry`, we could maybe create a new method that would be called and inside the new method you'd call -> `LivechatInquiry.queueInquiry` and, also, run a new callback method, so rather than implementing this logic here, you could implement the logic for the new callback I just mentioned, something like:  `livechat:afterQueueInquiry`, WDYT?

You could, even, improve the signature of the method `Livechat.queueInquiry` and support additional data, so you could also pass along the new value of `estimatedInactivityCloseTimeAt` and update the document at once.

renatobecker (changes requested):

app/models/server/models/LivechatInquiry.js
```diff
@@ -228,6 +232,48 @@ export class LivechatInquiry extends Base {
 
 		this.remove(query);
 	}
+
+	getAbandonedQueuedItems(date) {
```
renatobecker, line 236: IMO we should rename this method since we are not just talking about abandoned chats because the end-user may not have abandoned, actually the end-user was there waiting for assistance that never came.

ee/app/livechat-enterprise/server/lib/QueueInactivityMonitor.ts
```diff
@@ -0,0 +1,100 @@
...
+		let counter = 0;
+		if (!action || action === 'Nothing') {
+			return;
+		}
+
+		LivechatInquiry.getAbandonedQueuedItems(moment().utc()).forEach((inquiry: any) => {
+			switch (action) {
+				case 'Close_chat': {
+					counter++;
+					this.closeRooms(LivechatRooms.findOneByIdOrName(inquiry.rid));
```
renatobecker, line 86: Why not `LivechatRooms.findOneById` ?

ee/app/livechat-enterprise/server/lib/QueueInactivityMonitor.ts
```diff
@@ -0,0 +1,100 @@
...
+		this.running = false;
+		this._name = 'Omnichannel-Queue-Inactivity-Monitor';
+		this.logger = new Logger('QueueInactivityMonitor');
+		this.scheduler = new Agenda({
+			mongo: (MongoInternals.defaultRemoteCollectionDriver().mongo as any).client.db(),
+			db: { collection: SCHEDULER_NAME },
+			defaultConcurrency: 1,
+		});
+		this.user = Users.findOneById('rocket.cat');
+		this.message = TAPi18n.__('Closed_automatically_chat_queued_too_long');
```
renatobecker, line 39: Is there a way to select a global language to translate the message? Perhaps a server language or something. This way, English will always be the language used here.

KevLehman: Yup, we can get the value for the language setting.

ee/app/livechat-enterprise/server/lib/Helper.js
```diff
@@ -151,6 +154,21 @@ export const updatePredictedVisitorAbandonment = () => {
 	}
 };
 
+export const updateQueueInactivityTimeout = () => {
+	const queueAction = settings.get('Livechat_max_queue_wait_time_action');
```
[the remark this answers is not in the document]
KevLehman: Yup, I thought about that but in this specific case we can't, since that function is called on the changes of 2 different, dependant, settings. So, instead of having to call the function like ths: `updateQueueInactivityTimeout(undefined, valueOfUpdatedSetting) or with (valueOfUpdatedSetting, undefined)` depending on which setting was updated, I preferred to fetch the value on each call. Anyways, this should only be called when the setting is updated

ee/app/livechat-enterprise/server/hooks/afterReturnRoomAsInquiry.ts
KevLehman: Hey but this callback is called only on `returnRoomAsInquiry` method, this callback should just set the value again when the user returns the chat to the queue, and as far as i can see on `app/livechat/server/lib/Livechat.js L663` this should happen disregards of the routing algorithm.

The callback that sets the value when the chat first comes is on the `createLivechatInquiry` method (using the `livechat.beforeInquiry` cb) and that should also run on all routing algos, since it happens before queueing the inquiry.

app/models/server/models/LivechatInquiry.js
KevLehman: Yup, agreed with this.

ee/app/livechat-enterprise/server/lib/QueueInactivityMonitor.ts
KevLehman: 😬 I didn't find that method at first

ee/app/livechat-enterprise/server/hooks/afterReturnRoomAsInquiry.ts
renatobecker: Regarding `returnRoomAsInquiry`:
- The method is called when a user returns the chat to the queue manually(Only available on Manual Selection)
- The method is called when the agent didn't reply and the chat is auto-transferred, however, there's a condition: `RoutingManager.getConfig().autoAssignAgent` ==== `Manual Selection`.

Apart from that, the chat can be queued in other places, such as here:

https://github.com/RocketChat/Rocket.Chat/blob/6267bf4b6d726f7a3709c1a7ed79f7ed4243d44c/app/livechat/server/lib/Helper.js#L409-L412

So, as you can see, the chat will be queued, but the `estimatedInactivityCloseTimeAt` won't be set, that's the reason we should implement a more centralized solution that would cover all cases and would reduce complexity.

KevLehman: That's interesting to know. What about using the `livechat.beforeRouteChat` to set that value then? (or create another callback def at the same place as this) to avoid having to modify multiple files by changing the function signature?

renatobecker: IMHO we should wrap the method `LivechatInquiry.queueInquiry`(called twice in the codebase) in a new method that, as the last step, would call a new callback, perhaps -> `livechat:afterQueueInquired` and, then, implement the logic for the new callback.
In this case, we wouldn't need to implement callbacks for `livechat.beforeInquiry` nor `livechat:afterReturnRoomAsInquiry`, just one single callback method would do the trick.

renatobecker (approved):

PR #22972, opened 2021-08-19
Chore: Prevent new JS files being added

ggazzo:

We are moving our code base to TS, one way to help developers remember this is create a task that will notify you every time a new file is created.


KevLehman (approved):
Haven't tested on the repo, but I like this 🥇

sampaiodiego:
looks like the action didn't run :thinking:

PR #22976, opened 2021-08-19
[FIX] Error getting default agent when routing system algorithm is Auto Selection

renatobecker:
When the Omnichannel routing system algorithm is **Auto Selection** (Community version) an error is raised when getting the default agent. The cause of the issue was a wrong implementation of a callback method.
Another issue has been found and fixed: Guests associated with a department that no longer exists were not routed properly since the routing system was trying to find available agents associated with a department that doesn't exist anymore.


KevLehman (dismissed):
Tried on community, autoselection algorithm is no longer returning errors. However, there's still an issue with the routing itself (that afaik, won't be handled in this same PR)

app/livechat/server/hooks/beforeDelegateAgent.js
```diff
@@ -3,9 +3,7 @@ import { callbacks } from '../../../callbacks';
 import { settings } from '../../../settings';
 import { Users, LivechatDepartmentAgents } from '../../../models';
 
-callbacks.add('livechat.beforeDelegateAgent', (options = {}) => {
-	const { department, agent } = options;
-
+callbacks.add('livechat.beforeDelegateAgent', (agent, department) => {
```
ggazzo, line 6: I think it's best to keep using a key value object, if you need another different item, it won't break everywhere. This change actually doenst fix or solves anything
```suggestion
callbacks.add('livechat.beforeDelegateAgent', (agent, { department }) => {
```

app/livechat/server/lib/RoutingManager.js
```diff
@@ -162,7 +162,7 @@ export const RoutingManager = {
 	},
 
 	delegateAgent(agent, inquiry) {
-		const defaultAgent = callbacks.run('livechat.beforeDelegateAgent', { agent, department: inquiry?.department });
+		const defaultAgent = callbacks.run('livechat.beforeDelegateAgent', agent, inquiry?.department);
```
ggazzo, line 165:
```suggestion
		const defaultAgent = callbacks.run('livechat.beforeDelegateAgent', agent, { department: inquiry?.department });
```

ggazzo (changes requested):

renatobecker:
Suggestions applied.

KevLehman (dismissed):

ggazzo (approved):

PR #22985, opened 2021-08-20
[NEW][APPS] Get livechat's room transcript via bridge method

thassiov:

Adds a new method for retrieving a room's transcript via a new method in the Livechat bridge

## Further comments

Depends on [Apps-engine/pull/436](https://github.com/RocketChat/Rocket.Chat.Apps-engine/pull/436)


app/apps/server/bridges/livechat.ts
```diff
@@ -222,6 +223,20 @@ export class AppLivechatBridge extends LivechatBridge {
...
+	protected async _fetchLivechatRoomMessages(appId: string, roomId: string): Promise<Array<IMessage>> {
+		this.orch.debugLog(`The App ${ appId } is getting the transcript for livechat room ${ roomId }.`);
+		const messages = Livechat.getRoomMessages({ rid: roomId });
+
+		const result = [];
+
+		for await (const message of messages) {
+			const res = await this.orch.getConverters()?.get('messages').convertMessage(message);
+			result.push(res);
+		}
```
d-gubert, line 235: I would suggest this be changed into a `Promise.allSettled()` call, allowing the calls to `convertMessage` to be run concurrently

d-gubert (changes requested):

app/livechat/server/lib/Livechat.js
```diff
@@ -1097,6 +1097,25 @@ export const Livechat = {
 		return true;
 	},
 
+	getRoomMessages({ rid }) {
+		check(rid, String);
+
+		const isLivechat = Promise.await(Rooms.findByTypeInIds('l', [rid])).count();
+
+		if (!isLivechat) {
+			return [];
```
d-gubert, line 1106: I'm wondering if we should throw an error in this case 🤔

app/livechat/server/lib/Livechat.js
```diff
@@ -1097,6 +1097,25 @@ export const Livechat = {
...
+		}
+
+		const ignoredMessageTypes = ['livechat_navigation_history', 'livechat_transcript_history', 'command', 'livechat-close', 'livechat-started', 'livechat_video_call'];
+		const messageCursor = Promise.await(Messages.findVisibleByRoomIdNotContainingTypes(rid, ignoredMessageTypes, { sort: { ts: 1 } }));
```
murtaza98, line 1110:
```suggestion
		const messageCursor = Promise.await(Messages.findVisibleByRoomIdNotContainingTypes(rid, ignoredMessageTypes, { sort: { ts: 1 } })).toArray();
```

murtaza98: I think we can convert a cursor to array like this. This way we don't have to loop through the cursor and manually add each message to an empty array

app/livechat/server/lib/Livechat.js
```diff
@@ -1097,6 +1097,25 @@ export const Livechat = {
...
+
+		const messages = [];
```
murtaza98, line 1112:
```suggestion
```

app/livechat/server/lib/Livechat.js
```diff
@@ -1097,6 +1097,25 @@ export const Livechat = {
...
+
+		messageCursor.forEach((message) => messages.push(message));
```
murtaza98, line 1114:
```suggestion
```

app/livechat/server/lib/Livechat.js
murtaza98: This won't be required if we decide to apply my earlier suggestion

app/livechat/server/lib/Livechat.js
murtaza98: Same as above

app/livechat/server/lib/Livechat.js
```diff
@@ -1097,6 +1097,25 @@ export const Livechat = {
...
+	getRoomMessages({ rid }) {
+		check(rid, String);
+
+		const isLivechat = Promise.await(Rooms.findByTypeInIds('l', [rid])).count();
+
+		if (!isLivechat) {
+			return [];
+		}
+
+		const ignoredMessageTypes = ['livechat_navigation_history', 'livechat_transcript_history', 'command', 'livechat-close', 'livechat-started', 'livechat_video_call'];
```
KevLehman, line 1109: Should we include `user joining/leaving` message types? I guess those ones are `uj` or `ul` 🤔

app/apps/server/bridges/livechat.ts
```diff
@@ -222,6 +223,14 @@ export class AppLivechatBridge extends LivechatBridge {
 		return LivechatDepartment.findEnabledWithAgents().map(boundConverter);
 	}
 
+	protected async _fetchLivechatRoomMessages(appId: string, roomId: string): Promise<Array<IMessage>> {
+		this.orch.debugLog(`The App ${ appId } is getting the transcript for livechat room ${ roomId }.`);
+		const messages = Livechat.getRoomMessages({ rid: roomId });
+		const messageConverter = this.orch.getConverters()?.get('messages');
+		const boundMessageConverter = messageConverter.convertMessage.bind(messageConverter);
```
KevLehman, line 230: `messageConverter` may return undefined on line above, calling this would cause an error 👀

KevLehman (changes requested):

app/apps/server/bridges/livechat.ts
d-gubert: Typescript should have caught that, right? 🤔

KevLehman: That's true, TS should have caught it 👀 weird why it didn't

app/livechat/server/lib/Livechat.js
thassiov: It does make sense, so I'm throwing the 'invalid-room' error here

app/livechat/server/lib/Livechat.js
thassiov: The thing is this method call returns a cursor that does not have a `toArray` method included, unfortunately. That's why I'm looping through the query's contents.

The error:
```
TypeError: cursor.toArray is not a function                                                                                                                                                                                               
     at Object.getRoomMessages (app/livechat/server/lib/Livechat.js:1055:27)
```

KevLehman: I believe you don't need to call toArray here. Would be enough to call `.fetch` on the returned cursor, that will return an array

app/apps/server/bridges/livechat.ts
thassiov: Because of the optional chaining there TS said it was ok, but I'm throwing an error if this expression comes as `undefined` now

app/apps/server/bridges/livechat.ts
```diff
@@ -222,6 +223,14 @@ export class AppLivechatBridge extends LivechatBridge {
...
+		return await Promise.all(messages.map(boundMessageConverter)) as Array<IMessage>;
```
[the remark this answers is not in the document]
thassiov: Didn't need the `Promise.all` either. I removed it and made it work just like the other methods. The thing with this `await` was that, although is was working file without it, the linter was saying that the expression needed to be converted to `unknown`. Now it is not an issue anymore

app/livechat/server/lib/Livechat.js
thassiov: Didn't know about fetch. I'll try it

app/livechat/server/lib/Livechat.js
thassiov: If this data is not relevant, I can add them to the list

KevLehman: For now, I don't think it would affect 🤔 so let's leave it with the current list

KevLehman (approved):

PR #22987, opened 2021-08-20
[NEW] TaskRoom - GSOC

Jeanstaquet:
This project introduces a new type of room, the Task Room. In which you can create tasks. The task room uses a new renderer, React. 
Furthermore the tasks are stored in a new model in the DB, and in a new Front End Collection.


Jeanstaquet:
The PR was moved to #23003

PR #22991, opened 2021-08-20
[FIX] Fix autoselection algorithm on community

KevLehman:

- When using the autoselection algo on community editions, all agents were marked as unavailable due to an unapplied filter
- Fixed an issue when both user & system setting to manange EE max number of chats allowed were set to 0

### DIFF
--- a/ee/app/livechat-enterprise/server/hooks/applySimultaneousChatsRestrictions.ts
+++ b/ee/app/livechat-enterprise/server/hooks/applySimultaneousChatsRestrictions.ts
@@ -29,5 +29,18 @@ callbacks.add('livechat.applySimultaneousChatRestrictions', (_: any, { departmen
 		}
 		: {};
 
-	return { $match: { $or: [agentFilter, globalFilter] } };
+	return [
+		// we need to use this value to know when to apply global/agent filters
+		{ $addFields: { maxChatsPerSetting } },
+		{ $match: {
+			$and: [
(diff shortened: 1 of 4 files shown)


ee/app/livechat-enterprise/server/hooks/applySimultaneousChatsRestrictions.ts
```diff
@@ -29,5 +29,18 @@ callbacks.add('livechat.applySimultaneousChatRestrictions', (_: any, { departmen
...
 
-	return { $match: { $or: [agentFilter, globalFilter] } };
+	return [
+		// we need to use this value to know when to apply global/agent filters
+		{ $addFields: { maxChatsPerSetting } },
+		{ $match: {
+			$and: [
+				{ $or: [agentFilter, globalFilter] },
+				// when both settings are set at 0 means there's no limit, so agent is "available"
+				{ $and: [
```
renatobecker, line 39: We can't just apply `$and` condition here. Admin can set just the global setting, for instance, and the `$and` condition here will not work in this use case. I replaced `$and` with `$or` and then it worked, but then the setting per agent doesn't work, the agent can take as many chats he/she wants without restriction.
Please, test the following:
- Set just the global setting and replace `$and` with `$or`, the `global` setting won't work.
- Then set the agent limit, the feature doesn't work anymore.

renatobecker (changes requested):

PR #22994, opened 2021-08-20
Chore: Enable husky pre-push hook (back again)

sampaiodiego:

After recently husky lib update the pre-push hook was no longer working, this will make it work again.


d-gubert:
GIMME

ggazzo (approved):

PR #22998, opened 2021-08-21
[FIX] Send to channel functionality in threads/discussions 

### LINKED ISSUE #19167: "Send to Channel" send automatically disables checkbox
### Description:

Also Send to Channel tick box in thread does not persist 

### Steps to reproduce:

Set Send to Channel globally
Administration -> Accounts -> Default User Preferences -> Show thread messages in the main thread

Set Send to Channel individually
Account -> Preferences -> Show thread messages in the main thread

Reply to create a new thread. 'Also Send to Channel' is ticked.

Close thread box.

Open Thread box.

'Also Send to Channel' not ticked.

Tick box and send and the box unchecks again.

### Expected behavior:

If the setting is enable the 'Also Send To Channel' tick box should be checked at all times unless the user deselects it.

### Actual behavior:

When you close and the reopen the thread box the 'Also Send To Channel' is not checked.

Check the box and send and it unchecks again.

It does appear that messages are sent to the thread.

However, this is confusing a lot of users.

### Server Setup Information:

- Version of Rocket.Chat Server: open.rocket.chat - probably 3.7.x or 3.8.dev + tested on 3.5.4 
- Operating System: Yours
- Deployment Method: unknown
- Number of Running Instances: multiple 
- DB Replicaset Oplog: N/A
- NodeJS Version: N/A
- MongoDB Version: N/A

### Client Setup Information

- Desktop App or Browser Version: Chromium
- Operating System: Xubuntu 18.04

### Additional context

None

### Relevant logs:

None

Probably related:

https://github.com/RocketChat/Rocket.Chat/issues/18347
https://github.com/RocketChat/Rocket.Chat/pull/17416

chetas411:
https://user-images.githubusercontent.com/65273718/130312202-43daaa5e-f79a-4f41-8d42-c209a6ccd59b.mp4

## Issue(s)

 This PR fixes #19167. Now the checkbox for "Also Send to Channel" gets ticked already as per the user preference settings and it persists even after sending the message. Also the user can check and uncheck it and the messages will be sent accordingly to the main channel.


chetas411:
@dougfabris  can you review ?

chetas411:
@sampaiodiego @renatobecker @ggazzo @dougfabris can you approve the workflows to run as I am first time contributing to Rocket.chat and review the PR?

sampaiodiego:
thanks @chetas411 for you contribution.. I'll let the workflows run, but I'm not sure this is a good fix, not even if this is an issue at all..

I think the preference to "always show thread messages in channel" has no relation with the check box "also send to channel".. but I'll explain this better on the issue

chetas411:
> thanks @chetas411 for you contribution.. I'll let the workflows run, but I'm not sure this is a good fix, not even if this is an issue at all..
> 
> I think the preference to "always show thread messages in channel" has no relation with the check box "also send to channel".. but I'll explain this better on the issue

here in this issue  #19167, the checkbox does not persist the check and resets as soon as the message is sent in thread. Also there is no effect of checking the checkbox or uncheck it as the messages are sent to the main channel always if the preference to show messages in channel is set.

chetas411:
> > thanks @chetas411 for you contribution.. I'll let the workflows run, but I'm not sure this is a good fix, not even if this is an issue at all..
> > I think the preference to "always show thread messages in channel" has no relation with the check box "also send to channel".. but I'll explain this better on the issue
> 
> here in this issue #19167, the checkbox does not persist the check and resets as soon as the message is sent in thread. Also there is no effect of checking the checkbox or uncheck it as the messages are sent to the main channel always if the preference to "show thread messages in channel" is set.

@sampaiodiego Can you update  me on this ?

dougfabris:
@chetas411 Agreed with @sampaiodiego here. There's no relation between `Show thread messages in the main thread` preference and `Also send to channel` option. But if the user has the preference turned on doesn't make sense to keep `Also send to channel` displayed. So the correct fix here is to hide the `Also send to channel` option when the preference is turned on, since it's going to send to the channel anyway. Are you available to work on that?

dougfabris:
Hey @chetas411 
We decided to move forward with another approach, presented here #26002, thanks for your contribution!

PR #23008, opened 2021-08-23
[NEW] additional settings in bbb videoconference

JensBr1986:
add additional settings for bigbluebutton videoconference:

- selectable property of user for displayed user label in bbb-client (username, user, nickname)
- custom parameter for bbb-client on join (only start with 'userdata-bbb', see https://docs.bigbluebutton.org/admin/customize.html#passing-custom-parameters-to-the-client-on-join)
- automatic notification at the start of the meeting


dougfabris:
Since our latest major we moved the BBB from the core to an app. The app already has: 
- selectable property of user for displayed user label in bbb-client (username, user, nickname)
- automatic notification at the start of the meeting

We've implemented your suggestion for the custom parameters, you can check it [here](https://github.com/RocketChat/Apps.BigBlueButton/pull/7)
Thanks a lot for helping us!

PR #23032, opened 2021-08-25
[FIX] User list not being updated after creation/deletion of user

ostjen:
This PR aims to fix the behavior of users when they're deleted in admin/users by creating a specific callback for deleting users and reloading the user list afterwards. 

It also replicates this reload-after-done behavior on the creation of users.

**before**

https://user-images.githubusercontent.com/28611993/130843279-5c274bc2-42f5-4def-8f47-7929a23df3de.mov

**after**

https://user-images.githubusercontent.com/28611993/130842679-6a4f647f-b976-4f7d-a80e-87cee05b632a.mov


client/views/admin/users/AddUser.js
```diff
@@ -17,6 +17,8 @@ export function AddUser({ roles, ...props }) {
 	const { value: roleData } = useEndpointData('roles.list', '');
 	const [errors, setErrors] = useState({});
 
+	const { reloadTable, ...otherProps } = props;
```
tassoevan, line 20: I believe it's safe to do this destructuring at the function parameter declaration, i.e.:

```
export function AddUser({ roles, reloadTable, ...props }) {
```

I don't think `props` requires `reloadTable` property anywhere else.

client/views/admin/users/UsersPage.js
```diff
@@ -27,7 +29,49 @@ function UsersPage() {
...
+					sortFields.reduce((agg, [column, direction]) => {
+						agg[column] = sortDir(direction);
+						return agg;
+					}, {}),
+				),
+				...(itemsPerPage && { count: itemsPerPage }),
+				...(current && { offset: current }),
+			}),
+			[text, itemsPerPage, current, sortFields],
+		);
```
tassoevan, line 63: Those two function will be recreated each time `UsersPage` is re-rendered. Better move them to the parent scope.

client/views/admin/users/EditUser.js
```diff
@@ -29,6 +29,8 @@ const getInitialValue = (data) => ({
 function EditUser({ data, roles, ...props }) {
 	const t = useTranslation();
 
+	const { reloadTable, ...otherProps } = props;
```
tassoevan, line 32: I believe it's safe to do this destructuring at the function parameter declaration, i.e.:

```
export function AddUser({ roles, reloadTable, ...props }) {
```

I don't think `props` requires `reloadTable` property anywhere else.

client/views/admin/users/UserInfo.js
```diff
@@ -19,6 +19,8 @@ export function UserInfoWithData({ uid, username, ...props }) {
 	const showRealNames = useSetting('UI_Use_Real_Name');
 	const approveManuallyUsers = useSetting('Accounts_ManuallyApproveNewUsers');
 
+	const { reloadTable, ...otherProps } = props;
```
tassoevan, line 22: I believe it's safe to do this destructuring at the function parameter declaration, i.e.:

```
export function AddUser({ roles, reloadTable, ...props }) {
```

I don't think `props` requires `reloadTable` property anywhere else.

tassoevan (changes requested):

tassoevan (approved):

PR #23037, opened 2021-08-26
[FIX] "Read Only" and "Allow Reacting" system messages are missing in rooms

matheusbsilva137:

 - Add system message to notify changes on the **"Read Only"** setting;
 - Add system message to notify changes on the **"Allow Reacting"** setting;
 - Fix "Allow Reacting" setting's description (updated from "Only authorized users can write new messages" to "Only authorized users can react to messages").
system-messages

## Issue(s)

[Task - ClickUp](https://app.clickup.com/t/5k5qvt)

## Steps to test or reproduce

1. Create a new channel (not read-only);
2. Go to **Room Information > Edit > Read Only** and turn this setting on. **_Expected Behavior:_** a system message with the text "Room set as Read Only by _your.user_" should be sent;
3. Turn the "Read Only" setting off. **_Expected Behavior:_** a system message with the text "Room added writing permission by _your.user_" should be sent;
4. Turn the "Read Only" setting on again and then turn the "Allow Reacting" setting on. **_Expected Behavior:_** a system message with the text "Room allowed reacting by matheus.barbosa by _your.user_" should be sent;
5. Turn the "Allow Reacting" setting off. **_Expected Behavior:_** a system message with the text "Room disallowed reacting by _your.user_" should be sent;
**IMPORTANT:** Save the changes after changing each setting (the system messages will only be sent _after_ saving).


app/channel-settings/server/functions/saveReactWhenReadOnly.js
```diff
@@ -1,12 +1,21 @@
 import { Meteor } from 'meteor/meteor';
 import { Match } from 'meteor/check';
 
-import { Rooms } from '../../../models';
+import { Rooms, Messages } from '../../../models';
 
-export const saveReactWhenReadOnly = function(rid, allowReact) {
+export const saveReactWhenReadOnly = function(rid, allowReact, user, sendMessage = true) {
```
ostjen, line 6: why you need this `sendMessage`?

app/channel-settings/server/functions/saveReactWhenReadOnly.js
```diff
@@ -1,12 +1,21 @@
...
 	}
 
-	return Rooms.setAllowReactingWhenReadOnlyById(rid, allowReact);
+	const result = Rooms.setAllowReactingWhenReadOnlyById(rid, allowReact);
+
+	if (result && sendMessage) {
+		if (allowReact) {
+			Messages.createRoomAllowedReactingByRoomIdAndUser(rid, user);
+		} else {
+			Messages.createRoomDisallowedReactingByRoomIdAndUser(rid, user);
```
ostjen, line 17: wouldn't be cleaner to use a ternary operator?  at least it would avoid nesting, but idk

```
allowReact ? Messages.createRoomAllowedReactingByRoomIdAndUser(rid, user)
	   : Messages.createRoomDisallowedReactingByRoomIdAndUser(rid, user);
```

app/channel-settings/server/functions/saveRoomReadOnly.js
```diff
@@ -1,14 +1,23 @@
...
 	}
-	return Rooms.setReadOnlyById(rid, readOnly, hasPermission);
+	const result = Rooms.setReadOnlyById(rid, readOnly, hasPermission);
+
+	if (result && sendMessage) {
+		if (!readOnly) {
+			Messages.createRoomSetReadOnlyByRoomIdAndUser(rid, user);
+		} else {
+			Messages.createRoomRemovedReadOnlyByRoomIdAndUser(rid, user);
+		}
```
ostjen, line 20: same ternary op comment

app/channel-settings/server/functions/saveReactWhenReadOnly.js
matheusbsilva137: Some other similar functions (eg `saveRoomAnnouncement`, `saveRoomType`) in this directory use this parameter, so I decided to include it here too. Anyways, if we think it'd be better to remove it, I can do it 😄

ostjen: hmmmm I think it's worth checking if it makes any difference in this case, if it doesn't I think it's better to just leave as it is

matheusbsilva137: In the other functions that contain the `sendMessage` parameter, it is always omitted from their calls. So, I think this isn't relevant at all 🤷‍♂️ . Guess we can just leave it as it is, then

ostjen (dismissed):

ostjen (dismissed):

ostjen (approved):

PR #23050, opened 2021-08-27
[NEW] Prevent activating users beyond the license cap

pierre-lehnen-rc:
This change prevents admins from accidentally deactivating their enterprise license by adding more users than the license permits.

Blocking other role syncing options will be done in a separate PR.


app/dolphin/lib/common.js
```diff
@@ -25,7 +25,7 @@ function DolphinOnCreateUser(options, user) {
 	if (user && user.services && user.services.dolphin && user.services.dolphin.NickName) {
 		user.username = user.services.dolphin.NickName;
 	}
-	return user;
+	return options;
```
tassoevan, line 28: 🤔

PR #23055, opened 2021-08-27
[NEW] Banner for the updates regarding authentication services

g-thome:

Add a banner to inform admins about future authentication changes. This banner targets servers that use some sort of authentication service since they're the ones which this update concerns the most.

### DIFF
--- a/app/models/server/models/Settings.js
+++ b/app/models/server/models/Settings.js
@@ -242,6 +242,13 @@ export class Settings extends Base {
 			}
 		}
 	}
+
+	isOAuthEnabled() {
+		return !!this.findOne({
+			_id: { $regex: /^Accounts_OAuth_/ },
+			value: true,
+		});
+	}
 }
 
 export default new Settings('settings', true);
--- a/app/models/server/models/Settings.js
+++ b/app/models/server/models/Settings.js
@@ -242,13 +242,6 @@ export class Settings extends Base {
 			}
 		}
 	}
-
-	isOAuthEnabled() {
-		return !!this.findOne({
-			_id: { $regex: /^Accounts_OAuth_/ },
-			value: true,
-		});
-	}
 }
 
 export default new Settings('settings', true);
--- a/server/startup/migrations/v231.ts
+++ b/server/startup/migrations/v231.ts
@@ -25,7 +25,7 @@ Migrations.add({
 			blockId: 'attention',
 			text: {
 				type: TextObjectType.PLAINTEXT,
-				text: 'Please notice that after the next release (4.0) advanced functionalities of LDAP, SMAL, and Custom Oauth will be available just in Enterprise Edition and Gold plans. Click here for more info: http://www.example.com',
+				text: 'Please notice that after the next release (4.0) advanced functionalities of LDAP, SMAL, and Custom Oauth will be available just in Enterprise Edition and Gold plans. Click here for more info: https://forums.rocket.chat/t/upcoming-changes-to-identity-management-integrations/11994',
 				emoji: false,
 			},
 		};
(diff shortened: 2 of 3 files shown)


server/startup/migrations/v231.ts
```diff
@@ -0,0 +1,53 @@
...
+		if (!isAuthServiceEnabled) {
+			return;
+		}
+
+		const s: ISectionBlock = {
+			type: BlockType.SECTION,
+			blockId: 'attention',
+			text: {
+				type: TextObjectType.PLAINTEXT,
+				text: 'Please notice that after the next release (4.0) advanced functionalities of LDAP, SMAL, and Custom Oauth will be available just in Enterprise Edition and Gold plans. Click here for more info: https://go.rocket.chat/i/authentication-changes',
```
sampaiodiego, line 28:
```suggestion
				text: 'Please notice that after the next release (4.0) advanced functionalities of LDAP, SAML, and Custom Oauth will be available only on Enterprise Edition and Gold plan. Click here for more info: https://go.rocket.chat/i/authentication-changes',
```

server/startup/migrations/v231.ts
```diff
@@ -0,0 +1,53 @@
...
+				emoji: false,
+			},
+		};
+
+		const LDAPBanner = {
+			platform: [BannerPlatform.Web],
+			createdAt: new Date(),
+			expireAt: new Date('2021-09-27'),
```
sampaiodiego, line 36: maybe change to expire only so we're sure people will see it?
```suggestion
			expireAt: new Date('2022-09-27'),
```

server/startup/migrations/v231.ts
```diff
@@ -0,0 +1,53 @@
...
+import { Banner } from '../../sdk';
+import { settings } from '../../../app/settings/server';
+import { Settings } from '../../../app/models/server';
+
+Migrations.add({
+	version: 231,
+	up() {
+		const LDAPEnabled = settings.get('LDAP_Enable');
+		const SAMLEnabled = settings.get('SAML_Custom_Default');
+		const CustomOauthEnabled = Settings.isOAuthEnabled();
```
sampaiodiego, line 15: if this query will be used only on this migration, I think you write the query directly, instead perpetuating it on the model class.

app/models/server/models/Settings.js
```diff
@@ -242,6 +242,13 @@ export class Settings extends Base {
 			}
 		}
 	}
+
+	isOAuthEnabled() {
+		return !!this.findOne({
+			_id: { $regex: /^Accounts_OAuth_/ },
```
sampaiodiego, line 248:
```suggestion
			_id: /^Accounts_OAuth_/,
```

sampaiodiego (changes requested):

app/models/server/models/Settings.js
pierre-lehnen-rc, line 248: This will return true for any OAuth related settings, not just the "enabled" settings. For example, `Accounts_OAuth_Gitlab_merge_users` is boolean and may be turned on even if the gitlab OAuth is off.

pierre-lehnen-rc (changes requested):

server/startup/migrations/v231.ts
```diff
@@ -0,0 +1,58 @@
...
+import { Settings } from '../../../app/models/server';
+
+Migrations.add({
+	version: 231,
+	up() {
+		const LDAPEnabled = settings.get('LDAP_Enable');
+		const SAMLEnabled = settings.get('SAML_Custom_Default');
+
+		const query = {
+			_id: { $regex: /^Accounts_OAuth_/ },
```
sampaiodiego, line 17:
```suggestion
			_id: /^Accounts_OAuth_/,
```

sampaiodiego (changes requested):

app/models/server/models/Settings.js
g-thome: updated the rex to only match against the servicce enabled documents

pierre-lehnen-rc (approved):

sampaiodiego (approved):

PR #23076, opened 2021-08-31
[FIX] "Parent channel or group" search in discussions' creation throws "Unexpected end of JSON input" error

matheusbsilva137:

- Use `encodeURIComponent()` to encode values received by `_generateQueryFromParams()`.

## Issue(s)

[Task - ClickUp](https://app.clickup.com/t/5q8naj)

## Steps to test or reproduce

Open **Create New Discussion** and type "#" in the "**Parent channel or group**" field. **_Expected behavior:_** no error should be displayed (and the search's result should be empty).

## Further comments

The JSON input would end before expected when using the "#" char because this could be interpreted as the URL's **fragment** part.


tassoevan (changes requested):
It looks like theres a lot more to it than meets the eye. A private method on `APIClient` (`_generateQueryFromParams` to be more specific) is not encoding the values via `encodeURIComponent()`. so the `#` character is recognized as the start of the fragment part of the REST API's endpoint URL.

tassoevan (approved):

PR #23089, opened 2021-09-01
[FIX] Change way emails are validated on livechat registerGuest method

KevLehman:


app/livechat/server/lib/Helper.js
```diff
@@ -535,3 +535,11 @@ export const updateDepartmentAgents = (departmentId, agents, departmentEnabled)
 
 	return true;
 };
+
+export const validateEmail = (email) => {
+	const emailValidation = /^[a-zA-Z0-9.!#$%&'*+\/=?^_`{|}~-]+@[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?(?:\.[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?)*$/;
```
rodrigok, line 540: Please move this lite to outside the function, no need to assign it on call, it's a global constant

Issue #22365, opened 2021-06-15
Uploading an image while file upload is disabled


alueers118:
### Description:

We disabled all file uploads for our rocket chat instance. This works for "normal" file uploads. But if a user copies and pastes an image into the text box of a chat window, the dialog to upload a file is displayed. When file uploads are disabled, this dialog shouldn't even be displayed.

When the user tries to upload the image, the upload will be correctly blocked. But the displayed error message can't be customized/translated.

### Steps to reproduce:

1. Disable all fileuploads in rocket-chat administration
2. Copy and paste an image in the textbox of a chat window
3. Try to send the image

### Expected behavior:

The dialog to upload the image shouldn't even be displayed.

Alternatively the error message for the rejected file upload should be customizable via translation files.

### Actual behavior:

The dialog to upload an image is displayed. The upload will be blocked but the resulting error message can't be customized/translated.

![grafik](https://user-images.githubusercontent.com/85939627/122054276-cfa88780-cde7-11eb-8959-ca446a6ff3c8.png)

### Server Setup Information:

- Version of Rocket.Chat Server: 3.15.0

johncrisp:
Thanks for reporting this.

Please always include the full server information when reporting.

Someone will look in due course thanks.

Rhuancpq:
I will work on this.

PR #23099, opened 2021-09-02
[FIX]  upload file with paste event when upload are disabled

juniopereirab:
Co-authored-by: Rhuan Carlos <rhuancarlos.queiroz@gmail.com>

## Issue(s)

#22365 

## Steps to test or reproduce

1. Disable all fileuploads in rocket-chat administration
2. Copy and paste an image in the textbox of a chat window
3. Try to send the image


app/ui-message/client/messageBox/messageBox.js
```diff
@@ -398,17 +398,21 @@ Template.messageBox.events({
...
+				}))
+				.filter(({ file }) => file !== null);
+			if (files.length) {
+				event.preventDefault();
+				fileUpload(files, input, { rid, tmid });
+			}
+		}
+		else {
+			alert('file_upload_disabled');
 		}
```
ggazzo, line 416:
```suggestion
		if(!settings.get('FileUpload_Enabled')) {
			return;
		}
		const files = items
			.filter((item) => item.kind === 'file' && item.type.indexOf('image/') !== -1)
			.map((item) => ({
				file: item.getAsFile(),
				name: `Clipboard - ${ moment().format(settings.get('Message_TimeAndDateFormat')) }`,
			}))
			.filter(({ file }) => file !== null);
		if (files.length) {
			event.preventDefault();
			fileUpload(files, input, { rid, tmid });
		}

```

ggazzo (changes requested):
We dont need a visual interaction if the files are disabled (if you believe we do please create a modal instead of `alert`) and always return first if you can, thanks for your PR

app/ui-message/client/messageBox/messageBox.js
juniopereirab: Okay. Thank you for this comment. I will resolve based on your feedback

PR #23110, opened 2021-09-03
[FIX] Can't edit profile information if any field update setting is disabled

matheusbsilva137:

- Check which fields have been updated before throwing errors in `validateUserEditing`.

## Issue(s)

[Task - ClickUp](https://app.clickup.com/t/5t4h67)

## Steps to test or reproduce

1. Go to **Administration > Accounts** and disable "Allow Username Change", "Allow Email Change", "Allow Custom Status Message" and "Allow Name Change";
2. Still logged in as an admin, go to your own profile under **Administration > Users**;
3. Try to add/remove any role. **_Expected behavior:_** these operations should be performed succesfully;
4. Try to update the Username, Name, Email and Custom Status Message fields (one at a time). **_Expected behavior:_** these operations should **not** be performed succesfully;
5. Enable all the options that have been disabled in step 1 and repeat step 4. **_Expected behavior:_** these operations should be performed succesfully.

### DIFF
--- a/app/lib/server/functions/saveUser.js
+++ b/app/lib/server/functions/saveUser.js
@@ -13,6 +13,7 @@ import { validateUserRoles } from '../../../../ee/app/authorization/server/valid
 import { getNewUserRoles } from '../../../../server/services/user/lib/getNewUserRoles';
 import { saveUserIdentity } from './saveUserIdentity';
 import { checkEmailAvailability, checkUsernameAvailability, setUserAvatar, setEmail, setStatusText } from '.';
+import { Users } from '../../../models/server';
 
 let html = '';
 let passwordChangedHtml = '';
@@ -146,8 +147,13 @@ function validateUserEditing(userId, userData) {
 
 	const canEditOtherUserInfo = hasPermission(userId, 'edit-other-user-info');
 	const canEditOtherUserPassword = hasPermission(userId, 'edit-other-user-password');
+	const user = Users.findOneById(userData._id);
+	const editingUserRoles = userData.roles
+							&& user.roles.length === userData.roles.length
+							&& user.roles.every((role, index) => role === user[index]);
+	const editingField = (previousValue, newValue) => newValue && newValue !== previousValue;
 
-	if (userData.roles && !hasPermission(userId, 'assign-roles')) {
+	if (editingUserRoles && !hasPermission(userId, 'assign-roles')) {
 		throw new Meteor.Error('error-action-not-allowed', 'Assign roles is not allowed', {
 			method: 'insertOrUpdateUser',
 			action: 'Assign_role',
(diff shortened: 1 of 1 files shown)


app/lib/server/functions/saveUser.js
```diff
@@ -146,8 +147,13 @@ function validateUserEditing(userId, userData) {
 
 	const canEditOtherUserInfo = hasPermission(userId, 'edit-other-user-info');
 	const canEditOtherUserPassword = hasPermission(userId, 'edit-other-user-password');
+	const user = Users.findOneById(userData._id);
+	const editingUserRoles = userData.roles
+							&& user.roles.length === userData.roles.length
+							&& user.roles.every((role, index) => role === user[index]);
```
ostjen, line 153: I don't think this is necessary, and if you're editing user roles wouldn't `user.roles.length` be different from `userData.roles.length` ?

app/lib/server/functions/saveUser.js
```diff
@@ -146,8 +147,13 @@ function validateUserEditing(userId, userData) {
...
+	const editingField = (previousValue, newValue) => newValue && newValue !== previousValue;
```
ostjen, line 154: I think `isEditingField` would be a more meaningful function name

ostjen, line 154: checking if newValue exists wouldn't introduce an edge case like allowing to delete a custom status even though the edit flag is off?

ostjen (changes requested):

app/lib/server/functions/saveUser.js
matheusbsilva137: I agree this wouldn't be necessary if the `roles` field was sent only when it's being edited, but the client code sends this field (and many others) in `userData` even when they haven't been updated (that's why I decided to add these checks before throwing the "not allowed to edit" errors, otherwise we'd throw errors even when a user wasn't updating a blocked field).

app/lib/server/functions/saveUser.js
matheusbsilva137: Ohh, it would. I've just removed this check, think we're not missing anything by removing it. Tks 🤗

app/lib/server/functions/saveUser.js
ostjen: you're correct, that is necessary, but in this case if you add or remove roles the `user.roles.length` will be different from `userData.roles.length`, right? If that's true I guess the best approach is  just create another anon func to compare both arrays elements

matheusbsilva137: Ohh, you're right. I've just updated `isEditingUserRoles` to correct its value in this case. I'm not considering changes in the order of the `roles` array's elements as an update.

app/lib/server/functions/saveUser.js
```diff
@@ -146,8 +147,19 @@ function validateUserEditing(userId, userData) {
...
 
-	if (userData.roles && !hasPermission(userId, 'assign-roles')) {
+	const isEditingUserRoles = (previousRoles, newRoles) => {
+		if (!Array.isArray(previousRoles) || !Array.isArray(newRoles) || previousRoles.length !== newRoles.length) {
+			return true;
+		}
+		previousRoles.sort();
+		newRoles.sort();
+		return !previousRoles.every((role, index) => role === newRoles[index]);
+	};
```
ostjen, line 159: I think using the underscore is the cleanest way to go, you should test it, but I thinks something like this will work

`const  isEditingUserRoles  = (previousRoles, newRoles) => !_.isEqual(_.sortBy(previousRoles), _.sortBy(newRoles))`

It will sort both arrays and compare if the contain the same elements

ostjen (changes requested):

app/lib/server/functions/saveUser.js
matheusbsilva137: Much cleaner and it works, thanks 😄

ostjen (dismissed):

ostjen (dismissed):

ostjen (approved):

hatung:
hi, I got an error when call update user with my users was registered without email. 

POST: api/v1/users.update
Response:
```
{
    "success": false,
    "error": "Cannot read property '0' of undefined"
}
```

PR #23131, opened 2021-09-06
[IMPROVE] Add jobId to job data passed onto processor

marceloschmidt:
Adds job id to the jobContext data that is passed to the processor.
This information is then available when the processor runs the job.


marceloschmidt:
Closed because @d-gubert said it has been implemented on another PR

PR #23134, opened 2021-09-06
[NEW] Voip: Permissions

KevLehman:


app/authorization/server/startup.js
```diff
@@ -140,6 +140,16 @@ Meteor.startup(function() {
...
+
+		// VOIP Permissions
+		// allows to manage voip calls configuration
+		{ _id: 'manage-voip-calls', roles: ['livechat-manager', 'admin'] },
+		// allows to hook on an ongoing call and listen
+		{ _id: 'ghost-voip-calls', roles: ['livechat-manager', 'livechat-monitor'] },
+		// allows to receive a voip call
+		{ _id: 'receive-voip-calls', roles: ['livechat-agent'] },
+		// allows to perform an outgoing voip call
+		{ _id: 'send-voip-calls', roles: ['livechat-agent'] },
```
renatobecker, line 152: What about `start` rather then `send` ?
```suggestion
		{ _id: 'start-voip-calls', roles: ['livechat-agent'] },
```

KevLehman: It's way better 😬 I'm bad naming things

app/authorization/server/startup.js
```diff
@@ -140,6 +140,16 @@ Meteor.startup(function() {
...
+
+		// VOIP Permissions
+		// allows to manage voip calls configuration
+		{ _id: 'manage-voip-calls', roles: ['livechat-manager', 'admin'] },
+		// allows to hook on an ongoing call and listen
+		{ _id: 'ghost-voip-calls', roles: ['livechat-manager', 'livechat-monitor'] },
+		// allows to receive a voip call
+		{ _id: 'receive-voip-calls', roles: ['livechat-agent'] },
+		// allows to perform an outgoing voip call
+		{ _id: 'start-voip-calls', roles: ['livechat-agent'] },
```
renatobecker, line 152: Except for the `start-voip-calls` permission, all of the others above will be part of the EE version. Do you guys remember that?

KevLehman: You mean, the `receive-voip-calls` isn't it? The one for incoming calls. Also, the `manage-voip-calls` should also be on CE since that's for admins to manage the voip configuration.

I'll move the other 2 to EE.

renatobecker: You are right the only one available in CE is `receive-voip-calls`, I just fixed it in the previous comment. My bad, sorry. Regarding `manage-voip-calls`, we might need to break it down a little more. There'll be `admin` settings and `management` settings, but I just can see one setting here, so I'm a bit confused, to be honest.

amolghode1981 (commented):
Thanks for setting up the permission. Few minor suggestions. 
1. I agree with Renato. I think, for CE, manager role will almost be insignificant, or none. Admin user would need to configure VoIP settings(interface for management, calling. Setting up agent-extension association.)
2. manage-voip-calls will be a role for manager. Manager will monitor the queue as well as spy the call. So I think it would be clearer if we rename it to manage-voip-contactcenter or manage-voip-channel.
3. One minor suggestion. If we decide  to keep spying as a separate permission, Can we change ghost-voip-calls -> spy-voip-calls or supervise-voip-calls?
4. Can we have receive-voip-calls -> inbound-voip-calls and start-voip-calls ->outbound-voip-calls. The purpose is that these are terms used in the contact centre.

renatobecker:
> 4\. Can we have receive-voip-calls -> inbound-voip-calls and start-voip-calls ->outbound-voip-calls. The purpose is that these are terms used in the contact centre.

I really like that too. A more industry-standard terminology.

KevLehman:
Hey! Great comments on this one, yup, let me do the changes 😬

amolghode1981:
Looks good to me

Issue #22548, opened 2021-07-02
There is no stretch option when discussing strings


hanzhiyan588:
### Description:

Most users don't have this stretch option when discussing strings
![image](https://user-images.githubusercontent.com/63633319/124223961-5aeb8200-db37-11eb-8775-522d840b53c8.png)
![image](https://user-images.githubusercontent.com/63633319/124224009-748cc980-db37-11eb-913e-4edf2ed6d2e9.png)

### Expected behavior:

Every user should have this stretch option

### Server Setup Information:

- Version of Rocket.Chat Server: 3.16.0
- Operating System: Ubuntu 20.04
- Deployment Method: Docker-compose
- Number of Running Instances: 1
- DB Replicaset Oplog: 1
- NodeJS Version: 12.22.1
- MongoDB Version: 4.0

### Client Setup Information

- Desktop App or Browser Version: Chrome 90.0.4430.212 (Official Build) (64-bit)
- Operating System: Ubuntu 20.04

debdutdeb:
Hi there, another issue? ;) Awesome, thank you !!

I'm happy to see you filling up the bug template, but unfortunately we're going to need a bit more information. Specifically about the clients of the users who aren't seeing the "Expand" button.

Please update the issue with the information I just mentioned and let us know with a comment.

TTYL :)

hanzhiyan588:
> 你好，还有一个问题？;） 太棒了，谢谢你 ！！
> 
> 我很高兴看到您填写错误模板，但不幸的是，我们需要更多信息。特别是关于没有看到“展开”按钮的用户的客户。
> 
> 请使用我刚才提到的信息更新问题，并通过评论告诉我们。
> 
> 下次再谈 ：）
Sorry, the buttons are the key. Only one or two people have it, not all of them

inetfuture:
> Specifically about the clients of the users who aren't seeing the "Expand" button.

We can't find any patterns, some clients just don't see the button. We've tried upgrading chrome to the latest, clearing cache, no luck.

inetfuture:
@debdutdeb

yash-rajpal:
The Stretch button is only visible to those users whose display resolution is low, or if the website is zoomed in.

To confirm this, you can try to zoom in the website and the stretch button will be visible to you. To confirm this - 
![Stretch_Button](https://user-images.githubusercontent.com/58601732/125167403-e57a5400-e1bd-11eb-9ae8-f3411e8066af.gif)

[...]

PR #23136, opened 2021-09-07
[FIX] Thread view expand command in wide screens

edward-raven:
Allows users with wide screens stretch thread sidebar.

## Issue(s)
#22548

## Steps to test or reproduce
- Maximize window to 1920px width
- Open any thread sidebar

Expected: there is expand button
Actual: there is no expand button


ggazzo:
makes no sense letting the thread be expanded on >1920px  screens, because after that resolution the thread size is a percentage from the screen, and that value is actually bigger than the "expanded"  thread

ggazzo:
but thans

ggazzo:
on even larger screens the effect is the opposite the expanded gets smaller than the normal version :x

edward-raven:
Hi @ggazzo 
Thanks for reply.

You're right, on pretty wide screens (starts from ~2800px) the effect of the `Expand` button is opposite. 
However:
> makes no sense letting the thread be expanded on >1920px screens

it's no quite correct. The suggested fix allows the thread be expanded on greater **or equal** 1920px screens. Here are screenshot on 1920px screen.
Without fix:

With fix:



So, it looks as expected.

BTW: About 24% PC users have 1920x1080 screen resolution, while only ~2% have really wide screens (based on https://gs.statcounter.com/screen-resolution-stats/desktop/worldwide)

ggazzo (approved):

edward-raven:
@ggazzo Thanks for approval

PR #23169, opened 2021-09-10
[IMPROVE] Unit testing with DOM access

tassoevan:

Adds support to proper unit testing with DOM APIs via jsdom. As example, adds unit testing for two React components.


.mocharc.dom.json
```diff
@@ -0,0 +1,13 @@
...
+	"require": [
+		"ts-node/register",
+		"./client/lib/tests/registerJsdom.ts"
+	],
+	"reporter": "spec",
+	"ui": "bdd",
+	"extension": ["ts", "tsx"],
+	"spec": [
+		"client/**/*.spec.ts",
+		"client/**/*.spec.tsx"
```
sampaiodiego, line 11: this glob finds tests other than DOM unit tests (it finds `client/lib/minimongo/bson.spec.ts` for example).. idk how bad is that

sampaiodiego (commented):
I don't feel confident enough to evaluate this. I don't know how valuable are "mount tests".. I see the effort is minimum, but I'm not sure about the value they'd add.

tassoevan:
Stale. Now we prefer performing tests with the Testing Library.

PR #23171, opened 2021-09-10
[BREAK] LDAP Refactoring

pierre-lehnen-rc:


server/lib/ldap/Manager.ts
```diff
@@ -0,0 +1,403 @@
+import limax from 'limax';
+import { SHA256 } from 'meteor/sha';
+import { Meteor } from 'meteor/meteor';
+import { Accounts } from 'meteor/accounts-base';
```
sampaiodiego, line 4: since this file is imported by a service, **ideally** it should not import any Meteor dependency. Doing so it won't work as a dedicated service.. so this is something we'll need to take a look in the future.

server/lib/ldap/Manager.ts
```diff
@@ -0,0 +1,403 @@
...
+
+export class LDAPManager {
+	public static async login(username: string, password: string): Promise<LDAPLoginResult> {
+		logger.log({ msg: 'Init LDAP login', username });
+
+		if (settings.get('LDAP_Enable') !== true) {
+			return this.fallbackToDefaultLogin(username, password);
+		}
+
+		const ldap = this.getNewConnection();
```
sampaiodiego, line 31: does this mean for every new login attempt a new LDAP connection would be established? I didn't found if the connection is being closed afterwards, I guess we should?

sampaiodiego (changes requested):

package.json
```diff
@@ -68,6 +68,7 @@
 		"@types/bad-words": "^3.0.1",
 		"@types/bcrypt": "^5.0.0",
 		"@types/body-parser": "^1.19.0",
+		"@types/bunyan": "^1.8.7",
```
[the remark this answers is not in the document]
pierre-lehnen-rc: nope, wasn't changed on purpose.

app/importer/server/classes/VirtualDataConverter.ts
```diff
@@ -0,0 +1,154 @@
+import { Random } from 'meteor/random';
+
+import { IImportChannel } from '../definitions/IImportChannel';
+import { ImportDataConverter } from './ImportDataConverter';
+import type { IConverterOptions } from './ImportDataConverter';
+
+export class VirtualDataConverter extends ImportDataConverter {
+	protected _objects: Record<string, Array<Record<string, any>>>;
```
sampaiodiego, line 8: this could be better typed.. from what I could see, there are three distinct "types": `user` , `channel` and `message` .. also a `Map` seems to fit better.

this way, methods like `getUsersToImport` , `getMessagesToImport` and `getChannelsToImport` could also be better typed, instead of returning `any`

ee/server/services2/ldap/service.ts
```diff
@@ -0,0 +1,16 @@
+import '../../broker';
+
+import { api } from '../../../../server/sdk/api';
```
sampaiodiego, line 3: the `ee/server/services/*` doesn't work when being imported is bacause of:

https://github.com/RocketChat/Rocket.Chat/blob/0a293e547fcab88d548d14b6c8608b530b5d6f9d/.meteorignore#L1


the idea is that all EE services were intended to run on a separate process, which is not the case anymore with thi LDAP EE service..

with that in mind, I think we can use the name `local-services` or `internal-services` .. what do you think?

package.json
```diff
@@ -68,6 +68,7 @@
 		"@types/bad-words": "^3.0.1",
 		"@types/bcrypt": "^5.0.0",
 		"@types/body-parser": "^1.19.0",
+		"@types/bunyan": "^1.0.2",
```
sampaiodiego, line 71: I'd actually recommend to remove this dependency, as it seems to not be used.

server/startup/migrations/v236.ts
```diff
@@ -0,0 +1,125 @@
...
+
+		Settings.remove({
+			_id: {
+				$in: [
+					'LDAP_Sync_Now',
+					'LDAP_Test_Connection',
+					'LDAP_Sync_CustomFields',
+					'LDAP_Sync_User_Data',
+					'LDAP_Sync_User_Data_FieldMap',
+					'LDAP_Enable_LDAP_Roles_To_RC_Roles',
```
sampaiodiego, line 107: this settings is still used at https://github.com/RocketChat/Rocket.Chat/blob/0a293e547fcab88d548d14b6c8608b530b5d6f9d/app/statistics/server/lib/getServicesStatistics.ts#L43

also, this and some other settings are mentioned on this README file, which I think we can get rid off, right? 
https://github.com/RocketChat/Rocket.Chat/blob/0a293e547fcab88d548d14b6c8608b530b5d6f9d/ee/app/ldap-enterprise/README.md#L18

sampaiodiego (changes requested):

app/utils/server/lib/cron/Cronjobs.ts
```diff
@@ -1,15 +1,18 @@
 import { SyncedCron } from 'meteor/littledata:synced-cron';
 
+type scheduleType = 'cron' | 'text';
```
ggazzo, line 3:
```suggestion
type ScheduleType = 'cron' | 'text';
```

app/utils/server/lib/cron/Cronjobs.ts
```diff
@@ -1,15 +1,18 @@
...
+
 export interface ICronJobs {
-	add(name: string, schedule: string, callback: Function): void;
+	add(name: string, schedule: string, callback: Function, scheduleType?: scheduleType): void;
```
ggazzo, line 6:
```suggestion
	add(name: string, schedule: string, callback: Function, scheduleType?: ScheduleType): void;
```

definition/IUser.ts
```diff
@@ -33,6 +33,12 @@ export interface IUserEmailCode {
 type LoginToken = IMeteorLoginToken & IPersonalAccessToken;
 export type Username = string;
 
+export interface ILoginUsername {
+	username?: string;
+	email?: string;
+}
+export type LoginUsername = string | ILoginUsername;
+
```
ggazzo, line 41:
```suggestion
export type ILoginUsername = {
	username: string;
} | {
	email: string;
}
export type LoginUsername = string | ILoginUsername;

```

ggazzo (changes requested):
ta bonito mas ta errado ❤️

server/lib/ldap/Manager.ts
```diff
@@ -0,0 +1,403 @@
...
+			return;
+		}
+
+		const avatar = this.getAvatarFromUser(ldapUser);
+		if (!avatar) {
+			return;
+		}
+
+		logger.info('Syncing user avatar');
+		Meteor.defer(() => Meteor.runAsUser(user._id, () => setUserAvatar(user, avatar, 'image/jpeg', 'rest')));
```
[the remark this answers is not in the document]
pierre-lehnen-rc: The UploadFS class will throw an error if a file is inserted without a Fiber

server/services/ldap/service.ts
```diff
@@ -0,0 +1,16 @@
...
+
+export class LDAPService extends ServiceClass implements ILDAPService {
+	protected name = 'ldap';
+
+	async loginRequest(username: string, password: string): Promise<LDAPLoginResult> {
+		return LDAPManager.login(username, password);
+	}
+
+	async testConnection(): Promise<void> {
+		return LDAPManager.testConnection();
```
sampaiodiego, line 14: should this disconnect after a successful connection?

server/services/ldap/service.ts
```diff
@@ -0,0 +1,16 @@
...
+import { LDAPLoginResult } from '../../../definition/ldap/ILDAPLoginResult';
+
+export class LDAPService extends ServiceClass implements ILDAPService {
+	protected name = 'ldap';
+
+	async loginRequest(username: string, password: string): Promise<LDAPLoginResult> {
+		return LDAPManager.login(username, password);
+	}
+
+	async testConnection(): Promise<void> {
```
sampaiodiego, line 13: why not return `boolean` for a successful or not connection?

app/importer/server/classes/VirtualDataConverter.ts
```diff
@@ -0,0 +1,149 @@
...
+		}
+
+		const record = this.getVirtualRecordById(_id);
+
+		if (record) {
+			record.skipped = true;
+		}
+	}
+
+	protected getMessagesToImport(): any {
```
sampaiodiego, line 107:
```suggestion
	protected async getMessagesToImport(): Promise<IImportMessageRecord[]> {
```

app/importer/server/classes/VirtualDataConverter.ts
```diff
@@ -0,0 +1,149 @@
...
+		if (!this.useVirtual) {
+			return super.getMessagesToImport();
+		}
+
+		return this._messageRecords;
+	}
+
+	protected getChannelsToImport(): any {
```
sampaiodiego, line 115:
```suggestion
	protected async getChannelsToImport(): Promise<IImportChannelRecord[]> {
```

sampaiodiego (changes requested):

server/services/ldap/service.ts
pierre-lehnen-rc: It does!

sampaiodiego (dismissed):

sampaiodiego (approved):

joker234:
@pierre-lehnen-rc can you give a brief description what changed or better what needs to be done when upgrading and using the LDAP integration until the stable release 4.0.0? That would be very helpful!

lobo87:
Im kind of missing the feature where I can map LDAP groups to certain roles in RC. was that removed? didn't find anything related to that.

dennisvj:
> Im kind of missing the feature where I can map LDAP groups to certain roles in RC. was that removed? didn't find anything related to that.
> 
> due to this fact and syncing the users on log in, my admins lost their admin role.

Same here. Have I read it correctly it is now an Enterprise feat (Enhanced LDAP) ?

milton-rucks:
> > Im kind of missing the feature where I can map LDAP groups to certain roles in RC. was that removed? didn't find anything related to that.
> > due to this fact and syncing the users on log in, my admins lost their admin role.
> 
> Same here. Have I read it correctly it is now an Enterprise feat (Enhanced LDAP) ?

That's correct.

You can find more information here:

Forum announcement:
https://forums.rocket.chat/t/upcoming-changes-to-identity-management-integrations/11994

List of available login methods and features (EE vs CE):
https://docs.rocket.chat/quick-start/identity-management-ee-vs-ce

PR #23190, opened 2021-09-14
[IMPROVE] Canned response admin settings

tiagoevanp:


server/startup/migrations/v236.ts
```diff
@@ -0,0 +1,21 @@
+import { Migrations } from '../../../app/migrations/server';
+import { Settings } from '../../../app/models/server';
+
+Migrations.add({
+	version: 236,
+	up() {
+		Settings.removeById('Canned Responses');
```
KevLehman, line 7: Wasn't this removed some time ago too? 👀

KevLehman (approved):

server/startup/migrations/v236.ts
tiagoevanp: not for me... 👀

PR #22975, opened 2021-08-19
[IMPROVE] Change log format to JSON

sampaiodiego:
## Proposed changes (including videos or screenshots)

The idea is to have a new and standardized log format so looked like a good opportunity to also switch to a well known library.

- [x] Change log output in production environments to JSON
- [x] In dev env there is an easy to read representation of the JSON logs
- [x] Removed the ability to expand `Logger`, either via `methods`, `sections`, etc . If you want similar functionality you can create a child logger using `Logger.section` (probably need a better name for this, other than `section`)
- [x] Change log level of REST endpoints to `info` and add useful info to identify who is calling the endpoints
- [x] Removed the ability to create "boxes" from `Logger` (there are specific functions for it now)
- [x] Replaced most of the place using raw `console.log` and switched to `Logger`
- [x] Convert `SystemLogger` to use the new `Logger` class
- [x] Remove `LoggerManager` and its related settings

A few items that needs some attention:

* We should consider changing the default Log Level to INFO, so we can have better logs by default
* Previously log messages using `console.log` are now suppressed as the default log level is still ERROR and those logs were changed to INFO
* We could also consider creating a new log level to shows those previously shown logs even on ERROR level, but this doesn't feel right
* A further addition would be creating unit tests to make sure the log format is stable

Here is how the new logs format looks like:

```
{"level":30,"time":"2021-08-30T17:05:13.957Z","pid":90,"hostname":"4d7041154553","name":"System","msg":"Setting default file store to 'GridFS'"}
{"level":30,"time":"2021-08-30T17:05:18.318Z","pid":90,"hostname":"4d7041154553","name":"Search Logger","msg":"create search provider defaultProvider"}
{"level":30,"time":"2021-08-30T17:05:18.326Z","pid":90,"hostname":"4d7041154553","name":"Search Logger","msg":"create search provider chatpalProvider"}
{"level":30,"time":"2021-08-30T17:05:26.505Z","pid":90,"hostname":"4d7041154553","name":"StreamBroadcast","section":"Auth","msg":"prevent self connect 'undefined:3000'"}
{"level":50,"time":"2021-08-30T17:08:20.383Z","pid":90,"hostname":"4d7041154553","name":"System","msg":"Exception while invoking method getUserRoles 'Invalid user [error-invalid-user]'"}
[...]

PR #23205, opened 2021-09-15
Regression: View Logs administration page crashing

tassoevan:

Fixes the `stdout.queue` endpoint; makes the components type-safe.

## Issue(s)

From #22975


sampaiodiego:
I had worked on a fix for the backend part at #23194 ... it will be good to merge both though, as they seem complementary to each other.

ggazzo (approved):

Issue #22525, opened 2021-07-01
CAS login window won't close due to content security policy


Monschichi:
### Description:

When logging in via CAS the login window doesn't close with version 3.16.0.

### Steps to reproduce:

1. enable CAS login
2. login via CAS
3. CAS login window stays open
4. when manually closed the login to rocket chat works

### Expected behavior:

1. login via CAS
2. CAS login window closes automatically

### Actual behavior:

1. login via CAS
2. CAS login window does not close automatically

### Server Setup Information:

- Version of Rocket.Chat Server: 3.16.0
- Operating System: Debian 10
- Deployment Method: tar
- Number of Running Instances: 6
- NodeJS Version: 12.22.1
- MongoDB Version: 3.6.18

### Client Setup Information

- Desktop App or Browser Version: Firefox 89.0.2 and Chromium 91.0.4472.114
- Operating System: Linux

### Additional context

Until Version 3.15.0 the auto close worked fine.
The final CAS page https://DOMAIN/_cas/XXX contains `<html><head><script>window.close()</script></head></html>` and the content security policy is: `default-src 'self'; script-src 'self' 'unsafe-eval'; connect-src * 'self' data:; img-src data: 'self' http://* https://*; style-src 'self' 'unsafe-inline'; media-src 'self' http://* https://*; frame-src 'self' http://* https://*; font-src 'self' data:;`.
Developer console outputs: `Content Security Policy: The page’s settings blocked the loading of a resource at inline (“script-src”).`

### Relevant logs:

Developer console: `Content Security Policy: The page’s settings blocked the loading of a resource at inline (“script-src”).`

bseclier:
Thanks for posting the issue @Monschichi.
Same problem in 3.16.1, this is a blocking issue :(

goyome:
The chromium developer console add some information on the way to correct the problem.

`Refused to execute inline script because it violates the following Content Security Policy directive: "script-src 'self' 'unsafe-eval'". Either the 'unsafe-inline' keyword, a hash ('sha256-jqxtvDkBbRAl9Hpqv68WdNOieepg8tJSYu1xIy7zT34='), or a nonce ('nonce-...') is required to enable inline execution.`

goyome:
It seems to be linked to this change 
[...]

PR #23215, opened 2021-09-16
[FIX] Add CSP to authorize auto-close of CAS login window

goyome:

Add the hash of the JS inside the page that won't close ( window.close(); )

## Issue(s)

#22525


goyome:
@ggazzo @pierre-lehnen-rc Is there a problem with this PR ? 
Could you please review it ? We are enable to upgrade as we consider this bug as a blocking issue.

Thanks in advance

pierre-lehnen-rc (approved):

goyome:
Thanks for your review :+1:

PR #23219, opened 2021-09-16
[FIX] Mark agents as unavailable when they logout

KevLehman:


murtaza98 (approved):
LGTM!! :rocket: 

Just a side note - while testing I found that the translations were not getting applied on the following place on UI, so I've added a commit to this PR itself to fix it.

image

PR #23225, opened 2021-09-17
[NEW] stats on seats request 

g-thome:
Add seat request count on statistics report


ee/server/api/licenses.ts
```diff
@@ -20,6 +21,7 @@ API.v1.addRoute('licenses.requestSeatsLink', { authRequired: true }, {
 		}
 
 		const url = getSeatsRequestLink();
+		Promise.await(Analytics.saveSeatRequest());
```
gabriellsh, line 24: Not sure this should go here... This endpoint is called every time the Users page is accessed, and not when clicking the link...

gabriellsh (changes requested):

matheusbsilva137:
LGTM, just waiting for @ostjen 's review so that we can approve this 😄

matheusbsilva137 (approved):

ee/server/requestSeatsRoute.ts
```diff
@@ -0,0 +1,20 @@
+import { IncomingMessage, ServerResponse } from 'http';
+
+import { Meteor } from 'meteor/meteor';
+import { WebApp } from 'meteor/webapp';
+
+import { getSeatsRequestLink } from '../app/license/server/getSeatsRequestLink';
+import { Analytics } from '../../server/sdk';
+import { isEnterprise } from '../app/license/server/license';
```
matheusbsilva137, line 8: Hey @g-thome , following the changes that Rodrigo suggested to #23256 , I think we should create a regression to fix this import too :grimacing:

g-thome: I believe Rodrigo sees that import as problematic because I was importing EE outside of EE, so I think doing it over here is ok. Am I right? @rodrigok

PR #23239, opened 2021-09-20
[NEW] Add new Endpoints to manage VoIP server configs

murtaza98:

Corresponding ClickUp task: https://app.clickup.com/t/7qeckt

### DIFF
--- a/app/api/server/v1/voip/server-config.ts
+++ b/app/api/server/v1/voip/server-config.ts
@@ -1,4 +1,5 @@
 import { Match, check } from 'meteor/check';
+import { TAPi18n } from 'meteor/rocketchat:tap-i18n';
 
 import { API } from '../../api';
 import { hasPermission } from '../../../../authorization/server';
@@ -9,7 +10,7 @@ import { ICallServerConfigData, IManagementConfigData, ServerType } from '../../
 API.v1.addRoute('voipServerConfig.management', { authRequired: true }, {
 	get() {
 		if (!hasPermission(this.userId, 'manage-voip-contact-center-settings')) {
-			return API.v1.unauthorized('error-no-permission-manage-voip-contact-center-settings'); // TODO: create translations
+			return API.v1.unauthorized(TAPi18n._('error-insufficient-permission', { permission: 'manage-voip-contact-center-settings' }));
 		}
 
 		const config = Promise.await(Voip.getServerConfigData(ServerType.MANAGEMENT));
@@ -31,7 +32,7 @@ API.v1.addRoute('voipServerConfig.management', { authRequired: true }, {
 		}));
 
 		if (!hasPermission(this.userId, 'manage-voip-contact-center-settings')) {
-			return API.v1.unauthorized('error-no-permission-manage-voip-contact-center-settings'); // TODO: create translations
+			return API.v1.unauthorized(TAPi18n._('error-insufficient-permission', { permission: 'manage-voip-contact-center-settings' }));
 		}
 
 		const { host, port, serverName, username, password } = this.bodyParams;
@@ -41,9 +42,9 @@ API.v1.addRoute('voipServerConfig.management', { authRequired: true }, {
 		const result = Promise.await(Voip.addServerConfigData({
 			type: ServerType.MANAGEMENT,
 			host,
+			serverName,
 			configData: {
 				port,
-				serverName,
 				username,
 				password,
 			} as IManagementConfigData,
@@ -57,7 +58,7 @@ API.v1.addRoute('voipServerConfig.management', { authRequired: true }, {
 API.v1.addRoute('voipServerConfig.callServer', { authRequired: true }, {
 	get() {
 		if (!hasPermission(this.userId, 'manage-voip-call-settings')) {
-			return API.v1.unauthorized('error-no-permission-manage-voip-call-settings'); // TODO: create translations
+			return API.v1.unauthorized(TAPi18n._('error-insufficient-permission', { permission: 'manage-voip-call-settings' }));
 		}
 
 		const config = Promise.await(Voip.getServerConfigData(ServerType.CALL_SERVER));
@@ -71,26 +72,26 @@ API.v1.addRoute('voipServerConfig.callServer', { authRequired: true }, {
 	post() {
 		check(this.bodyParams, Match.ObjectIncluding({
 			host: String,
+			serverName: String,
 			websocketPort: Number,
 			websocketPath: String,
-			callServerName: String,
 		}));
 
 		if (!hasPermission(this.userId, 'manage-voip-call-settings')) {
-			return API.v1.unauthorized('error-no-permission-manage-voip-call-settings'); // TODO: create translations
+			return API.v1.unauthorized(TAPi18n._('error-insufficient-permission', { permission: 'manage-voip-call-settings' }));
 		}
 
-		const { host, websocketPort, websocketPath, callServerName } = this.bodyParams;
+		const { host, websocketPort, websocketPath, serverName } = this.bodyParams;
 
 		Promise.await(Voip.deleteServerConfigDataIfAvailable(ServerType.CALL_SERVER));
 
 		const result = Promise.await(Voip.addServerConfigData({
 			type: ServerType.CALL_SERVER,
 			host,
+			serverName,
 			configData: {
 				websocketPort,
 				websocketPath,
-				callServerName,
 			} as ICallServerConfigData,
 		}));
 
--- a/app/api/server/v1/voip/server-config.ts
+++ b/app/api/server/v1/voip/server-config.ts
@@ -42,8 +42,8 @@ API.v1.addRoute('voipServerConfig.management', { authRequired: true }, {
 		Promise.await(Voip.addServerConfigData({
 			type: ServerType.MANAGEMENT,
 			host,
-			serverName,
-			configActive: true,
+			name: serverName,
+			active: true,
 			configData: {
 				port,
 				username,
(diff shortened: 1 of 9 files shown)


definition/IVoipServerConfig.ts
```diff
@@ -2,11 +2,24 @@ import { IRocketChatRecord } from './IRocketChatRecord';
...
 
 export interface IVoipServerConfig extends IRocketChatRecord {
 	type: ServerType;
 	host: string;
+	configData: IManagementConfigData | ICallServerConfigData;
+}
+
+export interface IManagementConfigData {
 	port: number;
+	serverName: string; // Internal name to be used in reports
```
amolghode1981, line 16: serverName will be in IVoipServerConfig. This is a user defined name. This will be used in the reports to identify the server.

murtaza98: Okay got it. Let me move this property to the parent interface i.e. IVoipServerConfig

app/api/server/v1/voip/server-config.ts
```diff
@@ -18,6 +20,37 @@ API.v1.addRoute('voipServerConfig.management', { authRequired: true }, {
...
+			password: String,
+		}));
+
+		if (!hasPermission(this.userId, 'manage-voip-contact-center-settings')) {
+			return API.v1.unauthorized('error-no-permission-manage-voip-contact-center-settings'); // TODO: create translations
+		}
+
+		const { host, port, serverName, username, password } = this.bodyParams;
+
+		Promise.await(Voip.deleteServerConfigDataIfAvailable(ServerType.MANAGEMENT));
```
amolghode1981, line 39: I have some questions about how we intend to handle update/delete. 
1. This server information is going to be a part of our reports. If we delete (or update) the record completely, we will not have a way to trace back the server information from the report if someone deletes/modifies the record.
2. The evidence trail is important in case of auditing. So data actually does not get deleted for predefined duration. 
It is usually 3 years.   
So we need to keep in mind that while preparing the report, we actually duplicate the entire server record instead of just the server name. Can we add a before |deleteServerConfigDataIfAvailable| to document this fact?

murtaza98: Hi, @amolghode1981 Thanks for the feedback. Based on the use-case u shared, I think we'll need to update the DB schema as well to maintain the audit trail. However, before modifying the schema, I wanted to make sure I understand the overall requirements first, so can you please answer the following questions

1. Do we need to keep a track of users who created/edited/deleted a management config record?
2. Do we also need to keep a track of `call-server` type?
3. Will in the reports, there will ever be a query like "Show me server records active during particular time-frame?"

murtaza98: More info about point 3: This question is in the context of whether we need to store data about the duration of the record when it was active

amolghode1981: > Do we need to keep a track of users who created/edited/deleted a management config record?
No. We do not need to keep track of users who did the operations.
> Do we also need to keep a track of call-server type records?
We need to keep track of both. Management type record is used only to pull the data but the data dictates the behaviour of call server. E.g We will store that agent x served queue y on call server z. If the call server is changed to w which does not have a queue y, this record will not make sense if we do not maintain the call server reference. I think same holds true for management interface.
> Will in the reports, there will ever be a query like "Show me server records active during particular time-frame?", More info about point 3: This question is in the context of whether we need to store data about the duration of the record when it was active. 
I dont think that would be required. While inserting the data in the report datacollection, we will have to put the timestamp because most of the reports are going to be trend based reports (e.g Call served per agent per month or abandoned calls per month)

I am not sure if we need to modify the schema at this point of time. in CE, we do not have any reports or schemas related to it. Neither for CEI suggest that it would be good to just add a comment around deleteServerConfigDataIfAvailable to consider this fact.

definition/IVoipServerConfig.ts
```diff
@@ -2,11 +2,24 @@ import { IRocketChatRecord } from './IRocketChatRecord';
...
-	CALLS = 'calls',
+	CALL_SERVER = 'call-server',
 }
 
 export interface IVoipServerConfig extends IRocketChatRecord {
 	type: ServerType;
 	host: string;
+	serverName: string;
+	configData: IManagementConfigData | ICallServerConfigData;
+	configActive: boolean;
```
KevLehman, line 13:
```suggestion
	active: boolean;
```

definition/IVoipServerConfig.ts
```diff
@@ -2,11 +2,24 @@ import { IRocketChatRecord } from './IRocketChatRecord';
...
 export enum ServerType {
 	MANAGEMENT = 'management',
-	CALLS = 'calls',
+	CALL_SERVER = 'call-server',
 }
 
 export interface IVoipServerConfig extends IRocketChatRecord {
 	type: ServerType;
 	host: string;
+	serverName: string;
```
KevLehman, line 11:
```suggestion
	name: string;
```

app/models/server/raw/VoipServerConfiguration.ts
```diff
@@ -10,4 +10,8 @@ export class VoipServerConfigurationRaw extends BaseRaw<IVoipServerConfig> {
 	) {
 		super(col, trash);
 	}
+
+	removeByType(type: ServerType): Promise<DeleteWriteOpResultObject> {
```
KevLehman, line 14:
```suggestion
	removeOneByType(type: ServerType): Promise<DeleteWriteOpResultObject> {
```

app/api/server/v1/voip/server-config.ts
```diff
@@ -0,0 +1,102 @@
...
+			websocketPath: String,
+		}));
+
+		if (!hasPermission(this.userId, 'manage-voip-call-settings')) {
+			return API.v1.unauthorized(TAPi18n._('error-insufficient-permission', { permission: 'manage-voip-call-settings' }));
+		}
+
+		const { host, websocketPort, websocketPath, serverName } = this.bodyParams;
+
+		Promise.await(Voip.deactivateServerConfigDataIfAvailable(ServerType.CALL_SERVER));
```
KevLehman, line 87: Shouldn't this be handled inside addServerConfigData? 👀

server/services/voip/service.ts
```diff
@@ -17,6 +18,43 @@ export class VoipService extends ServiceClass implements IVoipService {
...
 
+	async addServerConfigData(config: Omit<IVoipServerConfig, '_id' | '_updatedAt'>): Promise<boolean> {
+		const { type } = config;
+
+		const existingConfig = await this.getServerConfigData(type);
+		if (existingConfig) {
+			throw new Error(`Error! There already exists an active record of type ${ type }`);
+		}
+
+		await this.VoipServerConfiguration.insertOne(config);
```
KevLehman, line 29: Return this right away, since the return value is not sent as an API response
```suggestion
		return !!await this.VoipServerConfiguration.insertOne(config);
```

(not sure if eslint will like the !!await 🙈 )

server/services/voip/service.ts
```diff
@@ -17,6 +18,43 @@ export class VoipService extends ServiceClass implements IVoipService {
...
+			throw new Error(`Error! No active record exists of type ${ type }`);
+		}
+
+		await this.VoipServerConfiguration.updateOne({ type, configActive: true }, config);
+
+		return true;
+	}
+
+	// in-future, if we want to keep a track of duration during which a server config was active, then we'd need to modify the
+	// 		IVoipServerConfig interface and add columns like "valid_from_ts" and "valid_to_ts"
```
KevLehman, line 48: Is the extra space here needed? 👀

app/api/server/v1/voip/server-config.ts
```diff
@@ -0,0 +1,102 @@
...
+		}
+
+		const config = Promise.await(Voip.getServerConfigData(ServerType.CALL_SERVER));
+		if (!config) {
+			return API.v1.notFound();
+		}
+
+		return API.v1.success({ ...config });
+	},
+	// NOTE: you can use this POST endpoint for both create and update operation
```
KevLehman, line 72: I'm not sure about the frequency of updating 🤔 don't you think this can cause the collection to be filled with a lot of stale data? Since we're just deactivating the previous config, we can have a lot of duplicated server settings here. 

@amolghode1981 , is that needed? 👀

KevLehman (changes requested):

app/api/server/v1/voip/server-config.ts
murtaza98: Yeah, that makes sense :+1: Done

app/api/server/v1/voip/server-config.ts
murtaza98: Hi, @KevLehman Yeah, we indeed need to keep a track of the deactivated server config for auditing purposes. Please refer to Amol's earlier comment [here](https://github.com/RocketChat/Rocket.Chat/pull/23239#discussion_r712049547) for more info

app/api/server/v1/voip/server-config.ts
```diff
@@ -0,0 +1,100 @@
...
+			password: String,
+		}));
+
+		if (!hasPermission(this.userId, 'manage-voip-contact-center-settings')) {
+			return API.v1.unauthorized(TAPi18n._('error-insufficient-permission', { permission: 'manage-voip-contact-center-settings' }));
+		}
+
+		const { host, port, serverName, username, password } = this.bodyParams;
+
+		Promise.await(Voip.deactivateServerConfigDataIfAvailable(ServerType.MANAGEMENT));
```
KevLehman, line 40: Did you forgot to remove this one? 👀

KevLehman (approved):
Just remove the extra call to deactivateServerConfig and we're good 🥳

app/api/server/v1/voip/server-config.ts
murtaza98: Ooops! :see_no_evil: My bad!! Glad you caught it :raised_hands:

PR #23256, opened 2021-09-21
[IMPROVE] Better text for auth banner

g-thome:

Change the text in the banner warning for auth changes


server/startup/migrations/v238.ts
```diff
@@ -0,0 +1,34 @@
+import { Mongo } from 'meteor/mongo';
+
+import { Migrations } from '../../../app/migrations/server';
+import { isEnterprise } from '../../../ee/app/license/server';
```
rodrigok, line 4: We should never import the EE folder directly.

server/startup/migrations/v238.ts
```diff
@@ -0,0 +1,34 @@
...
+import { Settings } from '../../../app/models/server/raw';
+// import { Banner } from '../../sdk';
+// import { BannerPlatform } from '../../../definition/IBanner';
+// import { IUser } from '../../../definition/IUser';
```
rodrigok, line 8: Remove these comments

server/startup/migrations/v238.ts
```diff
@@ -0,0 +1,34 @@
...
+		const LDAPEnabled = settings.get('LDAP_Enable');
+		const SAMLEnabled = settings.get('SAML_Custom_Default');
+
+		const isEE = isEnterprise();
+
+		if (!isEE && (isCustomOAuthEnabled || LDAPEnabled || SAMLEnabled)) {
+			return;
+		}
+
+		const msg = 'Please notice that after the next release (4.0) advanced functionalities of LDAP, SAML, and Custom Oauth will be available only in Enterprise Edition and Gold plan. Check the official announcement for more info: https://go.rocket.chat/i/authentication-changes';
```
rodrigok, line 29: Since this will be released on 4.0 the text seems wrong

server/startup/migrations/v238.ts
```diff
@@ -0,0 +1,34 @@
...
+			value: true,
+		};
+
+		const isCustomOAuthEnabled = !!Promise.await(Settings.findOne(query));
+		const LDAPEnabled = settings.get('LDAP_Enable');
+		const SAMLEnabled = settings.get('SAML_Custom_Default');
+
+		const isEE = isEnterprise();
+
+		if (!isEE && (isCustomOAuthEnabled || LDAPEnabled || SAMLEnabled)) {
```
rodrigok, line 25: Could check the banner collections if cloud already sent the alert earlier to not show it again

rodrigok (dismissed):

tassoevan (approved):

server/startup/migrations/v238.ts
g-thome: I don't think this will be necessary. Since I'm just updating the text field from these documents, everything else should work the same. In other words, this banner won't show up for people who clicked on the dismiss button.

server/startup/migrations/v238.ts
g-thome: done

server/startup/migrations/v238.ts
g-thome: done

server/startup/migrations/v238.ts
casalsgh: Confirmed with PM that this is the current approved message.

server/startup/migrations/v238.ts
matheusbsilva137: I think you haven't pushed your changes, @g-thome 😬

matheusbsilva137 (approved):

server/startup/migrations/v238.ts
```diff
@@ -0,0 +1,28 @@
...
+			value: true,
+		};
+
+		const isCustomOAuthEnabled = !!Promise.await(Settings.findOne(query));
+		const LDAPEnabled = settings.get('LDAP_Enable');
+		const SAMLEnabled = settings.get('SAML_Custom_Default');
+
+		if (isCustomOAuthEnabled || LDAPEnabled || SAMLEnabled) {
+			return;
+		}
```
rodrigok, line 21: Now that I understood this is updating an existing banner (if it exists) I don't believe we need these checks. The update will update only if the banner already exists

casalsgh (approved):

casalsgh (approved):

PR #23261, opened 2021-09-21
[FIX] Stop queue when Omnichannel is disabled or the routing method does not support it

KevLehman:

- Add missing key logs
- Stop queue (and logs) when livechat is disabled or when routing method does not support queue
- Stop ignoring offline bot agents from delegation (previously, if a bot was offline, even with "Assign new conversations to bot agent" enabled, bot will be ignored and chat will be left in limbo (since bot was assigned, but offline).


murtaza98 (approved):

ee/app/livechat-enterprise/server/lib/LivechatEnterprise.js
```diff
@@ -268,4 +268,10 @@ settings.onload('Livechat_Routing_Method', function() {
 	} queue`);
 
 	routingSupportsAutoAssign ? queueWorker.start() : queueWorker.stop();
+}
+
+settings.get('Livechat_enabled', (_, value) => {
+	value && settings.get('Livechat_Routing_Method') ? shouldQueueStart() : queueWorker.stop();
```
murtaza98, line 274: Although I see no harm in adding this - `settings.get('Livechat_Routing_Method')` - extra check here, I didn't quite understand when would the condition occur when the `Livechat_Routing_Method` setting value gets `undefined`.This is because this setting has a default value as **Auto Selection** - [Reference](https://github.com/RocketChat/Rocket.Chat/blob/43828017e3b62202ed9b1c6a3ea33d459e4084f2/app/livechat/server/config.js#L426) - plus the input type is select, so there's no way a user can add a blank string there

murtaza98 (approved):

ee/app/livechat-enterprise/server/lib/LivechatEnterprise.js
KevLehman: Yeah, I didn't thought this was needed at first 😬 but on a new generated server, the setting Livechat_enabled would be created before creating Livechat_Routing_Method. So, at one point, the server won't have that setting in place, and the RoutingManager will throw an error because the routing method hasn't been selected yet (cause the method is set in a setting callback)

murtaza98: Okay, that makes sense!! Thanks for the explanation :smile: :100:

PR #23277, opened 2021-09-23
[BREAK] Remove old migrations up to version 2.4.14

sampaiodiego:

To update to version 4.0.0 you'll need to be running at least version 3.0.0, otherwise you might loose some database migrations which might have unexpected effects.

This aims to clean up the code, since upgrades jumping 2 major versions are too risky and hard to maintain, we'll keep only migration from that last major (in this case 3.x).

Another changes included in this PR:

- Removed old `Migrations` class - functionality remains the same though
  - New migrations should use `addMigration()` instead of `Migrations.add()`
- Migrations code re-written in TS

Still missing some code clean up.


ggazzo (approved):

geekgonecrazy:
@debdutdeb This one is one you will want to pay attention to from snap perspective.  🙏

PR #23298, opened 2021-09-28
[IMPROVE] PoC: React Query

tassoevan:


tassoevan:
Stale. We are applying React Query now, but we don't have an implementation plan yet.

PR #23301, opened 2021-09-28
[FIX] Showing thumbnails on files list

ostjen:
before
image

after
image


app/api/server/v1/im.js
```diff
@@ -146,7 +146,8 @@ API.v1.addRoute(['dm.files', 'im.files'], { authRequired: true }, {
 		const { offset, count } = this.getPaginationItems();
 		const { sort, fields, query } = this.parseJsonQuery();
 
-		const ourQuery = Object.assign({}, query, { rid: findResult.room._id });
+		// thumbnails don't need to be fetched);
+		const ourQuery = Object.assign({}, query, { rid: findResult.room._id, name: new RegExp('^(?!thumb-).+') });
```
sampaiodiego, line 150: what if the *full* file name actually is `thumb-something.jpg`? we need a better query to filter thumb files.

sampaiodiego (changes requested):

app/api/server/v1/im.js
ostjen: yeah, I thought about that ... but is there a way to query for thumbnails without adding new fields to files?

sampaiodiego: looking at the fields added to thumb files, looks like there is currently no good fields to filter:
https://github.com/RocketChat/Rocket.Chat/blob/13e0983ce2fda48a51f42d9ba9fd5a096d548f07/app/file-upload/server/lib/FileUpload.js#L296

we might need to add a field specifying those files are related to some other file, so we don't need to list them.

server/startup/migrations/v243.ts
```diff
@@ -0,0 +1,43 @@
...
+		const images = Uploads.model.rawCollection().find({ typeGroup: 'image' });
+		const getAspectRatio = (image: Record<string, any>): number => image.identify.size.width / image.identify.size.height;
+
+		// proportion rate error margin 7%
+		const epsilon = 0.07;
+
+		// 2 seconds
+		const thumbCreationTimeout = 2000;
+
+		images.forEach(
```
sampaiodiego, line 17: since you're using `rawCollection` this `forEach` is `async`, and because of that the migration is "finished" **before** it loops over all documents, so the migration keeps running in background even after the servers starts, which is not ideal. I recommend using `fibers/future` to lock the migration during the execution.

app/models/server/models/Uploads.js
```diff
@@ -21,9 +21,7 @@ export class Uploads extends Base {
 			doc.instanceId = InstanceStatus.id();
 		});
 
-		this.tryEnsureIndex({ rid: 1 });
-		this.tryEnsureIndex({ uploadedAt: 1 });
-		this.tryEnsureIndex({ typeGroup: 1 });
+		this.tryEnsureIndex({ rid: 1, uploadedAt: -1, typeGroup: 1, _hidden: 1 });
```
sampaiodiego, line 24: I'm not sure this is the optimal index yet 🤔 

anyways, if we decide to remove the old indexes, we should include a `dropIndex` in the migration.

server/startup/migrations/v243.ts
```diff
@@ -0,0 +1,43 @@
+import { addMigration } from '../../lib/migrations';
+import { Uploads } from '../../../app/models/server';
+
+
+addMigration({
+	version: 243,
+	up() {
+		const images = Uploads.model.rawCollection().find({ typeGroup: 'image' });
+		const getAspectRatio = (image: Record<string, any>): number => image.identify.size.width / image.identify.size.height;
```
sampaiodiego, line 9: trying with an old database it found a some images without `identify` field, which broke the migration

sampaiodiego (changes requested):

ostjen:
nice

apps/meteor/server/startup/migrations/v282.ts
```diff
@@ -0,0 +1,67 @@
...
+	version: 282,
+	up() {
+		// thumbnails were introduced on 3.18 release ( 31 of august ) + add sort
+		const total = Promise.await(
+			Uploads.col.countDocuments({ typeGroup: 'image', uploadedAt: { $gte: new Date('2021-08-31T00:00:00.000Z') } }),
+		);
+
+		Uploads.col.dropIndex('rid_1');
+		Uploads.col.dropIndex('uploadedAt_1');
+		Uploads.col.dropIndex('typeGroup_1');
```
pierre-lehnen-rc, line 63:
```suggestion
```

The original PR was dropping those indexes because they were replaced with a single index with 4 columns.
The PR now has 4 different indexes - one for each column, so dropping them here is not necessary.

Either remove those lines or adjust the index definition in the model.

pierre-lehnen-rc (changes requested):

apps/meteor/server/models/raw/Uploads.ts
```diff
@@ -20,7 +20,7 @@ export class UploadsRaw extends BaseRaw<IUpload> implements IUploadsModel {
 	}
 
 	protected modelIndexes(): IndexDescription[] {
-		return [{ key: { rid: 1 } }, { key: { uploadedAt: 1 } }, { key: { typeGroup: 1 } }];
+		return [{ key: { rid: 1 } }, { key: { uploadedAt: 1 } }, { key: { typeGroup: 1 } }, { key: { _hidden: 1 } }];
```
sampaiodiego, line 23: a better index would be a compound with `rid` and `_hidden`, since both are used together.
```suggestion
		return [{ key: { rid: 1, _hidden: 1 } }, { key: { uploadedAt: 1 } }, { key: { typeGroup: 1 } }];
```

apps/meteor/server/startup/migrations/v282.ts
```diff
@@ -0,0 +1,67 @@
...
+
+	const batch = await Uploads.insertMany(thumbs, { ordered: false });
+
+	if (currentImages.length === batchSize) {
+		return migrateThumbnails(total, current + batchSize);
+	}
+	return batch;
+}
+
+addMigration({
```
sampaiodiego, line 53: I wonder if we should not have a migration at all.. per our new definitions migrations should be included on major versions only..

sampaiodiego (changes requested):
we're also missing tests.

sampaiodiego (approved):

PR #23302, opened 2021-09-28
[BREAK] Remove cordova compatibility setting

ostjen:
Cordova setting was previously deprecated, now is fully removed.


app/lib/server/startup/settings.js
```diff
@@ -916,12 +916,6 @@ settings.addGroup('General', function() {
 		public: true,
 	});
 
-	// Deprecated setting
-	this.add('Support_Cordova_App', false, {
```
rodrigok, line 920: Missing a migration to remove the setting from database

sampaiodiego (approved):

Issue #23285, opened 2021-09-24
Broken links in search when search term is part of the URL


SufiSR:
### Description:
The link in "Search Messages" open to a 404 page.
The URL is opened with <mark> tags.

### Steps to reproduce:
Click in lookup icon to open "Search Messages" box in chat.
Find some link in chat.
Click to open the highlighted link.
![search_issue](https://user-images.githubusercontent.com/83453553/134642920-a5d2ae9f-43e0-49eb-a18f-61d5e2a4bd3a.png)

### Expected behavior:
Link opens correctly.

### Actual behavior:
Link opens a 404 page.

### Server Setup Information:
Version of Rocket.Chat Server: 3.17.2
Operating System: Linux
NodeJS Version: v12.22.1
MongoDB Version: 3.6.14 / wiredTiger (oplog Enabled)

### Client Setup Information
Desktop App or Browser Version: Google Chrome 93.0.4577.82 (Offizieller Build) (64-Bit) / same challenge with Firefox
Operating System: Windows 10 

### Additional context
This was already closed as "could not reproduce" in issue #19369. The screenshot should help to reproduce.

ngulden:
We experience the same behavior in our team with our internal instance.

PR #20878, opened 2021-02-23
[FIX] Message search breaking URL, usertags and emojis

TBG-FR:
**Clean version of #18913** (still worth mentionning it for the discussions in comments about "missing" search results")

## Proposed changes

It has been reported that when searching for a message, the results were displaying with bugs : URLs where broken, emojis title too, ~~and some results are missing~~. This PR aims to fix these issues.

Elements resolved:
- [X] URLs are broken by search `<mark>`
- [X] Emojis are broken by search `<mark>`
- [X] `mailto:` links  are broken by search `<mark>`
- ~~[ ] Some search results are missing (is it really a bug ? See comment on #18913 )~~

## Issue(s)

#18770 => reports that highlighted links, emojis (and mentions ?) are broken, as well as missing results
#18696 => reports that highlighted links are broken
#18456 => report that highlighted links are broken

## How to test or reproduce

Go into a channel, and post these messages :
> smile
> :smile:
> smile.com
> smile-not.com
> smilenot.com

Then search for the word `smile` in that channel. You'll notice that: 
- you can't click on links anymore (they're broken by `<mark>` tags)
- if you hover on the smile emoji, its title is broken by `<mark>` tags too
- the `smilenot.com` link isn't part of the results

## Screenshots
See issues for screenshots

## Types of changes

- [X] Bugfix (non-breaking change which fixes an issue)
- [ ] Improvement (non-breaking change which improves a current function)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Hotfix (a major bugfix that has to be merged asap)
- [ ] Documentation Update (if none of the other choices apply)

## Checklist

- [X] I have read the [CONTRIBUTING](https://github.com/RocketChat/Rocket.Chat/blob/develop/.github/CONTRIBUTING.md#contributing-to-rocketchat) doc
- [X] I have signed the [CLA](https://cla-assistant.io/RocketChat/Rocket.Chat)
- [x] Lint and unit tests pass locally with my changes
- [ ] I have added tests that prove my fix is effective or that my feature works (if applicable)
- [ ] I have added necessary documentation (if applicable)
- [ ] Any dependent changes have been merged and published in downstream modules

#### Broken links/emojis
I fixed the `<mark>` tag issue by doing three `replace` steps
[...]

PR #23309, opened 2021-09-29
[FIX]: Broken links in search

abhinaypandey02:
Added a more precise regex to check and replace the searched text with a highlight. (Ignoring the links in href which was causing the issue)

The following image is taken from the issue #23285.
image
As the user mentioned the `<a>` tag doesn't seem to get the correct URL in the `href` attribute.
The reason being that if a searched text is present in the search result it gets replaced by `<mark>text<mark>`. But this also affectes the attributes. For instance, if we searched for `google.com`, `href="google.com"` would become `href="<mark>google.com</mark>"`, which results in wrong destination.

This was fixed using a more narrow regex search. Here I look for only those matches which does not come between `href="` and  `"`
Following this change the URL points to the right link.

## Issue(s)

#23285 
## Steps to test or reproduce

Reproduction steps mentioned in the issue
## Further comments

My first PR, please suggest better approaches, if any.

### DIFF
--- a/app/ui-message/client/message.js
+++ b/app/ui-message/client/message.js
@@ -46,7 +46,7 @@ const renderBody = (msg, settings) => {
 	}
 
 	if (searchedText) {
-		msg = msg.replace(new RegExp(searchedText, 'gi'), (str) => `<mark>${ str }</mark>`);
+		msg = msg.replace(new RegExp(`(?!href=".)${searchedText}(?!.*")`, 'gi'), (str) => `<mark>${ str }</mark>`);
 	}
 
 	return msg;


TBG-FR:
> > What's the difference with #20878 ?
> > A lot of issues are referencing that problem, and now we have 2 PRs, but still no answers from RocketChat maintainers....
> 
> I am sorry I didn't see that PR, I guess it's better maintained and I should close this one

No problem ! It hasn't been updated for a while though... 😅

Yopai:
Rather than trying to fix the regexp, wouldn't it be more advisable to accept the fact highlighting cannot be simply done with regular expressions ?

See this SO question and answer :

https://stackoverflow.com/a/8193700/2955802

That handle the problem stated here, which is not highlighting everything that is inside tags.

TBG-FR:
Can be closed, fixed by #20878 (merged) 😉

PR #23321, opened 2021-09-30
Regression: LDAP User Data Sync not always working

pierre-lehnen-rc:

### DIFF
--- a/app/importer/server/classes/ImportDataConverter.ts
+++ b/app/importer/server/classes/ImportDataConverter.ts
@@ -9,7 +9,7 @@ import { IImportChannel } from '../../../../definition/IImportChannel';
 import { IConversionCallbacks } from '../definitions/IConversionCallbacks';
 import { IImportUserRecord, IImportChannelRecord, IImportMessageRecord } from '../../../../definition/IImportRecord';
 import { Users, Rooms, Subscriptions, ImportData } from '../../../models/server';
-import { generateUsernameSuggestion, insertMessage, saveUserIdentity } from '../../../lib/server';
+import { generateUsernameSuggestion, insertMessage, saveUserIdentity, addUserToDefaultChannels } from '../../../lib/server';
 import { setUserActiveStatus } from '../../../lib/server/functions/setUserActiveStatus';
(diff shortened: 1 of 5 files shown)


app/importer/server/classes/ImportDataConverter.ts
```diff
@@ -202,43 +214,48 @@ export class ImportDataConverter {
...
-		if (userData.importIds.length) {
-			this.addUserToCache(userData.importIds[0], existingUser._id, existingUser.username);
+		if (userData.name || userData.username) {
+			saveUserIdentity({ _id, name: userData.name, username: userData.username, joinDefaultChannelsSilenced: true });
 		}
 
 		if (userData.avatarUrl) {
 			try {
-				Users.update({ _id: existingUser._id }, { $set: { _pendingAvatarUrl: userData.avatarUrl } });
+				Users.update({ _id }, { $set: { _pendingAvatarUrl: userData.avatarUrl } });
```
sampaiodiego, line 249: I noticed this can be added to `updateData` defined on line 207 as well, right? doesn't need an additional update.

app/importer/server/classes/ImportDataConverter.ts
```diff
@@ -202,43 +214,48 @@ export class ImportDataConverter {
...
 
-		this.updateUserId(userData._id, userData);
+		if (userData.utcOffset) {
+			Users.setUtcOffset(_id, userData.utcOffset);
+		}
 
-		if (userData.importIds.length) {
-			this.addUserToCache(userData.importIds[0], existingUser._id, existingUser.username);
+		if (userData.name || userData.username) {
+			saveUserIdentity({ _id, name: userData.name, username: userData.username, joinDefaultChannelsSilenced: true });
```
sampaiodiego, line 244: the user should be added to default channels only if it's being added, that's the default behavior everywhere else. as I mentioned in other comment, pls don't add this behavior to `saveUserIdentity`..

if you need to add the user to default channels call `addUserToDefaultChannels` directly. probably better call it from `insertUser`.

sampaiodiego (changes requested):

sampaiodiego (approved):

PR #23329, opened 2021-10-01
Chore: Add middlewares to micro services actions

sampaiodiego:

The middlewares implementation is just a boilerplate for now, just to demonstrate how the API works.

This adds the ability to add type-safed middlewares to any micro service actions. Let's say you want to add a middleware to `Room.create`, that's how you'd do it:
```ts
Room.use('create', (uid: string, room: any, /* next */): void => {
  console.log('running before the action', uid, room);

  room.extraData.description = 'overwritten from middleware';

  // next(); // not implemented yet
});
```
This example is already working with current implementation.

Thoughts?

### DIFF
--- a/server/sdk/lib/Api.ts
+++ b/server/sdk/lib/Api.ts
@@ -9,6 +9,8 @@ export class Api {
 
 	private broker: IBroker = new LocalBroker();
 
+	private middlewares: Record<string, Array<(...args: any[]) => any>> = {};
+
 	// set a broker for the API and registers all services in the broker
 	setBroker(broker: IBroker): void {
 		this.broker = broker;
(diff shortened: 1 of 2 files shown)


sampaiodiego:
@d-gubert given your suggestion:

```ts
Service.use('hook_name', (context: IHookContext, next?: Function) => void))
```

how do you think `IHookContext` should look like? or more specifically, how do you think it should represent the original function params? and maybe additional you think we could include now if any?

sampaiodiego:
btw, one thing this proposal doesn't cover, is how to make sure middlewares for external/standalone services are registered.

for example the [Account service](https://github.com/RocketChat/Rocket.Chat/blob/1cc427861954c6f4e626a8da573229d0443b79eb/server/sdk/types/IAccount.ts) that runs standalone when using micro services, if we allow that any code any where can define a middleware, we need to make sure that code is actually deployed along with the standalone service, which can be very tricky..

I was thinking on alternative APIs to solve this problem, one of them could be like [moleculer's mixins](https://moleculer.services/docs/0.14/services.html#Mixins), so you need to explicitly declare on the service definition it's mixins.. wdyt?

PR #23330, opened 2021-10-01
Regression: SAML identifier mapping

pierre-lehnen-rc:


ostjen (approved):

app/meteor-accounts-saml/tests/server.tests.ts
```diff
@@ -42,7 +42,7 @@ import '../../../definition/xml-encryption';
 
 const { expect } = chai;
 
-describe('SAML', () => {
+describe.only('SAML', () => {
```
sampaiodiego, line 45: oopsie

PR #23331, opened 2021-10-01
Regression: LDAP: Handle base authentication and prevent crash

rodrigok:

When AD requires TLS the auth crashes the server if StartTLS is not set, the error shows at the end because the code was not waiting on this operation.

### DIFF
--- a/server/lib/ldap/Connection.ts
+++ b/server/lib/ldap/Connection.ts
@@ -549,8 +549,7 @@ export class LDAPConnection {
 			await this.bindDN(this.options.authenticationUserDN, this.options.authenticationPassword);
 			this.usingAuthentication = true;
 		} catch (error) {
-			authLogger.info({ msg: 'Base Authentication Issue', dn: this.options.authenticationUserDN });
-			authLogger.debug({ msg: 'error', error });
+			authLogger.error({ msg: 'Base Authentication Issue', err: error, dn: this.options.authenticationUserDN });
 			this.usingAuthentication = false;
 		}
 	}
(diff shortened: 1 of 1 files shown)


pierre-lehnen-rc (dismissed):

server/lib/ldap/Connection.ts
```diff
@@ -545,12 +545,18 @@ export class LDAPConnection {
...
 
 		bindLogger.info({ msg: 'Binding UserDN', userDN: this.options.authenticationUserDN });
-		await this.bindDN(this.options.authenticationUserDN, this.options.authenticationPassword);
-		this.usingAuthentication = true;
+		try {
+			await this.bindDN(this.options.authenticationUserDN, this.options.authenticationPassword);
+			this.usingAuthentication = true;
+		} catch (error) {
+			authLogger.info({ msg: 'Base Authentication Issue', dn: this.options.authenticationUserDN });
+			authLogger.debug({ msg: 'error', error });
```
sampaiodiego, line 553: any reason this is not an "error" log? like should this be visible only if log level is set to INFO or DEBUG? if so, just discard my suggestion =)
```suggestion
			authLogger.error({ msg: 'Base Authentication Issue', err: error, dn: this.options.authenticationUserDN });
```

rodrigok: @sampaiodiego no idea, I just followed other examples in the code. @pierre-lehnen-rc WDYT?

pierre-lehnen-rc (approved):

PR #23332, opened 2021-10-01
Regression: Monitor permission to view Livechat rooms

tiagoevanp:


sampaiodiego (commented):
is this really a regression? how was this working before?

KevLehman:
> is this really a regression? how was this working before?

Nope, not a regression. Changed to fix.

PR #23340, opened 2021-10-02
[FIX] Attachments by ignored users not getting ignored

### LINKED ISSUE #23315: File posts by ignored users are still shown
### Description:

If a user whom you are ignoring posts a regular **text message**, it shows the metadata (username and timestamp), but not the message _itself_.  Instead, it shows "This message was ignored".

If that user posts a **file upload**, I would expect the image/document likewise to be concealed, same as for text messages.  However, the file is still shown inline!

### Steps to reproduce:

1. Log in as User A and ignore User B.
2. Log in as User B and post a text message and a file upload (e.g. an image).
3. Log in as User A and view the room.

### Expected behavior:

**Both** of User B's messages should be concealed, with an informative message of "This message was ignored".

### Actual behavior:

User B's text message is concealed, but the file attachment is displayed.

In this screenshot, user "noach" is ignoring user "noach.mp".  The latter posted an image, which is _shown_ rather than _hidden_.
Screen Shot 2021-09-30 at 14 21 26

### Server Setup Information:

- Version of Rocket.Chat Server: 3.17.0
- Operating System: Linux
- Deployment Method: tar
- Number of Running Instances: 30
- DB Replicaset Oplog: enaabled
- NodeJS Version: v12.22.1
- MongoDB Version: 4.2.15 / wiredTiger (oplog Enabled)

### Client Setup Information

- Desktop App or Browser Version: Chrome 93.0.4577.63
- Operating System: MacOS 10.15.7

abhinaypandey02:

As mentioned in #23315 , even after ignoring a user, the attachments were still visible. After having a look, I realised it was because of not verifying the ignored status before displaying the attachments.
So I have added a check in the template to verify that the user is not ignored.

## Issue(s)

Closes #23315

## Steps to test or reproduce

Mentioned in the issue
## Further comments

Please suggest if there is a better approach.

### DIFF
--- a/app/ui-message/client/message.html
+++ b/app/ui-message/client/message.html
@@ -123,7 +123,9 @@
 				{{/each}}
 			{{/if}}
 			{{#if hasAttachments}}
-				{{> reactAttachments attachments=msg.attachments file=msg.file }}
+				{{#unless isIgnored}}
+					{{> reactAttachments attachments=msg.attachments file=msg.file }}
+				{{/unless}}
 			{{/if}}
 
 


nmagedman:
This has been fixed by other means and may be closed.

Issue #23250, opened 2021-09-21
Adding Users via Federation stopped working in 3.18.1


qwertiko:
### Description:

Adding Users via Federation is not working after Updating from 3.17.1 to 3.18.1.
The error message "invalid user" is given.

### Steps to reproduce:

1. Go to Directory
2. Go to External Users
3. Search for Federated User
4. Click on the User to start a room

### Expected behavior:

User is added and room created

### Actual behavior:

The message "invalid user" is displayed

### Server Setup Information:

- Version of Rocket.Chat Server: 3.18.1
- Operating System: Ubuntu 20.04
- Deployment Method: tar
- Number of Running Instances: 2
- DB Replicaset Oplog: yes
- NodeJS Version: v12.22.1
- MongoDB Version: 4.2.14

### Client Setup Information

- Desktop App or Browser Version: both/Firefox/Chrome
- Operating System: manjaro/Fedora

### Additional context

Debug Log show the following:
Sep 21 12:10:37 laura rocketchat[14765]: server.js:204 Federation ➔ client.error afterCreateDirectRoom => Could not create federated room: TypeError: Cannot read property '0' of undefined
Sep 21 12:10:37 laura rocketchat[14765]:     at Object.normalizeUser (app/federation/server/normalizers/user.js:30:16)
Sep 21 12:10:37 laura rocketchat[14765]:     at app/federation/server/hooks/afterCreateDirectRoom.js:34:41
Sep 21 12:10:37 laura rocketchat[14765]:     at Array.map (<anonymous>)
Sep 21 12:10:37 laura rocketchat[14765]:     at app/federation/server/hooks/afterCreateDirectRoom.js:33:51
Sep 21 12:10:37 laura rocketchat[14765]:     at /opt/Rocket.Chat/programs/server/npm/node_modules/meteor/promise/node_modules/meteor-promise/fiber_pool.js:43:40

Since the error occured in the user.js file trying to get the property emails at the following place:
  const resource = _.pick(originalResource, '_id', 'username', 'type', 'emails', 'name', 'federation', 'isRemote', 'createdAt', '_updatedAt');
  const email = resource.emails[0].address;

I therefore logged via the console the array "resource" and got the following back:

{ _id: 'LdJc2Ad2iNrqtuw3L', username: 'USERNAME_REDACTED' }

There is clearly no email address that can be accessed in this array. But i am not sure why this is not the case as the user has every attribute set in the database.

col-panic:
[...]

PR #23347, opened 2021-10-03
[FIX] creating room with federated member

qwertiko:
When creating a room with a federated member the local user only has the attributes "_id" and "username" for the event normalizeUser.
In there the attribute emails is called but only defined two lines down.
Moved the definition before the call to resolve  "undefined user" error message on room creation.

## Issue(s)
#23250

## Steps to test or reproduce
Have two servers setup with RocketChat which communicate via Federation.
Create a room with a another user from the other server.


2disbetter:
Would be nice if we could get this reviewed and merged.

bkraul:
This issue persists on 4.1.0-develop. Please merge.

bkraul:
Is there any particular reason why thys PR can't be reviewed and merged?

bkraul (approved):

alansikora (approved):

PR #23372, opened 2021-10-05
[FIX] unwanted toastr error message when deleting user

ostjen:


app/lib/server/functions/getFullUserData.js
```diff
@@ -89,6 +89,10 @@ export function getFullUserDataByIdOrUsername({ userId, filterId, filterUsername
 		fields,
 	};
 	const user = Users.findOneByIdOrUsername(filterId || filterUsername, options);
+	if (!user) {
+		return;
```
tassoevan, line 93: As the function is expected to return _something_, it's nice to be a little more explicit on `return` statements:

```suggestion
		return null;
```

tassoevan (changes requested):

tassoevan (approved):

PR #23384, opened 2021-10-06
[FIX] Upgrade Highlight.js to latest major

tassoevan:

Bumps `highlight.js` to the latest version, as 9.x reached EOL.


tassoevan:
> You have added 1 .js files, please convert to ts(x). imports/client/highlight.js

It's not a file: it's a symlink to a directory named `highlight.js` ಠ_ಠ

tassoevan:
There is a **major bug** on Meteor's bundler preventing us to upgrade dependencies and go ahead with this change.

PR #23392, opened 2021-10-07
[IMPROVE] Allow Omnichannel to handle huge queues 

KevLehman:

### DIFF
--- a/server/startup/migrations/index.ts
+++ b/server/startup/migrations/index.ts
@@ -65,6 +65,5 @@ import './v238';
 import './v239';
 import './v240';
 import './v241';
-
-import './v243';
+import './v242';
 import './xrun';
--- a/server/startup/migrations/v242.ts
+++ b/server/startup/migrations/v242.ts
@@ -22,7 +22,7 @@ function removeSetting(): void {
 }
 
 addMigration({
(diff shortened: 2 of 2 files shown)


murtaza98:
@KevLehman JFYI: There are some conflicts here :grimacing:

murtaza98:
Recommending some cleanup tasks since we now no longer need `estimatedInactivityCloseTimeAt` property on inquiry docs.

We won't need these 2 methods anymore 
- [setEstimatedInactivityCloseTime](https://github.com/RocketChat/Rocket.Chat/blob/fe21ede7a1efa5e82c1f6fd66dffd4591de45969/app/models/server/models/LivechatInquiry.js#L244)
- [unsetEstimatedInactivityCloseTime](https://github.com/RocketChat/Rocket.Chat/blob/fe21ede7a1efa5e82c1f6fd66dffd4591de45969/app/models/server/models/LivechatInquiry.js#L252)

We can also remove this property from the search filter from [here](https://github.com/RocketChat/Rocket.Chat/blob/fe21ede7a1efa5e82c1f6fd66dffd4591de45969/app/models/server/models/LivechatInquiry.js#L52)

PR #20870, opened 2021-02-22
Improve: Switch storage in docker-compose

geekgonecrazy:
## Proposed changes (including videos or screenshots)

Removed the option to force --storageEngine=mmapv1  as we need to get people using wiredtiger

PR #23402, opened 2021-10-07
[FIX] Update Docker-compose.yml for Mongodb 4.2

0xsuid:

- Updated MongoDB Version to 4.2 
- changed storage engine to wiredTiger from mmapv1(mmapv1 no longer supported by MongoDB 4.2)
- smallfiles argument is also removed because it is also not supported by MongoDB 4.2.

## Issue(s)
Mongodb version 4.0 is deprecated and thus it is not compatible with latest release of RocketChat.
image

## Steps to test or reproduce
Run previous/older docker-compose file and rocket-chat will show following message:
```MongoDB version 4.0.27 is deprecated, please upgrade your installation.```

## Further comments
Mongodb storage engine migration guide [link](https://docs.rocket.chat/quick-start/installing-and-updating/docker-containers/mongodb-mmap-to-wiredtiger-migration) should be posted with deprecation message so it will be easier for administrator to update rocketchat. However, migration guide docker-compose.yml also needs to be updated for mongodb - from 4.0 to 4.2.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -31,19 +31,19 @@ services:
       - "traefik.frontend.rule=Host: your.domain.tld"
 
   mongo:
-    image: mongo:4.0
+    image: mongo:4.2
     restart: unless-stopped
     volumes:
      - ./data/db:/data/db
      #- ./data/dump:/dump
-    command: mongod --smallfiles --oplogSize 128 --replSet rs0 --storageEngine=mmapv1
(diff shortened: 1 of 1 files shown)


sampaiodiego:
thanks @0xsuid for the contribution.. we were some discussions in https://github.com/RocketChat/Rocket.Chat/pull/20870 on how is the way to actually update this file..

I'll try to come up with an end to that discussion so we can move this forward 👍

PR #23425, opened 2021-10-09
Chore: entrypoint script for docker image

debdutdeb:
The current loop assumes every error to be a connection failure, where that isn't necessarily the case.

One very common one is authentication failure. In such situations it's not quite logical to continue retrying whereas
the process should exit and let the admin fix the credentials first (or whatever the other issue should be).

This script filters out the actual connection failure & replicaset primary selection error and retries ONLY in these cases. Otherwise simply prints the issue and terminates the process.

Another benefit is by implementing the retry logic in the application container image we get rid of the necessity of implementing the same in/through different deployment methods. One immediate example that I can think of is caprover, which uses a compose like structure for its one click apps, but doesn't support the command key (iow expects the retry logic to be part of the image).

### DIFF
--- a/.docker/Dockerfile
+++ b/.docker/Dockerfile
@@ -46,4 +46,4 @@ ENV DEPLOY_METHOD=docker \
 
 EXPOSE 3000
 
-CMD ["node", "/app/entrypoint.js"]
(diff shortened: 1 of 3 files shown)


sampaiodiego:
I'm not sure about what "loop" you're talking about.

PR #23450, opened 2021-10-13
[IMPROVE] Support Markdown on Livechat Custom Message Setting

murtaza98:

Since we're now adding support for markdown text on Livechat offline Instructions message, we're adding this PR to provide a markdown editor in place of the generic text setting for the corresponding livechat setting
image

## Further comments

Note: I've moved the Markdown components I needed for this PR out of the EE folder and into our general community folder since this setting is also present on community version


KevLehman:
It works 🤔 but I feel the interaction a bit buggy

- Cut (ctrl-x) sometimes work, sometimes doesn't
- Link button doesn't work
- Sometimes, when deleting a long text, you're left with just one character that you cannot remove

PR #23461, opened 2021-10-14
[FIX] Undefined emoji on parsed message

tiagoevanp:


dougfabris:
With the new message parser, it doesn't make sense anymore!

PR #22588, opened 2021-07-06
[FIX] Direct Reply

ggazzo:
## Steps to test or reproduce

https://docs.rocket.chat/guides/administration/administration/settings/email/direct-reply

PR #23521, opened 2021-10-20
[FIX] Delay start of email inbox

KevLehman:


ggazzo:
and please check if you can https://github.com/RocketChat/Rocket.Chat/pull/22588 
I theory when this feature was developed the changes on the classes just broke direct reply(on team collaboration) making the feature unusable

PR #23533, opened 2021-10-21
[FIX] New specific endpoint for contactChatHistoryMessages with right permissions

tiagoevanp:

Anyone with 'View Omnichannel Rooms' permission can see the History Messages.


KevLehman (dismissed):

KevLehman (dismissed):

app/models/server/models/Messages.js
```diff
@@ -1260,6 +1260,19 @@ export class Messages extends Base {
 		return this.find(query);
 	}
 
+	findLivechatClosedMessages(rid, options) {
+		return this.find(
+			{
+				rid,
+				$or: [
+					{ t: { $exists: false } },
+					{ t: 'livechat-close' },
```
murtaza98, line 1269:
```suggestion
```
This condition has no effect because of the previous OR condition i.e. `t: { $exists: false }`, so I think we can safely remove it. In fact, this entire OR condition isn't required. we can simply keep the filter as 
```
{
     rid,
     t: { $exists: false },
}
```
and this should give you only the text messages for this room

KevLehman:
@ggazzo , endpoint was moved to TS, could you take a look 👀 ?

KevLehman:
@ggazzo pipeline is pasing at last! 🎉

ggazzo (approved):

PR #23468, opened 2021-10-15
[FIX]  Fixed E2E default room settings not being honoured

ostjen:

PR #23544, opened 2021-10-24
[FIX] Resolves issue with E2E default settings on private and Direct

TheDigitalEagle:
Change Log:
- Fixed Typo with E2E_Enable in the E2E server settings
- Made property public.

This has been tested and confirmed working in our internal systems. 

Additional notes:
We have further implemented Enforced E2E in our own company build for Private and DM (meaning users cannot disable). We'd be happy to upstream if there's interest as we're sure there's customers who require that as much as we did.


tassoevan:
Closing it in favor of #23468. I've merged commits from this branch. I believe @ostjen will update some tests and everything will be ready to merge.

PR #23547, opened 2021-10-25
[IMPROVE] Engagement Dashboard

tassoevan:

- Adds helpers `onToggledFeature` for server and client code to handle license activation/deactivation without server restart;
- Replaces usage of `useEndpointData` with `useQuery` (from [React Query](https://react-query.tanstack.com/));
- Introduces `view-engagement-dashboard` permission.


matheusbsilva137 (changes requested):
The "Users by time of day" chart is now displaying 5 days instead of 7 when using Local Timezone.

## Before (`develop`):
print-ed-local-before

## After:
print-ed-local-after

ggazzo (approved):

PR #23587, opened 2021-10-28
[FIX] Omnichannel status being changed on page refresh

KevLehman:

### DIFF
--- a/app/livechat/server/business-hour/AbstractBusinessHour.ts
+++ b/app/livechat/server/business-hour/AbstractBusinessHour.ts
@@ -47,7 +47,7 @@ export abstract class AbstractBusinessHourBehavior {
 	}
 
 	async changeAgentActiveStatus(agentId: string, status: string): Promise<any> {
-		return this.UsersRepository.setLivechatStatus(agentId, status);
+		return this.UsersRepository.setLivechatStatusIf(agentId, status, { livechatStatusSystemModified: true }, { livechatStatusSystemModified: true });
 	}
 }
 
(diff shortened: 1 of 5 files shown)


cauefcr:
I've tested this commit and it fixes the issue highlighted in https://app.clickup.com/t/8z67gc

murtaza98:
Hi, @KevLehman I noticed we have a duplicate method - `setLivechatStatusIf`, in the following 2 files. Both of them have identical logic and params.
- [models folder](https://github.com/RocketChat/Rocket.Chat/blob/fix/omni-status-available-on-refresh/app/models/server/models/Users.js#L272)
- [Raw model folder](https://github.com/RocketChat/Rocket.Chat/blob/b67c612635978feff76cc384d3e7adc08bf11bf9/app/models/server/raw/Users.js#L262)

So I was wondering if we can it from one of them? Preferably from the models folder since we're transitioning to using raw models

cauefcr (approved):
There has been no regression of functionality regarding the new changes, it properly saves the state of the Live Chat agent between sessions.

sdelcroix:
Hi,

please excuse me if this PR is not related to the following issue I have, I guessed that just because of the title.

Indeed when I put myself offline for Omnichannel and I refresh my browser, my status switch to online. You can see that behavior on the following video.
I've just updated my Rocket.Chat server snap to 4.1.2.

https://user-images.githubusercontent.com/6229909/145061016-8a15afa0-c51d-40c7-8919-389d7a4400a3.mov

KevLehman:
Hey @sdelcroix , do you have other things enabled? Like, business hours?

sdelcroix:
@KevLehman Yes we have business hours. But in the case show in the video :

- I'm outside the business hours
- the account is an admin and is not member of any omnichannel department 

I've also tried with a user with "livechat-agent" role, member of an omnichannel department, outside of business hours and the issue occurs.

PR #23591, opened 2021-10-29
Chore: Removing hubot from docker-compose

geekgonecrazy:

Remove hubot from docker-compose.  This is forcing everyone to spin up Hubot every time they deploy Rocket.Chat and not that many people are using it.  So we are wasting resources on peoples machines by forcing it


debdutdeb:
@geekgonecrazy I already have a pr for the compose file update. 

Also just opened another one for the development environment setup using docker compose #23594

debdutdeb (approved):

PR #23594, opened 2021-10-30
Chore: platform agnostic development environment setup using docker compose

debdutdeb:

## Further comments

Testing this PR - 

existing local repo

```
git fetch origin pull/23594/head:pr-23594
git checkout pr-23594
docker compose run build
```

else

```
git clone -b dev-env-setup https://github.com/rocketchat/rocket.chat
cd rocket.chat
docker compose run build
```

### DIFF
--- a/.docker/Dockerfile.meteor
+++ b/.docker/Dockerfile.meteor
@@ -18,7 +18,7 @@ RUN apt update && \
 USER meteor
 WORKDIR /home/meteor
 
-COPY --from=0 /.meteor .meteor
+COPY --from=meteor-download /.meteor .meteor
 
 VOLUME [ "/home/meteor/.meteor/packages" ]
 
--- a/.docker/Dockerfile.meteor
+++ b/.docker/Dockerfile.meteor
@@ -1,0 +1,25 @@
+FROM ubuntu:focal
+
+ARG RELEASE
+
+ADD https://s3.amazonaws.com/com.meteor.static/packages-bootstrap/$RELEASE/meteor-bootstrap-os.linux.x86_64.tar.gz /
+
+RUN tar xzf /meteor-bootstrap-os.linux.x86_64.tar.gz
+
+FROM ubuntu:focal
+
+# Rocket.Chat build dependencies: g++ git build-essential (make + etc.) python
+RUN apt update && \
+    apt install --no-install-recommends -y \
+        g++ sudo git build-essential python && \
+    useradd -mG sudo meteor && \
+    echo '%meteor ALL=(ALL) NOPASSWD: ALL' > /etc/sudoers.d/meteor
+
+USER meteor
+WORKDIR /home/meteor
+
+COPY --from=0 /.meteor .meteor
+
+VOLUME [ "/home/meteor/.meteor/packages" ]
+
+ENV PATH=/home/meteor/.meteor:$PATH
--- a/.docker/Dockerfile.meteor
+++ b/.docker/Dockerfile.meteor
@@ -1,4 +1,4 @@
-FROM ubuntu:focal
+FROM ubuntu:focal as meteor-download
 
 ARG RELEASE
 
--- a/.docker/Dockerfile.meteor
+++ b/.docker/Dockerfile.meteor
@@ -2,9 +2,9 @@ FROM ubuntu:focal
 
 ARG RELEASE
 
-ADD https://s3.amazonaws.com/com.meteor.static/packages-bootstrap/$RELEASE/meteor-bootstrap-os.linux.x86_64.tar.gz /
+ADD https://packages.meteor.com/bootstrap-link?arch=os.linux.x86_64&release=$RELEASE /meteor.tar.gz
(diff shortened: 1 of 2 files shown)


.docker/Dockerfile.meteor
```diff
@@ -0,0 +1,25 @@
+FROM ubuntu:focal
+
+ARG RELEASE
+
+ADD https://s3.amazonaws.com/com.meteor.static/packages-bootstrap/$RELEASE/meteor-bootstrap-os.linux.x86_64.tar.gz /
```
geekgonecrazy, line 5: Is there a non-s3 path we could link to?  I know if it was our s3 I wouldn't want people linking direct to it like this.  Can be expensive.  AWS likes to get you on bandwidth

.docker/Dockerfile.meteor
```diff
@@ -0,0 +1,25 @@
...
+RUN apt update && \
+    apt install --no-install-recommends -y \
+        g++ sudo git build-essential python ca-certificates && \
+    useradd -mG sudo meteor && \
+    echo '%meteor ALL=(ALL) NOPASSWD: ALL' > /etc/sudoers.d/meteor
+
+USER meteor
+WORKDIR /home/meteor
+
+COPY --from=0 /.meteor .meteor
```
geekgonecrazy, line 21: Is the `--from` here needed?  I thought only needed when reversing a different stage of a multi-stage dockerfile

docker-compose.yml
```diff
@@ -1,102 +1,28 @@
...
-  #  ports:
-  #    - 80:80
-  #    - 443:443
-  #  volumes:
-  #    - /var/run/docker.sock:/var/run/docker.sock
+      - type: bind
+        source: .
+        target: /mnt
+      - type: volume
+        source: packages
```
geekgonecrazy, line 21: Should this instead be .meteor/packages as well.  That way it's not another folder to ignore in gitignore?

.docker/Dockerfile.meteor
debdutdeb: I took this directly from the install script - https://install.meteor.com - line no. 133.

Since their script is using it I figured we could too. Not sure if we can use some other endpoint redirect in this case.

debdutdeb: JFYI - the new `npm i -g meteor` doesn't work for meteor@2.2 which we need (`npm i -g meteor@2.3.1` see https://www.npmjs.com/package/meteor for version correlation between the npm package and meteor).

.docker/Dockerfile.meteor
debdutdeb: It is reversing the first stage. The downloaded archive isn't needed in the final image. 

The same *can* be achieved without the multi-stage way - but this just felt slightly more natural to me (also was my first thought).

docker-compose.yml
debdutdeb: I'm not sure what you mean. I'm not yet a 100% well versed in meteor build system. The only data that I found needed to be saved across runs was `$HOME/.meteor/packages`.

.docker/Dockerfile.meteor
debdutdeb: Ok - scratch that - found one https://github.com/meteor/meteor/blob/290bf55e3c8c002ff35c0f9553383d7a2f9f4c2e/npm-packages/meteor-installer/install.js#L43

https://packages.meteor.com/bootstrap-link?arch=os.linux.x86_64&release=2.2 

This should work :)

docker-compose.yml
debdutdeb: I'm not sure what you mean

.docker/Dockerfile.meteor
```diff
@@ -0,0 +1,25 @@
+FROM ubuntu:focal
```
geekgonecrazy, line 1:
```suggestion
FROM ubuntu:focal as meteor-download
```

.docker/Dockerfile.meteor
```diff
@@ -0,0 +1,25 @@
...
+RUN apt update && \
+    apt install --no-install-recommends -y \
+        g++ sudo git build-essential python ca-certificates && \
+    useradd -mG sudo meteor && \
+    echo '%meteor ALL=(ALL) NOPASSWD: ALL' > /etc/sudoers.d/meteor
+
+USER meteor
+WORKDIR /home/meteor
+
+COPY --from=0 /.meteor .meteor
```
geekgonecrazy, line 21:
```suggestion
COPY --from=meteor-download /.meteor .meteor
```

nileshtrivedi:
[This comment](https://github.com/caprover/one-click-apps/issues/66#issuecomment-963867908) might be relevant:

> It is recommended to add a password to Mongo's configuration. After all, passwords are required in most cases.

> I tried several existing configurations, but I couldn't complete the password configuration without manual operation. I hope the official can make a configuration so that we can deploy the environment with password only by changing the password.

PR #20234, opened 2021-01-16
[FIX] Line breaks with Marked Markdown

Karting06:
## Proposed changes (including videos or screenshots)

Fix missing line breaks when using Marked Markdown. This also fixes ASCII emojis that weren't converted to an emoji.

## Issue(s)

Fixes: #11675 (line break issue)
Fixes: #11214 (emoji)

## Steps to test or reproduce

See details for the line break issue : #11675
## Further comments

The implementation of the _Original_ Markdown in Rocket.Chat doesn't add a paragraph HTML tag to each message (`<p>message</p>`) while Marked.js did. I changed the paragraph renderer to fix that. This also fixed the problem with ASCII emojis (`:)`) not being parsed by the emoji parser.

Marked Markdown also parsed multiple line breaks as a new paragraph while Original Markdown doesn't. I slightly changed the paragraph rule of Marked Markdown (GFM & normal) in order to keep 1 paragraph. The line breaks are now consistent with Original Markdown.

I created one pull request for both fixes in order to avoid to break the line break workaround.

Issue #11675, opened 2018-08-04
[Marked MD Parser] Line break in messages


a5his:
### Description:
Line break is not working as expected. I have chose `Marked` from the Markdown parser options through settings. This works with `Original` option. After going through the code I figured out that the line breaks `\n` are removed and messages are rendered inside `p` tag by below code from file `package/rocketchat-markedown/parser/marked/marked.js`.
```
msg.html = _marked(s.unescapeHTML(msg.html), {
		gfm,
		tables,
		breaks,
		pedantic,
		smartLists,
		smartypants,
		renderer,
		sanitize: true,
		highlight
	});
```
Just wanted to confirm that whether is intentional or not? In addition, the sub options under `Marked` 
`Enable Marked Breaks` setting is not affecting anything in the view either you chose `true` or `false`. 

Since `p` tag is used to render these message, is this UI bug?

### Steps to reproduce:

1. Go to any channel and type your message with line breaks using `Shift` plus `Enter` key.

### Expected behavior:

The line break should have actually separated the messages.

### Actual behavior:

The line break is not separating the messages.

### Server Setup Information:

- Version of Rocket.Chat Server: 0.63.3
- Operating System: Darwin
- Deployment Method: 
- Number of Running Instances: 1
- DB Replicaset Oplog: Enabled
- NodeJS Version: v8.11.1
- MongoDB Version: 

### Screenshot
![markdown](https://user-images.githubusercontent.com/330362/43676034-1b3ed720-9809-11e8-8394-c9c130c6b5d2.gif)

felixmind:
This is a real problem. And it does not allow to work well.

naiaokami:
Yes; this is an issue.

horstepipe:
Hello team,
just to verify, is there no way to insert a line break into a message at the moment?

reetp:
As far as I am aware this was fixed ages ago and probably should be closed.
Seems to work perfectly at open.rocket.chat 
Please test and advise.

horstepipe:
no this doesn't work when you set markdown to "marked". open.rocket.chat seems to have it set to "original", there the line breaks work, but most other markdowns don't work when you have set it to "original".

horstepipe:
So even a simple list like 
`
* list item 1
* list item 2
`

is not being rendered properly when you set it to "original"

gabriellsh:
Still happens 3.0.0-develop

sst2010:
[...]

Issue #11214, opened 2018-06-21
"Marked" Markdown parser interfering with Emoji transformation 


frdmn:
### Description:

When you switch from `Original` to  `Marked` Markdown parser in the administration UI (Messages → Markdown → Markdown Parser), normal text emoticons (like `:)`, `:-(` or `:o`) won't be transformed into their corresponding Emoji counterpart

### Steps to reproduce:

1. Switch to `Marked`
2. Type `:)` 
3. Look at the text emoticon rather than Emoji

### Expected behavior:

Transform text emoticon to Emoji

### Actual behavior:

![bildschirmfoto-2018-06-21-um-16 17](https://user-images.githubusercontent.com/944459/41725069-4127ce7c-756f-11e8-9670-452d3bb0d0b2.jpg)

### Server Setup Information:

- Version of Rocket.Chat Server: 0.65.1
- Operating System: Ubuntu
- Deployment Method: Docker
- Number of Running Instances: 1
- DB Replicaset Oplog: yes
- NodeJS Version: dockerized
- MongoDB Version: dockerized

chefino:
Running manual install on Ubuntu 20.04, RC version 3.0.12 and issue still persists :-/
Is there some alternative Markdown processor instead of the featureless Original or buggy Marked?

dougfabris:
Closing this stale issue, since it's already been fixed since the new message parser implementation!

PR #23596, opened 2021-10-30
[IMPROVE] Upgrade Marked to 3.0.8 + fix linebreaks

Karting06:

- Updated Marked to current version (3.0.8) & types package.
- Implements the fix of #20234 
- Based on PR of @dougfabris (#23380)

## Issue(s)

Fixes: #11675 (line break issue)
Fixes: #11214 (emoji)
Closes: #20234 
Closes: #23380
## Steps to test or reproduce

See details for the line break issue : #11675

## Further comments

The implementation of the _Original_ Markdown in Rocket.Chat doesn't add a paragraph HTML tag to each message (`<p>message</p>`) while Marked.js did. I changed the paragraph renderer to fix that. This also fixed the problem with ASCII emojis (`:)`) not being parsed by the emoji parser.

Marked Markdown also parsed multiple line breaks as a new paragraph while Original Markdown doesn't. I slightly changed the paragraph rule of Marked Markdown (GFM & normal) in order to keep 1 paragraph. The line breaks are now consistent with Original Markdown.

Regarding the new Marked.js version:
The current version of Marked.js don't have separated `em` & `strong` rules anymore instead they use `emStrong`. 
I couldn't migrate the custom `inlineLexer` rules for `strong` & `em`. I also don't really know where they are used in order to test it.


danel1:
@ggazzo @dougfabris can we get this in a release soon? Would be awesome!

dougfabris:
@Karting06 Thanks a lot for your help. We discussed internally and we are planning to deprecate the marked in the future, replacing it with our own marked solution. Because of that we won't upgrade the marked package. Due this decision I'll close this one and I'm look forward to merge the fixes you have done here: #20234

Karting06:
Ok, thanks for the follow-up @dougfabris. 
I hope that Rocket.Chat's marked solution will be similar to the "normal" marked ;)

danel1:
Sorry @dougfabris to disturb you again. Do you really think it's good idea to "re-invent the wheel"? Why not use a framework that is known and well developed? You're just adding another component that you have to maintain yourselves. I think there are enough problems and bugs with self-developed stuff already and ever new release brings more problems than fixing.. IMHO this decision should be overthinked...

Issue #23593, opened 2021-10-30
German (and possibly other) translations of user activity wrong since 4.1.0


chotaire:
### Description:

German translation of user activity is wrong since 4.1.0:

`$USER is schreibt` should be: `$USER schreibt`
`$USER1 und $USER2 are schreiben` should be: `$USER1 und $USER2 schreiben`

This may very well be broken in other translations too.

### Steps to reproduce:

1. Install Rocket.Chat 4.1.0 and switch to German
2. Add three users
3. Watch one and/or two users writing on channel

### Expected behavior:

See description

### Actual behavior:

See description

### Server Setup Information:

- Version of Rocket.Chat Server: 4.1.0
- Operating System: Linux
- Deployment Method: Docker
- Number of Running Instances: 1
- DB Replicaset Oplog: active
- NodeJS Version: v12.22.1
- MongoDB Version: 5.0.3 / wiredTiger

### Client Setup Information

- Desktop App or Browser Version: Any
- Operating System: Any

### Additional context

None

### Relevant logs:

None

chotaire:
According to https://github.com/RocketChat/Rocket.Chat/commit/428d70ae5f3c3d58686d683f0dca635c8f191440 this is not a problem with the actual translation, the problem lays somewhere inside the code.

Karting06:
I agree with you.
Starting with this [PR](https://github.com/RocketChat/Rocket.Chat/pull/22392), the translation isn't linked to the action anymore. 
It's like this: `is/are + action` and both are translated as 2 distinct words which lead to strange things in other language (same problem in French & Dutch). (And "is" & "are" aren't present in the English translation file, so the translations are missing for all languages...)


I'm also a translator on the Lingohub project. When translating, you don't have a context for the words which doesn't help (especially in this case). For French & Dutch, I changed the translation in order to correct it. 
But I don't think that there is an alternative in German (something like `ist + action`)? (in Dutch, I used `is aan het typen`)

hexabinaer:
@Karting06 thanks, you comment has been very helpful for me. I'd already been halfway through the on-boarding process to Lingohub just for this issue.

We've already come up with creative interpretations like informal/dialect interpretations in our internal team channel. 
>"is am Schreiben dranne"

Thanks for taking care of German translations, very much appreciated (several people maintaining Drupal translations say hello).

Karting06:
Hi @hexabinaer 
[...]

PR #23597, opened 2021-10-30
[FIX] Add missing is/are translation for user actions

Karting06:

Add missing `is` & `are` translation key in order to better translate the `$USER1 is typing`. 

## Issue(s)

It doesn't fully fixed it but it fixes it for some languages: #23593
## Steps to test or reproduce

See: #23593


chotaire:
Not all languages feature present progressive I believe. This will likely require "is" and "are" to be left empty in affected language translations, however translators will just fill them because the strings exist. 

This entire PR that changed typing indicators complicates things to an extreme and I would rather see that reverted. There should not be a case where message typing indicators demand present progressive, and then the strings are even split. This really isn't good at all!

As an example, there is a present progressive in German, however in the case of "is writing" or "are writing" it would actually not be used because it just wouldn't sound right.

chotaire:
Oh, despite my concerns, @Karting06 the translations you added also need to be referenced within the code. For now this shouldn't do anything.

Karting06:
@chotaire It does work (tested in French). 
It's used [here](https://github.com/RocketChat/Rocket.Chat/commit/a0249005961091c06b820d807d42e36f75be4996#diff-c825f97d19958626d81c98c83366f003657fe9631c93609ec62add90d5bf4a29)

Karting06:
But, I completely agree with you. 
The core code should be rewritten instead of simplifying to `is + action`.

dougfabris:
> But, I completely agree with you. The core code should be rewritten instead of simplifying to `is + action`.

@Karting06 Would you like to help us with that? 😄

Karting06:
@dougfabris, okay. But that will be in a few weeks ;)

ankar84:
Any news guys?

TBG-FR:
I think you got it right, it should be reworked to create `[is] action` strings for each `action`
- That will handle languages using `is action` and languages using only `action`
- That will give more context to people doing translations, hence getting them right (for exemple `recording` is mistranslated in French because of that)

I'd be willing to do these modifications, if anyone has already some ideas or guidelines, feel free to give them 😄 

While I'm here, how do you apply to be able to help translating RocketChat in LingoHub ? I sent a mail years ago and never been added 😅

ankar84:
> While I'm here, how do you apply to be able to help translating RocketChat in LingoHub ? I sent a mail years ago and never been added

Ask @sampaiodiego at open to send invite link. He invited me to LingoHub.

TBG-FR:
@sampaiodiego If you can send me an invitation, I'd be happy to contribute !
You can also give my your thoughts about #27004 😉

TBG-FR:
@Mugros @ankar84 I'd be happy to have your review 😄

TBG-FR:
@dougfabris Thank you for your reactivity ! I didn't receive any invitation, but my account wasn't created with my Github public email and wasn't linked to Github, I fixed it, can you invite me again ? Thanks a lot !

dougfabris:
@TBG-FR Already sent to your GitHub public email 😃 
Thanks for helping us!

Karting06:
Hi @TBG-FR, thanks for taking the time to fix it.
Didn't have the time in the last weeks.

Nice to see that it will be solved.
Thanks @dougfabris

TBG-FR:
Hello @dougfabris , sorry to bother you but I can now access the RocketChat project on LingoHub, but I can't do anything
- Dashboard displays "0 targeted language"
- When I try to translate an (not yet translated) element, it's greyed out

Can you check the perms ? The new RocketChat version introduced new strings, and they need translations 😅 

Many thanks in advance ! 🙏

dougfabris:
@TBG-FR I changed your permissions on the projects, can you check it again?
Thanks for helping us, mate!

PR #23607, opened 2021-11-01
[FIX] App update flow failing in HA setups

d-gubert:

The flow for app updates is broken in specific scenarios with HA setups. Here we change the method calls in the Apps-Engine to avoid race conditions


thassiov (approved):
LGTM!

PR #23608, opened 2021-11-01
[FIX] Advanced LDAP Sync Features

pierre-lehnen-rc:
Issues fixed:
 - [X] Background Sync: LDAP Connection was closed before the advanced searches were performed
 - [X] LDAP user's roles were always changed to `user` whenever their data was updated on Rocket.Chat.
 - [X] Team Sync was only compatible with a very specific LDAP configuration.
 - [X] When debug logs were active, LDAP sync was logging image data as a giant encoded string, polluting the log file with useless information.


matheusbsilva137 (dismissed):

app/importer/server/classes/ImportDataConverter.ts
```diff
@@ -235,7 +242,13 @@ export class ImportDataConverter {
...
 		this.addUserImportId(updateData, userData);
 		this.addUserEmails(updateData, userData, existingUser.emails || []);
-		Users.update({ _id }, updateData);
+
+		if (Object.keys(updateData.$set).length === 0) {
+			delete updateData.$set;
+		}
+		if (Object.keys(updateData).length > 0) {
+			Users.update({ _id }, updateData);
+		}
```
sampaiodiego, line 251: I think changing it to a single statement makes it easier to read, wdyt? 
```suggestion
		if (Object.keys(updateData.$set).length > 0) {
			Users.update({ _id }, updateData);
		}
```

ee/server/lib/ldap/Manager.ts
```diff
@@ -45,13 +45,13 @@ export class LDAPEEManager extends LDAPManager {
 				} else if (updateExistingUsers) {
 					await this.updateExistingUsers(ldap, converter);
 				}
+
+				converter.convertUsers({
```
sampaiodiego, line 49: maybe it is even better removing the inner try/finally? it'd work having a single one. also easier to understand.

ee/server/settings/ldap.ts
```diff
@@ -200,20 +200,34 @@ export function addSettings(): void {
...
-				this.add('LDAP_Query_To_Get_User_Teams', '(&(ou=*)(uniqueMember=uid=#{username},dc=example,dc=com))', {
+				this.add('LDAP_Teams_BaseDN', '', {
+					type: 'string',
+					enableQuery: enableQueryTeams,
+					invalidValue: '',
+				});
+				this.add('LDAP_Teams_Name_Field', 'ou,cn', {
+					type: 'string',
+					enableQuery: enableQueryTeams,
+					invalidValue: 'ou,cn',
```
sampaiodiego, line 224: the `invalidValue` could be an empty string right? that's the idea, so when there is no EE license, it will return the `invalidValue`, being a way to prevent an EE feature comes "pre-configured".
```suggestion
					invalidValue: '',
```

ee/server/settings/ldap.ts
```diff
@@ -200,20 +200,34 @@ export function addSettings(): void {
...
+				});
+
+				this.add('LDAP_Query_To_Get_User_Teams', '(&(ou=*)(uniqueMember=#{userdn}))', {
 					type: 'string',
-					enableQuery: { _id: 'LDAP_Enable_LDAP_Groups_To_RC_Teams', value: true },
-					invalidValue: '(&(ou=*)(uniqueMember=uid=#{username},dc=example,dc=com))',
+					enableQuery: enableQueryTeams,
+					invalidValue: '(&(ou=*)(uniqueMember=#{userdn}))',
```
sampaiodiego, line 230: same reason as above
```suggestion
					invalidValue: '',
```

sampaiodiego (changes requested):

app/importer/server/classes/ImportDataConverter.ts
pierre-lehnen-rc: No, `updateData` can have an `$addToSet` too.

sampaiodiego: ok.. I didn't realize the calls could mutate `updateData` 👍

sampaiodiego (approved):

PR #23628, opened 2021-11-04
Update ru.i18n.json

iBelow:
The period at the end of the password is misleading. Users copy the password along with the dot at the end of the sentence

image


ankar84:
I just corrected that problem at Lingohub.
Thank you for pointing in that issue!

dougfabris:
As @ankar84 mentioned it was already fixed on lingohub 😄 
@iBelow Thanks for your contribution!

PR #23650, opened 2021-11-06
[IMPROVE] Drag and Drop UI for files only

### LINKED ISSUE #16872: If You Drag Text While Editing It Client Gets In "Drop To Upload File" State
## My Setup

- Operating System: Windows 7
- App Version: 2.17.2
- Installation type: exe

- [x ] I have tested with the latest version
- [x ] I can simulate the issue easily

## Description
If text is highlighted and dragged, it goes into drag and drop mode which causes RocketChat to *think* you're trying to upload a file.

### Current Behavior
Type in some text in a chat or private message, highlight some of that text and drag it.  The window will go into "Drop to upload file mode".

### Expected Behavior
If possible it would be best if RocketChat disabled text dragging in the text edit window or at least somehow realize it's not a file being offered.      Plan B would be to have a cancel or "X" button in the 'drop file to upload' mode.

Escape usually closes upload mode *but* a small percentage of times, it doesn't and I have to actually drop a file and cancel that to make it work normally again which is annoying.

sidmohanty11:
fixes #16872

Please review this PR and suggest for required changes @gabriellsh 
Thank you

Now,

rocket-chat-files

Previously,

previous-rc-files

## Issue(s)

#16872

### DIFF
--- a/app/ui/client/views/app/room.js
+++ b/app/ui/client/views/app/room.js
@@ -424,8 +424,13 @@ export const dropzoneEvents = {
 	'dragenter .dropzone'(e) {
 		const types = e.originalEvent && e.originalEvent.dataTransfer && e.originalEvent.dataTransfer.types;
 
-		if (types != null && types.length > 0 && _.some(types, (type) => type.indexOf('text/') === -1 || type.indexOf('text/uri-list') !== -1 || type.indexOf('text/plain') !== -1) && userCanDrop(this._id)) {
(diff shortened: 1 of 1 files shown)


ankar84:
Nice fix for a long term issue!

sidmohanty11:
@ggazzo as mentioned in the issue, due to dragging on text too the UI shows "Drag files here". With the changes I made the UI shows "Drag files here" when any file is dragged upon.

PR #23661, opened 2021-11-08
[FIX] Performance issues when running Omnichannel job queue dispatcher

renatobecker:


ee/app/livechat-enterprise/server/lib/Helper.js
```diff
@@ -126,14 +139,11 @@ export const processWaitingQueue = async (department) => {
...
 		const { _id: rid, servedBy: { _id: agentId } } = room;
-		helperLogger.debug(`Inquiry ${ inquiry._id } taken succesfully by agent ${ agentId }. Notifying`);
+		helperLogger.debug(`Inquiry ${ inquiry._id } taken successfully by agent ${ agentId }. Notifying`);
 		return setTimeout(() => {
 			propagateAgentDelegated(rid, agentId);
 		}, 1000);
 	}
-
-	const { departmentId } = room || {};
-	await debouncedDispatchWaitingQueueStatus(departmentId);
```
renatobecker, line 136: FYI: This method was being called whenever the waiting queue was processed but there's no reason for that.

renatobecker:
> Let me see if i understood everything:
> 
> * This PR adds a new setting (so livechat does not gather statistics when its not being used), 
> * an index (such that the query for closed chats is faster),
> * a new property on the room object (queuedAt, but why?),
- Because the current information displayed is misleading. The **queue time** is the amount of time between the moment when the chat arrived in the queue and the moment the chat was taken by the agent. The same chat might be returned to the queue until it's solved, so `ts` field is not the correct field to calc the `queue time` information.
> * removed the dependency on the RoutingManager on the livechat room hook (so that the debouncedDispatchWaitingQueueStatus always happens on closing the room, but why?)
- Because `debouncedDispatchWaitingQueueStatus` must be called regardless of the routing system algorithm in place.

felipetomm:
@renatobecker, i see the big benefit to disabled it when most used channels is not widget, but just for information my self, when the new setting `Omnichannel_calculate_dispatch_service_queue_statistics` is disabled then does this cause an impact on use of livechat widget right? Maybe the queue statistics can be calculated only for livechat widget rooms because many client use the different channels by widget and Apps. Currently in `rocketchat_room` collection has the `source.type` property.

renatobecker:
> @renatobecker, i see the big benefit to disabled it when most used channels is not widget, but just for information my self, when the new setting `Omnichannel_calculate_dispatch_service_queue_statistics` is disabled then does this cause an impact on use of livechat widget right? Maybe the queue statistics can be calculated only for livechat widget rooms because many client use the different channels by widget and Apps. Currently in `rocketchat_room` collection has the `source.type` property.

Hey @felipetomm the ability to disable the queue statistics dispatcher doesn't impact on the widget, in terms of UI/UX the only difference is that that information will no longer show up on the top of the widget.
Also, we're planning to refactor the whole `Waiting Queue` feature so more improvements will come out very soon.

felipetomm:
> > @renatobecker, i see the big benefit to disabled it when most used channels is not widget, but just for information my self, when the new setting `Omnichannel_calculate_dispatch_service_queue_statistics` is disabled then does this cause an impact on use of livechat widget right? Maybe the queue statistics can be calculated only for livechat widget rooms because many client use the different channels by widget and Apps. Currently in `rocketchat_room` collection has the `source.type` property.
> 
> Hey @felipetomm the ability to disable the queue statistics dispatcher doesn't impact on the widget, in terms of UI/UX the only difference is that that information will no longer show up on the top of the widget. Also, we're planning to refactor the whole `Waiting Queue` feature so more improvements will come out very soon.

Nice @renatobecker , great work!

ee/app/livechat-enterprise/server/settings.ts
```diff
@@ -113,6 +113,37 @@ export const createSettings = (): void => {
...
 			});
 
+			this.add('Omnichannel_calculate_dispatch_service_queue_statistics', true, {
+				type: 'boolean',
+				group: 'Omnichannel',
+				section: 'Queue_management',
+				i18nLabel: 'Omnichannel_calculate_dispatch_service_queue_statistics',
+				enableQuery: [{ _id: 'Livechat_waiting_queue', value: true }, omnichannelEnabledQuery],
+				enterprise: true,
+				invalidValue: 0,
```
murtaza98, line 123:
```suggestion
				invalidValue: false,
```

KevLehman (approved):

PR #23694, opened 2021-11-11
[NEW] Allow Omnichannel statistics to be collected.

cauefcr:

This PR adds the possibility for business stakeholders to see what is actually being used of the Omnichannel integrations. 

## Issue(s)

https://app.clickup.com/t/1nfxcz6


cauefcr:
Sorry about forgetting to open this PR.

app/statistics/server/lib/statistics.js
```diff
@@ -114,7 +114,18 @@ export const statistics = {
 		statistics.totalLivechatAgents = Users.findAgents().count();
 
 		// livechat enabled
-		statistics.livechatEnabled = settings.get('Livechat_enabled');
+		statistics.livechatEnabled = settings.get('Livechat	_enabled');
```
murtaza98, line 117:
```suggestion
		statistics.livechatEnabled = settings.get('Livechat_enabled');
```

murtaza98:
Hi, @cauefcr I was wondering if there is a way to test the changes for this PR locally?

app/statistics/server/lib/statistics.js
```diff
@@ -114,7 +114,18 @@ export const statistics = {
...
+
+		// Count and types of omnichannel rooms
+		statistics.omnichannelSources = _.reduce(Promise.await(RoomsRaw.allRoomSourcesCount().toArray()), (acc, curr) => {
```
KevLehman, line 120: why using lodash reduce instead of a normal reduce?

KevLehman (changes requested):

app/statistics/server/lib/statistics.js
cauefcr: I was just following the usage on the file, i can change to a regular reduce too.

cauefcr:
> Hi, @cauefcr I was wondering if there is a way to test the changes for this PR locally?

Yes! We can change the endpoint that is being used to send the statistics to a local version of statscollector (on file https://github.com/RocketChat/Rocket.Chat/blob/develop/server/cron/statistics.js), which is able to run locally if we remove the fleetcommand middleware from the statsHandler endpoint (https://github.com/RocketChat/statscollector/blob/master/router/router.go), and change the config file to use your local mongo.

app/models/server/raw/Rooms.js
```diff
@@ -397,4 +397,22 @@ export class RoomsRaw extends BaseRaw {
...
 		return this.col.findOne({ $or: [{ name }, { fname: name }] }, options);
 	}
+
+	allRoomSourcesCount() {
+		return this.col.aggregate([
+			{
+				$match: {
+					source: {
+						$exists: true,
+					},
```
murtaza98, line 407:
```suggestion
					},
					t: 'l',
```

murtaza98: Although I do understand that source property will only exist on live-chat conversations, I'm still suggesting this to be extra cautious 😁

murtaza98 (dismissed):

KevLehman:
@cauefcr do you know how often this data will be pulled? If not, maybe we need an index on the `source` prop so it can be used on queries 👀

murtaza98:
> @cauefcr do you know how often this data will be pulled? If not, maybe we need an index on the `source` prop so it can be used on queries eyes

@KevLehman This happens every 12 minutes if I understand [this](https://github.com/RocketChat/Rocket.Chat/blob/a858dbc237fc9c628bed2444ce0d47a2cf6d7897/server/cron/statistics.js#L59) format correctly

KevLehman:
Nice, I'll recommend to add an index on "source" property then 👀 since there are customers with tons of rooms. 
(You can find examples by looking at the LivechatRooms.js model under app/models/)

app/statistics/server/lib/statistics.js
murtaza98:
```suggestion
		statistics.omnichannelSources = Promise.await(RoomsRaw.allRoomSourcesCount().toArray()).map((source) => ({
			id: source._id.id,
			alias: source._id.alias,
			type: source._id.type,
			count: source.count,
		}));
```

Looking at the data, I found that the **map** operation fits better here instead of **reduce** operation since our original and final array always has a 1:1 mapping

app/models/server/raw/Rooms.js
```diff
@@ -3,6 +3,15 @@ import { escapeRegExp } from '@rocket.chat/string-helpers';
 import { BaseRaw } from './BaseRaw';
 
 export class RoomsRaw extends BaseRaw {
+	indexes=[
```
murtaza98, line 6: @cauefcr This index was not applied when I restarted my local server. I feel this has something to do with this file still being in .js instead of .ts :thinking: 
Anyways, when I applied the same index in [this](https://github.com/RocketChat/Rocket.Chat/blob/7660464d086ecdaef18cf3bd9eae7684ff317f18/app/models/server/models/LivechatRooms.js#L23) file, I could see that it was properly applied. So maybe we need to move this over there. @KevLehman WDYT? Can you check if the indexes are getting applied on your local?

cauefcr: weird, it got applied on mine, i'll move it over.

murtaza98 (approved):
LGTM

KevLehman (approved):

PR #23698, opened 2021-11-12
[NEW] Add API to Expose Unread Message count

nmagedman:
## Proposed changes

Add API for getting a user’s count of unread messages.
Useful for websites that want to integrate with RocketChat,
showing the unread count _within_ the website UI and linking
to the RC web app.

Screen Shot - Unread Messaage count in site navbar

## Further comments
In this implementation, we expose this data via new API endpoint.  An alternative approach would be to add the unread count to `users.info`.  That latter endpoint already includes unread counts on a per-room basis, which could then be summed together.

Cc: @milton-rucks


milton-rucks:
@nmagedman we are still analyzing the solution for this one

Issue #15369, opened 2019-09-13
Cannot reference discussion by its name


tyteen4a03:
### Description:

It is not possible to reference a discussion by its name in channels.

### Steps to reproduce:

1. Create a new discussion with any name.
2. Go to the channel and type in # which should bring up the autocomplete menu.

### Expected behavior:

It should show the name of the discussion.

### Actual behavior:

It shows the ID of the discussion.

<img width="205" alt="Screenshot 2019-09-13 at 23 00 16" src="https://user-images.githubusercontent.com/440015/64897150-45fbe100-d67a-11e9-86de-4bd29b1fe09f.png">

### Server Setup Information:

- Version of Rocket.Chat Server: Latest that is running on RC cloud
- Operating System: 
- Deployment Method: 
- Number of Running Instances: 
- DB Replicaset Oplog: 
- NodeJS Version: 
- MongoDB Version:

eengstrom:
It's not so much that you can't reference the discussion, but that the discussion is named for some random UID generated (e.g. `#x5bX9...` in your example.  Would be nice to be able to construct a human readable version or alias for that, minimally.

dasBuero:
This is really annoying as we often use discussions to structure channels and urgently need a working referencing. Hope to see a solution soon.

xmedeko:
It's OK for most discussion to have random name, but I would like to be able to rename a few of them, turn them public and be able to reference them, too.

Kayoku:
Hi,
I'm Interested too for my community to had the possibility to name our discussion with a human readable name, is that possible ? :-)
Thanks!

tyteen4a03:
Still very much an issue.

adam-bapm:
I just stumbled across this as well and would appreciate this feature.

vladborovtsov:
Just another +1 :)

deepesharora:
I'd like to add my +1 here as well. 
Discussions are awesome the way they are setup....they are really subtopics of channels. So no point if you tell me to convert a discussion to a channel..its not quite the right answer. 
What will be awesome is instead of referencing the discussion by its random id, you allow us to reference it by its 'human-friendly' name.  Internally its ok to continue to use the id...just the mobile and desktop apps should go 
#<random-id>-"human-friendly-name"

This avoids the risk of ending up with 2 discussions that have the same human-friendly name. 
my 2c 👍

jwhall:
[...]

PR #23710, opened 2021-11-13
[FIX] reference discussion by its name in the popup

Aman-Maheshwari:
- [x]  I have read the Contributing Guide 
  - [x] I have signed the CLA - https://cla-assistant.io/RocketChat/Rocket.Chat
  - [ ] I have added necessary documentation (if applicable)
  - [ ] Any dependent changes have been merged and published in downstream modules

In the popup, for discussion only the random discussion ID was visible,  but now user can see the name of the discussion and also the name of the channel to which the discussion belongs to (in case of the same discussion name but different channel). Also using ID avoid the problem that could arise with the discussion having same name.

once send in the channel, corresponding discussion ID is converted to the link(as was before), but in the side the human readable name for the discussion is shown.

## Before

https://user-images.githubusercontent.com/50165440/141660219-6288f7ea-b719-4b63-819d-63eb2ee61fe3.mp4

## After

https://user-images.githubusercontent.com/50165440/141660224-33630090-4298-4fec-b5d7-f2e6bbd22f37.mp4

## Issue(s)

#15369 
## Steps to test or reproduce

Mentioned in the issue

### DIFF
--- a/app/ui-message/client/popup/messagePopupChannel.html
+++ b/app/ui-message/client/popup/messagePopupChannel.html
@@ -1,4 +1,4 @@
 <template name="messagePopupChannel">
 	{{> icon icon=channelIcon }}
-	{{name}}
+	{{fname}}{{#if topic}} - [{{> icon icon=roomIcon }}{{topic}}]{{/if}}
 </template>
--- a/app/ui-message/client/popup/messagePopupChannel.js
+++ b/app/ui-message/client/popup/messagePopupChannel.js
@@ -1,3 +1,4 @@
+import { Meteor } from 'meteor/meteor';
 import { Template } from 'meteor/templating';
 
 import { roomTypes } from '../../../utils';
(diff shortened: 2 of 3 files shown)


geekgonecrazy:
IMO we have to figure out a better way to show the discussion in the message list too.  Cause that is still super confusing if not

paulopera:
It would be really nice if RoCket.Chat adopt this kind of approach to reference a discussion in a DM or Channel. My only concern is that **IF** I reference a discussion with a specific **channel/person** and that **channel/person** doesn't belong to the **Parent channel or group** could possibly create some **"awkward situation"** with all the people involved. Such as a new employe that doesn't have the clearance to access the discussion or someone that doesn't need to have access.
Some kind of check to verify if everyone in that particular **Channel/DM** has the rights to read the "to be referenced discussion" and then enable sending it.
---
Just after writing the text above I tested sending a url from the discussion to a person and if that person doesn't have the rights to access the discussion it shows a **red alert** on the screen.
So the App indeed make a check but I don't know how much it would impact in **performance of the application** validating it when trying to reference the discussion like described before.

PR #23726, opened 2021-11-16
Chore: index cleansing omnichannel

KevLehman:

Reasoning behind index removals:

- AgentActivity { date } => replaced by compound index
- AgentActivity { agentId } => replaced by compound index
- DepartmentAgents { agentId } => replaced by compound index
- DepartmentAgents { departmentEnabled } => index on a property with just 2 possible values, not adding much value
- Inquiry { message } => not being used
- Inquiry { name } => not being used
- Inquiry { status } => replaced by compound index
- Room { open } => replaced by compound
- 'v.token' => replaced by compound index

Why?
- The unused indexes were measured on conditions on 100k rooms + 20k inquiries.
- The preffered usage of compound indexes over normal indexes is based on https://docs.mongodb.com/manual/core/index-compound/#prefixes (I'm adjusting to validate all removed indexes are covered with a compound prefix 👀 )

## Further comments

Pls let me know any suggestions 👀 or another things that can be added/removed

### DIFF
--- a/app/models/server/models/LivechatRooms.js
+++ b/app/models/server/models/LivechatRooms.js
@@ -9,7 +9,6 @@ export class LivechatRooms extends Base {
 	constructor(...args) {
 		super(...args);
 
-		// this.tryEnsureIndex({ open: 1 }, { sparse: true }); covered by compound index
 		this.tryEnsureIndex({ departmentId: 1 }, { sparse: true });
 		this.tryEnsureIndex({ 'metrics.chatDuration': 1 }, { sparse: true });
 		this.tryEnsureIndex({ 'metrics.serviceTimeDuration': 1 }, { sparse: true });
@@ -17,7 +16,6 @@ export class LivechatRooms extends Base {
 		this.tryEnsureIndex({ 'omnichannel.predictedVisitorAbandonmentAt': 1 }, { sparse: true });
 		this.tryEnsureIndex({ closedAt: 1 }, { sparse: true });
 		this.tryEnsureIndex({ servedBy: 1 }, { sparse: true });
-		// this.tryEnsureIndex({ 'v.token': 1 }, { sparse: true }); covered by following index
 		this.tryEnsureIndex({ 'v.token': 1, 'email.thread': 1 }, { sparse: true });
 		this.tryEnsureIndex({ 'v._id': 1 }, { sparse: true });
 		this.tryEnsureIndex({ t: 1, departmentId: 1, closedAt: 1 }, { partialFilterExpression: { closedAt: { $exists: true } } });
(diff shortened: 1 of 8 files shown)


app/models/server/models/LivechatRooms.js
```diff
@@ -9,18 +9,17 @@ export class LivechatRooms extends Base {
...
 		this.tryEnsureIndex({ 'metrics.serviceTimeDuration': 1 }, { sparse: true });
 		this.tryEnsureIndex({ 'metrics.visitorInactivity': 1 }, { sparse: true });
 		this.tryEnsureIndex({ 'omnichannel.predictedVisitorAbandonmentAt': 1 }, { sparse: true });
 		this.tryEnsureIndex({ closedAt: 1 }, { sparse: true });
 		this.tryEnsureIndex({ servedBy: 1 }, { sparse: true });
-		this.tryEnsureIndex({ 'v.token': 1 }, { sparse: true });
 		this.tryEnsureIndex({ 'v.token': 1, 'email.thread': 1 }, { sparse: true });
 		this.tryEnsureIndex({ 'v._id': 1 }, { sparse: true });
 		this.tryEnsureIndex({ t: 1, departmentId: 1, closedAt: 1 }, { partialFilterExpression: { closedAt: { $exists: true } } });
+		this.tryEnsureIndex({ open: 1, 'v.token': 1 }, { sparse: true });
```
sampaiodiego, line 22: for compound sparse indexes, even if any of the fields are empty the index will be stored anyways.. so it compound indexes it makes sense to be sparse only if **both** fields can be empty, is that the case? otherwise I'd say this to not be sparse.

sampaiodiego: btw, I've seen a lot of queries using `t` and `v.token` together, but no index with both together, maybe there is an opportunity for better index?

KevLehman: > btw, I've seen a lot of queries using t and v.token together, but no index with both together, maybe there is an opportunity for better index?

For that case, I'd say it's better to keep the v.token index alone 🤔 . I mean, even if we query for t & v.token, the `t` filter only has around 4 possible values, so the index would have a lot of repeated data. While v.token should limit a lot the results 👀 wdyt?

> for compound sparse indexes, even if any of the fields are empty the index will be stored anyways.. so it compound indexes it makes sense to be sparse only if both fields can be empty, is that the case? otherwise I'd say this to not be sparse.

I didn't know this 👀 for this specific index, i'd say it's not needed for it to be sparse. I'll revisit others to check if they should/n't be sparse

sampaiodiego:
bump 😬

KevLehman:
Closing for now

PR #23732, opened 2021-11-17
[NEW] Rate limiting for user registering

ostjen:
Introducing a new setting that controls the number of calls to the rate limiter specifically for registering users


app/lib/server/startup/settings.ts
```diff
@@ -3003,6 +3003,7 @@ settingsRegistry.addGroup('Rate Limiter', function() {
 		this.add('API_Enable_Rate_Limiter', true, { type: 'boolean' });
 		this.add('API_Enable_Rate_Limiter_Dev', true, { type: 'boolean', enableQuery: { _id: 'API_Enable_Rate_Limiter', value: true } });
 		this.add('API_Enable_Rate_Limiter_Limit_Calls_Default', 10, { type: 'int', enableQuery: { _id: 'API_Enable_Rate_Limiter', value: true } });
+		this.add('API_Enable_Rate_Limiter_Limit_RegisterUser', 1, { type: 'int', enableQuery: { _id: 'API_Enable_Rate_Limiter', value: true } });
```
rodrigok, line 3006: This should be in a different section, it's not API or DDP only setting, it's a feature limit setting.

server/methods/registerUser.js
```diff
@@ -97,3 +97,9 @@ Meteor.methods({
 		return userId;
 	},
 });
+
+RateLimiter.limitMethod('registerUser',
```
rodrigok, line 101: Since the value came from a setting you will need to watch the setting and on change remove the limiter and recreate it.

app/api/server/v1/users.js
```diff
@@ -283,7 +283,9 @@ API.v1.addRoute('users.list', { authRequired: true }, {
 	},
 });
 
-API.v1.addRoute('users.register', { authRequired: false }, {
+API.v1.addRoute('users.register', { authRequired: false,
```
rodrigok, line 286: Here you will need to keep the object reference and update the values on setting changes, after each change you need to call the API method `reloadRoutesToRefreshRateLimiter` to refresh the limiters

rodrigok (approved):

rodrigok (approved):

PR #23736, opened 2021-11-17
[NEW] : API endpoints for Agent-Extension Association and Database changes

amolghode1981:
Description : This commit adds the required APIs and permissions to run these APIs for following :
1. Create agent-extension association. (Access to admin)
2. Get extension associated with given agent name. (Access to agent, admin and manager)
3. Delete extension of a given agent. (Access to admin)
4. List all free extensions. (Access to admin)
5. Get the list of agent-extension association. (Access to admin)

It adds necessary functions in the omnichannel-voip service and adds corresponding types.

In the database, it adds a new field |extension| to existing meteor |users| document.

This PR adds following APIs and necessary permissions
1. Create agent-extension association. (Access to admin)
2. Get extension associated with given agent name. (Access to agent, admin and manager)
3. Delete extension of a given agent. (Access to admin)
4. List all free extensions. (Access to admin)
5. Get the list of agent-extension association. (Access to admin)

Along with these APIs, it also adds the code in omnichannel-voip service to provide necessary data.

To search all free extensions, it uses Voip service (Which internally uses asterisk connector) to pull the information from Asterisk. It then looks up in to the database for all those agents which have extensions associated with it.

Then it finds the difference between allextensions and associatedextensions.

To get the allocation details, it gets all those agents from the data base which have extension associated with it.

Testing Notes :
1. Check that the API permission is correctly coded. 
2. check all the APIs added work properly.

## Further comments
We are using one service from another here. omnichannel-voip uses voip service.


server/services/omnichannel-voip/service.ts
```diff
@@ -1,10 +1,93 @@
...
+		super();
+		this.users = new UsersRaw(db.collection('users'));
+		this.logger = new Logger('OmnichannelVoipService');
+	}
+
 	getConfiguration(): any {
 		return {};
 	}
+
+	private async getAllocatedExtesionAllocationData(projections: any): Promise<any> {
```
murtaza98, line 38:
```suggestion
	private async getAllocatedExtensionAllocationData(projections: any): Promise<any> {
```
Minor typo fix

server/services/omnichannel-voip/service.ts
```diff
@@ -1,10 +1,93 @@
...
+		const roles: string[] = ['livechat-agent', 'livechat-manager', 'admin'];
```
murtaza98, line 39: I'd suggest to put declare private methods after public methods, since it makes the code easier to follow. FYI: we have [this](https://palantir.github.io/tslint/rules/member-ordering/) tslint rule for this as well

server/services/omnichannel-voip/service.ts
```diff
@@ -1,10 +1,93 @@
...
+		const options = {
+			sort: {
+				username: 1,
+			},
+			projection: {
```
murtaza98, line 44:
```suggestion
			projection,
```

murtaza98: we can simply keep this like this and avoid the following reassinging later
`options.projection = projections;`

server/services/omnichannel-voip/service.ts
```diff
@@ -1,10 +1,93 @@
...
+		options.projection = projections;
+		const query = {
+			extension: { $exists: true },
+		};
+		const data = await this.users.findUsersInRolesWithQuery(roles, query, options).toArray();
+		return data;
+	}
+
+	async getAvailableExtensions(): Promise<IOmnichannelVoipServiceResult> {
+		const allExtensions = Promise.await(Voip.getExtensionList()) as IVoipConnectorResult;
```
murtaza98, line 56:
```suggestion
		const allExtensions = await Voip.getExtensionList();
```

murtaza98: Since its already a typescript file, we can use this shortcut instead of calling Promise.await.

Also the return type of this method is already sepcified, so we don't have to override the type

app/api/server/v1/voip/omnichannel.ts
```diff
@@ -0,0 +1,94 @@
...
+		const user = Users.findOneByAgentUsername(this.bodyParams.username, {
+			fields: { _id: 1, username: 1 },
+		});
+		if (!user) {
+			return API.v1.notFound();
+		}
+		Users.setExtension(user._id, this.bodyParams.extension);
+		return API.v1.success();
+	},
+	delete() {
```
amolghode1981, line 50: @renatobecker @KevLehman and @murtaza98 I have added delete API. This delete API will simply delete the extension for the user. So we do not have a trail of the past extensions atleast in the user document.

Nevertheless, we can either keep this trail in the server side or client side logging. We need to design the fields which needs to be logged for auditing purpose. 

Alternatively we can maintain the extension array and it will contain 
extension: [
    {
     _id: <Extension number>
    isActive: <boolean>
    },
...
]

We will not delete the extension. on Deletion we will just make isActive false.
While creating, we shall just push a new element in the array and set isActive = true.

Reasoning about this : On a sunny day, a delete should just delete the element but when it is with any telephony systems certain records are necessary to be maintained. e.g Call detail records **(CDR)** are necessary for GDPR. Some countries like India has strict toll compliance guidelines (And hence we do not yet have publicly sold VoIP trunks). They too need CDR. We can argue that we are not a pbx because maintaining the CDR is the responsibility of the pbx (SIP server). Nevertheless, we need some information that can trace an individual agent if something goes wrong. So if we simply delete the extension, there is no tracing of an individual thru the database. We can definitely trace back thru logs and login/logout time of an agent but that would need a multilevel redirection of data.

The long story short,  we need to decide on what we need to maintain for the audit trail and I would like to hear your thoughts.

app/models/server/models/Users.js
```diff
@@ -53,6 +53,7 @@ export class Users extends Base {
 		this.tryEnsureIndex({ 'services.saml.inResponseTo': 1 });
 		this.tryEnsureIndex({ openBusinessHours: 1 }, { sparse: true });
 		this.tryEnsureIndex({ statusLivechat: 1 }, { sparse: true });
+		this.tryEnsureIndex({ extension: 1 }, { sparse: true });
```
murtaza98, line 56: Can we also add definitions for this new property in the user model i.e. [here](https://github.com/RocketChat/Rocket.Chat/blob/c54ef380d29f99bfde38c31f8dccbf841ffef455/definition/IUser.ts#L110)

server/services/omnichannel-voip/service.ts
```diff
@@ -1,10 +1,93 @@
...
+		const filtered = _.difference(_.pluck(allExtensions.result as IVoipExtensionBase [], 'extension'),
+			_.pluck(allocatedExtensions, 'extension'));
+		this.logger.debug({ msg: 'getAvailableExtensions()', found: filtered.length });
+		const result: IOmnichannelVoipServiceResult = {
+			result: filtered,
+		};
+		this.logger.debug({ msg: 'getAvailableExtensions() result length ',
+			length: result.result.length,
+		});
+		return Promise.resolve(result);
```
murtaza98, line 69: Since this method is already async, we don't need to explicity resolve promise

murtaza98:
```suggestion
		return result;
```

server/services/omnichannel-voip/service.ts
```diff
@@ -1,10 +1,93 @@
...
+			length: allocatedExtensions.length,
+		});
+		const result: IOmnichannelVoipServiceResult = {
+			result: _.map(allocatedExtensions, (user: any) => {
+				const agent: IAgentExtensionMap = {
+					_id: user._id,
+					agentName: user.username,
+					extension: user.extension,
+				};
+				return agent;
```
murtaza98, line 88: Since is object isn't being used anywhere else, then Y not simply return the object directly instead of storing it in another variable

murtaza98: Also FYI: all array already have inbuilt operations like map, reduce, pluck, etc, so you can reuse that directly instead of importing it from underscore package.

Eg: this whole code can be simplified like this
```
			result: allocatedExtensions.map((user: any) => ({
				_id: user._id,
				agentName: user.username,
				extension: user.extension,
			} as IAgentExtensionMap)),
```

server/services/omnichannel-voip/service.ts
```diff
@@ -1,10 +1,93 @@
...
+		const allocatedExtensions = await this.getAllocatedExtesionAllocationData({
+			username: 1,
+			roles: 1,
+			extension: 1,
+		});
+		this.logger.debug({ msg: 'getExtensionAllocationDetails() all extension length ',
+			length: allocatedExtensions.length,
+		});
+		const result: IOmnichannelVoipServiceResult = {
+			result: _.map(allocatedExtensions, (user: any) => {
```
murtaza98, line 82: Since is object isn't being used anywhere else, then Y not simply return the object directly instead of storing it in another variable

murtaza98: Also FYI: all array already have inbuilt operations like map, reduce, pluck, etc, so you can reuse that directly instead of importing it from underscore package.

Eg: this whole code can be simplified like this
```
			result: allocatedExtensions.map((user: any) => ({
				_id: user._id,
				agentName: user.username,
				extension: user.extension,
			} as IAgentExtensionMap)),
```

server/services/omnichannel-voip/service.ts
```diff
@@ -1,10 +1,93 @@
...
+				const agent: IAgentExtensionMap = {
+					_id: user._id,
+					agentName: user.username,
+					extension: user.extension,
+				};
+				return agent;
+			}),
+		};
+		return Promise.resolve(result);
```
murtaza98, line 91: Since this method is already async, we don't need to explicity resolve promise

app/authorization/server/functions/upsertPermissions.js
```diff
@@ -145,6 +145,9 @@ export const upsertPermissions = () => {
 		// allows to manage voip calls configuration
 		{ _id: 'manage-voip-call-settings', roles: ['livechat-manager', 'admin'] },
 		{ _id: 'manage-voip-contact-center-settings', roles: ['livechat-manager', 'admin'] },
+		// allows agent-extension association.
+		{ _id: 'managage-agent-extension-association', roles: ['admin'] },
```
murtaza98, line 149:
```suggestion
		{ _id: 'manage-agent-extension-association', roles: ['admin'] },
```
Minor typo

app/authorization/server/functions/upsertPermissions.js
```diff
@@ -145,6 +145,9 @@ export const upsertPermissions = () => {
...
+		{ _id: 'query-agent-extension-association', roles: ['livechat-manager', 'admin', 'livechat-agent'] },
```
murtaza98, line 150:
```suggestion
		{ _id: 'view-agent-extension-association', roles: ['livechat-manager', 'admin', 'livechat-agent'] },
```
Suggesting this since we usually use "view" verb as a prefix for settings with only read access

app/api/server/v1/voip/omnichannel.ts
```diff
@@ -0,0 +1,94 @@
+import { Match, check } from 'meteor/check';
+
+import { API } from '../../api';
+import { Users } from '../../../../models/server/index';
+import { hasPermission } from '../../../../authorization/server/index';
+import { LivechatVoip } from '../../../../../server/sdk';
+
+API.v1.addRoute('omnichannel.agent.extension', { authRequired: true }, {
```
murtaza98, line 8:
```suggestion
API.v1.addRoute('omnichannel.agent.extension', { authRequired: true, permissionsRequired: ['query-agent-extension-association'] }, {
```

You don't have to explicity check for permission in your code. you can simply pass the permission you want on this route and then addRoute will take care of the rest

app/models/server/models/Users.js
murtaza98: @Amol, will these extensions always be unique for each user doc? If yes, then I'd recommended adding a `unique` index here as well for further optimizations during search

app/models/server/models/Users.js
```diff
@@ -1622,6 +1623,43 @@ Find users to send a message by email if:
 			},
 		});
 	}
+
+	// Voip functions
```
murtaza98, line 1627: Since our end goal on backend is to move all query method to use the raw model instead of this traditional model, I'd recommened moving all these methods to the respective raw User model i.e. [this one](https://github.com/RocketChat/Rocket.Chat/blob/develop/app/models/server/raw/Users.js)

app/api/server/v1/voip/omnichannel.ts
murtaza98: Just realised that this comment won't be applicable to this route since each method required a different permission. So you can ingore this over here,

app/api/server/v1/voip/omnichannel.ts
```diff
@@ -0,0 +1,94 @@
...
+		if (!user) {
+			return API.v1.notFound();
+		}
+		Users.unsetExtension(user._id);
+		return API.v1.success();
+	},
+});
+
+// Get free extensions
+API.v1.addRoute('omnichannel.extension.free', { authRequired: true }, {
```
murtaza98, line 69:
```suggestion
API.v1.addRoute('omnichannel.extension.free', { authRequired: true, permissionsRequired: ['manage-agent-extension-association'] }, {
```

You don't have to explicity check for permission in your code. you can simply pass the permission you want on this route and then addRoute will take care of the rest

app/api/server/v1/voip/omnichannel.ts
```diff
@@ -0,0 +1,94 @@
...
+		const extension = Promise.await(LivechatVoip.getAvailableExtensions());
+		if (!extension) {
+			return API.v1.notFound();
+		}
+		return API.v1.success(extension);
+	},
+});
+
+// Get free extensions
+API.v1.addRoute('omnichannel.extension.allocated', { authRequired: true }, {
```
murtaza98, line 83: Same goes here

app/api/server/v1/voip/omnichannel.ts
```diff
@@ -0,0 +1,94 @@
...
+			return API.v1.unauthorized();
+		}
+		check(this.requestParams(), Match.ObjectIncluding({
+			username: String,
+		}));
+		const user = Users.findOneByAgentUsername(this.requestParams().username, {
+			fields: { _id: 1 },
+		});
+		if (!user) {
+			return API.v1.notFound();
```
murtaza98, line 21:
```suggestion
			return API.v1.notFound('User not found');
```
Trying to be explicit here since this method returns 2 no found errors

app/api/server/v1/voip/omnichannel.ts
```diff
@@ -0,0 +1,94 @@
...
+		}
+		const extension = Users.getExtension(user._id, {
+			fields: { _id: 1, username: 1, extension: 1 },
+		});
+		if (!extension) {
+			return API.v1.notFound();
```
murtaza98, line 27:
```suggestion
			return API.v1.notFound('Extension not found');
```

app/api/server/v1/voip/omnichannel.ts
```diff
@@ -0,0 +1,94 @@
...
+	},
+});
+
+// Get free extensions
+API.v1.addRoute('omnichannel.extension.free', { authRequired: true }, {
+	get() {
+		if (!hasPermission(this.userId, 'managage-agent-extension-association')) {
+			return API.v1.unauthorized();
+		}
+		const extension = Promise.await(LivechatVoip.getAvailableExtensions());
```
murtaza98, line 74:
```suggestion
		const extension = await LivechatVoip.getAvailableExtensions();
```

app/api/server/v1/voip/omnichannel.ts
```diff
@@ -0,0 +1,94 @@
...
+	},
+});
+
+// Get free extensions
+API.v1.addRoute('omnichannel.extension.allocated', { authRequired: true }, {
+	get() {
+		if (!hasPermission(this.userId, 'managage-agent-extension-association')) {
+			return API.v1.unauthorized();
+		}
+		const association = Promise.await(LivechatVoip.getExtensionAllocationDetails());
```
murtaza98, line 88:
```suggestion
		const association = LivechatVoip.getExtensionAllocationDetails();
```

app/api/server/v1/voip/omnichannel.ts
murtaza98: Since its already a typescript file, we can use this shortcut instead of calling Promise.await.

app/api/server/v1/voip/omnichannel.ts
murtaza98: If such a level of traceability is something that's required, then we certainly will have to create a new collection to store all these extensions. 
Amol, I'm a little concerned about storing the trace info on the user's document because I'm worried about increasing the size of the user's document in case an agent's extension changes many times. Hence, I'm suggesting moving this trace data into another collection.

Following is my proposal for the new collection
```
{
	_id: '123', 
	extension: '123123' // note, this wouldn't be required incase we're sure that all extension will be unique. hen we can maintain this data directly within _id
	history: [
		{
			userId: 'current-active-user for this extension',
			validFrom: '21st Oct'
		},
		{
			userId: 'xyz',
			validFromTs: '17th Oct',
			validToTs: '20th Oct'
		},
		{
			userId: 'abc',
			validFromTs: '1th Sept',
			validToTs: '10th Oct'
		},
	]
}
```

Then in the user's collection, we can just store the extension ID of the corresponding extension doc

server/services/omnichannel-voip/service.ts
amolghode1981: Fixed

app/api/server/v1/voip/omnichannel.ts
amolghode1981: Hello @murtaza98 while its a good design, I feel that it would be an overkill to utilise this for extension. The reason is as follows :
1. Extensions are going to remain fairly constant. The chances that the user extension may change for are
   a. PBX changes the extension partitions.
   b. Agent becomes non-voip agent.   
2. validFromTs-validToTs may not be needed as this can be referred from logs.

Furthermore the problem of traceability does not end here. What happens if the agent gets deleted? (Do we hard delete the agent today?).

More and more we discuss about this, I think we need to device an external mechanism for traceability.

But I am open to implement the design you suggested if that is the optimal way. @renatobecker what do you think about the modification / deletion frequency for the agent's extension?

app/models/server/models/Users.js
amolghode1981: Yes. These extensions will be unique for each agent/user. i.e 2 user's do not share the same extension. So I shall add unique index. Thanks for that suggestion.

Added the property in new user model IUser.

server/services/omnichannel-voip/service.ts
amolghode1981: Fixed.

app/authorization/server/functions/upsertPermissions.js
amolghode1981: Fixed.

app/authorization/server/functions/upsertPermissions.js
amolghode1981: Fixed.

app/api/server/v1/voip/omnichannel.ts
amolghode1981: Makes sense. No change as such on this. But will keep this in mind for other REST

app/api/server/v1/voip/omnichannel.ts
amolghode1981: Fixed.

app/api/server/v1/voip/omnichannel.ts
amolghode1981: Fixed.

app/api/server/v1/voip/omnichannel.ts
amolghode1981: Fixed.

app/api/server/v1/voip/omnichannel.ts
amolghode1981: Fixed.

server/services/omnichannel-voip/service.ts
amolghode1981: Fixed.

server/services/omnichannel-voip/service.ts
amolghode1981: Fixed.

server/services/omnichannel-voip/service.ts
amolghode1981: Fixed.

server/services/omnichannel-voip/service.ts
amolghode1981: This does not seem to work as expected. If we simply want to use async, we need to make the get function async.
But framework doesnt seem to call get in await mode. So whatever is returned is some object which does not have the desired result. I tried to see if I can find the type. But that too is not present. So we have to use Promise.await, which will not need the get method to be async.

app/api/server/v1/voip/omnichannel.ts
amolghode1981: This does not seem to work as expected. If we simply want to use async, we need to make the get function async.
But framework doesnt seem to call get in await mode. So whatever is returned is some object which does not have the desired result. I tried to see if I can find the type. But that too is not present. So we have to use Promise.await, which will not need the get method to be async.

app/api/server/v1/voip/omnichannel.ts
amolghode1981: This does not seem to work as expected. If we simply want to use async, we need to make the get function async.
But framework doesnt seem to call get in await mode. So whatever is returned is some object which does not have the desired result. I tried to see if I can find the type. But that too is not present. So we have to use Promise.await, which will not need the get method to be async.

app/models/server/models/Users.js
amolghode1981: @murtaza98  How do we do it? REST API layer seem to use the traditional model. Can you please point me to how to use this?

amolghode1981: I tried that but didnt work just by moving the APIs there.

app/api/server/v1/voip/omnichannel.ts
murtaza98: Hi @amolghode1981 

If the extension data is always short, then I agree with storing it directly within users collection. 

Regarding your second point of **validFromTs-validToTs**, here are you referring to Rocket.chat server logs or some other server logs. If it's Rocket.Chat logs, then  I wouldn't recommend relying on it for this kind of info, if it's that valuable, because logs are easily purged and so this info can easily get lost. Hence I'd recommend storing it in DB instead

Next this question
> What happens if the agent gets deleted? (Do we hard delete the agent today?).

Yes, we indeed hard-delete a user right now. So if this kind of traceability is required, then we'd indeed have to override that logic by somehow preventing a user to be deleted.

app/api/server/v1/voip/omnichannel.ts
murtaza98: Okay, that makes sense. Please ignore this suggestion

app/api/server/v1/voip/omnichannel.ts
murtaza98: Okay, that makes sense. Please ignore this suggestion

app/models/server/models/Users.js
murtaza98: @amolghode1981 you'll have to modify the rest layer API's to use the new raw models instead of the traditional collection. You can import the raw collection there using 
`import { Users } from '../../../../models/server/raw';` instead of `import { Users } from '../../../../models/server/index';`

Sidenote: Notice in the new file which you're importing above, [here](https://github.com/RocketChat/Rocket.Chat/blob/f1ce17dd71060a6faecde14e73f047aaad72afd0/app/models/server/raw/index.ts#L76) we're combining the Raw and traditional models, so this way using the Raw model, you'll be able to use all the traditional model method, but you won't get typescript definitions for them. If you want typescript definitions, then you'll have to declare them inside Raw models

app/api/server/v1/voip/omnichannel.ts
amolghode1981: @murtaza98 Thanks for the explanations. Few more points from my side 
1. Extension is going remain fairly simple. At the most it will be an object with 2 fields, 
{
  extension: <allocated extension>,
 server: <server-id from the voip server table>
},
2. But above said scheme too is needed if we ever support multiple VoIP servers on our platform. So theoretically it will be possible, but we may not ever take that way. 
3. I am referring rocket.chat logs here, we also should make sure that the client side logs too are received on the server and stored. I think we shall address the traceability issue at higher level because we need to address more things around it. 
@renatobecker what do you think?

app/models/server/models/Users.js
amolghode1981: Fixed.

app/api/server/v1/voip/omnichannel.ts
renatobecker: I didn't get what's exactly the meaning of traceability here: if the idea is to design a way to know who was the agent/user who was using an extension, I assume that the extension number will be stored in the room document, right? We need to store VoIP data/context inside the room, I mean, relevant information and then we'll have traceability because we'll know what was the extension number of each call and who was the agent associated. If I'm missing anything, please lemme know.
Also, I think we should have an extension history collection to store associations (extension x user) and, also, make it available on the UI, it's definitely something useful. 
BTW: We should bring this discussion to open server and discuss it internally as we're talking about business rules.

amolghode1981: @renatobecker @KevLehman @murtaza98 Resolving this comment as we have agreed upon creating a separate collection for maintaining Voip activity **Omnichannel call data record OCDR**. With this approach, user document will contain a simple extension string. The traceability will be possible via looking at the OCDR collection.

server/services/omnichannel-voip/service.ts
```diff
@@ -1,10 +1,84 @@
...
+import { Logger } from '../../lib/logger/Logger';
+import { Voip } from '../../sdk';
+import { IAgentExtensionMap, IOmnichannelVoipServiceResult } from '../../../definition/ILivechatVoipServiceResult';
+/**
+ * Note (Amol) :
+ * Why cant we get the direct access to the model here and have to work with
+ * raw model? This is still a server code. So ideally it should have access to the model
+ * like the REST apis have.
+ *
+ * Check this during the code review and after code review is done, delete this comment.
```
murtaza98, line 17: Is it okay to remove this comment now? Let me know if you have any more doubts regarding models

server/services/omnichannel-voip/service.ts
```diff
@@ -1,10 +1,84 @@
...
+		};
+		return this.users.findUsersInRolesWithQuery(roles, query, options).toArray();
+	}
+
 	getConfiguration(): any {
 		return {};
 	}
+
+	async getAvailableExtensions(): Promise<IOmnichannelVoipServiceResult> {
+		const allExtensions = await Voip.getExtensionList() as IVoipConnectorResult;
```
KevLehman, line 55: Shouldn't this call be already typed without the coercion? 👀

server/services/omnichannel-voip/service.ts
```diff
@@ -1,10 +1,84 @@
...
+			extension: { $exists: true },
+		};
+		return this.users.findUsersInRolesWithQuery(roles, query, options).toArray();
+	}
+
 	getConfiguration(): any {
 		return {};
 	}
+
+	async getAvailableExtensions(): Promise<IOmnichannelVoipServiceResult> {
```
KevLehman, line 54: How often will this be called/modified? Do you think it's worth to add a small cache to the extension associations to avoid having to go to the db each time? We do have some npm packages to manage timed cache

KevLehman (changes requested):

app/api/server/v1/voip/omnichannel.ts
```diff
@@ -0,0 +1,102 @@
...
+		const user = Promise.await(Users.findOneByAgentUsername(this.bodyParams.username, {
+			projection: {
+				_id: 1,
+				username: 1,
+			},
+		}));
+		if (!user) {
+			return API.v1.notFound();
+		}
+		Users.setExtension(user._id, this.bodyParams.extension);
```
murtaza98, line 54:
```suggestion
		Promise.await(Users.setExtension(user._id, this.bodyParams.extension));
```

app/api/server/v1/voip/omnichannel.ts
```diff
@@ -0,0 +1,102 @@
...
+		const user = Promise.await(Users.findOneByAgentUsername(this.requestParams().username, {
+			projection: {
+				_id: 1,
+				username: 1,
+			},
+		}));
+		if (!user) {
+			return API.v1.notFound();
+		}
+		Users.unsetExtension(user._id);
```
murtaza98, line 73:
```suggestion
		Promise.await(Users.unsetExtension(user._id));
```

app/models/server/models/Users.js
```diff
@@ -53,6 +53,7 @@ export class Users extends Base {
 		this.tryEnsureIndex({ 'services.saml.inResponseTo': 1 });
 		this.tryEnsureIndex({ openBusinessHours: 1 }, { sparse: true });
 		this.tryEnsureIndex({ statusLivechat: 1 }, { sparse: true });
+		this.tryEnsureIndex({ extension: 1 }, { sparse: true, unique: true });
```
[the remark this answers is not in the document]
amolghode1981: Yes. Thats correct.

server/services/omnichannel-voip/service.ts
amolghode1981: Removed the comment.

app/api/server/v1/voip/omnichannel.ts
```diff
@@ -0,0 +1,102 @@
+import { Match, check } from 'meteor/check';
+
+import { API } from '../../api';
+import { Users } from '../../../../models/server/raw/index';
+import { hasPermission } from '../../../../authorization/server/index';
+import { LivechatVoip } from '../../../../../server/sdk';
+
+API.v1.addRoute('omnichannel.agent.extension', { authRequired: true }, {
+	// Get the extensions associated with the agent passed as request params.
+	get() {
```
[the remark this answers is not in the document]
amolghode1981: I tried those. But I didnt get the data on the client side as expected.  I think that even though the function could be async, the framework does not call it with wait. So the data that we get on client side is not what we expect.
Correct me if I am missing something.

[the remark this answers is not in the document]
KevLehman: Ohh, I believe we're not using the most recent of develop yet 👀 support for async route handlers was introduced in this month. We should update base branches

app/api/server/v1/voip/omnichannel.ts
```diff
@@ -0,0 +1,102 @@
...
+		if (!hasPermission(this.userId, 'view-agent-extension-association')) {
+			return API.v1.unauthorized();
+		}
+		check(this.requestParams(), Match.ObjectIncluding({
+			username: String,
+		}));
+		const user = Promise.await(Users.findOneByAgentUsername(this.requestParams().username, {
```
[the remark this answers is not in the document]
amolghode1981: I tried making get asyn and use simple await for Users.findOneByAgentUsername. But I didnt get the data on the client side as expected. I think that even though the function could be async, the framework does not call it with wait. So the data that we get on client side is not what we expect.
Correct me if I am missing something.

app/api/server/v1/voip/omnichannel.ts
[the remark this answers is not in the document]
amolghode1981: Fixed.

app/api/server/v1/voip/omnichannel.ts
```diff
@@ -0,0 +1,102 @@
...
+		if (!user) {
+			return API.v1.notFound();
+		}
+		Users.unsetExtension(user._id);
+		return API.v1.success();
+	},
+});
+
+// Get free extensions
+API.v1.addRoute('omnichannel.extension.free', { authRequired: true,
```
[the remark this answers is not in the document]
amolghode1981: Will think about a different query parameter. I feel allocated free are opposites of each other. So I have discomfort in imagining when allocated = false, return free.

I can pass type = allocated | free and make the function accordingly.

server/services/omnichannel-voip/service.ts
```diff
@@ -1,10 +1,84 @@
...
+import { IOmnichannelVoipService } from '../../sdk/types/IOmnichannelVoipService';
 import { ServiceClass } from '../../sdk/types/ServiceClass';
+import { IVoipConnectorResult } from '../../../definition/IVoipConnectorResult';
+import { IVoipExtensionBase } from '../../../definition/IVoipExtension';
+import { Logger } from '../../lib/logger/Logger';
+import { Voip } from '../../sdk';
+import { IAgentExtensionMap, IOmnichannelVoipServiceResult } from '../../../definition/ILivechatVoipServiceResult';
+/**
+ * Note (Amol) :
+ * Why cant we get the direct access to the model here and have to work with
```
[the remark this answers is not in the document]
amolghode1981: Understand

server/services/omnichannel-voip/service.ts
amolghode1981: Extension list is coming from Asterisk so I dont think we can cache that. We may think of local allocations to be cached. But I will prefer taking  a conservative approach till we need optimisation.

app/api/server/v1/voip/omnichannel.ts
amolghode1981: Fixed.

app/api/server/v1/voip/omnichannel.ts
amolghode1981: Fixed.

server/services/omnichannel-voip/service.ts
amolghode1981: Fixed.

murtaza98 (approved):

murtaza98 (approved):

app/api/server/v1/voip/omnichannel.ts
```diff
@@ -0,0 +1,134 @@
...
+			},
+		}).toArray();
+		if (!extension) {
+			return API.v1.notFound('Extension not found');
+		}
+		return API.v1.success({ extension });
+	},
+
+	// Create agent-extension association.
+	async post() {
```
KevLehman, line 38: Should we validate extension is a number?

app/api/server/v1/voip/omnichannel.ts
```diff
@@ -0,0 +1,134 @@
...
+				const extension = await Users.getVoipExtensionByUserId(user._id, {
+					projection: {
+						_id: 1,
+						username: 1,
+						extension: 1,
+					},
+				}).toArray();
+				const freeExt = await LivechatVoip.getFreeExtensions();
+				const extensions = [];
+				for (let i = 0; i < extension.length; i++) {
```
KevLehman, line 122: I'd say you can benefit from the spread syntax here 👀 For example:

```
const available = [...freeExt, ...extension];
return API.v1.success({ available });
```

app/api/server/v1/voip/omnichannel.ts
amolghode1981: Well, SIP extensions can be non numeric texts. https://en.wikipedia.org/wiki/SIP_URI_scheme#:~:text=A%20SIP%20address%20is%20written,voip%2Dprovider.example.net&text=As%20a%20SIP%20address%20is,may%20contain%20non%2Dnumeric%20characters.

Right now I would keep it simple. But once we have people using it and we discover different formats people use, we can add some validations. 

Furthermore, as Diego pointed out yesterday, while doing this association, we may have to check on asterisk if the extension really exists. So I would prefer keeping it simple at this point of time.

app/api/server/v1/voip/omnichannel.ts
```diff
@@ -0,0 +1,134 @@
...
+		const user = await Users.findOneByAgentUsername(username, {
+			projection: {
+				_id: 1,
+				username: 1,
+			},
+		});
+		if (!user) {
+			return API.v1.notFound();
+		}
+		await Users.setExtension(user._id, extension);
```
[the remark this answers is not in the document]
amolghode1981: Fixed.

app/api/server/v1/voip/omnichannel.ts
```diff
@@ -0,0 +1,134 @@
...
+// Get free extensions
+API.v1.addRoute('omnichannel/extension', { authRequired: true,
+	permissionsRequired: ['manage-agent-extension-association'],
+}, {
+	async get() {
+		check(this.queryParams, Match.ObjectIncluding({
+			type: String,
+			agentId: Match.Maybe(String),
+		}));
+		const { type, username } = this.queryParams;
```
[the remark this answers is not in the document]
amolghode1981: Thanks for catching this. Fixed.

app/api/server/v1/voip/omnichannel.ts
amolghode1981: Correct me if I am wrong, but that may not work here by just using spread syntax. Extension is an allocated extension and it is coming from the database. freeExt is coming from connector and it is string array. So spread syntax my not just work.

app/models/server/raw/Users.js
```diff
@@ -768,4 +768,44 @@ export class UsersRaw extends BaseRaw {
...
+
+		return this.findOne(query, options);
+	}
+
+	getVoipExtensionByUserId(userId, options) {
+		const query = {
+			_id: userId,
+			extension: { $exists: true },
+		};
+		return this.find(query, options);
```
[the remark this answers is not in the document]
amolghode1981: Ok Cool. Fixed. this

definition/ILivechatVoipServiceResult.ts
```diff
@@ -0,0 +1,9 @@
+export interface IAgentExtensionMap {
```
KevLehman, line 1: Can you rename this file to `IOmnichannelVoipServiceResult`?

amolghode1981: Fixed.

KevLehman (approved):

PR #23737, opened 2021-11-17
[NEW] Allow registering by REG_TOKEN environment variable

geekgonecrazy:

You can provide the REG_TOKEN environment variable containing a registration token and it will automatically register to your cloud account.  This simplifies the registration flow

### DIFF
--- a/app/cloud/server/index.js
+++ b/app/cloud/server/index.js
@@ -6,9 +6,12 @@ import { getWorkspaceAccessToken } from './functions/getWorkspaceAccessToken';
 import { getWorkspaceAccessTokenWithScope } from './functions/getWorkspaceAccessTokenWithScope';
 import { getWorkspaceLicense } from './functions/getWorkspaceLicense';
 import { getUserCloudAccessToken } from './functions/getUserCloudAccessToken';
+import { retrieveRegistrationStatus } from './functions/retrieveRegistrationStatus';
 import { getWorkspaceKey } from './functions/getWorkspaceKey';
 import { syncWorkspace } from './functions/syncWorkspace';
+import { connectWorkspace } from './functions/connectWorkspace';
 import { settings } from '../../settings/server';
+import { SystemLogger } from '../../../server/lib/logger/system';
 
 const licenseCronName = 'Cloud Workspace Sync';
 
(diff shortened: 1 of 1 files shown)


app/cloud/server/index.js
```diff
@@ -34,6 +37,22 @@ Meteor.startup(function() {
...
+
+	if (process.env.REG_TOKEN && process.env.REG_TOKEN !== '' && !workspaceRegistered) {
+		try {
+			SystemLogger.info('REG_TOKEN Provided. Attempting to register');
+
+			if (!connectWorkspace(process.env.REG_TOKEN)) {
+				throw new Error('Couldn\'t register with token.  Please make sure token is valid or hasn\'t already been used');
+			}
+
+			console.log('Successfully registered with token provided by REG_TOKEN!');
```
KevLehman, line 51: Use SystemLogger here

app/cloud/server/index.js
```diff
@@ -34,6 +37,22 @@ Meteor.startup(function() {
 			job: syncWorkspace,
 		});
 	});
+
+	const { workspaceRegistered } = retrieveRegistrationStatus();
+
+	if (process.env.REG_TOKEN && process.env.REG_TOKEN !== '' && !workspaceRegistered) {
```
KevLehman, line 43: Maybe we can change this to test for negative to avoid nesting the if 👀

KevLehman (changes requested):

felipe-menelau (approved):
There a github warning regarding a Forbidden non-null assertion on the diff, but I'm guessing it's unrelated?
Otherwise LGTM!

app/cloud/server/index.js
geekgonecrazy: Which variation?  I want this one to always log

app/cloud/server/index.js
graywolf336: @KevLehman can you make a suggestion here with an example of what you mean?

KevLehman: It's something like:
```
if (!process.env.REG_TOKEN || process.env.REG_TOKEN.trim() === '' || workspaceRegistered) return;
try { ... }
```

app/cloud/server/index.js
pierre-lehnen-rc: On the default log level we only log errors. The next level is `SystemLogger.info`

graywolf336: So then console.log is the only option for logging 100% of the time without the log being a false error?

geekgonecrazy: Definitely seems this way

PR #23742, opened 2021-11-17
Chore: Update base docker image from node:12.22.1-buster-slim to node:12.22.7-buster-slim

rodrigok:


sampaiodiego:
should we close this one? 😬

PR #23745, opened 2021-11-18
Chore: Generic Table 

ggazzo:
We failed developing the V1, in the end of the day the generic was not so generic as we wanted, so I created a new one, a little less magic and demands a little more effort, but its generic at least, I moved one Table as proof of concept and changed the V1 internally to use the V2 until we finish the migration

:v:


client/components/GenericTable/GenericTable.tsx
```diff
@@ -57,41 +56,29 @@ const GenericTable: {
...
 		pagination = true,
 		...props
-	},
-	ref,
+	}: GenericTableProps<FilterProps, ResultProps>,
+	ref: Ref<HTMLElement>,
 ) {
 	const t = useTranslation();
 
 	const [filter, setFilter] = useState(paramsDefault);
```
ujorgeleite, line 64: From where we receive paramsDefault ?

client/components/GenericTable/V2/GenericTable.tsx
```diff
@@ -0,0 +1,25 @@
+/* eslint-disable react/no-multi-comp */
```
ujorgeleite, line 1: Is possible to refactor to remove this lint disable rule?

client/components/GenericTable/GenericTable.tsx
```diff
@@ -104,28 +91,18 @@ const GenericTable: {
...
-										))}
-									{children && (results ? results.map(children) : <Loading />)}
-								</Table.Body>
-							</Table>
-						</ScrollableContentWrapper>
-					</Box>
+					<GenericTableV2 fixed={fixed} ref={ref}>
+						{header && <GenericTableHeader>{header}</GenericTableHeader>}
+						<GenericTableBody>
+							{!results && <GenericTableLoadingTable headerCells={headerCells} />}
```
ujorgeleite, line 97: Can we break this with isLoading method to clarify the intention of this block and make it easy to read?

client/components/GenericTable/GenericTable.tsx
ggazzo: It doesn't matter actually, since this is the old GenericTable I kept the definitions and behaviors even if makes no sense

client/views/admin/customEmoji/CustomEmoji.tsx
```diff
@@ -0,0 +1,121 @@
...
+					{/* {phase === AsyncStatePhase.RESOLVED &&
+						!data.emojis.length
+						))} */}
+				</GenericTableBody>
+			</GenericTable>
+			{phase === AsyncStatePhase.RESOLVED && (
+				<Pagination
+					current={current}
+					itemsPerPage={itemsPerPage}
+					count={data?.count || 0}
```
gabriellsh, line 111: I think this should be `data?.total` like in the JS version. Table is not showing extra pages...

gabriellsh (changes requested):

gabriellsh (approved):
LGTM!

PR #23750, opened 2021-11-18
[NEW] REST endpoints to manage Omnichannel Business Units

KevLehman:

Basic documentation about endpoints can be found at https://www.postman.com/kaleman960/workspace/rocketchat-public-api/request/3865466-71502450-8c8f-42b4-8954-1cd3d01fcb0c

### DIFF
--- a/definition/ILivechatBusinessUnit.ts
+++ b/definition/ILivechatBusinessUnit.ts
@@ -1,0 +1,9 @@
+export interface ILivechatBusinessUnit {
+	_id: string;
+	name: string;
+	visibility: 'public' | 'private';
+	type: string;
+	numMonitors: number;
+	numDepartments: number;
+	_updatedAt: Date;
+}
--- a/ee/app/livechat-enterprise/server/api/units.ts
+++ b/ee/app/livechat-enterprise/server/api/units.ts
@@ -1,14 +1,15 @@
 import { API } from '../../../../../app/api/server';
 import { findUnits, findUnitById, findUnitMonitors } from './lib/units';
 import { LivechatEnterprise } from '../lib/LivechatEnterprise';
+import { ILivechatBusinessUnit } from '../../../../../definition/ILivechatBusinessUnit';
 
 API.v1.addRoute('livechat/units.list', { authRequired: true }, {
-	get() {
+	async get() {
 		const { offset, count } = this.getPaginationItems();
 		const { sort } = this.parseJsonQuery();
 		const { text } = this.queryParams;
 
-		const response = Promise.await(findUnits({
+		const response = await findUnits({
 			userId: this.userId,
 			text,
 			pagination: {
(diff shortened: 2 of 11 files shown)


ggazzo (dismissed):
what do you think about creating this endpoints under a ts file?

KevLehman:
It was a small pain to test (I had a lot of issues with the `PUT` method, so I changed it to be POST for now :sad:) but it's there.
@ggazzo would you mind taking a look? 👀

definition/ILivechatBusinessUnit.ts
```diff
@@ -0,0 +1,9 @@
+export interface ILivechatBusinessUnit {
```
murtaza98, line 1:
```suggestion
export interface IOmnichannelBusinessUnit {
```

murtaza98: Can we please change this to omnichannel, since by livechat we usually refer to the Livechat widget, but these units will be used across entire omnichannel

ee/app/livechat-enterprise/server/api/units.ts
```diff
@@ -0,0 +1,92 @@
...
+
+		return API.v1.success(await findUnitMonitors({
+			userId: this.userId,
+			unitId,
+		}));
+	},
+});
+
+API.v1.addRoute('livechat/units', { authRequired: true, permissionsRequired: ['manage-livechat-units'] }, {
+	async get() {
```
murtaza98, line 50: @KevLehman Y r we creating a new endpoint when we already have one with the same functionality? Talking about this one here - `livechat/units.list` - GET

ee/app/livechat-enterprise/server/api/units.ts
```diff
@@ -0,0 +1,92 @@
...
+		})));
+	},
+	async post() {
+		const { unitData, unitMonitors, unitDepartments } = this.bodyParams?.();
+		return LivechatEnterprise.saveUnit(null, unitData, unitMonitors, unitDepartments);
+	},
+});
+
+API.v1.addRoute('livechat/units/:id', { authRequired: true, permissionsRequired: ['manage-livechat-units'] }, {
+	async get() {
```
murtaza98, line 72: Same goes here. I think we already have an endpoint which fetches this data - `livechat/units.getOne` (GET)

KevLehman: Yup, the original idea was to mark the others as deprecated in favor of these ones (which are a bit more restful than `.getOne`, etc) but I had some issues that I'm still solving with the deprecationWarning helper.

murtaza98: Ahh.. okay makes sense. It's fine then.

murtaza98 (approved):

ee/app/livechat-enterprise/server/api/units.ts
```diff
@@ -0,0 +1,93 @@
...
+		const { id } = this.urlParams;
+		const { unit } = await findUnitById({
+			userId: this.userId,
+			unitId: id,
+		}) as { unit: IOmnichannelBusinessUnit };
+
+		return API.v1.success(unit);
+	},
+	async post() {
+		const { unitData, unitMonitors, unitDepartments } = this.bodyParams?.();
```
renatobecker, line 83: What happens inside `LivechatEnterprise.saveUnit` if either `unitMonitors` or `unitDepartments` are not provided?

murtaza98: It will throw an error. We have a check for that over there - see [here](https://github.com/RocketChat/Rocket.Chat/blob/d6a25f550ac8aa59f235f332e9648f29cfc3f421/ee/app/livechat-enterprise/server/lib/LivechatEnterprise.js#L75)

KevLehman: Yup, all of those are required params 👀

ggazzo:
> It was a small pain to test (I had a lot of issues with the `PUT` method, so I changed it to be POST for now :sad:) but it's there. @ggazzo would you mind taking a look? 👀
It was in the right path but I dont think that was enough , we have a lot of implicit `any` through the functions, that gives us almost zero guarantee about the types.

@murtaza98 I invite you to review more calmly, spending more time on the task and be more discerning in your reviews.

tassoevan:
Just found an issue [adding more strict typings](#23735). I've spotted an endpoint without `params`:

```ts
	'livechat/units.getOne': {
		GET: () => (IOmnichannelBusinessUnit);
	};
```

And here is how is implemented:

```ts
API.v1.addRoute('livechat/units.getOne', { authRequired: true }, {
	async get() {
		const { id } = this.urlParams; // we DON'T have an `:id` at the path pattern
		const { unit } = await findUnitById({
			userId: this.userId,
			unitId: id,
		}) as { unit: IOmnichannelBusinessUnit };

		return API.v1.success(deprecationWarning({ response: unit, endpoint: 'livechat/units.getOne' }));
	},
});
```

This is wrong by definition, and I confirm it here:

```js
function UnitEditWithData({ unitId, reload, title }) {
	const query = useMemo(() => ({ unitId }), [unitId]); // the params should at least be { unitId: string }
	const { value: data, phase: state, error } = useEndpointData('livechat/units.getOne', query);
```

You guys authorize me to make the relevant changes? I'll mark @RocketChat/Omnichannel as reviewer.

tassoevan:
A small nitpick: it's better to enforce the type at the variable rather than asserting its value:

```ts
		const { unit }: { unit: IOmnichannelBusinessUnit } = await findUnitById({
			userId: this.userId,
			unitId: id,
		});
```

The reason is not so obvious, but I think is worthy to share that, once `findUnitById` get typed, we can skip a wrong type assertion and let a failure point for `tsc` help us.

ggazzo:
that why I always strongly suggest convert the files affected to TS , to prevent fix -> regression loops

PR #23758, opened 2021-11-19
Chore: Type omnichannel models

KevLehman:


ggazzo (commented):
we should avoid using `any` at all costs, as temporary alternative `unknown` can be used

ggazzo (approved):

PR #23769, opened 2021-11-22
Chore: Update settings.ts

ggazzo:


sampaiodiego (approved):

sampaiodiego (changes requested):
migration file has no extension and it is missing being imported.

sampaiodiego (approved):

PR #23781, opened 2021-11-23
[IMPROVE] Rewrite remove room invite modal

dougfabris:
### before
image

### after
image

## Steps to test or reproduce

- First create a invite link: go to some room, click on members and invite link button
- Go to admin > Invites
- Click on the "X" button in some invite row to see the changes


client/views/admin/invites/InviteRow.tsx
```diff
@@ -1,25 +1,43 @@
...
+	const daysToExpire = ({
+		expires,
+		days,
+	}: {
+		expires: IInvite['expires'];
+		days: IInvite['days'];
+	}): string => {
 		if (days > 0) {
-			if (expires < Date.now()) {
+			if (expires && expires.getTime() < new Date().getTime()) {
```
ujorgeleite, line 40: maybe decoupling a method isExpired here. Is possible to get days null or undefined?

client/views/admin/invites/InviteRow.tsx
```diff
@@ -29,7 +47,13 @@ function InviteRow({ _id, createdAt, expires, days, uses, maxUses, onRemove }) {
...
-	const maxUsesLeft = ({ maxUses, uses }) => {
+	const maxUsesLeft = ({
+		maxUses,
+		uses,
+	}: {
+		maxUses: IInvite['maxUses'];
+		uses: IInvite['uses'];
+	}): number | string => {
 		if (maxUses > 0) {
 			if (uses >= maxUses) {
```
ujorgeleite, line 58: Here we can decouple a method for excededMaximun or something like this I think

ujorgeleite (changes requested):

client/views/admin/invites/InviteRow.tsx
dougfabris: It's possible to get null

client/views/admin/invites/InviteRow.tsx
dougfabris: I don't think it deserves a method to check this simple thing!

client/views/admin/invites/InviteRow.tsx
```diff
@@ -0,0 +1,130 @@
...
+			return maxUses - uses;
+		}
+
+		return t('Unlimited');
+	};
+
+	const handleRemoveButtonClick = async (event: MouseEvent<HTMLElement>): Promise<void> => {
+		event.stopPropagation();
+
+		imperativeModal.open({
```
gabriellsh, line 79: Please, use the modal context istead of this.

gabriellsh (changes requested):

client/views/admin/invites/InviteRow.tsx
ujorgeleite: When someone goes offer manutention on this code, maybe will be better if can read what are you checking directly and not to be necessary to interpret the two-block of validation, that was my motivation to suggest this for you, but I'll approve the review.

ujorgeleite (dismissed):

gabriellsh (approved):

gabriellsh (approved):

PR #23791, opened 2021-11-25
[FIX] Modal keeps state if reset too fast.

gabriellsh:

Queued updates so the Modal has a chance to close.


tassoevan (changes requested):
I've suggested privately to @gabriellsh to use unique `key` props to ensure the modal state is dropped between two `setModal()` calls. It will require only the last `useEffect` hook.

tassoevan (approved):

PR #23792, opened 2021-11-25
Chore: PoC prettier

KevLehman:


ggazzo:
❤️ 

my little secret: 
image
I have been using for a while, I'm 100% with you

sampaiodiego (changes requested):
have you tried using the client's prettier config? https://github.com/RocketChat/Rocket.Chat/blob/develop/client/.prettierrc

or maybe another question, should we use the same config for the entire repo?

KevLehman:
> I have been using for a while, I'm 100% with you

I knew your formatting was too perfect <3 

> or maybe another question, should we use the same config for the entire repo?

Well that's a good question. I don't know 😬 should we? Would be good to have an "standard" style for the whole codebase, but, FE & BE may have different needs (?) (not sure tho)

sampaiodiego:
I think it works if we have a `.prettierrc` on root and others with specific rules in different folders.. maybe you can try that? 😬

KevLehman:
> I think it works if we have a .prettierrc on root and others with specific rules in different folders.. maybe you can try that? 😬

That could be, any rule you want in particular for BE? Maybe Gazzo can share his prettier 👀 (for reference purposes only)

sampaiodiego:
I guess you can move the one from `/client/.prettierrc` to root and see what happens 😬

PR #23816, opened 2021-11-29
Chore: Centralize email validation functionality

KevLehman:

- Create lib for validating emails
- Modify places that validate emails to use the new central function

### DIFF
--- a/lib/emailValidator.ts
+++ b/lib/emailValidator.ts
@@ -1,5 +1,5 @@
 export const validateEmail = (email: string, options: { style: string } = { style: 'basic' }): boolean => {
-	const basicEmailRegex = /^.+@.+$/;
+	const basicEmailRegex = /^[^@]+@[^@]+$/;
 	const rfcEmailRegex = /^[a-zA-Z0-9.!#$%&'*+\/=?^_`{|}~-]+@[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?(?:\.[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?)*$/;
 
 	switch (options.style) {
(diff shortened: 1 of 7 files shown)


lib/emailValidator.ts
```diff
@@ -0,0 +1,13 @@
...
+	const basicEmailRegex = /^.+@.+$/;
+	const rfcEmailRegex = /^[a-zA-Z0-9.!#$%&'*+\/=?^_`{|}~-]+@[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?(?:\.[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?)*$/;
+
+	switch (options.style) {
+		case 'basic':
+			return basicEmailRegex.test(email);
+		case 'rfc':
+			return rfcEmailRegex.test(email);
+		default:
+			return basicEmailRegex.test(email);
```
tassoevan, line 11:
```suggestion
		case 'rfc':
			return rfcEmailRegex.test(email);
		case 'basic':
		default:
			return basicEmailRegex.test(email);
```

lib/emailValidator.ts
```diff
@@ -0,0 +1,13 @@
+export const validateEmail = (email: string, options: { style: string } = { style: 'basic' }): boolean => {
+	const basicEmailRegex = /^.+@.+$/;
```
sampaiodiego, line 2: looking into some regex you modified to use this function I found this one interesting: `[^@].*@[^@]`, it makes sure only a single `@` is allowed.. wdyt? 👀

lib/emailValidator.ts
KevLehman: I thought we had the no-fallthrough rule enabled 🤔

lib/emailValidator.ts
KevLehman: image
Not sure 🙈 if we should use it tho

sampaiodiego: oopsie.. looks like I missread the regex 🙈

sampaiodiego: ok, this one 😬  `^[^@]+@[^@]+$`

sampaiodiego (approved):

sampaiodiego:
Although I have approved it already, what about writing some unit tests for it? maybe we can have an agreement that all functions in `/lib` should have unit tests.. wdyt?

PR #23824, opened 2021-11-30
[CHORE] Improvements on role syncing (ldap, oauth and saml)

pierre-lehnen-rc:
- created a single syncUserRoles function to be used by all login methods that can add/remove user roles.
- converted `addUserRoles` and `removeUserFromRoles` to typescript
- validate number of user seats available before converting a guest into a full user through role syncing.


tassoevan (approved):

server/startup/initialData.js
```diff
@@ -28,7 +28,7 @@ Meteor.startup(async function() {
 			type: 'bot',
 		});
 
-		addUserRoles('rocket.cat', 'bot');
+		addUserRolesSync('rocket.cat', 'bot');
```
sampaiodiego, line 31: this function is already `async`, so:
```suggestion
		await addUserRoles('rocket.cat', 'bot');
```

server/startup/initialData.js
```diff
@@ -112,7 +112,7 @@ Meteor.startup(async function() {
 
 			Accounts.setPassword(id, process.env.ADMIN_PASS);
 
-			addUserRoles(id, 'admin');
+			addUserRolesSync(id, 'admin');
```
sampaiodiego, line 115:
```suggestion
			await addUserRoles(id, 'admin');
```

server/startup/initialData.js
```diff
@@ -138,7 +138,7 @@ Meteor.startup(async function() {
 		const oldestUser = Users.getOldest({ _id: 1, username: 1, name: 1 });
 
 		if (oldestUser) {
-			addUserRoles(oldestUser._id, ['admin']);
+			addUserRolesSync(oldestUser._id, ['admin']);
```
sampaiodiego, line 141:
```suggestion
			await addUserRoles(oldestUser._id, ['admin']);
```

server/startup/initialData.js
```diff
@@ -189,7 +189,7 @@ Meteor.startup(async function() {
 
 		Accounts.setPassword(adminUser._id, adminUser._id);
 
-		addUserRoles(adminUser._id, ['admin']);
+		addUserRolesSync(adminUser._id, ['admin']);
```
sampaiodiego, line 192:
```suggestion
		await addUserRoles(adminUser._id, ['admin']);
```

app/authentication/server/startup/index.js
```diff
@@ -295,7 +295,7 @@ Accounts.insertUserDoc = _.wrap(Accounts.insertUserDoc, function(insertUserDoc,
 		}
 	}
 
-	addUserRoles(_id, roles);
+	Promise.await(addUserRoles(_id, roles));
```
sampaiodiego, line 298: any reason to not use the "sync" version here?

server/lib/arrayUtils.ts
```diff
@@ -0,0 +1,4 @@
+export function ensureArray<T>(param: T | Array<T>): Array<T> {
```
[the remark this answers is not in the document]
sampaiodiego: agreed..

I've used `[].concat(...)` in many places already, do you guys think we need a utility function just for this? I mean, it is very small and feels very intuitive.. 😬

sampaiodiego (changes requested):
I saw you have created "sync" variants of some old functions.. the way we did before was keeping the old function as they were but creating an "async" version of them instead (see `hasPermission` and `hasPermissionAsync`). it concerns me a little bit we mix both.. should we stick to what we've been doing far? wdyt?

server/lib/arrayUtils.ts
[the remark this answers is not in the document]
pierre-lehnen-rc: It's not so small in typescript.

pierre-lehnen-rc:
> I saw you have created "sync" variants of some old functions.. the way we did before was keeping the old function as they were but creating an "async" version of them instead (see `hasPermission` and `hasPermissionAsync`). it concerns me a little bit if we mix both "patterns".. should we stick to what we've been doing far? wdyt?

I think at some point we should switch to having the Async be the default and the Sync be the exception, but it would be best to update the old functions when we do so. I can switch the names around for this one for now.

sampaiodiego:
> I think at some point we should switch to having the Async be the default and the Sync be the exception, but it would be best to update the old functions when we do so. I can switch the names around for this one for now.

yes, that's definitely the idea. the "sync" versions are around just to make our lives easier during the migration. but at any time that we found we don't need the "sync" version anymore we should get rid of them.

sampaiodiego (approved):
Just need to fix conflicts

sampaiodiego (approved):

PR #23845, opened 2021-12-02
[FIX] Add user when mentioned in thread for public rooms

### LINKED ISSUE #17599: Adding users with mentions doesn't work inside of threads
### Description:
 
When adding users to public channels via `@user`-mention it only works in the main Chat, not in a thread.

### Steps to reproduce:

1. in a Public room
2. where `user` isnt in yet
3. in a Thread
4. mention someone `@user`

### Expected behavior:

5. `user` would join the room.

### Actual behavior:

5. `user` does not join the room.

### Server Setup Information:

- Version of Rocket.Chat Server: 3.2.1
- Operating System: lubuntu
- Deployment Method: docker
- DB Replicaset Oplog: ?
- NodeJS Version: ?
- MongoDB Version: ?

### Client Setup Information

- Desktop App or Browser Version: Firefox 76
- Operating System: Qubes OS (Debian VM)

yash-rajpal:
## Issue(s)

fix #17599

### DIFF
--- a/app/threads/server/hooks/aftersavemessage.js
+++ b/app/threads/server/hooks/aftersavemessage.js
@@ -1,11 +1,12 @@
 import { Meteor } from 'meteor/meteor';
 
-import { Messages } from '../../../models/server';
+import { Messages, Subscriptions } from '../../../models/server';
 import { callbacks } from '../../../callbacks/server';
 import { settings } from '../../../settings/server';
 import { reply } from '../functions';
 import { updateThreadUsersSubscriptions, getMentions } from '../../../lib/server/lib/notifyUsersOnMessage';
-import { sendMessageNotifications } from '../../../lib/server/lib/sendNotificationsOnMessage';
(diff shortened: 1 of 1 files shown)


tassoevan (changes requested):
1. It seems we've been invoking `callJoinRoom` just to make sure the notifications reach the mentioned users. Maybe is time to acknowledge that these behaviors are unrelated: we need to _join mentioned users_ and _notify them_, as separated features (despite being sequential). Could you split the callbacks on `afterSaveMessage`, each one handling one of those? Doing it will enable us to handle all the joinings in a single place without needing to check if a message is at some thread or not.
2. One thing to check when you're performing join in a room is that channels, despite being public, can be nested in private teams. We need to ensure that the user is on the same team in this case.

tassoevan (changes requested):
One thing that worries me now is that we're mixing asynchronous code with fiber code; as example, `Messages` and `Subscriptions` models are supposed to be used in fibers. The problem looks a little bit worse when we go down to helper functions. I wanna get some feedback from @pierre-lehnen-rc on how we can make this hook be more reliable.

Issue #11598, opened 2018-07-29
After updating to 0.68.0 Sidebar shows Usernames instead of Real Names


johnyb0y:
### Description:

After updating to 0.68.0 the **left sidebar** displays usernames instead of Real names.
I already searched settings (twice) and as far as I can tell every setting is set to Use Real Names.
Have i overlooked something ?

**EDIT:** Just to be clear: Before the update the sidebar showed the Real Names, not the usernames.

### Steps to reproduce:

1. Update to 0.68.0 from 0.65.2

### Expected behavior:

Setting shouldn't change. Also toggling the various "Use Real Names" settings should have an effect and configure the sidebar to use Real Names again.

### Actual behavior:

Username is shown and settings don't seem to change anything regarding the sidebar.

### Server Setup Information:

- Version of Rocket.Chat Server: 0.68.0
- Operating System: Debian 9.5
- Deployment Method: tar
- Number of Running Instances: 2
- DB Replicaset Oplog: Yes
- NodeJS Version: 8.11.3
- MongoDB Version: Up-to-date Debian Stable Package

Thanks.

himpierre:
Just took a look. I guess it was always like that? The sidbar only shows the username. No? Just checked my 0.66.0 installation.

baevga:
I have the same problem too. i was using 0.65.1 version and can to show realnames in sidebar. After update this option not work and all current chats became a login. We using numeric logins and that so uncomfortable

johnyb0y:
@himpierre Nope, it worked before. We used real names in sidebar since at least 6 months.

baevga:
@johnyb0y 
I'm talking about  that problem https://prnt.sc/kcdist . I deleted one chat  and start new one. After that it has real name in it's title like that http://prntscr.com/kcdjhk

Your issue about another problem?

johnyb0y:
That looks like the same problem I'm experiencing. But for me, deleting a chat and starting a new one doesn't help.

You also said something about your login policy not working anymore. That might be worth a separate issue.

gabrielloberg:
Same problem here.

himpierre:
Which setting enables that feature? I have set "Set a User Name to Alias in Message" to true, but thats it.

johnyb0y:
@himpierre which Setting specifically no idea. There are various „Use real name“ Settings spread across the configuration menu.

baevga:
[...]

PR #23848, opened 2021-12-02
[FIX] DMs being created with username instead of user's name

### LINKED ISSUE #23093: [Bug] 4.1.2 does not show the real name of new users in several places.
### Description:

The option to use real name is checked from previous versions.

After upgrading to 3.18.0 new registered users do not show real name in several places.

On the top and left side bar the user is shown and the real name should be shown.

### Steps to reproduce:

1. Register a new user manually, also tested with users coming from LDAP

### Actual behavior:

image

### Server Setup Information:

- Version of Rocket.Chat Server: 3.18.0
- Operating System: Ubuntu 20.04
- Deployment Method: tar
- DB Replicaset Oplog: wiredTiger
- NodeJS Version: 12.22.1
- MongoDB Version: 4.4.8

### Additional context

@pierre-lehnen-rc and @ankar84

This issue is closely related to #19046 since version 3.9.3. The difference is that it now directly affects the new users who register in the system.

The option **Use real name** is updated correctly, even if the modification comes from LDAP, the only problem is that these values are not shown in the left column and in the title of the chat, however in the messages it is shown correctly.

gabriellsh:
## Issue(s)

closes #23093


ankar84:
@gabriellsh hey!
What do you think is that PR fix that issue https://github.com/RocketChat/Rocket.Chat/issues/23093 ?
Or it fix different issue?

gabriellsh:
> @gabriellsh hey! What do you think is that PR fix that issue #23093 ? Or it fix different issue?

hey @ankar84 ! Reading through that issue, I believe this is the fix. Thanks for the tip! I'll close it when this is merged.

dougfabris (approved):

sampaiodiego (commented):
do you mind adding tests @gabriellsh ? 🤗 

thx for the fix anyways

tests/end-to-end/api/04-direct-message.js
```diff
@@ -481,34 +485,55 @@ describe('[Direct Messages]', function() {
...
+			userCredentials = await login(user.username, password);
+			otherUserCredentials = await login(otherUser.username, password);
+			thirdUserCredentials = await login(thirdUser.username, password);
 		});
 
 		after(async () => {
-			if (roomId) {
-				await deleteRoom({ type: 'd', roomId });
-			}
+			Promise.all(Object.values(roomIds).map((roomId) => deleteRoom({ type: 'd', roomId })));
```
sampaiodiego, line 515:
```suggestion
			await Promise.all(Object.values(roomIds).map((roomId) => deleteRoom({ type: 'd', roomId })));
```

tests/end-to-end/api/04-direct-message.js
```diff
@@ -517,35 +542,66 @@ describe('[Direct Messages]', function() {
...
+					expect(res.body.room).to.have.property('usernames').and.to.have.members([user.username]);
+					roomIds = { ...roomIds, self: res.body.room._id };
 				})
 				.end(done);
 		});
+
+		async function testRoomFNameForUser(credentials, rid, fullName) {
+			return request.get(api('subscriptions.getOne'))
+				.set(credentials)
+				.query({ roomId: rid })
```
sampaiodiego, line 579: being very picky here.. but since there is a `credentials` variable on the scope above, would be good to call this different to avoid confusion..

what about changing to `roomId` as well to make things cleaner 😬 

```suggestion
		async function testRoomFNameForUser(testCredentials, roomId, fullName) {
			return request.get(api('subscriptions.getOne'))
				.set(testCredentials)
				.query({ roomId })
```

tests/end-to-end/api/04-direct-message.js
```diff
@@ -517,35 +542,66 @@ describe('[Direct Messages]', function() {
...
+				.expect('Content-Type', 'application/json')
+				.expect((res) => {
+					expect(res.body).to.have.property('success', true);
+					expect(res.body).to.have.property('subscription').and.to.be.an('object');
+					expect(res.body.subscription).to.have.property('fname', fullName);
+				});
+		}
+
+		describe('Rooms fullName', () => {
+			it('should be own user\'s name for self dm', async () => {
```
sampaiodiego, line 590:
```suggestion
			it('should be own user\'s name for self DM', async () => {
```

tests/end-to-end/api/04-direct-message.js
```diff
@@ -517,35 +542,66 @@ describe('[Direct Messages]', function() {
...
+				await testRoomFNameForUser(userCredentials, roomIds.self, userFullName);
+			});
+
+			it('should be other user\'s name concatenated for multiple users\'s dm for every user', async () => {
```
sampaiodiego, line 594:
```suggestion
			it('should be other user\'s name concatenated for multiple users\'s DM for every user', async () => {
```

tests/end-to-end/api/04-direct-message.js
```diff
@@ -517,35 +542,66 @@ describe('[Direct Messages]', function() {
...
+				await testRoomFNameForUser(userCredentials, roomIds.multipleDm, [otherUserFullName, thirdUserFullName].join(', '));
+				await testRoomFNameForUser(otherUserCredentials, roomIds.multipleDm, [userFullName, thirdUserFullName].join(', '));
+				await testRoomFNameForUser(thirdUserCredentials, roomIds.multipleDm, [userFullName, otherUserFullName].join(', '));
+			});
+
+			it('should be other user\'s name for dm for both users', async () => {
```
sampaiodiego, line 600:
```suggestion
			it('should be other user\'s name for DM for both users', async () => {
```

sampaiodiego (changes requested):
found a missing `await` on the `Promise.all` call..

added other suggestions for improving readability and consistency :)

sampaiodiego:
oh, awesome job overall btw.. that's a single line fix that required a lot of tests to prove it works.. thanks for this great improvements

sampaiodiego (approved):

a89h4ya:
#11598 says that the issue is also in 3.18. Shouldnt the fix be backported to that major version too?

PR #23865, opened 2021-12-04
[FIX] undefined error at toolbox as props.id isn't present

sidmohanty11:
props.id is undefined each time, does it check for the label name? 
  

Before,

https://user-images.githubusercontent.com/73601258/144701246-47815f4a-d8aa-46ff-a506-60f19528f8f6.mp4

After,

https://user-images.githubusercontent.com/73601258/144701251-9ed78152-abbc-45a6-9627-6bd4afa0cd56.mp4

## Steps to test or reproduce

Click on the 3 dots present in the Header of any channel, screen goes blank.


sidmohanty11:
Sorry, didn't see that already a PR has been made

PR #23875, opened 2021-12-06
Chore: update docker image base to latest node 12 patch

debdutdeb:


.docker-mongo/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM node:12.22.1-buster-slim
+FROM node:12.22.7-buster-slim
```
sampaiodiego, line 1:
```suggestion
FROM node:12.22.8-buster-slim
```

.docker/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM node:12.22.1-buster-slim
+FROM node:12.22.7-buster-slim
```
sampaiodiego, line 1:
```suggestion
FROM node:12.22.8-buster-slim
```

sampaiodiego (changes requested):
there was a new patch in the meanwhile 😬

sampaiodiego:
@debdutdeb I've updated the node version on github workflows and everywhere else, so we're sure we're building and testing using the same version as we're publishing our docker images  👍

sampaiodiego (approved):

PR #23904, opened 2021-12-08
[FIX] Jitsi call already ended

### LINKED ISSUE #23903: Jitsi: Click to join button returns "call already ended" after 10s when Jitsi is configured to open "always in new window"
### Description:

Hello Devs.

I know there are a lot of similar tickets and most of them are fixed but my issue isn't. When setting Jitsi to "always open in new window" in Administration -> Video Conference -> Jitsi the "click to join" button only works for 10s. After that the famous "call already ended" message appears although the jitsi room still exists. One still can join the jitsi call via phone icon. Interesting enough.

### Steps to reproduce:

1. Set Administration -> Video Conference -> Jitsi -> "always open in new window" 
2. Open Jitsi Room via phone icon.
3. Wait 15s
4. Try to join the room via "click to join"
5. "call already ended" appears

### Expected behavior:

"click to join" button works as long as the jitsi room exists.

### Actual behavior:

After 10s "click to join" does not work anymore.

### Server Setup Information:

- Version of Rocket.Chat Server: 4.2.0
- Operating System: Linux
- Deployment Method: tar
- DB Replicaset Oplog: sure
- NodeJS Version: 12.22.6
- MongoDB Version: 4.4.8

### Client Setup Information

- Desktop App or Browser Version: Firefox 95 / App 3.7.0
- Operating System: Linux

Aman-Maheshwari:
- [x] I have read the Contributing Guide
  - [x] I have signed the CLA 
  - [ ] I have added necessary documentation (if applicable)
  - [ ] Any dependent changes have been merged and published in downstream modules

 - Fix Jitsi timeout update -- which caused the "Jitsi call already ended" error when trying to join a call some time after its creation;

## Before

https://user-images.githubusercontent.com/50165440/145302813-02e5bddb-3cf5-4236-9e8d-61d097efebad.mp4

## After 

https://user-images.githubusercontent.com/50165440/145302063-97830dbe-1b96-4b5b-9767-7875b4c92e55.mp4

## Issue(s)

Fix #23903 

## Steps to test or reproduce

Mentioned in the issue

### DIFF
--- a/client/views/room/contextualBar/Call/Jitsi/CallJitsiWithData.js
+++ b/client/views/room/contextualBar/Call/Jitsi/CallJitsiWithData.js
@@ -34,7 +34,6 @@ const querySettings = {
 };
 
 const CallJitsiWithData = ({ rid }) => {
-	// console.log('rid = ', rid);
 	const user = useUser();
 	const { connected } = useConnectionStatus();
 	const [accessToken, setAccessToken] = useSafely(useState());
--- a/client/views/room/contextualBar/Call/Jitsi/CallJitsiWithData.js
+++ b/client/views/room/contextualBar/Call/Jitsi/CallJitsiWithData.js
@@ -34,6 +34,7 @@ const querySettings = {
 };
 
 const CallJitsiWithData = ({ rid }) => {
+	// console.log('rid = ', rid);
 	const user = useUser();
 	const { connected } = useConnectionStatus();
 	const [accessToken, setAccessToken] = useSafely(useState());
@@ -130,7 +131,7 @@ const CallJitsiWithData = ({ rid }) => {
 				return jitsi.dispose();
 			}
 			try {
-				await updateTimeout(rid, false);
+				return updateTimeout(rid, false);
 			} catch (error) {
 				dispatchToastMessage({ type: 'error', message: t(error.reason) });
 				clear();
@@ -144,7 +145,7 @@ const CallJitsiWithData = ({ rid }) => {
 
 		if (new Date() - new Date(room.jitsiTimeout) + TIMEOUT > DEBOUNCE) {
 			try {
-				await updateTimeout(rid, false);
+				return updateTimeout(rid, false);
 			} catch (error) {
 				dispatchToastMessage({ type: 'error', message: t(error.reason) });
 				clear();
@@ -182,7 +183,7 @@ const CallJitsiWithData = ({ rid }) => {
 
 			jitsi.on('HEARTBEAT', testAndHandleTimeout);
 
-			shouldDispose = jitsi.openNewWindow;
+			shouldDispose = !jitsi.openNewWindow;
 		}
 
 		fetchData().then(() => {
--- a/client/views/room/contextualBar/Call/Jitsi/CallJitsiWithData.js
+++ b/client/views/room/contextualBar/Call/Jitsi/CallJitsiWithData.js
@@ -163,7 +163,6 @@ const CallJitsiWithData = ({ rid }) => {
 			jitsi.off('HEARTBEAT', testAndHandleTimeout);
 			jitsi.dispose();
 		};
-		const none = () => {};
 
 		try {
 			if (jitsi.needsStart) {
@@ -179,7 +178,9 @@ const CallJitsiWithData = ({ rid }) => {
 		}
 		jitsi.on('HEARTBEAT', testAndHandleTimeout);
 
-		return jitsi.openNewWindow ? none : clear;
+		return () => {
+			if (!jitsi.openNewWindow) clear();
+		};
 	}, [
 		accepted,
 		jitsi,
--- a/client/views/room/contextualBar/Call/Jitsi/CallJitsiWithData.js
+++ b/client/views/room/contextualBar/Call/Jitsi/CallJitsiWithData.js
@@ -155,41 +155,31 @@ const CallJitsiWithData = ({ rid }) => {
 	});
 
 	useEffect(() => {
-		let shouldDispose = false;
-
-		async function fetchData() {
-			if (!accepted || !jitsi) {
-				return;
-			}
-
-			const clear = () => {
-				jitsi.off('HEARTBEAT', testAndHandleTimeout);
-				jitsi.dispose();
-			};
-
-			try {
-				if (jitsi.needsStart) {
-					jitsi.start(ref.current);
-					await updateTimeout(rid, true);
-				} else {
-					await updateTimeout(rid, false);
-				}
-			} catch (error) {
-				dispatchToastMessage({ type: 'error', message: t(error.reason) });
-				clear();
-				handleClose();
-			}
-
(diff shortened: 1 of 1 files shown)


Aman-Maheshwari:
@dudanogueira Thank you for the appreciation. :smile:

arunodhayam10d:
Hi Guys, I am using rocket chat in docker version 4.2.0. I am also facing this call ended issue. Is this fixed in docker-version??

client/views/room/contextualBar/Call/Jitsi/CallJitsiWithData.js
```diff
@@ -182,7 +182,7 @@ const CallJitsiWithData = ({ rid }) => {
 
 			jitsi.on('HEARTBEAT', testAndHandleTimeout);
 
-			shouldDispose = jitsi.openNewWindow;
+			shouldDispose = !jitsi.openNewWindow;
```
matheusbsilva137, line 185:
```suggestion
			shouldDispose = jitsi.openNewWindow;
```
With this change, calls are not working when the "Always open in new window" setting is off (since the call would be disposed). I think we could keep it as it was and your fix would still work well, right?

matheusbsilva137 (changes requested):

client/views/room/contextualBar/Call/Jitsi/CallJitsiWithData.js
Aman-Maheshwari: can you tell me what should be the behaviour ? In which state should we dispose the call and in which we should not ?
As per the suggested changes if `shouldDispose = jitsi.openNewWindow`  then 
```
fetchData().then(() => {
			if (shouldDispose) {
				jitsi.dispose();
			}
		});
```

this always runs and dispose the call as soon as it starts when the call is opened in the new window.

client/views/room/contextualBar/Call/Jitsi/CallJitsiWithData.js
```diff
@@ -155,41 +155,31 @@ const CallJitsiWithData = ({ rid }) => {
...
+			handleClose();
 		}
+		jitsi.on('HEARTBEAT', testAndHandleTimeout);
 
-		fetchData().then(() => {
-			if (shouldDispose) {
-				jitsi.dispose();
-			}
-		});
+		return jitsi.openNewWindow ? none : clear;
```
matheusbsilva137, line 182:
```suggestion
		if (!jitsi.openNewWindow) {
			return clear;
		}
```
Maybe we could change the return statement so that we could delete this `none` function, wdyt?

client/views/room/contextualBar/Call/Jitsi/CallJitsiWithData.js
```diff
@@ -155,41 +155,31 @@ const CallJitsiWithData = ({ rid }) => {
...
-				}
-			} catch (error) {
-				dispatchToastMessage({ type: 'error', message: t(error.reason) });
-				clear();
-				handleClose();
+		const clear = () => {
+			jitsi.off('HEARTBEAT', testAndHandleTimeout);
+			jitsi.dispose();
+		};
+		const none = () => {};
```
matheusbsilva137, line 166:
```suggestion
```
If we decide to change the return statement, then we could also erase this line.

matheusbsilva137 (changes requested):

client/views/room/contextualBar/Call/Jitsi/CallJitsiWithData.js
yash-rajpal: Done.

client/views/room/contextualBar/Call/Jitsi/CallJitsiWithData.js
yash-rajpal: I Agree. But I think its a good practice to always return a cleanup function and handle all checks inside that function.

matheusbsilva137 (approved):

matheusbsilva137:
Merged! Thanks @Aman-Maheshwari for proposing the changes and @yash-rajpal for working on some improvements 🤗

camherry:
Hi,
The bug still seems to be present, I'm currently on version 4.3.2 rocket-chat docker image, and when a second participant wants to join me in video jitsi, he gets message "call already ended", so he has to click on the phone icon to join me, what generates a new button in the conversation.
Thanks for your help.

bseclier:
Hello, 
I can confirm that it still doesn't work in 4.3.1 nor in 4.3.2...
This bug has been fixed in the past, but it is back, it doesn't give a good image to the solution in my university :(

bseclier:
Still here in 4.4.1 :(

matheusbsilva137:
Hey @bseclier and @camherry . Thanks for pointing this is still an issue for you
However, I couldn't reproduce it in 4.4.1. Could you please provide more info about your environment?
Are you experiencing this issue while following the same steps as in the "Before" video in this PR's description?

bseclier:
Hello @matheusbsilva137 
I've made more tests, I appears that the problem occurs only with the desktop client. I've tried with both our test and production server instances (docker 4.4.1 for test and docker 4.3.1 prod).
I want the jitsi to open in a new window. In firefox and chromium, it opens a new tab and everything works well.
In the desktop client, it opens a popup and the problem become the same as the "Before" video.

My desktop client is the snap package with 3.7.7 version (it was the same behavior with 3.7.X). I hope it'll help.
Cheers,

himpierre:
Same experience here. It worked for a while. Now with the desktop client 3.7.7 the "call already ended" problem is back. Server version 4.4.1. What puzzles me is with desktop client 3.7.7 and the setting "always open in a new window" the desktop client now does not open a tab in the default web browser anymore (firefox for example) but spawns a popup from the desktop client.

jeanfbrito:
> Same experience here. It worked for a while. Now with the desktop client 3.7.7 the "call already ended" problem is back. Server version 4.4.1. What puzzles me is with desktop client 3.7.7 and the setting "always open in a new window" the desktop client now does not open a tab in the default web browser anymore (firefox for example) but spawns a popup from the desktop client.

The description says that will use the browser if this is not enabled. Then you need to disable to use browser for video chat.
"When set Video Chat will be opened using an application's window, otherwise the default browser will be used."

image

himpierre:
@jeanfbrito Interesting! Thanks!. When disabling this setting (did not know it existed) the video chat opens in firefox again *and* guess what, the "call already ended" problem went away. Maybe with this information you are able to fix the desktop client Jean?

bseclier:
Hello @jeanfbrito 
This is a workaroud, indeed, but it doesn't solve the initial problem :)

jeanfbrito:
@himpierre @bseclier 
We are investigating this issue on Desktop App. If it happens outside the desktop app, please let us know.

bseclier:
Thank you @jeanfbrito 
As I said above, I have no problem with firefox nor chromium, only in the desktop app.

PR #23957, opened 2021-12-15
[FIX] Error when syncing user to a room with special characters in the name

pierre-lehnen-rc:


ee/server/lib/ldap/Manager.ts
```diff
@@ -244,9 +244,9 @@ export class LDAPEEManager extends LDAPManager {
 	private static createRoomForSync(channel: string): IRoom | undefined {
 		logger.debug(`Channel '${ channel }' doesn't exist, creating it.`);
 
-		const roomOwner = settings.get('LDAP_Sync_User_Data_Channels_Admin') || '';
+		const roomOwner = settings.get<string>('LDAP_Sync_User_Data_Channels_Admin') || '';
 		// #ToDo: Remove typecastings when createRoom is converted to ts.
```
pierre-lehnen-rc, line 248:
```suggestion
```

server/lib/findRoomByNonValidatedName.ts
```diff
@@ -0,0 +1,22 @@
...
+	let channelName = s.trim(name);
+	try {
+		// TODO evaluate if this function call should be here
+		channelName = getValidRoomName(channelName, undefined, { allowDuplicates: true });
+	} catch (e) {
+		console.error(e);
+	}
+
+	return Rooms.findOneByName(channelName);
+};
```
ggazzo, line 22: I dont know why you removed the logic from the model, but I would change this function to async, and use the raw model instead. and I know its not our default, but I would use the raw collection as the first parameter injecting the model and removing the import (or I would keep this one as a method of Room raw collection)

PR #23975, opened 2021-12-16
Handle translation of the buttons names

d-gubert:

The names registered by the buttons weren't being translated

### DIFF
--- a/app/ui-message/client/actionButtons/messageAction.ts
+++ b/app/ui-message/client/actionButtons/messageAction.ts
@@ -11,8 +11,8 @@ const getIdForActionButton = ({ appId, actionId }: IUIActionButton): string => `
 // eslint-disable-next-line no-void
 export const onAdded = (button: IUIActionButton): void => void MessageAction.addButton({
 	id: getIdForActionButton(button),
-	icon: button.icon || '',
-	label: t(Utilities.getI18nKeyForApp(button.nameI18n, button.appId)),
+	// icon: button.icon || '',
+	label: t(Utilities.getI18nKeyForApp(button.labelI18n, button.appId)),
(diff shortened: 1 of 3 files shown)


app/ui-message/client/actionButtons/messageAction.ts
```diff
@@ -1,6 +1,8 @@
 import { IUIActionButton } from '@rocket.chat/apps-engine/definition/ui';
 
+import { Utilities } from '../../../apps/lib/misc/Utilities';
 import { MessageAction, messageArgs } from '../../../ui-utils/client';
+import { t } from '../../../utils/client';
```
AllanPazRibeiro, line 5: It would be more "readable" if you set a alias to this import as:

`{ t as translator}`

app/ui-message/client/actionButtons/messageBox.ts
```diff
@@ -5,27 +5,32 @@ import { Rooms } from '../../../models/client';
 import { messageBox } from '../../../ui-utils/client';
 import { applyButtonFilters } from './lib/applyButtonFilters';
 import { triggerActionButtonAction } from '../ActionManager';
+import { t } from '../../../utils/client';
```
AllanPazRibeiro, line 8: here too

AllanPazRibeiro (approved):

app/ui-message/client/actionButtons/messageAction.ts
d-gubert: I agree with readability, but it is used as `t` everywhere else in the code base. So I guess consistency wins in this one? 😛

AllanPazRibeiro: surething

PR #23976, opened 2021-12-16
[FIX] Owner not being able to leave team when he is the only participant

### LINKED ISSUE #23956: Leaving Team with one member gives error to make someone else admin
### Description:
When a user click on leave a team in which there is only one member. he is prompted to make someone else admin before leaving instead of prompting to whether delete the team or not.

### Steps to reproduce:

1. Click on Create New -> Team
2. Then Click on three dots on team name and select leave.

### Expected behavior:
Should either delete the team or prompt the user to confirm whether to delete and leave the team or not.

Screenshot 2021-12-16 at 12 40 34 AM
Screenshot 2021-12-16 at 12 40 42 AM
Screenshot 2021-12-16 at 12 40 47 AM

### Client Setup Information

- Desktop App or Browser Version: Brave
- Operating System: MacOS v12.0.01

aswiniip:
## Issue(s)

Fixes #23956 

## Steps to test or reproduce

Mentioned in the issue


aswinidev:
This is an initial commit i need to make some changes like the sidebar menu option is still having the same issue but the leave option from the team info menu is fixed now.

gabriellsh:
This looks super weird TBH. You click `Leave` and endup in a modal that'll delete the channel, even though there's no clear indicative of that (e.g. the title is very misleading). Feels like bad UX.

It is way better than just locking the user, but I think we can do better. Maybe we could have a warning modal with an action button to open the delete modal. WDYT?

apps/meteor/client/views/teams/contextualBar/info/TeamsInfoWithData.js
```diff
@@ -66,10 +67,15 @@ const TeamsInfoWithLogic = ({ room, openEditing }) => {
 	const canDelete = usePermission('delete-team', room._id);
 	const canEdit = usePermission('edit-team-channel', room._id);
 
+	const { value } = useEndpointData(
```
gabriellsh, line 70: Use `useQuery` instead.

gabriellsh (changes requested):

dougfabris:
> This looks super weird TBH. You click `Leave` and endup in a modal that'll delete the channel, even though there's no clear indicative of that (e.g. the title is very misleading). Feels like bad UX.
> 
> It is way better than just locking the user, but I think we can do better. Maybe we could have a warning modal with an action button to open the delete modal. WDYT?

@gabriellsh I'm planning to do the last option.

gabriellsh:
Closing for the same reason as in https://github.com/RocketChat/Rocket.Chat/issues/23956#issuecomment-5319323558

PR #23981, opened 2021-12-17
[FIX] Enter not working on modal's multi-line input

murtaza98:

Right now, if we try to press enter for a new line on multi-line modal input... it auto triggers the submit event. This PR fixes this behaviour by not submitting the modal in case the enter was pressed within an input text with multiline expected

## Issue(s)

**Before:**
https://user-images.githubusercontent.com/34130764/146526766-d927326d-26d8-4ccb-85eb-8afa573b97c8.mp4

**After:**
https://user-images.githubusercontent.com/34130764/146526743-d7b063f3-c9f6-4e4f-a39a-480bd1b4d3dc.mp4


client/views/blocks/ModalBlock.js
```diff
@@ -59,6 +59,12 @@ function ModalBlock({ view, errors, appId, onSubmit, onClose, onCancel }) {
 		(event) => {
 			if (event.keyCode === 13) {
 				// ENTER
+				if (
+					event.srcElement &&
+					(event.srcElement.nodeName === 'TEXTAREA' || event.srcElement.localName === 'textarea')
```
tiagoevanp, line 64: `srcElement` have some compatibility problems with firefox, how about change to `currentTarget`?

client/views/blocks/ModalBlock.js
```diff
@@ -59,6 +59,12 @@ function ModalBlock({ view, errors, appId, onSubmit, onClose, onCancel }) {
...
+				) {
+					return;
+				}
 				return onSubmit(event);
```
tiagoevanp, line 68: In my point of view, is not necessary return something, how about if/else block only?

tiagoevanp (changes requested):

client/views/blocks/ModalBlock.js
murtaza98: Yeah makes sense. I've modified it in a way to avoid blank return

client/views/blocks/ModalBlock.js
murtaza98: I couldn't find `currentTarget` prop attached to this event. The closest I found to this was `target` prop, so I've it to use that instead. Let me know if that should work or not? Thanks!!

tiagoevanp: perfect!

tiagoevanp (approved):

Issue #18264, opened 2020-07-15
Additional alias names of custom emojis not accepted anymore


frdmn:
### Description:

[According to our docs](https://docs.rocket.chat/guides/administrator-guides/custom-emoji), it should be possible to add multiple alias names to custom emojis, separated by space character: 

> Enter a name for your emoji, add space separated list of the aliases for the emoji which can be entered to display the emoji in the chats

This does not seem to be working anymore, the space character will be replaced by a dash instead.

### Steps to reproduce:

1. Add a new custom emoji
2. Give it multiple aliases, e.g. `first second`
3. Hit save, see how the alias gets transformed into: `first_second` 

### Expected behavior:

It should separate the alias input by space, as stated in the docs, and accept multiple alias names.

### Actual behavior:

Space character gets replaced by dash. Multiple aliases not possible.

### Server Setup Information:

- Version of Rocket.Chat Server: 3.4.0
- Operating System: Linux
- Deployment Method: Docker
- Number of Running Instances: Docker 
- DB Replicaset Oplog: Yes
- NodeJS Version: Docker
- MongoDB Version: Docker

##### Demo video

[<img width="796" alt="Bildschirmfoto 2020-07-15 um 10 45 26" src="https://user-images.githubusercontent.com/944459/87524368-6163be00-c688-11ea-964d-873f1d86ffd9.png">](https://up.frd.mn/0zWo1s8hFd.mp4)

LeeThompson:
Seeing this too.

LeeThompson:
This *appears* to be fixed as of 3.7.0.

LeeThompson:
(This stat:stale thing because it hasn't been addressed or added to and then automatically closed is why these bugs stay.)

RavuAlHemio:
I'm trying to resolve this in #23990; may I request a review?

paulchen:
Today I investigated into several problems regarding custom emojis (cf. https://github.com/RocketChat/Rocket.Chat/issues/26979, https://github.com/RocketChat/Rocket.Chat/issues/26980).

In that process, I stumbled upon the problem described in this issue as well which still persists in Rocket.Chat 5.1.4.

codewithmitesh:
Hey, @dudanogueira I want to work on the Issue, Can you assign it to me?

RavuAlHemio:
I‘ve already proposed a fix in #23990; unfortunately, it appears the dev team has not yet come around to reviewing it…

[...]

PR #23990, opened 2021-12-20
[FIX] Slugify emoji aliases after splitting them

RavuAlHemio:
This allows to set multiple emoji aliases again. Otherwise the comma or
space would have been replaced by an underscore before the splitting
occurs.

Split multiple emoji aliases on spaces or commas before slugifying them

## Issue(s)

#18264

## Steps to test or reproduce

Attempt to add an emoji with aliases specified as `test1 test2` or `test1,test2`. The aliases are currently converted into a single alias `test1_test2`; after this fix, they are correctly split up into the aliases `test1` and `test2`.


RavuAlHemio:
For completeness's sake: the issue appears to have been solved as part of #32991.

PR #24000, opened 2021-12-21
Chore: Prettier for us all

sampaiodiego:

I've moved the file `.prettierrc` that was previously only being applied to `/client` folder.

In order to make prettier and eslint to run smoothly, I had to remove few eslint rules applied to TS files.

When reviewing this PR I recommend looking the individual commits. The majority of the changes were done via `eslint --fix`.

Why having Prettier and ESLint together https://prettier.io/docs/en/comparison.html

## Issue(s)

Closes #23792


d-gubert (approved):

KevLehman (approved):
image

PR #24018, opened 2021-12-23
Regression: Remove self from fallback departments dropdown

KevLehman:

### DIFF
--- a/client/components/RoomIcon/OmnichannelRoomIcon/OmnichannelCoreSourceRoomIcon.tsx
+++ b/client/components/RoomIcon/OmnichannelRoomIcon/OmnichannelCoreSourceRoomIcon.tsx
@@ -27,5 +27,5 @@ export const OmnichannelCoreSourceRoomIcon = ({
 	size: ComponentProps<typeof Icon>['size'];
 }): ReactElement => {
 	const icon = iconMap[room.source?.type || 'other'] || 'headset';
-	return <Icon name={icon} size={size} color={colors[room.v.status || 'offline']} />;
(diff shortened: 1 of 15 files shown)


definition/IRoom.ts
```diff
@@ -107,7 +107,7 @@ export enum OmnichannelSourceType {
 	OTHER = 'other', // catch-all source type
 }
 
-export interface IOmnichannelRoom extends Omit<IRoom, 'default' | 'featured' | 'broadcast' | 'featured' | ''> {
+export interface IOmnichannelRoom extends Omit<IRoom, 'default' | 'featured' | 'broadcast' | ''> {
```
KevLehman, line 110: This is an unrelated change to remove duplicated 'featured' from Omit.

client/components/RoomIcon/OmnichannelRoomIcon/OmnichannelCoreSourceRoomIcon.tsx
```diff
@@ -27,5 +27,5 @@ export const OmnichannelCoreSourceRoomIcon = ({
 	size: ComponentProps<typeof Icon>['size'];
 }): ReactElement => {
 	const icon = iconMap[room.source?.type || 'other'] || 'headset';
-	return <Icon name={icon} size={size} color={colors[room.v.status || 'offline']} />;
+	return <Icon name={icon} size={size} color={colors[room.v?.status || 'offline']} />;
```
KevLehman, line 30: Again, unrelated change, just for safety in the case a room doesn't have a `v` prop. This doesn't happen very often, but if it happens the UI will be unusable.

tiagoevanp (approved):

PR #24022, opened 2021-12-23
Chore: Update copyright notices

tassoevan:

Update date range in copyright notices to `2015-2022`.

## Further comments

> Though not required by law, there's no good reason to omit these notices either since they serve as a clear record of authorship. In particular, they serve as a reminder that the work is under copyright protection and therefore deter copyright infringement.
> https://opensource.stackexchange.com/a/6393


sampaiodiego:
we're 3 days away from 2022.. should we wait and update to 2022 instead?

sampaiodiego (approved):

Issue #23896, opened 2021-12-08
[bug] Message Erasure Type `Keep Messages And Username` not work as expected


ankar84:
### Description:

In **Admin UI - Message** we have `Message Erasure Type` option.
Default is `Delete All Messages`
But if you set `Keep Messages And Username` - you get same behavior and  `Delete All Messages` set - all users messages are deleted.

### Steps to reproduce:

1. Set `Keep Messages And Username` value in **Admin UI - Message**
2. Delete user that had messages in DM and other chats
3. No messages of deleted user remain

### Expected behavior:

User account should be deleted, but messages in chat history should remain and be authored as deleted user (when that user was on server, for example UserA @usera)

### Actual behavior:

![image](https://user-images.githubusercontent.com/4023037/145171435-87d30ef2-e745-41e2-883a-c044f2ee3c6d.png)

### Server Setup Information:

- Version of Rocket.Chat Server: 4.2.0
- Operating System: CentOS7
- Deployment Method: docker
- Number of Running Instances: 10
- DB Replicaset Oplog: Enabled
- NodeJS Version: 12.22.1
- MongoDB Version: 4.2.17

### Client Setup Information

- Desktop App or Browser Version: Chrome 96
- Operating System: Windows 10

### Additional context

Third option look like work fine
![image](https://user-images.githubusercontent.com/4023037/145171998-389166f9-5f49-4b93-a2e1-a249ad2b9f56.png)

arshxyz:
Hi, I was able to reproduce this issue.

Looking further into it. It seems like the messages are NOT deleted on the server side. Hence, a reload of the channel will show the "mistakenly deleted" messages.

The messages are removed from the frontend due to a websocket message [here](https://github.com/RocketChat/Rocket.Chat/blob/028ede9c804ae3aace4c0d917b3ca7ce9e040875/server/modules/listeners/listeners.module.ts?L62#L62-L66).

Every deletion regardless of `messageErasureType` emits the `user.deleted` websocket message which results in removal of the  messages from the frontend if the page is open.

arshxyz:
I'd love to work on this issue, please assign it to me. However I don't know which way would be better - 
1. Skip broadcasting `user.deleted` for `messageErasureType` of `Keep`?
2. Create a new Event `user.keepdeleted` for this so that it's easier to bind an action to that broadcast in future?
[...]

PR #24024, opened 2021-12-25
[FIX] Message Erasure Type "Keep" Messages not working

arshxyz:
## Proposed changes
Check thread on https://github.com/RocketChat/Rocket.Chat/issues/23896

## Issue(s)
https://github.com/RocketChat/Rocket.Chat/issues/23896


murtaza98 (changes requested):
Hey @arshxyz I feel the first approach you mentioned [here](https://github.com/RocketChat/Rocket.Chat/issues/23896#issuecomment-988779797) looks more elegant since there is no need to send a broadcast in case the event is not valid. I'd recommend using that solution instead. Thanks!!

arshxyz:
I've made the changes accordingly, thanks

server/modules/listeners/listeners.module.ts
```diff
@@ -59,10 +59,14 @@ export class ListenersModule {
 			});
 		});
 
-		service.onEvent('user.deleted', ({ _id: userId }) => {
-			notifications.notifyLoggedInThisInstance('Users:Deleted', {
-				userId,
-			});
+		service.onEvent('user.deleted', ({ _id: userId }, keepMessages) => {
```
murtaza98, line 62: Do we still need this change?

server/sdk/lib/Events.ts
```diff
@@ -36,7 +36,7 @@ export type EventSignatures = {
 	'stream'([streamer, eventName, payload]: [string, string, string]): void;
 	'subscription'(data: { action: string; subscription: Partial<ISubscription> }): void;
 	'user.avatarUpdate'(user: Partial<IUser>): void;
-	'user.deleted'(user: Partial<IUser>): void;
+	'user.deleted'(user: Partial<IUser>, keepMessages: boolean): void;
```
murtaza98, line 39: Same goes here

arshxyz: nope, both changes have been removed

arshxyz: should be fine now

murtaza98 (approved):

PR #24026, opened 2021-12-25
[FIX] this.input can be undefined causing a crash

grahhnt:

Add a null check to prevent crashing


grahhnt:
Someone beat me to it: #24033

PR #24035, opened 2021-12-27
Regression: Ensure room action buttons only appear inside menu

d-gubert:

Currently, action buttons registered by apps to appear in the ROOM_ACTION context show in the first position of the list, but since they don't have an icon they are effectively invisible in the tab bar.

Here we change the order configuration of the button so we make sure it only shows inside the room menu


AllanPazRibeiro (approved):
LGTM

PR #24037, opened 2021-12-28
[FIX] Inconsistent validation of user's access to rooms

pierre-lehnen-rc:
The `canAccessRoom` function was being called with insufficient data in several places, resulting in some false negatives.


app/api/server/v1/chat.js
```diff
@@ -494,7 +496,9 @@ API.v1.addRoute('chat.syncThreadsList', { authRequired: true }, {
 			updatedSinceDate = new Date(updatedSince);
 		}
 		const user = Users.findOneById(this.userId, { fields: { _id: 1 } });
-		const room = Rooms.findOneById(rid, { fields: { t: 1, _id: 1 } });
+		const roomFields = Object.assign({ t: 1, _id: 1 }, roomAccessAttributes);
```
ggazzo, line 499:
```suggestion
		const roomFields = { ...roomAccessAttributes, t: 1, _id: 1 };
```

ostjen (approved):

PR #24039, opened 2021-12-28
[FIX] Omnichannel Current chats pagination not working

murtaza98:
## Issue(s)

Omnichannel - Current Chats not refreshing while clicking on pagination
## Steps to test or reproduce

Steps to reproduce:
1. Go to Omnichannel => Current Chats
2. Add some filter or just Clear all Filters to see all the chats
3. Click to navigate between the pages

Expected Behavior:
After navigating between the pages it is supposed to refresh and display different chats

Actual Behavior:
Even clicking on any page option it doesn't refresh and the chats remain the same of the first page


KevLehman:
@murtaza98 , is this a fix or a regression? Cause it was working before candidate 🙈

KevLehman:
This is behaving a bit weird 🤔 
1. Filter by status is not working: current chats page is only showing closed chats. If you select "Open" from dropdown, it will rerender to show only closed ones.
2. Filter by department is not working

Can you take a look at those? 👀

yash-rajpal:
Yes, I think Filter by fields is not be working as of now.

#23970 Fixes this issue. Please apply this fix and check if filter works. 👀

KevLehman:
Hey, I integrated the changes from the other PR into this one 👀 and fixed the Omni-specific issue.

Here, the issue is that both, GenericTable & FilterByText were trying to filter by params, and they were not liking each other. Moved the filters to GenericTable again should also fix it the issue on the other view 🤗

murtaza98:
New changes look good to me :+1: Just added a small change to remove the unused method. Thanks, guys!!

Harmeet221 (approved):

amolghode1981 (approved):

PR #24044, opened 2021-12-28
[IMPROVE] Rewrite roomNotFound to React Component

dougfabris:
### before
image
image

### after
image
image


ggazzo:
check the variation='danger'

app/ui-utils/client/lib/openRoom.js
```diff
@@ -91,9 +91,11 @@ export const openRoom = async function (type, name, render = true) {
 		} catch (error) {
 			c.stop();
 			if (type === 'd') {
-				const result = await callWithErrorHandling('createDirectMessage', ...name.split(', '));
-				if (result) {
+				try {
+					const result = await callWithErrorHandling('createDirectMessage', ...name.split(', '));
```
tassoevan, line 95: `callWithErrorHandling` is redundant here as you still have to catch the exception. Better to use `call`:

```suggestion
					const { rid } = await call('createDirectMessage', ...name.split(', '));
```

client/views/room/Room/RoomNotFound/RoomNotFound.tsx
```diff
@@ -0,0 +1,48 @@
...
+type RoomNotFoundSession = {
+	type: IRoom['t'];
+	name: IRoom['name'];
+	error: Meteor.Error;
+};
+
+const RoomNotFound = (): ReactElement => {
+	const { isMobile } = useLayout();
+	const t = useTranslation();
+	const { name, type } = useSession('roomNotFound') as RoomNotFoundSession;
```
tassoevan, line 22: As `error` is ignored at `RoomNotFound`, I wonder if we can pass `name` and `type` as props, using the template's data context. @ggazzo thoughts?

tassoevan (dismissed):

client/views/room/Room/RoomNotFound/RoomNotFound.tsx
dougfabris: As we changed the approach here with the design team, we won't use the room data anymore

gabriellsh (approved):

PR #24048, opened 2021-12-29
[NEW] SlashCommand to create discussions

debdutdeb:
## Further comments

- `/discuss #slashcommand Merge this PR` will create a new discussion named `Merge this PR` in `#slashcommand` room.
- `/discuss I hope there is no problem in this PR` will create a new discussion named `I hope ...` in the room where the command is executed, if the room is either a public channel or private group.
- `/discuss` in any thread will create a discussion with that thread as it's parent, the thread message as title, and the thread followers as members.
- `/discussion I like this` in any thread will create a discussion with that thread as it's parent, `I like this` as title, and the thread followers as members.


ggazzo:
did you try create this functionality by some app?

debdutdeb:
Hey @ggazzo , yes, tried with an app, but thought, it'd be better to make it part of core, along with `create`. Is there a problem 🤔 ?

debdutdeb:
Moved to an app :)

PR #24053, opened 2021-12-29
[IMPROVE] Throw 404 error in invalid endpoints

matheusbsilva137:

- Throw 404 error when trying to call invalid endpoints.

## Steps to test or reproduce

**_Expected output to an invalid endpoint:_**
```
{
  "statusCode": 404,
  "body": {
    "success": false,
    "error": "Resource not found"
  }
}
```


app/api/server/api.js
```diff
@@ -779,6 +780,21 @@ const createApi = function _createApi(_api, options = {}) {
 	return _api;
 };
 
+WebApp.connectHandlers.use('/api/v1', function(req, res) {
+	res.writeHead(404, {
+		'Content-Type': 'application/json',
+		'Access-Control-Allow-Origin': '*',
```
sampaiodiego, line 786: I don't think this header is needed 🤔

sampaiodiego (changes requested):
You'll need to fix new eslint/prettier rules 😬

app/api/server/api.js
matheusbsilva137: Yeah, it isn't. Removed 🙂

pierre-lehnen-rc (dismissed):

pierre-lehnen-rc (approved):

PR #24067, opened 2021-12-30
[FIX] Omnichannel enabled setting not working when creating rooms

murtaza98:
## Steps to test or reproduce

Turn off Omnichannel from Admin -> Omnichannel
image

The start an omnichannel conversation from anywhere like LiveChat, API, app, etc. The simplest way would be to go to `{serverUrl}/livechat` and send a message.

**Before:** The conversation will be started

**Expected Behaviour:** An error should be thrown stating that omnichannel is disabled

### DIFF
--- a/app/livechat/server/lib/Livechat.js
+++ b/app/livechat/server/lib/Livechat.js
@@ -49,6 +49,7 @@ const dnsResolveMx = Meteor.wrapAsync(dns.resolveMx);
 export const Livechat = {
 	Analytics,
 	historyMonitorType: 'url',
+	omnichannelEnabled: false,
 
 	logger,
 	webhookLogger: logger.section('Webhook'),
@@ -156,6 +157,9 @@ export const Livechat = {
 	},
 
 	async getRoom(guest, message, roomInfo, agent, extraData) {
+		if (!this.omnichannelEnabled) {
+			throw new Meteor.Error('error-omnichannel-is-disabled');
+		}
 		Livechat.logger.debug(`Attempting to find or create a room for visitor ${guest._id}`);
 		let room = LivechatRooms.findOneById(message.rid);
 		let newRoom = false;
@@ -1406,4 +1410,8 @@ export const Livechat = {
 
 settings.watch('Livechat_history_monitor_type', (value) => {
 	Livechat.historyMonitorType = value;
+});
+
+settings.watch('Livechat_enabled', (value) => {
+	Livechat.omnichannelEnabled = value;
 });
--- a/app/livechat/server/lib/Livechat.js
+++ b/app/livechat/server/lib/Livechat.js
@@ -49,7 +49,6 @@ const dnsResolveMx = Meteor.wrapAsync(dns.resolveMx);
 export const Livechat = {
 	Analytics,
 	historyMonitorType: 'url',
-	omnichannelEnabled: false,
 
 	logger,
 	webhookLogger: logger.section('Webhook'),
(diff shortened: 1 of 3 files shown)


packages/rocketchat-i18n/i18n/en.i18n.json
```diff
@@ -4771,4 +4772,4 @@
   "Your_temporary_password_is_password": "Your temporary password is <strong>[password]</strong>.",
   "Your_TOTP_has_been_reset": "Your Two Factor TOTP has been reset.",
   "Your_workspace_is_ready": "Your workspace is ready to use 🎉"
-}
\ No newline at end of file
+}
```
murtaza98, line 4775:
```suggestion
}```

File: app/livechat/server/lib/Livechat.js [hunk 2]
```diff
@@ -156,6 +157,9 @@ export const Livechat = {
 	},
 
 	async getRoom(guest, message, roomInfo, agent, extraData) {
+		if (!this.omnichannelEnabled) {
```
sampaiodiego, line 160:
```suggestion
		if (!settings.get('Livechat_enabled')) {
```

app/livechat/server/lib/Livechat.js
```diff
@@ -1407,3 +1411,7 @@ export const Livechat = {
 settings.watch('Livechat_history_monitor_type', (value) => {
 	Livechat.historyMonitorType = value;
 });
+
+settings.watch('Livechat_enabled', (value) => {
```
sampaiodiego, line 1415: I see this is a pattern we've been using (just reading the above lines), but there is no value "watching" a setting to the value to somewhere else. using `settings.get(settingName)` is meant for that purpose.

we need to start using `watch` only when we need to trigger additional tasks, or we need to watch for multiple settings to compose a single value. ;)

packages/rocketchat-i18n/i18n/en.i18n.json
```diff
@@ -1751,6 +1751,7 @@
   "error-not-allowed": "Not allowed",
   "error-not-authorized": "Not authorized",
   "error-office-hours-are-closed": "The office hours are closed.",
+  "error-omnichannel-is-disabled": "Error! Omnichannel is disabled",
```
sampaiodiego, line 1754: is this going to be used somewhere?

sampaiodiego (changes requested):

app/livechat/server/lib/Livechat.js
murtaza98: I went with this approach since I was hoping to cache this setting's value in the memory so we don't have to hit the DB to get its value, every time a room gets created. Let me know if you still think it would be better to use settings.get() here instead of settings.watch()

app/livechat/server/lib/Livechat.js
murtaza98: This same comment applies here as well -> https://github.com/RocketChat/Rocket.Chat/pull/24067/files#r776976552

packages/rocketchat-i18n/i18n/en.i18n.json
murtaza98: I realised this isn't getting used anywhere. So reverting it now

app/livechat/server/lib/Livechat.js
sampaiodiego: I thought so, but the point is that `settings.get()` is already cached, it doesn't hit the DB, that's why I'm suggesting to use it =)

if you look the `settings` variable is an instance of `CachedSettings` class.. 😉  but we should also add a comment on the `settings.get` method, so it is clear to everyone that the value is cached 👍

app/livechat/server/lib/Livechat.js
murtaza98: This has been applied

app/livechat/server/lib/Livechat.js
murtaza98: Thanks, I wasn't aware of that. I've updated the code to use `settings.get()` now. 

PS: Also added a comment on the cached setting method so it is more clear that the value is cached

sampaiodiego (approved):

Issue #23997, opened 2021-12-20
Error when importing users from csv file: only first user imported with incomplete information, not working with TOTP


jahohu:
When using the import function in the admin panel -> CSV file -> Users in a csv file (in a zip file), only the first user / line gets imported and displays "error" after the import. The user is created, but not automatically joined to default channels, the display name is not imported although it was present in the CSV. Users can then be edited manually in order to set all fields correctly. An error is also displayed when the CSV only contains one line.

Additionally, when using TOTP with the admin account that uploads the file, a message like TOTP not available (sorry, no screenshot taken) shows up and results in the same situation with only the first line imported. The same message appears when trying to edit a user with a TOTP-enabled admin user.

**Versions in user**
Operating system/version: Managed virtual server, Debian 10
Rocket.Chat server version: approx. 4.0, beginning of November, still persistent with 4.2.1
Rocket.Chat client version: same results in Rocket Chat Electron desktop 3.5.7, 3.6.0 on Linux Mint 19 / Ubuntu 18 LTS as well as Windows 10 Pro 64-bit (virtual machine), and in the browser client the last versions of Firefox on same OS.
Is there any setting relevant changed? No, only the somehow related fix #23734 has been applied and is working

**Steps to reproduce:**
- Create a csv file users.csv and zip it as described in the [guide](https://docs.rocket.chat/guides/administration/admin-panel/import/csv), containing one or more lines with user information
- From any client (browser / desktop), with an admin user, navigate to the admin panel: Import -> Import New File -> Import Type: CSV -> Select the file -> Click Import. This lists the correct and complete content of the csv file. When clicking upload, the described error message shows up and only one user, the first line, is imported.

**Expected behavior**
All users/lines from the csv file should be imported with all information arriving in the user accounts (currently, display name is missing) without error messages, and also while using a TOTP-enabled admin user. All this worked before, in server versions prior to 4.0, with any client for the admin panel.

imanshafiei540:
I have the same issue with the web version.
RocketChat Version: 4.2.
[users.csv](https://github.com/RocketChat/Rocket.Chat/files/7771064/users.csv)
1

fatemeh-mgsdi:
I'm having the same issue with Rocketchat version 4.2.1. Any updates?

jahohu:
**Update:**
[...]

PR #24073, opened 2021-12-31
[FIX] Not importing users from csv file

kulkarnigaurav38:
The users.csv file was being read properly, but the importing process throws an error when trying to import users details. Further details are mentioned in the issue.

https://user-images.githubusercontent.com/59637086/147810177-48c5ed58-7f0a-4e58-a0e6-4717471b868a.mp4

## Issue(s)
 #23997

## Further comments

Error which showed up earlier in the logs is solved now
image


app/lib/server/functions/setRealName.js
```diff
@@ -41,7 +41,5 @@ export const _setRealName = function (userId, name, fullUser) {
 };
 
 export const setRealName = RateLimiter.limitFunction(_setRealName, 1, 60000, {
-	0() {
-		return !Meteor.userId() || !hasPermission(Meteor.userId(), 'edit-other-user-info');
-	}, // Administrators have permission to change others names, so don't limit those
+	0() { return !Meteor.userId || !hasPermission(Meteor.userId, 'edit-other-user-info'); }, // Administrators have permission to change others names, so don't limit those
```
tassoevan, line 44: This doesn't look right. `Meteor.userId` is a function, therefore `!Meteor.userId` is always `false`. The first parameter of `hasPermission(userId, permissionId, scope)` is supposed to be the value returned by calling `Meteor.userId()`.

tassoevan (changes requested):

PR #24078, opened 2022-01-01
[IMPROVE] Added a Reset Button in the Account Profile Page

aswinidev:

https://user-images.githubusercontent.com/76481696/147860107-b31ba229-fb86-471b-8060-283ce9a6fa30.mp4

## Further comments

I added a reset button which resets all the changes made while editing the profile. It seemed like a necessary improvement because whenever i had to not save the changes, i had to go to a different page and come back


murtaza98:
Hey, @aswinidev This indeed seems like a good improvement. I wanted to suggest a couple of improvements here
1. Make the reset button theme different from the Save button (a positive action)
2. The reset button should only be visible (or clickable) when the user has changed something which isn't yet saved. Look into the `Cancel` button on the Admin settings panel for inspiration.
image

aswinidev:
Sorry for the late reply. Must have missed my mails :disappointed_relieved: . I made the changes. @murtaza98 please review!

murtaza98 (approved):

PR #24095, opened 2022-01-03
[FIX] Workaround for use of the default settings collection

MartinSchoeler:
This is a workaround for us to swiftly be able to use the default settings collection and system.

This will allow us to use the default settings ui for now, after the MVP we will discuss again a better way to maintain traceability.

### DIFF
--- a/server/services/voip/service.ts
+++ b/server/services/voip/service.ts
@@ -11,6 +11,7 @@ import { Commands } from './connector/asterisk/Commands';
 import { IVoipConnectorResult } from '../../../definition/IVoipConnectorResult';
 import { IQueueMembershipDetails, IRegistrationInfo, isIExtensionDetails } from '../../../definition/IVoipExtension';
 import { IQueueDetails, IQueueSummary } from '../../../definition/ACDQueues';
+import { settings } from '../../../app/settings/server';
 
 export class VoipService extends ServiceClass implements IVoipService {
 	protected name = 'voip';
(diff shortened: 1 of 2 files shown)


ggazzo (changes requested):
use return first

tiagoevanp (approved):

server/services/voip/service.ts
```diff
@@ -84,7 +85,34 @@ export class VoipService extends ServiceClass implements IVoipService {
...
+							password: settings.get('VoIP_Management_Server_Password'),
+						},
+				  }
+				: {
+						configData: {
+							websocketPort: parseInt(settings.get('VoIP_Server_Websocket_Port')),
+							websocketPath: settings.get('VoIP_Server_Websocket_Path'),
+						},
+				  }),
+		} as IVoipServerConfig;
```
sampaiodiego, line 115: looks like the type casting is not needed.
```suggestion
		};
```

server/services/voip/service.ts
```diff
@@ -84,7 +85,34 @@ export class VoipService extends ServiceClass implements IVoipService {
...
 	async getServerConfigData(type: ServerType): Promise<IVoipServerConfig | null> {
-		return this.VoipServerConfiguration.findOne({ type, active: true });
+		// TODO: Decide the approach we will take regarding settings after the MVP,
+		// For now this work around should be enough.
+
+		// const config = this.VoipServerConfiguration.findOne({ type, active: true });
+
+		const management = type === ServerType.MANAGEMENT;
+		return {
+			_id: '',
```
sampaiodiego, line 95: looks like `IVoipServerConfig` doesn't need extend `IRocketChatRecord` anymore.. so you don't need to provide `_id: ''` (which doesn't make sense anyways)

server/services/voip/service.ts
```diff
@@ -84,7 +85,34 @@ export class VoipService extends ServiceClass implements IVoipService {
...
+			type,
+			host: settings.get(management ? 'VoIP_Management_Server_Host' : 'VoIP_Server_Host'),
+			name: settings.get(management ? 'VoIP_Management_Server_Name' : 'VoIP_Server_Name'),
+			active: true,
+			_updatedAt: new Date(),
```
sampaiodiego, line 100: same as before, not extending `IRocketChatRecord` this is not required

sampaiodiego (changes requested):

sampaiodiego (approved):
oh, not sure how my suggestions didn't break other parts of the code.. 🙊  I see `IVoipServerConfig` still being used as a type for a database record, like on this model: 
https://github.com/RocketChat/Rocket.Chat/blob/780dabee9d3d15b2b30876a6d6adadf76f52f628/app/models/server/raw/VoipServerConfiguration.ts#L4

I'll approve this PR anyways as it seems that there still work needed to make this "voip server config" thing to work, and I think it can be done on the target branch.

Issue #24099, opened 2022-01-04
Incoming integrations broken after upgrade 3.18.0 to 4.3.0


TomaszDom:
### Description:

Incoming integrations no longer work after upgrading from 3.18.0 to 4.3.0.

All integrations return the following message from the server:

`{"success":false,"error":"Invalid href value provided"}`

This behaviour is the same with old and newly recreated integrations, when using the sample curl command the behaviour is the same.

### Steps to reproduce:

1. Go to Administration -> Integrations
2. Click New and fill in the form, make sure the bot user is in the room where you want to send messages
3. Enable integration, save it, run the sample `curl` command provided in the form, e.g.:

`curl -X POST -H 'Content-Type: application/json' --data '{"text":"Example message","attachments":[{"title":"Rocket.Chat","title_link":"https://rocket.chat","text":"Rocket.Chat, the best open source chat","image_url":"/images/integration-attachment-example.png","color":"#764FA5"}]}' https://[REDACTED/hooks/[REDACTED]`

### Expected behavior:

Sample message sent correctly to the room.

### Actual behavior:

No message appears in the room, curl command receives reponse:

`{"success":false,"error":"Invalid href value provided"}`

### Server Setup Information:

- Version of Rocket.Chat Server: 4.3.0
- Operating System: Ubuntu Server 
- Deployment Method: tar
- Number of Running Instances: 2
- DB Replicaset Oplog: yes
- NodeJS Version: 12.22.1
- MongoDB Version: 4.0.27

### Client Setup Information

- Desktop App or Browser Version: all
- Operating System: all

### Relevant logs:
```

{"level":30,"time":"2022-01-04T06:44:19.114Z","pid":24320,"hostname":"[REDACTED]","name":"Integrations","section":"Incoming WebHook","msg":"Post integration:","name":"test2"} 
{"level":20,"time":"2022-01-04T06:44:19.115Z","pid":24320,"hostname":"[REDACTED]","name":"Integrations","section":"Incoming WebHook","urlParams":{"integrationId":"[REDACTED]","token":"[REDACTED]"},"bodyParams":{"text":"Example message","attachments":[{"title":"Rocket.Chat","title_link":"https://rocket.chat","text":"Rocket.Chat, the best open source chat","image_url":"/images/integration-attachment-example.png","color":"#764FA5"}]}} 
[...]

PR #24106, opened 2022-01-04
[FIX] webhook incoming payload and curl example

dudanogueira:

The current example was relying on images hosted on the server itself (/images/integration-attachment-example.png) which doesn't seem to work anymore. Changing to an image at https://open.rocket.chat/images/logo/logo.svg

## Issue(s)

#24099

## Further comments

This will help people experiment with this incoming web hook successfully.


debdutdeb (dismissed):

client/views/admin/integrations/exampleIncomingData.js
```diff
@@ -10,7 +10,7 @@ export function useExampleData({ additionalFields, url }) {
 					title: 'Rocket.Chat',
 					title_link: 'https://rocket.chat',
 					text: 'Rocket.Chat, the best open source chat',
-					image_url: '/images/integration-attachment-example.png',
+					image_url: 'https://open.rocket.chat/images/logo/logo.svg',
```
sampaiodiego, line 13: I have a few concerns:

- The load is probably negligible but still I don't think Open server should be the one serving those images. A better alternative would be github (any image from https://github.com/RocketChat/Rocket.Chat.Artwork) or another CDN where we have images hosted;
- we're creating an external dependency, which is also not ideal (airgapped environments will also not work).. so instead we could chose another internal image.. the server logo might be best choice, as it will probably never be deleted.

sampaiodiego (changes requested):

geekgonecrazy (changes requested):
I agree.  Resource absolutely cannot be an external image. 

A couple of important things to always keep in mind:

If you ever link somewhere it's gonna need to exist forever and always be available.  We cannot make people upgrade and the version including the reference might be used for many years to come.  And if we ever do either of those.. the same problem exists again.

Best course of action is to use another resource inside of the workspace.  

It's honestly the best for the user and for us both.

sampaiodiego:
@dudanogueira are you willing to do the changes we mentioned?

StefanIGit:
I found it too :( and I can fix the it be adding the servers own domain... like "https://my.rocketchatdomain.tld/images/integration-attachment-example.png"
So can you just add the servers URL in it and in the curl example too?
`curl -X POST -H 'Content-Type: application/json' --data '{"emoji":":warning:","text":"Example message","attachments":[{"title":"Rocket.Chat","title_link":"https://rocket.chat","text":"Rocket.Chat, the best open source chat","image_url":"https://{DOMAIN}/images/integration-attachment-example.png","color":"#764FA5"}]}' https://{DOMAIN}/hooks/{TOKEN}`

sampaiodiego:
closing it meanwhile

dudanogueira:
Sorry, missed this one :(
This issue was fixed here :) https://github.com/RocketChat/Rocket.Chat/pull/27290

PR #24111, opened 2022-01-05
Chore: Fix Houston `getNodeNpmVersions` regex to correctly get Node and Npm complete versions

debdutdeb:


sampaiodiego (approved):
good catch :D

sampaiodiego:
I've updated the PR title to `Chore` since it's not a product fix 😉

PR #24114, opened 2022-01-06
[NEW] E2E password generator

ostjen:
Created a new password generator for e2e private keys using readable words


tassoevan:
`generateMnemonicPhrase` may be a good addition to our `@rocket.chat/string-helpers` package (currently on Fuselage monorepo), but I'm wonder if rely on `Math.random()` is fine. The previous solution was based on `meteor/random`, which differs in the pseudorandom generator used.

ostjen:
I can switch to meteor/random, but I don't think it'd be good too still rely on meteor ... and @sampaiodiego do you think the `generateMnemonicPhrase` should be on `@rocket.chat/string-helpers`?

app/e2e/client/helper.js
```diff
@@ -122,6 +124,19 @@ export async function readFileAsArrayBuffer(file) {
...
+	const result = new Array(n);
+	let len = words.length;
+	const taken = new Array(len);
+
+	while (n--) {
+		const x = Math.floor(Math.random() * len);
+		result[n] = words[x in taken ? taken[x] : x];
+		taken[x] = --len in taken ? taken[len] : len;
+	}
+	return result.join(' ');
```
sampaiodiego, line 137: not good having spaces on the password, right? 😬 
```suggestion
	return result.join('-');
```
this char could be received as a param as well =)

app/e2e/client/wordList.ts
```diff
@@ -0,0 +1,1635 @@
+export const words = [
```
sampaiodiego, line 1: what if storing this as a JSON and parsing it when needed? I mean, we're keeping this array in memory even if we don't use it.

sampaiodiego (changes requested):

sampaiodiego:
I'm not particularly worried about `Math.random()` .. if the issue is that it is using repeated words too often we can maybe change it to make sure it always use different words? or maybe adding a few more `Math.random()` calls would be enough

app/e2e/client/helper.js
ostjen: The space was actually discussed with the security team, we did it inspired by the metamask's secret recovery phrase... but I agree the separator can be a parameter

ostjen:
> I'm not particularly worried about `Math.random()` .. if the issue is that it is using repeated words too often we can maybe change it to make sure it always use different words? or maybe adding a few more `Math.random()` calls would be enough

I don't see `Math.random()` as a problem too, I think what @tassoevan was worried was just that we used meteor pseudorandom algo in the previous solution

tassoevan:
@sampaiodiego @ostjen The problem with `Math.random` is [not being cryptographically secure](https://security.stackexchange.com/questions/181580/why-is-math-random-not-designed-to-be-cryptographically-secure), so it's not an option for generating passwords. Unsurprisingly, sometimes Snyk marks `Math.random()` as a vulnerability. This is better: https://developer.mozilla.org/en-US/docs/Web/API/Crypto/getRandomValues

sampaiodiego:
> @sampaiodiego @ostjen The problem with `Math.random` is [not being cryptographically secure](https://security.stackexchange.com/questions/181580/why-is-math-random-not-designed-to-be-cryptographically-secure), so it's not an option for generating passwords. Unsurprisingly, sometimes Snyk marks `Math.random()` as a vulnerability. This is better: https://developer.mozilla.org/en-US/docs/Web/API/Crypto/getRandomValues

the point is that `Math.random` is **not** being used to generate the password itself.. it's being used to get a index position of words array list..

tassoevan:
> it's being used to get an index position of words array list..

Which are combined to generate a password.

sampaiodiego (changes requested):
I don't know the root cause but while testing this I found out the keys were not successfully being generated anymore. not sure if the new password is not compatible (?)

ostjen:
> I don't know the root cause but while testing this I found out the keys were not successfully being generated anymore. not sure if the new password is not compatible (?)

tested here and it's working fine apparently 🤔

pierre-lehnen-rc (changes requested):

tassoevan (approved):

PR #24117, opened 2022-01-06
[FIX] Custom Emoji Image preview

### LINKED ISSUE #24082: Custom Emoji - Failed to load emoji thumbnail after uploaded
### Description:

I am getting the error related to content security policy after uploading a custom emoji -> I think it's causing thumbnail is not rendered

### Steps to reproduce:

1. Sign in as an admin user
2. Go to Administration >> Custom Emoji
3. Click + button to add a custom emoji
3. Upload an image with .png format
4. Observe the emoji thumbnail after uploaded

### Expected behavior:

Thumbnail of emoji should be loaded

### Actual behavior:

Thumbnail of emoji is not loaded

### Relevant info:

https://app.birdeatsbug.com/sessions/Ha1EWvkklqDHvN6-PPVVh?

sidmohanty11:
By using URL.createObjectURL() it produces a blob url which is not present in cors content-security-policy header so, its showing that error.

By making a data url, the image preview works fine.

Before,

custom-img-preview-rc3

After,

custom-img-preview-rc1

also if any error, (for example - if we upload a video mp4 file) 

custom-img-preview-rc2

## Issue(s)

Closes #24082


tassoevan:
Unless I'm doing some fundamental mistake, `data:` and `blob:` URIs are similar in terms of security, therefore a better correction would be to change the Content Security Policy of `img-src` to include `blob:` aside `data:`.

sidmohanty11:
Done! @tassoevan

dougfabris (approved):

PR #24121, opened 2022-01-07
[IMPROVE] Descriptive tooltip for Encrypted Key on Room Header

yash-rajpal:

### DIFF
--- a/client/views/room/Header/icons/Encrypted.js
+++ b/client/views/room/Header/icons/Encrypted.js
@@ -3,6 +3,7 @@ import colors from '@rocket.chat/fuselage-tokens/colors';
 import React, { memo } from 'react';
 
 import Header from '../../../../components/Header';
+import { usePermission } from '../../../../contexts/AuthorizationContext';
 import { useMethod } from '../../../../contexts/ServerContext';
 import { useSetting } from '../../../../contexts/SettingsContext';
 import { useTranslation } from '../../../../contexts/TranslationContext';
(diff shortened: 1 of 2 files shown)


albuquerquefabio (dismissed):

server/startup/migrations/v253.ts
```diff
@@ -0,0 +1,9 @@
+import { Permissions } from '../../../app/models/server/raw';
+import { addMigration } from '../../lib/migrations';
+
+addMigration({
+	version: 253,
+	up() {
+		return Permissions.update({ _id: 'toggle-room-e2e-encryption' }, { $addToSet: { roles: 'admin' } });
```
matheusbsilva137, line 7:
```suggestion
		return upsertPermissions();
```
I think you could use the `upsertPermissions` function instead, so you don't have to update it manually (e.g. see migration 244). This should have the same effect

server/startup/migrations/v253.ts
```diff
@@ -0,0 +1,9 @@
+import { Permissions } from '../../../app/models/server/raw';
```
matheusbsilva137, line 1:
```suggestion
import { upsertPermissions } from '../../../app/authorization/server/functions/upsertPermissions';
```
And this is the import you should use to make the other suggested change work.

matheusbsilva137 (changes requested):

server/startup/migrations/v253.ts
yash-rajpal: Done!

server/startup/migrations/v253.ts
yash-rajpal: Done! Thanks for the review!

matheusbsilva137 (dismissed):

albuquerquefabio (approved):

PR #24131, opened 2022-01-10
[FIX] Check if `file.uploadedAt` is present for user avatars

onigoetz:

* Fix `Cannot read property 'toUTCString' of undefined` in `server/routes/avatar/user.js`

## Steps to test or reproduce

We had this error in the logs of our instance (Rocket.Chat 4.2.0 running in Docker):

```
=== UnHandledPromiseRejection ===
TypeError: Cannot read property 'toUTCString' of undefined
    at server/routes/avatar/user.js:49:50
    at /app/bundle/programs/server/npm/node_modules/meteor/promise/node_modules/meteor-promise/fiber_pool.js:43:40
```

I don't know how this can be reproduced

but I see that in the code for `avatar/room.js` there is a similar `if (file.uploadedAt)` : 

https://github.com/RocketChat/Rocket.Chat/blob/0179f3fd9c39b8e3c82c72d5051273638e3c0fcf/server/routes/avatar/room.js#L39-L48

So I reproduced the change in this file.


pierreozoux:
Thanks @onigoetz for the fix, I also removed `EXIT_UNHANDLEDPROMISEREJECTION` in the mean time.
I'm wondering if I should then restart my app server once a day, to be sure the "delayed job" continue to work.

I'll see and report here if needed.

PR #23801, opened 2021-11-26
[NEW] Livechat voip/contexts providers

MartinSchoeler:

PR #24140, opened 2022-01-11
[NEW] Implementation of a call feature Hold-unhold

amolghode1981:
Hold unhold is a call feature that puts a call on hold. When a call is put on hold by an agent, it triggers a re-negotiation of SIP.  Which means that the media flows changed. When the call is put on hold, the agent and customer should not hear each other. Furthemore, the far end should hear the music on hold (if configured on the call server).

In ordered to do this, VoIPUser.ts should implement the code that triggers the re-invite. Important thing to note here is that hold-unhold feature can work only when the answer has been sent (or received) or the call is established. Because we are not handling the outbound calling right now, we will only check for ANSWER_SENT OR IN_CALL as a valid state for holding the call.

Call hold feature also needs to have toggling of the receiving of media as well as sending of the media. Which means, all the receiving streams and sending streams need to be disabled when the call gets hold. 

A new call state should be introduced to indicate the call hold state. When the call gets unhold, the call state changes to IN_CALL.

This workspace does not add the UI buttons but wires everything up to UI. It is a question of adding appropriate buttons and calling the correct action in the file VoiceController.tsx. 

## Issue(s)
https://app.clickup.com/t/7qdt7t
## Further comments
This feature can be tested when the UI buttons are in place. Following things need to be tested :
1. While in the call, when hold button is clicked, far end starts hearing the configured Music on hold.
2. Far end and near end (The one which put the call on hold) do not hear each other.
3. When unheld, they start hearing each other and far end stops hearing music on hold.
4. Can can be terminated while on hold.


tiagoevanp (commented):
I did a simple commit adjusting some TS types. If @amolghode1981 agrees, all LGTM.

definition/voip/VoIpCallerInfo.ts
```diff
@@ -7,18 +7,8 @@ export interface IState {
...
-	| {
-			state: 'OFFER_RECEIVED';
-			userState: UserState;
-			callerInfo: ICallerInfo;
-	  }
-	| {
-			state: 'IN_CALL';
-			userState: UserState;
-			callerInfo: ICallerInfo;
-	  }; // TODO: Check for additional properties and States (E.g. call on hold, muted, etc)
```
amolghode1981, line 24: @tiagoevanp I am not sure why these changes were done in this way. These changes were added https://github.com/RocketChat/Rocket.Chat/pull/23801/files#diff-d9b296df208a294ef147f00110a129467d20958e462293be58b54d333fb6f02a
So @MartinSchoeler may be able to put some light on this.

tiagoevanp (approved):

PR #24157, opened 2022-01-12
[IMPROVE] Endpoints

albuquerquefabio:
 **/room.export -** The logic was fixed.
 **/dns.resolve.txt -** For post request all params must be passed by body request
 **/dns.resolve.srv and /dns.resolve.txt -**  These routes should handle DNS queries error exceptions

## Issue(s)

- /room.export 
  - When one tries to export to email, that is with the param `{"type": "email"}`, we get the error "Cannot read property 'map' of undefined" ( even when "toUsers" and "toEmails" is provided)
  - Subject can not be empty and must be validate.

## Further comments
About the /dns.resolve.txt must be analyzed the impact. With this update all params are passed by body request which is the right way when using POST method.


app/api/server/v1/dns.ts
```diff
@@ -64,28 +64,40 @@ API.v1.addRoute(
...
-			const resolved = await resolveSRV(url);
+			try {
+				const resolved = await resolveSRV(url);
 
-			return API.v1.success({ resolved });
+				return API.v1.success({ resolved });
+			} catch (error: any) {
+				const { code, syscall, hostname } = error;
+
+				return API.v1.failure(`${syscall} ${code} ${hostname}`);
```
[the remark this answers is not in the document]
albuquerquefabio: Yes, I agree! But, in this case the right type is `NodeJS.ErrnoException` and this is not acceptable and you get this message `Catch clause variable type annotation must be 'any' or 'unknown' if specified.`
Even if you use this `// @ts-ignore` to ignore the warning your type automatically is `any`.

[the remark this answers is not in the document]
tassoevan: Got it. It looks like an issue with TypeScript 4.5 (probably the one active in your IDE). Let me try to implement some type guards for it...

server/lib/isErrnoException.ts
```diff
@@ -0,0 +1,12 @@
...
+export const isErrnoException = (error: unknown): error is NodeJS.ErrnoException => {
+	return (
+		isArbitraryObject(error) &&
+		error instanceof Error &&
+		(typeof error.errno === 'number' || typeof error.errno === 'undefined') &&
+		(typeof error.code === 'string' || typeof error.code === 'undefined') &&
+		(typeof error.path === 'string' || typeof error.path === 'undefined') &&
+		(typeof error.syscall === 'string' || typeof error.syscall === 'undefined')
+	);
+};
```
pierre-lehnen-rc, line 12: This will also return true for any Error instance that doesn't have those 4 attributes.

pierre-lehnen-rc (changes requested):

definition/rest/v1/rooms.ts
```diff
@@ -38,4 +38,26 @@ export type RoomsEndpoints = {
...
+				| {
+						rid: IRoom['_id'];
+						type: 'email';
+						toUsers: string[];
+						toEmails: string[];
+						subject: string;
+						messages: string[];
+				  },
+		) => {
+			missing: string[];
```
pierre-lehnen-rc, line 60: This return type is only compatible with the 'email' type. When using the 'file' type it'll not have any result.

server/lib/isErrnoException.ts
tassoevan: It's fine. The definition of `NodeJS.ErrnoException` ensures all these fields are optional. It's better to think about what is **NOT** a `NodeJS.ErrnoException`: something that is not an instance of `Error` and/or has one of the attributes with a _different_ type. For instance:

```ts
const valid = new Error();
isErrnoException(valid); // true

const invalid = Object.assign(new Error(), { errno: 'should be a number here' });
isErrnoException(invalid); // false
```

In that context, `isErrnoException` checks if the object's shape is compliant. Something like you are suggesting may be called `isStrictErrnoException`.

app/api/server/v1/rooms.js
```diff
@@ -549,15 +550,17 @@ API.v1.addRoute(
...
-				}
+				check(
+					this.bodyParams,
+					Match.ObjectIncluding({
+						subject: String,
+						toUsers: Match.Where((x) => Array.isArray(x) && x.length > 0 && typeof x[0] === 'string'),
+						toEmails: Match.Where((x) => Array.isArray(x) && x.length > 0 && typeof x[0] === 'string'),
+						messages: Match.Where((x) => Array.isArray(x) && x.length > 0 && typeof x[0] === 'string'),
+					}),
+				);
```
pierre-lehnen-rc, line 561:
```suggestion
				const nonEmptyStringArray = Match.Where((x) => {
					check(x, [String]);
					return x.length > 0;
				});

				check(
					this.bodyParams,
					Match.ObjectIncluding({
						subject: String,
						toUsers: nonEmptyStringArray,
						toEmails: nonEmptyStringArray,
						messages: nonEmptyStringArray,
					}),
				);
```

definition/rest/v1/rooms.ts
tassoevan: I had to rollback as there is a design limitation on how to represent overloads in REST API operation (i.e. how different payloads and responses are at the same method and path).

pierre-lehnen-rc (approved):

matheusbsilva137 (dismissed):

pierre-lehnen-rc (changes requested):

apps/meteor/app/channel-settings/server/methods/saveRoomSettings.js
```diff
@@ -241,8 +241,8 @@ const settingSavers = {
 	encrypted({ value, room, rid, user }) {
 		saveRoomEncrypted(rid, value, user, Boolean(room.encrypted) !== Boolean(value));
 	},
-	favorite({ value, rid }) {
-		Rooms.saveFavoriteById(rid, value.favorite, value.defaultValue);
+	favorite({ value, rid, user }) {
+		Subscriptions.setFavoriteByRoomIdAndUserId(rid, user._id, value.favorite);
```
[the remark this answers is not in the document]
albuquerquefabio: Done

apps/meteor/app/channel-settings/server/methods/saveRoomSettings.js
```diff
@@ -241,8 +241,8 @@ const settingSavers = {
 	encrypted({ value, room, rid, user }) {
 		saveRoomEncrypted(rid, value, user, Boolean(room.encrypted) !== Boolean(value));
 	},
-	favorite({ value, rid }) {
-		Rooms.saveFavoriteById(rid, value.favorite, value.defaultValue);
+	favorite({ value, rid, user }) {
+		Subscriptions.updateOne({ rid, 'u._id': user._id }, { $set: { f: value ?? true } });
```
pierre-lehnen-rc, line 245: value is an object with two properties (`favorite` and `defaultValue`)

pierre-lehnen-rc (changes requested):

apps/meteor/app/channel-settings/server/methods/saveRoomSettings.js
albuquerquefabio: Updated

apps/meteor/app/channel-settings/server/functions/saveFavoriteRoom.ts
```diff
@@ -0,0 +1,16 @@
...
+import { Subscriptions } from '@rocket.chat/models';
+
+export const saveFavoriteRoom = function (rid: IRoom['_id'], uid: IUser['_id'], favorite?: boolean): Promise<UpdateResult> {
+	return Subscriptions.updateOne(
+		{
+			rid,
+			'u._id': uid,
+		},
+		{
+			...(favorite && { $set: { f: true } }),
```
LucianoPierdona, line 12: can we update this to be only in one single line?
```suggestion
			...(favorite ? { $set: { f: true } } : { $unset: { f: '' } }),
```

apps/meteor/app/channel-settings/server/methods/saveRoomSettings.js
```diff
@@ -294,7 +296,8 @@ Meteor.methods({
 			settings = { encrypted: settings.encrypted };
 		}
 
-		if (room.broadcast && (settings.readOnly || settings.reactWhenReadOnly)) {
+		const checkReadOnly = Object.keys(settings).includes('readOnly') || Object.keys(settings).includes('reactWhenReadOnly');
```
LucianoPierdona, line 299: just ignore if you think it's not really important in this case, but could you assign `Object.keys(settings)` to one single variable? the way it's done we're reading the fields twice.

```suggestion
                const settingsKey = Object.keys(settings);
		const checkReadOnly = settingsKey.includes('readOnly') || settingsKey.includes('reactWhenReadOnly');
```

apps/meteor/client/views/room/contextualBar/ExportMessages/MailExportForm.tsx
```diff
@@ -97,7 +97,7 @@ const MailExportForm: FC<MailExportFormProps> = ({ onCancel, rid }) => {
 	const roomsExport = useEndpoint('POST', '/v1/rooms.export');
 
 	const handleSubmit = async (): Promise<void> => {
-		if (toUsers.length === 0 && additionalEmails === '') {
+		if (toUsers === undefined || (toUsers?.length === 0 && additionalEmails === '')) {
```
LucianoPierdona, line 100:
```suggestion
		if (!toUsers || (toUsers.length === 0 && !additionalEmails)) {
```

apps/meteor/server/lib/channelExport.ts
```diff
@@ -53,7 +53,7 @@ type ISentViaEmail = {
 };
 
 export const sendViaEmail = (data: ExportEmail, user: IUser): ISentViaEmail => {
-	const emails = data.toEmails.map((email) => email.trim()).filter(Boolean);
+	const emails = data.toEmails.map((email) => email.trim()).filter(Boolean) || [];
```
LucianoPierdona, line 56: is there any case where it might fall into the `or`
```suggestion
	const emails = data.toEmails.map((email) => email.trim()).filter(Boolean);
```

apps/meteor/app/channel-settings/server/methods/saveRoomSettings.js
tassoevan, line 299:
```suggestion
		const checkReadOnly = 'readOnly' in settings || 'reactWhenReadOnly' in settings;
```

tassoevan (changes requested):

tassoevan (dismissed):

apps/meteor/app/api/server/v1/rooms.js
```diff
@@ -559,14 +559,8 @@ API.v1.addRoute(
...
 			if (type === 'email') {
-				const { toUsers, toEmails, subject, messages } = this.bodyParams;
-
-				if ((!toUsers || toUsers.length === 0) && (!toEmails || toEmails.length === 0)) {
-					throw new Meteor.Error('error-invalid-recipient');
-				}
-
-				if (messages.length === 0) {
-					throw new Meteor.Error('error-invalid-messages');
+				if ((!toEmails && !toUsers) || (!toUsers?.filter(Boolean).length && !toEmails?.filter(Boolean).length)) {
```
pierre-lehnen-rc, line 562:
```suggestion
				if (!toUsers?.filter(Boolean).length && !toEmails?.filter(Boolean).length) {
```

apps/meteor/app/api/server/v1/dns.ts
```diff
@@ -111,20 +126,28 @@ API.v1.addRoute(
...
 			check(
-				this.queryParams,
+				this.bodyParams,
 				Match.ObjectIncluding({
 					url: String,
 				}),
 			);
 
-			const { url } = this.queryParams;
+			const { url } = this.bodyParams;
```
pierre-lehnen-rc, line 135:
```suggestion
			const { url } = this.bodyParams;
```
This fix is a breaking change. 
Add a fallback to make it still work with queryParams, just ensure it shows a deprecation notice in that case.

apps/meteor/client/views/room/contextualBar/ExportMessages/MailExportForm.tsx
```diff
@@ -97,7 +97,7 @@ const MailExportForm: FC<MailExportFormProps> = ({ onCancel, rid }) => {
 	const roomsExport = useEndpoint('POST', '/v1/rooms.export');
 
 	const handleSubmit = async (): Promise<void> => {
-		if (toUsers.length === 0 && additionalEmails === '') {
+		if (!toUsers || (toUsers.length === 0 && !additionalEmails)) {
```
pierre-lehnen-rc, line 100: This conditional feels wrong. Should it really have a different behavior when `toUsers` is not set than when it is set to an empty array?
If that's correct, please add a comment explaining the reason.

pierre-lehnen-rc (changes requested):

apps/meteor/client/views/room/contextualBar/ExportMessages/MailExportForm.tsx
tassoevan: Actually, it's never going to be `undefined.`

apps/meteor/app/channel-settings/server/methods/saveRoomSettings.ts
```diff
@@ -333,12 +355,13 @@ Meteor.methods({
...
 
 		// saving data
-		for await (const setting of Object.keys(settings)) {
+		keys.forEach((setting) => {
 			const value = settings[setting];
 
-			const saver = await settingSavers[setting];
+			const saver = settingSavers[setting];
 			if (saver) {
 				saver({
```
pierre-lehnen-rc, line 366: One of the savers is async... But it was already being called without await so it's not a blocker for this PR - but probably a bug.

apps/meteor/app/channel-settings/server/methods/saveRoomSettings.ts
```diff
@@ -1,10 +1,11 @@
 import { Meteor } from 'meteor/meteor';
 import { Match } from 'meteor/check';
+import type { IRoom, RoomType, IRoomWithRetentionPolicy, IUser } from '@rocket.chat/core-typings';
 import { TEAM_TYPE } from '@rocket.chat/core-typings';
+import { Rooms } from '@rocket.chat/models';
```
pierre-lehnen-rc, line 5: If you change the code to use the raw model then you can't assume that everything will be running in sync anymore.

pierre-lehnen-rc (changes requested):

apps/meteor/app/channel-settings/server/methods/saveRoomSettings.ts
```diff
@@ -199,64 +217,65 @@ const settingSavers = {
...
+	async systemMessages({ value, room, rid }: { value: boolean; room: IRoom; rid: IRoom['_id']; user: IUser }) {
 		if (JSON.stringify(value) !== JSON.stringify(room.sysMes)) {
-			saveRoomSystemMessages(rid, value, user);
+			saveRoomSystemMessages(rid, value);
 		}
 	},
-	joinCode({ value, rid }) {
-		Rooms.setJoinCodeById(rid, String(value));
+	async joinCode({ value, rid }: { value: string; rid: IRoom['_id'] }) {
+		await Rooms.setJoinCodeById(rid, String(value));
```
pierre-lehnen-rc, line 242: The function can now be removed from the old model as it is no longer used anywhere

apps/meteor/app/channel-settings/server/methods/saveRoomSettings.ts
```diff
@@ -199,64 +217,65 @@ const settingSavers = {
...
 	},
-	default({ value, rid }) {
-		Rooms.saveDefaultById(rid, value);
+	async default({ value, rid }: { value: boolean; rid: IRoom['_id'] }) {
+		await Rooms.saveDefaultById(rid, value);
```
pierre-lehnen-rc, line 245: The function can now be removed from the old model as it is no longer used anywhere

packages/model-typings/src/models/IRoomsModel.ts
```diff
@@ -46,6 +46,22 @@ export interface IRoomsModel extends IBaseModel<IRoom> {
...
+	saveRetentionMaxAgeById(rid: IRoom['_id'], maxAge: number): Promise<UpdateResult>;
+
+	saveRetentionExcludePinnedById(rid: IRoom['_id'], excludePinned: boolean): Promise<UpdateResult>;
+
+	saveRetentionFilesOnlyById(rid: IRoom['_id'], filesOnly: boolean): Promise<UpdateResult>;
+
+	saveRetentionIgnoreThreadsById(rid: IRoom['_id'], ignoreThreads: boolean): Promise<UpdateResult>;
+
+	saveRetentionOverrideGlobalById(rid: IRoom['_id'], overrideGlobal: boolean): Promise<UpdateResult>;
+
```
pierre-lehnen-rc, line 64: Those 7 functions were not created in the raw model.

apps/meteor/app/channel-settings/server/methods/saveRoomSettings.ts
```diff
@@ -199,64 +217,65 @@ const settingSavers = {
...
 	},
-	featured({ value, rid }) {
-		Rooms.saveFeaturedById(rid, value);
+	async featured({ value, rid }: { value: boolean; rid: IRoom['_id'] }) {
+		await Rooms.saveFeaturedById(rid, value);
```
pierre-lehnen-rc, line 248: This will throw an error as the function was not created in the raw model.

apps/meteor/app/channel-settings/server/methods/saveRoomSettings.ts
```diff
@@ -199,64 +217,65 @@ const settingSavers = {
...
 	},
-	retentionEnabled({ value, rid }) {
-		Rooms.saveRetentionEnabledById(rid, value);
+	async retentionEnabled({ value, rid }: { value: boolean; rid: IRoom['_id'] }) {
+		await Rooms.saveRetentionEnabledById(rid, value);
```
pierre-lehnen-rc, line 251: This will throw an error as the function was not created in the raw model.

apps/meteor/app/channel-settings/server/methods/saveRoomSettings.ts
```diff
@@ -199,64 +217,65 @@ const settingSavers = {
...
 	},
-	retentionMaxAge({ value, rid }) {
-		Rooms.saveRetentionMaxAgeById(rid, value);
+	async retentionMaxAge({ value, rid }: { value: number; rid: IRoom['_id'] }) {
+		await Rooms.saveRetentionMaxAgeById(rid, value);
```
pierre-lehnen-rc, line 254: This will throw an error as the function was not created in the raw model.

apps/meteor/app/channel-settings/server/methods/saveRoomSettings.ts
```diff
@@ -199,64 +217,65 @@ const settingSavers = {
...
 	},
-	retentionExcludePinned({ value, rid }) {
-		Rooms.saveRetentionExcludePinnedById(rid, value);
+	async retentionExcludePinned({ value, rid }: { value: boolean; rid: IRoom['_id'] }) {
+		await Rooms.saveRetentionExcludePinnedById(rid, value);
```
pierre-lehnen-rc, line 257: This will throw an error as the function was not created in the raw model.

apps/meteor/app/channel-settings/server/methods/saveRoomSettings.ts
```diff
@@ -199,64 +217,65 @@ const settingSavers = {
...
 	},
-	retentionFilesOnly({ value, rid }) {
-		Rooms.saveRetentionFilesOnlyById(rid, value);
+	async retentionFilesOnly({ value, rid }: { value: boolean; rid: IRoom['_id'] }) {
+		await Rooms.saveRetentionFilesOnlyById(rid, value);
```
pierre-lehnen-rc, line 260: This will throw an error as the function was not created in the raw model.

apps/meteor/app/channel-settings/server/methods/saveRoomSettings.ts
```diff
@@ -199,64 +217,65 @@ const settingSavers = {
...
 	},
-	retentionIgnoreThreads({ value, rid }) {
-		Rooms.saveRetentionIgnoreThreadsById(rid, value);
+	async retentionIgnoreThreads({ value, rid }: { value: boolean; rid: IRoom['_id'] }) {
+		await Rooms.saveRetentionIgnoreThreadsById(rid, value);
```
pierre-lehnen-rc, line 263: This will throw an error as the function was not created in the raw model.

apps/meteor/app/channel-settings/server/methods/saveRoomSettings.ts
```diff
@@ -199,64 +217,65 @@ const settingSavers = {
...
 	},
-	retentionOverrideGlobal({ value, rid }) {
-		Rooms.saveRetentionOverrideGlobalById(rid, value);
+	async retentionOverrideGlobal({ value, rid }: { value: boolean; rid: IRoom['_id'] }) {
+		await Rooms.saveRetentionOverrideGlobalById(rid, value);
```
pierre-lehnen-rc, line 266: This will throw an error as the function was not created in the raw model.

pierre-lehnen-rc (changes requested):

PR #24161, opened 2022-01-13
[IMPROVE] Extended message Actions menu when shift key is pressed while hovering 

try-catch-stack:
After using RocketChat for a while , I feel that it would be really great to have this feature of showing a extended messageActions menu with delete,copy and few more buttons when a particular key like the shift key is pressed while hovering over a message.
It'll make deleting,copying and other similar actions really convenient as we'll not have to go through the menu every time for these.

I also posted it on feature request repo and after getting a positive response from @dudanogueira , started working on it.
The same can be found on https://github.com/RocketChat/feature-requests/issues/535

This feature adds a extended message actions menu with some additional options like delete ,reply, copy , star etc. which appears when we hover over a message while pressing shift key . This makes deleting, replying and similar actions convenient as it can be done in a single click.

https://user-images.githubusercontent.com/77742477/149315735-16eca76e-906e-45dd-aa38-3c4fa44579c6.mp4

## Further comments

This feature is work in progress and has a lot of room for improvement. So I invite the dev team and other contributors to provide their valuable feedback and suggestions.

### DIFF
--- a/app/message-star/client/actionButton.js
+++ b/app/message-star/client/actionButton.js
@@ -38,7 +38,7 @@ Meteor.startup(function () {
 			return !message.starred || !message.starred.find((star) => star._id === u._id);
 		},
 		order: 9,
-		group: 'menu',
+		group: ['menu', 'extended'],
 	});
 
 	MessageAction.addButton({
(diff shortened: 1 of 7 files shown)


ankar84:
Looks good! Thank you!

PR #24162, opened 2022-01-13
[NEW][EE] Allow to filter departments by Business Units on Livechat

murtaza98:


app/livechat/server/api/lib/livechat.js
```diff
@@ -107,10 +110,24 @@ export function normalizeHttpHeaderData(headers = {}) {
 	const httpHeaders = Object.assign({}, headers);
 	return { httpHeaders };
 }
-export async function settings() {
+export async function settings(businessUnitIds = '') {
```
renatobecker, line 113:
```suggestion
export async function settings({ businessUnit = '' }) {
```

Array is not expected, just one single business unit (Id / Name) will be passed as parameter.

app/livechat/server/api/v1/config.js
```diff
@@ -10,16 +10,18 @@ API.v1.addRoute('livechat/config', {
...
 			const enabled = Livechat.enabled();
 
 			if (!enabled) {
 				return API.v1.success({ config: { enabled: false } });
 			}
 
-			const config = Promise.await(settings());
+			const { token, department, businessUnitIds } = this.queryParams;
+
+			const config = Promise.await(settings(businessUnitIds));
```
renatobecker, line 23: Have you thought about passing an object instead of a single parameter?

ee/app/livechat-enterprise/server/hooks/findDepartmentsWithBusinessUnits.ts
```diff
@@ -0,0 +1,15 @@
+import { LivechatDepartment } from '../../../../../app/models/server';
+import { callbacks } from '../../../../../lib/callbacks';
+
+callbacks.add(
+	'livechat.findDepartmentsWithBusinessUnits',
+	async ({ businessUnitIds } = { businessUnitIds: [] }) => {
+		console.log('enterprise code called', businessUnitIds);
+
+		return LivechatDepartment.findByUnitIds(businessUnitIds, {
```
renatobecker, line 9: What if the BU doesn't exist? We need to verify/validate before reading departments.

renatobecker (dismissed):

app/livechat/server/api/v1/config.js
murtaza98: Since it was just a single param, so I initially went like this... but let me change it

KevLehman (changes requested):

app/livechat/server/api/lib/livechat.js
```diff
@@ -107,10 +110,23 @@ export function normalizeHttpHeaderData(headers = {}) {
...
 }
-export async function settings() {
+export async function settings({ businessUnit = '' }) {
 	const initSettings = Livechat.getInitSettings();
 	const triggers = await findTriggers();
-	const departments = findDepartments();
+	let departments = [];
+	if (businessUnit) {
+		logger.debug('Attempting to get departments connected to business unit', businessUnit);
+		departments = await callbacks.run('livechat.findDepartmentsWithBusinessUnits', { businessUnit });
```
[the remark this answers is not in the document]
renatobecker: Great approach, @KevLehman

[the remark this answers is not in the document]
murtaza98: Good idea!! Added this change. Please check again :smiley:

app/livechat/server/api/v1/config.js
```diff
@@ -10,16 +10,18 @@ API.v1.addRoute('livechat/config', {
...
 			const enabled = Livechat.enabled();
 
 			if (!enabled) {
 				return API.v1.success({ config: { enabled: false } });
 			}
 
-			const config = Promise.await(settings());
+			const { token, department, businessUnit } = this.queryParams;
+
+			const config = Promise.await(settings({ businessUnit }));
```
KevLehman, line 23: You can make the `get` async now, and use normal await instead of this one 👀 
```suggestion
			const config = await settings({ businessUnit });
```

KevLehman (changes requested):

KevLehman (approved):

KevLehman (approved):

PR #24173, opened 2022-01-13
[FIX] respect `Accounts_Registration_Users_Default_Roles

### LINKED ISSUE #24148: Role `user` assign to registered users despite of `Default Roles for Users` settings
### Description:

Some versions ago a very good setting was added to **Admin UI - Account - Registration**
It's a `Default Roles for Users`
image
As you can see on a screen we assign special restricted role `invited-user` to self registered users.
We are using secret URL and invite links to register external users.

In description of settings you see this:
`Default roles (comma-separated) users will be given when registering through manual registration (including via API)`
But in fact it's not true! Role `user` is added to registered user anyway!
I believe it's a bug! Because if I need that role `user` assigned to new registered user account I will set `Default Roles for Users` with that role.

As you can see, there is no role `user` in list, but anyway all new registered users get that `user` role for some strange reason.

### Steps to reproduce:

1. 
2. 
3. 

### Expected behavior:

Role `user` should be added to new registered users in it set in `Default Roles for Users` setting of **Admin UI - Account - Registration** and should not be added if it's not in list of `Default Roles for Users` setting

### Actual behavior:

Role `user` added to new registered users for some reason

### Server Setup Information:

- Version of Rocket.Chat Server: 4.2.2
- Operating System: CentOS7
- Deployment Method: docker
- NodeJS Version: 12.22.1
- MongoDB Version: 4.4.11

### Client Setup Information

- Desktop App or Browser Version: 3.7.5

### Additional context

We used our own patch in 3.x versions to remove that `user` role, but now on 4.x server version with our patch users gets an error on registration process

Here is a patch
`Roles.removeUserRoles(userId, ['user']);`

And here is an error:
`{"level":50,"time":"2022-01-11T11:17:19.603Z","pid":10,"hostname":"27525ea62f59","name":"System","msg":"Exception while invoking method registerUser 'Cannot read property 'removeUserRoles' of undefined'"}`

### Relevant logs:

No

debdutdeb:

## Issue(s)

closes #24148

### DIFF
--- a/app/authentication/server/startup/index.js
+++ b/app/authentication/server/startup/index.js
@@ -8,7 +8,7 @@ import { escapeRegExp, escapeHTML } from '@rocket.chat/string-helpers';
 import * as Mailer from '../../../mailer/server/api';
 import { settings } from '../../../settings/server';
 import { callbacks } from '../../../../lib/callbacks';
-import { Users, Settings } from '../../../models/server';
+import { Settings, Users } from '../../../models/server';
 import { Roles, Users as UsersRaw } from '../../../models/server/raw';
 import { addUserRoles } from '../../../authorization/server';
 import { getAvatarSuggestionForUser } from '../../../lib/server/functions';
@@ -213,8 +213,6 @@ Accounts.onCreateUser(function (options, user = {}) {
 });
 
 Accounts.insertUserDoc = _.wrap(Accounts.insertUserDoc, function (insertUserDoc, options, user) {
-	const noRoles = !user?.hasOwnProperty('globalRoles');
-
 	const globalRoles = [];
 
 	if (Match.test(user.globalRoles, [String]) && user.globalRoles.length > 0) {
@@ -278,27 +276,22 @@ Accounts.insertUserDoc = _.wrap(Accounts.insertUserDoc, function (insertUserDoc,
 		}
 	}
 
-	if (noRoles || roles.length === 0) {
-		const hasAdmin = Users.findOne(
-			{
-				roles: 'admin',
-				type: 'user',
-			},
-			{
-				fields: {
-					_id: 1,
-				},
-			},
-		);
-
-		if (hasAdmin) {
-			roles.push('user');
-		} else {
+	/**
+	 * if settings shows setup wizard to be pending
+	 * and no admin's been found,
+	 * and existing role list doesn't include admin
+	 * create this user admin.
+	 * count this as the completion of setup wizard step 1.
+	 */
+	if (settings.get('Show_Setup_Wizard') === 'pending') {
+		// prettier-ignore
+		if (
+			!roles.includes('admin')
+			&&
+			Users.findOne({ roles: 'admin', type: 'user' }, { fields: { _id: 1 } }) === undefined
+		)
 			roles.push('admin');
-			if (settings.get('Show_Setup_Wizard') === 'pending') {
-				Settings.updateValueById('Show_Setup_Wizard', 'in_progress');
-			}
-		}
+		Settings.updateValueById('Show_Setup_Wizard', 'in_progress');
 	}
 
 	addUserRoles(_id, roles);


ankar84:
We merge that and tested - it works perfect - new registered user gets only roles from list, and no `user` role!
Thank you, @debdutdeb 
Wait it to be merged in main codebase!

ankar84:
Hope that PR will land to 4.4.0 release!
@sampaiodiego please consider to merge it.

app/authentication/server/startup/index.js
```diff
@@ -278,27 +276,22 @@ Accounts.insertUserDoc = _.wrap(Accounts.insertUserDoc, function (insertUserDoc,
...
-			roles.push('user');
-		} else {
+	/**
+	 * if settings shows setup wizard to be pending
+	 * and no admin's been found,
+	 * and existing role list doesn't include admin
+	 * create this user admin.
+	 * count this as the completion of setup wizard step 1.
+	 */
+	if (settings.get('Show_Setup_Wizard') === 'pending') {
```
matheusbsilva137, line 286: Using this as the "first" condition would only allow us promoting users to admin (when there are no other admins in the server) when the Setup Wizard is still pending, right? This is a slight change from what we were doing, but I'm afraid this could create a unwanted behavior by leaving the server with no admins 🤔 
Note that, previously, we could promote a new user to admin regardless of the setup wizard status.

matheusbsilva137 (changes requested):

app/authentication/server/startup/index.js
debdutdeb: I though about that, but then iirc we don't allow deletion of the last admin via the app any way.

I also just tested it,

```
{
  "success": false,
  "error": "Leaving the app without admins is not allowed [error-action-not-allowed]",
  "errorType": "error-action-not-allowed",
  "details": {
    "method": "deleteUser",
    "action": "Remove_last_admin"
  }
}
```

The endpoint was called by a non-admin user with just user-delete permission.

This means the server can end up with no admin iff it's done directly by modifying the collection. In which case, rectifying would still be possible.

matheusbsilva137: > Doesn't seem like it's necessary, at least one admin's existence is enforced via the app/api.

Yup, agree. Still, I'm not sure we're not missing any edge case 🤔
I've just created #24308 based on your branch with some changes: I noticed we would miss the "user" role when creating users by the `userHandler.js` file, so I added this role as the default there. I've also added the `findOneByRolesAndType` method to improve legibility.
Please take a look there and feel free to request changes or approve it if you think it's ok 🙂

ankar84: > I noticed we would miss the "user" role when creating users by the `userHandler.js` file, so I added this role as the default there. I've also added the `findOneByRolesAndType` method to improve legibility.

That's not correct. Check issue that this PR intended to fix https://github.com/RocketChat/Rocket.Chat/issues/24148

debdutdeb:
Just tested the changes, doesn't seem to be working :/

app/authentication/server/startup/index.js
```diff
@@ -278,26 +276,18 @@ Accounts.insertUserDoc = _.wrap(Accounts.insertUserDoc, function (insertUserDoc,
...
+	 * create this user admin.
+	 * count this as the completion of setup wizard step 1.
+	 */
+	const hasAdmin = Users.findOneByRolesAndType('admin', 'user', { fields: { _id: 1 } });
+	if (!roles.includes('admin') && !hasAdmin) {
+		roles.push('admin');
+		if (settings.get('Show_Setup_Wizard') === 'pending') {
+			Settings.updateValueById('Show_Setup_Wizard', 'in_progress');
 		}
 	}
```
debdutdeb, line 292:
```suggestion
	/**
	 * if settings shows setup wizard to be pending
	 * and existing role list doesn't include admin
	 * create this user admin.
	 * count this as the completion of setup wizard step 1.
	 */
	if (settings.get('Show_Setup_Wizard') === 'pending') {
		if (roles.includes('admin')) {
		    roles.push('admin');
		}
		Settings.updateValueById('Show_Setup_Wizard', 'in_progress');
	}
```

Hey @matheusbsilva137 - take a look at this, https://github.com/RocketChat/Rocket.Chat/blob/443866e706602a2cf33a56e4e3b25ccacf7e7c42/server/startup/initialData.js#L138-L145

No admin situation is already handled at startup, so I don't think we need to do re-do that here. Could be wrong. I'm just leaving this  as a suggestion here.

ostjen (approved):

matheusbsilva137 (dismissed):

app/authentication/server/startup/index.js
matheusbsilva137: I've just realized this approach could create a security issue 🤔 
Since the `Show_Setup_Wizard` setting is _not hidden_, a user may have the permission to edit it (eg to `pending`) and, then, create a new user who will become an admin (even though the server already has an admin user).

matheusbsilva137 (approved):

PR #24177, opened 2022-01-14
[IMPROVE] lib/Statistics improved and metrics collector

albuquerquefabio:

- On `statistics` object the property `get`  is an async function now.
- We need to collect additional data of feature activation through the statistics collector.
  - Some codes were splitted into another file just to organize.


app/statistics/server/lib/statistics.js
```diff
@@ -215,33 +214,35 @@ export const statistics = {
...
-				{},
-				{
-					projection: {
-						_id: 0,
-						type: 1,
-						enabled: 1,
-						scriptEnabled: 1,
-					},
-					readPreference,
+		// If getSettingsStatistics9) throw error save data as empty object
```
ostjen, line 230: typo

server/lib/statistics/getSettingsStatistics.ts
```diff
@@ -0,0 +1,160 @@
...
+			{ key: 'WebRTC_Enable_Direct', alias: 'webRTCEnableDirect' },
+			{ key: 'Canned_Responses_Enable', alias: 'cannedResponses' },
+		];
+
+		// Mapping only _id values
+		const settingsIDs = settingsBase.map((el) => el.key);
+
+		const settingsArray = await Settings.findByIds(settingsIDs).toArray();
+		const settingsStatistics = settingsArray
+			.map((el): ISettingStatistics => {
```
tassoevan, line 146: I believe the `Cursor` object returned by `Settings.findByIds()` has a `.map()` method:

```suggestion
		// no intermediary `settingsArray`
		const settingsStatistics = await Settings.findByIds(settingsIDs)
			.map((el): ISettingStatistics => {
```

But the result of it may be `Promise<ISettingStatistics[]>`, so you may need an `await` before applying `.filter` and `.reduce` methods.

tassoevan:
> * 1 for Syntax error

I think there are some calls of `Promise.await` to be replaced with `await`.

server/lib/statistics/getSettingsStatistics.ts
albuquerquefabio: Unfortunately the `Cursor` return only `ISetting` and to get `ISetting[]` is necessary `.toArray()`

tassoevan: `Cursor<ISetting>` is almost like a stream of `ISetting` objects, where `.toArray()` works like a "collector" or "aggregator". You could chain `.map()` and `.filter()` generating new `Cursor<T>` instances before invoking `.toArray()`. Take a look at my commit d077c8f.

tassoevan (approved):

PR #24180, opened 2022-01-14
[NEW] Livechat voip/queue events

amolghode1981:
This PR implements continuous monitor in the asterisk connector. Asterisk continuously generates a stream of events on various activities. Management server user defined in asterisk's manager.conf determines the events to be sent on this user. This PR's main focus is to find out calls in the queue and source queue of a call.

To do this, it monitors following events
    

    agentcalled (For finding out the source queue of a call)  (message is sent to a to a single agent)
    queuecallerjoin   (for calls in the queue) (message is sent to all the extensions in the queue)
    agentconnect (for calls in the queue) (message is sent to all the extensions in the queue)
    queuememberadded (for calls in the queue) (message is sent to all the extensions in the queue)
    queuememberremoved (for calls in the queue) (message is sent to all the extensions in the queue)
    queuecallerabandon (for calls in the queue) (message is sent to all the extensions in the queue)
 
These events are then captured, processed and necessary information extracted. In some cases, the code also uses other connector commands (e.g queue_status) to determine certain parameters which are necessary to derive certain information.

e.g When a customer calls a queue, we need to find out all the members of the queue. This is achieved using queue_status command. Once the members are found necessary information is sent to all the members in the queue.

When the members receive the notification, the have aggregator, which makes some calculation and determines the calls waiting in a queue. 

## Issue(s)
There are some improvements necessary in the future :
1. We do not have a way to configure the events we want to watch for from our admin page. Currently these values are hardcoded. This should be configurable based on feature (source queue of a call, calls waiting in queue). So to support a feature, the continuous monitor uses certain events.
2. There is a lot of data that asterisk sends on the connection. This can be configured by correctly configuring the asterisk user on manager.conf of asterisk. This needs to be documented.
3. We do not have a strategy to reinitiate the connection or check the connectivity.
## Further comments
Testing Notes :
For determining source queue of a call :
1. Verify that the user has correct information when the call is ringing. Also verify that the user whose call is ringing gets this information. 
2. When user is a part of 2 queues, verify that user receives correct information.
3. When user does not accept the call, verify that the information is cleared.

For determining calls waiting in the queue :
1. Verify that user receives the aggregated calls in queue information when the agent is part of multiple queues.
2. Verify that on customer calling queue, agent accepting call, agent added/removed from the queue, customer abandoning queue events are properly managed so that the agent has correct and the latest view.


app/lib/server/functions/getFullUserData.js
```diff
@@ -30,6 +30,7 @@ const fullFields = {
 	requirePasswordChange: 1,
 	requirePasswordChangeReason: 1,
 	roles: 1,
+	extension: 1,
```
KevLehman, line 33: I added this for now, but, if we don't want to leak this prop the client, maybe we should remove it, wdyt?

server/services/voip/connector/asterisk/CommandHandler.ts
```diff
@@ -60,13 +63,11 @@ export class CommandHandler {
...
-			config.host,
-			(config.configData as IManagementConfigData).port.toString(),
-			(config.configData as IManagementConfigData).username,
-			(config.configData as IManagementConfigData).password,
-		);
+		connection.connect(config.host, config.configData.port.toString(), config.configData.username, config.configData.password);
 		this.connections.set(commandType, connection);
+		this.continuousMonitor = CommandFactory.getCommandObject(Commands.event_stream, this.db);
+		this.continuousMonitor.connection = this.connections.get(this.continuousMonitor.type) as IConnection;
+		this.continuousMonitor.initMonitor({ data: 'test' });
```
KevLehman, line 70: 👀

server/services/voip/lib/Helper.ts
```diff
@@ -0,0 +1,33 @@
...
+
+export function getServerConfigDataFromSettings(type: ServerType): IVoipCallServerConfig | IVoipManagementServerConfig {
+	switch (type) {
+		case ServerType.CALL_SERVER: {
+			const serverCofig: IVoipCallServerConfig = {
+				type: ServerType.CALL_SERVER,
+				host: settings.get<string>('VoIP_Server_Host'),
+				name: settings.get<string>('VoIP_Server_Name'),
+				configData: {
+					websocketPort: settings.get<number>('VoIP_Server_Websocket_Port'),
```
KevLehman, line 12:
```suggestion
					websocketPort: Number(settings.get<number>('VoIP_Server_Websocket_Port')),
```

server/services/voip/lib/Helper.ts
```diff
@@ -0,0 +1,33 @@
...
+			return serverCofig;
+		}
+
+		case ServerType.MANAGEMENT: {
+			const serverCofig: IVoipManagementServerConfig = {
+				type: ServerType.MANAGEMENT,
+				host: settings.get<string>('VoIP_Management_Server_Host'),
+				name: settings.get<string>('VoIP_Management_Server_Name'),
+				configData: {
+					port: settings.get<number>('VoIP_Management_Server_Port'),
```
KevLehman, line 25:
```suggestion
					port: Number(settings.get<number>('VoIP_Management_Server_Port')),
```

KevLehman (changes requested):

server/services/voip/connector/asterisk/Command.ts
```diff
@@ -80,11 +82,22 @@ export class Command {
...
+
+	get db(): any {
+		return this._db;
+	}
+
+	set db(db: Db) {
+		this._db = db;
+	}
+
+	constructor(command: string, parametersNeeded: boolean, db?: Db) {
```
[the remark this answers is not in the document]
amolghode1981: Fixed. I did it for the reason that it is in the base class. But not all Command objects need to have the DB.

server/services/voip/connector/asterisk/ami/ContinuousMonitor.ts
```diff
@@ -0,0 +1,234 @@
...
+		// Setup necessary command event handlers based on the command
+		this.connection.on('queuecallerjoin', new CallbackContext(this.onEvent.bind(this), this));
+		this.connection.on('agentcalled', new CallbackContext(this.onEvent.bind(this), this));
+		this.connection.on('agentconnect', new CallbackContext(this.onEvent.bind(this), this));
+		this.connection.on('queuememberadded', new CallbackContext(this.onEvent.bind(this), this));
+		this.connection.on('queuememberremoved', new CallbackContext(this.onEvent.bind(this), this));
+		this.connection.on('queuecallerabandon', new CallbackContext(this.onEvent.bind(this), this));
+	}
+
+	resetEventHandlers(): void {
```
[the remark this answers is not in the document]
amolghode1981: Fixed.

server/services/voip/service.ts
```diff
@@ -126,39 +127,39 @@ export class VoipService extends ServiceClass implements IVoipService {
...
 		};
-		membershipDetails.extension = requestParams.extension;
-		const queueSummary = Promise.await(this.commandHandler.executeCommand(Commands.queue_summary)) as IVoipConnectorResult;
-		for (const queue of queueSummary.result as IQueueSummary[]) {
-			const queueDetails = Promise.await(
-				this.commandHandler.executeCommand(Commands.queue_details, { queueName: queue.name }),
-			) as IVoipConnectorResult;
-			this.logger.debug({ msg: 'API = voip/queues.getCallWaitingInQueuesForThisExtension queue details = ', result: queueDetails });
-			if (!(queueDetails.result as unknown as IQueueDetails).members) {
+		const queueSummary = (await this.commandHandler.executeCommand(Commands.queue_summary)) as IVoipConnectorResult;
```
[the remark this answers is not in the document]
amolghode1981: Queue membership wont change too often.I can imagine it to be changing (A safest measure) once a day. In reality it will not change for weeks or months once the initial configuration is done. But Queue summary and queue details also contain some other more dynamic information which we will need. e.g available agents, talk time for the queue, longest hold time. There are few ways I think we can use for optimisation
1. Memoize the part that does not change. Change the APIs in such a way that if slow changing information is asked for, we give it from the memory.
2. We provide a different set of APIs for getting dynamically changing parameter. (Same as above but slightly altered.)

I think we should keep it as is for MVP and then look at the optimisation part. This will also help gathering the customer data on the usage. Furthermore, we should configure the memoization parameters per command.

app/lib/server/functions/getFullUserData.js
amolghode1981: If the client wants to login to SIP, it is going to query this property. So it is ok if we pass it on.

server/services/voip/connector/asterisk/CommandHandler.ts
amolghode1981: Fixed.

server/services/voip/connector/asterisk/ami/ContinuousMonitor.ts
```diff
@@ -0,0 +1,234 @@
...
+	// Not sure if we should do it here.
+	private async getQueueDetails(queueName: string): Promise<IQueueDetails> {
+		const queue = new ACDQueue(Commands.queue_details.toString(), true);
+		queue.connection = this.connection;
+		const queueDetails = await queue.executeCommand({ queueName });
+		return queueDetails.result as unknown as IQueueDetails;
+	}
+
+	/*
+	private async findMemberUsers(queueName: string): Promise<string[]> {
```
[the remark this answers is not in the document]
amolghode1981: Fixed.

server/services/voip/lib/Helper.ts
amolghode1981: Fixed. But I have a question. If we are already reading it as a number, why do we want to convert it to number again by using Number()?

server/services/voip/lib/Helper.ts
amolghode1981: Fixed.

server/services/voip/service.ts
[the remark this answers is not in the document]
KevLehman: Yup agreed, should be after-mvp. But yeah, if won't change too often, a memoize would come handy, and we can have both, a memoized and a direct call depending on how we plan to use it 😬

server/services/voip/lib/Helper.ts
KevLehman: Oh, that's because we already have those values as strings on our db, so i didn't want to disrupt the work. Before merging into develop, this can be removed as clients will have those values already as numbers

KevLehman (approved):

KevLehman (approved):

PR #24204, opened 2022-01-18
[FIX] MAU when using micro services

sampaiodiego:


app/statistics/server/lib/SAUMonitor.ts
```diff
@@ -0,0 +1,342 @@
...
+			day: { $lte: yesterday.day },
+		};
+
+		await aggregates.dailySessionsOfYesterday(Sessions.col, yesterday).forEach(async (record) => {
+			await Sessions.updateOne(
+				{ _id: `${record.userId}-${record.year}-${record.month}-${record.day}` },
+				{ $set: record },
+				{ upsert: true },
+			);
+		});
```
ggazzo, line 333:
```suggestion
		await aggregates.dailySessionsOfYesterday(Sessions.col, yesterday).map(async (record) => Sessions.updateOne(
				{ _id: `${record.userId}-${record.year}-${record.month}-${record.day}` },
				{ $set: record },
				{ upsert: true },
			));
```

ee/server/services/ddp-streamer/Client.ts
```diff
@@ -22,21 +63,25 @@ export class Client extends EventEmitter {
...
-	constructor(public ws: WebSocket, public meteorClient = false) {
+	constructor(public ws: WebSocket, public meteorClient = false, req: IncomingMessage) {
 		super();
 
 		this.connection = {
+			id: this.session,
+			instanceId: server.id,
 			onClose: (fn): void => {
 				this.on('close', fn);
 			},
```
ggazzo, line 82: maybe we should check if this is intended, looks more a leak but not sure

ggazzo (changes requested):

ggazzo (approved):

PR #24207, opened 2022-01-18
[FIX] Remove redundant tooltip from setting inputs

try-catch-stack:
Settings Inputs that are used in the sections like the Administration settings  shows a tooltip on hovering over the label of that input (Refer the video below) which contains the _id for that input and is redundant considering label is already present . These tooltips hence should not be present as they provide no extra information.

Before:

https://user-images.githubusercontent.com/77742477/149975456-01fc608b-a470-4d23-811a-11fb6999e3e0.mp4

After:

https://user-images.githubusercontent.com/77742477/149975473-ab783193-a8f8-4227-a98e-229bbb7c547c.mp4

## Further comments

If they are present intentionally , hint or some sort of description should be shown instead of the _id.

### DIFF
--- a/client/views/admin/settings/inputs/AssetSettingInput.js
+++ b/client/views/admin/settings/inputs/AssetSettingInput.js
@@ -51,9 +51,7 @@ function AssetSettingInput({ _id, label, value = {}, asset, fileConstraints = {}
 
 	return (
 		<>
-			<Field.Label htmlFor={_id} title={_id}>
-				{label}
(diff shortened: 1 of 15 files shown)


try-catch-stack:
What do you think @dougfabris @sampaiodiego @debdutdeb ?

debdutdeb:
Closing this since this is fixing a non-existent issue/changing an expected behavior.

I'll reach out to you on open server :)

PR #24211, opened 2022-01-19
Regression: Fix handling of http requests in apps bridge

d-gubert:

Changes made during Meteor upgrade broke HTTP requests made in Rocket.Chat Apps


app/apps/server/bridges/http.ts
```diff
@@ -78,16 +78,28 @@ export class AppHttpBridge extends HttpBridge {
...
-				}
+			const body = Buffer.from(await response.arrayBuffer());
+
+			if (request.encoding === null) {
+				/**
+				 * So, here we have a problem. The property is not appropriately typed
+				 * in the Apps-Engine. Thus, even though it was  previously supported to
+				 * return it as a Buffer. Apps ended up needing to use that same type
+				 * assertion you see below.
+				 */
```
d-gubert, line 92: @sampaiodiego would like your thoughts here on how to best handle the situation

d-gubert: Something interesting to add is that changing the type in the Apps-Engine can be a breaking change 😬

cauefcr: This fixes my inability to get a binary response from an http request in an app.

d-gubert: @cauefcr You should be able to do that in previous versions of Rocket.Chat 🤔

cauefcr: I was using a workspace on the develop branch like a dummy. :see_no_evil:

app/apps/server/bridges/http.ts
```diff
@@ -78,16 +78,33 @@ export class AppHttpBridge extends HttpBridge {
...
 
-			if (request.hasOwnProperty('encoding')) {
-				if (request.encoding === null) {
-					result.content = Buffer.from(await response.arrayBuffer()).toString();
-				} else {
-					result.content = await response.text();
-				}
+			const body = Buffer.from(await response.arrayBuffer());
+
+			if (request.encoding === null) {
```
AllanPazRibeiro, line 86: maybe you can check if the encoding is not null by: 

`if (!request.encoding)`

AllanPazRibeiro (approved):

app/apps/server/bridges/http.ts
d-gubert: Here we need a strict comparison, as `undefined` for example, is a valid parameter to the `Buffer.toString` function

sampaiodiego (approved):

PR #24218, opened 2022-01-19
[FIX] Fixing the changing custom status behavior

AllanPazRibeiro:
This PR has the following fixes and additions:

- Fix on the custom status change when it is updated for both App and server
- Adds the custom status text on the user card tooltip


d-gubert (changes requested):

app/lib/server/functions/setStatus.ts
```diff
@@ -0,0 +1,24 @@
+import { UserStatus } from '../../../../definition/UserStatus';
+import { api } from '../../../../server/sdk/api';
+import { Users as UsersRaw } from '../../../models/server/raw';
+
+export const setUserStatus = async function (userId: string, status: UserStatus, statusText = ''): Promise<boolean> {
+	if (!userId) {
+		return false;
+	}
+
+	const user = await UsersRaw.findOneById(userId);
```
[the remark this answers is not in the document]
AllanPazRibeiro: You're right, fixing that

app/lib/server/functions/setStatus.ts
```diff
@@ -0,0 +1,24 @@
...
+
+	if (!user) {
+		return false;
+	}
+
+	await UsersRaw.updateStatusByAppId(user._id, status);
```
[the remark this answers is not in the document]
AllanPazRibeiro: got it

app/apps/server/bridges/users.ts
```diff
@@ -85,14 +98,14 @@ export class AppUserBridge extends UserBridge {
 
 	protected async update(user: IUser & { id: string }, fields: Partial<IUser>, appId: string): Promise<boolean> {
 		this.orch.debugLog(`The App ${appId} is updating a user`);
+		console.debug(`The App ${appId} is updating a user`);
```
d-gubert, line 101:
```suggestion
```

app/apps/server/bridges/users.ts
```diff
@@ -1,11 +1,24 @@
...
 import { Users } from '../../../models/server';
 import { Subscriptions, Users as UsersRaw } from '../../../models/server/raw';
 import { AppServerOrchestrator } from '../orchestrator';
+import { UserStatus } from '../../../../definition/UserStatus';
+import { api } from '../../../../server/sdk/api';
+
+const updateUserPresence = function (user: IUser, status = UserStatus.OFFLINE, statusText = ''): void {
+	const { id, username } = user;
+
+	Meteor.call('setUserStatus', status, statusText);
```
d-gubert, line 16: We have to avoid using `Meteor.call` to avoid depending on the framework as much as possible. Is there an alternative to calling this method, as it wasn't being called before?

app/apps/server/bridges/users.ts
d-gubert, line 101:
```suggestion
```

d-gubert (changes requested):

app/apps/server/bridges/users.ts
AllanPazRibeiro: got it and done

d-gubert (approved):

d-gubert (approved):

app/apps/server/bridges/users.ts
```diff
@@ -90,17 +91,19 @@ export class AppUserBridge extends UserBridge {
...
-			delete fields.statusText;
-		}
-
 		if (!Object.keys(fields).length) {
 			return true;
 		}
 
+		const { status } = fields;
+		delete fields.status;
+
```
ggazzo, line 100: const { status, ...updatedFields } = fields;

		await UsersRaw.update({ _id: user.id }, { $set: updatedFields });

		if (status) {
			UserPresence.setDefaultStatus(user.id, status);
		}

PR #24230, opened 2022-01-20
[FIX] Prevent overriding omni-room source props from apps

murtaza98:


KevLehman:
@murtaza98 , is this PR needed?

Issue #24238, opened 2022-01-20
[BUG] No Errors for Invalid Name or Password on Setup Wizard


samad-yar-khan:
### Description: 

Whenever we setup a Rocket.Chat server, we have to go through a setup wizard. While filling the Admin info form in setup wizard we are not able to submit the form if the Name and Password fields are empty, but there is no message showing us that an empty Name or Password is Invalid.

### Steps to reproduce:

1. Start a new Rocket.Chat server and enter the setup wizard.
2.  Fill User Name and Password. 
3. Click On Continue

### Expected behavior:

1. There should be an error message below the message name input field indicating that an empty name is invalid.
2. There should be a error message below the password field for invalid or weak passwords.

![image](https://user-images.githubusercontent.com/70485812/150405158-e36896ea-0b68-435a-b683-fbbfd4f1d0a9.png)

### Actual behavior:

1.  No error message is shown below empty Name and Password field, user is given no indication why they are not able to submit the form.

![error_in_rocket_chat](https://user-images.githubusercontent.com/70485812/150404771-9e56653e-c8e6-49d5-aaa5-1f3667c5e45b.JPG)

### Server Setup Information:

- Version of Rocket.Chat Server: 4.4.0-develop 
- Operating System: Ubuntu (Using Gitpod)
- Number of Running Instances: 1
- DB Replicaset Oplog: Enabled
- NodeJS Version: 14.18.2 - x64
- MongoDB Version: 4.4.4

### Client Setup Information

- Desktop App or Browser Version: Google Chrome Browser Version 97.0.4692.71
- Operating System: Windows 10

Ayush01729:
Hello @samad-yar-khan, I am Ayush and I am currently working on this issue.

PR #24240, opened 2022-01-20
[FIX] No Errors for Invalid Name or Password on Setup Wizard #24238

samad-yar-khan:
-  [x] I have read the Contributing Guide - https://github.com/RocketChat/Rocket.Chat/blob/develop/.github/CONTRIBUTING.md#contributing-to-rocketchat doc
  -  [x] I have signed the CLA - https://cla-assistant.io/RocketChat/Rocket.Chat
  -  [ ] I have added necessary documentation (if applicable)
  - [x] Any dependent changes have been merged and published in downstream modules

- I have added a new feature in the SetupWizard which shows an error if the Name field of the form is empty.
- I have added a password validator for the SetupWizard which displays an error in the following three conditions : I password field is empty , If the password is less than 8 lettes or if the password does not contain a number. This ensures that the password set by the user is strong.  

## Issue(s)

Fixes : #24238 

## Steps to test or reproduce

Mentioned in the Issue

## Before : 

error_in_rocket_chat

## After : 
image

lessThan8

atleast a number


lib/passwordValidator.ts
```diff
@@ -0,0 +1,13 @@
...
+
+	if (!password) {
+		errorMessage = 'Password cannot be empty';
+	} else if (password.length < 8) {
+		errorMessage = 'Password must be 8 characters long';
+	} else if (!passwordRegex.test(password)) {
+		errorMessage = 'Invalid password. Must contain one number';
+	}
+	return errorMessage;
+};
```
debdutdeb, line 13:
```suggestion
export const validatePassword = (password: string): string => {
	const passwordRegex = /(?=.*\d)(?=.*[!@#$%^&*])(?=.*[A-Z])(?=.*[a-z]).{8,}/;

	if (!password) {
		return 'Password cannot be empty';
	} else if (password.length < 8) {
		return 'Password must be at least 8 characters long';
	} else if (!passwordRegex.test(password)) {
		return 'Invalid password. Must contain one number, one capital letter, one small letter and one special character (!@#$%^&*)';
	}
	return '';
};
```

debdutdeb (changes requested):

lib/passwordValidator.ts
samad-yar-khan: @debdutdeb After Implementing the suggested changes we will get ES Lint errors saying that the else after a return statement is redundant. I will try to clean this up in some other way.

debdutdeb:
Closing, let's take this to the dm on open.

PR #24243, opened 2022-01-21
Regression: Remove extra call to `useOutsideClick` hook not following the function signature

tassoevan:

It migrates `client/sidebar/header/actions/Search` component to TypeScript and mitigates a invalid call to `Array.prototype.every`:

image


rique223 (approved):
LGTM

PR #24247, opened 2022-01-21
[FIX] Cleaning up some hard coding in the Voip code.

amolghode1981:
1. useVoipClient hook had hardcoded value for the extension. It was always using single extension regardless of what user is logging in. This has been fixed. Now it fetches the registration information based on the userId.
2. To support above, a new API has been added to app/api/server/v1/voip/extensions.ts which returns registrationinfo when userId is passed. Throws error if the agent is not associated with extension.
3. Removed user's personal account from app/lib/server/startup/settings.ts. Instead used organization specific account name for Voip Management Server.

## Issue(s)
1. When you have no agent-extension association, you would always see an error when you hover over "make agent available" phone button on top left of side bar. 

## Further comments
Testing Notes :
1. Test that proper registration information is returned when agent-extension association is done.
2. Test that proper error is shown when extension is not associated with the agent.
3. Make sure that you can change the settings's Voip Manager's user name. Make sure that system uses newly added username for connecting to asterisk.

### DIFF
--- a/client/providers/CallProvider/hooks/useVoipClient.ts
+++ b/client/providers/CallProvider/hooks/useVoipClient.ts
@@ -21,7 +21,7 @@ export const isUseVoipClientResultError = (result: UseVoipClientResult): result 
 	!!(result as UseVoipClientResultError).error;
 
 export const isUseVoipClientResultLoading = (result: UseVoipClientResult): result is UseVoipClientResultLoading =>
-	result === undefined || Object.keys(result).length === 0;
+	!result || !Object.keys(result).length;
 
 export const useVoipClient = (): UseVoipClientResult => {
 	const registrationInfo = useEndpoint('GET', 'connector.extension.getRegistrationInfoByUserId');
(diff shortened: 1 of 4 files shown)


client/providers/CallProvider/hooks/useVoipClient.ts
```diff
@@ -21,50 +21,46 @@ export const isUseVoipClientResultError = (result: UseVoipClientResult): result
 	!!(result as UseVoipClientResultError).error;
 
 export const isUseVoipClientResultLoading = (result: UseVoipClientResult): result is UseVoipClientResultLoading =>
-	Object.keys(result).length === 0;
+	result === undefined || Object.keys(result).length === 0;
```
KevLehman, line 24:
```suggestion
	!result || !Object.keys(result).length;
```

Both undefined and 0 are falsy values, so using the ! the code looks cleaner

KevLehman (changes requested):

client/providers/CallProvider/hooks/useVoipClient.ts
amolghode1981: fixed

KevLehman (approved):

PR #24252, opened 2022-01-21
[FIX] Avoid updating all rooms with visitor abandonment queries

KevLehman:

### DIFF
--- a/ee/app/livechat-enterprise/server/lib/Helper.js
+++ b/ee/app/livechat-enterprise/server/lib/Helper.js
@@ -171,13 +171,8 @@ export const updatePredictedVisitorAbandonment = () => {
 	}
 };
 
-let queueTimeout;
-
-settings.watch('Livechat_max_queue_wait_time', (value) => {
-	queueTimeout = value;
-});
-
 export const updateQueueInactivityTimeout = () => {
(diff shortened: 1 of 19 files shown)


ee/app/livechat-enterprise/server/lib/Helper.js
```diff
@@ -173,7 +173,7 @@ export const updatePredictedVisitorAbandonment = () => {
 
 let queueTimeout;
 
-settings.get('Livechat_max_queue_wait_time', (value) => {
+settings.watch('Livechat_max_queue_wait_time', (value) => {
```
sampaiodiego, line 176: 👀  good find..

may I ask you to actually remove this watcher and use `settings.get` whenever `queueTimeout` is being used? this is an old pattern I think we should get rid of

sampaiodiego (changes requested):

ee/app/livechat-enterprise/server/lib/Helper.js
KevLehman: Ah, didn't know it was an old pattern 👀 let me change them

KevLehman:
@sampaiodiego , ready for your eyes again.

sampaiodiego (approved):

PR #24256, opened 2022-01-21
[IMPROVE] Perform endpoint operations with less DB calls

KevLehman:

- Changed the way some endpoints did multiple operations to do those in a single batch instead of 1 by 1.
- Added missing indexes to models
- Converted some endpoint handlers to `async` to use normal `await` instead of Promise.await


app/livechat/server/api/lib/queue.js
```diff
@@ -6,7 +6,7 @@ export async function findQueueMetrics({ userId, agentId, includeOfflineAgents,
 		throw new Error('error-not-authorized');
```
KevLehman, line 6: This file has an optimization

app/livechat/server/api/v1/customField.js
```diff
@@ -53,16 +53,12 @@ API.v1.addRoute('livechat/custom.fields', {
 			throw new Meteor.Error('invalid-token');
```
KevLehman, line 53: This file has an optimization

app/livechat/server/api/v1/visitor.ts
```diff
@@ -47,17 +47,25 @@ API.v1.addRoute('livechat/visitor', {
 		});
```
KevLehman, line 47: This file has an optimization

app/livechat/server/lib/Livechat.js
```diff
@@ -522,6 +522,31 @@ export const Livechat = {
 		return LivechatVisitors.updateLivechatDataByToken(token, key, value, overwrite);
```
KevLehman, line 522: This file has an optimization

app/models/server/models/LivechatAgentActivity.js
```diff
@@ -4,9 +4,8 @@ export class LivechatAgentActivity extends Base {
 	constructor(...args) {
```
KevLehman, line 4: This file has an optimization

app/models/server/models/LivechatDepartment.js
```diff
@@ -16,6 +16,7 @@ export class LivechatDepartment extends Base {
 			numAgents: 1,
```
KevLehman, line 16: This file has an optimization

app/models/server/models/LivechatDepartmentAgents.js
```diff
@@ -11,9 +11,9 @@ export class LivechatDepartmentAgents extends Base {
 		super('livechat_department_agents');
```
KevLehman, line 11: This file has an optimization

app/models/server/models/LivechatRooms.js
```diff
@@ -22,6 +22,7 @@ export class LivechatRooms extends Base {
 		this.tryEnsureIndex({ 'v._id': 1 }, { sparse: true });
```
KevLehman, line 22: This file has an optimization

app/models/server/models/LivechatVisitors.js
```diff
@@ -11,6 +11,7 @@ export class LivechatVisitors extends Base {
```
KevLehman, line 11: This file has an optimization

app/models/server/raw/LivechatRooms.js
```diff
@@ -84,13 +84,23 @@ export class LivechatRoomsRaw extends BaseRaw {
 			firstParams.push(matchUsers);
```
KevLehman, line 84: This file has an optimization

KevLehman:
I marked which files have relevant updates and not just `async` additions

app/livechat/server/lib/Contacts.js
```diff
@@ -55,7 +55,7 @@ export const Contacts = {
 
 		LivechatVisitors.updateById(contactId, updateUser);
 
-		const rooms = LivechatRooms.findByVisitorId(contactId).fetch();
+		const rooms = LivechatRooms.findOpenByVisitorId(contactId).fetch();
```
cauefcr, line 58: Only open rooms receive visitor registration?

app/models/server/models/LivechatVisitors.js
```diff
@@ -81,6 +82,22 @@ export class LivechatVisitors extends Base {
...
+		const user = this.findOne(query, { fields: { livechatData: 1 } });
+		const updates = {};
+
+		fields.map((f) => {
+			if (!f.overwrite && user.livechatData && typeof user.livechatData[f.key] !== 'undefined') {
+				return null;
+			}
+			updates[`livechatData.${f.key}`] = f.value;
+			return null;
+		});
```
murtaza98, line 96:
```suggestion
		fields.forEach((f) => {
			if (!f.overwrite && user.livechatData && typeof user.livechatData[f.key] !== 'undefined') {
				return;
			}
			updates[`livechatData.${f.key}`] = f.value;
		});
```

murtaza98: Since we're not converting/mapping an array here, we can use foreach loop instead

app/models/server/models/LivechatRooms.js
```diff
@@ -81,6 +82,22 @@ export class LivechatRooms extends Base {
...
+		const room = this.findOne(query, { fields: { livechatData: 1 } });
+		const updates = {};
+
+		fields.map((f) => {
+			if (!f.overwrite && room.livechatData && typeof room.livechatData[f.key] !== 'undefined') {
+				return null;
+			}
+			updates[`livechatData.${f.key}`] = f.value;
+			return null;
+		});
```
murtaza98, line 96:
```suggestion
		fields.forEach((f) => {
			if (!f.overwrite && room.livechatData && typeof room.livechatData[f.key] !== 'undefined') {
				return;
			}
			updates[`livechatData.${f.key}`] = f.value;
		});
```

murtaza98: Since we're not converting/mapping an array here, we can use `forEach` loop instead.

app/livechat/server/lib/Livechat.js
```diff
@@ -522,6 +522,31 @@ export const Livechat = {
...
 	},
 
+	setMultipleCustomFields(token, fields = []) {
+		check(token, String);
+		const customfieldKeys = fields.map((f) => f.key);
+		const customFields = LivechatCustomField.findByIds(customfieldKeys, { fields: { _id: 1, scope: 1 } }).fetch();
```
murtaza98, line 528:
```suggestion
		const customFieldKeys = fields.map((f) => f.key);
		const customFields = LivechatCustomField.findByIds(customFieldKeys, { fields: { _id: 1, scope: 1 } }).fetch();
```

murtaza98: Minor typo

app/models/server/models/LivechatAgentActivity.js
```diff
@@ -4,9 +4,8 @@ export class LivechatAgentActivity extends Base {
...
 		super(...args);
 
-		this.tryEnsureIndex({ date: 1 });
```
murtaza98, line 7: I think we still need this index since I found it getting used [here](https://github.com/RocketChat/Rocket.Chat/blob/ee120705befb1f1be89bed7eec2b106673ffd2ea/app/models/server/raw/LivechatAgentActivity.ts#L78). Let me know if I'm missing something here

app/models/server/models/LivechatRooms.js
```diff
@@ -22,6 +22,7 @@ export class LivechatRooms extends Base {
 		this.tryEnsureIndex({ 'v._id': 1 }, { sparse: true });
 		this.tryEnsureIndex({ t: 1, departmentId: 1, closedAt: 1 }, { partialFilterExpression: { closedAt: { $exists: true } } });
 		this.tryEnsureIndex({ source: 1 }, { sparse: true });
+		this.tryEnsureIndex({ 'open': 1, 'v.token': 1 });
```
murtaza98, line 25: Do we need to add sparse options here since only livechat rooms will have this property set?

app/livechat/server/lib/Contacts.js
KevLehman: No, but we only care about open rooms when updating contact info. The old query would have updated all rooms, even if they were months old 😬

app/models/server/models/LivechatRooms.js
KevLehman: Could be 👀 let me check

murtaza98 (dismissed):

app/livechat/server/api/v1/contact.js
```diff
@@ -28,12 +28,12 @@ API.v1.addRoute(
...
 			}
 		},
-		get() {
+		async get() {
 			check(this.queryParams, {
 				contactId: String,
 			});
 
-			const contact = Promise.await(LivechatVisitors.findOneById(this.queryParams.contactId));
+			const contact = await LivechatVisitors.findOneById(this.queryParams.contactId);
```
sampaiodiego, line 36: from what I see, `LivechatVisitors` is not a raw model, so it was not even needed to `await` here 😬

app/models/server/models/LivechatVisitors.js
```diff
@@ -11,6 +11,7 @@ export class LivechatVisitors extends Base {
 
 		this.tryEnsureIndex({ token: 1 });
 		this.tryEnsureIndex({ 'phone.phoneNumber': 1 }, { sparse: true });
+		this.tryEnsureIndex({ 'open': 1, 'v.token': 1 }, { sparse: true });
```
sampaiodiego, line 14: I couldn't find `v.token` on this model, is this index correct?

app/models/server/models/LivechatRooms.js
```diff
@@ -81,6 +82,21 @@ export class LivechatRooms extends Base {
 		return this.update(query, update);
 	}
 
+	updateBatchDataByToken(token, fields) {
+		const query = { 'v.token': token, 'open': true };
```
sampaiodiego, line 86: changing the query to use an existing index:
```suggestion
		const query = { t: 'l', 'v.token': token, 'open': true };
```

app/models/server/models/LivechatRooms.js
```diff
@@ -22,6 +22,7 @@ export class LivechatRooms extends Base {
 		this.tryEnsureIndex({ 'v._id': 1 }, { sparse: true });
 		this.tryEnsureIndex({ t: 1, departmentId: 1, closedAt: 1 }, { partialFilterExpression: { closedAt: { $exists: true } } });
 		this.tryEnsureIndex({ source: 1 }, { sparse: true });
+		this.tryEnsureIndex({ 'open': 1, 'v.token': 1 }, { sparse: true });
```
sampaiodiego, line 25: looking the existing queries, I think a better index would be:
```
t: 1, 
'v.token': 1,
open: 1,
'email.thread' :1,
```
this could be the only index with `v.token` and it probably fit all queries (except the one at `updateDataByToken` that would require a change to include `t: 'l'` filter as well)

app/livechat/server/api/v1/contact.js
KevLehman: Good catch 👀 this was not a raw model and we were "awaiting" it lol. Let me fix it

app/models/server/models/LivechatDepartmentAgents.js
```diff
@@ -11,9 +11,9 @@ export class LivechatDepartmentAgents extends Base {
...
 
 		this.tryEnsureIndex({ departmentId: 1 });
-		this.tryEnsureIndex({ departmentEnabled: 1 });
-		this.tryEnsureIndex({ agentId: 1 });
-		this.tryEnsureIndex({ username: 1 });
+		this.tryEnsureIndex({ departmentEnabled: 1, agentId: 1 });
+		this.tryEnsureIndex({ username: 1, departmentId: 1 });
```
sampaiodiego, line 15: instead of creating a new index, I'd change the existing one with `departmentId: 1` to a compound index with `username`.

app/models/server/models/LivechatDepartmentAgents.js
```diff
@@ -11,9 +11,9 @@ export class LivechatDepartmentAgents extends Base {
 		super('livechat_department_agents');
 
 		this.tryEnsureIndex({ departmentId: 1 });
-		this.tryEnsureIndex({ departmentEnabled: 1 });
-		this.tryEnsureIndex({ agentId: 1 });
-		this.tryEnsureIndex({ username: 1 });
+		this.tryEnsureIndex({ departmentEnabled: 1, agentId: 1 });
```
sampaiodiego, line 14: don't forget the migration 😬

app/models/server/models/LivechatAgentActivity.js
```diff
@@ -4,9 +4,9 @@ export class LivechatAgentActivity extends Base {
...
 		this.tryEnsureIndex({ agentId: 1, date: 1 }, { unique: true });
-		this.tryEnsureIndex({ agentId: 1 });
+		this.tryEnsureIndex({ lastStoppedAt: 1 });
```
sampaiodiego, line 8: I found only with `lastStoppedAt: { $exists: false }`, if this is really the query using `lastStoppedAt` this index will not help

app/models/server/models/LivechatAgentActivity.js
```diff
@@ -4,9 +4,9 @@ export class LivechatAgentActivity extends Base {
 	constructor(...args) {
 		super(...args);
 
-		this.tryEnsureIndex({ date: 1 });
 		this.tryEnsureIndex({ agentId: 1, date: 1 }, { unique: true });
-		this.tryEnsureIndex({ agentId: 1 });
```
sampaiodiego, line 9: pls remove this index in a migration as well

app/models/server/models/LivechatRooms.js
```diff
@@ -333,6 +349,16 @@ export class LivechatRooms extends Base {
 		return this.find(query);
 	}
 
+	findOpenByVisitorId(visitorId) {
+		const query = {
+			't': 'l',
+			'v._id': visitorId,
```
sampaiodiego, line 355: not sure if that would help, but is it possible to change to use `v.token` instead of `v._id`? trying to use the best index here =)

sampaiodiego (changes requested):

app/livechat/server/api/v1/contact.js
sampaiodiego: yep.. that is not an issue though.. not sure how harmful it is to let the `await` there

KevLehman: I don't think it affects 🤔 there are other places where we "promise.await" on meteor models lol. Maybe the biggest issue is the overhead on node 🤷‍♂️

app/models/server/models/LivechatRooms.js
KevLehman: The thing with adding "t" as part of the index is that "t" doesn't change, and "t" holds very few possible values. With that, t won't discriminate so much to be part of an index imho 👀 

do you think it's worth it?

(about the "email.thread", yep, I can add it there)

app/models/server/models/LivechatRooms.js
KevLehman: There's no index on t+v.token+open 🤔

app/models/server/models/LivechatDepartmentAgents.js
KevLehman: Oh, but I was trying to use the prefixes here:
- username + departmentId will cover queries on both and on username
- departmentEnabled + agentId will cover queries on both
- agentId + departmentId will cover both and agentId only
- departmentId will cover only the ones using departmentId

There's one query that only uses username, so the index may be worth there

We do need some telemetry on index usage on real production envs tho :(

app/models/server/models/LivechatDepartmentAgents.js
KevLehman: 😬 i'm scared of index migrations now lol

app/models/server/models/LivechatAgentActivity.js
KevLehman: Yup, that's the only one right now. I'm planning to change some queries around the feature and this index will come handy 🙈 so that's why it's here now

app/models/server/models/LivechatDepartmentAgents.js
sampaiodiego: maybe my comment was not scoped enough 😬 

I'm talking about only regarding the new index `{ username: 1, departmentId: 1 }`.. because there is no query using only `username` , if you have a compound index with `departmentId` + `username` , it will alone cover both cases:

- a find only using `departmentId`
- a find using both `departmentId` and `username`

since there is already a index for `departmentId` only, I suggested then to change that one to include `username` as well.

app/models/server/models/LivechatRooms.js
sampaiodiego: I'm suggesting one in the comment below 🙈

app/models/server/models/LivechatRooms.js
KevLehman: Good catch, I reused one function we already had, but we can use another that fetch using token

app/models/server/models/LivechatRooms.js
sampaiodiego: my suggestion to add `t` was because I've seen almost all queries using it as well. but yeah, maybe that is not that useful on the index and mongo could already use an index without it.

the overall goal of my suggestion is to have **less** indexes covering **more** queries.. so with change we would end up with 3 indexes with `v.token`:

```
		this.tryEnsureIndex({ 'v.token': 1 }, { sparse: true });
		this.tryEnsureIndex({ 'v.token': 1, 'email.thread': 1 }, { sparse: true });
		this.tryEnsureIndex({ 'open': 1, 'v.token': 1 }, { sparse: true });
```

looking the queries, I think a single index can handle all queries:

```
this.tryEnsureIndex({
    t: 1, // as per messages above, maybe not needed
    'v.token': 1, // the main field of the index (present in all queries)
    'email.thread': 1,
    open: 1, // this last two I'm not sure the "order", need to understand which one happens more
}, { sparse: true });
```

sampaiodiego:
fix conflicts pls? 🙈

KevLehman:
Fixed 😬

KevLehman:
Redoing this is easier at this point lol

PR #24259, opened 2022-01-22
Regression: Fix Alpine release tag

sampaiodiego:

The code was still relying in `GIT_TAG` that was removed in latest changes.

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -536,7 +536,9 @@ jobs:
       run: |
         cd /tmp/build
 
-        DOCKER_TAG="${GITHUB_REF#*tags/}"
+        GIT_TAG="${GITHUB_REF#*tags/}"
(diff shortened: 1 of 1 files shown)


debdutdeb:
Why not use `GITHUB_REF_NAME` or `github.ref_name` instead? I see `GIT_TAG` being used in a lot of places.

sampaiodiego:
> Why not use `GITHUB_REF_NAME` or `github.ref_name` instead? I see `GIT_TAG` being used in a lot of places.

I think the concern was that `GITHUB_REF_NAME` can be also the branch name, but since we have the condition `github.event_name == 'release'` it should be safe using it. 

maybe we can switch.. not an issue.

debdutdeb:
I guess to make sure `github.event.release.tag_name` should also work.

KevLehman (approved):

PR #24276, opened 2022-01-24
Regression: Fix incompatibility of apps http requests

d-gubert:

HTTP GET and HEAD requests made with an empty object as `data` were breaking, as the bridge converted this to the request's body as `'{}'` but meteor's new lib doesn't allow for body content on either of this request methods.

To maintain compatibility, we forced an empty body whenever we have a GET or HEAD request. This was probably the case previously, with the body of requests made with this methods being ignored either before being sent or in the third party server receiving the request

### DIFF
--- a/app/apps/server/bridges/http.ts
+++ b/app/apps/server/bridges/http.ts
@@ -6,6 +6,8 @@ import { IHttpBridgeRequestInfo } from '@rocket.chat/apps-engine/server/bridges'
 import { AppServerOrchestrator } from '../orchestrator';
 import { getUnsafeAgent } from '../../../../server/lib/getUnsafeAgent';
 
+const isGetOrHead = (method: string): boolean => ['GET', 'HEAD'].includes(method.toUpperCase());
+
 export class AppHttpBridge extends HttpBridge {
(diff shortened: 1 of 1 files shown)


app/apps/server/bridges/http.ts
```diff
@@ -6,6 +6,8 @@ import { IHttpBridgeRequestInfo } from '@rocket.chat/apps-engine/server/bridges'
 import { AppServerOrchestrator } from '../orchestrator';
 import { getUnsafeAgent } from '../../../../server/lib/getUnsafeAgent';
 
+const isGetOrHead = (method: string): boolean => ['GET', 'HEAD'].includes(method.toUpperCase());
```
sampaiodiego, line 9: it hurts me deep down in my heart having to upper case `method` every time 🙈

d-gubert: Calling a method every time is bad but creating an array every time doesn't raise any flags? 😛

sampaiodiego: yep.. I agree hence I didn't ask for changes.. I was just expressing my feeling :P

sampaiodiego (approved):

PR #24289, opened 2022-01-25
Regression: Create migration to fix index issue at boot

KevLehman:


sampaiodiego (changes requested):
pls revert b983476602b676015b1d4ed5f094461781df6187

sampaiodiego (approved):

PR #24297, opened 2022-01-26
[IMPROVE] OTR Message

albuquerquefabio:
- Now we are sending OTR messages through streamer without saving them on the database.
- The OTR Key now is updated without ending OTR.
- Old OTR methods were removed.


app/otr/client/rocketchat.otr.js
```diff
@@ -94,7 +96,8 @@ Meteor.startup(function () {
 				}
 				if (data.userId !== Meteor.userId()) {
 					return otrRoom.encryptText(ack).then((ack) => {
-						Meteor.call('updateOTRAck', message._id, ack);
+						message.msg = message.msg !== msg ? msg : message.msg;
```
tassoevan, line 99: This whole `Meteor.startup()` call must be moved to an isolated module under `client/startup` for the sake of standardization.

xxxmadxxx:
Captura de Tela 2022-03-29 às 16 26 01

I'm checking something related to this feature and I notice that when we send a message we're using websockets but also we're sending a HTTP POST, there's any reason for this request?

tassoevan (dismissed):

PR #24305, opened 2022-01-26
[FIX] Prevent Apps Bridge to remove visitor status from room

KevLehman:


tiagoevanp (commented):
Passing through the tests! 😃 
Waiting on @d-gubert review... 🙈

d-gubert:
@KevLehman the `status` property is already accepted in the apps-engine side?

KevLehman:
(I hope I understood the question 😬, but will try)

Yep, they can use the `status` prop, and they can pass it as a param. What was happening is that, some apps trigger a "room update" after doing some initial modifications (like, setting custom fields). The app at that point can update the full room object, even if the update was just a key 👀 . But, the problem was that apps-engine was returning a "smaller" `room.v` prop with just `_id, username, token` even if the `status` was already set (by the default values on omnichannel side).

d-gubert (approved):

PR #24310, opened 2022-01-27
[FIX] Implement client errors on ddp-streamer

sampaiodiego:
## Further comments

Created as draft to discuss about a few things:

- should this PR go through all errors thrown from existing services and change them to use one of the new error types?
- open a discussion regarding the new error type names, currently `MethodError` and `ClientSafeError`


ee/server/services/account/Account.ts
```diff
@@ -56,12 +57,21 @@ const loginViaResume = async (resume: string): Promise<false | ILoginResult> =>
 	}
 
 	const { when } = user.services?.resume?.loginTokens?.find((token) => token.hashedToken === hashedToken) || {};
+	if (!when) {
```
rodrigok, line 60: Are you sure this is the right behavior? I thought tokens without when would never expire (e.g. application tokens)

ee/server/broker.ts
```diff
@@ -288,6 +289,35 @@ const {
...
+
+		if (isClientSafe) {
+			return new ClientSafeError(message, reason, details);
+		}
+
+		return undefined;
+	}
+
+	extractPlainError(err: Error | ClientSafeError): Errors.PlainMoleculerError {
+		console.log('extractPlainError ->', err);
```
rodrigok, line 308:
```suggestion
```

ee/server/services/account/Account.ts
```diff
@@ -56,12 +57,21 @@ const loginViaResume = async (resume: string): Promise<false | ILoginResult> =>
...
+	if (new Date() >= tokenExpires) {
+		throw new ClientSafeError(403, 'Your session has expired. Please log in again.');
+	}
 
 	return {
 		uid: user._id,
 		token: resume,
-		hashedToken,
-		tokenExpires: when ? _tokenExpiration(when) : undefined,
+		hashedToken, // TODO should we return the hashed token? Meteor does not.
```
rodrigok, line 73: Does it affect any other code if not return here? If not we should not return.

ee/server/services/ddp-streamer/Server.ts
```diff
@@ -67,18 +85,34 @@ export class Server extends EventEmitter {
...
 			this.nosub(client, packet, error.toString());
+
+			if (error instanceof ClientSafeError) {
+				return this.nosub(client, packet, error.toJSON());
+			}
+
+			// default errors are logged to the console and redacted from the client
+			logger.error({ msg: 'Subscribe error', error });
+
+			return this.nosub(client, packet, {
```
rodrigok, line 109: Why not pass the error instance and instead of accepting any record to return as error?

ee/server/services/ddp-streamer/Server.ts
```diff
@@ -67,18 +85,34 @@ export class Server extends EventEmitter {
 	async subscribe(client: Client, packet: IPacket): Promise<void> {
 		try {
 			if (!this._subscriptions.has(packet.name)) {
-				throw new Error(`Subscription '${packet.name}' doesn't exist`);
+				// TODO should not be MethorError
```
rodrigok, line 88: Why not?

rodrigok (changes requested):

ee/server/services/account/Account.ts
sampaiodiego: I don't think it affects anything.. as stated, Meteor doesn't return that as well, so our clients definitely don't use it.. just need to make sure we don't use it **before** returning to client.

ee/server/services/ddp-streamer/Server.ts
sampaiodiego: I think the name is misleading.. this is a *subscription* , so using a *MethodError* seems wrong.. 

In my local changes I've removed both error types and kept just a single one called `MeteorError`, to match the `errorType` property.. but I was not sure about that as well.

ee/server/services/account/Account.ts
sampaiodiego: I was not, and you're definitely right.. 👍  I'll revert this change

ee/server/services/ddp-streamer/Server.ts
sampaiodiego: good idea!

ee/server/services/account/Account.ts
rodrigok: That was my question :) if it's used somewhere else here. If not I think we should not return

ee/server/services/ddp-streamer/Server.ts
```diff
@@ -34,24 +38,32 @@ export class Server extends EventEmitter {
...
-			return this.result(client, packet, null, error.toString());
+		} catch (err: unknown) {
+			if (err instanceof MeteorError) {
+				return this.result(client, packet, null, err);
+			}
+
+			// default errors are logged to the console and redacted from the client
+			logger.error({ msg: 'Method call error', err });
+
+			return this.result(client, packet, null, new MeteorError(500, 'Internal server error'));
```
rodrigok, line 66: I think this check should be inside the result function (or in a "send" function called by it), checking if the error is client save to return it or to print and return 500

sampaiodiego: is your goal to remove the duplicated error handling? if that is the case, I can create a helper to do that, so the result would be something like:

```ts
catch (err: unknown) {
  return this.result(client, packet, null, handleInternalException(err));
}
// ...
catch (err: unknown) {
  return this.nosub(client, packet, handleInternalException(err));
}
```

so the logic to either log the error or return would be only into `handleInternalException`

ee/server/broker.ts
```diff
@@ -288,6 +289,30 @@ const {
 	SKIP_PROCESS_EVENT_REGISTRATION = 'true',
 } = process.env;
 
+class CustomRegenerator extends Errors.Regenerator {
+	restoreCustomError(plainError: any): Error | undefined {
+		const { message, reason, details, errorType } = plainError;
+
+		if (errorType === 'Meteor.Error') {
+			return new MeteorError(message, reason, details);
```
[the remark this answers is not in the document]
sampaiodiego: it is a "constant" of the `MeteorError` class, so need to transfer over the wire.. it is restored when a new instance of `MeteorError` is created.

ee/server/services/ddp-streamer/Server.ts
sampaiodiego: went ahead and implemented `handleInternalException`

ee/server/services/account/Account.ts
sampaiodiego: it is used on the logout method.. so I removed the TODO 😉

PR #24311, opened 2022-01-27
Chore: Convert to typescript the slash commands invite files

eduardofcabrera:

Convert to typescript the slash commands invite files


ostjen (approved):
lgtm

PR #24319, opened 2022-01-27
[NEW] Create server event when a setting is updated

KevLehman:

### DIFF
--- a/apps/meteor/app/api/server/v1/settings.ts
+++ b/apps/meteor/app/api/server/v1/settings.ts
@@ -8,8 +8,9 @@ import {
 	isSettingsUpdatePropsActions,
 	isSettingsUpdatePropsColor,
 } from '@rocket.chat/rest-typings';
+import farmhash from 'farmhash';
 
-import { Settings } from '../../../models/server/raw';
+import { Settings, ServerEvents } from '../../../models/server/raw';
 import { hasPermission } from '../../../authorization/server';
 import { API, ResultFor } from '../api';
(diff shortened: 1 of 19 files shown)


lib/settingsChangeWatcher.ts
```diff
@@ -0,0 +1,37 @@
...
+	const currentSetting = settings.get(settingId);
+	if (currentSetting === targetValue) {
+		logger.debug(`Cannot store change metadata for ${settingId}. Setting didn't change`);
+		return;
+	}
+	const event = {
+		t: IServerEventType.SETTING_MODIFIED,
+		ts: new Date(),
+		u: {
+			_id: uid,
```
murtaza98, line 20: Since we're just storing the id of the user, so if this user is deleted from the server, then it means we won't be able to track back to the updater.
So, wanted to confirm if this is the intended behaviour? I feel this info might be required for audit purposes...Would it be okay if we also put the username or name within this user object?  .... maybe we can even create a setting for this to make it configurable

app/models/server/raw/Settings.ts
```diff
@@ -57,7 +63,10 @@ export class SettingsRaw extends BaseRaw<ISetting> {
 				value,
 			},
 		};
+		console.log('--------------------------');
+		console.log(_id, value, _editor, uid);
```
murtaza98, line 67:
```suggestion
```

KevLehman:
Closing as this is a conflict more than a PR right now. I'll re-create it

PR #24322, opened 2022-01-27
Regression: Fix OmnichannelAppSourceRoomIcon sizes

tiagoevanp:


KevLehman:
So, the actual size is OK (16x16) but the icon still looks bigger than the others 🤔 
image

tiagoevanp:
> So, the actual size is OK (16x16) but the icon still looks bigger than the others 🤔 image
> 
> I believe it's not because of the code itself, but icons like the "email" one have a padding surrounding it 👀 not sure if that would be a problem cc @renatobecker @casalsgh

This is because of SVG usage against PNG images designed by our Figma icons! Design people create icons with a safe padding on it, we get this image icons with this safe squared area on it.

image

murtaza98 (approved):

PR #20594, opened 2021-02-04
[FIX] Channel name of messages in global search

RonLek:
## Proposed changes (including videos or screenshots)

On searching for a message in global search the name of the current channel was being shown even if the message belonged to different channels.

### Initially

Screenshot from 2021-02-04 12-01-59

### Now

Screenshot from 2021-02-04 12-03-08

## Issue(s)

Fixes #20574 

## Steps to test or reproduce

1. Enable Global Search (under Settings > Search)
2. Go back to the UI, and select a room
3. Search for a string that is located in a room other than the current room.

Issue #21659, opened 2021-04-20
Global Search with unchecked box


xi67:
### Description:

When you click on the conversation search (the right one), the global search is unchecked but the search is global.
The global search box does not reflect the actual search.

### Steps to reproduce:

1. Click on the conversation search button
2. add some text
3. the search is global without checked it

### Expected behavior:

The box is checked by default as the default search is global or let the box unchecked but change the actual default search.

### Server Setup Information:

- Version of Rocket.Chat Server: 3.13.2
- Operating System: Debian
- Deployment Method: tar
- Number of Running Instances: 1
- DB Replicaset Oplog: wiredTiger
- NodeJS Version: 12.22.1
- MongoDB Version: 4.2.13

### Client Setup Information

- Desktop App or Browser Version: App 3.1.1 and bowser
- Operating System: Windows 10

johncrisp:
Hi and thanks for reporting this.

Can you try testing on https://open.rocket.chat as well just to see if it still exists in latest?

Thanks.

aaa-a-a-a-aaa:
Global search appears to be disabled on open.rocket.chat so I don't see how that would be possible.

danielfl:
**Workaround**: check and then uncheck the Global search checkbox.

scrivna:
I can confirm that we also have this issue on our instance running 3.15.0 (current latest version)

ankar84:
Hey, @johncrisp 
I can confirm that bug on latest 3.15.0 too
It's a front-end bug I guess.

ankar84:
And still an issue in 3.16.0
![image](https://user-images.githubusercontent.com/4023037/123749915-9f201d80-d8be-11eb-9405-09b6c262ed60.png)

xi67:
Stil present on 3.16.2

ankar84:
Still an issue in 3.16.3

debdutdeb:
Present in 3.17.0

debdutdeb:
We had a meeting with the product team today. Good news, this should be fixed in the next September or October release. 

Thank you everyone here for keeping the issue updated 🙌🏼

noobbbbb:
What's This?
https://github.com/RocketChat/Rocket.Chat/pull/20594

ankar84:
> We had a meeting with the product team today. Good news, this should be fixed in the next September or October release.
> 
> Thank you everyone here for keeping the issue updated 🙌🏼

[...]

PR #24326, opened 2022-01-28
[FIX] Global Search checkbox inital state

dpanchuk:
This is fix for search checkbox.
The problem was in giving "global search enabled" option value to "search all" payload by default.
#21659

### DIFF
--- a/app/search/client/provider/result.js
+++ b/app/search/client/provider/result.js
@@ -79,7 +79,7 @@ Template.DefaultSearchResultTemplate.onCreated(function () {
 
 	// global search
 	this.globalSearchEnabled = this.data.settings.GlobalSearchEnabled;
-	this.data.parentPayload.searchAll = this.globalSearchEnabled;
+	this.data.parentPayload.searchAll = (!!this.globalSearchEnabled && !!this.data.parentPayload.searchAll);
 
 	this.hasMore = new ReactiveVar(true);
 


glebPo:
👍

serjotaylor:
finally!!!

ankar84:
@ggazzo please take a look at that fix
cc @dudanogueira @debdutdeb

ankar84:
There is another community fix https://github.com/RocketChat/Rocket.Chat/pull/20594

app/search/client/provider/result.js
```diff
@@ -79,7 +79,7 @@ Template.DefaultSearchResultTemplate.onCreated(function () {
 
 	// global search
 	this.globalSearchEnabled = this.data.settings.GlobalSearchEnabled;
-	this.data.parentPayload.searchAll = this.globalSearchEnabled;
+	this.data.parentPayload.searchAll = (!!this.globalSearchEnabled && !!this.data.parentPayload.searchAll);
```
debdutdeb, line 82:
```suggestion
	this.data.parentPayload.searchAll = false;
```
You can just set it to false

debdutdeb (changes requested):
HI @dpanchuk thanks for this PR!

Initially I was thinking about changing the element to reflect what actually is being done (i.e. keep it checked instead of unchecked if global search is enabled).

BUT, I like the idea of not defaulting to that behaviour but leave it as a toggle for the users to switch manually, who actually is meaning to use it.

dudanogueira:
Hi! Thanks for your contribution. We will close this PR in favor of 
https://github.com/RocketChat/Rocket.Chat/pull/25568

PR #24330, opened 2022-01-28
Chore: Convert to typescript the open slash commands files

eduardofcabrera:

Convert to typescript the open slash commands files


app/slashcommands-open/client/client.ts
```diff
@@ -0,0 +1,63 @@
...
+	let room = params.trim();
+	const typeStr = room[0];
+	let type: string[] | null;
+	if (typeStr === '#') {
+		type = ['c', 'p'];
+	} else if (typeStr === '@') {
+		type = ['d'];
+	} else {
+		type = null;
+	}
```
ostjen, line 18: why did you removed the dict?

ostjen, line 18: why did you removed the dict?

ostjen, line 18: why did you removed the dict?

ostjen, line 18: why did you remove the dict?

ostjen (changes requested):

app/slashcommands-open/client/client.ts
eduardofcabrera: Because it was giving type erros, so I would need to create an interface to the dict. I prefered to just create an if-else to the cases.

eduardofcabrera: It was giving that string can't be used to index the dict.

app/slashcommands-open/client/client.ts
```diff
@@ -1,26 +1,22 @@
...
-		return;
-	}
-
 	let room = params.trim();
 	const type = dict[room[0]];
 	room = room.replace(/#|@/, '');
-
 	const query = {
 		name: room,
+		t: {},
```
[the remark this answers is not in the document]
eduardofcabrera: on line 27, requires the t property of the query, and the typescript was complaining about typing, so I just set it to update later.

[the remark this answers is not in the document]
ostjen: ok, but wouldn't be better to do the line 27 assignment here, instead of initializing with an empty object and reassigning after?

ostjen (changes requested):

app/slashcommands-open/client/client.ts
```diff
@@ -0,0 +1,55 @@
...
+import { Subscriptions, ChatSubscription } from '../../models/client';
+
+function Open(_command: 'open', params: string, _item: Record<string, string>): void {
+	const dict: Record<string, string[]> = {
+		'#': ['c', 'p'],
+		'@': ['d'],
+	};
+
+	let room = params.trim();
+	const type = dict[room[0]];
```
ostjen, line 15: since you removed the check for type existing it'd be better to
`const type = dict[room[0]] || {};`

ggazzo (approved):

PR #24365, opened 2022-01-31
[IMPROVE] Rewrite SaveToWebdav Modal to React Component

dougfabris:


client/views/room/webdav/SaveToWebdavModal.tsx
```diff
@@ -0,0 +1,123 @@
...
+
+	const handleSaveFile = ({ accountId }: { accountId: IWebdavAccount['_id'] }): void => {
+		setIsLoading(true);
+
+		const {
+			url,
+			attachment: { title },
+		} = data;
+
+		const fileRequest = new XMLHttpRequest();
```
gabriellsh, line 53: We need a way to close this request if the user wants to. I suggest a cancel button to close the modal and a useEffect to cancel the request on unMount. I think you can save the request in a ref so you can do this.

gabriellsh (changes requested):
I think we need to cancel the

client/views/room/webdav/SaveToWebdavModal.tsx
dougfabris: We agreed to skip this for now!

gabriellsh (approved):

ggazzo (approved):

PR #24366, opened 2022-01-31
Chore: Set Docker image tag to latest only when really latest

debdutdeb:


.github/workflows/build_and_test.yml
```diff
@@ -466,6 +466,22 @@ jobs:
 
         aws s3 cp $ROCKET_DEPLOY_DIR/ s3://download.rocket.chat/build/ --recursive
 
+        # before updating release store, save the current latest stable release
```
sampaiodiego, line 469: remove commented code? 😬

.github/workflows/build_and_test.yml
```diff
@@ -565,9 +581,31 @@ jobs:
         docker build -t $IMAGE_NAME:$DOCKER_TAG .
         docker push $IMAGE_NAME:$DOCKER_TAG
 
-        # build and push the broader alias tag
-        docker tag $IMAGE_NAME:$DOCKER_TAG $IMAGE_NAME:$RELEASE
-        docker push $IMAGE_NAME:$RELEASE
+        # is_gt() {
```
sampaiodiego, line 584: remove this one as well 😬

.github/workflows/build_and_test.yml
```diff
@@ -565,9 +581,31 @@ jobs:
...
+        #   local v2=(${2//./ })
+
+        #   for i in $(seq 0 2); do
+        #     (( ${v1[$i]} < ${v2[$i]} )) && return 1
+        #     (( ${v1[$i]} > ${v2[$i]} )) && return 0
+        #   done
+        # }
+
+        if [[ $RELEASE == 'latest' ]]; then
+          # https://gist.github.com/sampaiodiego/3992a07fdc4e43fd776acd7f36821ea1
```
sampaiodiego, line 595: it lives here now, no need to mention the gist

.github/workflows/build_and_test.yml
```diff
@@ -565,9 +581,31 @@ jobs:
...
+          CURRENT_LATEST="$(
+            git -c 'versionsort.suffix=-' ls-remote -t --exit-code --refs --sort=-v:refname https://github.com/RocketChat/Rocket.Chat '*' |
```
sampaiodiego, line 597: do you know if there is an env var or something else we can use instead of having the repo's url hardcoded here?

sampaiodiego (changes requested):

.github/workflows/build_and_test.yml
debdutdeb: forgot 🙈

.github/workflows/build_and_test.yml
debdutdeb: `GITHUB_REPOSITORY` should do the trick

sampaiodiego (approved):

PR #24388, opened 2022-02-01
[IMPROVE][ENTERPRISE] Improve how micro services are loaded

ggazzo:


ee/server/NetworkBroker.ts
```diff
@@ -0,0 +1,225 @@
...
+
+const lifecycle: { [k: string]: string } = {
+	created: 'created',
+	started: 'started',
+	stopped: 'stopped',
+};
+
+const {
+	INTERNAL_SERVICES_ONLY = 'false',
+	// SERVICES_ALLOWED = '',
```
sampaiodiego, line 23: to remove `SERVICES_ALLOWED` you need to add the `internal` flag on every internal service currently missing the flag (like the `TeamService` for example)

ee/server/NetworkBroker.ts
```diff
@@ -0,0 +1,225 @@
...
+		try {
+			await this.broker.waitForServices(
+				method.split('.')[0],
+				this.isActionWhitelisted(method) ? waitForServicesWhitelistTimeout : waitForServicesTimeout,
+			);
+		} catch (err) {
+			console.error(err);
+		}
+
+		return this.call(method, data);
```
sampaiodiego, line 98: I noticed you're not using the `context` stored on `asyncLocalStorage` here anymore, any reason for that?

ee/server/NetworkBroker.ts
```diff
@@ -0,0 +1,225 @@
...
+		if (!this.isServiceAllowed(instance)) {
+			return;
+		}
+
+		const name = instance.getName();
+
+		if (!this.isServiceInternal(instance)) {
+			instance.onEvent('shutdown', async (services) => {
+				const service = services[name];
+				if (!service) {
```
sampaiodiego, line 115:
```suggestion
				if (!services[name]) {
```

ee/server/NetworkBroker.ts
```diff
@@ -0,0 +1,225 @@
...
+			actions: {},
+			...(name !== 'license'
+				? {
+						dependencies: ['license'],
+				  }
+				: {}),
+			events: instance.getEvents().reduce<Record<string, Function>>((map, eventName) => {
+				map[eventName] = /^\$/.test(eventName)
+					? (data: Parameters<EventSignatures[typeof eventName]>): void => {
+							instance.emit(eventName, data);
```
sampaiodiego, line 133: any reason this is not `instance.emit(eventName, ...data)` anymore?

ee/app/license/server/license.internalService.ts
```diff
@@ -26,6 +26,44 @@ class LicenseService extends ServiceClass implements ILicense {
 		onModule((licenseModule) => {
 			api.broadcast('license.module', licenseModule);
 		});
+
+		this.onEvent('$services.changed', async () => {
+			// if (hasModule('scalability')) {
```
sampaiodiego, line 31: should this be uncommented? (is that a real word? 😳 )

server/services/startup.ts
```diff
@@ -1,28 +1,40 @@
...
-api.registerService(new LDAPService());
 api.registerService(new SAUMonitorService());
+api.registerService(new TeamService(db));
+api.registerService(new UiKitCoreApp());
+
+// if TRANSPORTER env var it means the process is running in micro services mode
+// in that case we don't need to register services that will run separately
+if (!process.env.TRANSPORTER?.match(/^(?:nats|TCP)/)) {
+	(async (): Promise<void> => {
+		const { Authorization } = await import('./authorization/service');
```
KevLehman, line 36: I'm curious 🤔 can this be a separate function?
Like:
```
api.importAndRegister(path: string, ...deps: any[])
```

That way, we won't have to import things here and the iife won't grow too big 😬

sampaiodiego: hmm.. interesting 🤔  I'll take a look and see if it works

sampaiodiego: this is definitely a good idea.. moleculer has the same feature to create/register services by using `broker.loadService('./file.js')` .. 

this is not something I'll provide on this PR, but I'll take notes to look at this in the future 👍 

thx again for bringing up this idea @KevLehman

sampaiodiego (dismissed):

sampaiodiego (approved):

PR #24398, opened 2022-02-03
[IMPROVE] Room and Channel Wrong Format Logo Issue

nishant23122000:
I created one PR in 2021 with the same issue, but after gaining experience in one year, I feel like the recent solution is not robust because it's handling only JPEF, PNG. 

## Issue(s)
You can check last year's PR #21298 where you will find the issue also.


guijun13:
Already fixed here: https://github.com/RocketChat/Rocket.Chat/pull/25912

Issue #24399, opened 2022-02-03
[BUG] Audio File Must Be Save As MP3


nishant23122000:
### Description:
When the user records the audio, It must be store in an mp3 format.

### Steps to reproduce:

1. Go to any room
2. Record the Audio
3. Now one Dialog box will open
4. remove the .mp3 and add any other format or leave it blank without any format
5. Then go to the search File section on the top right to search all files
6. Download the file which you saved
7. If you save the file as .img then it will download the .img file and vice versa for other formats

### Expected behavior:

When the user records audio, then in the dialog box we should not give permission to change the file format otherwise it will download another file format when we download it from the search file section.

### Actual behavior:

It allows the user to change the file format in the input box and that will give a different file format while downloading the file.

https://user-images.githubusercontent.com/53515714/152321364-400dd265-59ef-42fb-b5de-e7c98d62b502.mp4

### Server Setup Information:

- Version of Rocket.Chat Server:  4.4.0
- Operating System: Windows 10
- Deployment Method: 
- Number of Running Instances: 
- DB Replicaset Oplog: 
- NodeJS Version: 14.18.2
- MongoDB Version: 4.4.4

### Client Setup Information

- Desktop App or Browser Version:
- Operating System:

PR #24400, opened 2022-02-03
[FIX] Audio Should Not Save Without Any extention

nishant23122000:
You can find more details in the issue #24399

## Issue(s)
#24399

## Steps to test or reproduce
Already in #24399

## Further comments

After Resolving Bug 

https://user-images.githubusercontent.com/53515714/152334329-9c13b0a2-1923-42c9-b81f-afc751c3cd2f.mp4


debdutdeb (commented):
Hi, and thanks for this PR. Here is the problem with this change from my perspective,

1. mp3 is not the *only* audio format that web, i.e. rocketchat supports, but this enforces that.
2. why is it a bad thing to be able to change the extension? Obviously you can see the web doesn't care. 
3. the *default* filllers, iow extensions, are present in both cases (uploading or downloading). there is no practical reason to change them, unless there is, in which case we should just trust the user to know what they're doing.

This isn't solving any issue or improving anything, again from my perspective.

PR #24403, opened 2022-02-03
Chore: update CONTRIBUTING.md

debdutdeb:

- [x] copied existing developer guidelines from handbook
- [x] moved file to project root for easier reach/recognition


sampaiodiego (commented):
I just found this.. do we still want to move everything from our handbook to here only?

debdutdeb:
Yes, that'd be really good. we can keep a copy there, but this being here makes sense.

sampaiodiego:
it was here at some point, and then we decided to move it there.
I don't think we should have it duplicated though.
I don't remember exactly why we decided to move it to the handbook only, but I'd like to know if this is something someone else agreed moving back here.

PR #24407, opened 2022-02-03
[FIX] Skip cloud steps for registered servers on setup wizard

dougfabris:


server/methods/getSetupWizardParameters.js
```diff
@@ -5,11 +5,12 @@ import { Settings } from '../../app/models';
 Meteor.methods({
 	getSetupWizardParameters() {
 		const settings = Settings.findSetupWizardSettings().fetch();
-		const allowStandaloneServer = process.env.DEPLOY_PLATFORM !== 'rocket-cloud';
+		const [registerServerSetting] = Settings.findById('Register_Server').fetch();
+		const canSkipRegistration = registerServerSetting.value || process.env.DEPLOY_PLATFORM === 'rocket-cloud';
```
geekgonecrazy, line 9: I don't think this will cover what you're trying to do.  

Register_Server is a boolean that controls if a user "wants" the server to be registered or allowed to communicate with Cloud  It doesn't reflect in any way the actual registration status of the workspace.

The reliable indicator that a workspace is actually registered would be: 
https://github.com/RocketChat/Rocket.Chat/blob/7632f12cfcc7ed8ee8f843587fdff63b101cc765/app/cloud/server/functions/retrieveRegistrationStatus.js#L7

```
settings.get('Cloud_Workspace_Client_Id')
```

You'll notice this is the `settings` and not the `Settings` also.  Need to use the raw in this case otherwise we will be back here fixing this having bugs in a few days.

The cached Settings isn't the most reliable in this case.  Because the registration could have happened on another instance and some how not propagated to the cache via oplog/change stream.  

Before we made the change to the raw settings it would randomly return false when it was actually true.  Causing a lot of issues as you can imagine

client/views/setupWizard/steps/OrganizationInfoStep.tsx
```diff
@@ -41,6 +44,9 @@ const OrganizationInfoStep = (): ReactElement => {
 	const organizationSizeOptions = getSettingOptions(settings, 'Size', t);
 
 	const handleSubmit: ComponentProps<typeof OrganizationInfoPage>['onSubmit'] = async (data) => {
+		if (canDeclineServerRegistration) {
```
geekgonecrazy, line 47: IMO reading this with out the context I think I'd be confused.  

`canDeclineServerRegistration` to me means the user has the ability or not to choose not to register.   But unless I'm not understanding.. that's not the case.

What this is actually controlling is **if** the registration will be skipped.  If so I'd propose: `skipCloudRegistration` or `serverAlreadyRegistered`

Something that when we come back we know what it actually means.  (same for the other occurrences)

geekgonecrazy (approved):
LGTM 👍 thanks for addressing the feedback 🙏

tassoevan (approved):

PR #24408, opened 2022-02-03
[NEW] Connectivity check between RC and asterisk

amolghode1981:
This PR verifies the connectivity of the rocketchat server and asterisk.
These changes also verify the authentication credentials too. 
Changes :
1. Modified server/services/voip/connector/asterisk/ami/AMIConnection.ts to avoid connection race.
In earlier version of the code, there was an issue because of which if we check the connection immediately after creating it, it would not be up. This was because the Connection happens asynchronously.  
This is fixed by handling the connection on success callback of each step such as connect/login. If we get an error in any of the stage, we reject the promise.
The caller waits for on the promise

Added a new API 
Added i get api voipServerConfig.management.checkconnection in app/api/server/v1/voip/server-config.ts
This API needs
{
	host: String,
	port: String,
	username: String,
	password: String,
}

Necessary wiring around the core connectivity check functionality.

Return values : 
In case of success following object is returned
{
	status: 'connected'
}

In case of error following object is returned
{
	status: 'connection-error',
	error: error message describing the reason of the error such as auth failure, connection failure, address does not exist etc.
}

## Further comments
Testing Notes :
1. Check that the connection successfully happens.
2. Enter wrong asterisk IP check for proper failure.
3. Enter wrong asterisk management port. Check for proper failure.
4. Add incorrect username. Check for proper failure.
5. Add incorrect password. Check for proper failure.
6. Make asterisk non reachable. Check for proper failure.


app/api/server/v1/voip/server-config.ts
```diff
@@ -59,6 +59,30 @@ API.v1.addRoute(
...
+				Match.ObjectIncluding({
+					host: String,
+					port: String,
+					username: String,
+					password: String,
+				}),
+			);
+			const { host, port, username, password } = this.requestParams();
+			const status = await Voip.checkManagementConnection(host, port, username, password);
+			return API.v1.success({ ...status });
```
KevLehman, line 81:
```suggestion
			return API.v1.success(await Voip.checkManagementConnection(host, port, username, password));
```
No need to reassign the result if you're not changing it 👀

server/services/voip/connector/asterisk/CommandHandler.ts
```diff
@@ -60,12 +61,17 @@ export class CommandHandler {
...
-		);
+		try {
+			await connection.connect(
+				config.host,
+				(config.configData as IManagementConfigData).port.toString(),
+				(config.configData as IManagementConfigData).username,
+				(config.configData as IManagementConfigData).password,
+			);
+		} catch (error) {
+			this.logger.warn('Management server connection error');
```
KevLehman, line 72: Pls log the error

KevLehman (dismissed):

server/services/voip/connector/asterisk/CommandHandler.ts
amolghode1981: Fixed.

app/api/server/v1/voip/server-config.ts
amolghode1981: Fixed.

KevLehman (approved):

PR #24421, opened 2022-02-05
[NEW] VoIP buttons to mute and hold the call

tiagoevanp:

image


MartinSchoeler (approved):
While I don't agree with some nomenclature changes the PR is working fine and the code looks ok, so I'll approve this and later we can discuss better the naming and structure :)

Issue #24419, opened 2022-02-05
[BUG] Should Display appropriate Message


nishant23122000:
### Description:
When there are no messages to prune, It should give the appropriate response.

### Steps to reproduce:

1. Go to any Channel
2. Click on Prune Messages on the top right
3. set the time periods where no message exists to prune
4. prune all messages or any particular user messages
5. It will show an undefined error

### Expected behavior:

When there is no message to prune for given time periods, we should display an appropriate message.

### Actual behavior:

It showing undefined if there is no message to prune.

### Server Setup Information:

- Version of Rocket.Chat Server: 4.5.0 
- Operating System: linux
- Deployment Method: 
- Number of Running Instances: 
- DB Replicaset Oplog: 
- NodeJS Version: 14.18.2
- MongoDB Version: 4.4.4

### Client Setup Information

- Desktop App or Browser Version:
- Operating System:

Sayan-dev:
Does this needs work or is it fixed?

PR #24424, opened 2022-02-06
[FIX] Prune Message issue

nishant23122000:

https://user-images.githubusercontent.com/53515714/152670869-59134755-18a6-4bc4-be3f-e013f5d63480.mp4

## Issue(s)
#24419

## Steps to test or reproduce
Mention in issue #24419

## Further comments

**new Throw()**  return an object which contains a message and name key, and we were printing the error object directly that's why it gives undefined. we can get error message by error.message


murtaza98 (approved):
LGTM!

PR #24441, opened 2022-02-08
[FIX] GDPR action to forget visitor data on request

KevLehman:
## Further comments

- The issue was present on an old fix that prevented admins/users to remove visitors when they had open rooms. However, this affected GDPR "right to forget". This PR fixes that


murtaza98 (dismissed):
LGTM!

murtaza98 (approved):

PR #24452, opened 2022-02-08
Chore: Create TS Interface for OTR Room

kibonusp:


KevLehman:
The OTR feature was removed on v8. Thanks!

PR #24467, opened 2022-02-10
Chore: Improve PR title validation regex

sampaiodiego:

I don't know how it happened, but looks like the regex was accepting some invalid titles like for example:

```
[IMPROVE][EE] Description
```

In this case, it should be invalid because the seconds group (`[EE]`) is invalid, but since `[IMPROVE]` is valid, the regex has a match and so the title is considered valid.

The new regex will validate the second group as well, if present, and allow only the values `[ENTERPRISE]` and `[APPS]`.

I've used this list of valid and invalid titles to check the regex:

```
[NEW] ok
[BREAK] ok
[IMPROVE] ok
[FIX] ok
[BREAK][ENTERPRISE] ok
[BREAK][APPS] ok
Regression: ok
Chore: ok
Revert: ok
i18n: ok
Bump ok
Release 1.2.3
Merge master into develop

[NEW]not ok
[NEW][NOPE] not ok
Regression:not ok
Bump: not ok
Release version 1.2.3 not ok
```

I'm considering having somes tests to make sure the regex works as expected and having a reference to check at some point. wdyt?


.github/pr-title-checker-config.json
```diff
@@ -4,7 +4,7 @@
 		"color": "B60205"
 	},
 	"CHECKS": {
-		"regexp": "^(?:(?:\\[(NEW|BREAK|IMPROVE|FIX)\\](\\[(ENTERPRISE|APPS)\\])?|(?:Regression|Chore|Revert|i18n):)|(?:Bump) .+|Release [0-9]+\\.[0-9]+\\.[0-9]+|Merge master into develop)",
+		"regexp": "^(?:(?:\\[(NEW|BREAK|IMPROVE|FIX)\\](?:[^[]|\\[(ENTERPRISE|APPS)\\])|(?:Regression|Chore|Revert|i18n):)|(?:Bump) .+|Release [0-9]+\\.[0-9]+\\.[0-9]+|Merge master into develop)",
```
debdutdeb, line 7: The current regex, my guess, seems to be failing because the first group check is loose - if it is in the given string or not. So as long as the given string prefix *contains* one of the given groups, it's fine. I think the better solution is to introduce another group for just the title prefix, as in 

```suggestion
		"regexp": "^(?:(?:(?:\\[(NEW|BREAK|IMPROVE|FIX)\\](\\[(ENTERPRISE|APPS)\\])?|(?:Regression|Chore|Revert|i18n):)|(?:Bump)) .+|Release [0-9]+\\.[0-9]+\\.[0-9]+|Merge master into develop)",
```

Adding as a suggestion so that the change is more visible

debdutdeb: very wrong "on behalf of" 🙈

debdutdeb: Having said that, `[^[]` works too, but is slgihtly less readable without context for those who're not as comfortable with regexes.

.github/pr-title-checker-config.json
```diff
@@ -4,7 +4,7 @@
 		"color": "B60205"
 	},
 	"CHECKS": {
-		"regexp": "^(?:(?:\\[(NEW|BREAK|IMPROVE|FIX)\\](\\[(ENTERPRISE|APPS)\\])?|(?:Regression|Chore|Revert|i18n):)|(?:Bump) .+|Release [0-9]+\\.[0-9]+\\.[0-9]+|Merge master into develop)",
+		"regexp": "^(?:(?:(?:\\[(NEW|BREAK|IMPROVE|FIX)\\](\\[(ENTERPRISE|APPS)\\])?|(?:Regression|Chore|Revert|i18n):)|(?:Bump)) .+|Release [0-9]+\\.[0-9]+\\.[0-9]+|Merge master into develop)",
```
debdutdeb, line 7:
```suggestion
		"regexp": "^(?:(?:(?:\\[(NEW|BREAK|IMPROVE|FIX)\\](\\[(ENTERPRISE|APPS)\\])?|(?:Regression|Chore|Revert|i18n):)|(?:Bump)) .+|Release [0-9]+\\.[0-9]+\\.[0-9]+|Merge master into develop)$",
```

We should also make sure nothing sneaks in after Release x.y.z and Merge master into develop 😬

sampaiodiego: that's true for `Relase x.y.z` but not for `Merge master ..` as you can at https://github.com/RocketChat/Rocket.Chat/pull/24363

debdutdeb: Ah .. well then we can use three separate groups, but tbh release ones will only be internal and that too not in hundreds, so doesn't matter ig

debdutdeb (approved):

PR #24475, opened 2022-02-10
[IMPROVE] Skip encryption for slash commands in E2E rooms

yash-rajpal:

Currently Slash Commands don't work in an E2EE room, as we encrypt the message before slash command is detected by the server, So removed encryption for slash commands in e2e rooms.

## Steps to test or reproduce

- Create an E2EE room
- Try to execute any slash command in it.
- Slash command sent as normal encrypted message.

### Before
image

### After
image


client/startup/e2e.ts
```diff
@@ -116,7 +116,7 @@ Meteor.startup(() => {
 
 			subscription.encrypted ? e2eRoom.resume() : e2eRoom.pause();
 
-			if (!(await e2eRoom.shouldConvertSentMessages())) {
+			if (!(await e2eRoom.shouldConvertSentMessages(message))) {
```
albuquerquefabio, line 119:
```suggestion
                        const isConvert = await e2eRoom.shouldConvertSentMessages(message)
			if (!isConvert) {
```

albuquerquefabio (changes requested):

client/startup/e2e.ts
yash-rajpal: Done! Thanks!

albuquerquefabio (approved):

PR #24484, opened 2022-02-11
[NEW] Livechat Voip Room

tiagoevanp:

### DIFF
--- a/app/livechat/client/tabBar.ts
+++ b/app/livechat/client/tabBar.ts
@@ -3,7 +3,7 @@ import { lazy } from 'react';
 import { addAction } from '../../../client/views/room/lib/Toolbox';
 
 addAction('room-info', {
-	groups: ['live'],
+	groups: ['live', 'voip'],
 	id: 'room-info',
 	title: 'Room_Info',
 	icon: 'info-circled',
@@ -12,7 +12,7 @@ addAction('room-info', {
 });
 
 addAction('contact-chat-history', {
-	groups: ['live'],
+	groups: ['live', 'voip'],
 	id: 'contact-chat-history',
 	title: 'Contact_Chat_History',
 	icon: 'clock',
--- a/client/sidebar/footer/voip/index.tsx
+++ b/client/sidebar/footer/voip/index.tsx
@@ -1,16 +1,29 @@
 import React, { ReactElement, useCallback, useState } from 'react';
 
+import { roomTypes } from '../../../../app/utils/client';
 import { useCallActions, useCallerInfo } from '../../../contexts/CallContext';
+import { useEndpoint } from '../../../contexts/ServerContext';
 import { useTranslation } from '../../../contexts/TranslationContext';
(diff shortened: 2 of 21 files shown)


app/livechat/client/tabBar.ts
```diff
@@ -3,7 +3,7 @@ import { lazy } from 'react';
 import { addAction } from '../../../client/views/room/lib/Toolbox';
 
 addAction('room-info', {
-	groups: ['live'],
```
MartinSchoeler, line 6: Will we uncomment this later on?

app/livechat/server/api/v1/visitor.ts
```diff
@@ -37,13 +37,15 @@ API.v1.addRoute('livechat/visitor', {
 		}
 
 		guest.connectionData = normalizeHttpHeaderData(this.request.headers);
-		const visitorId = Livechat.registerGuest(guest);
+		const visitorId = Livechat.registerGuest(guest as any);
```
MartinSchoeler, line 40: Is this `any` really necessary? If so, pls use `unknown` instead

client/sidebar/footer/voip/index.tsx
```diff
@@ -1,17 +1,41 @@
...
 	const callActions = useCallActions();
+	const user = useUser();
 
 	const [muted, setMuted] = useState(false);
 	const [paused, setPaused] = useState(false);
 
+	const visitorEndpoint = useEndpoint('POST', 'livechat/visitor');
+	const voipEndpoint = useEndpoint('GET', 'voip/room');
+
+	const openRoom = useCallback(async () => {
```
MartinSchoeler, line 23: How do you feel about moving the function to the voice context and only calling it here?

MartinSchoeler (changes requested):

client/views/room/Header/RoomHeader.tsx
```diff
@@ -1,6 +1,6 @@
 import React, { FC } from 'react';
 
-import { IOmnichannelRoom } from '../../../../definition/IRoom';
```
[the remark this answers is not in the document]
tiagoevanp: This isn't an Omnichannel exclusive component... just taking vantage to fix this together...

app/livechat/client/tabBar.ts
tiagoevanp: Yes, doing that on a different PR

app/livechat/server/api/v1/visitor.ts
tiagoevanp: yes... is necessary! The Livechat.registerGuest function is in a .js file with a destructured object as guest... I can't pass this as unknown

MartinSchoeler: Then can you add a TODO there saying that we need to convert the js file and the name of the file. I can create a task for it, so we don't forget

MartinSchoeler (approved):

PR #24485, opened 2022-02-11
[FIX] Skip admin info in setup wizard for servers with admin registered

dougfabris:


gabriellsh (changes requested):
As in our previous conversation, the flag for if there's an admin already should come from the server. Current behaviour in this PR doesn't account for if the admin is logged out.

Also, only logged admins should be able to view the setup wizard, which isn't the current case.

PR #24506, opened 2022-02-14
Chore: added settings endpoint types

felipe-rod123:

Created typing for endpoint definitions on `settings.ts`.

### DIFF
--- a/definition/rest/v1/settings.ts
+++ b/definition/rest/v1/settings.ts
@@ -18,9 +18,9 @@ export type OauthCustomConfiguration = {
 	authorizePath: string;
 	scope: string;
 	loginStyle: 'popup' | 'redirect';
-	tokenSentVia: string;
-	identityTokenSentVia: string;
-	keyField: string;
+	tokenSentVia: 'header' | 'payload';
+	identityTokenSentVia: 'default' | 'header' | 'payload';
+	keyField: 'username' | 'email';
 	usernameField: string;
 	emailField: string;
 	nameField: string;
--- a/definition/rest/v1/settings.ts
+++ b/definition/rest/v1/settings.ts
@@ -10,38 +10,38 @@ type SettingsUpdatePropsActions = {
 export type OauthCustomConfiguration = {
 	_id: string;
 	clientId?: string;
-	custom: boolean; // unknown
+	custom: boolean;
 	service?: string;
-	serverURL: string; // unknown
-	tokenPath: string; // unknown
-	identityPath: string; // unknown
-	authorizePath: string; // unknown
-	scope: string; // unknown
+	serverURL: string;
+	tokenPath: string;
+	identityPath: string;
+	authorizePath: string;
+	scope: string;
 	loginStyle: 'popup' | 'redirect';
-	tokenSentVia: string | boolean; // unknown
-	identityTokenSentVia: string | boolean; // unknown
-	keyField: string | boolean; // unknown
-	usernameField: string | boolean; // unknown
-	emailField: string | boolean; // unknown
-	nameField: string | boolean; // unknown
-	avatarField: string | boolean; // unknown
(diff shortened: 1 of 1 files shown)


definition/rest/v1/settings.ts
```diff
@@ -10,38 +10,38 @@ type SettingsUpdatePropsActions = {
...
+	showButton: boolean;
+
+	appId: string;
 	consumerKey?: string;
 
-	clientConfig: unknown;
-	buttonLabelText: unknown;
-	buttonLabelColor: unknown;
-	buttonColor: unknown;
+	clientConfig: { provider: SettingValue };
```
sampaiodiego, line 41: as you can see below, `clientConfig` can have different values, so maybe let this as `unknown`:

https://github.com/RocketChat/Rocket.Chat/blob/d42f1643f1b6ceff55ae079509879a4433741ffb/app/lib/server/startup/oAuthServicesUpdate.js#L100

definition/rest/v1/settings.ts
```diff
@@ -10,38 +10,38 @@ type SettingsUpdatePropsActions = {
...
+	buttonLabelText: SettingValue;
+	buttonLabelColor: SettingValue;
+	buttonColor: SettingValue;
```
sampaiodiego, line 44: what's the reason to use `SettingValue`? looks like they're just `string`

definition/rest/v1/settings.ts
```diff
@@ -10,38 +10,38 @@ type SettingsUpdatePropsActions = {
...
-	showButton: unknown;
-
-	appId: unknown;
+	tokenSentVia: string | boolean;
+	identityTokenSentVia: string | boolean;
+	keyField: string | boolean;
+	usernameField: string | boolean;
+	emailField: string | boolean;
+	nameField: string | boolean;
+	avatarField: string | boolean;
```
sampaiodiego, line 27: I didn't find any place using these keys as `boolean`, only as `string` .. I'd like to understand why you added `boolean` for them :)

sampaiodiego (changes requested):

definition/rest/v1/settings.ts
felipe-rod123: On the `app/meteor-accounts-saml/server/lib/settings.ts`, there's a `settings.get` that returns a SettingValue, so I wasn't sure if this could affect something. But I'll change it to string! Thanks!

definition/rest/v1/settings.ts
```diff
@@ -10,38 +10,38 @@ type SettingsUpdatePropsActions = {
...
-	mapChannels: unknown;
-	channelsMap: unknown;
-	channelsAdmin: unknown;
-	mergeUsers: unknown;
-	mergeRoles: unknown;
-	accessTokenParam: unknown;
-	showButton: unknown;
-
-	appId: unknown;
+	tokenSentVia: string;
```
pierre-lehnen-rc, line 21:
```suggestion
	tokenSentVia: 'header' | 'payload';
```

definition/rest/v1/settings.ts
```diff
@@ -10,38 +10,38 @@ type SettingsUpdatePropsActions = {
...
+	identityTokenSentVia: string;
```
pierre-lehnen-rc, line 22:
```suggestion
	identityTokenSentVia: 'default' | 'header' | 'payload';
```

definition/rest/v1/settings.ts
```diff
@@ -10,38 +10,38 @@ type SettingsUpdatePropsActions = {
...
+	keyField: string;
```
pierre-lehnen-rc, line 23:
```suggestion
	keyField: 'username' | 'email';
```

pierre-lehnen-rc (changes requested):

ggazzo (approved):

Issue #23029, opened 2021-08-25
Loading new messages while scrolling up moves you to undesired position


GirardR1006:
Hello and good afternoon everyone :)

### Description:

In a channel, when my users scroll up to the point new messages must be loaded, they are moved to an undesired location in the channel history. Even more, the client keep loading for new messages, which makes the location to change multiple type. This is particularly annoying when trying to scroll up on a high-activity channel, since users loose focus on where they were. 

### Steps to reproduce:

1. load a channel with at least 50 messages
2. hit CTRL+R
3. scroll up until you reach the top of loaded messages

### Expected behavior:

Location should stay at the first loaded message (meaning should not change at all when new messages are loaded)

### Actual behavior:

Location changes to a different message several times, which lead to confusion while scrolling up. Attached video to illustrate this

https://user-images.githubusercontent.com/19275558/130788257-cf80d779-ee94-4da5-859d-3e05378a21bc.mp4

### Server Setup Information:

- Version of Rocket.Chat Server: 3.16.3
- Operating System: Debian
- Deployment Method: Manual
- Number of Running Instances: 1
- DB Replicaset Oplog: mmapv1
- NodeJS Version: v12.18.4
- MongoDB Version: 4.0.25 

### Client Setup Information

- Desktop App or Browser Version: multiples: Firefox 91.0.1 and Chrome
- Operating System: Linux, Windows, macOS

debdutdeb:
Hi. Thanks for reporting this. I _should_ be asking you to test this on the latest, but I know I have come across it multiple times, especially when trying to jump to a message from search. 

@ankar84, have you faced this on your end? Anything you can share?

GirardR1006:
> Hi. Thanks for reporting this. I _should_ be asking you to test this on the latest, but I know I have come across it multiple times, especially when trying to jump to a message from search.
> 
> @ankar84, have you faced this on your end? Anything you can share?

I updated on 3.18.0 with no change on my side

emikolajczak:
This is very old bug
Related to:
https://github.com/RocketChat/Rocket.Chat/issues/19871

Vringe:
Is there any update?
This is a very annoying issue for over a year now and it affects almost all users.

GirardR1006:
Updated to latest version did not fix it. That seem quite a serious issue for me, I am not really sure why it does not have any more exposure.

lelvisl:
Same problem on 4.3.2

[...]

Issue #19871, opened 2020-12-15
Scrolling forward while a channel is loading new messages causes auto-scroll to the very bottom to occur


jessey-git:
### Description:

Despite the person on #19112 saying this was fixed, it is definitely not fixed on 3.9.1...

Scrolling forward while a channel is loading new messages causes auto-scroll to the very bottom to occur. This makes catching up on the previous night's conversations nearly impossible.

The poor behavior of paging-in messages (in both scroll-back and scroll-forward cases) continues to be a source of absolute rage for members.

As this is now the third time I'm reporting this same bug where it was incorrectly closed each time. Why is auto-scroll even enabled at this point. It is broken and useless.

### Steps to reproduce:

1. Go to a channel with a lot of new messages (come in first thing in morning to a channel)
2. Click the "Jump to first message" or similar to jump to your last read message
3. Start scrolling forward reading the messages in order
4. You'll soon hit a point, if the channel is large enough, where new messages have to be paged in
5. If you hit this point the chat window will basically just auto-scroll to the bottom; leaving you having to fish back upwards over several hours of chats to find your last point manually...

### Expected behavior:
When paging in new messages in both scroll forward and scroll back cases DO NOT continue scrolling. Immediately STOP at the exact location the user was already at.

### Actual behavior:

Incorrect, annoying, auto-scrolling occurs unnecessarily.

### Server Setup Information:

blender.chat
- Version of Rocket.Chat Server: 3.9.1
- Operating System: linux
- Deployment Method: 
- Number of Running Instances: 
- DB Replicaset Oplog: 
- NodeJS Version: 
- MongoDB Version:

### Client Setup Information

- Desktop App or Browser Version: Desktop Chrome Browser Version 87.0.4280.88 (Official Build) (64-bit)
- Operating System: Win10

kakkou:
I tried 3.9.1 but I don't think it's fixed.
Even now, when I start reading old messages, the scroll position is changed without permission, and the location cannot be maintained.


Discord does not change the current scroll position at all when reading old messages.
Also, Rocket.Chat Mobile does not have this issue.

z3braman:
I concur that this problem occurs every time I use rocket chat per the OP description. I am unable to cite the version as it is viewed via Safari browser and the site maintained by service I subscribe to. Yes, I am enraged when this occurs. Trying to be efficient when catching up on messages gets defeated when the scroll causes the cursor to move to previous day information.

lelvisl:
[...]

PR #24520, opened 2022-02-16
[FIX] Scrolling-up moves to unexpected position

Dnouv:

On the channel or teams page having more than 50+ chats, when a user scrolls to load old messages calling the `loadHistory()` method, once the messages are fetched and rendered, the scroll position moves out of the current viewport to a random message instead of the last seen message.
This PR implements a `scrollIntoView` method on the last appeared message, which scrolls the message viewport such that the last seen message always appears at the bottom for every load of old messages.

Here is a clip showing the made changes, 

https://user-images.githubusercontent.com/61188295/154330267-59006ea9-7e5c-448e-95fc-78d6d511c3f0.mp4

During the first call to load history messages, the last seen message was "52" after the old messages are loaded the chat window is scrolled back to the position where "52" is at the bottom.

## Issue(s)

#23029 
#19871 

## Steps to test or reproduce

Mentioned in issue #23029 the affected screens would be the `message wrapper`

### DIFF
--- a/app/ui-utils/client/lib/RoomHistoryManager.js
+++ b/app/ui-utils/client/lib/RoomHistoryManager.js
@@ -235,6 +235,7 @@ export const RoomHistoryManager = new (class extends Emitter {
 		waitAfterFlush(() => {
 			const heightDiff = wrapper.scrollHeight - previousHeight;
 			wrapper.scrollTop = scroll + heightDiff;
+			lastMessage && document.getElementById(lastMessage._id).scrollIntoView({ behavior: 'smooth', block: 'end', inline: 'nearest' });
 		});
 
 		room.isLoading.set(false);


ankar84:
Nice improvement for UX

Dnouv:
@debdutdeb, could you please review this PR.
Thank you!

app/ui-utils/client/lib/RoomHistoryManager.js
```diff
@@ -235,6 +235,7 @@ export const RoomHistoryManager = new (class extends Emitter {
 		waitAfterFlush(() => {
 			const heightDiff = wrapper.scrollHeight - previousHeight;
 			wrapper.scrollTop = scroll + heightDiff;
+			lastMessage && document.getElementById(lastMessage._id).scrollIntoView({ behavior: 'smooth', block: 'end', inline: 'nearest' });
```
debdutdeb, line 238: This scrolls the dom twice, which is not very efficient 🤔

debdutdeb (changes requested):

app/ui-utils/client/lib/RoomHistoryManager.js
Dnouv: Thanks a lot for finally reviewing this PR. Actually, in both (before and after changes) cases DOM scrolls twice, the difference is in here it scrolls smoothly.

debdutdeb: When the message box is refreshed with history the whole view gets to the top, so to *fix* that the view is then scrolled back down with `Element.scrollTop`. So yes it **was** scrolling twice already, but now it's doing it three times - 1) scrolling to the top with all the new messages 2) scrolling down to get the user to the expected position 3) scrolling again because the last one didn't quite work as expected - I just didn't count the initial scroll the first time.

does this make sense?

Dnouv: Yeah, now I understand. So, now I have changed the current logic to adapt to the currently used code structure. Now, after the new messages are loaded the DOM is scrolled back such that the last visible message before loading new messages appears at the top, here is a video, showing the behavior.

https://user-images.githubusercontent.com/61188295/156872481-74addaa1-0038-45d8-874e-948164789d84.mp4

Please review once again. Thank you!

app/ui-utils/client/lib/RoomHistoryManager.js
```diff
@@ -234,7 +232,15 @@ export const RoomHistoryManager = new (class extends Emitter {
...
-			wrapper.scrollTop = scroll + heightDiff;
+			let toScroll = wrapper.scrollHeight;
+			let len = Math.ceil(Math.log10(wrapper.scrollHeight + 1));
+
+			while (len > 2 && heightDiff < toScroll) {
+				toScroll -= heightDiff;
+				len = Math.ceil(Math.log10(toScroll + 1));
+			}
+
+			wrapper.scrollTop = toScroll + heightDiff;
```
debdutdeb, line 243: Ok, so, like the clever algorithm ;) 

But, two points, 1) code's glancability is too low; 2) don't think it still solves the problem;

Can you check a suspicion of mine please? There, if i'm not  mistaken, is a *race condition*. Think about someone else in the same room sending messages while another person is  attempting to load history. Multiple people at the same time doing the same. You save the *current* (really the key word should be "*instanteneous*") height, do your calculations, but in the meantime the div height is still increasing/has increased. That invalidates any calculations done on the old data. And we're left with random scrolls.

debdutdeb (commented):
Neither approving nor requesting any specific changes for now. It'd be great if you could confirm the suspicion.

I do like your initial way of using scrollToView since that is per-client and doesn't depend on the environment/get affected by environmental changes. If you can confirm I'll talk to the frontend team about that.

app/ui-utils/client/lib/RoomHistoryManager.js
Dnouv: Thanks for the review, yeah right, my bad, I missed `race condition`, to re-verify the code for this specific case while loading history I sent a too-long message, and yeah it broke. Well, I did verify this scenario for the current in-use code and it also breaks so should I revert the changes to `scrollIntoView` it scrolls the DOM thrice but works fine in this scenario. Please let me know whether or not we're going with this (scrollIntoView).
Thank you!

debdutdeb: > it scrolls the DOM thrice but works fine in this scenario  

Definitely like scrollIntoView, but don't scroll three times. u can just clean up the other scrolling bit. 

Before implementing can you also check if there's any cross-browser/cross-version compatibility issues with this method?

Another thing imo would be great is if you can highlight the last seen message after scrolling to the right place for  like a second. IIRC this same file has example for this - uses jquery. Because of the inevitable two scroll situation and crowdedness, even if the view positions itself correctly reliably, it kinda becomes hard to track the last viewed message. The highlight will be able to help with that.

Additionally, please go through the code multiple times, and really clean up any redundant code. Thanks!

Sing-Li:
> Nice improvement for UX

Totally concurred that this needs to be fixed.    The "user experience" is down right *janky* and irritating.     It makes Rocket.Chat feels _second class_ when compared with the likes of Slack or Microsoft Team.

app/ui-utils/client/lib/RoomHistoryManager.js
Dnouv: Thank you for the valuable feedback debdutdeb and Sing-Li , I'll try to clean up as much as possible and will also add a highlight on the last message. 
I'll be looking forward to your guidance if got stuck anywhere. Thank you!

Dnouv: I have removed the thrice scrolling part and also added the highlight function to the last seen message; here is a vid showing the changes.

https://user-images.githubusercontent.com/61188295/156938325-a69a118f-e8f0-4673-aebc-ead8638ef07e.mp4

Please have a review. Thank you!

app/ui-utils/client/lib/RoomHistoryManager.js
```diff
@@ -233,8 +224,9 @@ export const RoomHistoryManager = new (class extends Emitter {
 		}
 
 		waitAfterFlush(() => {
-			const heightDiff = wrapper.scrollHeight - previousHeight;
-			wrapper.scrollTop = scroll + heightDiff;
+			lastMessage &&
+				($(`#${lastMessage._id}`).addClass('highlight'),
+				document.getElementById(lastMessage._id).scrollIntoView({ block: 'end', inline: 'nearest' }));
```
debdutdeb, line 229: Hm can you simplify this a little bit? Best to do `if (!lastMessage) return;` then the rest. Also, can we not store `$('#' + lastMessage._id, wrapper)` and get the `Element` from there instead if calling `document.getElementById` again?

app/ui-utils/client/lib/RoomHistoryManager.js
```diff
@@ -245,6 +237,7 @@ export const RoomHistoryManager = new (class extends Emitter {
 	}
 
 	async getMoreNext(rid, limit = defaultLimit) {
+		console.log('getmore');
```
debdutdeb, line 240:
```suggestion
```

debdutdeb (changes requested):

Dnouv:
Thanks a lot for the review @debdutdeb 
I have made the required changes. Please have a review.
Thank you!

ankar84:
> > Nice improvement for UX
> 
> Totally concurred that this needs to be fixed. The "user experience" is down right _janky_ and irritating. It makes Rocket.Chat feels _second class_ when compared with the likes of Slack or Microsoft Team.

Exactly! Totally agree with Sing!!!
That idea with me al the time I use RC. If all that little problems will be solved some day - RC will be very first among all messengers for sure!

ankar84:
@debdutdeb any updates here?

Issue #24523, opened 2022-02-17
[IMPROVE] No Scroll Down Button Available


nishant23122000:
### Description:

I was showing all the past messages of the GSoC 2021 channel and I found that there is no button that redirects me to the bottom message directly. When there are new messages then the "new message" button will pop up.

### Steps to reproduce:

1. Go to any room Which contains many messages
2. Scroll up inside messages
3. Then if you want to scroll the bottom message, there is no button. (**alt+down arrow** will not direct redirect to bottom )

### Expected behavior:

We should add a **scroll down** button like the **new message** button which redirects to the bottom message.

### Actual behavior:

No button or key which direct redirects to the bottom message.

### Server Setup Information:

- Version of Rocket.Chat Server: 4.5.0
- Operating System: Windows 10
- Deployment Method: 
- Number of Running Instances: 
- DB Replicaset Oplog: 
- NodeJS Version: 14.18.2
- MongoDB Version: 4.4.4

### Client Setup Information

- Desktop App or Browser Version:
- Operating System:

nishant23122000:
Hi @SanjaySinghRajpoot great to see that you are interested to work on this issue, but i already fixed it. You can check the logic part and let me know if you need any help from myside. #24526.

PR #24526, opened 2022-02-17
[IMPROVE] No Scroll Down Button 

nishant23122000:

When the user scrolls up, then scroll down button will pop up on the bottom side. see the below video.

https://user-images.githubusercontent.com/53515714/154499894-a519544e-fcf8-4313-b161-217ef9a57cdd.mp4

There is also a **new message** button which will pop up, when a new message comes, the **scroll down** button will automatically close and give priority to the open **new message** button.

https://user-images.githubusercontent.com/53515714/154500165-b3938ffe-c869-4372-a672-181b7b1f305e.mp4

## Issue(s)

Issue Number :  #24523

## Steps to test or reproduce
Mentioned in #24523

## Further comments

I copy the style of the new-message class in the scroll-down class, because if I use the same then It causes issues in room.js **new-message** logics.

### DIFF
--- a/app/theme/client/imports/general/base_old.css
+++ b/app/theme/client/imports/general/base_old.css
@@ -1705,6 +1705,28 @@
 		}
 	}
 
+	& .scroll-down {
+		position: absolute;
+		z-index: 2;
+		bottom: 8px;
+		left: 50%;
+
+		width: 130px;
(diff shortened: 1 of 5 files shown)


nishant23122000:
@dougfabris Can you review this.

PR #24527, opened 2022-02-17
[NEW] Voip rooms endpoint

KevLehman:

### DIFF
--- a/app/livechat/client/voip.ts
+++ b/app/livechat/client/voip.ts
@@ -1,7 +1,7 @@
 import moment from 'moment';
 
 import { MessageTypes, IMessageType } from '../../ui-utils/client';
-import { IVoipMessage } from '../../../definition/IMessage';
+import { IMessage, isVoipMessage } from '../../../definition/IMessage';
 
 type IMessageFuncReturn = { at: string } | { at: string; time: string } | { comment: string } | { duration: string };
 
@@ -10,10 +10,13 @@ const messageTypes: IMessageType[] = [
 		id: 'voip-call-started',
 		system: true,
 		message: 'Voip_call_started',
-		data(message: IVoipMessage): IMessageFuncReturn {
-			const seconds = message?.voipData?.callWaitingTime || 0;
+		data(message: IMessage): IMessageFuncReturn {
+			if (!isVoipMessage(message)) {
+				return { at: '', time: '' };
+			}
+			const seconds = message.voipData.callWaitingTime || 0;
 			return {
-				at: message.voipData?.callStarted?.toString() || 'unknown date',
+				at: message.voipData.callStarted?.toString() || 'unknown date',
 				time: moment.duration(seconds, 'seconds').humanize(),
 			};
 		},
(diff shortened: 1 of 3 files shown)


app/api/server/v1/voip/rooms.ts
```diff
@@ -7,7 +7,21 @@ import { API } from '../../api';
 import { VoipRoom, LivechatVisitors, Users } from '../../../../models/server/raw';
 import { LivechatVoip } from '../../../../../server/sdk';
 import { ILivechatAgent } from '../../../../../definition/ILivechatAgent';
+import { hasPermission } from '../../../../authorization/server';
+import { typedJsonParse } from '../../../../../lib/typedJSONParse';
 
+type DateParam = { start?: string; end?: string };
```
tiagoevanp, line 13: Wait, are returning a empty object at the end... even if date is a DateParam instead a string?

app/api/server/v1/voip/rooms.ts
```diff
@@ -7,7 +7,21 @@ import { API } from '../../api';
...
+const validateDateParams = (property: string, date: DateParam | string): DateParam => {
```
tassoevan, line 14: This function is doing two things, which is a violation of the single-responsibility principle. Better split it into parsing and validation functions.

app/livechat/client/voip.ts
```diff
@@ -4,9 +4,7 @@ import { MessageTypes } from '../../ui-utils/client';
 import { IVoipMessage } from '../../../definition/IMessage';
 
 type IMessageFuncReturn = {
-	at?: string;
-	duration?: string;
-	time?: string;
+	[k: string]: string | undefined;
```
tassoevan, line 7: Here you're doing `type IMessageFuncReturn = Record<string, string | undefined>`, which is not good. I believe `IMessageFuncReturn` should be a union.

```ts
type IMessageFuncReturn =
  | {
    at: string | undefined;
    time: string;
  }
  | {
    duration: string;
  }
  | {
    at: string;
  }
  | {
    comment: string;
  }
```

Not ideal, but at least type-safe.

app/api/server/v1/voip/rooms.ts
```diff
@@ -176,7 +232,7 @@ API.v1.addRoute(
 			}
 			const language: string = rcSettings.get('Language') || 'en';
 			const comment = TAPi18n.__('Closed_by_visitor', { lng: language });
-			const closeResult = await LivechatVoip.closeRoom(visitor, room);
+			const closeResult = await LivechatVoip.closeRoom(visitor, room, this.user);
```
tiagoevanp, line 235: Where the endpoint get the comments/tags and send to voip service as params?

tiagoevanp (changes requested):

app/api/server/v1/voip/rooms.ts
KevLehman: Oh, that's a good thing :) didn't know we're following solid-esque principles now.

May I suggest to share that on the engineering channel instead of this PR? That way others can benefit too from new practices 😄 
Did we had a discussion about them 👀 ?

app/api/server/v1/voip/rooms.ts
KevLehman: oops

app/livechat/client/voip.ts
tiagoevanp: @KevLehman ?

KevLehman: Can you check again? I changed this to a union (not my personal favorite but works) and added some types to the MessageTypes class.

tiagoevanp (approved):

Issue #24482, opened 2022-02-11
[BUG]: Empty role is created in new role 


Himanshu664:
### Description:
when I create new role from permissions, it can be created empty.

### Steps to reproduce:

1. Go to Admin
2. Click on Permissions
3. Click on New Role
4. Save without giving role name
### Expected behavior:

Role name and  Role description should made mandatory.

### Actual behavior:

https://user-images.githubusercontent.com/76220055/153590526-d583398b-bc62-4b2b-a045-5811c77d99b6.mp4

### Server Setup Information:
Version of Rocket.Chat Server: 4.5.0-develop
Operating System: Linux ubuntu 20.04
NodeJS Version: v14.18.2
MongoDB Version: 4.4.4 / wiredTiger (oplog Enabled)

### Client Setup Information
Desktop App or Browser Version: Chrome 97.0.4692.99 (Official Build) (64-bit)
Operating System: Linux Ubuntu 20.04

prsndp:
@Himanshu664 I would like to work on this issue if it is open officially? thanks:)

J-Ankit2020:
@Himanshu664  I would like to work on this issue

dougfabris:
Already fixed here: #25063
Thanks for reporting @Himanshu664

PR #24532, opened 2022-02-17
[FIX] Fix regarding new role creation issue

Sayan-dev:

This solution provides a validator that checks the role fields and controls saving of form until filled up

## Issue(s)
#24482

## Steps to test or reproduce

Mentioned in the issue
## Further comments

I would have considered making a validator feature globally for all the forms, but I sufficed to make a local validator function that checks whether the required field is empty or not, similarly with the error text and input border color, a global feature possibility remains. This is a local solution, I would love to work to create a global solution.

# Final Result
https://user-images.githubusercontent.com/42710260/154562544-de07efa2-bde3-47e6-8d18-eabb30b6893f.mp4

### DIFF
--- a/client/views/admin/permissions/NewRolePage.js
+++ b/client/views/admin/permissions/NewRolePage.js
@@ -20,32 +20,49 @@ const NewRolePage = () => {
 		description: '',
 		scope: 'Users',
 		mandatory2fa: false,
+		error: { name: '' },
 	});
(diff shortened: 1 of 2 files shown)


dougfabris:
Hey @Sayan-dev 
We ended up solving it using a different approach here: #25063
Thanks a lot for your contribution!

PR #24534, opened 2022-02-17
[FIX] Validate room access

albuquerquefabio:

The request must be blocked If the user has no permission to view rooms.

## Issue(s)

You can access Direct Rooms even if "View Direct Room" permissions are disabled.
You can access Public Rooms even if "View Public Room" permissions are disabled.
You can access Private Rooms even if "View Private Room" permissions are disabled.
You can access Omnichannel Rooms even if "View Omnichannel Room" permissions are disabled.


server/services/authorization/canAccessRoom.ts
```diff
@@ -55,7 +74,16 @@ const roomAccessValidators: RoomAccessValidator[] = [
...
 			return false;
 		}
-		if (await Subscriptions.countByRoomIdAndUserId(room._id, user._id)) {
+
+		const hasViewPermission = await roomTypePermission(room.t, user._id);
+
+		const hasSubscriptions = await Subscriptions.countByRoomIdAndUserId(room._id, user._id);
+
+		if (!hasViewPermission) {
+			return false;
```
pierre-lehnen-rc, line 83: this will not block the access, it will only ignore the subscriptions.

albuquerquefabio: I did some changes and this [commit](https://github.com/RocketChat/Rocket.Chat/pull/24534/commits/b14482faa69e16143eff915b658a0cad44f572b5) can be the best approach.

server/services/authorization/canAccessRoom.ts
```diff
@@ -17,6 +18,23 @@ async function canAccessPublicRoom(user: Partial<IUser>): Promise<boolean> {
 	return Authorization.hasPermission(user._id, 'view-c-room');
 }
 
+const roomViewPermission = async (roomType: RoomType, uid: string): Promise<boolean> => {
```
matheusbsilva137, line 21:
```suggestion
const hasRoomViewPermission = async (roomType: RoomType, uid: string): Promise<boolean> => {
```
Maybe we could change this function name to improve legibility

server/services/authorization/canAccessRoom.ts
```diff
@@ -85,10 +106,11 @@ export const canAccessRoom: RoomAccessValidator = async (room, user, extraData):
 	// if (!room || !user) {
 	// 	return false;
 	// }
-
 	for await (const roomAccessValidator of roomAccessValidators) {
 		if (await roomAccessValidator(room, user, extraData)) {
-			return true;
+			if (await roomViewPermission(room.t, user._id)) {
```
matheusbsilva137, line 111:
```suggestion
			if (await hasRoomViewPermission(room.t, user._id)) {
```
If we decide to update the function's name, then we'd have to update it here too

matheusbsilva137 (changes requested):

server/services/authorization/canAccessRoom.ts
albuquerquefabio: Updated

matheusbsilva137 (dismissed):
LGTM

matheusbsilva137 (dismissed):

apps/meteor/server/services/authorization/canAccessRoom.ts
```diff
@@ -89,7 +109,9 @@ export const canAccessRoom: RoomAccessValidator = async (room, user, extraData):
 
 	for await (const roomAccessValidator of roomAccessValidators) {
 		if (await roomAccessValidator(room, user, extraData)) {
-			return true;
+			if (await hasRoomViewPermission(room.t, user._id)) {
```
pierre-lehnen-rc, line 112: There's no need to run this inside the validator loop; Instead, run it first and if there's no permission, return false;

Also: please double-check if this won't break livechat visitor access. I think it will.

apps/meteor/client/lib/rooms/roomTypes/livechat.ts
```diff
@@ -70,6 +70,9 @@ roomCoordinator.add(LivechatRoomType, {
 	},
 
 	findRoom(identifier) {
+		if (!hasPermission('view-l-room')) {
+			return null;
+		}
 		return ChatRoom.findOne({ _id: identifier });
```
pierre-lehnen-rc, line 76: returning null here will trigger a new request to the server to load the room data.

apps/meteor/client/lib/rooms/roomTypes/private.ts
```diff
@@ -88,6 +88,9 @@ roomCoordinator.add(PrivateRoomType, {
 	},
 
 	findRoom(identifier) {
+		if (!hasPermission('view-p-room')) {
```
pierre-lehnen-rc, line 91: same as above

apps/meteor/client/lib/rooms/roomTypes/public.ts
```diff
@@ -87,6 +87,9 @@ roomCoordinator.add(PublicRoomType, {
 	},
 
 	findRoom(identifier) {
+		if (!hasPermission('view-c-room')) {
```
pierre-lehnen-rc, line 90: same

apps/meteor/client/lib/rooms/roomTypes/voip.ts
```diff
@@ -24,6 +24,9 @@ roomCoordinator.add(VoipRoomType, {
 	},
 
 	findRoom(identifier) {
+		if (!hasPermission('view-l-room')) {
```
pierre-lehnen-rc, line 27: same

pierre-lehnen-rc (changes requested):

apps/meteor/client/lib/rooms/roomTypes/livechat.ts
albuquerquefabio: Done for all similar

pierre-lehnen-rc: Undefined will have the same effect. If the room is not found locally it'll be loaded from the server.

apps/meteor/client/lib/rooms/roomTypes/livechat.ts
```diff
@@ -70,6 +70,9 @@ roomCoordinator.add(LivechatRoomType, {
 	},
 
 	findRoom(identifier) {
+		if (!hasPermission('view-l-room')) {
+			return;
+		}
```
pierre-lehnen-rc, line 75:
```suggestion
```

It's better not to have this test at all. It'll only cause the room data to be loaded from the server every time.

apps/meteor/client/lib/rooms/roomTypes/private.ts
```diff
@@ -88,6 +88,9 @@ roomCoordinator.add(PrivateRoomType, {
...
+			return;
+		}
```
pierre-lehnen-rc, line 93:
```suggestion
```

It's better not to have this test at all. It'll only cause the room data to be loaded from the server every time.

apps/meteor/client/lib/rooms/roomTypes/public.ts
```diff
@@ -87,6 +87,9 @@ roomCoordinator.add(PublicRoomType, {
...
+			return;
+		}
```
pierre-lehnen-rc, line 92:
```suggestion
```

It's better not to have this test at all. It'll only cause the room data to be loaded from the server every time.

apps/meteor/client/lib/rooms/roomTypes/voip.ts
```diff
@@ -24,6 +24,9 @@ roomCoordinator.add(VoipRoomType, {
 	},
 
 	findRoom(identifier) {
+		if (!hasPermission('view-l-room') || !hasPermission('view-v-room')) {
+			return;
+		}
```
pierre-lehnen-rc, line 29:
```suggestion
```

It's better not to have this test at all. It'll only cause the room data to be loaded from the server every time.

apps/meteor/server/services/authorization/canAccessRoom.ts
```diff
@@ -85,6 +105,10 @@ export const canAccessRoom: RoomAccessValidator = async (room, user, extraData):
 	// 	return false;
 	// }
 
+	if (room && user) {
+		if (!(await hasRoomViewPermission(room.t, user._id))) return false;
+	}
+
```
pierre-lehnen-rc, line 111:
```suggestion
```
Do not **block** the access. Instead change each validator to stop **granting** the access when the permission is missing.

pierre-lehnen-rc (changes requested):

pierre-lehnen-rc (approved):

ggazzo (approved):

PR #24542, opened 2022-02-18
[FIX] Date Message Export Filter Fix

eduardofcabrera:

Fix message export filter to get all messages between "from date" and "to date", including "to date".

## Issue(s)

Date Message Export Filter wasn't considering the limit day.


app/api/server/v1/rooms.js
```diff
@@ -525,18 +525,23 @@ API.v1.addRoute(
...
+				dateTo.setDate(dateTo.getDate() + 1);
+
 				sendFile(
 					{
 						rid,
 						format,
-						...(dateFrom && { dateFrom: new Date(dateFrom) }),
-						...(dateTo && { dateTo: new Date(dateTo) }),
+						...(dateFrom && { dateFrom }),
+						...(dateTo && { dateTo }),
```
ostjen, line 544: based on this `sendFile` parameter type
```
type ExportFile = {
	rid: string;
	dateFrom: Date;
	dateTo: Date;
	format: 'html' | 'json';
};
```

I think you can send only the dates, without any destructuring

ostjen (changes requested):

ostjen (approved):

PR #24549, opened 2022-02-19
[FIX] Fixed css class names and comments

Biki-das:
I found a few css  class names and comments which had odd naming, this invalid class names may have resulted in styles not applied . this Pr fixes those and a fixes few comment typos


Biki-das:
@debdutdeb could you review?

Biki-das:
@murtaza98 can you review?

app/theme/client/imports/general/animations.css
```diff
@@ -104,7 +104,7 @@
 	}
 }
 
-@keyframes spinh {
+@keyframes spin {
```
debdutdeb, line 107: Don't think this is a typo, h for horizontal (rotation across Y axis)

debdutdeb (commented):
Thanks for the contribution! rest lgtm :) I'll ask someone who can confirm.

app/theme/client/imports/general/animations.css
Biki-das: ok sure will revert it

PR #24563, opened 2022-02-21
[NEW] Detect the abrupt disconnection of agent's client while in call to close the room

amolghode1981:

When the agent disconnects abruptly, the room should be closed. But because the agent is on browser, agent will not be
able to call the necessary APIs gracefully. The reason is that the browser may just crash or the agent may just force the tab close or refresh (Even though the client can prevent it on client side, it can not assure the graceful cleanup). So the solution is to detect this scenario on server and close the associated room. When such forceful tab close happens, Asterisk sends AMI event called 'ContactStatus' where the field contactstatus = 'Removed'
Connector will handle the ContactStatus event and if the contactstatus is removed, server will gracefully close the room on server.
Important thing to note is that the aor field givesthe extension record for which this event is raised. Connector will use this aor to detect the room for this extension.
Though this event is immediately raised, there is a theoretical race condition. The agent may have switched to another room meanwhile. But this is practically impossible to happen.

## Steps to test or reproduce
1. Make a call to Queue.
2. Let the agent accept the call.
3. Refresh/close/kill the browser.

## Further comments
Testing notes :
1. Verify that room is closed gracefully when such scenario happens.
2. Use test browser client or desktop phone and try the same scenario (Close the browser tab or kill the desktop client from task manager/activity monitor/kill comand). Verify the behaviour.


casalsgh:
Based on squads comments LGTM

PR #24565, opened 2022-02-21
[IMPROVE] Add OTR Room States

yash-rajpal:

Earlier OTR room uses only 2 states, we need more states to support future features. 
This adds more states for the OTR contextualBar.

- Expired
Screen Shot 2022-04-20 at 13 55 52

- Declined
Screen Shot 2022-04-20 at 13 49 28

- Error
Screen Shot 2022-04-20 at 13 55 26


apps/meteor/app/otr/client/rocketchat.otr.room.js
```diff
@@ -52,8 +61,6 @@ OTR.Room = class {
 	}
 
 	acknowledge() {
-		APIClient.v1.post('statistics.telemetry', { params: [{ eventName: 'otrStats', timestamp: Date.now(), rid: this.roomId }] });
```
dougfabris, line 55: it's a product decision to remove this one? 🤔

apps/meteor/app/otr/client/rocketchat.otr.room.js
```diff
@@ -72,15 +80,14 @@ OTR.Room = class {
...
 		this.reset();
+		this.setState(OtrRoomState.NOT_STARTED);
 		Notifications.notifyUser(this.peerId, 'otr', 'end', {
 			roomId: this.roomId,
 			userId: this.userId,
 		});
 	}
 
 	reset() {
-		this.establishing.set(false);
```
dougfabris, line 82: Unncessary?

apps/meteor/app/otr/client/rocketchat.otr.room.js
```diff
@@ -275,11 +285,11 @@ OTR.Room = class {
 
 				(async () => {
 					const { username } = await Presence.get(data.userId);
-					if (data.refresh && this.established.get()) {
+					if (data.refresh) {
```
dougfabris, line 288: shouldn't we check the established state?

apps/meteor/app/otr/client/rocketchat.otr.room.js
yash-rajpal: Before these new OTR States, reset function reseted the whole OTR room. 

But now we have to maintain different OTR states after reset and we want to know why was it reseted to show in contextual bar, OTR room can be rested and can be in many states like declined, timeout etc. So instead of updating state from inside reset function, I do it whenever reset is called with the state in which OTR is in.

apps/meteor/app/otr/client/rocketchat.otr.room.js
yash-rajpal: Removed by mistake, maybe while solving monorepo conflicts.

apps/meteor/app/otr/client/rocketchat.otr.room.js
yash-rajpal: Done!

dougfabris (approved):

PR #24577, opened 2022-02-22
[FIX] Jump To Recent Message Button Not Showing

nishant23122000:
Now when the user scrolls up to see old messages, jump to the recent message button will popup. 

https://user-images.githubusercontent.com/53515714/155124132-d9effe6b-c18b-4f87-bd05-a321862f07bc.mp4

 If the **new message** button is already there then **jump to recent button** will close.

https://user-images.githubusercontent.com/53515714/155124099-9e40cb15-ae4b-405a-877a-1eaac2fee001.mp4

## Issue(s)
#24523

## Steps to test or reproduce
mentioned in ##24523

### DIFF
--- a/app/ui/client/views/app/room.js
+++ b/app/ui/client/views/app/room.js
@@ -554,10 +554,12 @@ Meteor.startup(() => {
 				addClass.forEach((message) => $(`.messages-box #${message}`).addClass('selected'));
 			}
 		},
-		'click .jump-recent button'(e, template) {
+		'click .jump-recent button'(e) {
 			e.preventDefault();
-			template.atBottom = true;
-			RoomHistoryManager.clear(template && template.data && template.data._id);
+			const wrapperObj = document.getElementsByClassName('wrapper')[0];
+			wrapperObj.style.scrollBehavior = 'smooth';
+			const { scrollHeight, clientHeight } = wrapperObj;
+			wrapperObj.scrollTop = scrollHeight - clientHeight;
 		},
 		'load .gallery-item'(e, template) {
 			template.sendToBottomIfNecessary();
@@ -614,6 +616,10 @@ Meteor.startup(() => {
 					RoomHistoryManager.getMoreNext(this._id);
 				}
 			}
+			const justRecent = document.getElementsByClassName('jump-recent')[0];
+			const { scrollHeight, scrollTop, clientHeight } = e.target;
+			if (Math.ceil(scrollTop) < scrollHeight - clientHeight) justRecent.classList.remove('not');
+			else justRecent.classList.add('not');
 		}, 100),
 
 		'click .time a'(e) {
@@ -841,6 +847,7 @@ Meteor.startup(() => {
 
 		const wrapperUl = this.find('.wrapper > ul');
 		const newMessage = this.find('.new-message');
+		const justRecent = this.find('.jump-recent');
 
 		const template = this;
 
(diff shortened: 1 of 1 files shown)


app/ui/client/views/app/room.js
```diff
@@ -554,10 +554,12 @@ Meteor.startup(() => {
...
 		},
-		'click .jump-recent button'(e, template) {
+		'click .jump-recent button'(e) {
 			e.preventDefault();
-			template.atBottom = true;
-			RoomHistoryManager.clear(template && template.data && template.data._id);
+			const wrapperObj = document.getElementsByClassName('wrapper')[0];
+			wrapperObj.style.scrollBehavior = 'smooth';
+			const { scrollHeight, clientHeight } = wrapperObj;
+			wrapperObj.scrollTop = scrollHeight - clientHeight;
```
tassoevan, line 562: It doesn't work well when messages have to be loaded from server while going to the bottom.

nishant23122000: Hey, @tassoevan Can you check now, added a smooth scroll inside sentToBottom.

nishant23122000:
Hello @tassoevan , Can you check now?

tassoevan (dismissed):

nishant23122000:
@tassoevan Don't you think we should remove this [useEffect](https://github.com/RocketChat/Rocket.Chat/blob/d332c3674a1021032c19880f35330eccb20b9979/apps/meteor/client/views/room/components/body/RoomBody.tsx#L243) as there is no sense to call `sendToBottomIfNecessary` when section resizes.

Also calling `sendToBottom` when the window resizes is not a good user experience from my perspective.

tassoevan:
@nishant23122000 `sendToBottomIfNecessary` is supposed to invoke `sendToBottom` only if `atBottom` is `true`. I understand it as follows: if you scrolled to the end of the list of messages and the window shrinks, it should stick to the bottom.

nishant23122000:
@tassoevan I agree that the reason for putting `ResizeObserver` is to trigger when the window shrinks, but when the user slightly moves up from the latest message it triggers [this](https://github.com/RocketChat/Rocket.Chat/blob/303e8364a1a5e290505a3f35f8496e8e3fee4e06/apps/meteor/client/views/room/components/body/RoomBody.tsx#L251) line. and at that time `atBottom` is **false** so the user again moves to the latest message which is inconsistent. 

https://user-images.githubusercontent.com/53515714/212235648-425f2518-2a79-4a71-9974-1fb36cf4d9b4.mov

tassoevan:
@nishant23122000 Ouch, now I get it. Do you have a better alternative to keep the sticky behavior?

nishant23122000:
> @nishant23122000 `sendToBottomIfNecessary` is supposed to invoke `sendToBottom` only if `atBottom` is `true`. I understand it as follows: if you scrolled to the end of the list of messages and the window shrinks, it should stick to the bottom.

@tassoevan  If we only want to handle window shrink then we can use `addEventListener('resize')` because the current implementation only deals with `wrapperRef.current?.querySelector('ul')` instead we should focus on the whole window dimension changes.

nishant23122000:
> @nishant23122000 what about rescaling by toggling "expand" button on thread tab header?

@tassoevan It will not affect by these changes. Will work as it is.

tassoevan:
> It will not affect by these changes. Will work as it is.

Ok. Let's try it.

nishant23122000:
@tassoevan As I suggested a solution to use `resize` **addEventListener**, this approach will not work because this event will also trigger when the **thread** is open and we **resize** the window.

seems like replacing `sendToBottomIfNecessary` with `sendToBottom` will be the best solution to remove that **inconsistency** in scrolling.

This resize observer was triggering multiple times because of **sendToBottomIfNecessary** ( Collision of some `useEffect` dependencies 🥲 and dependencies chaining ).

nishant23122000:
@tassoevan any update on this? let me know if any changes are required.

PR #24578, opened 2022-02-22
[FIX] - unread message alert

filipemarins:
## Further comments

Fix unread message alert behavior


ee/server/services/ecdh-proxy/lib/server.ts
```diff
@@ -5,6 +5,7 @@ import { Readable } from 'stream';
 import WebSocket from 'ws';
 import cookie from 'cookie';
 import express, { Request, Response } from 'express';
+// eslint-disable-next-line import/no-unresolved
```
gabriellsh, line 8: Hmmm, I wouldn't change this file in this PR. I usually only generate errors in the local checks, but the CI runs ok. Let's remove this change and if needed implement it in a different PR directly to base.

gabriellsh (changes requested):

PR #24583, opened 2022-02-22
Regression: Error when trying to load name of dm rooms for avatars and notifications

pierre-lehnen-rc:


server/lib/rooms/roomTypes/direct.ts
```diff
@@ -53,7 +61,13 @@ roomCoordinator.add(DirectMessageRoomType, {
 				return undefined;
 			}
 
-			return Subscriptions.findOneByRoomIdAndUserId(room._id, Meteor.userId());
+			const uid = userId || getCurrentUserId();
+			if (uid) {
+				return Subscriptions.findOneByRoomIdAndUserId(room._id, uid);
```
sampaiodiego, line 66: from what I can see only `fname` and `name` fields are being used from the returned subscription, so it would be good to only return them from the find:
```suggestion
				return Subscriptions.findOneByRoomIdAndUserId(room._id, uid, { fields: { name: 1, fname: 1 } });
```

server/lib/rooms/roomTypes/direct.ts
```diff
@@ -53,7 +61,13 @@ roomCoordinator.add(DirectMessageRoomType, {
...
+			}
+
+			// If we don't know what user is requesting the roomName, then any subscription will do
+			return Subscriptions.findOne({ rid: room._id });
```
sampaiodiego, line 70: same as above
```suggestion
			return Subscriptions.findOne({ rid: room._id }, { fields: { name: 1, fname: 1 } });
```

server/lib/rooms/roomTypes/livechat.ts
```diff
@@ -22,16 +22,16 @@ roomCoordinator.add(LivechatRoomType, {
...
 		return room.name || room.fname || room.label;
 	},
 
 	canAccessUploadedFile({ rc_token: token, rc_rid: rid }): boolean {
 		return token && rid && LivechatRooms.findOneOpenByRoomIdAndVisitorToken(rid, token);
 	},
 
-	getNotificationDetails(room: IRoom, _user, notificationMessage: string) {
-		const title = `[Omnichannel] ${this.roomName(room)}`;
+	getNotificationDetails(room: IRoom, _sender, notificationMessage: string, userId?) {
```
sampaiodiego, line 33: how did TS not screamed on the missing type?
```suggestion
	getNotificationDetails(room: IRoom, _sender: unknown, notificationMessage: string, userId?: string) {
```

sampaiodiego (changes requested):

server/lib/rooms/roomTypes/livechat.ts
pierre-lehnen-rc: The function signature is already typed on the `IRoomTypeServerDirectives` interface.

albuquerquefabio (dismissed):

server/lib/rooms/roomTypes/livechat.ts
tassoevan: Wipe the all parameter types out to keep consistency.

sampaiodiego (approved):

PR #24590, opened 2022-02-23
[FIX] Duplicated 'name' log key

sampaiodiego:

Production logs (in JSON format) had a duplicated `name` key. While this is ignored in JS it may be a problem for different JSON parsers.

Before:
```
I20220223-00:10:38.662(-3)? {"level":51,"time":"2022-02-23T03:10:38.662Z","pid":13033,"hostname":"Diegos-MacBook-Pro.local","name":"","name":"Migrations","msg":"Not migrating, already at version 256"}
```

After:
```
I20220223-00:10:38.662(-3)? {"level":51,"time":"2022-02-23T03:10:38.662Z","pid":13033,"hostname":"Diegos-MacBook-Pro.local","name":"Migrations","msg":"Not migrating, already at version 256"}

```

### DIFF
--- a/server/lib/logger/getPino.ts
+++ b/server/lib/logger/getPino.ts
@@ -21,7 +21,6 @@ const mainPino = pino({
 		subscription: 35,
 		startup: 51,
 	},
-	name: '',
 	level: 'warn',
 	timestamp: pino.stdTimeFunctions.isoTime,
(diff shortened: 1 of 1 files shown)


debdutdeb (approved):
🥂

PR #24594, opened 2022-02-23
Regression: Bunch of settings fixes for VoIP

MartinSchoeler:
This fixes: 
- Misspellings
- Renaming VoIP section to Call Center
- Removed default values from settings
- Added Password field to management password setting

Pls note that you might still find an empty 'voip' section in your settings if you used a previous develop version, this is normal as we use collections for our settings


app/lib/server/startup/settings.ts
```diff
@@ -3219,8 +3219,8 @@ settingsRegistry.addGroup('VoIP', function () {
 				},
 			});
 
-			this.add('VoIP_Management_Server_Password', 'secure_password', {
-				type: 'string',
+			this.add('VoIP_Management_Server_Password', '', {
```
KevLehman, line 3222: These values were left as references of what should be on those settings 🤔 any reason to remove them?

MartinSchoeler: Since they were actual values instead of placeholders, we don't usually use values like this

KevLehman: Fair enough, can we add some indication for the format of hostnames as placeholder text? So they know where to put a domain url (a.b.c) or a websocket url 👀

KevLehman (approved):

PR #24607, opened 2022-02-24
[FIX] VoIP Enable/Disable setting on CallContext/CallProvider Notifications

tiagoevanp:


client/providers/CallProvider/CallProvider.tsx
```diff
@@ -85,13 +84,15 @@ export const CallProvider: FC = ({ children }) => {
...
 
 	useEffect(() => {
-		Notifications.onUser('callerjoined', handleQueueJoined);
-		Notifications.onUser('agentcalled', handleAgentCalled);
-		Notifications.onUser('agentconnected', handleAgentConnected);
-		Notifications.onUser('queuememberadded', handleMemberAdded);
-		Notifications.onUser('queuememberremoved', handleMemberRemoved);
-		Notifications.onUser('callabandoned', handleCallAbandon);
-		Notifications.onUser('call.callerhangup', handleCallHangup);
+		if (voipEnabled) {
```
KevLehman, line 87: What if i enable voip, the events get registered, and then i disable them? Will the event listeners be destroyed? 👀

tiagoevanp: ok... fixing

client/providers/CallProvider/CallProvider.tsx
```diff
@@ -143,23 +129,58 @@ export const CallProvider: FC = ({ children }) => {
...
-		handleAgentConnected,
-		handleCallHangup,
-		handleAgentCalled,
-	]);
+		if (!voipEnabled || !user) {
+			return;
+		}
+
+		const unsubscribeFromCallerJoined = subscribeToNotifyUser(`${user._id}/callerjoined`, handleQueueJoined);
+		return unsubscribeFromCallerJoined;
```
tassoevan, line 137: image

```suggestion
		return subscribeToNotifyUser(`${user._id}/callerjoined`, handleQueueJoined);
```

tassoevan (approved):

KevLehman:
@tiagoevanp , not sure if this should be handled on this PR, but, if the socket is already stablished, and then you disable voip, the ws continues receiving messages 👀 (it's not finished)

Rest looks good

tiagoevanp:
> 

Thanks for raise this @KevLehman!

We both agree on proceed with the merge and work in another time with the solution for this different problem found.

PR #24612, opened 2022-02-24
[FIX] Rooms' names turn lower case on CSV import

guijun13:

* Change 'Settings' import to not get cached configs
* Remove update `UI_Allow_room_names_with_special_chars` value


apps/meteor/app/utils/lib/getValidRoomName.js
```diff
@@ -2,15 +2,14 @@ import { Meteor } from 'meteor/meteor';
...
-import { Rooms } from '../../models';
+import { Settings, Rooms } from '../../models';
 import { validateName } from '../../lib/server/functions/validateName';
 
 export const getValidRoomName = (displayName, rid = '', options = {}) => {
 	let slugifiedName = displayName;
 
-	if (settings.get('UI_Allow_room_names_with_special_chars')) {
-		const cleanName = limax(displayName);
+	if (Settings.findOneById('UI_Allow_room_names_with_special_chars').value) {
```
matheusbsilva137, line 11: Since we're not changing the value of this setting before starting the importing process (due to the changes you've made in the `ImporterBase.js` file), I guess there's no problem on using the cached settings here. wdyt? Maybe we could keep using the `settings` import in this file

matheusbsilva137 (changes requested):

apps/meteor/app/utils/lib/getValidRoomName.js
guijun13: I agree with you @matheusbsilva137

matheusbsilva137 (approved):

PR #24622, opened 2022-02-25
Regression: Fix Wrap Up is not completing successfully

murtaza98:
## Issue(s)

- When the agent enters a new tag via Tag Input the tag value doesn't appear under the input component as it happens on the chat close Modal(Only happens when the agent is able to enter unregistered tags)  
- It seems neither notes nor tags are being stored in the room object.  
- Cancel: Wrap Up modal is closed but the room panel doesn't disappear (Unexpected behavior)


KevLehman:
Closing in favor of Martin's PR

PR #24629, opened 2022-02-25
[FIX] "Match error" when converting a team to a channel

matheusbsilva137:

- Fix "Match error"  when trying to convert a channel to a team;

## Steps to test or reproduce

1. Create a team and add some rooms to it;
2. Go to **Room Information > Convert to Channel**. No errors should be displayed and the list of rooms should be shown.

### DIFF
--- a/app/api/server/v1/teams.ts
+++ b/app/api/server/v1/teams.ts
@@ -356,7 +356,8 @@ API.v1.addRoute(
 				return API.v1.unauthorized();
 			}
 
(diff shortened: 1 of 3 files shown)


ostjen (commented):
isn't there a way to still use booleans? for me it doesn't make sense converting to string and write checks like `canUserDelete === 'true'`

ostjen (approved):
got it! lgtm

PR #24683, opened 2022-03-03
[FIX] no id of room closer in livechat-close message

cuonghuunguyen:
In `app/apps/server/bridges/livechat.ts:114`, we converted the user to app interface and the use that user for Rocket.Chat method. Due to the unmatch interface, this cause some troubles.
We should use method `convertToRocketChat` instead


debdutdeb (approved):
lgtm!

murtaza98 (approved):

PR #24708, opened 2022-03-04
Chore: Cancel running jobs if PR is updated

debdutdeb:
Co-authored-by: Thassio Victor <tvmcarvalho@gmail.com>

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -10,7 +10,7 @@ on:
       - develop
 
 concurrency:
-  group: ${{ github.workflow }}-${{ github.head_ref }}
+  group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
   cancel-in-progress: true
 
 env:
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -9,6 +9,10 @@ on:
     branches:
       - develop
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref }}
+  cancel-in-progress: true
+
 env:
   CI: true
   MONGO_URL: mongodb://localhost:27017


sampaiodiego:
that's awesome.. how can we test it? force push? develop merge?

sampaiodiego:
looks like it works 😃 

looking into GH docs, I found their example using `group: ${{ github.workflow }}-${{ github.ref }}` https://docs.github.com/en/actions/using-jobs/using-concurrency#example-only-cancel-in-progress-jobs-or-runs-for-the-current-workflow

any special reason you've used `github.head_ref` instead of `github.ref`? [I found](https://docs.github.com/en/actions/learn-github-actions/contexts) `head_ref` is available only for pull requests.

debdutdeb:
IIUC `github.ref` will also cancel any running workflows in favor of the latest commit (in develop). We want that behavior only in prs not on develop (where each commit should have its own run). 

That said I see a flaw here in this pr along the same line - let ne fix that 🙈

.github/workflows/build_and_test.yml
```diff
@@ -9,6 +9,10 @@ on:
     branches:
       - develop
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref }}
```
debdutdeb, line 13:
```suggestion
  group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
```

sampaiodiego:
> IIUC `github.ref` will also cancel any running workflows in favor of the latest commit (in develop). We want that behavior only in prs not on develop (where each commit should have its own run).

oh.. how did you catch that? how can we test it?

debdutdeb:
I'm just going to create a private repo to test it 😬

KevLehman:
I love this

debdutdeb:
@sampaiodiego check out the newest private repo (in our org) - see the actions. The very first one, cancelled, was with `github.ref`, after the second commit. The next one didn't get dropped once changed it to head_ref ||  run_id 😬

sampaiodiego:
playing with the other repo I found using `group: ${{ github.workflow }}-${{ github.ref }}` had the best and expected results 🤔

.github/workflows/build_and_test.yml
```diff
@@ -9,6 +9,10 @@ on:
     branches:
       - develop
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
```
sampaiodiego, line 13:
```suggestion
  group: ${{ github.workflow }}-${{ github.ref }}
```

sampaiodiego (changes requested):

.github/workflows/build_and_test.yml
debdutdeb: Found this, https://github.com/nodejs/node/blob/818284b68737bbb9098cf12b1b221cbb24f06ce8/.github/workflows/build-tarball.yml#L27-L29

Not sure about just using .ref 🙈 I'm using the other one in here https://github.com/RocketChat/RC4Community/blob/d7c98d2eea15100b58adc3a209224526e59fb232/.github/workflows/build_and_test.yml#L8-L10 and so far working as expected.

PR #24709, opened 2022-03-04
[FIX] API Error preventing adding an email to users without one (like bot/app users)

debdutdeb:

### DIFF
--- a/app/lib/server/functions/saveUser.js
+++ b/app/lib/server/functions/saveUser.js
@@ -207,7 +207,7 @@ export function validateUserEditing(userId, userData) {
 	}
 
 	if (
-		user.emails[0] &&
+		user.emails?.[0] &&
 		isEditingField(user.emails[0].address, userData.email) &&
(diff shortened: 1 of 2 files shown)


sampaiodiego (commented):
I need to ask for a test case 😬

debdutdeb:
Hm but the api doesn't allow no-email user creation. Not sure how can I add an e2e for this.

sampaiodiego:
> Hm but the api doesn't allow no-email user creation. Not sure how can I add an e2e for this.

true.. I thought about that after I send the comment.. 😞

sampaiodiego:
ohh. you can create a test case adding an email to `rocket.cat` user.

sampaiodiego (approved):

dudanogueira:
This seems to also be affecting slack imports:
https://forums.rocket.chat/t/cannot-read-property-0-of-undefined-when-editing-user/13604

PR #24751, opened 2022-03-08
Chore: Update Site Url Description. No trailing slash.

dudanogueira:
Change site url description to not suggest leaving the trailing slash.

## Steps to test or reproduce
- Set admin > general > site url with a trailing slash
- restart the server (yes, necessary)
- go to http://localhost:3000/livechat
- Live chat will not load or not behave as expected

### DIFF
--- a/packages/rocketchat-i18n/i18n/af.i18n.json
+++ b/packages/rocketchat-i18n/i18n/af.i18n.json
@@ -2298,7 +2298,7 @@
   "since_creation": "sedert%s",
   "Site_Name": "Webbladnaam",
   "Site_Url": "Site-URL",
-  "Site_Url_Description": "Voorbeeld: https://chat.domain.com/",
+  "Site_Url_Description": "Voorbeeld: https://chat.domain.com",
   "Size": "grootte",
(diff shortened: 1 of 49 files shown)


debdutdeb:
Hey @dudanogueira - what was the issue exactly?

dudanogueira:
> Hey @dudanogueira - what was the issue exactly?

it will fail to load the livechat widget, or when loading with a cached js in browser, it will work erratically:
https://forums.rocket.chat/t/not-sync-at-livechat-page/12917

debdutdeb:
Hm. But this isn't really a fix though. This isn't a documentation error rather a bug in our code.

dudanogueira:
Agreed! We should fix both.

Nice PR ;)
https://github.com/RocketChat/Rocket.Chat/pull/24762

Also, it would also be nice to have the trailing slash removed while saving, or validating it before saving.

debdutdeb:
> Also, it would also be nice to have the trailing slash removed while saving, or validating it before saving.

I do agree. Maybe another pr someday :)

PR #24789, opened 2022-03-10
[FIX] VoIP button gets disabled whenever user status changes

amolghode1981:
Whenever the user state is changed (Explicitly or automatically by the virtue of user's inactivity)
Voip component gets reloaded. Also during the reload of the page, the Voip component gets initialised multiple times, causing multiple websockets opened to asterisk.
This happens because of the following facts:

1. The useEffect of useVoipClient, which creates the Voip component had user in its dependency list.
2. Whenever any user property changes, it would cause the useEffect to trigger and execute the entire initialisation of voip component.
3. When state change happens, it triggers update on all the components. So OmnichannelCallToggleReady gets re-rendered, also useVoipClient causes reinitialisation (user state change).
4. Now, when the useEffect of OmnichannelCallToggleReady gets executed, voipClient.getRegistrarState() returns the value which is not |registered| because the state change has caused reinitialisation of the voip client  which is not yet registered.
5. So again, the phone button value gets reset and disabled, the agent becomes unavailable.

Fix:
Even though the user state may change, user identity does not change. SIP client reinitialisation should happen only when userId changes.
So in summary the fix is not to add the entire user Object in the dependency list of the useEffect of useVoipClient hook in useVoipClient hook file. Instead

1. Create a state userId.
2. Add another useEffect watching the user. Whenever user becomes available or any user change happens, update the userId
3. The useEffect which does the voip client initialisation, should have the userId in the dependency list.

When this is done, the Voip initialisation will happen only when userId changes.

## Issue(s)
[](https://app.clickup.com/t/25jwe53)

## Steps to test or reproduce

1. Configure Voip, assign extension to the user, login to the extension
2. Change the state from available to busy, this should disable the button.
3. login to the extension.
4. Goto My Account -> Preferences -> User Preferences
5. change the "Idle Time Limit" to some small value (1 min is minimum)
6. Do not interact with the page for 1 minute, whenever the user becomes idle, it should disable the voip button.

## Further comments
Test cases :
Configure Voip, Assign extension to the user.
1. When page is loaded, refresh the page couple of times. Verify on asterisk that you see "_WebSocket connection from 'Your public IP:port' for protocol 'sip' accepted using version '13'_" is displayed only once per reload.
2. Change the user state explicitly. Verify that if the user is available for taking calls, he/she remains in the same state. Verify this by calling the user.
3. Do not interact with the agent's ui for the timelimit mentioned in above section ("Idle Time Limit"). Verify that if the user is available for taking calls, he/she remains in the same state. Verify this by calling the user.

### DIFF
--- a/client/providers/CallProvider/hooks/useVoipClient.ts
+++ b/client/providers/CallProvider/hooks/useVoipClient.ts
@@ -35,15 +35,16 @@ export const useVoipClient = (): UseVoipClientResult => {
 	const user = useUser();
 	const iceServers = useWebRtcServers();
 
+	const [userId, setUserId] = useSafely(useState<string | null>(null));
 	const [result, setResult] = useSafely(useState<UseVoipClientResult>({}));
 
 	useEffect(() => {
-		if (!user || !user?._id || !voipEnabled) {
+		if (!userId || !voipEnabled) {
 			setResult({});
 			return;
 		}
 
-		registrationInfo({ id: user._id }).then(
+		registrationInfo({ id: userId }).then(
 			(data) => {
 				let parsedData: IRegistrationInfo;
 				if (isSignedResponse(data)) {
@@ -82,7 +83,14 @@ export const useVoipClient = (): UseVoipClientResult => {
 			// client?.disconnect();
 			// TODO how to close the client? before creating a new one?
 		};
-	}, [user, iceServers, registrationInfo, setResult, membership, voipEnabled]);
+	}, [userId, iceServers, registrationInfo, setResult, membership, voipEnabled]);
 
+	useEffect(() => {
+		if (!user || !user?._id) {
+			setResult({});
+			return;
+		}
+		setUserId(user._id);
+	}, [setResult, setUserId, user]);
 	return result;
 };


MartinSchoeler (changes requested):
If we only need the userId, instead of using `useUser` we can use `useUserId` from the same file, this way we don't need to create additional useStates, nor watch changes to the user object with `useEffect`

Take a look at https://github.com/RocketChat/Rocket.Chat/blob/new/reload-issue/client/contexts/UserContext.ts#L70

amolghode1981:
> If we only need the userId, instead of using `useUser` we can use `useUserId` from the same file, this way we don't need to create additional useStates, nor watch changes to the user object with `useEffect`
> 
> Take a look at https://github.com/RocketChat/Rocket.Chat/blob/new/reload-issue/client/contexts/UserContext.ts#L70

Fixed.

client/providers/CallProvider/hooks/useVoipClient.ts
```diff
@@ -32,18 +32,17 @@ export const useVoipClient = (): UseVoipClientResult => {
 	const voipEnabled = useSetting('VoIP_Enabled');
 	const registrationInfo = useEndpoint('GET', 'connector.extension.getRegistrationInfoByUserId');
 	const membership = useEndpoint('GET', 'voip/queues.getMembershipSubscription');
-	const user = useUser();
+	const userId = useUserId();
```
KevLehman, line 35: Hey 👀 not sure of a different approach here, but, I did use "user.extension" prop on other fix to prevent this call from happening if the user doesn't have an extension associated.

But, that breaks the fix 🤔 since I was using the "useUser" hook, unless we have a "useUserExtension" hook somewhere (or we create one)

amolghode1981: @KevLehman Can you please point me to the task? There are few of such kind which are dependent.

amolghode1981: As @KevLehman pointed out, the fix was breaking https://github.com/RocketChat/Rocket.Chat/pull/24752. Next commit fixes this.

KevLehman (approved):

PR #24790, opened 2022-03-10
[FIX] Sequential messages

gabriellsh:

### DIFF
--- a/client/views/room/MessageList/MessageList.tsx
+++ b/client/views/room/MessageList/MessageList.tsx
@@ -25,7 +25,7 @@ export const MessageList: FC<{ rid: IRoom['_id'] }> = ({ rid }) => {
 	const messages = useMessages({ rid });
 	const subscription = useUserSubscription(rid);
 	const isBroadcast = Boolean(subscription?.broadcast);
-	const messageGroupingPeriod = useSetting('Message_GroupingPeriod') as number;
+	const messageGroupingPeriod = Number(useSetting('Message_GroupingPeriod'));
 	const format = useFormatDate();
 
 	return (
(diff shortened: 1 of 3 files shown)


client/views/room/MessageList/MessageList.tsx
```diff
@@ -24,6 +25,7 @@ export const MessageList: FC<{ rid: IRoom['_id'] }> = ({ rid }) => {
 	const messages = useMessages({ rid });
 	const subscription = useUserSubscription(rid);
 	const isBroadcast = Boolean(subscription?.broadcast);
+	const messageGroupingPeriod = useSetting('Message_GroupingPeriod') as number;
```
filipemarins, line 28: maybe this?

```suggestion
	const messageGroupingPeriod = Number(useSetting('Message_GroupingPeriod'));
```

filipemarins, line 28: Maybe this?

```suggestion
	const messageGroupingPeriod = Number(useSetting('Message_GroupingPeriod'));
```

client/views/room/MessageList/lib/isMessageSequential.ts
```diff
@@ -1,15 +1,14 @@
-import { differenceInMinutes } from 'date-fns';
+import { differenceInSeconds } from 'date-fns';
 
+import { MessageTypes } from '../../../../../app/ui-utils/client';
 import { IMessage } from '../../../../../definition/IMessage';
 
-// TODO: use settings config to define time range instead of hardcoded value
-
-export const isMessageSequential = (current: IMessage, previous: IMessage | undefined): boolean => {
+export const isMessageSequential = (current: IMessage, previous: IMessage | undefined, groupingRange: number): boolean => {
```
filipemarins, line 6: maybe a default value to groupingRage?

filipemarins (approved):

client/views/room/MessageList/lib/isMessageSequential.ts
gabriellsh: Hmmm, I think we don't intend this function to be called without this groupingRange, specially because it depends on a setting. Since there's no real use for it outside this scope, I think it's better to narrow it's use as much as possible.

Issue #23041, opened 2021-08-26
Error when creating a new chat with a user in the general channel (createDirectMessage error)


IvanOpanasiuk:
### Description:

In the channel 'general' I press to create a chat with a user with whom there was no chat yet. And I get an error 'createDirectMessage'. 

### Steps to reproduce:

1. Go to channel general
2. Click on User (click on a user with whom there is no chat for the createDirectMessage method to work)
3. Click on 'Direct Message' button

<img width="545" alt="Снимок экрана 2021-08-26 в 11 09 35" src="https://user-images.githubusercontent.com/15374759/130926121-4c99ef8d-4223-4cce-a898-5c26abb1a785.png">

### Actual behavior:

I get a dark screen and the console shows that there is an error in the createDirectMessage method

<img width="1769" alt="Снимок экрана 2021-08-26 в 11 10 22" src="https://user-images.githubusercontent.com/15374759/130926258-b9306adc-0f3b-4ab6-a1e3-f53134a527cb.png">

After page reload everything works

I can assume that the problem is that in the 'createDirectMessage' method the 'Users.findOneByUsernameIgnoringCase()' is executed and returns 'undefined'

<img width="1045" alt="Снимок экрана 2021-08-26 в 11 20 31" src="https://user-images.githubusercontent.com/15374759/130928256-040f0a5c-2fd0-4e29-bd67-3b411e0f1f95.png">

https://user-images.githubusercontent.com/15374759/130930409-36fdfbcf-b63f-4514-9f89-7fdb2375707c.mov

### Server Setup Information:

- Version of Rocket.Chat Server: 3.17.1
- Operating System: mac macOS
- NodeJS Version: 14.16.1
- MongoDB Version: 4.4.0

### Client Setup Information

-  Browser Version:
- Operating System: macOS

### Relevant logs:

<img width="1157" alt="Снимок экрана 2021-08-26 в 11 15 19" src="https://user-images.githubusercontent.com/15374759/130927015-55196a3d-dddb-43bc-a6cf-0714a85a6445.png">

<img width="1790" alt="Снимок экрана 2021-08-26 в 11 16 28" src="https://user-images.githubusercontent.com/15374759/130927126-370c8f97-1d76-46e9-816b-849c81a02c3e.png">

debdutdeb:
I just tried to reproduce this on 3.17.2 and I couldn't reproduce this. @ankar84 have you seen this issue on your end?

ankar84:
[...]

PR #24623, opened 2022-02-25
[FIX] Opening a new DM from user card

tassoevan:
## Proposed changes (including videos or screenshots)

A race condition on `useRoomIcon` -- delayed merge of rooms and subscriptions -- was causing a UI crash whenever someone tried to open a DM from the user card component.

## Issue(s)

Closes #24579

PR #24797, opened 2022-03-10
[FIX] Create direct message from search

debdutdeb:
## Further comments

Probably not the best solution but it works :/

### DIFF
--- a/app/models/server/models/Users.js
+++ b/app/models/server/models/Users.js
@@ -768,6 +768,21 @@ export class Users extends Base {
 		return this.findOne(query, options);
 	}
 
+	findOneByIdOrUsernameIgnoringCase(idOrUsername, options = {}) {
+		const query = {
+			$or: [
+				{
+					_id: idOrUsername,
+				},
+				{
+					username: new RegExp(`^${escapeRegExp(idOrUsername)}$`, 'i'),
+				},
+			],
+		};
+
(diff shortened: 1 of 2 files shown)


ankar84:
Hey Deb!
What that PR should fix?
Creating DM was fixed in https://github.com/RocketChat/Rocket.Chat/pull/24623
Is your case still not fixed?

debdutdeb:
Screenshot_2022-03-11-11-49-36-157_com android chrome
No looks like it hasn't.

debdutdeb:
Closes #23041

ankar84:
> No looks like it hasn't.

Try develop branch.
I tested - all works perfect. 
Fix not in 4.5.1

debdutdeb:
Ah crap. Thanks for confirming :P

debdutdeb:
@ankar84 can you double check please? I just heard that this is not working in 4.5.2 :/

ankar84:
> @ankar84 can you double check please? I just heard that this is not working in 4.5.2 :/

I will do it today, because it was all perfect in develop branch...

debdutdeb:
Hm thanks. I'm outside so can't test it rn.

pierre-lehnen-rc:
The search result was generating a link with the user id instead of the username, but that was fixed on #24955, so this change will no longer be necessary.

PR #24798, opened 2022-03-10
[FIX] Add title with date and time

filipemarins:


ankar84:
Could be solution for https://github.com/RocketChat/feature-requests/issues/482 if I get that PR right

PR #24804, opened 2022-03-11
[FIX] "livechat/webrtc.call" endpoint not working

murtaza98:
## Issue(s)

We're getting the following error when calling the `livechat/webrtc.call` endpoint.
`TypeError: Cannot destructure property 'businessUnit' of 'undefined' as it is undefined.`


cauefcr (approved):
Looks good to me.

PR #24805, opened 2022-03-11
[FIX] Critical: Incorrect visitor getting assigned to a chat from apps

murtaza98:
## Steps to test or reproduce

**Prerequisite:**  Server with the Facebook app installed and set up. Plus the FB account should have at least 2 page.
1. Customer sends a message to a FB Page.
2. RC displays it on the Queue 
3. RC Agent takes it and a room is started
4. Another customer sends another message from a different FB Page.
5. RC displays it on the Queue 
6. The new customer goes to the same room previously created
7. RC Agent sends a message and only the first customer gets it.

Expected Behavior:
FB App should create an individual room for any new customer from any FB Page.

Actual Behavior:
FB App creates a single room for all customers from any FB Page associated with the FB

### DIFF
--- a/app/livechat/server/lib/Livechat.js
+++ b/app/livechat/server/lib/Livechat.js
@@ -318,7 +318,7 @@ export const Livechat = {
 		if (user) {
 			Livechat.logger.debug('Found matching user by token');
 			userId = user._id;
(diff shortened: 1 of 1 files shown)


cauefcr (approved):
Looks good, let's have QA check it out.

PR #24806, opened 2022-03-11
[NEW][APPS] Ability for Rocket.Chat Apps to delete messages

kulkarnigaurav38:

Added a new delete method on the messages bridge for deleting messages via the Apps-Engine.

### DIFF
--- a/app/apps/server/bridges/messages.ts
+++ b/app/apps/server/bridges/messages.ts
@@ -50,6 +50,11 @@ export class AppMessageBridge extends MessageBridge {
 		updateMessage(msg, editor);
 	}
 
+	protected async delete(messageId: string, appId: string): Promise<void> {
+		this.orch.debugLog(`The App ${appId} is deleting a message.`);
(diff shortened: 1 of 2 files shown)


app/apps/server/bridges/messages.ts
```diff
@@ -50,6 +50,11 @@ export class AppMessageBridge extends MessageBridge {
 		updateMessage(msg, editor);
 	}
 
+	protected async delete(messageId: string, appId: string): Promise<void> {
+		this.orch.debugLog(`The App ${appId} is deleting a message.`);
+		Messages.removeById(messageId);
```
murtaza98, line 55: We cannot delete the database entry directly from here. There are a couple of checks that need to be performed before we actually delete a message from DB. You can take a look at this file to understand more about this: `app/lib/server/functions/deleteMessage.ts`.

d-gubert (changes requested):
The change mentioned by @murtaza98 needs to be applied

Issue #24816, opened 2022-03-13
[BUG] Opening Multiple VRDialog With Different Timer


nishant23122000:
### Steps to reproduce:

1. Go To any room, click on the + icon in the message input field.
2. click on Video Message
3. Start recording
4. Again click on the + icon(without closing the VR dialog)  it will close the VR dialog.
5. Again Open the VR Dialog and start the recording.
6. It will start the second VR dialog timer from 0 seconds. 

### Expected behavior:
When the user again clicks on the + button, we should reset and remove the old state of the VR dialog, otherwise, multiple dialogs will run parallelly.

### Actual behavior:
When we click + button without closing the VR dialog, and then if we open it again, 2 parallel dialogs work with a different timer. and if we send this video, then it gives a file empty error.

https://user-images.githubusercontent.com/53515714/158059295-1c1be992-b904-435e-b187-dc21105154d7.mp4

### Server Setup Information:

- Version of Rocket.Chat Server: 4.6.0
- Operating System: Windows 10
- Deployment Method: 
- Number of Running Instances: 
- DB Replicaset Oplog: 
- NodeJS Version: 14.18.3
- MongoDB Version: 4.4.4

### Client Setup Information

- Desktop App or Browser Version:
- Operating System:

### Additional context
getting file empty error if we click timer multiple times.

https://user-images.githubusercontent.com/53515714/158059301-2669a05b-8ed5-40f0-b206-4940f2dd6deb.mp4

dougfabris:
Hi there, thanks for the contribution! 🚀 💯 

Sorry, this error does not happen on the latest and it was fixed here: #27850
[Rocket.Chat](https://rocket.chat/)  provides support for, at least, 3 months for each release and 6 months for the latest releases of each major version

I would suggest you update your Rocket.Chat version to the latest stable version, which you can find in [Releases Page](https://github.com/RocketChat/Rocket.Chat/releases) and, check it [HERE](https://docs.rocket.chat/resources/get-support/enterprise-support#rocket.chat-versions) the table with versions and End of Life dates. 

---

**Questions? Help needed? Feature Requests?** 

- Join our [Open Server](https://open.rocket.chat/) in the #support channel and feel free to raise a question
- Join our [Community Forum](https://forums.rocket.chat/) and search/create a post there

PR #24824, opened 2022-03-14
[FIX] Opening Multiple VRDialog With Different Timer

nishant23122000:
### This fix contains several changes in UX

1> When use stops the video record, then the ok button will enable otherwise it will be disabled. this will now work when the user toggles between pause and start multiple times. (In the old case, it was not working).

2> There is no need to close VRecDialog when it's already open and when the user clicks on the video message. otherwise, there might be chances that multiple timers run parallelly if the user clicks the timer. if the user wants to close the dialog box, then he can simply close it by clicking the close button. so that the old video state will clear.

https://user-images.githubusercontent.com/53515714/158149062-b2e6ae30-8ea6-4ff9-b80d-5431f595184d.mp4

## Issue(s)
#24816

## Steps to test or reproduce
#24816


dougfabris:

Closing this PR because it's stale now and it was already fixed here: #31830 27850

---

PR #24828, opened 2022-03-14
[IMPROVE] Trigger roomLeft integration when user removed from room

equal-matt:

When a user is removed from a room, the `roomLeft` event is triggered.

## Further comments

This seems like the simplest way of achieving parity with [what happens in the hubot-rocketchat lib](https://github.com/RocketChat/hubot-rocketchat/blob/c45faf61a47caba84ed1afb669313ff86150caa7/index.js#L110). This integration is currently not fit for purpose (for us) because we would miss too many instances of users leaving without this event firing on removals by admins/owners/etc.


equal-matt:
@pierre-lehnen-rc - sorry to tag you directly but there doesn't seem to be a good way to get a maintainer's attention noted anywhere and I see you've approved another PR here. Could I get a review for this? We'd really like to move to webhooks over a bot in every room and this is currently blocking us.

PR #24835, opened 2022-03-15
[NEW] Upgrade Tab

gabriellsh:

image


client/views/admin/upgrade/UpgradePage/UpgradePage.tsx
```diff
@@ -0,0 +1,98 @@
...
+
+	try {
+		parsedMessage = JSON.parse(e.data);
+	} catch (error) {
+		return;
+	}
+
+	if (messageIsNavigation(parsedMessage)) {
+		return parsedMessage.goTo;
+	}
```
ggazzo, line 48:
```suggestion
	try {
		const parsedMessage = JSON.parse(e.data);
		if (messageIsNavigation(parsedMessage)) {
			return parsedMessage.goTo;
		}
	} catch (error) {
		return;
	}
```

client/views/admin/sidebar/AdminSidebarPages.tsx
```diff
@@ -1,19 +1,36 @@
...
 import Sidebar from '../../../components/Sidebar';
+import { useEndpoint } from '../../../contexts/ServerContext';
 import { itemsSubscription } from '../sidebarItems';
+import UpgradeTab from './UpgradeTab';
 
 type AdminSidebarPagesProps = {
 	currentPath: string;
 };
 
+const useUpgradeTabParams = (): [data: UpgradeTabVariants | false, trialEndDate: string | undefined] => {
```
ggazzo, line 16: why not use the react-query result instead?

client/views/admin/upgrade/UpgradePage/UpgradePage.tsx
dougfabris: if we use this approach the catch block will be unnecessary and we're not going to be able to validate the `parsedMessage` 😞

client/views/hooks/useUpgradeTabParams.ts
```diff
@@ -0,0 +1,16 @@
+import { useCallback } from 'react';
+import { useQuery } from 'react-query';
+
+import { UpgradeTabVariants } from '../../../lib/getUpgradeTabType';
+import { useEndpoint } from '../../contexts/ServerContext';
+
+export const useUpgradeTabParams = (): [data: UpgradeTabVariants | false, trialEndDate: string | undefined, isLoading: boolean] => {
```
tassoevan, line 7: Why not return `UseQueryResult<{ tabType: UpgradeTabVariants; trialEndDate: string }, Error>` and let the caller use the query as it wants?

gabriellsh: Fixed :)

client/views/admin/upgrade/UpgradePage/UpgradePage.tsx
```diff
@@ -0,0 +1,107 @@
...
+import PageHeader from '../../../../components/Page/PageHeader';
+import { useLayout } from '../../../../contexts/LayoutContext';
+import { useRouteParameter, useQueryStringParameter } from '../../../../contexts/RouterContext';
+import { useAbsoluteUrl } from '../../../../contexts/ServerContext';
+import { useLanguage } from '../../../../contexts/TranslationContext';
+import UpgradePageError from '../UpgradePageError';
+
+const iframeStyle = { width: '100%', height: '100%' };
+
+const urlMap = {
```
tassoevan, line 15:
```suggestion
const urlMap: Record<UpgradeTabVariants, string> = {
```

client/views/admin/upgrade/UpgradePage/UpgradePage.tsx
```diff
@@ -0,0 +1,107 @@
...
+	if (date) {
+		urlParams.set('date', date);
+	}
+
+	return `${urlMap[type]}?${urlParams.toString()}`;
+};
+
+type NavigationMessage = { goTo: string };
+
+const messageIsNavigation = (message: unknown): message is NavigationMessage => {
```
tassoevan, line 34: Wording here is a little weird...

```suggestion
const isNavigationMessage = (message: unknown): message is NavigationMessage => {
```

client/views/admin/upgrade/UpgradePage/UpgradePage.tsx
```diff
@@ -0,0 +1,107 @@
...
+
+	const getAbsoluteUrl = useAbsoluteUrl();
+
+	const { isMobile } = useLayout();
+
+	const ref = useRef<HTMLIFrameElement>(null);
+	const hasConnection = navigator.onLine;
+
+	useEffect(() => {
+		const navigate = (e: MessageEvent<string>): void => {
```
tassoevan, line 72:
```suggestion
		const handleNavigationMessage = (e: MessageEvent<string>): void => {
```

lib/getUpgradeTabType.ts
```diff
@@ -0,0 +1,38 @@
...
+	hasGoldLicense: boolean;
+};
+
+export const getUpgradeTabType = ({
+	registered,
+	hasValidLicense,
+	isTrial,
+	hasGoldLicense,
+	hadExpiredTrials,
+}: UpgradeTabConditions): UpgradeTabVariants | false => {
```
tassoevan, line 17: Why `false` instead of `undefined`?

tassoevan (changes requested):
Navigation seems broken. At first, I was directed to `/admin/info`. Then got stuck at `/admin/upgrade/goFullyFeatured` while going back on history...

client/views/hooks/useUpgradeTabParams.ts
```diff
@@ -0,0 +1,25 @@
...
+import { useEndpoint } from '../../contexts/ServerContext';
+
+const placeholderData = {
+	tabType: false as false,
+	trialEndDate: undefined,
+};
+
+export const useUpgradeTabParams = (): UseQueryResult<
+	{ tabType: UpgradeTabVariants | false; trialEndDate: string | undefined },
+	unknown
```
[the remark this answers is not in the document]
tassoevan: It's because `result` is not typed, therefore the generic parameters of the `useQuery()` call end up assigning `unknown` for `TError`. Typing like `const result: UseQueryResult<{...}, Error> = useQuery(...);` or returning it directly (without assigning to a variable) is enough.

app/api/server/v1/cloud.ts
```diff
@@ -90,3 +94,41 @@ API.v1.addRoute(
...
+
+			const upgradeTabType = getUpgradeTabType({
+				registered: workspaceRegistered,
+				hasValidLicense: licenses.length > 0,
+				hadExpiredTrials,
+				isTrial,
+				hasGoldLicense,
+			});
+
+			return API.v1.success({ tabType: upgradeTabType, trialEndDate });
```
pierre-lehnen-rc, line 131: Move this logic to the client side using `/api/v1/licenses.get`;
You can create a new endpoint to get the `retrieveRegistrationStatus` if necessary;
That way we don't have to mix CE and EE code.

tassoevan (approved):

tassoevan (approved):

PR #24837, opened 2022-03-15
[IMPROVE] Standarize queue behavior for managers and agents when subscribing

KevLehman:

### DIFF
--- a/app/livechat/server/api/lib/inquiries.js
+++ b/app/livechat/server/api/lib/inquiries.js
@@ -43,7 +43,7 @@ export async function findInquiries({ userId, department: filterDepartment, stat
 				$and: [{ defaultAgent: { $exists: true } }, { 'defaultAgent.agentId': userId }],
 			},
 			{ ...(department && { department }) },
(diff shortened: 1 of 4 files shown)


app/livechat/server/api/lib/inquiries.js
```diff
@@ -48,6 +43,8 @@ export async function findInquiries({ userId, department: filterDepartment, stat
 				$and: [{ defaultAgent: { $exists: true } }, { 'defaultAgent.agentId': userId }],
 			},
 			{ ...(department && { department }) },
+			// Add _always_ the "public queue" to returned list of inquiries (task: 26uu9ep)
```
tiagoevanp, line 46: remove clickup task reference on comment

tiagoevanp (changes requested):

app/livechat/server/api/lib/inquiries.js
KevLehman: Fixed!

tiagoevanp (approved):

PR #24880, opened 2022-03-18
[NEW] Create marketplace app details page new header

rique223:

Implemented a new header for the marketplaces app details page. 
-Changed the size of the app name;
-Implemented the app description field on the header;
-Changed the "metadata" section of the header(The part with the version and author information) now it also shows the last time the app was updated;
-Created a chip that will show when an app is part of one or more bundles and inform which are the bundles;
-Implemented a tooltip for the bundle chips;
-Created a new button + data badge component to substitute the current App Status;
-Changed the title of the "purchase button". Now it shows different text based on the "purchase type" of the app;
-Created a new Pricing & Status display which shows the price when the app is not bought/installed and shows the app status(Enabled/Disabled) when it is bought/installed;

Demo gif:
new_app_details_header


client/views/admin/apps/AppDetailsHeader.tsx
```diff
@@ -0,0 +1,93 @@
...
+	return (
+		<Box display='flex' flexDirection='row' mbe='x20' w='full'>
+			<AppAvatar size='x124' mie='x20' iconFileContent={iconFileContent} iconFileData={iconFileData} />
+			<Box display='flex' flexDirection='column'>
+				<Box display='flex' flexDirection='row' alignItems='center' mbe='x8'>
+					<Box fontScale='h1' mie='x8'>
+						{name}
+					</Box>
+					{Boolean(bundledIn.length) &&
+						bundledIn.map((bundle) => (
```
ujorgeleite, line 33: I suggest breaking the content of this map into a new component to make easy the manutenabillity.

ujorgeleite (changes requested):

client/views/admin/apps/AppDetailsHeader.tsx
rique223: Solved

PR #24925, opened 2022-03-23
Chore: add some missing REST definitions

gerzonc:

On the [mobile client](https://github.com/RocketChat/Rocket.Chat.ReactNative), we made an effort to collect more `REST API` definitions that are missing on the server side during our migration to TypeScript. Since we're both migrating to TypeScript, we thought it would be a good idea to share those so you guys can benefit from our initiative.

## Further comments

If there is anything else needed (tests, comments, etc) please let me know 🙈

### DIFF
--- a/definition/IUser.ts
+++ b/definition/IUser.ts
@@ -100,6 +100,16 @@ export interface IRole {
 	_id: string;
 }
 
+export interface IGetRoomRoles {
+	_id: string;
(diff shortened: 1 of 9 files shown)


ankar84:
Great initiative, IMO, @gerzonc 
Thank you!

ggazzo (approved):

PR #24949, opened 2022-03-25
[FIX] Regression: Wrong type on a value and missing teams statistics re-added

cauefcr:
After seeing the logs for the statscollector this error presented itself.


app/statistics/server/lib/statistics.ts
```diff
@@ -157,11 +157,17 @@ export const statistics = {
 
 		statsPms.push(
 			LivechatBusinessHours.col.count().then((count) => {
+				let strat = settings.get('Livechat_enable_business_hours');
```
KevLehman, line 160: Should this stat be called "strategy"? For me,. this looks like just "enabled" and strategy would be "Livechat_business_hour_type" setting (Single/Multiple) (this is an ee setting, strategy is always "single" on community, so the EE stats should override this key)

cauefcr: changes made, now we collect it on the EE statistics.

PR #24950, opened 2022-03-25
[FIX] While removing an agent association, extension status remains same.

amolghode1981:
Clickup task : https://app.clickup.com/t/23zh1kx

This issue happens because of the async nature of the things.
When association is dropped, agent triggers unregister. But agent and admin are 2 different entities.
On admin, whenever the association is deleted, the reload() function gets called. This requeries the endpoints again.

While on agent, whenever it sees that the user does not have the extension anymore, it stops the SIP user agent.
Which calls unregister and then disconnects the web socket.

In all practical cases, agent and admin will be different entities. So as soon as admin disassociates the extension, it is impossible to get the correct state of
endpoint because the admin queries the endpoint list before the unregistration is completed by the agent.

Solution :
We need to implement a notification mechanism where such async race conditions are handled. For this particular case,
we are going to notify all the admin users whenever the device state changes. The device state change event is sent by asterisk to the connector
when device is unregistered. When server receives this event from asterisk, it notifies all the admin users.

The admin user which is in the process of disassociation in (client/views/admin/settings/groups/voip/RemoveAgentButton.tsx) waits for this event.
Once this event is received, it calls reload.

If the admin navigates away from the "Extensions" tab and comes back to it, the status will be restored.

Note :

Stale status condition will be true for the following scenario.

Admin is on extension tab. Agent is on another window. Agent registers/unregisters.
Admin will not get this status.

## Issue(s)
Clickup task : https://app.clickup.com/t/23zh1kx
## Steps to test or reproduce
1. Login as admin.
2. Associate agent with an extension.
3. From agent, "Enable Voice Operations" (Register to asterisk by clicking up the phone icon)
4. From admin, delete the association for the user.
5. Notice that the state of the deleted association. The extension is still shown as registered.
## Further comments
Stale status condition will be true for the following scenario.

Admin is on extension tab. Agent is on another window. Agent registers/unregisters.
Admin will not get this status.

### DIFF
--- a/client/views/admin/settings/groups/voip/RemoveAgentButton.tsx
+++ b/client/views/admin/settings/groups/voip/RemoveAgentButton.tsx
@@ -1,18 +1,21 @@
 import { Table, Icon, Button } from '@rocket.chat/fuselage';
 import { useMutableCallback } from '@rocket.chat/fuselage-hooks';
-import React, { FC } from 'react';
+import React, { FC, useEffect } from 'react';
 
 import GenericModal from '../../../../../components/GenericModal';
 import { useSetModal } from '../../../../../contexts/ModalContext';
-import { useEndpoint } from '../../../../../contexts/ServerContext';
+import { useEndpoint, useStream } from '../../../../../contexts/ServerContext';
 import { useToastMessageDispatch } from '../../../../../contexts/ToastMessagesContext';
 import { useTranslation } from '../../../../../contexts/TranslationContext';
+import { useUserId } from '../../../../../contexts/UserContext';
 
-const RemoveAgentButton: FC<{ username: string; reload: () => void }> = ({ username, reload }) => {
+const RemoveAgentButton: FC<{ username: string; reload: () => void; extension: string }> = ({ username, reload, extension }) => {
 	const removeAgent = useEndpoint('DELETE', 'omnichannel/agent/extension');
 	const setModal = useSetModal();
 	const dispatchToastMessage = useToastMessageDispatch();
 	const t = useTranslation();
+	const subscribeToNotifyUser = useStream('notify-user');
+	const userId = useUserId();
 
 	const handleRemoveClick = useMutableCallback(async () => {
 		try {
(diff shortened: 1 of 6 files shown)


server/modules/listeners/listeners.module.ts
```diff
@@ -290,40 +290,8 @@ export class ListenersModule {
 			notifications.notifyUserInThisInstance(userId, 'callabandoned', { queuename, queuedcallafterabandon });
 		});
 
-		service.onEvent('notify.desktop', (uid, notification): void => {
-			notifications.notifyUserInThisInstance(uid, 'notification', notification);
-		});
-
```
MartinSchoeler, line 296: Is there a reason for removing these listeners?

server/sdk/lib/Events.ts
```diff
@@ -1,28 +1,22 @@
-import { IUIKitInteraction } from '@rocket.chat/apps-engine/definition/uikit';
```
MartinSchoeler, line 1: Same here, it seems this file has unnecessary removals, maybe you were working on an outdated branch?

client/views/admin/settings/groups/voip/RemoveAgentButton.tsx
```diff
@@ -1,28 +1,42 @@
...
 	const handleRemoveClick = useMutableCallback(async () => {
 		try {
 			await removeAgent({ username });
 		} catch (error: any) {
 			dispatchToastMessage({ type: 'error', message: error });
 		}
-		reload();
 	});
 
+	useEffect(() => {
```
MartinSchoeler, line 28: Instead  of a useEffect here, maybe you can add this action to to the `handleRemoveClick` or create a mutableCallback

MartinSchoeler: Since the extension state change happens from an action done by the user, It doesn't make much sense to listen to changes and then to the action, is it better to act when the event happens

client/views/admin/settings/groups/voip/VoipExtensionsPage.tsx
```diff
@@ -86,7 +86,11 @@ const VoipExtensionsPage: FC = () => {
 						{queues?.length > 2 && `+${(queues.length - 2).toString()}`}
 					</Box>
 				</Table.Cell>
-				{username ? <RemoveAgentButton username={username} reload={reload} /> : <AssignAgentButton extension={extension} reload={reload} />}
+				{username ? (
+					<RemoveAgentButton username={username} reload={reload} extension={extension} />
```
MartinSchoeler, line 90: Since `handleExtensionStateChange` needs a bunch of data that exists on this file, I think it would be better to declare `handleExtensionStateChange` in this component and pass the function as props, instead of passing the username and requesting the user id inside `removeAgentButton`

MartinSchoeler (changes requested):

server/modules/listeners/listeners.module.ts
amolghode1981: It was overwritten by mistake. I have fixed the issue.

server/sdk/lib/Events.ts
amolghode1981: It was overwritten by mistake. I have fixed the issue in the latest commit.

PR #24955, opened 2022-03-26
[FIX] room message not load when is a new message

filipemarins:

When the room object is searched for the first time, it does not exist on the front object yet (subscription), adding a fallback search for room list will guarantee to search the room details.

before:
https://user-images.githubusercontent.com/9275105/160223241-d2319f3e-82c5-47d6-867f-695ab2361a17.mp4

after:
https://user-images.githubusercontent.com/9275105/160223244-84d0d2a1-3d95-464d-8b8a-e264b0d4d690.mp4

## Steps to test or reproduce

Create a direct message to a user that has never chatted before. It should not open the chat room details.

Screenshot:
BEGINING_DM_INCOMPLETE

The room detail only appears after page refresh:
update_after_refresh_page

 This is an intermittent issue.

### DIFF
--- a/app/ui-utils/client/lib/RoomManager.js
+++ b/app/ui-utils/client/lib/RoomManager.js
@@ -16,6 +16,7 @@ import { CachedCollectionManager } from '../../../ui-cached-collection';
 import { getConfig } from '../../../../client/lib/utils/getConfig';
 import { ROOM_DATA_STREAM } from '../../../utils/stream/constants';
 import { callWithErrorHandling } from '../../../../client/lib/utils/callWithErrorHandling';
+import { call } from '../../../../client/lib/utils/call';
 import { RoomManager as NewRoomManager } from '../../../../client/lib/RoomManager';
 import { roomCoordinator } from '../../../../client/lib/rooms/roomCoordinator';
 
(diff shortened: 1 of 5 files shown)


app/ui-utils/client/lib/RoomManager.js
```diff
@@ -85,15 +86,15 @@ export const RoomManager = new (function () {
...
+					Object.entries(openedRooms).forEach(async ([typeName, record]) => {
 						if (record.active !== true || record.ready === true) {
 							return;
 						}
 
 						const type = typeName.substr(0, 1);
 						const name = typeName.substr(1);
 
-						const room = roomCoordinator.getRoomDirectives(type)?.findRoom(name);
+						const room = roomCoordinator.getRoomDirectives(type)?.findRoom(name) || (await call('getRoomByTypeAndName', type, name));
```
pierre-lehnen-rc, line 97: Move this to the `findRoom` function on `client/lib/rooms/roomTypes/direct.ts`

pierre-lehnen-rc (changes requested):

pierre-lehnen-rc (approved):
We should still investigate what caused this in the first place.

PR #24973, opened 2022-03-28
[FIX] Description text error and icon missing Chatpal

dan0bar:
## Issue(s)

The description of some settings has been parsed wrong, hence the messages are now written in Markdown and not HTML. Also the icons by search results In Chatpal plug-in didn't appear. 

## Steps to test or reproduce

1. Create a channel and set up Chatpal search
2. Write some test messages
3. Try to search for them with the search plugin
4. Next to your username should appear your avatar


tkurz:
Hello Rocket.Chat Team!

This fix brings back the Chatpal Provider Plugin to full functionality.
@sampaiodiego as you did some reviews in the past maybe you can take a look. it is quite a small fix actually but important for Chatpal.

Thx and best regards
Thomas

sampaiodiego:
hey there.. I'm sorry taking long to respond.

about the code, I didn't found any anything actually settings a value to `avatar_random_${username}` .. am I missing something?

dan0bar:
@sampaiodiego Hey Diego!
What do you mean by setting a value to `avatar_random_${username}`? I didn't really get it...

sampaiodiego:
so, at [this line](https://github.com/RocketChat/Rocket.Chat/pull/24973/files#diff-042e544918978ddebbc777f4153057044344c269ff48e595c41ad1f692084c2eR16) you're reading a value from session like `Session.get('avatar_random_${username}')` .. then I tried to find in the code if we have something actually setting a value to that key.. I was expecting to find something like `Session.set('avatar_random_${username}', 'some value')` but I haven't found anything.

dan0bar:
@sampaiodiego You were right. I removed the key and cache lines and it still works fine. I guess I used this lines from the previous version of `getUserAvatarURl` function.

PR #24984, opened 2022-03-29
[FIX] Prevent sequential messages edited icon to hide on hover

dougfabris:
### before
Screen Shot 2022-03-29 at 13 35 56

### after
Screen Shot 2022-03-29 at 11 48 05

### DIFF
--- a/app/theme/client/imports/general/base_old.css
+++ b/app/theme/client/imports/general/base_old.css
@@ -2005,10 +2005,6 @@
 
 			width: 60px;
 		}
-
-		&:hover .edited {
-			display: none;
-		}
 	}
 
 	& .body {


ggazzo (approved):

ivans-netto:
> > **Current**: image **Expected**: image **Description**: The distance from message to icon should be 8px, the same distance we have from avatar and user name. Note the icon contains its own canvas so the 8px gap should be from the edge of the icon canvas to the edge of the text element.
> 
> are you sure this review is really necessary ? we are not working/changing anything related with design, and we are about to change everything as soon as we finish the rewrite (planned to the next release 4.7.0 or 5.0.0 or whatever )

IMO it should be aligned in deliverable, but up to @ggazzo and you @dougfabris

PR #25000, opened 2022-03-31
[NEW] Allow filtering by custom fields on omnichannel's current chats panel

murtaza98:

New UI demo which allows you to filter using custom fields values
https://www.loom.com/share/b198ba1f53b940fc87383a3439556da9

### DIFF
--- a/client/components/VerticalBar/VerticalBar.tsx
+++ b/client/components/VerticalBar/VerticalBar.tsx
@@ -1,9 +1,13 @@
 import { Box } from '@rocket.chat/fuselage';
-import React, { FC, ComponentProps, memo } from 'react';
+import React, { FC, ComponentProps, memo, CSSProperties } from 'react';
 
 import { useLayoutContextualBarPosition, useLayoutSizes } from '../../providers/LayoutProvider';
 
-const VerticalBar: FC<ComponentProps<typeof Box>> = ({ children, ...props }) => {
+const VerticalBar: FC<ComponentProps<typeof Box> & { overridePosition?: CSSProperties['position'] }> = ({
+	children,
+	overridePosition,
+	...props
+}) => {
 	const sizes = useLayoutSizes();
 	const position = useLayoutContextualBarPosition();
 	return (
@@ -18,7 +22,7 @@ const VerticalBar: FC<ComponentProps<typeof Box>> = ({ children, ...props }) => 
 			borderInlineStartColor='neutral-300'
 			borderInlineStartStyle='solid'
 			height='full'
-			position={position}
+			position={overridePosition || position}
 			zIndex={5}
 			insetInlineEnd={'none'}
 			insetBlockStart={'none'}
--- a/client/views/omnichannel/currentChats/FilterByText.tsx
+++ b/client/views/omnichannel/currentChats/FilterByText.tsx
@@ -18,10 +18,11 @@ import RemoveAllClosed from './RemoveAllClosed';
 
 type FilterByTextType = FC<{
 	setFilter: Dispatch<SetStateAction<any>>;
+	setCustomFieldsSearchBarOpen: Dispatch<SetStateAction<boolean>>;
 	reload?: () => void;
 }>;
 
-const FilterByText: FilterByTextType = ({ setFilter, reload, ...props }) => {
+const FilterByText: FilterByTextType = ({ setFilter, reload, setCustomFieldsSearchBarOpen, ...props }) => {
 	const setModal = useSetModal();
 	const dispatchToastMessage = useToastMessageDispatch();
 	const t = useTranslation();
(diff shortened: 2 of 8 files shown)


client/views/omnichannel/currentChats/FilterByText.tsx
```diff
@@ -74,11 +76,46 @@ const FilterByText: FilterByTextType = ({ setFilter, reload, ...props }) => {
...
 	const onSubmit = useMutableCallback((e) => e.preventDefault());
-	const reducer = function (acc: any, curr: string): any {
-		acc[curr] = '';
-		return acc;
+
+	const handleCustomFieldsFilterToggle = (): void => setCustomFieldsSearchBarOpen(true);
+
+	const { value: allCustomFieldsDefinition, phase: stateCustomFields } = useEndpointData('livechat/custom-fields');
+
+	const jsonConverterToValidFormat = (customFields: Omit<ILivechatCustomField, '_updatedAt'>[]): CustomFieldsData => {
```
KevLehman, line 84: If possible, can we move this outside of comp into its own function?

client/components/VerticalBar/VerticalBar.tsx
```diff
@@ -1,9 +1,13 @@
 import { Box } from '@rocket.chat/fuselage';
-import React, { FC, ComponentProps, memo } from 'react';
+import React, { FC, ComponentProps, memo, CSSProperties } from 'react';
 
 import { useLayoutContextualBarPosition, useLayoutSizes } from '../../providers/LayoutProvider';
 
-const VerticalBar: FC<ComponentProps<typeof Box>> = ({ children, ...props }) => {
+const VerticalBar: FC<ComponentProps<typeof Box> & { overridePosition?: CSSProperties['position'] }> = ({
```
MartinSchoeler, line 6: Why are these changes necessary? I would leave changing this generic component as a last resort. What exactly are you trying to fix here? Maybe we can find a better solution

MartinSchoeler (changes requested):

client/views/omnichannel/currentChats/FilterByText.tsx
```diff
@@ -74,11 +76,46 @@ const FilterByText: FilterByTextType = ({ setFilter, reload, ...props }) => {
...
+					}),
+				};
+			}
+		});
+		return jsonObj;
 	};
 
+	const jsonCustomField = useMemo(
+		() => (allCustomFieldsDefinition?.customFields ? jsonConverterToValidFormat(allCustomFieldsDefinition.customFields) : {}),
+		// eslint-disable-next-line react-hooks/exhaustive-deps
```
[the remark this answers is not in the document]
murtaza98: I was getting an eslint warning about adding `jsonConverterToValidFormat` to this dependency array, hence I'd added this before. However now I've removed it since you recommended moving that function outside of this component, so this won't be required

client/views/omnichannel/currentChats/FilterByText.tsx
murtaza98: Sure, its done, please check

client/views/omnichannel/currentChats/FilterByText.tsx
```diff
@@ -45,7 +47,8 @@ const FilterByText: FilterByTextType = ({ setFilter, reload, ...props }) => {
 	const [from, setFrom] = useLocalStorage('from', '');
 	const [to, setTo] = useLocalStorage('to', '');
 	const [tags, setTags] = useLocalStorage<never | { label: string; value: string }[]>('tags', []);
-	const [customFields, setCustomFields] = useLocalStorage<any[]>('tags', []);
+	const [customFields, setCustomFields] = useLocalStorage<{}>('customFields', {});
```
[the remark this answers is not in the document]
murtaza98: I'm not sure, I went with this approach since I wanted to make sure that the state is preserved when the customer moves away from this component (like if they open an ongoing chat), and then come back. If there is some other approach to handle this, do let me know

client/components/VerticalBar/VerticalBar.tsx
MartinSchoeler: Indeed, since the vertical bar is not made to fit inside a `filterByText` component. It is made to work on a page. The solution here is to move the component up the component tree and use it on the `CurrentChatPage.tsx`. You will probably have to move some code that handles the changes to the params up as well to prevent multiple sources changing the same state, leading to unaccounted side effects.

murtaza98: I'd actually done something similar in my previous commit [here](https://github.com/RocketChat/Rocket.Chat/pull/25000/commits/e35b6c4710119998347d0b20e4f77a30668c75df) but I was running into infinite re-render loop since every time I update the custom fields in the parent component (i.e. `CurrentChatsPage`), it was re-rendering the `GenericTable` component, that was internally re-rendering the `FilterByText`, that was finally triggering the re-rendering of the main component (i.e. `CurrentChatsRoute`) and this was leading to an infinite loop.

MartinSchoeler: or we also could separate the form from the table (since it is a huge one that we barely support) and let the params handle the table data.

KevLehman:
@murtaza98 , pls also fix the conflicts 👀

PR #25012, opened 2022-03-31
[FIX] Regression: re-adding team statistics and new omnichannel statistic

cauefcr:
Sorry data team, the missing stats are on me.

### DIFF
--- a/app/statistics/server/lib/statistics.ts
+++ b/app/statistics/server/lib/statistics.ts
@@ -449,39 +449,21 @@ export const statistics = {
 			}),
 		);
 
-		statsPms.push(
-			new Promise<void>((resolve) => {
-				statistics.totalOTRRooms = Rooms.find({ createdOTR: true }).count();
-				resolve();
-			}),
-		);
+		statistics.totalOTRRooms = Rooms.countOTRstarted();
+		statsPms.push(Analytics.resetSeatRequestCount());
 
-		statsPms.push(
-			new Promise<void>((resolve) => {
-				statsPms.push(Analytics.resetSeatRequestCount());
-
-				statistics.dashboardCount = settings.get('Engagement_Dashboard_Load_Count');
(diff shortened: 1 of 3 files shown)


KevLehman (approved):

app/statistics/server/lib/statistics.ts
```diff
@@ -449,15 +449,32 @@ export const statistics = {
 			}),
 		);
 
-		statsPms.push(Analytics.resetSeatRequestCount());
+		statsPms.push(
+			new Promise<void>((resolve) => {
+				statistics.totalOTRRooms = Rooms.find({ createdOTR: true }).count();
```
sampaiodiego, line 454: the field `createdOTR` is missing an index and this query should be inside the model.. the method can be named liked `Rooms.findByCreatedOTR()`

app/statistics/server/lib/statistics.ts
```diff
@@ -449,15 +449,32 @@ export const statistics = {
...
+				resolve();
+			}),
+		);
+
+		statsPms.push(
+			new Promise<void>((resolve) => {
```
sampaiodiego, line 460: there is no need for this promise.

ee/app/license/server/getStatistics.ts
```diff
@@ -62,6 +64,13 @@ export async function getStatistics(): Promise<ENTERPRISE_STATISTICS> {
 		}),
 	);
 
+	statsPms.push(
+		new Promise<void>((resolve) => {
+			statistics.businessHourType = settings.get('Livechat_business_hour_type');
```
sampaiodiego, line 69: this doesn't need to be in a promise.

sampaiodiego (changes requested):

KevLehman (changes requested):
Can you create an index on `message.t`? Looks like we don't have one and that may affect perf (since messages collection is the biggest one)

PR #25016, opened 2022-04-01
[BREAK] Deactivated team members are added to auto-join rooms

matheusbsilva137:

- Do not add deactivated users to auto-join rooms.

## Steps to test or reproduce

1. Create a new team and add a team member to it;
2. Go to **Administration > Users** and deactivate one of the team members;
3. Return to the team and add an auto-join channel/group. **_Expected behavior:_** the deactivated user should not be added to the room.

### DIFF
--- a/apps/meteor/app/models/server/raw/Users.js
+++ b/apps/meteor/app/models/server/raw/Users.js
@@ -147,6 +147,12 @@ export class UsersRaw extends BaseRaw {
 		return this.find(query, options);
 	}
 
+	findActive(query, options = {}) {
+		Object.assign(query, { active: true });
+
+		return this.find(query, options);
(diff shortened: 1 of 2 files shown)


server/services/team/service.ts
```diff
@@ -653,7 +653,7 @@ export class TeamService extends ServiceClassInternal implements ITeamService {
 			};
 		}
 
-		const users = await this.Users.find({ ...query }).toArray();
+		const users = await this.Users.findActive({ ...query }).toArray();
```
pierre-lehnen-rc, line 656: This change will also affect the 'teams.members' endpoint

pierre-lehnen-rc (approved):

PR #25048, opened 2022-04-04
Chore: Tests with Playwright (task: ROC-29, 10-user-preferences)

tmontini:
Migrate cypress tests to playwright
- include 10-user-preferences


tmontini:
Updated PR: https://github.com/RocketChat/Rocket.Chat/pull/25122

PR #25053, opened 2022-04-04
Remove deps after using them

sampaiodiego:

Some installed deps are not required to run rocket.chat, so we can remove them.

This reduces the decompressed image size in ~20%:

```
🕙 15:54:39 ➜ docker images
REPOSITORY                         TAG                     IMAGE ID       CREATED          SIZE
rc-alpine-new                      latest                  1e23ec3224ec   20 seconds ago   871MB
rc-alpine-old                      latest                  0967489d355c   3 minutes ago    1.1GB
```

### DIFF
--- a/.docker/Dockerfile.alpine
+++ b/.docker/Dockerfile.alpine
@@ -1,12 +1,13 @@
 FROM node:14.18.3-alpine3.15
 
-RUN apk add --no-cache python3 make g++ libc6-compat ttf-dejavu
+RUN apk add --no-cache ttf-dejavu
 
 ADD . /app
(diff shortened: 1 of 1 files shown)


debdutdeb (approved):
We should do the same with the other image.

sampaiodiego:
> We should do the same with the other image.

we do already.. it is not as clean as on Alpine though:

https://github.com/RocketChat/Rocket.Chat/blob/3bedb58e33f38509f87253e4ddad6aafd782f5ac/.docker/Dockerfile#L20-L29

PR #25054, opened 2022-04-04
Chore: Store login token on session collection

albuquerquefabio:


apps/meteor/server/hooks/sauMonitorHooks.ts
```diff
@@ -6,14 +6,31 @@ import { Meteor } from 'meteor/meteor';
...
+		type?: string;
+		allowed?: boolean;
+		methodName?: string;
+		methodArguments: any[];
+	}) => {
+		const {
+			methodArguments,
+			connection: { httpHeaders },
+		} = info;
+		const { resume } = methodArguments.reduce((a, b) => Object.assign(a, b), {});
```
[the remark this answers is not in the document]
albuquerquefabio: Done!

apps/meteor/server/hooks/sauMonitorHooks.ts
```diff
@@ -6,14 +6,31 @@ import { Meteor } from 'meteor/meteor';
...
+		let loginToken = '';
+		if (resume) {
+			loginToken = Accounts._hashLoginToken(resume);
+		}
```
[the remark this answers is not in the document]
albuquerquefabio: Done!

apps/meteor/app/statistics/server/lib/SAUMonitor.ts
```diff
@@ -192,11 +192,12 @@ export class SAUMonitorClass {
 		const ip = connection.clientAddress || connection.httpHeaders?.['x-real-ip'] || connection.httpHeaders?.['x-forwarded-for'];
 
 		const host = connection.httpHeaders?.host || '';
-
+		const { loginToken = '' } = connection;
```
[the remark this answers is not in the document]
debdutdeb: I was thinking more like if there is no token, there is no session, right? 🤔 so basically not setting it to default or making the schema uneven. Instead, adding this as  a condition on Session.createOrUpdate.

apps/meteor/server/lib/getClientAddress.ts
```diff
@@ -1,8 +1,12 @@
...
 		return '';
 	}
-	return connection.clientAddress || connection.httpHeaders['x-real-ip'];
-}
+
+	const { clientAddress, httpHeaders } = connection;
+	const xRealIp = Array.isArray(httpHeaders?.['x-real-ip']) ? httpHeaders['x-real-ip'][0] : httpHeaders['x-real-ip'];
+
+	return clientAddress || xRealIp || '';
+};
```
debdutdeb, line 12: Why this change? x-real-ip is never an array, did you encounter it being so?

packages/core-typings/src/IMethodThisType.ts
```diff
@@ -1,16 +1,18 @@
...
 export interface IMethodThisType {
-	isSimulation: boolean;
-	userId: string | null;
-	connection: IMethodConnection | null;
-	setUserId(userId: string): void;
-	unblock(): void;
-	twoFactorChecked: boolean | undefined;
+  isSimulation: boolean;
+  userId: string | null;
+  connection: ISocketConnection | null;
```
debdutdeb, line 14: Same here, why do we need to switch to ISocketConnection? I'm not seeing any changes to existing code that are using these types (as in IMethodThisType).

albuquerquefabio: Everything about socket events expects connection as ISocketConnection

[apps/meteor/server/services/sauMonitor/events.ts](https://github.com/RocketChat/Rocket.Chat/blob/chore/session-login-token/apps/meteor/server/services/sauMonitor/events.ts#L5)

apps/meteor/server/lib/getClientAddress.ts
albuquerquefabio: When we use **ISocketConnection** the object **httpHeaders** can be `string | string[] | undefined`

debdutdeb (approved):

PR #25075, opened 2022-04-05
[NEW] Mention History

pierre-lehnen-rc:


app/lib/server/lib/sendNotificationsOnMessage.js
```diff
@@ -77,6 +78,10 @@ export const sendNotification = async ({
 
 	const isHighlighted = messageContainsHighlight(message, subscription.userHighlights);
 
+	if (hasMentionToUser || isHighlighted || (!disableAllMessageNotifications && (hasMentionToAll || hasMentionToHere))) {
```
pierre-lehnen-rc, line 81: `hasMentionToUser` is also true for thread replies, which should not generate a mention record.

app/models/server/raw/Mentions.ts
```diff
@@ -0,0 +1,9 @@
+import { BaseRaw, IndexSpecification } from './BaseRaw';
+import { IMention } from '../../../../definition/IMention';
+
+export class MentionsRaw extends BaseRaw<IMention> {
+	protected indexes: IndexSpecification[] = [
+		{ key: { uid: 1 }, unique: true },
```
sampaiodiego, line 6: this index is not needed and will actually prevent to mentions to the same user to be inserted.

PR #25087, opened 2022-04-06
[NEW] Add expire index to integration history

geekgonecrazy:
Adds an index to expire after 30 days integration history items.  With out it if you have a particularly failure heavy integration you could end up with terabytes worth of failure logs.  

Lets save hard drives and clean up logs. :)

### DIFF
--- a/apps/meteor/app/models/server/raw/IntegrationHistory.ts
+++ b/apps/meteor/app/models/server/raw/IntegrationHistory.ts
@@ -4,7 +4,10 @@ import { BaseRaw, IndexSpecification } from './BaseRaw';
 
 export class IntegrationHistoryRaw extends BaseRaw<IIntegrationHistory> {
 	protected modelIndexes(): IndexSpecification[] {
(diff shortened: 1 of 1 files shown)


geekgonecrazy:
https://github.com/RocketChat/Rocket.Chat/blob/integrationHistoryExpire/app/models/server/raw/index.ts#L119-L120

This calls the constructor which looks like: https://github.com/RocketChat/Rocket.Chat/blob/integrationHistoryExpire/app/models/server/raw/BaseRaw.ts#L88-L89

should be creating the indexes?  

Either way.. this makes sure the index exists.  I'm at your mercy for corrections

geekgonecrazy:
Seems related to: https://github.com/RocketChat/Rocket.Chat/pull/25101

sampaiodiego:
> Seems related to: #25101

not totally related.. but after #25101 is merged we won't rely on `IndexSpecification` field to create indexes anymore.. so this will need to be rewritten using new API

sampaiodiego (approved):

PR #25092, opened 2022-04-07
[FIX] Message preview not available for queued chats

murtaza98:
## Issue(s)

Preview message was not shown for queued chats
image

## Further comments

The solution was to make sure the Livechat Inquiry corresponding to the Room have the `lastMessage` property set


KevLehman (approved):

KevLehman (approved):

apps/meteor/app/livechat/server/hooks/saveLastMessageToInquiry.ts
```diff
@@ -0,0 +1,35 @@
+import { isOmnichannelRoom } from '@rocket.chat/core-typings';
+
+import { callbacks } from '../../../../lib/callbacks';
+import { LivechatInquiry } from '../../../models/server/raw';
+import { settings } from '../../../settings/server';
+import { RoutingManager } from '../lib/RoutingManager';
+
+callbacks.add(
+	'afterSaveMessage',
+	async (message, room) => {
```
sampaiodiego, line 10:
```suggestion
	async (message: IMessage, room: IRoom) => {
```

apps/meteor/app/livechat/server/hooks/saveLastMessageToInquiry.ts
```diff
@@ -0,0 +1,35 @@
...
+		if (!isOmnichannelRoom(room)) {
+			return message;
+		}
+
+		// skip callback if message was edited
+		if ((message as any).editedAt) {
```
sampaiodiego, line 16: pls avoid `any` as must as possible 🙈 
```suggestion
		if (message.editedAt) {
```

sampaiodiego (dismissed):

apps/meteor/app/livechat/server/hooks/saveLastMessageToInquiry.ts
sampaiodiego: in this case you could also use `isEditedMessage` from `IMessage.ts`

apps/meteor/app/livechat/server/hooks/saveLastMessageToInquiry.ts
sampaiodiego: in this case you could also use `isEditedMessage` from `IMessage.ts`

KevLehman: Nice catch, let me change it 👀

apps/meteor/app/livechat/server/hooks/saveLastMessageToInquiry.ts
KevLehman: It's actually detecting the right typings from the "aftersavemessage" callback signature, so i think we can skip these 👀

KevLehman (approved):

apps/meteor/app/livechat/server/hooks/saveLastMessageToInquiry.ts
sampaiodiego: cool.. I thought it was needed to add them to avoid the `(message as any)` in the lines below

murtaza98:
Hey @KevLehman we had a minor issue with the isEditedMethod for which I pushed a new commit. Can you please review and approve it again? Thanks!!

packages/core-typings/src/IMessage/IMessage.ts
```diff
@@ -155,7 +155,7 @@ export interface IEditedMessage extends IMessage {
 }
 
 export const isEditedMessage = (message: IMessage): message is IEditedMessage =>
-  "editedAt" in message && "editedBy" in message;
+  "editedAt" in message && "editedBy" in message && !!(message as IEditedMessage).editedAt && !!(message as IEditedMessage).editedBy?._id;
```
KevLehman, line 158: 🤔 i can understand the check for editedBy.id, but why do we need the !!msg.editedAt?

murtaza98: I was running into a situation where the `editedAt` property was set on the message but the value was `undefined`. So that was resulting in `"editedAt" in message` to be true :( So to solve this I added this extra check

KevLehman: So what about `message?.editedAt && message?.editedBy?._id`? That way you check if keys are there and if they have values 👀

murtaza98: That will throw an error since the return of this function is expecting a return type of boolean however `message.editedAt` is of type Date

KevLehman: Oh, yeah, didn't mean to copy literally 🙈 , but to remove the "editedAt" in message && "editedBy" in message

ggazzo: > I was running into a situation where the editedAt property was set on the message but its value 
do you know how or why is this happening? maybe a migration and fixing the object before insert, if not we should consider the whole typing since `undefined` is allowed

ggazzo (dismissed):

packages/core-typings/src/IMessage/IMessage.ts
murtaza98: @ggazzo I found that the app bridges were the root cause of those undefined properties so I've modified them to prevent this.

apps/meteor/client/sidebar/RoomList/SideBarItemTemplateWithData.js
```diff
@@ -136,7 +136,11 @@ const propsAreEqual = (prevProps, nextProps) => {
 	if (prevProps.room._updatedAt?.toISOString() !== nextProps.room._updatedAt?.toISOString()) {
 		return false;
 	}
-	if (prevProps.room.lastMessage?._updatedAt?.toISOString() !== nextProps.room.lastMessage?._updatedAt?.toISOString()) {
+	if (
+		typeof prevProps.lastMessage?._updatedAt === 'object' &&
```
KevLehman, line 140: If _updatedAt happens to be null, this will fail 👀

apps/meteor/client/sidebar/RoomList/SideBarItemTemplateWithData.js
```diff
@@ -136,7 +136,11 @@ const propsAreEqual = (prevProps, nextProps) => {
...
+		typeof nextProps.lastMessage?._updatedAt === 'object' &&
+		prevProps.lastMessage?._updatedAt?.toISOString() !== nextProps.lastMessage?._updatedAt?.toISOString()
```
KevLehman, line 142: Since you're checking that lastMessage?._updated is of type object (so it exists), does typescript still requires you to write the ? op?

KevLehman (changes requested):

apps/meteor/client/sidebar/RoomList/SideBarItemTemplateWithData.js
murtaza98: Good catch!! Modified it to handle this case

apps/meteor/client/sidebar/RoomList/SideBarItemTemplateWithData.js
murtaza98: yeah, ideally it shouldn't but since I copied the earlier condition, so those checks still exist there. But now since I've removed this object check, this won't be required

KevLehman (approved):

PR #25101, opened 2022-04-07
[FIX] Database indexes not being created

sampaiodiego:

Indexes from raw models using the property `indexes` were not being created due to an issue on OOP works in JS.

### DIFF
--- a/app/models/server/raw/Analytics.ts
+++ b/app/models/server/raw/Analytics.ts
@@ -1,14 +1,16 @@
 import { Random } from 'meteor/random';
 import { AggregationCursor, Cursor, SortOptionObject, UpdateWriteOpResult } from 'mongodb';
 
-import { BaseRaw, IndexSpecification } from './BaseRaw';
+import { BaseRaw } from './BaseRaw';
 import { IAnalytic } from '../../../../definition/IAnalytic';
 import { IRoom } from '../../../../definition/IRoom';
 
 type T = IAnalytic;
 
 export class AnalyticsRaw extends BaseRaw<T> {
-	protected indexes: IndexSpecification[] = [{ key: { date: 1 } }, { key: { 'room._id': 1, 'date': 1 }, unique: true }];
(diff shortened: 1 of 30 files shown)


ggazzo (changes requested):
check the warnings `Missing return type on function`

sampaiodiego:
I saw the warnings.. TS can correctly infer the missing returning type based on the parent class, that's why I didn't added them..

also, if we care about that warning specifically we should turn them into errors don't you think?

ggazzo:
Yep , but the infer can lead you to add undesirable results, thats why we force you to explicit define the return
> TS can correctly infer the missing returning type based on the parent class, that's why I didn't added them..

sampaiodiego:
> Yep , but the infer can lead you to add undesirable results, 

I don't understand why it lead to undesirable results, can you please explain that?

> thats why we force you to explicit define the return

this is not accurate, there is nothing forcing anything, they're just warnings.. and this also relates to my previous question that you probably didn't see: 

> also, if we care about that warning specifically we should turn it into errors don't you think? I mean, if you have to check the linting logs to ask for changes if a warning show up, IMO our linting rules are bad and should be improved, so we can focus on what matters.

ggazzo:
> > Yep , but the infer can lead you to add undesirable results,
> 
> I don't understand why it lead to undesirable results, can you please explain that?

imagine that we have a colleague, who is working in the following code:..
```
const fn = () => { 

	if(a){
		return false
	}

	return true;
}
``

somehow he decides to do something like 


```
const fn = () => { 

	if(a){
		return false
	}

	if(b) {
		return 'yes';
	}

	return true;
}

The ts will infer the result  as `boolean | string` and everything will "look good", until in same place we realise having a string instead a boolean is a problem;

that why we suggest always starting by the result definition


I agreed it should be an error, if we deal with warnings the same way we handle erros

ggazzo (approved):
Warnings are a parallel discussion ;)

PR #25108, opened 2022-04-08
[IMPROVE] Unify voip streams into single stream

KevLehman:


packages/core-typings/src/voip/IVoipClientEvents.ts
```diff
@@ -0,0 +1,54 @@
+export type VoipPropagatedEvents =
+	| 'agentcalled'
```
sampaiodiego, line 2: is this a good moment to rename these events to make them easier to read? 😬 
like `agent-called` or `queue-member-added`

KevLehman: 😬 well... it won't affect anyone to change it lol

KevLehman:
@d-gubert @amolghode1981 this would be a good addition to the release and forgot to put on milestone 😬 pls don't kill me

apps/meteor/server/modules/listeners/listeners.module.ts
```diff
@@ -271,23 +271,13 @@ export class ListenersModule {
...
-			notifications.notifyUserInThisInstance(userId, 'queuememberremoved', { queuename, queuedcalls });
+
+		service.onEvent('voip.events', (userId, data): void => {
+			notifications.notifyUserInThisInstance(userId, 'voip.events', data);
 		});
-		service.onEvent('queue.callabandoned', (userId, queuename: string, queuedcallafterabandon: string): void => {
-			notifications.notifyUserInThisInstance(userId, 'callabandoned', { queuename, queuedcallafterabandon });
+
+		service.onEvent('call.callerhangup', (userId, data): void => {
+			notifications.notifyUserInThisInstance(userId, 'call.hangup', data);
```
amolghode1981, line 280: Any specific reason for changing the event name from call.callerhangup to call.hangup? Everywhere else we are using the same name of the event to notify the user.

apps/meteor/client/providers/CallProvider/CallProvider.tsx
```diff
@@ -160,7 +132,7 @@ export const CallProvider: FC = ({ children }) => {
 			openWrapUpModal();
 		};
 
-		return subscribeToNotifyUser(`${user._id}/call.callerhangup`, handleCallHangup);
+		return subscribeToNotifyUser(`${user._id}/call.hangup`, handleCallHangup);
```
amolghode1981, line 135: Any specific reason for changing the callerhangup to call.hangup?

amolghode1981 (commented):
I have some confusion. Hence adding a generic comment. The title says that "Unify voip streams into single stream". When I saw the implementation of "notifyUserInThisInstance", it is using streamUser (https://github.com/RocketChat/Rocket.Chat/blob/e1588e086fe66e46562393df1a91a95fb82ae6ef/apps/meteor/server/modules/notifications/notifications.module.ts#L489)

Isnt it a single stream? Or Am I missing something here?

amolghode1981 (dismissed):

PR #25147, opened 2022-04-11
[IMPROVE] User actions in the members menu

dougfabris:
### before
image

### after
image


ankar84:
Nice that remove feature is located in that menu. Before that remove from room was only when you open user card in member list. That inconvenient and not consistent from UX aspect.

ankar84:
Yes, totally not clear that it's there!

PR #25153, opened 2022-04-11
[NEW] Description tab marketplace

rique223:

Implemented markdown on the description section of the app details page, now the description will show the detailedDescription.rendered (as rendered JSX) information in case it exists and show the description (a.k.a. short description) information in case it doesn't. Unfortunately, as of right now no app has a visual example of a markdown description and because of that, I will not be able to provide a demo image/gif for this PR.

### DIFF
--- a/apps/meteor/client/views/admin/apps/AppDetailsPage.tsx
+++ b/apps/meteor/client/views/admin/apps/AppDetailsPage.tsx
@@ -82,8 +82,8 @@ const AppDetailsPage: FC<{ id: string }> = function AppDetailsPage({ id }) {
 	const isLogsTabSelected = isSelectedLogs;
 	const isLogsTabEnabled = Boolean(installed);
 
-	const isSettingsTabSelected = Boolean(settings && Object.values(settings).length && isSelectedSettings);
-	const isSettingsTabEnabled = Boolean(settings && Object.values(settings).length && installed);
+	const isSettingsTabSelected = Boolean(isSelectedSettings && settings && Object.values(settings).length);
+	const isSettingsTabEnabled = Boolean(installed && settings && Object.values(settings).length);
 
 	return (
 		<Page flexDirection='column'>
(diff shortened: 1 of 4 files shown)


apps/meteor/client/views/admin/apps/AppDetailsPage.tsx
```diff
@@ -76,15 +76,15 @@ const AppDetailsPage: FC<{ id: string }> = function AppDetailsPage({ id }) {
...
-	const isSettingsTabSelected = Boolean(settings && Object.values(settings).length && isSelectedSettings);
-	const isSettingsTabEnabled = Boolean(settings && Object.values(settings).length && installed);
-
 	const isDetailsTabSelected = isSelectedDetails;
 	const areApisVisible = Boolean(isSelectedDetails && !!showApis);
 
 	const isLogsTabSelected = isSelectedLogs;
 	const isLogsTabEnabled = Boolean(installed);
 
+	const isSettingsTabSelected = Boolean(settings && Object.values(settings).length && isSelectedSettings);
```
ujorgeleite, line 85: I suggest putting isSelectedSettings in the front of the condition, because in this case if the settings aren't selected don't need to verify if the settings were filled.

apps/meteor/client/views/admin/apps/AppDetailsPage.tsx
```diff
@@ -76,15 +76,15 @@ const AppDetailsPage: FC<{ id: string }> = function AppDetailsPage({ id }) {
...
+	const isSettingsTabEnabled = Boolean(settings && Object.values(settings).length && installed);
```
ujorgeleite, line 86: The same for installed.

apps/meteor/client/views/admin/apps/AppDetailsPageContent.tsx
```diff
@@ -1,9 +1,8 @@
 /* eslint-disable @typescript-eslint/no-use-before-define */
```
ujorgeleite, line 1: This exception is necessary?

ujorgeleite (changes requested):

apps/meteor/client/views/admin/apps/AppDetailsPage.tsx
rique223: Solved

ujorgeleite (approved):

Issue #25155, opened 2022-04-12
[Bug] Admin UI web page crashed when deactivating user who is owner of at least one private group


ankar84:
### Description:

Admin UI web page crashed when deactivating user who is owner of at least one chat.
If you deactivations user that not an owner of any chat - it's OK.

### Steps to reproduce:

1. Create a new user, called UserA for example
2. Login as UserA and create private group chat
3. Login as Administrator in Admin UI - Users try to Deactivate UserA
4. App crashes

### Expected behavior:

Admin UI not crashed on deactivating any users

### Actual behavior:

![deactivate-owner](https://user-images.githubusercontent.com/4023037/162879926-c0c03257-05df-410c-8835-6dc85edb0c17.gif)
![image](https://user-images.githubusercontent.com/4023037/162880350-30f18b89-befb-499e-9998-96fdfaff67d2.png)

### Server Setup Information:

- Version of Rocket.Chat Server: 4.6.1
- Operating System: Centos7
- Deployment Method: docker
- Number of Running Instances: 12
- DB Replicaset Oplog: Enabled
- NodeJS Version: v14.18.3
- MongoDB Version: 4.4.8

### Client Setup Information

- Desktop App or Browser Version: Chrome 100
- Operating System: Windows 10

### Additional context

If you deactivations user that not an owner of any chat - it's OK.
So the problem is in changing owner in chat to another user

### Relevant logs:

```
POST https://rc.company.com/api/v1/users.setActiveStatus 400
send @ 1cd5dca1e2c9f2d70e8c63a6b36daf8ea031d466.js?meteor_js_resource=true:1053
ajax @ 1cd5dca1e2c9f2d70e8c63a6b36daf8ea031d466.js?meteor_js_resource=true:1053
d @ 1cd5dca1e2c9f2d70e8c63a6b36daf8ea031d466.js?meteor_js_resource=true:1493
_jqueryCall @ 1cd5dca1e2c9f2d70e8c63a6b36daf8ea031d466.js?meteor_js_resource=true:1493
a._jqueryCall @ 1cd5dca1e2c9f2d70e8c63a6b36daf8ea031d466.js?meteor_js_resource=true:1495
await in a._jqueryCall (async)
post @ 1cd5dca1e2c9f2d70e8c63a6b36daf8ea031d466.js?meteor_js_resource=true:1493
[...]

PR #24933, opened 2022-03-24
[FIX] Deactivating user breaks if user is the only room owner

sidmohanty11:
When we send details as `String(rooms)` it returns us a string =>  "[object Object]" so we cannot destructure out `shouldChangeOwner, shouldBeRemoved` which breaks the whole app.

## Before

https://user-images.githubusercontent.com/73601258/160000871-cfc2f2a5-2a59-4d27-8049-7754d003dd48.mp4

## After
https://user-images.githubusercontent.com/73601258/159998287-681ab475-ff33-4282-82ff-db751c59a392.mp4

## Issue(s)

Fixes #24934
Fixes #25155

PR #25156, opened 2022-04-12
[FIX] Client breaks when trying to deactivate a user who's the last owner of one or more room[s]

debdutdeb:
## Issue(s)

#25155


debdutdeb:
Ftr I don't know if it's deja vu or not, but i could swear I saw a similar patch a while ago. Or maybe it was in a different section altogether.

ankar84:
> Ftr I don't know if it's deja vu or not, but i could swear I saw a similar patch a while ago. Or maybe it was in a different section altogether.

Actually I have the same filling!

ankar84:
@debdutdeb what do you think that fix could be included in some patch version like 4.6.2?

sidmohanty11:
@debdutdeb I think you saw it here #24933 :)

debdutdeb:
Ahhhh yes! Thanks @sidmohanty11 ! I'm closing this one in favor of that.

PR #25160, opened 2022-04-12
Chore: Move voip's Wrap-up and On-hold functionality to EE (Backend)

murtaza98:


KevLehman:
What if, instead of using callbacks, we create an enterprise service that overrides the close function on this service? Like with `overrideClassOnLicense` func, and then on that function we do the things to store the close room message 👀 

That way we don't need to use callbacks inside services

apps/meteor/server/services/omnichannel-voip/service.ts
```diff
@@ -267,87 +268,75 @@ export class OmnichannelVoipService extends ServiceClassInternal implements IOmn
...
-	}
-
-	// Comment can be used to store wrapup call data
 	async closeRoom(
 		closerParam: ILivechatVisitor | ILivechatAgent,
 		room: IVoipRoom,
 		user: IUser,
-		comment?: string,
-		tags?: string[],
+		options: { comment?: string; tags?: string[] },
```
KevLehman, line 275: Should options be optional? 👀

apps/meteor/server/services/omnichannel-voip/service.ts
```diff
@@ -267,87 +268,75 @@ export class OmnichannelVoipService extends ServiceClassInternal implements IOmn
...
+		return true;
+	}
+
+	async getRoomClosingData(
+		closerParam: ILivechatVisitor | ILivechatAgent,
+		room: IVoipRoom,
+		options: { comment?: string; tags?: string[] },
+		sysMessageId: 'voip-call-wrapup' | 'voip-call-ended-unexpectedly',
+	): Promise<{ closeInfo: IRoomClosingInfo; closeSystemMsgData: any }> {
+		console.log('Calling community version');
```
murtaza98, line 306:
```suggestion
```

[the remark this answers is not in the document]
murtaza98: :see_no_evil:

apps/meteor/server/services/omnichannel-voip/service.ts
murtaza98: I see no issues with that, let me do that...

KevLehman:
CI is failing 😬

KevLehman (commented):
Currently on CE, it shows you the modal to close the conversation but doesn't store the data. Would be better to not to show the modal at all 👀 

The other thing: when looking at the calls panel, the call has both values as "Not available", would be good to not to show them if there's no valid license, maybe @MartinSchoeler can help you with how to achieve that 🤗

apps/meteor/app/api/server/v1/voip/rooms.ts
```diff
@@ -218,10 +218,12 @@ API.v1.addRoute(
 			check(this.bodyParams, {
 				rid: String,
 				token: String,
-				comment: Match.Maybe(String),
-				tags: Match.Maybe([String]),
+				options: Match.Maybe({
+					comment: Match.Maybe(String),
+					tags: Match.Maybe([String]),
+				}),
```
ggazzo, line 224: what is the point of pass an object option instead shallow props?

packages/core-typings/src/IMessage/IMessage.ts
```diff
@@ -232,3 +232,5 @@ export type IMessageInbox = IMessage & {
 
 export const isIMessageInbox = (message: IMessage): message is IMessageInbox => 'email' in message;
 export const isVoipMessage = (message: IMessage): message is IVoipMessage => 'voipData' in message;
+
+export type IOmniRoomClosingMessage = Pick<IMessage, 't' | 'groupable'> & Partial<Pick<IMessage, 'msg'>>;
```
ggazzo, line 236: looks wrong, this looks more like a `system message` than IOminichannelRoomClosingSomething
I think this is a internal type, not a core one

ggazzo (changes requested):
could you check my comment? thanks

murtaza98:
> Currently on CE, it shows you the modal to close the conversation but doesn't store the data. Would be better to not to show the modal at all eyes
> 
> The other thing: when looking at the calls panel, the call has both values as "Not available", would be good to not to show them if there's no valid license, maybe @MartinSchoeler can help you with how to achieve that hugs

Well, both these behaviours were actually intentional since I was planning to keep this PR isolated to backend only change. However, I don't mind adding the frontend changes too in here if the frontend guys are okay with it. @MartinSchoeler WDYT?

apps/meteor/app/api/server/v1/voip/rooms.ts
murtaza98: @ggazzo  I went with this after seeing a pattern across our code-base like [here](https://github.com/RocketChat/Rocket.Chat/blob/9578e15300465ff6f1074088b3b35366c567bb89/apps/meteor/app/livechat/server/methods/closeRoom.js#L8) where any params used by EE licence were bundled together. But let me know if you think we should still change this

packages/core-typings/src/IMessage/IMessage.ts
murtaza98: Changes done :heavy_check_mark:

apps/meteor/server/services/omnichannel-voip/helper.ts
```diff
@@ -0,0 +1,37 @@
+import { IVoipRoom } from '@rocket.chat/core-typings';
+
+import { PbxEvent } from '../../../app/models/server/raw';
+
+export const calculateOnHoldTimeForRoom = async (room: IVoipRoom, closedAt: Date): Promise<number> => {
+	if (!room || !room.callUniqueId) {
```
ggazzo, line 6: room is not optional there is no need to test if its empty or not

murtaza98:
> Currently on CE, it shows you the modal to close the conversation but doesn't store the data. Would be better to not to show the modal at all eyes
> 
> The other thing: when looking at the calls panel, the call has both values as "Not available", would be good to not to show them if there's no valid license, maybe @MartinSchoeler can help you with how to achieve that hugs

Hi @KevLehman The Frontend guys would like to keep this PR isolated from any frontend changes so it is easier to review. So what do you say about that? Should we merge this after this month's release?

murtaza98:
moving this to draft since some changes are also required for On-hold feature migration

murtaza98:
> moving this to draft since some changes are also required for On-hold feature migration

Done with this change. This PR is again ready for review :smile:

KevLehman:
Noticed you moved some files to `livechat-enterprise`, could you please create a `voip-enterprise` instead? 🙈 since the functionality you moved was not from livechat but from voip 👀

murtaza98:
> Noticed you moved some files to `livechat-enterprise`, could you please create a `voip-enterprise` instead? see_no_evil since the functionality you moved was not from livechat but from voip eyes

That's a good idea. I've moved them into their own folder now :+1: 

PS: Seems like we're facing the same build issue on this PR too :(

KevLehman (commented):
Apart from that comment, function-wise this seems to be working as expected 👀

KevLehman (approved):

KevLehman (dismissed):

ggazzo (approved):

KevLehman (approved):

PR #25165, opened 2022-04-12
[NEW] Add user events for apps

tapiarafael:
## Further comments
It depents on [Rocket.chat-Apps.engine PR](https://github.com/RocketChat/Rocket.Chat.Apps-engine/pull/499)


apps/meteor/app/authentication/server/startup/index.js
```diff
@@ -210,6 +212,18 @@ Accounts.onCreateUser(function (options, user = {}) {
...
 
 	callbacks.run('onCreateUser', options, user);
+
+	// App IPostUserCreated event hook
+	try {
+		Promise.await(Apps.triggerEvent(AppEvents.IPostUserCreated, { user, performedBy: Meteor.user() }));
+	} catch (error) {
+		if (error instanceof AppsEngineException) {
+			throw new Meteor.Error('error-app-prevented', error.message);
+		}
```
d-gubert, line 222: App events that happen after the fact cannot prevent it from happening, by definition - in this case, this validation can be removed to avoid confusion. Same applies to the other triggers

d-gubert (changes requested):

thassiov (dismissed):
LGTM
now we need to update the apps engine version here after merging the other pr to get rid of that syntax error

d-gubert (approved):

PR #25170, opened 2022-04-13
[FIX] Client disconnection on network loss

amolghode1981:

Agent gets disconnected (or Unregistered) from asterisk in multiple ways. The goal is that agent should remain online
unless agent explicitly logs off.
Agent can stop receiving calls in multiple ways due to network loss. Network loss can happen in following ways.
1. User tries to switch the network. User experiences a glitch of disconnectivity. This can be simulated by turning the network off
in the network tab of chrome's dev tool. This can disconnect the UA if the disconnection happens just before the registration refresh.
2. Second reason is when computer goes in sleep mode.
3. Third reason is that when asterisk is crashed/in maintenance mode/explicitly stopped.

Solution:
The idea is to detect the network disconnection and start the start the attempts to reconnect.
The detection of the disconnection does not happen in case#1. The SIPUA's UserAgent transport does not
call onDisconnected when network loss of such kind happens. To tackle this problem, window's online and offline event handlers are
used.

The number of retries is configurable but ideally it is to be kept at -1. Whenever disconnection happens, it should keep on trying to
reconnect with increasing backoff time. This behaviour is useful when the asterisk is stopped.

When the server is disconnected, it should be indicated on the phone button.

## Issue(s)
https://app.clickup.com/t/245c0d8

It is ambiguous on what should happen when the disconnection happens while the agent is on call.
## Steps to test or reproduce
There are atleast 4 known ways to reproduce the problem
1. Change the network. e.g switch from home broadband to tethered network.
2. Go to chrome's DevTools -> Network and change No Throttling to offline.
3. Put the computer to sleep.
4. Stop asterisk

## Further comments
Default expiration time of Asterisk is 600 seconds. So unregistration/re-registration will happen only after 10 minutes.

There are several test-cases. This feature should be tested in following scenarios.
Testcase#1
1. Turn off the local network. 
2. Make a call to the extension.
3. Verify that the network is actually disconnected by verifying that the UA does not receive the call.
4. Bring back the network.
5. Verify on asterisk that the userAgent is connected by running command pjsip show endpoint xxxx. In the result of this command verify that the extension status is "not in use" and NOT "unavailable"

Testcase#2.
1. Verify steps 2 onwards by  changing DevTools -> Network  No throttle to offline.
2. Verify step 2 onwards by switching the network.

Testcase#3
1. Put computer on sleep.
2. Make a call to the extension from another device.
3. Verify that the agent is actually disconnected by verifying that the UA does not receive the call.
4. Wake up the computer.
5. Verify on asterisk that the userAgent is connected by running command pjsip show endpoint xxxx. In the result of this command verify that the extension status is "not in use" and NOT "unavailable"

Tasecase#4.
1. Stop Asterisk.
2. Make a call to the extension from another device.
3. Verify that the asterisk  is actually stopped by verifying that the UA does not receive the call.
4. Restart asterisk.
5. Verify on asterisk that the userAgent is connected by running command pjsip show endpoint xxxx. In the result of this command verify that the extension status is "not in use" and NOT "unavailable"

Testcase#5
1. Goto Administrator -> Call Center -> Server Configuration.
2. Change retry count to 2.
3. Stop the asterisk.
4. Let the code retry for 2 times.
5. Make sure that after 2 retries, a proper error is shown as a tooltip of the phone button.
6. Make sure that UA does not receive any calls.


apps/meteor/client/lib/voip/VoIPUser.ts
```diff
@@ -54,22 +55,160 @@ export class VoIPUser extends Emitter<VoipEvents> implements OutgoingRequestDele
...
+		};
+
+		this.userAgent = new UserAgent(this.userAgentOptions);
+		this.userAgent.transport.isConnected();
+		this._opInProgress = Operation.OP_CONNECT;
+		try {
+			this.registerer = new Registerer(this.userAgent, { expires: 60 });
+			this.userAgent.transport.onConnect = this.onConnected.bind(this);
+			this.userAgent.transport.onDisconnect = this.onDisconnected.bind(this);
+			window.addEventListener('online', this.onlineNetworkHandler);
```
KevLehman, line 141: There's just one scenario that i think this won't work, and it's when asterisk ip is blocked by a firewall. In that case, the browser will be "online" but asterisk won't be reachable. Not sure how to overcome that tho

apps/meteor/packages/rocketchat-i18n/i18n/en.i18n.json
```diff
@@ -4002,6 +4002,7 @@
   "Sidebar": "Sidebar",
   "Sidebar_list_mode": "Sidebar Channel List Mode",
   "Sign_in_to_start_talking": "Sign in to start talking",
+  "Signalling_connection_disconnected": "Signalling connection disconnected",
```
KevLehman, line 4005:
```suggestion
  "Signaling_connection_disconnected": "Signaling connection disconnected",
```

KevLehman (changes requested):

KevLehman:
And there's some conflicts :(

apps/meteor/app/lib/server/startup/settings.ts
```diff
@@ -3209,6 +3209,14 @@ settingsRegistry.addGroup('Call_Center', function () {
 					value: true,
 				},
 			});
+			this.add('VoIP_Retry_Count', '-1', {
```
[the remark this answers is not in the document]
amolghode1981: Fixed.

apps/meteor/client/lib/voip/VoIPUser.ts
```diff
@@ -54,22 +55,160 @@ export class VoIPUser extends Emitter<VoipEvents> implements OutgoingRequestDele
...
 
+	private _connectionState: ConnectionState = 'INITIAL';
+
 	private _held = false;
 
 	private mode: WorkflowTypes;
 
 	private queueInfo: QueueAggregator;
 
+	// private networkMonitor: NetworkMonitor;
```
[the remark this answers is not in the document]
amolghode1981: Fixed.

apps/meteor/client/sidebar/sections/components/OmnichannelCallToggleReady.tsx
```diff
@@ -5,20 +5,25 @@ import React, { ReactElement, useEffect, useState } from 'react';
...
 	const voipClient = useCallClient();
-	const [onCall, setOnCall] = useState(false);
+	const [disableButtonClick, setDisableButtonClick] = useState(false);
+	const [networkStatus, setNetworkStatus] = useState<NetworkState>('online');
 	const callerInfo = useCallerInfo();
 	const callActions = useCallActions();
 
 	const getTooltip = (): string => {
+		if (networkStatus === 'offline') {
+			return t('Signalling_connection_disconnected');
```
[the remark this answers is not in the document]
amolghode1981: Fixed.

apps/meteor/client/lib/voip/VoIPUser.ts
amolghode1981: Connection is going to happen in this.userAgent.start() function. When the connection could not be established for any reason such as firewall blocking the address, ideally userAgent.start() should throw exception. We are catching that exception. I have tested this function when there is no network and it throws an exception. I will also test it for firewall scenario.

apps/meteor/packages/rocketchat-i18n/i18n/en.i18n.json
amolghode1981: Fixed.

KevLehman (approved):

apps/meteor/client/lib/voip/VoIPUser.ts
KevLehman: Yeah, after testing, I think that if a customer has asterisk IP blocked by firewall, then it's beyond our reach. Client will attempt reconnections to asterisk and will fail, but since that will be on customer's infrastructure, we don't have power to decide there.

@amolghode1981 , what do you think on sharing this as part of onboarding to customers? 👀

packages/core-typings/src/voip/SignallingSocketEvents.ts
```diff
@@ -0,0 +1,8 @@
+export type SignallinSocketEvents = {
```
sampaiodiego, line 1: small typo (missing `g`) .. also suggested to use single `l` per Kevin's suggestion on the translation.. would be good to have the file name changed as well.
```suggestion
export type SignalingSocketEvents = {
```

apps/meteor/definition/voip/SignallingSocketEvents.ts
```diff
@@ -0,0 +1,8 @@
+export type SignallinSocketEvents = {
+	connected: undefined;
+	disconnected: undefined;
+	connectionerror: unknown;
+	localnetworkonline: undefined;
+	localnetworkoffline: undefined;
+};
+export type SocketEventKeys = keyof SignallinSocketEvents;
```
sampaiodiego, line 8: this file seems duplicated from `packages/core-typings/src/voip/SignallingSocketEvents.ts`

sampaiodiego (changes requested):

packages/core-typings/src/voip/SignallingSocketEvents.ts
amolghode1981: Fixed. Renamed the file.

apps/meteor/packages/rocketchat-i18n/i18n/en.i18n.json
sampaiodiego: looks like you renamed just the key, but not the value 🙈

sampaiodiego (dismissed):

apps/meteor/packages/rocketchat-i18n/i18n/en.i18n.json
amolghode1981: Fixed. :P

apps/meteor/client/lib/voip/VoIPUser.ts
amolghode1981: Yeah we should notify the customer about this.

KevLehman (dismissed):

sampaiodiego (dismissed):

KevLehman (approved):

PR #25208, opened 2022-04-18
Regression: yarn dev triggers build dependencies

ggazzo:

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,7 +10,7 @@
 		"build:services": "turbo run build --filter=rocketchat-services...",
 		"build:ci": "turbo run build:ci",
 		"testunit": "turbo run testunit",
-		"dev": "turbo run dev --parallel --filter=@rocket.chat/meteor...",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -10,7 +10,7 @@
 		"build:services": "turbo run build --filter=rocketchat-services...",
 		"build:ci": "turbo run build:ci",
 		"testunit": "turbo run testunit",
-		"dev": "turbo run dev --parallel --filter=@rocket.chat/meteor",
+		"dev": "turbo run dev --parallel --filter=@rocket.chat/meteor...",
```
sampaiodiego, line 13: they can't be in parallel since they depend on each other.
```suggestion
		"dev": "turbo run dev --filter=@rocket.chat/meteor...",
```

sampaiodiego (changes requested):

sampaiodiego (approved):

PR #25217, opened 2022-04-18
[IMPROVE] Performance for some Omnichannel features

KevLehman:


apps/meteor/ee/app/livechat-enterprise/server/methods/getUnitsFromUserRoles.ts
```diff
@@ -1,15 +1,22 @@
 import { Meteor } from 'meteor/meteor';
+import mem from 'mem';
 
 import { hasAnyRoleAsync } from '../../../../../app/authorization/server/functions/hasRole';
 import LivechatUnit from '../../../models/server/models/LivechatUnit';
 
+export async function getUnitsFromUserRoles(user: Meteor.User | null): Promise<{ [k: string]: any }[] | undefined> {
```
sampaiodiego, line 7: can we change this function to receive only `userId`? it seems it uses only the `_id`..

the reason I'm asking is that since `Meteor.User` is a complex object, in order to memoize the result correctly, the input should be stringified first.

apps/meteor/server/startup/migrations/v261.ts
```diff
@@ -0,0 +1,25 @@
...
+import { ILivechatVisitor } from '../../../definition/ILivechatVisitor';
+import { Users } from '../../../app/models/server';
+
+// Convert all visitor emails to lowercase
+addMigration({
+	version: 261,
+	up() {
+		const updates: unknown[] = [];
+		Users.find({ 'visitorEmails.address': /[A-Z]/ })
+			.limit(6000)
```
sampaiodiego, line 11: missing a loop? 😬

apps/meteor/server/startup/migrations/index.ts
```diff
@@ -83,4 +83,6 @@ import './v256';
 import './v257';
 import './v258';
 import './v259';
+import './v260';
+import './v261';
```
sampaiodiego, line 87: can you please merge both migrations into a single one?

apps/meteor/app/models/server/models/LivechatVisitors.js
```diff
@@ -200,7 +202,7 @@ export class LivechatVisitors extends Base {
 
 	findOneGuestByEmailAddress(emailAddress) {
 		const query = {
-			'visitorEmails.address': new RegExp(`^${escapeRegExp(emailAddress)}$`, 'i'),
+			'visitorEmails.address': emailAddress,
```
sampaiodiego, line 205: I know I proposed these changes in the first place, but taking a wider look now, I wonder if we should do `String(emailAddress).toLowerCase()` here, instead of changing the case in every place using this function.. wdyt?

sampaiodiego (changes requested):

sampaiodiego (approved):

PR #25224, opened 2022-04-19
Chore: Add yarn plugin to check node and yarn version

ggazzo:
After some people struggling with node version....


sampaiodiego:
how it works? 🤔

ggazzo:
> how it works? 🤔

if you run `yarn` or any other task based on `yarn` the plugin will be triggered, and if the versions are invalid a message will be displayed and the process `exit(1)`

gabriellsh (approved):
Works nice!

KevLehman (approved):

PR #25226, opened 2022-04-19
[FIX] Add the requested setting for BioLength and NickName

### LINKED ISSUE #24968: Replacing magic numbers in the code by settings
There are a few "magic numbers" in the code, such as in https://github.com/RocketChat/Rocket.Chat/blob/a1c41a16c550968761dd0be84673318e7bc63b10/app/lib/server/functions/saveUser.js#L231

In this example, is there a reason why `260` has not been replaced by something like `settings.get('Accounts_MaxBioLength')`? (With the corresponding setting added to `app/lib/server/startup/settings.ts`).

aakash-gitdev:
Added the requested setting for BioLength and NickName also.

## Issue(s)
Closes #24968 

## Further comments
I want someone to kindly review this PR. Also should I add "i18nLabel" field in the add settings of 'NickName' and 'BioLength'? If yes then any kind of help in finding it's location would be great.

### DIFF
--- a/apps/meteor/app/lib/server/startup/settings.ts
+++ b/apps/meteor/app/lib/server/startup/settings.ts
@@ -134,6 +134,15 @@ settingsRegistry.addGroup('Accounts', function () {
 		this.add('Accounts_DefaultUsernamePrefixSuggestion', 'user', {
 			type: 'string',
 		});
+		this.add('Accounts_DefaultUsernamePrefixSuggestion', 'user', {
+			type: 'string',
+		});
+		this.add('Accounts_MaxBioLength', '260', {
+			type: 'int',
+		});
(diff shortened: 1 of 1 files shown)


gabriellsh (changes requested):
There is a couple of things I'd like to point out.

First, answering your question, yes you should add translations when you add a reference to it in a PR.

The other thing is, you added the settings, but you didn't use them. 😛 

So, that was mostly a review so far, but I'd like to comment that we shouldn't add setting for this kind of thing. IMO it adds little value and a lot of complexity for maintenance in the future. Instead of adding a setting, you could just add something like `const MAX_BIO_LENGTH = 260;`. This way, it's no longer a magic number and we don't need a whole setting for it.

Anyway, thanks for the help!

gabriellsh:
closed in favor of #25231

PR #25230, opened 2022-04-19
[FIX] VoIP disabled/enabled  sequence puts voip agent in error state

amolghode1981:

Initially it was thought that the issue occurs because of the race condition while changing the client settings vs those settings reflected on server side. So a natural solution to solve this is to wait for setting change event 'private-settings-changed'. Then if 'VoIP_Enabled' is updated and it is true, set voipEnabled  to true in useVoipClient.ts (on client side)

It was realised that the race does not happen because of the database or server noticing the changes late. But because of the time taken to establish the AMI connection with Asterisk.

Solution:

1. Change apps/meteor/app/voip/server/startup.ts. When VoIP_Enabled is changed, await for Voip.init() to complete and then broadcast connector.statuschanged  with changed value.
2. From apps/meteor/server/modules/listeners/listeners.module.ts use notifyLoggedInThisInstance to notify all logged in users on current instance.
3. in apps/meteor/client/providers/CallProvider/hooks/useVoipClient.ts add the event handler that receives this event. Change voipEnabled from constant to state. Change this state based on the 'value' that is received by the handler.

## Issue(s)
https://app.clickup.com/t/22rjza5

### DIFF
--- a/apps/meteor/app/voip/server/startup.ts
+++ b/apps/meteor/app/voip/server/startup.ts
@@ -2,14 +2,11 @@ import debounce from 'lodash.debounce';
 
 import { settings } from '../../settings/server';
 import { Voip } from '../../../server/sdk';
-import { api } from '../../../server/sdk/api';
 
 const debouncedRefresh = debounce(Voip.refresh, 1000);
 
-settings.watch('VoIP_Enabled', async (value: boolean) => {
-	const data = value ? await Voip.init() : Voip.stop();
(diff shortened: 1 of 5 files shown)


KevLehman:
Hey @amolghode1981 , can you try to replay this on top of #25108 ? That way, we'll use the same stream 👀

KevLehman:
Did a small update to the place where the event is being fired 👀 shouldn't affect anything

KevLehman (approved):

PR #25231, opened 2022-04-19
[IMPROVE] Added MaxNickNameLength and MaxBioLength constants

### LINKED ISSUE #24968: Replacing magic numbers in the code by settings

aakash-gitdev:
added above mentioned constants to remove magic numbers in a simple manner

## Issue(s)
closes #24968

### DIFF
--- a/apps/meteor/app/lib/server/functions/saveUser.js
+++ b/apps/meteor/app/lib/server/functions/saveUser.js
@@ -227,8 +227,9 @@ export function validateUserEditing(userId, userData) {
 }
 
 const handleBio = (updateUser, bio) => {
+	const MaxBioLength = 260;
 	if (bio && bio.trim()) {
-		if (typeof bio !== 'string' || bio.length > 260) {
+		if (typeof bio !== 'string' || bio.length > MaxBioLength) {
 			throw new Meteor.Error('error-invalid-field', 'bio', {
 				method: 'saveUserProfile',
 			});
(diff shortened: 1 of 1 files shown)


aakash-gitdev:
@gabriellsh can you please check that are these changes follwing with what you were suggesting?

gabriellsh:
I added a commit with a few code style fixes. It's the same thing but with expected formating.

Btw, I just wanted to point out that it's best to keep iterating over the same PR than opening new ones. You could have changed the last one you opened instead of making a new one for this. It's not a big problem, but it's very important in order for us to avoid duped/irrelevant PRs. It's also better because we can keep the conversation history.

Thanks for your comprehension :)

aakash-gitdev:
> I added a commit with a few code style fixes. It's the same thing but with expected formating.
> 
> Btw, I just wanted to point out that it's best to keep iterating over the same PR than opening new ones. You could have changed the last one you opened instead of making a new one for this. It's not a big problem, but it's very important in order for us to avoid duped/irrelevant PRs. It's also better because we can keep the conversation history.
> 
> Thanks for your comprehension :)

Thanks for your consideration. Being new to open source I am gradually learning these minute points and yeah from next time onwards I will keep in mind to keep iterating over same PR if the topic is same. Btw can you please guide me where should I learn the expected formatting styles so that in future I can make a fully acceptable pull request?

gabriellsh (approved):

PR #25232, opened 2022-04-19
[NEW] Session endpoints 

albuquerquefabio:


packages/core-typings/src/ISession.ts
```diff
@@ -50,3 +52,15 @@ export type UserSessionAggregation = Pick<ISession, '_id'> & {
...
 export type DeviceSessionAggregation = Pick<ISession, '_id'> & { type: string; name: string; version: string; count: number; time: number };
 export type OSSessionAggregation = Pick<ISession, '_id'> & { name: string; version: string; count: number; time: number };
+
+// Session for Multi Device Manager
+
+export interface IMDMSessionParams {
+	count?: number;
+	offset?: number;
+	search?: string | null;
+}
```
ggazzo, line 62: wrong place this is not a core typing ;)

ggazzo (changes requested):

apps/meteor/app/models/server/raw/Sessions.ts
```diff
@@ -737,12 +739,333 @@ export const aggregates = {
 };
 
 export class SessionsRaw extends BaseRaw<ISession> {
+	protected indexes: IndexSpecification[] = [
+		{ key: { instanceId: 1, sessionId: 1, year: 1, month: 1, day: 1 } },
+		{ key: { instanceId: 1, sessionId: 1, userId: 1 } },
```
KevLehman, line 744: If you change this idx to have `sessionId` as first, you won't need an individual idx on 746

KevLehman: Ref: https://www.mongodb.com/docs/manual/core/index-compound/#prefixes

apps/meteor/app/models/server/raw/Sessions.ts
```diff
@@ -737,12 +739,333 @@ export const aggregates = {
...
+		},
+	];
+
 	private secondaryCollection: Collection<ISession>;
 
 	constructor(public readonly col: Collection<ISession>, public readonly colSecondary: Collection<ISession>, trash?: Collection<ISession>) {
 		super(col, trash);
 
 		this.secondaryCollection = colSecondary;
+		this.col.createIndexes(this.indexes);
```
KevLehman, line 782: Isn't this done on `BaseRaw`? 👀

apps/meteor/app/models/server/raw/Sessions.ts
```diff
@@ -737,12 +739,333 @@ export const aggregates = {
...
+				createdAt: {
+					$first: '$createdAt',
+				},
+				loginAt: {
+					$first: '$loginAt',
+				},
+			},
+		};
+
+		const [docTotal] = await this.col
```
KevLehman, line 871: Instead of running the agg 2 times, you can take a look at the `$facet` stage 👀 
https://www.mongodb.com/docs/manual/reference/operator/aggregation/facet/

(If you search for $facet on the repo, you'll find 2 examples of this being used to paginate aggregations on a single call 👀 )

apps/meteor/app/models/server/raw/Sessions.ts
```diff
@@ -737,12 +739,333 @@ export const aggregates = {
...
+					$first: '$mostImportantRole',
+				},
+				roles: {
+					$first: '$roles',
+				},
+			},
+		};
+		const limitOperator = { $limit: count };
+
+		const [docTotal] = await this.col
```
KevLehman, line 996: Same here 😬

apps/meteor/app/models/server/raw/Users.js
```diff
@@ -174,6 +174,31 @@ export class UsersRaw extends BaseRaw {
...
+			.project({
+				_id: 1,
+				username: 1,
+				name: 1,
+				avatarOrigin: 1,
+				avatarETag: 1,
+			})
+			.skip(offset)
+			.limit(count)
+			.toArray();
```
KevLehman, line 199: as a note 👀 you can pass all of these calls to the 2nd param on `find`.

apps/meteor/ee/server/api/sessions.ts
```diff
@@ -0,0 +1,167 @@
...
+			check(
+				this.queryParams,
+				Match.ObjectIncluding({
+					offset: Match.Maybe(String),
+					count: Match.Maybe(String),
+					search: Match.Maybe(String),
+				}),
+			);
+
+			try {
```
KevLehman, line 30: no need for the try/catch if you're not doing something special with the error 👀

apps/meteor/ee/server/api/sessions.ts
```diff
@@ -0,0 +1,167 @@
...
+			}
+
+			if (!hasLicense('device-management')) {
+				return API.v1.unauthorized();
+			}
+			check(this.bodyParams, {
+				sessionId: String,
+			});
+
+			try {
```
KevLehman, line 59: Same here 😬

apps/meteor/ee/server/api/sessions.ts
```diff
@@ -0,0 +1,167 @@
...
+
+API.v1.addRoute(
+	'sessions/list.all',
+	{ authRequired: true, twoFactorRequired: true },
+	{
+		async get() {
+			if (!this.userId) {
+				API.v1.failure('error-invalid-user');
+			}
+			if (!hasRole(this.userId, 'admin')) {
```
KevLehman, line 92: Instead of checking for a role, can we check for a permission? 👀

KevLehman (changes requested):

apps/meteor/app/models/server/raw/Sessions.ts
albuquerquefabio: Just works for me when I add this line. But, I need to add this before this [patch release](#25101). 
I skip it, I'll remove it!

apps/meteor/app/models/server/raw/Sessions.ts
albuquerquefabio: So nice!
We have many things that follow this pattern above. Maybe updating this pattern is a good task that we can do in a near future. :smiley:

apps/meteor/ee/server/api/sessions.ts
albuquerquefabio: Yeah, we can do it. But in this case, we don't have this kind of permissions yet.

albuquerquefabio: https://github.com/RocketChat/Rocket.Chat/pull/25419

apps/meteor/app/models/server/raw/Sessions.ts
```diff
@@ -745,17 +747,313 @@ export class SessionsRaw extends BaseRaw<ISession> {
...
+		};
+
+		const queryArray = [matchOperator, sortOperator, groupOperator, projectOperator, facetOperator];
+
+		const [
+			{
+				docs: sessions,
+				count: [{ total } = { total: 0 }],
+			},
+		] = (await this.col.aggregate(queryArray).toArray()) as unknown as { docs: ISession[]; count: { total: number }[] }[];
```
KevLehman, line 878: you can pass a type param to the aggregate function iirc

apps/meteor/app/models/server/raw/Sessions.ts
```diff
@@ -745,17 +747,313 @@ export class SessionsRaw extends BaseRaw<ISession> {
...
+		};
+
+		const queryArray = [matchOperator, sortOperator, groupOperator, facetOperator];
+
+		const [
+			{
+				docs: sessions,
+				count: [{ total } = { total: 0 }],
+			},
+		] = (await this.col.aggregate(queryArray).toArray()) as unknown as {
```
KevLehman, line 1023: same here 👀

apps/meteor/client/views/room/MessageList/hooks/useMessages.ts
```diff
@@ -29,15 +31,15 @@ const removePossibleNullValues = ({
...
-	...(editedBy !== null && { editedBy }),
-	...(editedAt !== null && { editedAt }),
-	...(emoji !== null && { emoji }),
-	...(avatar !== null && { avatar }),
-	...(alias !== null && { alias }),
-	...(customFields !== null && { customFields }),
-	...(groupable !== null && { groupable }),
-	...(attachments !== null && { attachments }),
-	...(reactions !== null && { reactions }),
+	...(isNotNullOrUndefined(editedBy) && { editedBy }),
```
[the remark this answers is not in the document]
albuquerquefabio: I did not add it. Maybe some merge brings this piece of code.

apps/meteor/app/models/server/raw/Sessions.ts
albuquerquefabio: Done

apps/meteor/app/models/server/raw/Sessions.ts
albuquerquefabio: Done

packages/core-typings/src/ISession.ts
```diff
@@ -27,10 +27,14 @@ export interface ISession {
 	sessionId: string;
 	_updatedAt: Date;
 	createdAt: Date;
+	loginToken?: string;
```
debdutdeb, line 30:
```suggestion
```

apps/meteor/app/statistics/server/lib/SAUMonitor.ts
```diff
@@ -190,6 +190,7 @@ export class SAUMonitorClass {
 			return;
 		}
 
+
```
debdutdeb, line 193:
```suggestion
```

packages/core-typings/src/ISession.ts
albuquerquefabio: But we need this object

apps/meteor/app/models/server/raw/Sessions.ts
KevLehman: @sampaiodiego , were you aware of this? 👀 like, index creation is not working on baseraw

apps/meteor/app/models/server/raw/Sessions.ts
```diff
@@ -761,17 +763,315 @@ export class SessionsRaw extends BaseRaw<ISession> {
...
+
+		const [
+			{
+				docs: sessions,
+				count: [{ total } = { total: 0 }],
+			},
+		] = await this.col
+			.aggregate<{
+				docs: ISession & { _user: Pick<IUser, 'name' | 'username' | 'avatarETag' | 'avatarOrigin'> }[];
+				count: { total: number }[];
```
KevLehman, line 1042: This part could be its own type now, something like:
`WithItemCount<T> = T & { count: { total: number }[] }`

So that you can reuse it on other places

apps/meteor/app/models/server/raw/Sessions.ts
```diff
@@ -761,17 +763,315 @@ export class SessionsRaw extends BaseRaw<ISession> {
...
 	protected modelIndexes(): IndexSpecification[] {
 		return [
-			{ key: { instanceId: 1, sessionId: 1, year: 1, month: 1, day: 1 } },
-			{ key: { instanceId: 1, sessionId: 1, userId: 1 } },
-			{ key: { instanceId: 1, sessionId: 1 } },
-			{ key: { sessionId: 1 } },
-			{ key: { userId: 1 } },
-			{ key: { year: 1, month: 1, day: 1, type: 1 } },
-			{ key: { type: 1 } },
+			{ key: { createdAt: 1 } },
```
KevLehman, line 1051: A reminder that, if you removed an index that was already present on develop, pls create a migration removing the index from the db :)

apps/meteor/ee/server/api/sessions.ts
```diff
@@ -0,0 +1,162 @@
...
+			});
+
+			const { sessionId } = this.bodyParams;
+			const sessionObj = await Sessions.findOneBySessionIdAndUserID(sessionId, this.userId);
+
+			if (!sessionObj) {
+				return API.v1.notFound('Session not found');
+			}
+
+			Meteor.users.update(this.userId, {
```
KevLehman, line 62: Can the "normal" user model be used here? 👀

apps/meteor/ee/server/api/sessions.ts
```diff
@@ -0,0 +1,162 @@
...
+
+API.v1.addRoute(
+	'sessions/list.all',
+	{ authRequired: true, twoFactorRequired: true },
+	{
+		async get() {
+			if (!this.userId) {
+				API.v1.failure('error-invalid-user');
+			}
+			if (!hasPermission(this.userId, 'view-device-management')) {
```
KevLehman, line 84: You can move this check to the 2 param of `addRoute` as `permissionsRequired: ['view-device-management']`

apps/meteor/ee/server/api/sessions.ts
```diff
@@ -0,0 +1,162 @@
...
+			}
+			if (!hasLicense('device-management')) {
+				return API.v1.unauthorized();
+			}
+
+			check(this.bodyParams, {
+				sessionId: String,
+			});
+
+			try {
```
KevLehman, line 139: No need to use try/catch on handler if you're not doing something special with the error value 👀

KevLehman (changes requested):

apps/meteor/app/models/server/raw/Sessions.ts
pierre-lehnen-rc: It was fixed one or two releases ago.

apps/meteor/ee/server/api/sessions.ts
pierre-lehnen-rc: Yes, `Meteor.users` shouldn't be used.
And preferably the raw model should be used instead of the normal one.

apps/meteor/app/models/server/raw/Sessions.ts
```diff
@@ -1295,17 +1599,22 @@ export class SessionsRaw extends BaseRaw<ISession> {
...
+				logoutBy: userId,
 			},
 		};
 
-		return this.updateMany(query, update);
+		if (!session?.loginToken) {
+			return this.updateOne(query, update);
+		}
+
+		return this.updateMany({ userId, loginToken: session?.loginToken }, update);
```
pierre-lehnen-rc, line 1617: Why was this change necessary?

I'm not following the logic here.

apps/meteor/app/models/server/raw/Sessions.ts
pierre-lehnen-rc: I agree. And since not all session attributes are being projected, the resulting type is not a full `ISession`, in this case it's better to set the type picking only the attributes that are actually projected.

packages/core-typings/src/ISession.ts
```diff
@@ -30,7 +30,10 @@ export interface ISession {
 	host: string;
 	ip: string;
 	loginAt: Date;
+	logoutAt?: Date;
 	closedAt?: Date;
+	logoutBy?: string;
```
pierre-lehnen-rc, line 35:
```suggestion
	logoutBy?: IUser['_id'];
```
When the attribute name doesn't specify what kind of information is in the attribute, it's a good practice to link the type directly like this so we don't have to look up if this uses an id or an username or something else.

pierre-lehnen-rc (changes requested):

apps/meteor/app/models/server/raw/Sessions.ts
KevLehman: I quite remember that, but was not sure, thanks Pierre!

So, @albuquerquefabio , you can trust on BaseRaw to create indexes now 👀

apps/meteor/ee/server/api/sessions.ts
```diff
@@ -0,0 +1,162 @@
...
+
+				Meteor.users.update(sessionObj.userId, {
+					$pull: {
+						'services.resume.loginTokens': {
+							hashedToken: sessionObj.loginToken,
+						},
+					},
+				});
+
+				await Sessions.updateMany({ loginToken: sessionObj.loginToken }, { $set: { logoutAt: new Date(), logoutBy: this.userId } });
```
KevLehman, line 155: Pls move this update query to the model 👀

apps/meteor/ee/server/api/sessions.ts
```diff
@@ -0,0 +1,189 @@
...
+
+			const sortKeys = ['loginAt', 'device.name', 'device.os.name', '_user.username', '_user.name'];
+			if (!Object.keys(sort).filter((key) => sortKeys.includes(key)).length) {
+				return API.v1.failure('error-invalid-sort');
+			}
+
+			const searchUser = await Users.find<Pick<IUser, '_id'>>(
+				{ $text: { $search: search }, active: true },
+				{
+					projection: { _id: 1, score: { $meta: 'textScore' } },
```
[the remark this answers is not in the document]
albuquerquefabio: No, it is to sort by the best text score in this case. I need only the Ids.

apps/meteor/ee/server/api/sessions.ts
```diff
@@ -0,0 +1,189 @@
...
+import { IUser } from '@rocket.chat/core-typings';
+
+import { Users, Sessions } from '../../../app/models/server/raw/index';
+import { API } from '../../../app/api/server/api';
+import { hasLicense } from '../../app/license/server/license';
+
+API.v1.addRoute(
+	'sessions/list',
+	{ authRequired: true },
+	{
```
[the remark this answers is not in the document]
albuquerquefabio: Yes, but an EE key is necessary. This endpoint only shows information about the user who requested it.

apps/meteor/app/models/server/raw/Sessions.ts
```diff
@@ -30,6 +33,9 @@ type DestructuredDateWithType = {
 type DestructuredRange = { start: DestructuredDate; end: DestructuredDate };
 type DateRange = { start: Date; end: Date };
 
+type CustomSortOp = 'loginAt' | 'device.name' | 'device.os.name';
+type CustomSortOpAdmin = CustomSortOp | 'loginAt' | 'device.name' | 'device.os.name' | '_user.username' | '_user.name';
```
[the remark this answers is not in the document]
albuquerquefabio: Thank you!

apps/meteor/ee/server/api/sessions.ts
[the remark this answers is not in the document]
albuquerquefabio: I made some tests here and regarding the score on projection is optional. We can use the sort to organize the best content based on the search score.

apps/meteor/ee/definition/rest/v1/sessions/sessions.ts
```diff
@@ -0,0 +1,29 @@
+import type { ISession, DeviceManagementSession, DeviceManagementPopulatedSession } from '@rocket.chat/core-typings';
+
+import type { PaginatedRequest } from '../../../../../../../packages/rest-typings/src/helpers/PaginatedRequest';
```
KevLehman, line 3: Can't you import these form `@rocket-chat/rest-typings`? 👀

apps/meteor/ee/definition/rest/v1/sessions/sessions.ts
```diff
@@ -0,0 +1,29 @@
...
+import type { PaginatedResult } from '../../../../../../../packages/rest-typings/src/helpers/PaginatedResult';
+import { SessionsProps } from './SessionsProps';
```
KevLehman, line 5: Do `import type` work here? 👀

apps/meteor/ee/server/api/sessions.ts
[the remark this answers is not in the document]
KevLehman: Nice!

apps/meteor/ee/server/api/sessions.ts
```diff
@@ -0,0 +1,181 @@
...
+import { IUser } from '@rocket.chat/core-typings';
+
+import { isSessionsProps } from '../../definition/rest/v1/sessions/SessionsProps';
+import { Users, Sessions } from '../../../app/models/server/raw/index';
+import { API } from '../../../app/api/server/api';
+import { hasLicense } from '../../app/license/server/license';
+
+API.v1.addRoute(
+	'sessions/list',
+	{ authRequired: true },
```
KevLehman, line 11: This is the only handler here using Check instead of AJV 🙈

albuquerquefabio: No issues

KevLehman (approved):
All good for me :) Let's see other reviewer's comments

PR #25252, opened 2022-04-20
Chore: Tests with Playwright (task: ROC-25, 06-message)

weslley543:


apps/meteor/tests/e2e/utils/mocks/subscriptionMock.ts
```diff
@@ -0,0 +1,83 @@
+export const subscriptionMock = [
```
ggazzo, line 1: hey what do you think about using Types, like IUser and IRoom, just to prevent us to create some wrong mock?

weslley543: fix

ggazzo (approved):

PR #25265, opened 2022-04-22
Chore: Convert `UserStatusMenu` to TS

debdutdeb:


apps/meteor/client/components/UserStatusMenu.tsx
```diff
@@ -58,7 +78,7 @@ const UserStatusMenu = ({ onChange, optionWidth = undefined, initialStatus = 'of
 
 	return (
 		<>
-			<Button ref={ref} small square ghost onClick={onClick} onBlur={hide} onKeyUp={handleKeyUp} onKeyDown={handleKeyDown} {...props}>
+			<Button ref={ref} small square ghost onClick={onClick} onBlur={hide} onKeyUp={handleKeyUp} onKeyDown={handleKeyDown} margin={margin}>
```
KevLehman, line 81: 👀 what happened to the other props?

debdutdeb: No other props are passed except for margin. So didn't make sense using props. Just flattened it. If need to extend in the future then we can adjust.

apps/meteor/client/components/UserStatusMenu.tsx
```diff
@@ -39,9 +55,13 @@ const UserStatusMenu = ({ onChange, optionWidth = undefined, initialStatus = 'of
 		hide();
 	});
 
-	const ref = useRef();
+	const ref = useRef<HTMLElement>(null);
```
ggazzo, line 58: why not?
```suggestion
	const ref = useRef<HTMLButtonElement>(null);
```

ggazzo (dismissed):

apps/meteor/client/components/UserStatusMenu.tsx
debdutdeb: I don't exactly remember 🙈 let me revise this :p

apps/meteor/client/components/UserStatusMenu.tsx
```diff
@@ -1,17 +1,33 @@
...
+	onChange,
+	initialStatus = UserStatusType.OFFLINE,
+	optionWidth = undefined,
+	placement = 'bottom-end',
+}: UserStatusMenuProps): React.ReactElement => {
 	const t = useTranslation();
-	const [status, setStatus] = useState(initialStatus);
-	const allowInvisibleStatus = useSetting('Accounts_AllowInvisibleStatusOption');
+	const [status, setStatus] = useState<UserStatusType>(initialStatus);
+	const allowInvisibleStatus = useSetting('Accounts_AllowInvisibleStatusOption') as boolean;
```
gabriellsh, line 27:
```suggestion
	const allowInvisibleStatus = Boolean(useSetting('Accounts_AllowInvisibleStatusOption'));
```

apps/meteor/client/components/UserStatusMenu.tsx
```diff
@@ -1,17 +1,33 @@
...
+import type { Placements } from '@rocket.chat/fuselage-hooks';
 import React, { useRef, useCallback, useState, useMemo, useEffect } from 'react';
 
 import { useSetting } from '../contexts/SettingsContext';
 import { useTranslation } from '../contexts/TranslationContext';
 import { UserStatus } from './UserStatus';
 
-const UserStatusMenu = ({ onChange, optionWidth = undefined, initialStatus = 'offline', placement = 'bottom-end', ...props }) => {
+type UserStatusMenuProps = {
+	margin: string;
```
gabriellsh, line 11:
```suggestion
	margin: ComponentProps<typeof Box>['margin'];
```

apps/meteor/client/components/UserStatusMenu.tsx
```diff
@@ -1,17 +1,33 @@
...
+	onChange: (type: UserStatusType) => void;
+	initialStatus?: UserStatusType;
+	optionWidth?: any;
```
gabriellsh, line 14:
```suggestion
	optionWidth?: ComponentProps<typeof Option>['width'];
```

apps/meteor/client/components/UserStatusMenu.tsx
```diff
@@ -1,17 +1,33 @@
...
+	placement?: Placements;
+};
+
+const UserStatusMenu = ({
+	margin,
+	onChange,
+	initialStatus = UserStatusType.OFFLINE,
+	optionWidth = undefined,
+	placement = 'bottom-end',
+}: UserStatusMenuProps): React.ReactElement => {
```
gabriellsh, line 24:
```suggestion
}: UserStatusMenuProps): ReactElement => {
```

It's better to import `ReactElement` from react instead of using the namespace.

apps/meteor/client/components/UserStatusMenu.tsx
```diff
@@ -20,14 +36,14 @@ const UserStatusMenu = ({ onChange, optionWidth = undefined, initialStatus = 'of
 			</Box>
 		);
 
-		const statuses = [
-			['online', renderOption('online', t('Online'))],
-			['away', renderOption('away', t('Away'))],
-			['busy', renderOption('busy', t('Busy'))],
+		const statuses: Array<[value: UserStatusType, label: React.ReactElement]> = [
```
gabriellsh, line 39:
```suggestion
		const statuses: Array<[value: UserStatusType, label: ReactElement]> = [
```

apps/meteor/client/components/UserStatusMenu.tsx
```diff
@@ -1,17 +1,33 @@
...
+}: UserStatusMenuProps): React.ReactElement => {
 	const t = useTranslation();
-	const [status, setStatus] = useState(initialStatus);
-	const allowInvisibleStatus = useSetting('Accounts_AllowInvisibleStatusOption');
+	const [status, setStatus] = useState<UserStatusType>(initialStatus);
+	const allowInvisibleStatus = useSetting('Accounts_AllowInvisibleStatusOption') as boolean;
 
 	const options = useMemo(() => {
-		const renderOption = (status, label) => (
+		const renderOption = (status: UserStatusType, label: string): React.ReactElement => (
```
gabriellsh, line 30:
```suggestion
		const renderOption = (status: UserStatusType, label: string): ReactElement => (
```

gabriellsh (changes requested):

apps/meteor/client/components/UserStatusMenu.tsx
debdutdeb: Any reason for this? I thought we had value types for each setting? in that sense if this setting's value type is bool should we not get bool? in which case running through the constructor feels unnecessary

tassoevan: I agree with @debdutdeb, we already trust the returned value of `useSetting()` in many situations.

tassoevan (approved):

PR #25271, opened 2022-04-22
[IMPROVE] Fail fast on callbacks

KevLehman:


apps/meteor/package.json
```diff
@@ -13,7 +13,7 @@
 		"chat"
 	],
 	"scripts": {
-		"start": "meteor",
+		"start": "SERVER_NODE_OPTIONS=--max-semi-space-size=128 meteor",
```
sampaiodiego, line 16: do we want this? 👀

apps/meteor/app/models/server/raw/Subscriptions.ts
```diff
@@ -198,4 +207,62 @@ export class SubscriptionsRaw extends BaseRaw<T> {
...
+			},
+			'open': { $ne: true },
+		};
+
+		const update = {
+			$set: {
+				open: true,
+			},
+		};
+		return this.update(query, update, { multi: true });
```
sampaiodiego, line 266:
```suggestion
		return this.updateMany(query, update);
```

apps/meteor/app/models/server/raw/Subscriptions.ts
```diff
@@ -198,4 +207,62 @@ export class SubscriptionsRaw extends BaseRaw<T> {
...
+			$set: {
+				alert: true,
+				open: true,
+			},
+			$inc: {
+				unread: inc,
+			},
+		};
+
+		return this.update(query, update, { multi: true });
```
sampaiodiego, line 232:
```suggestion
		return this.updateMany(query, update);
```

apps/meteor/app/models/server/models/Subscriptions.js
```diff
@@ -12,8 +12,8 @@ export class Subscriptions extends Base {
 	constructor(...args) {
 		super(...args);
 
-		this.tryEnsureIndex({ rid: 1 });
 		this.tryEnsureIndex({ rid: 1, ls: 1 });
+		this.tryEnsureIndex({ 'rid': 1, 'userHighlights.0': 1 });
```
sampaiodiego, line 16: I've found a query using this inside an `$or` at:

https://github.com/RocketChat/Rocket.Chat/blob/86136cea1fff6aa68ebb0c46208923bbac510856/apps/meteor/app/lib/server/lib/sendNotificationsOnMessage.js#L254

not sure if the new index will be used for that query as well 🤔

apps/meteor/app/models/server/models/Subscriptions.js
```diff
@@ -30,7 +30,6 @@ export class Subscriptions extends Base {
 		this.tryEnsureIndex({ emailNotifications: 1 }, { sparse: 1 });
 		this.tryEnsureIndex({ autoTranslate: 1 }, { sparse: 1 });
 		this.tryEnsureIndex({ autoTranslateLanguage: 1 }, { sparse: 1 });
-		this.tryEnsureIndex({ 'userHighlights.0': 1 }, { sparse: 1 });
```
sampaiodiego, line 33: if we decide to go ahead an remove this index and the one `{ rid: 1 }`, a migration removing them will be needed.

apps/meteor/app/models/server/models/LivechatRooms.js
```diff
@@ -9,6 +9,7 @@ export class LivechatRooms extends Base {
 	constructor(...args) {
 		super(...args);
 
+		this.tryEnsureIndex({ _id: 1, t: 1 });
```
sampaiodiego, line 12: if the query is already using `_id: 'something'` which is also unique, this might not be needed, right?

apps/meteor/app/lib/server/lib/notifyUsersOnMessage.js
```diff
@@ -85,7 +86,7 @@ const getUserIdsFromHighlights = (rid, message) => {
 		.map(({ u: { _id: uid } }) => uid);
 };
 
-export function updateUsersSubscriptions(message, room) {
+/* export function updateUsersSubscriptions(message, room) {
```
sampaiodiego, line 89: any reason to keep the old code?

apps/meteor/package.json
KevLehman: :oh-no:

apps/meteor/app/models/server/models/LivechatRooms.js
KevLehman: Ouch, this shouldn't be here, thanks!

PR #25274, opened 2022-04-22
Chore: Convert customSounds folder to ts

juliajforesti:


apps/meteor/client/components/GenericModal.tsx
```diff
@@ -14,8 +14,8 @@ type GenericModalProps = RequiredModalProps & {
 	title?: string | ReactElement;
 	icon?: ComponentProps<typeof Icon>['name'] | ReactElement | null;
 	confirmDisabled?: boolean;
-	onCancel?: () => void;
-	onClose: () => void;
+	onCancel: () => void;
```
dougfabris, line 17: Are you sure you need these changes? It's going to impact a lot of components

apps/meteor/packages/rocketchat-i18n/i18n/en.i18n.json
```diff
@@ -1,5023 +1,5025 @@
 {
```
dougfabris, line 1: 👀 really?

dougfabris (changes requested):

apps/meteor/client/components/GenericModal.tsx
juliajforesti: just changed it back to optional...

ggazzo (approved):

PR #25276, opened 2022-04-22
Chore: Add typings for /v1/webdav.getMyAccounts

sampaiodiego:


packages/rest-typings/src/v1/webdav.ts
```diff
@@ -0,0 +1,9 @@
+import type { IWebdavAccount } from "@rocket.chat/core-typings";
+
+export type WebdavEndpoints = {
+	'webdav.getMyAccounts': {
+		GET: () => {
+			accounts: Pick<IWebdavAccount, '_id' | 'username' | 'serverURL' | 'name'>[];
```
ggazzo, line 6: what about create a type to hold this variation? does make any sense sharing this on other place or not?

sampaiodiego: idk actually, there is no other place using it.. 🤷‍♂️

KevLehman (dismissed):

PR #25285, opened 2022-04-22
Chore: Rewrite 2fa to typescript

KevLehman:
## Further comments

There's just one file i left as JS as the typings got really complicated and it was taking some liberties from JS that made the typing difficult


apps/meteor/app/2fa/server/code/TOTPCheck.ts
```diff
@@ -20,6 +20,10 @@ export class TOTPCheck implements ICodeCheck {
 			return false;
 		}
 
+		if (!user.services?.totp?.secret) {
```
ggazzo, line 23: please @RocketChat/backend not sure if returning is the best case, maybe throw an error?

sampaiodiego: returning `false` is fine.. it means it cannot not be verified by this class.

sampaiodiego: any other concern @ggazzo ?

PR #25292, opened 2022-04-24
[FIX] Reload Undefined While Creating Room

nishant23122000:
Getting Undefined error inside console when we create any new room.

### Before :

https://user-images.githubusercontent.com/53515714/164981162-23f7f407-031e-4caa-97f3-8723924215ba.mp4

## Further comments
Added condition before execute reload function.

### DIFF
--- a/apps/meteor/client/sidebar/header/CreateChannelWithData.js
+++ b/apps/meteor/client/sidebar/header/CreateChannelWithData.js
@@ -86,7 +86,7 @@ const CreateChannelWithData = ({ onClose, teamId = '', reload }) => {
 		}
 
 		onClose();
-		reload();
+		reload && reload();
 	}, [broadcast, createChannel, createPrivateChannel, description, encrypted, name, onClose, readOnly, teamId, type, users, reload]);
 
 	return (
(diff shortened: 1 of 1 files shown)


apps/meteor/client/sidebar/header/CreateChannelWithData.js
```diff
@@ -86,7 +86,7 @@ const CreateChannelWithData = ({ onClose, teamId = '', reload }) => {
 		}
 
 		onClose();
-		reload();
+		reload && reload();
```
debdutdeb, line 89:
```suggestion
		reload?.();
```

Let's use the chain syntax. 

Better to do either if (reload) reload(); OR, this ^ imo.

debdutdeb (changes requested):

apps/meteor/client/sidebar/header/CreateChannelWithData.js
nishant23122000: Right. Thank you

dougfabris:
Closing this one because it's already fixed here: #27469
Thanks for contributing @nishant23122000

PR #25298, opened 2022-04-25
Chore: Convert email inbox feature to TypeScript

ujorgeleite:
chore: add endpoint types for email-inboxes
chore: add migration for some e-mail-inboxes files

Wasn't migrated all file folders on this pr, but we can merge, and after we will migrate in the next engineering day.


apps/meteor/app/api/server/v1/email-inbox.ts
```diff
@@ -11,10 +11,10 @@ API.v1.addRoute(
 	'email-inbox.list',
 	{ authRequired: true },
 	{
-		get() {
+		async get() {
 			const { offset, count } = this.getPaginationItems();
-			const { sort, query } = this.parseJsonQuery();
-			const emailInboxes = Promise.await(findEmailInboxes({ userId: this.userId, query, pagination: { offset, count, sort } }));
+			const { sort, query }: { sort: {}; query: {} } = this.parseJsonQuery();
```
KevLehman, line 16: shouldn't this get the right types as Record (for both)

apps/meteor/app/api/server/v1/email-inbox.ts
```diff
@@ -25,40 +25,48 @@ API.v1.addRoute(
...
+				}),
+				imap: Match.ObjectIncluding({
 					server: String,
+					port: Number,
 					username: String,
+					password: String,
+					secure: Boolean,
 				}),
 			});
+			if (!this.bodyParams._id) {
```
KevLehman, line 55: This endpoint serves for both update/creation. This validation will break it for creation path.

KevLehman (changes requested):

apps/meteor/app/api/server/v1/email-inbox.ts
albuquerquefabio: Yes, this should. But, `sort` and `query` return as **any**

albuquerquefabio: Updated

tassoevan (approved):

PR #25314, opened 2022-04-26
Regression: Fix size of custom emoji and render emoji on thread message preview

filipemarins:
## Issue(s)

Fix: emoji name display `:undefined:` in message template
Fix: size of custom emoji when is `BigEmoji`
Fix: size of message reaction when is custom emoji
Fix: the thread preview was not parsing the message

## Further comments

The custom emoji was not rendering with the correct size. And also add the message render to thread preview message.
Fix the message reaction size of custom emoji.


apps/meteor/client/components/Message/MessageBodyRender/BigEmoji.tsx
```diff
@@ -1,18 +1,29 @@
...
-		))}
-	</>
-);
+const BigEmoji: FC<BigEmojiProps> = ({ value, disableBigEmoji }) => {
+	const bigEmojiClass = css`
+		> span {
+			width: ${disableBigEmoji ? '1em' : '44px'};
+			height: ${disableBigEmoji ? '1em' : '44px'};
+		}
+	`;
```
ggazzo, line 18: do not always use styling on application side, please adjust component on fuselage side

ggazzo (changes requested):

ggazzo (dismissed):
consider create a MessageBodyRender for Threads like ThreadBodyRender instead of polluting the components

apps/meteor/client/components/Message/MessageBodyRender/BigEmoji.tsx
filipemarins: This isn't a fuselage component, I am styling the component rendered by the emoji parser... I've fixed the stylesheet used by the parse, but @gabriellsh asked to not rely on external CSS and create the style on react component.

filipemarins:
> consider create a MessageBodyRender for Threads like ThreadBodyRender instead of polluting the components

I've created the `MessageRender` component, the problem with the `ThreadBody` name is that can cause some mistakes with thread message rendering, instead of thread preview.

Also, the `MessageRender` was reused on message content.

gabriellsh (approved):

PR #25323, opened 2022-04-27
Regression: Better MongoDB connection management for micro services

sampaiodiego:

#25301 added error check for MongoDB connections but only for `authorization` service, this one adds to every service using `getConnection` or `getCollection` .

### DIFF
--- a/apps/meteor/ee/server/services/authorization/service.ts
+++ b/apps/meteor/ee/server/services/authorization/service.ts
@@ -4,9 +4,6 @@ import { api } from '../../../../server/sdk/api';
 import { Authorization } from '../../../../server/services/authorization/service';
 import { getConnection } from '../mongo';
 
-getConnection()
-	.then((db) => api.registerService(new Authorization(db)))
-	.catch((error) => {
-		console.error(error);
-		process.exit(1);
-	});
+getConnection().then((db) => {
+	api.registerService(new Authorization(db));
+});
--- a/apps/meteor/ee/server/services/mongo.ts
+++ b/apps/meteor/ee/server/services/mongo.ts
@@ -29,8 +29,9 @@ function connectDb(poolSize: number): Promise<MongoClient> {
 	});
 }
 
+let db: Db;
+
 export const getConnection = ((): ((poolSize?: number) => Promise<Db>) => {
-	let db: Db;
 	let client: Promise<MongoClient>;
 
 	return async (poolSize = 5): Promise<Db> => {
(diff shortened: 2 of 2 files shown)


ggazzo (dismissed):

apps/meteor/ee/server/services/authorization/service.ts
```diff
@@ -4,9 +4,6 @@ import { api } from '../../../../server/sdk/api';
 import { Authorization } from '../../../../server/services/authorization/service';
 import { getConnection } from '../mongo';
 
-getConnection()
-	.then((db) => api.registerService(new Authorization(db)))
-	.catch((error) => {
-		console.error(error);
-		process.exit(1);
-	});
+getConnection().then((db) => {
```
KevLehman, line 7: no `catch` here? 👀

apps/meteor/ee/server/services/mongo.ts
```diff
@@ -14,23 +15,40 @@ export enum Collections {
...
+			client.then((c) => {
+				db = c.db(name);
+			});
+		}
+		// if getConnection was called multiple times before it was connected, wait for the connection
+		return (await client).db(name);
+	};
+})();
+
 export async function getCollection<T>(name: Collections): Promise<Collection<T>> {
```
KevLehman, line 51: i don't know how often we call `getCollection`, but if it's often, maybe we can cache `db` and call getConnection again only if db doesn't have a value instead of evry time we call getCollection 👀

apps/meteor/ee/server/services/authorization/service.ts
sampaiodiego: I've moved the `catch` to on `getConnection` 😉

apps/meteor/ee/server/services/mongo.ts
KevLehman: Yup, but it means you're spinning up a promise that won't be needed if db is already set 👀

apps/meteor/ee/server/services/authorization/service.ts
KevLehman: Oh nice, i'm not a fan of calling a process.exit on a promise chain 🙈 but i guess it won't affect anything.

apps/meteor/ee/server/services/mongo.ts
KevLehman: Something like this:
```
let db: Db;

async function getConnection() => client.connect().asfasdfdsaf
async function getCollection() => {
  if (!db)
      db = await getConnection();
   
   return db.collection('blablabla')
}
```

This way, the getConnection promise is only called (hopefully) once and getCollection manages the fetching of it, avoiding to call the extra promise

KevLehman (approved):

PR #25334, opened 2022-04-28
Chore: create api/server/v1 endpoint definitions

felipe-rod123:

Converted some files from JS to TS from the `apps/meteor/app/api/server/v1/` folder, and created their endpoint definitions on `packages/rest-typings/src/v1/`


packages/rest-typings/src/v1/statsEndpoints.ts
```diff
@@ -0,0 +1,28 @@
+import type { IUser } from '@rocket.chat/core-typings';
+import type { TelemetryEvents } from '../../../../apps/meteor/server/sdk/types/ITelemetryEvent';
```
ggazzo, line 2: move `TelemetryEvents` to 'core-typings'

PR #25342, opened 2022-04-29
Chore: Convert getStatistics

sampaiodiego:


ankar84:
Great PR, Diego!
What do you think if we can do all getStatistics stuff on secondary mongodb node?

ggazzo (approved):

sampaiodiego:
> Great PR, Diego! What do you think if we can do all getStatistics stuff on secondary mongodb node?

hey @ankar84 .. this is something we have been discussing and we've agreed on doing so.. I can't say when it is going to be done though.

PR #25356, opened 2022-05-01
Chore: [WIP] _BaseDb ts rewrite

debdutdeb:


debdutdeb:
PS: i'm not 100% sure about the oplog entry type. there's a little inconsistency where the idl definition != query data; also need to validate both object and object2 types. this also includes sharding fields which atm we don't care about s o maybe it'll be better to use ReplicaSetOplogEntry instead of OplogEntry - won't be touching this avenue for now though.

Issue #25355, opened 2022-05-01
[IMPROVE] Redirect To Current Chat Section


nishant23122000:
### Description:

It's a good user Experience if we redirect back to the current chat section when the manager manually closes or forward the chat.

### Steps to reproduce:

1. Go to Omnichannel
2. Go To Current Chats Section
3. Open any chat with open Status.
4.  Try to close it

### Expected behavior:

When the user(Manager) closes the chat or forward it we should redirect to **/omnichannel/current**.

### Actual behavior:

Currently It redirects to the **/home** when we close the chat or forward it.

When we close Chat : 

https://user-images.githubusercontent.com/53515714/166132971-b1cad137-459e-4f86-8c88-038f63541c78.mp4

When we forward chat :

https://user-images.githubusercontent.com/53515714/166132984-b4fbc7f4-c4cc-4de6-9e5c-cf1f547621c1.mp4

### Server Setup Information:

- Version of Rocket.Chat Server: 4.7.0
- Operating System: Windows 10
- Deployment Method: 
- Number of Running Instances:  
- DB Replicaset Oplog: 
- NodeJS Version:  14.18.3
- MongoDB Version: 4.4.4

### Client Setup Information

- Desktop App or Browser Version:
- Operating System:

florisvangeel:
Interesting feature request, 
what if there is no last chat?

KevLehman:
We deprecated this screen in recent versions. Thanks for reporting!

PR #25357, opened 2022-05-01
[IMPROVE] Redirect To Current Chats Section

nishant23122000:

https://user-images.githubusercontent.com/53515714/166134949-36986441-42d2-443b-ace2-ca64871def9f.mp4

## Issue(s)
#25355

## Steps to test or reproduce
#25355

## Further comments
Will work for forward chat also...

### DIFF
--- a/apps/meteor/app/livechat/client/views/app/livechatReadOnly.js
+++ b/apps/meteor/app/livechat/client/views/app/livechatReadOnly.js
@@ -83,7 +83,8 @@ Template.livechatReadOnly.onCreated(async function () {
 			// this will force to refresh the room
 			// since the client wont get notified of room changes when chats are on queue (no one assigned)
 			// a better approach should be performed when refactoring these templates to use react
-			FlowRouter.go('/omnichannel/current');
+			if (window.location.pathname.includes('current')) FlowRouter.go('/omnichannel/current');
+			else FlowRouter.go('/');
 			ChatRoom.remove(this.rid);
 			CachedChatRoom.save();
 		}
--- a/apps/meteor/app/livechat/client/views/app/livechatReadOnly.js
+++ b/apps/meteor/app/livechat/client/views/app/livechatReadOnly.js
@@ -83,9 +83,9 @@ Template.livechatReadOnly.onCreated(async function () {
 			// this will force to refresh the room
 			// since the client wont get notified of room changes when chats are on queue (no one assigned)
 			// a better approach should be performed when refactoring these templates to use react
+			FlowRouter.go('/omnichannel/current');
 			ChatRoom.remove(this.rid);
 			CachedChatRoom.save();
-			return FlowRouter.go('/home');
 		}
 
 		this.inquiry.set(inquiry);
--- a/apps/meteor/client/views/room/Header/Omnichannel/QuickActions/hooks/useQuickActions.tsx
+++ b/apps/meteor/client/views/room/Header/Omnichannel/QuickActions/hooks/useQuickActions.tsx
@@ -166,7 +166,8 @@ export const useQuickActions = (
 					throw new Error(departmentId ? t('error-no-agents-online-in-department') : t('error-forwarding-chat'));
 				}
 				dispatchToastMessage({ type: 'success', message: t('Transferred') });
-				FlowRouter.go('/omnichannel/current');
+				if (window.location.pathname.includes('current')) FlowRouter.go('/omnichannel/current');
+				else FlowRouter.go('/');
 				closeModal();
 			} catch (error: any) {
 				handleError(error);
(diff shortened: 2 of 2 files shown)


debdutdeb:
admittedly I've never tried omnichannel, so correct me if I'm saying something wrong here - 

 believe there are two previous_views, one, from /home -> sidebar::select_chat -> close_chat; another, /omnichannel/current_chats -> select_chat -> close_chat

with this patch, if someone is in the first path, will also get forwarded to current chats whereas he/she'd be expecting to go back /home. rather, what would be better, IMO, is if the user gets forwarded to the previous_path - not something static since the previous location is also not static.

nishant23122000:
Hmm, I was not aware of that, this chat section is also inside the sidebar menu (First path). so I feel like having the static route would be the best option. but now it seems like we need a dynamic way to handle the previous route.

nishant23122000:
I think we can check the active path before changing it to the previous path... (just normal if condition).

dudanogueira:
Not sure about the code base, but from a user perspective, whenever closing or forwarding a chat, a manager that have access to Current Chats and **accessed that chat from the Current Chat** view should be redirected back to Current Chats.

Consider you have 10 chats to close or forward, manually. Right now it gets you back to home instead of current chats, so you need to click on your avatar, and Omnichanel and then the next chat x 10 times.

This will be a great improvement on UX

nishant23122000:
Yes absolutely, I will deep dive into the codebase again and make sure we did not face any regression.

geekgonecrazy:
@dudanogueira from what I’ve seen looks like we had excellent feedback from PM on this one.  Would be good to summarize this here 🙏

geekgonecrazy:
I’m going to grab a few excerpts from conversation had internally about this.  

Initial PM Response 
> Design wise, users should not be redirected without a proactive interaction between the user and the UI. What is proposed is the introduction of a specific business rule for his use case, I get the point, but that's not scalable, so unfortunately we can't accept the contribution.

> FYI - The entire Current Chats list is gonna be rewritten and new UI/UX improvements will be introduced in short-term. It's part of the product roadmap, but not a priority now.

—
Further stating of the pain it solves:

> When the *manager* closes or forwards a chat, and have accessed it from the Current Chats view, it is redirected to the home page. 

> This is not be the expected behavior, as the previous view was *Current Chats* and not *Home*. This will improve UX for managers that needs to close of forward a lot of chats.

> Consider you have 10 chats to close. With this new behavior, after closing/forwarding the chat, you are right back at the Current Chats, and with just 1 click, you can access the next  chat.  

> Right now, you need to click on your avatar (1 click), then Omnichannel (2 clicks) and then on the next chat (3 clicks).

—

And a special rare look into a PMs mind.  And why I really wanted to share this.  

> Cool, now we know what the real pain point is so we can talk about problems instead of a small feature and "productize" the solution that will cover needs at scale (I love it :blush: ).
> 
> **Assume you have 10 chats to close. With this new behavior, after closing/forwarding the chat, you are back to the current chats, and with just 1 click, you can access the next chat.**
> 
> **Now, you need to click on your avatar (1 click), then Omnichannel (2 clicks) and then the next chat (3 clicks).**
> 
> Okay, so the solution solves the community member problem, but that also introduces coupling, when what we really want is cohesion, right? Look, the proposed solution starts from a need of **manager** users, they have a specific panel (Current Chats) to handle open chats and so on. However, the "solution" also touches on parts of the product that are designed for agents, not managers, and this is where we have a problem. The community member is incorporating a potential solution that should improve the manager's efficiency and productivity, but what if another community member replaces this "improvement" because of another specific need? For example, what if someone asks for another business rule when a manager/agent opens a chat in the **Omnichannel Contact Center** panel? Which business rule should we keep?
>
> As I said, the motivation behind the proposed solution is fair, but the solution itself does not meet the needs of the product because the best solution, in this case, is to improve the **Current Chats** list, turning it into a real management tool where managers will be able to do their work within the right domain/scope.
> 
> As I mentioned before, we have plans to replace the **Current Chats** list with a new panel called **Conversations** and there managers won't just find CTA's for specific needs but will, also, be able to watch the conversations and real-time so that they won't need to open the entire Chat Room because the CTA's they need are placed there. 
> 
> What can be done now? In case the community member is willing to help us release a solution that will help users at scale, he can add new CTA's on the Current Chat List row. If you look at the very right side of each chat row you'll find a button there allowing managers to delete closed chats. the delete button can be replaced with a popup/dropdown component where managers could find not just the delete chat button but, also, other two ones: **Close Chat** and **Forward Chat.**
> 
> This is how we can guide the community member through a solution that will remain in the codebase even after replacing the Current Chats List with a more robust management tool and not only the community member and existing customers will benefit from this great improvement, but we will also improve the UI/UX, which means a greater product sense.

KevLehman:
Based on the discussion, should we close this? 👀

PR #25358, opened 2022-05-02
Chore: Convert assets endpoint to Typescript

MarcosSpessatto:


apps/meteor/app/api/server/v1/assets.ts
```diff
@@ -44,7 +41,7 @@ API.v1.addRoute(
 	{
 		post() {
 			const { assetName, refreshAllClients } = this.bodyParams;
-			const isValidAsset = Object.keys(RocketChatAssets.assets).includes(assetName);
+			const isValidAsset = Object.keys(RocketChatAssets.assets).includes(assetName as string);
```
ggazzo, line 44: what about insert a proper test to guarantee this is actually a string?

apps/meteor/app/assets/server/assets.ts
```diff
@@ -301,23 +321,25 @@ export const RocketChatAssets = new (class {
 		return assetValue.cache;
 	}
 
-	getURL(assetName, options = { cdn: false, full: true }) {
-		const asset = settings.get(assetName);
+	public getURL(assetName: string, options = { cdn: false, full: true }): string {
+		const asset = settings.get<Record<string, any>>(assetName);
```
sampaiodiego, line 325: this returns a `IRocketChatAsset` right?
```suggestion
		const asset = settings.get<IRocketChatAsset>(assetName);
```

apps/meteor/app/assets/server/assets.ts
```diff
@@ -333,37 +355,37 @@ function addAssetToSetting(asset, value) {
 			asset,
 			public: true,
 			wizard: value.wizard,
-		},
+		} as Record<string, any>,
```
sampaiodiego, line 358: any special reason that you had to add this? is complaining some options are not in `ISettingAddOptions`? maybe the best should be adding them there.

apps/meteor/app/assets/server/assets.ts
```diff
@@ -333,37 +355,37 @@ function addAssetToSetting(asset, value) {
...
 	);
 
-	const currentValue = settings.get(key);
+	const currentValue = settings.get<Record<string, any>>(key);
```
sampaiodiego, line 361: same as before
```suggestion
	const currentValue = settings.get<IRocketChatAsset>(key);
```

apps/meteor/app/assets/server/assets.ts
```diff
@@ -174,39 +188,43 @@ const assets = {
...
+	public setAsset(binaryContent: BufferEncoding, contentType: string, asset: string): void {
+		if (!getAssetByKey(asset)) {
 			throw new Meteor.Error('error-invalid-asset', 'Invalid asset', {
 				function: 'RocketChat.Assets.setAsset',
 			});
 		}
 
 		const extension = mime.extension(contentType);
-		if (assets[asset].constraints.extensions.includes(extension) === false) {
+		if (getAssetByKey(asset).constraints.extensions.includes(extension as string) === false) {
```
sampaiodiego, line 212: I saw the return of `mime.extension` can actually be a `boolean`, do you know when it can return a `boolean`? and what we should do in that case?

apps/meteor/app/assets/server/assets.ts
```diff
@@ -466,29 +488,29 @@ Meteor.methods({
...
 WebApp.connectHandlers.use(
 	'/assets/',
-	Meteor.bindEnvironment(function (req, res, next) {
+	Meteor.bindEnvironment((req: Request, res: Response, next: NextFunction) => {
 		const params = {
 			asset: decodeURIComponent(req.url.replace(/^\//, '').replace(/\?.*$/, '')).replace(/\.[^.]*$/, ''),
 		};
 
-		const file = assets[params.asset] && assets[params.asset].cache;
+		const file = getAssetByKey(params.asset) && getAssetByKey(params.asset).cache;
```
sampaiodiego, line 496: why not store `getAssetByKey(params.asset)` in a variable and use it in the rest of the function? you checking if it has values in a lot of places

apps/meteor/app/assets/server/assets.ts
```diff
@@ -521,5 +545,5 @@ WebApp.connectHandlers.use(
 		res.setHeader('Content-Length', file.size);
 		res.writeHead(200);
 		res.end(file.content);
-	}),
+	}) as any,
```
sampaiodiego, line 548: is this really needed?

sampaiodiego (changes requested):
as a rule of thumb, when I see a type casting, I think there is something wrong.. I'd ask to avoid as must as possible. I've made several change requests and questions on most of them.

apps/meteor/app/api/server/v1/assets.ts
```diff
@@ -44,7 +42,10 @@ API.v1.addRoute(
 	{
 		post() {
 			const { assetName, refreshAllClients } = this.bodyParams;
-			const isValidAsset = Object.keys(RocketChatAssets.assets).includes(assetName);
+			if (!isAssetsUnsetAssetProps(this.bodyParams)) {
+				return API.v1.failure('invalid-body-params', isAssetsUnsetAssetProps.errors?.map((e) => e.message).join('\n '));
+			}
+			const isValidAsset = Object.keys(RocketChatAssets.assets).includes(assetName as string);
```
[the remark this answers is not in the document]
MarcosSpessatto: For some reason the type for this property is `string | undefined` 🤔

apps/meteor/app/assets/server/assets.ts
```diff
@@ -416,7 +438,7 @@ Meteor.methods({
 			});
 		}
 
-		const _hasPermission = hasPermission(Meteor.userId(), 'manage-assets');
+		const _hasPermission = hasPermission(Meteor.userId() as string, 'manage-assets');
```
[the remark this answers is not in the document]
MarcosSpessatto: The validation is already there. But even with the validation, I need to cast it, any other idea?

apps/meteor/app/api/server/v1/assets.ts
```diff
@@ -38,13 +36,16 @@ API.v1.addRoute(
 	},
 );
 
-API.v1.addRoute(
+API.v1.addRoute<'assets.unsetAsset', { authRequired: true }>(
```
ggazzo, line 39:
```suggestion
API.v1.addRoute(
```

ggazzo (dismissed):

ggazzo (approved):

PR #25365, opened 2022-05-02
Chore: Tests with Playwright (task: ROC-43)

weslley543:

- Add coverage collected by the playwright test
- Removed unused methods in page object


.github/workflows/build_and_test.yml
```diff
@@ -595,38 +603,6 @@ jobs:
           docker build -t $IMAGE_NAME .
           docker push $IMAGE_NAME
 
-  services-image-build-check:
```
ggazzo, line 598: why this test is no longer available ?

PR #25382, opened 2022-05-03
Chore: Improve template generation with eslint

ggazzo:

### DIFF
--- a/_templates/package/new/.eslintrc.ejs.t
+++ b/_templates/package/new/.eslintrc.ejs.t
@@ -1,0 +1,15 @@
+---
+to: packages/<%= name %>/.eslintrc
+---
+{
+	"extends": ["@rocket.chat/eslint-config"],
+	"overrides": [
+		{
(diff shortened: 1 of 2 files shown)


_templates/package/new/.eslintrc.ejs.t
```diff
@@ -0,0 +1,15 @@
+---
+to: packages/<%= name %>/.eslintrc
+---
+{
+	"extends": ["@rocket.chat/eslint-config"],
+	"overrides": [
```
sampaiodiego, line 6: are these overrides always needed?

sampaiodiego (changes requested):

_templates/package/new/.eslintrc.ejs.t
sampaiodiego: if every package will have this override, doesn't make sense to make it default, so moving it to the eslint package?

PR #25398, opened 2022-05-04
[FIX] Upgrade tab loader in incorrect position

guijun13:

- Add invisible prop to iframe when loading state is active.

## Steps to test or reproduce

- Administration > Upgrade tab page
- The loader should be in the center of the page and it should disappear when content appears.


apps/meteor/client/views/admin/upgrade/UpgradePage/UpgradePage.tsx
```diff
@@ -95,11 +93,21 @@ const UpgradePage = (): ReactElement => {
 			{isMobile && <PageHeader title='' />}
 			{!hasConnection && <UpgradePageError />}
 			{hasConnection && isLoading && (
-				<Box pb='x24'>
+				<Box width={'100%'} height={'100%'} position={'absolute'} display={'flex'} justifyContent={'center'} alignItems={'center'}>
```
tassoevan, line 96: It's safe to write string values for JSX element attributes without curly brackets:

```suggestion
				<Box width='100%' height='100%' position='absolute' display='flex' justifyContent='center' alignItems='center'>
```

tassoevan (approved):

PR #25400, opened 2022-05-04
[NEW] [VideoConference] Direct call negotiation between clients

pierre-lehnen-rc:
### New endpoint
The `video-conference.start` endpoint is used to request a new video call based on a room ID. For now it only accepts regular direct message rooms, in which case it only returns a call ID and user ID. This endpoint will still be expanded on.

### VideoConfManager

This client class is responsible for managing the status of videocalls. It encapsulates all the communication with the server and remote clients. 

So far it has three public entry points:

#### Start Call

function: `VideoConfManager.startCall(roomId)`

This is the entry point for every video call. This function will:

1. Call `video-conference.start` to get a new call ID
2. In case of a regular DM, it'll then send this call ID to the remote client every 3 seconds for up to 30 seconds.
3. It'll then wait for an acceptance from the remote client or for the timeout of the final attempt to communicate.

If the VideoConfManager is already calling a remote user when this function is called, it'll throw an exception.

Other room types and Multi user DMs are not yet supported.

#### Accept Incoming Direct Call
function: `VideoConfManager.acceptIncomingCall(callId)`

This function is used to accept an incoming direct call. It'll notify the remote user that the call has been accepted. If the callId is not the ID of a call that is currently ringing, it'll throw an exception. 

#### Mute Incoming Call
function `VideoConfManager.muteIncomingCall(callId)`

This function will prevent the `direct/ringing` event from being triggered for this callId again. It doesn't affect the workflow in any other way. 

#### Events

The VideConfManager extends the Emitter class to provide events that the client can listen to in order to react to status changes.

All of those events will always have the same params: `{ uid, rid, callId }`. The `uid` will always be the id of the remote user.

1. **direct/cancel** : This event is triggered when we give up trying to call a remote user. Use it to stop rendering any "calling" animation.
2. **direct/ringing**: This event is triggered when a remote user is calling us. It can be triggered multiple times for the same call. Use this to start displaying a "ringing" animation.
3. **direct/lost**: We lost an incoming call, either because the remote user gave up on calling or because the connection with the remote user was affected.
4. **direct/failed**: We tried to accept an incoming call but the process failed. Either because of a race condition or because the connection with the remote user was affected. Use this to warn the user that an error occurred.
5. **direct/accepted**: A remote user has accepted our call. This is currently the end of the workflow.


apps/meteor/app/api/server/v1/videoConference.ts
```diff
@@ -23,3 +26,27 @@ API.v1.addRoute(
 		},
 	},
 );
+
+API.v1.addRoute(
+	'video-conference.start',
```
ggazzo, line 31: I dont think we need `.start`, since we are using `POST` [POST] video-conference should be enough

pierre-lehnen-rc: This is the standard for the v1 API.
v2 will be RESTful

PR #25403, opened 2022-05-05
[NEW] [VideoConference] System Messages

pierre-lehnen-rc:

This PR adds 6 new system messages in preparation for the new video conference:

1. `video-direct-calling` to be used when an user is calling another
2. `video-direct-missed` to be used when a direct call was not answered
3. `video-direct-started` to be used when a direct call was accepted and started
4. `video-direct-ended` to be used when a provider confirms that a direct call has ended
5. `video-conference-started` to be used when a group conference is started
6. `video-conference` to be used to list active group conferences in a room.

Those message types are only used to display localized messages, their actions and visual elements will be attached to the actual message objects dynamically.


packages/core-typings/src/IMessage/IMessage.ts
```diff
@@ -232,3 +242,11 @@ export type IMessageInbox = IMessage & {
 
 export const isIMessageInbox = (message: IMessage): message is IMessageInbox => 'email' in message;
 export const isVoipMessage = (message: IMessage): message is IVoipMessage => 'voipData' in message;
+
+export interface IVideoConferenceMessage extends IMessage {
+	videoConference: {
+		name?: string;
+	};
```
sampaiodiego, line 249:
```suggestion
	videconf: {
		title?: string;
	};
```

apps/meteor/packages/rocketchat-i18n/i18n/en.i18n.json
```diff
@@ -4699,6 +4699,13 @@
   "Video_message": "Video message",
   "Videocall_declined": "Video Call Declined.",
   "Video_and_Audio_Call": "Video and Audio Call",
+  "video_conference_ended": "__username__ started __conference__ video conference.",
```
sampaiodiego, line 4702: will be there an ended event for video conference?
```suggestion
  "video_conference_ended": "__username__ ended __conference__ video conference.",
```

apps/meteor/app/lib/lib/MessageTypes.ts
```diff
@@ -310,6 +311,73 @@ Meteor.startup(function () {
...
+		system: true,
+		message: 'video_direct_ended',
+	});
+	MessageTypes.registerType({
+		id: 'video-conference-started',
+		system: true,
+		message: 'video_conference_started',
+		data(message) {
+			return {
+				conference: (isVideoConferenceMessage(message) && message.videoConference.name) || '',
```
sampaiodiego, line 355: should we provide something in case the title is empty? right now it would render:
`diego.sampaio started  video conference`

not sure this will look good..

sampaiodiego (changes requested):

apps/meteor/packages/rocketchat-i18n/i18n/en.i18n.json
pierre-lehnen-rc: Actually the message is the same in both cases, it's only the uikit blocks that are going to be different, so there is no need for an extra message type here, I'll remove this one.

apps/meteor/app/lib/lib/MessageTypes.ts
pierre-lehnen-rc: I've discussed with the design team and we'll keep it empty as it'll only be empty in case of error.

packages/core-typings/src/IMessage/IMessage.ts
```diff
@@ -232,3 +241,11 @@ export type IMessageInbox = IMessage & {
 
 export const isIMessageInbox = (message: IMessage): message is IMessageInbox => 'email' in message;
 export const isVoipMessage = (message: IMessage): message is IVoipMessage => 'voipData' in message;
+
+export interface IVideoConferenceMessage extends IMessage {
+	videoConf: {
+		title?: string;
```
pierre-lehnen-rc, line 247:
```suggestion
		title: string;
```

PR #25410, opened 2022-05-05
[FIX] fixes HTML sanitizing error.

MartinSchoeler:

If the user sent a HTML message over our product to a livechat user the HTML would get rendered on the message box, this prevents it from happening.

## Steps to test or reproduce

try to send some html tags in livechat

## Further comments

**Will be ready to review as soon as https://github.com/RocketChat/Rocket.Chat/pull/25312 is merged**


ggazzo:
could we use a better description?

cauefcr:
> could we use a better description?

if the user sent a HTML message over our product to a livechat user the HTML would get rendered on the message box, this prevents it from happening.

MartinSchoeler:
Okay so let's try to fix another related issue on this PR as well, we are currently having some issue with escaped characters like this 

image

cauefcr:
> Okay so let's try to fix another related issue on this PR as well, we are currently having some issue with escaped characters like this
> 
> image

Solved the problem, the cause was using the parse function twice :see_no_evil:.

KevLehman (approved):
Function-wise it works. Waiting for other reviewers :)

tiagoevanp (approved):

thrakks:
Is there any possibility of making this toggle-able?

I have users that display useful information in channels using html tags for tables etc. and are definitely not considered an internal threat vector when we only provide rocketchat internally.

thrakks:
Also, this is still partially an "issue" in 5.1.0. Specifically HTML code will render in search:

image

thrakks:
And it's broken html render in admin menus:
image

PR #25421, opened 2022-05-06
Chore: Rewrite mail-messages to ts

KevLehman:


apps/meteor/app/mail-messages/server/functions/sendMail.ts
```diff
@@ -47,18 +51,20 @@ export const sendMail = function (from, subject, body, dryrun, query) {
 				});
 			});
 	}
-	return Meteor.users.find(userQuery).forEach(function (user) {
-		if (user && user.emails && Array.isArray(user.emails) && user.emails.length) {
+
+	return Meteor.users.find(userQuery).forEach(function (u) {
+		const user: Pick<IUser, '_id' | 'name' | 'emails'> & { createdAt?: Date } = u;
```
sampaiodiego, line 56: why not rely on `createdAt` from `IUser`? any reason for it to be optional here?

sampaiodiego (changes requested):

apps/meteor/app/mail-messages/server/functions/sendMail.ts
KevLehman: Not actually. Forgot to update the type to the other i used

sampaiodiego (approved):

PR #25436, opened 2022-05-09
[NEW] Ability for RC server to check the business hour for a specific department

murtaza98:
## Steps to test or reproduce

1. Assign the department X to the LiveChat using the widget API code
2. Enable and setup multiple business hours
3. Assign the department X to one of the business hours
4. Enable the Display offline form option

**Expected Behavior:**
As long as the business hour assigned to department X is closed, the offline form should always be displayed for the visitor.

**Actual Behavior:**
Sometimes, when the business hour is closed, the LiveChat widget is online instead of offline form. Then, when a visitor initiates a chat, an alert is displayed on the widget saying that there is no agent available for chatting.

## Further comments

Related ClickUp task: https://app.clickup.com/t/2fdmbc1


packages/livechat/src/routes/SwitchDepartment/connector.js
```diff
@@ -16,10 +16,10 @@ const SwitchDepartmentConnector = ({ ref, ...props }) => (
 					fontColor: customFontColor,
 					iconColor: customIconColor,
 				} = {},
+				guest,
```
murtaza98, line 19: General note regarding SwitchDepartment component: Earlier this component was storing the department value in `state.department` property, however throughout the rest of the codebase, the department value was stored under `state.guest.department`. So this new change makes sure that this component follows the same pattern as the rest of the codebase & stores the department value in `state.guest.department`

packages/livechat/src/lib/hooks.js
```diff
@@ -60,13 +60,17 @@ const api = {
...
+	async setDepartment(value) {
+		const { config: { departments = [] }, department: existingDepartment } = store.state;
 
 		const dept = departments.find((dep) => dep._id === value || dep.name === value);
 		const department = (dept && dept._id) || '';
 
 		updateIframeGuestData({ department });
+
+		if (department !== existingDepartment) {
+			await loadConfig();
```
KevLehman, line 72: won't this cause the same issue we had some time ago where the room got "discarded" on the UI so the visitor had to open a new one on livechat?

murtaza98: Yups, you're right, that would indeed be a problem. I think we'll need to reload the messages too after config changes

murtaza98: Tested the new change and it works. Here's the recording of my test -> https://www.loom.com/share/fab58ba6e1fe48238784d15c5c0b2bc3 Let me know if I miss anything here

packages/livechat/widget-demo.html
```diff
@@ -48,7 +48,7 @@
 			var h = d.getElementsByTagName(s)[0],
 				j = d.createElement(s);
 			j.async = true;
-			j.src = 'build/rocketchat-livechat.min.js?_=' + Math.random();
+			j.src = 'dist/rocketchat-livechat.min.js?_=' + Math.random();
```
murtaza98, line 51: Not connected to this task in any way... but needed to update it since with the new monorepo structure, `build` folder is no longer generated & this js file resides in the `dist` folder instead

KevLehman (dismissed):

tiagoevanp (changes requested):
This fix doesn't look correlated to the problem mentioned by the customer.

packages/livechat/src/lib/main.js
```diff
@@ -42,27 +42,30 @@ export const loadConfig = async () => {
...
 		...businessUnit && { businessUnit },
+		...department && { department },
 	});
 
 	await store.setState({
 		config,
 		agent: agent && agent.hiddenInfo ? { hiddenInfo: true } : agent, // TODO: revert it when the API is updated
 		room,
-		user,
+		user: guest,
```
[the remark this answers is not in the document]
murtaza98: Yeah, it seems both these variables were holding the same data so removed `guest` & kept `user`

KevLehman (approved):

tiagoevanp (approved):

PR #25438, opened 2022-05-09
[FIX] Display relevant error toast message

Dnouv:

This PR fixes the toast error message on "Users" tab by catching and showing the relevant and brief error info.

image

## Issue(s)

Currently if a admin tries to save a user with the existing username, he will be prompted with an toast message "Error" which does not helps the user to rectify the issue, however the correct error is returned from the server.

## Steps to test or reproduce

1. Go to the "Users" tab.
2. Try to add a new user with the same name or e-mail address as the existing one.
3. An toast message "Error" is shown.
Expected:
An message with "_E-mail_ or _username_ already exists"


dudanogueira:
Closing in favor of https://github.com/RocketChat/Rocket.Chat/pull/26008

PR #25441, opened 2022-05-09
[NEW] Use setting to determine if initial general channel is needed

felipe-menelau:

- Adds flag responsible for overwriting #general channel creation

## Issue(s)

On cloud-side the Saas trial flow is getting redone, and in the new flow many pieces of information once captured by the wizard are now being captured on the trial flow, for that reason saas instances will get injected on to them the flag that skips the setup-wizard with the consequence that #general channel is no longer gettin auto-created. The proposed fix is another flag that lets cloud override the previous behaviour and still auto-create the general channel for saas instances that skipped wizard.


apps/meteor/app/lib/server/startup/settings.ts
```diff
@@ -9,6 +9,12 @@ settingsRegistry.add('uniqueID', process.env.DEPLOYMENT_ID || Random.id(), {
 	public: true,
 });
 
+settingsRegistry.add('Force_Create_General_Channel', 'false', {
```
geekgonecrazy, line 12:
```suggestion
settingsRegistry.add('Needs_Create_General_Channel', 'true', {
```

apps/meteor/server/startup/initialData.js
```diff
@@ -10,9 +10,17 @@ import { settings } from '../../app/settings/server';
 import { checkUsernameAvailability, addUserToDefaultChannels } from '../../app/lib/server';
 import { Settings } from '../../app/models/server/raw';
 import { validateEmail } from '../../lib/emailValidator';
+import { SystemLogger } from '../lib/logger/system';
 
 Meteor.startup(async function () {
-	if (settings.get('Show_Setup_Wizard') === 'pending' && !Rooms.findOneById('GENERAL')) {
+	if (
+		(settings.get('Show_Setup_Wizard') === 'pending' || settings.get('Force_Create_General_Channel') === true) &&
```
geekgonecrazy, line 17:
```suggestion
		(settings.get('Show_Setup_Wizard') === 'pending' || settings.get('Needs_Create_General_Channel') === true) &&
```

apps/meteor/server/startup/initialData.js
```diff
@@ -10,9 +10,17 @@ import { settings } from '../../app/settings/server';
...
+		!Rooms.findOneById('GENERAL')
+	) {
+		if (settings.get('Force_Create_General_Channel')) {
+			SystemLogger.info('FORCE_CREATE_GENERAL Set to true. Forcing channel creation if not already present');
+		}
```
geekgonecrazy, line 22:
```suggestion
```

apps/meteor/server/startup/initialData.js
```diff
@@ -10,9 +10,17 @@ import { settings } from '../../app/settings/server';
...
+
 		Rooms.createWithIdTypeAndName('GENERAL', 'c', 'general', {
 			default: true,
 		});
```
geekgonecrazy, line 26:
```suggestion
		});
		
		Settings.updateValueById('Needs_Create_General_Channel', false)
```

geekgonecrazy (commented):
Dropping a few comments here for a suggestion to alter it.

Making the setting true by default.  Then no matter what on Rocket.Chat wizard or no wizard it will create this channel if the value is true.   Then once created its set to false keeping it from being created by true.

The problem with the other flow is if this value stays true by injecting (like we'd have to do on SaaS) it'll keep trying to create that channel over and over. But we actually just want it to create regardless of the wizard.  This prevents us from needing to pass this value in.

Added benefit.. if anyone loses their General channel they can set this back to true and it'll re-create it :)

sampaiodiego:
I was doing some changes while you're adding comments @geekgonecrazy 😬  pls take another look =)

geekgonecrazy:
@sampaiodiego I think you must have been looking inside of my mind.  Thats exactly what I was suggesting 😂

sampaiodiego (approved):

PR #25444, opened 2022-05-09
[FIX] Removing user also removes them from Omni collections

cauefcr:
https://app.clickup.com/t/23tt12m

## Steps to test or reproduce

out

### DIFF
--- a/apps/meteor/app/lib/server/functions/deleteUser.ts
+++ b/apps/meteor/app/lib/server/functions/deleteUser.ts
@@ -72,7 +72,7 @@ export async function deleteUser(userId: string, confirmRelinquish = false): Pro
 			LivechatUnitMonitors.removeByMonitorId(userId);
 		}
 
-		if (user.roles.find((r: string) => r === 'livechat-agent' || r === 'livechat-monitor')){
+		if (user.roles.find((r: string) => r === 'livechat-agent' || r === 'livechat-monitor')) {
 			// Remove user as contact manager for all visitors
 			LivechatVisitors.removeContactManagerByUsername(user.username);
 		}
--- a/apps/meteor/app/lib/server/functions/deleteUser.ts
+++ b/apps/meteor/app/lib/server/functions/deleteUser.ts
@@ -62,12 +62,12 @@ export async function deleteUser(userId: string, confirmRelinquish = false): Pro
 
 		Subscriptions.removeByUserId(userId); // Remove user subscriptions
 
-		if (user.roles.find((r: string) => r === 'livechat-agent') !== undefined) {
+		if (user.roles.find((r: string) => r === 'livechat-agent')) {
 			// Remove user as livechat agent
 			LivechatDepartmentAgents.removeByAgentId(userId);
 		}
 
-		if (user.roles.find((r: string) => r === 'livechat-monitor') !== undefined) {
+		if (user.roles.find((r: string) => r === 'livechat-monitor')) {
 			// Remove user as Unit Monitor
 			LivechatUnitMonitors.removeByMonitorId(userId);
 		}
--- a/apps/meteor/app/lib/server/functions/deleteUser.ts
+++ b/apps/meteor/app/lib/server/functions/deleteUser.ts
@@ -16,7 +16,7 @@ import { LivechatUnitMonitors } from '../../../../ee/app/models/server';
 
 export async function deleteUser(userId: string, confirmRelinquish = false): Promise<void> {
 	const user = Users.findOneById(userId, {
-		fields: { username: 1, avatarOrigin: 1, federation: 1 },
+		fields: { username: 1, avatarOrigin: 1, federation: 1, roles: 1 },
 	});
 
 	if (!user) {
--- a/apps/meteor/app/lib/server/functions/deleteUser.ts
+++ b/apps/meteor/app/lib/server/functions/deleteUser.ts
@@ -72,9 +72,10 @@ export async function deleteUser(userId: string, confirmRelinquish = false): Pro
 			LivechatUnitMonitors.removeByMonitorId(userId);
 		}
 
-		// Remove user as contact manager for all visitors
-		LivechatVisitors.removeContactManagerByUsername(user.username);
-
+		if (user.roles.find((r: string) => r === 'livechat-agent' || 'livechat-monitor')){
+			// Remove user as contact manager for all visitors
+			LivechatVisitors.removeContactManagerByUsername(user.username);
+		}
 		// removes user's avatar
 		if (user.avatarOrigin === 'upload' || user.avatarOrigin === 'url') {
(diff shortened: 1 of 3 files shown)


apps/meteor/app/lib/server/functions/deleteUser.ts
```diff
@@ -61,14 +63,27 @@ export async function deleteUser(userId: string, confirmRelinquish = false): Pro
 
 		Subscriptions.removeByUserId(userId); // Remove user subscriptions
 
+		if (hasRole(userId, 'livechat-agent')) {
```
murtaza98, line 66: Although there isn't anything wrong with this logic, I feel this is not required at all and we can just call the database directly here without this check & it will not delete anything for users with no `livechat-agent` role. Calling this method is just adding another database call which I don't feel is required here, so this could be a minor optimization here.

apps/meteor/app/lib/server/functions/deleteUser.ts
```diff
@@ -61,14 +63,27 @@ export async function deleteUser(userId: string, confirmRelinquish = false): Pro
...
+			// Remove user as livechat agent
+			LivechatDepartmentAgents.removeByAgentId(userId);
+		}
+
+		if (hasRole(userId, 'livechat-monitor')) {
```
murtaza98, line 71: Same comment goes here

apps/meteor/app/models/server/models/LivechatVisitors.js
```diff
@@ -247,6 +247,20 @@ export class LivechatVisitors extends Base {
 		const query = { _id };
 		return this.remove(query);
 	}
+
+	removeContactManagerByUsername(manager) {
+		const query = {
+			contactManager: {
+				username: manager,
```
murtaza98, line 254: Since this is a query, do you mind adding an index on this property?

apps/meteor/app/lib/server/functions/deleteUser.ts
cauefcr: Well, this also prevents from making a unnecessary call on deletion when the is user is not an agent.

apps/meteor/app/models/server/models/LivechatVisitors.js
cauefcr: Makes sense, adding!

KevLehman (changes requested):

apps/meteor/app/lib/server/functions/deleteUser.ts
```diff
@@ -61,14 +62,27 @@ export async function deleteUser(userId: string, confirmRelinquish = false): Pro
...
 		Subscriptions.removeByUserId(userId); // Remove user subscriptions
 
+		if (user.roles.find((r: string) => r === 'livechat-agent') !== undefined) {
+			// Remove user as livechat agent
+			LivechatDepartmentAgents.removeByAgentId(userId);
+		}
+
+		if (user.roles.find((r: string) => r === 'livechat-monitor') !== undefined) {
+			// Remove user as Unit Monitor
+			LivechatUnitMonitors.removeByMonitorId(userId);
```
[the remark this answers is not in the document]
cauefcr: I believe it will stay without monitors, but let me check.

apps/meteor/app/lib/server/functions/deleteUser.ts
```diff
@@ -61,14 +62,27 @@ export async function deleteUser(userId: string, confirmRelinquish = false): Pro
 
 		Subscriptions.removeByUserId(userId); // Remove user subscriptions
 
+		if (user.roles.find((r: string) => r === 'livechat-agent') !== undefined) {
+			// Remove user as livechat agent
+			LivechatDepartmentAgents.removeByAgentId(userId);
```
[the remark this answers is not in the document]
cauefcr: Good question, let me check.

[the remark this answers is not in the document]
cauefcr: The user gets disconected, and the livechat room becomes unusable from the client side.

apps/meteor/app/lib/server/functions/deleteUser.ts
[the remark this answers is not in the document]
cauefcr: The unit stays without users, nothing more.

apps/meteor/app/lib/server/functions/deleteUser.ts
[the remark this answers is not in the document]
KevLehman: This would be a nice feature 🤔

apps/meteor/app/lib/server/functions/deleteUser.ts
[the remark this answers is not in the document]
KevLehman: There's a thread for discussing this 👀

apps/meteor/app/lib/server/functions/deleteUser.ts
```diff
@@ -61,14 +62,27 @@ export async function deleteUser(userId: string, confirmRelinquish = false): Pro
...
+		}
+
+		if (user.roles.find((r: string) => r === 'livechat-monitor')) {
+			// Remove user as Unit Monitor
+			LivechatUnitMonitors.removeByMonitorId(userId);
+		}
+
+		// Remove user as contact manager for all visitors
+		LivechatVisitors.removeContactManagerByUsername(user.username);
```
KevLehman, line 76: This should happen only when the user is agent/monitor/manager too 👀

KevLehman (changes requested):

KevLehman:
btw, pipe is failing :(

apps/meteor/app/models/server/models/LivechatVisitors.js
```diff
@@ -13,6 +13,7 @@ export class LivechatVisitors extends Base {
 		this.tryEnsureIndex({ 'visitorEmails.address': 1 }, { sparse: true });
 		this.tryEnsureIndex({ name: 1 }, { sparse: true });
 		this.tryEnsureIndex({ username: 1 });
+		this.tryEnsureIndex({ contactManager: 1 }, { sparse: true });
```
KevLehman, line 16: Instead of an index on full contactManager obj, try with an index on "contactManager.username" as it's the used prop 👀

KevLehman (changes requested):

apps/meteor/app/models/server/models/LivechatVisitors.js
cauefcr: makes sense, doing it.

apps/meteor/app/lib/server/functions/deleteUser.ts
cauefcr: done!

KevLehman (approved):

PR #25448, opened 2022-05-09
[FIX] Overwritten settings not getting cached

sampaiodiego:

Settings overwritten by environment variables were getting stored in the cache during startup, causing `watch*` callbacks to be executed using the old values.

### DIFF
--- a/apps/meteor/app/settings/server/SettingsRegistry.ts
+++ b/apps/meteor/app/settings/server/SettingsRegistry.ts
@@ -163,12 +163,15 @@ export class SettingsRegistry {
 					}),
 				},
 			);
+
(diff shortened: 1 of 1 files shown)


sampaiodiego:
this is good for review again =) I've completely changed the initial changes, I'm not forcing cache to be updated anymore when a setting value is overwritten, it is now setting up oplog listener before doing the updates, so if an update happens the correct flow will still happen:

```mermaid
graph TD;
    A([update setting on db])-->B([oplog event triggered]);
    B-->C([received by listener]);
    C-->D([update cache]);
    D-->E([trigger watchers]);
```

ggazzo (approved):

PR #25451, opened 2022-05-10
[FIX] allow only livechat-agents to be contact manager for any omnichannel contact 

murtaza98:
## Issue(s)

image

## Further comments

Tasks:

- [x] Frontend change on auto-select component
- [x] Backend API checks

### DIFF
--- a/apps/meteor/app/livechat/server/lib/Contacts.js
+++ b/apps/meteor/app/livechat/server/lib/Contacts.js
@@ -1,13 +1,26 @@
 import { check } from 'meteor/check';
+import { Meteor } from 'meteor/meteor';
 import s from 'underscore.string';
 
 import { LivechatVisitors, LivechatCustomField, LivechatRooms, Rooms, LivechatInquiry, Subscriptions } from '../../../models';
+import { Users } from '../../../models/server/raw';
 
 export const Contacts = {
 	registerContact({ token, name, email, phone, username, customFields = {}, contactManager = {} } = {}) {
 		check(token, String);
 
 		const visitorEmail = s.trim(email).toLowerCase();
+
+		if (contactManager?.username) {
+			// verify if the user exists with this username and has a livechat-agent role
+			const user = Users.findOneByUsername(contactManager.username, { projection: { roles: 1 } });
+			if (!user) {
+				throw new Meteor.Error('error-contact-manager-not-found', `No user found with username ${contactManager.username}`);
+			}
+			if (!user.roles || !Array.isArray(user.roles) || !user.roles.includes('livechat-agent')) {
+				throw new Meteor.Error('error-invalid-contact-manager', 'The contact manager must have the role "livechat-agent"');
+			}
+		}
 
 		let contactId;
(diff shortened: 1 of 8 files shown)


KevLehman (commented):
What if some random customer has a non-agent as contact manager? 👀

murtaza98:
> What if some random customer has a non-agent as contact manager? :eyes:

But that's not the way this feature was designed for. The whole idea of the Contact manager is that if that manager is online and the visitor tries to contact the company, then the manager the chat reaches directly to this manager instead of any other agent. So if a non-agent is assigned as a contact manager to any contact, then they won't be able to take the chat (since only agents can serve a chat) - hence I believe we need to add this filter over here to prevent such cases

KevLehman:
No no, that's fine. I mean, what happens if some customer _already_ has non-agents as contact managers after this fix is merged 👀 
Will UI crash? Will something odd happen? 😬

murtaza98:
> No no, that's fine. I mean, what happens if some customer _already_ has non-agents as contact managers after this fix is merged eyes Will UI crash? Will something odd happen? grimacing

@KevLehman I checked this case & I didn't find anything odd. Nothing breaks on the UI, however the only thing is that the existing contact manager (who isn't a agent) won't appear in the select dropdown

KevLehman (dismissed):

murtaza98:
Making this a draft for now since blocked on agent's page not getting loaded

PS: suggestions from the QA team to handle a case for a user losing an agent role.

murtaza98:
Hi @KevLehman @RocketChat/chat-engine May I ask for your approvals again? I lost earlier approvals due to some changes which QA team suggested

apps/meteor/app/livechat/server/lib/Contacts.js
```diff
@@ -1,14 +1,27 @@
...
 
 		const visitorEmail = s.trim(email).toLowerCase();
 
+		if (contactManager?.username) {
+			// verify if the user exists with this username and has a livechat-agent role
+			const user = Promise.await(Users.findOneByUsername(contactManager.username, { projection: { roles: 1 } }));
+			if (!user) {
+				throw new Meteor.Error('error-contact-manager-not-found', `No user found with username ${contactManager.username}`);
+			}
+			if (!user.roles || !Array.isArray(user.roles) || !user.roles.includes('livechat-agent')) {
```
KevLehman, line 20: Personally i don't like checking for roles here, but for now that would work.

Pls create a task for changing this check to be a permission in the future 👀

murtaza98: I agree to tackle this issue separately since we also have similar usage elsewhere in the code like https://github.com/RocketChat/Rocket.Chat/blob/993745d3996450036e8c39150a0b6926cdbbd869/apps/meteor/app/livechat/server/api/lib/users.js#L51

Here's the task for it btw: https://app.clickup.com/t/2f82np6

ggazzo (approved):

PR #25489, opened 2022-05-12
[NEW] [VideoConference] Join and Cancel Direct Call

pierre-lehnen-rc:


apps/meteor/app/models/server/raw/VideoConference.ts
```diff
@@ -1,9 +1,18 @@
...
 
 export class VideoConferenceRaw extends BaseRaw<IVideoConference> {
 	protected modelIndexes(): IndexSpecification[] {
 		return [{ key: { rid: 1, status: 1, createdAt: 1 }, unique: false }];
 	}
+
+	public updateOneById(
+		_id: string,
+		update: UpdateQuery<IVideoConference> | Partial<IVideoConference>,
+		options?: UpdateOneOptions & { multi?: boolean },
```
sampaiodiego, line 14: Update One should not have a `multi` option, it is intended to update just **one** record..

apps/meteor/server/services/video-conference/service.ts
```diff
@@ -0,0 +1,186 @@
...
+		// #ToDo: Load this from the Apps-Engine
+		return JitsiApp.customizeUrl(call, user, options);
+	}
+
+	private async addUserToCall(
+		callId: IVideoConference['_id'],
+		{ _id, username, name }: AtLeast<IUser, '_id' | 'username' | 'name'>,
+	): Promise<void> {
+		this.VideoConference.updateOneById(callId, {
+			$addToSet: {
```
sampaiodiego, line 176: move this operation to model pls.

apps/meteor/server/services/video-conference/service.ts
```diff
@@ -0,0 +1,186 @@
...
+
+		const user = await this.Users.findOneById<IUser>(caller, {});
+		if (!user) {
+			throw new Error('failed-to-load-own-data');
+		}
+
+		const callId = new ObjectID().toHexString();
+		// #ToDo: Generate the 'calling' message using the callId
+
+		await this.VideoConference.insertOne({
```
sampaiodiego, line 120: an idea that is not standardized nor well defined, but what do you think of having a `create` method, so you provide the real data and it adds the "hardcoded" one?

apps/meteor/server/services/video-conference/service.ts
```diff
@@ -0,0 +1,186 @@
...
+			// Are you trying to call yourself?
+			throw new Error('invalid-call-target');
+		}
+
+		const user = await this.Users.findOneById<IUser>(caller, {});
+		if (!user) {
+			throw new Error('failed-to-load-own-data');
+		}
+
+		const callId = new ObjectID().toHexString();
```
sampaiodiego, line 117: the `_id` generation should be in the model layer.

apps/meteor/server/services/video-conference/service.ts
```diff
@@ -0,0 +1,186 @@
...
+	}
+
+	private async getUrl(
+		call: IVideoConference,
+		user: AtLeast<IUser, '_id' | 'username' | 'name'>,
+		options: VideoConferenceJoinOptions,
+	): Promise<string> {
+		if (!call.url) {
+			call.url = await this.generateNewUrl(call);
+			this.VideoConference.updateOneById(call._id, { $set: { url: call.url } });
```
sampaiodiego, line 164: this should be in the model as well `VideoConference.setUrlById`

sampaiodiego (changes requested):

apps/meteor/server/services/video-conference/service.ts
```diff
@@ -0,0 +1,238 @@
...
+		this.Rooms.incMsgCountById(rid, 1);
+		return id;
+	}
+
+	private async createDirectCallMessage(rid: IRoom['_id'], user: IUser): Promise<IMessage['_id']> {
+		return this.createMessage('video-direct-calling', rid, user);
+	}
+
+	private async changeMessageType(messageId: string, newType: MessageTypesValues): Promise<void> {
+		this.Messages.setTypeById(messageId, newType);
```
sampaiodiego, line 165: since this method is returning `void`, if we don't `await` here, others awaiting for `changeMessageType` will have no effect.
```suggestion
		await this.Messages.setTypeById(messageId, newType);
```

apps/meteor/server/services/video-conference/service.ts
```diff
@@ -0,0 +1,238 @@
...
+
+	private async joinDirect(
+		call: IVideoConference,
+		user: AtLeast<IUser, '_id' | 'username' | 'name'>,
+		options: VideoConferenceJoinOptions,
+	): Promise<string> {
+		const url = this.getUrl(call, user, options);
+
+		if (!call.users.find(({ _id }) => _id === user._id)) {
+			this.addUserToCall(call._id, user);
```
sampaiodiego, line 207: should this await to make sure the user was added?
```suggestion
			await this.addUserToCall(call._id, user);
```

apps/meteor/server/services/video-conference/service.ts
```diff
@@ -0,0 +1,238 @@
...
+
+		// #ToDo: Load this from the Apps-Engine
+		return JitsiApp.customizeUrl(call, user, options);
+	}
+
+	private async addUserToCall(
+		callId: IVideoConference['_id'],
+		{ _id, username, name }: AtLeast<IUser, '_id' | 'username' | 'name'>,
+	): Promise<void> {
+		this.VideoConference.addUserById(callId, { _id, username, name });
```
sampaiodiego, line 236:
```suggestion
		await this.VideoConference.addUserById(callId, { _id, username, name });
```

apps/meteor/server/services/video-conference/service.ts
```diff
@@ -0,0 +1,238 @@
...
+		const user = await this.Users.findOneById(uid);
+		if (!user) {
+			throw new Error('failed-to-load-own-data');
+		}
+
+		if (call.messages.started) {
+			await this.changeMessageType(call.messages.started, 'video-direct-missed');
+		}
+
+		this.VideoConference.setEndedById(call._id, { _id: user._id, name: user.name, username: user.username });
```
sampaiodiego, line 127: it is not mandatory to `await` here, but I think we should, to make sure when this function returns, the data is correctly saved on DB
```suggestion
		await this.VideoConference.setEndedById(call._id, { _id: user._id, name: user.name, username: user.username });
```

apps/meteor/server/services/video-conference/service.ts
```diff
@@ -0,0 +1,238 @@
...
+	private Users: UsersRaw;
+
+	private Rooms: RoomsRaw;
+
+	private VideoConference: VideoConferenceRaw;
+
+	constructor(db: Db) {
+		super();
+
+		this.Users = new UsersRaw(db.collection('users'));
```
sampaiodiego, line 69: nothing wrong here, just something for us to think about, we've been using this pattern in a few places, I wonder if we can came up with a "better" solution to use models inside services.. 

I think just an `enum` of collection names would help.. this is one of my main concerns.

sampaiodiego (changes requested):

PR #25491, opened 2022-05-12
[IMPROVE] Avoid using omnichannel-queue collection

KevLehman:


murtaza98:
Rebasing the PR to run latest tests

apps/meteor/app/models/server/raw/LivechatInquiry.ts
```diff
@@ -34,4 +34,49 @@ export class LivechatInquiryRaw extends BaseRaw<ILivechatInquiryRecord> {
...
+	async findNextAndLock(queue?: string): Promise<ILivechatInquiryRecord | undefined> {
+		const date = new Date();
+		const result = await this.col.findOneAndUpdate(
+			{
+				status: LivechatInquiryStatus.QUEUED,
+				...(queue && { queue }),
+				$or: [
+					{
+						locked: true,
+						lockedAt: {
```
murtaza98, line 47: Since we're querying based on these fields, I think we should add them as indexes

apps/meteor/app/models/server/models/OmnichannelQueue.js
```diff
@@ -1,9 +0,0 @@
-import { Base } from './_Base';
-
-export class OmnichannelQueue extends Base {
-	constructor() {
-		super('omnichannel_queue');
```
murtaza98, line 5: Wouldn't this also need a migration to drop the old collection?

apps/meteor/app/models/server/models/LivechatInquiry.ts
```diff
@@ -12,17 +15,18 @@ export class LivechatInquiry extends Base {
 		this.tryEnsureIndex({ status: 1 }); // 'ready', 'queued', 'taken'
 		this.tryEnsureIndex({ queueOrder: 1, estimatedWaitingTimeQueue: 1, estimatedServiceTimeAt: 1 });
 		this.tryEnsureIndex({ 'v.token': 1, 'status': 1 }); // visitor token and status
+		this.tryEnsureIndex({ locked: 1, lockedAt: 1 }, { sparse: true }); // locked and lockedAt
```
murtaza98, line 18:
```suggestion
		this.tryEnsureIndex({ locked: 1, lockedAt: 1 }, { sparse: true }); // locked and lockedAt
		this.tryEnsureIndex({ queue: 1 }, { sparse: true }); // queue
```

Adding index on the queue prop as well since it is getting used in query

KevLehman: Where's that prop? 👀

murtaza98: image
Here it is

KevLehman: Well, that wasn't a new prop but a bug lol good catch there

murtaza98 (approved):
LGTM!

apps/meteor/server/startup/migrations/v269.ts
```diff
@@ -1,10 +1,13 @@
...
 import { addMigration } from '../../lib/migrations';
-import { Rooms } from '../../../app/models/server';
 
+// Remove Deprecated Omnichannel Queue Collection
 addMigration({
 	version: 269,
-	up() {
-		Rooms.tryDropIndex({ 'tokenpass.tokens.token': 1 });
-		Rooms.tryDropIndex({ tokenpass: 1 });
+	async up() {
```
sampaiodiego, line 8: oops.. looks like this migration was moved to 270.. this one can be reverted 😬

sampaiodiego (changes requested):

apps/meteor/server/startup/migrations/v269.ts
murtaza98: Revered the old migration :heavy_check_mark:

apps/meteor/server/startup/migrations/v273.ts
```diff
@@ -0,0 +1,13 @@
...
+
+import { addMigration } from '../../lib/migrations';
+
+// Remove Deprecated Omnichannel Queue Collection
+addMigration({
+	version: 273,
+	async up() {
+		// Remove collection
+		const { mongo } = MongoInternals.defaultRemoteCollectionDriver();
+		await mongo.db.dropCollection('rocketchat_omnichannel_queue');
```
sampaiodiego, line 11: if for some reason the collection doesn't exists, this will fail and halt the start up

sampaiodiego (changes requested):

apps/meteor/server/startup/migrations/v273.ts
KevLehman: What would be the right way of removing it then? 👀 
(This collection has been for quite a while, someone would have to update from a version older than 1.5 years for it to not be there [The col was there when i joined haha])

But it could happen if someone manually rolls back the migration lock number 🤔

sampaiodiego: maybe just wrap it into a try/catch :awesome-face:

KevLehman: image

murtaza98 (approved):

sampaiodiego (approved):

PR #25494, opened 2022-05-12
[FIX] Ordered and unordered list styles, Line breaks.

gabriellsh:

Also removed the message.md cache from server, since changes in the parser might break messages in the future (and will in this specific case).


apps/meteor/client/views/room/MessageList/components/EncryptedMessageRender.tsx
```diff
@@ -1,13 +1,13 @@
...
-import React, { ReactElement, useMemo } from 'react';
+import React, { ReactElement } from 'react';
 
 import MessageBodyRender from '../../../../components/Message/MessageBodyRender';
 import { useMessageActions } from '../../contexts/MessageContext';
+import { useParsedMessage } from '../hooks/useParsedMessage';
 
 const EncryptedMessageRender = ({ message }: { message: IMessage }): ReactElement => {
-	const tokens = useMemo(() => parser(message.msg), [message.msg]);
+	const tokens = useParsedMessage(message.msg);
```
ggazzo, line 10: if this message is encrypted there is a place were we decrypt, why not parse this value there?

ggazzo (changes requested):

apps/meteor/ee/server/services/package.json
```diff
@@ -26,7 +26,7 @@
 		"@rocket.chat/apps-engine": "^1.31.0",
 		"@rocket.chat/core-typings": "workspace:^",
 		"@rocket.chat/emitter": "~0.31.9",
-		"@rocket.chat/message-parser": "~0.31.9",
+		"@rocket.chat/message-parser": "0.31.12-dev.10",
```
tassoevan, line 29:
```suggestion
		"@rocket.chat/message-parser": "next",
```

tassoevan (changes requested):

ggazzo (approved):

PR #25495, opened 2022-05-12
WIP just load matrix if enabled

ggazzo:


geekgonecrazy (commented):
Seems like better solution would be to in the package try catch the require?

geekgonecrazy:
this will fix: https://github.com/RocketChat/Rocket.Chat/issues/25396 now with the failure handling being present

geekgonecrazy (approved):
Tested and it works.  Added a try catch to handle the failure that happens on alpine and older ubuntu installs.  It toggles off the feature when this failure happens and spits some verbose logs out

PR #25505, opened 2022-05-13
[NEW] Engagement Metrics - Phase 2

matheusbsilva137:

Add the following new statistics (metrics):
 - Total Broadcast rooms
 - Total rooms with an active Livestream;
 - Total triggered emails;
 - Total subscription roles;
 - Total User Roles;
 - Total uncaught exceptions;
 - `homeTitleChanged`: boolean value to indicate whether the `Layout_Home_Title` setting has been changed;
 - `homeBodyChanged`: boolean value to indicate whether the `Layout_Home_Body` setting has been changed;
 - `customCSSChanged`: boolean value to indicate whether the `theme-custom-css` setting has been changed;
 - `onLogoutCustomScriptChanged`: boolean value to indicate whether the `Custom_Script_On_Logout` setting has been changed;
 - `loggedOutCustomScriptChanged`: boolean value to indicate whether the `Custom_Script_Logged_Out` setting has been changed;
 - `loggedInCustomScriptChanged`: boolean value to indicate whether the `Custom_Script_Logged_In` setting has been changed;
 - `matrixBridgeEnabled`: boolean value to indicate whether the Matrix bridge has been enabled;


apps/meteor/server/models/raw/Roles.ts
```diff
@@ -164,6 +164,14 @@ export class RolesRaw extends BaseRaw<IRole> implements IRolesModel {
 		return this.find(query, options || {}) as P extends Pick<IRole, '_id'> ? Cursor<P> : Cursor<IRole>;
 	}
 
+	findByScope<P>(scope: IRole['scope'], options?: FindOneOptions<IRole>): P extends Pick<IRole, '_id'> ? Cursor<P> : Cursor<IRole> {
+		const query: FilterQuery<IRole> = {
+			scope,
+		};
+
+		return this.find(query, options || {}) as P extends Pick<IRole, '_id'> ? Cursor<P> : Cursor<IRole>;
+	}
```
sampaiodiego, line 173: I'm not sure why other methods were created like this, but I'd say you don't need this generic `P` and all this crazy, you could simply do:
```suggestion
	findByScope(scope: IRole['scope'], options?: FindOneOptions<IRole>): Cursor<IRole> {
		const query = {
			scope,
		};

		return this.find(query, options || {});
	}
```

apps/meteor/server/models/raw/Rooms.js
```diff
@@ -466,6 +466,24 @@ export class RoomsRaw extends BaseRaw {
...
+	findByBroadcast(options) {
+		return this.find(
+			{
+				broadcast: true,
+			},
+			options,
+		);
+	}
+
+	findByActiveLivestream(options) {
```
sampaiodiego, line 478: what does `Active` mean here? just asking =)

sampaiodiego (changes requested):

apps/meteor/server/models/raw/Rooms.js
matheusbsilva137: It means that the room announcement hasn't been cleared yet. When a livestream ends, all the content from the `streamingOptions` object is erased, so it's safe to use this field to check if there's an ongoing livestream.

tassoevan (approved):

ggazzo (approved):

PR #25515, opened 2022-05-15
[NEW] Ability for Rocket.Chat Apps to delete messages (Derived from #24806)

murtaza98:
## Further comments

Corresponding apps-engine PR: https://github.com/RocketChat/Rocket.Chat.Apps-engine/pull/507


debdutdeb:
I'm thinking whether this path should just be app deleting a message or any user deleting the message .. kinda like how we now send a message through an app - need to explicitely set the user performing the action, which can either be an user or the app user, set by the app itself.

murtaza98:
> I'm thinking whether this path should just be app deleting a message or any user deleting the message .. kinda like how we now send a message through an app - need to explicitely set the user performing the action, which can either be an user or the app user, set by the app itself.

That's a good point! Perhaps we can do the same thing here by adding a param asking for the user who is deleting the message. I'd like to keep that optional and use app-user if its not provided. WDYT?

murtaza98:
> I'm thinking whether this path should just be app deleting a message or any user deleting the message .. kinda like how we now send a message through an app - need to explicitely set the user performing the action, which can either be an user or the app user, set by the app itself.

That's a good point! Perhaps we can do the same thing here by adding a param asking for the user who is deleting the message. I'd like to keep that optional and use app-user if its not provided. WDYT?

PR #25520, opened 2022-05-16
[FIX] User abandonment setting is working again

cauefcr:

A setting watcher and the query for grabbing abandoned chats were broken, now they're not.


apps/meteor/ee/app/livechat-enterprise/server/lib/VisitorInactivityMonitor.js
```diff
@@ -98,7 +98,7 @@ export class VisitorInactivityMonitor {
 		if (!action || action === 'none') {
 			return;
 		}
-		LivechatRooms.findAbandonedOpenRooms(new Date()).forEach((room) => {
+		LivechatRooms.findAbandonedOpenRooms(new Date().toISOString()).forEach((room) => {
```
murtaza98, line 101:
```suggestion
		LivechatRooms.findAbandonedOpenRooms(new Date()).forEach((room) => {
```

I don't think this is required since you're converting to string over here & then in `.findAbandonedOpenRooms` method you're converting it back to Date object. I already tested this in my local without string conversion and it seems to be working. Can you please check this part again. I feel we only need to change the way we watch the settings over here

apps/meteor/ee/app/models/server/models/LivechatRooms.js
```diff
@@ -38,7 +38,7 @@ LivechatRooms.prototype.setPredictedVisitorAbandonment = function (roomId, willB
 
 LivechatRooms.prototype.findAbandonedOpenRooms = function (date) {
 	return this.find({
-		'omnichannel.predictedVisitorAbandonmentAt': { $lte: date },
+		'omnichannel.predictedVisitorAbandonmentAt': { $lte: new Date(date) },
```
murtaza98, line 41:
```suggestion
		'omnichannel.predictedVisitorAbandonmentAt': { $lte: date },
```

murtaza98: Same comment goes [here](https://github.com/RocketChat/Rocket.Chat/pull/25520/files#r874730462)

cauefcr: you were right

KevLehman (approved):

PR #25554, opened 2022-05-18
Chore: Convert apps/meteor/client/components/UserAutoComplete

juliajforesti:


apps/meteor/client/components/UserAutoComplete/UserAutoComplete.tsx
```diff
@@ -17,28 +30,29 @@ const UserAutoComplete = (props) => {
 		useMemo(() => query(debouncedFilter, conditions), [filter]),
 	);
 
-	const options = useMemo(() => (data && data.items.map((user) => ({ value: user.username, label: user.name }))) || [], [data]);
+	const options = useMemo(() => data?.items.map((user) => ({ value: user.username, label: user.name || '' })) || [], [data]);
```
ggazzo, line 33:
```suggestion
	const options = useMemo(() => data?.items.map((user) => ({ value: user.username, label: user.name || user.username })) || [], [data]);
```

PR #25566, opened 2022-05-19
Chore: converted bridges/converters to TS

felipe-rod123:

Converted `apps/meteor/app/apps/server/converters/` files to Typescript.


apps/meteor/app/models/server/raw/Uploads.ts
```diff
@@ -12,7 +12,7 @@ import {
 	WithId,
 	WriteOpResult,
 } from 'mongodb';
-import { IUpload as T } from '@rocket.chat/core-typings';
+import { IUpload, IUpload as T } from '@rocket.chat/core-typings';
```
graywolf336, line 15: What's the reason for this?

Issue #20595, opened 2021-02-04
[BUG] Global search not turned off by default


RonLek:
### Description:

When global search is turned on the default behaviour **is** of global search which should **not** be the case as addressed [here](https://github.com/RocketChat/Rocket.Chat/pull/19777)

### Steps to reproduce:

1. Enable Global Search from Administration.
2. Click on search icon in a channel.
3. Search for any string without checking the global search checkbox.

### Expected behavior:

The search shouldn't be global since the global search checkbox is checked off initially.

### Actual behavior:

The search is a global search.

https://user-images.githubusercontent.com/28918901/106859375-f2fc5100-66e8-11eb-9889-d1041c007dc2.mp4

### Server Setup Information:

- Version of Rocket.Chat Server: 3.12.0-develop
- Operating System: Ubuntu 18.04 LTS
- DB Replicaset Oplog: Enabled
- NodeJS Version: 12.18.4
- MongoDB Version: 4.2.8

### Client Setup Information

- Desktop App or Browser Version: Chromium
- Operating System: Ubuntu 18.04 LTS

ankar84:
Confirm that issue

PR #25568, opened 2022-05-19
[FIX] Set default value "false" for global search.

ivan1kazantsev:

If we disable the "global search", then the "global search" checkbox removes from the UI. It's not what we want. I think, we should set the default to "false", because the default is to search inside the channel/dialog.

## Issue(s)

Fixes: [#20595](https://github.com/RocketChat/Rocket.Chat/issues/20595)
## Steps to test or reproduce

1. Enable the "global search" in admin settings.
2. Go to channel or dialog.
3. Use search and be happy.


katop (dismissed):

debdutdeb (approved):

wreiske:
I just did this in my local build because this has been so incredibly painful for us... Went to go push a PR and found this one. Can we get this merged, please?

debdutdeb (approved):

vinogradovia:
#21659

PR #25579, opened 2022-05-20
Chore: Typescript migration for couple of endpoints in livechat/server/api/v1

murtaza98:


apps/meteor/app/livechat/server/lib/Contacts.ts
```diff
@@ -1,16 +1,28 @@
...
+		customFields = {},
+		contactManager,
+	}: Pick<ILivechatVisitorDTO, 'token' | 'name' | 'email' | 'username' | 'phone'> & {
+		customFields?: Record<string, unknown>;
+		contactManager?: ILivechatVisitor['contactManager'];
+	}): string {
 		check(token, String);
 
-		const visitorEmail = s.trim(email).toLowerCase();
+		const visitorEmail = email && s.trim(email).toLowerCase();
```
KevLehman, line 22: Can we just use `email.trim()` instead of the underscore one?

apps/meteor/app/livechat/server/lib/Contacts.ts
```diff
@@ -1,16 +1,28 @@
...
 
 		let contactId;
-		const updateUser = {
+		const updateUser: any = {
```
KevLehman, line 25: 😬

KevLehman: I believe you can use one of the mongo types (like `UpdateQuery` for representing this (and avoid `any`)

apps/meteor/app/livechat/server/lib/Contacts.ts
```diff
@@ -43,12 +55,14 @@ export const Contacts = {
 		updateUser.$set.phone = (phone && [{ phoneNumber: phone }]) || null;
 		updateUser.$set.visitorEmails = (visitorEmail && [{ address: visitorEmail }]) || null;
 
-		const allowedCF = LivechatCustomField.find({ scope: 'visitor' }, { fields: { _id: 1 } }).map(({ _id }) => _id);
+		const allowedCF = (LivechatCustomField.find({ scope: 'visitor' }, { fields: { _id: 1 } }) as ILivechatCustomField[]).map(
```
KevLehman, line 58: Does this fit in one line yet? 🙈

apps/meteor/app/livechat/server/lib/Contacts.ts
```diff
@@ -43,12 +55,14 @@ export const Contacts = {
...
+			({ _id }) => _id,
+		);
 
 		const livechatData = Object.keys(customFields)
 			.filter((key) => allowedCF.includes(key) && customFields[key] !== '' && customFields[key] !== undefined)
 			.reduce((obj, key) => {
-				obj[key] = customFields[key];
+				(obj as any)[key] = customFields[key];
```
KevLehman, line 65: I always wanted to try this: https://fettblog.eu/typescript-better-object-keys/ to have a typed Object.keys 👀 so we don't have to typecast as any. maybe you can give it a try 😬

KevLehman (changes requested):

apps/meteor/app/livechat/server/lib/Contacts.ts
murtaza98: Interesting!! Thanks for sharing :hugs: Was able to apply this here :smile:

apps/meteor/app/livechat/server/lib/Contacts.ts
murtaza98: Yups, it seems like it does. Even if I split it into multiple lines, the prettier auto-format just reverts it back to this :bug:

PR #25586, opened 2022-05-20
Chore: Convert additionalForms

ggazzo:


MartinSchoeler (approved):
None of the forms seems to be broken, LGTM

PR #25601, opened 2022-05-23
Chore: add Ajv JSON Schema to api/v1

felipe-rod123:

This pull request adds Ajv JSON Schema validation to `apps/meteor/app/api/server/v1/` and `packages/rest-typings/src/v1/`, where needed.


ggazzo (changes requested):
put some `Prop` at the end of types and functions assertions

pierre-lehnen-rc (changes requested):

.yarnrc.yml
```diff
@@ -11,3 +11,5 @@ plugins:
     spec: '@yarnpkg/plugin-typescript'
 
 yarnPath: .yarn/releases/yarn-3.2.0.cjs
+
+checksumBehavior: 'update'
```
pierre-lehnen-rc, line 15:
```suggestion
```

Leave this out of this PR.

packages/rest-typings/src/v1/omnichannel.ts
```diff
@@ -13,20 +13,778 @@ import type {
...
+			items: {
+				type: 'string',
+			},
+		},
+	},
+	required: ['upsert', 'remove'],
+	additionalProperties: false,
+};
+
+export const isLivechatDepartmentDepartmentIdAgentsPOSTProps = ajv.compile<LivechatDepartmentDepartmentIdAgentsPOST>(
```
KevLehman, line 148: This kind of types would become more common, would be good to define an "standard" for it, like `GETLivechatDepartmentIdAgents` and `POSTLivechatDepartmentIdAgents`, prefixing the method to all typedefs

packages/rest-typings/src/v1/omnichannel.ts
```diff
@@ -13,20 +13,778 @@ import type {
...
+		sort: {
+			type: 'string',
+			nullable: true,
+		},
+		query: {
+			type: 'string',
+			nullable: true,
+		},
+	},
+	required: [],
```
KevLehman, line 774: I believe required can be omitted when no prop is required

packages/rest-typings/src/v1/users.ts
```diff
@@ -1,26 +1,140 @@
 import type { ITeam, IUser } from '@rocket.chat/core-typings';
+import Ajv from 'ajv';
+
+const ajv = new Ajv({
```
KevLehman, line 4: Looking at the times we're using this, would be good to have it centralized, in case we ever want to add/change a config option 🙈 

(Not sure if on the scope of this PR, but)

ggazzo: yep I'm planning to centralize them

apps/meteor/client/components/Omnichannel/hooks/useDepartmentsList.ts
```diff
@@ -44,7 +43,7 @@ export const useDepartmentsList = (
 				count: end + start,
 				sort: `{ "name": 1 }`,
 				excludeDepartmentId: options.excludeDepartmentId,
-				enabled: options.enabled,
+				enabled: options.enabled ? 'true' : 'false',
```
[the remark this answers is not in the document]
ggazzo: makes sense but I prefer being a little more verbose here, and maybe if somebody uses this function in a javascript code and pass a wrong value it will lead in to a problem

pierre-lehnen-rc (approved):

PR #25604, opened 2022-05-23
[FIX] useCurrentChatTags is not a function

MartinSchoeler:
This should fix the error `useCurrentChatTags is not a function` when trying to close a livechat conversation


KevLehman (dismissed):

tiagoevanp (dismissed):
Don't forget to fix the hook! For now... is what we can do...

PR #21779, opened 2021-04-23
[FIX] Remove margin from quote inside quote

tiagoevanp:
## Proposed changes (including videos or screenshots)

image

PR #25613, opened 2022-05-23
[FIX] Quote message spacing

hugocostadev:
Fix Quote message margins

Was applied default margins for the quote message.

The change was applied only in the single quote, when it has more than one quote was applied only in the inner quote to prevent overwriting this [PR](https://github.com/RocketChat/Rocket.Chat/pull/21779).

## Issue(s)

The quote margins is off, should be 16px in all sides

## Steps to test or reproduce

1. Log in to the [rocket.chat](http://rocket.chat/) 
2. Open a room (DM/Discussion/Channel/Team)
3. Send a message
4. Quote the previous message


apps/meteor/client/components/Message/Attachments/QuoteAttachment.tsx
```diff
@@ -18,6 +18,11 @@ const hover = css`
 			border-inline-start-color: ${colors.n600} !important;
 		}
 	}
+
+	blockquote.rcx-attachment__details {
+		blockquote.rcx-attachment__details {
+			margin-block-end: -12px
+	}
```
tassoevan, line 25: Why not pass directly to `<Attachment.Details ... />`? Even better, why not move it into `Attachment.Details`? It'd avoid a higher-specificity CSS selector, and it could be more succint:

```
& > & {
  margin-block-end: -12px;
}
```

tassoevan (changes requested):

apps/meteor/client/components/Message/Attachments/QuoteAttachment.tsx
hugocostadev: Because not all `<Attachment.Details ... />` will be used in Quote (or a blockquote), if we move to Details all the Details for other Attachments (files, images, pinned messages and etc...) will be impacted, and is possible to include a normal blockquote with Markdown inside `<Attachment.Details ... />`
 
I decided to put at `QuoteAttachment` because this is a component only used for Quote messages and the margin only should be applied at the blockquote within another blockquote.

tassoevan (approved):

PR #25614, opened 2022-05-23
[FIX] Thread Message Preview parser

gabriellsh:


hugocostadev:
Hi @gabriellsh , I found some problems, check below:

Feedbacks:
1. Sometimes the "Send to channel" checkbox does not work, which happened with the user role only
2. Sometimes without the checkbox checked it still send to channel
image

3. The big Emoji seems to be smaller than the normal emoji
image

4. The list index are mismatching in thread preview and thread box
image

5. Some inline code items aren't showing the border bottom
image

6. The Heading markdown are not being parsed correctly at Thread Box (I think that this is expected since it's using the old parser, right? )
But I don't know, because I tested in the open.rocket.chat workspace and seems different
image

PR #25627, opened 2022-05-24
Regression: Subscription menu not appearing for non installed but subscribed apps

rique223:

Fixed a problem on which the AppMenu component did not appear for apps that had an active subscription but weren't installed, now the rendering of the component is also based on the isSubscribed flag, and the appearance of the uninstall and enable/disable options are based on the app.installed flag so that the correct options appear on all the edge cases.
Demo gif:
subscription-manager-fix

### DIFF
--- a/packages/core-typings/src/Apps.ts
+++ b/packages/core-typings/src/Apps.ts
@@ -49,6 +49,7 @@ export type App = {
 	iconFileContent: string;
 	installed?: boolean;
 	isEnterpriseOnly?: boolean;
+	isSubscribed: string;
 	bundledIn: {
 		bundleId: string;
 		bundleName: string;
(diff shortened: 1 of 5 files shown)


packages/core-typings/src/Apps.ts
```diff
@@ -49,6 +49,7 @@ export type App = {
 	iconFileContent: string;
 	installed?: boolean;
 	isEnterpriseOnly?: boolean;
+	isSubscribed: string;
```
graywolf336, line 52: This should be a boolean...

rique223: Oh lord, that was totally an oversight. Pushing a fix commit right now.

rique223: Solved.

geekgonecrazy (approved):

ggazzo (approved):

PR #25631, opened 2022-05-25
[FIX] Messages spacing

hugocostadev:

Adding `sequential` prop to Message component from Fuselage

## Issue(s)

The Message component was not sending the `sequential` prop to Fuselage component that is trying to access it, because of that the className was never generated.


gabriellsh (changes requested):
Hey @hugocostadev! I noticed that we might have missed a few more margins when calculating the message size. Theres an additional margin of 2px in a child element we didn't account for, so we need to make another adjustment in fuselage and change the 8px margin to be 6px.

gabriellsh (approved):

PR #25643, opened 2022-05-26
[IMPROVE] Differ Voip calls from Incoming and Outgoing

murtaza98:

Updated this column and its respective endpoints to support inbound/outfound call definitions
image

## Further comments

Clickup: https://app.clickup.com/t/22bmc0f


KevLehman:
The pipe fails, and seems like an actual fail 👀 can you take a look?

apps/meteor/app/api/server/v1/voip/rooms.ts
```diff
@@ -89,10 +89,11 @@ API.v1.addRoute(
 				token: String,
 				agentId: Match.Maybe(String),
 				rid: Match.Maybe(String),
+				direction: Match.Maybe(String),
```
KevLehman, line 92: You can pass a custom validator here to do what you're doing on the if at 119

KevLehman (commented):
Pls check if voip/rooms endpoint can be filtered by the new prop too :)

apps/meteor/client/views/omnichannel/directory/calls/CallTable.tsx
```diff
@@ -54,6 +55,21 @@ const CallTable: FC = () => {
 	const query = useQuery(debouncedParams, debouncedSort, userIdLoggedIn);
 	const directoryRoute = useRoute('omnichannel-directory');
 
+	const resolveDirectionLabel = useCallback(
+		(direction: Partial<IVoipRoom['direction']>) => {
```
tiagoevanp, line 59: Partial is not necessary on this case, right? There is a specific portion of IVoipRoom defined as a type here (direction)... Partial is used when you don't know from what portion of the type must be used. I mean... It looks like an 'any' inside of a specific type defined!

tiagoevanp (changes requested):

apps/meteor/client/views/omnichannel/directory/calls/CallTable.tsx
murtaza98: Yeah, it makes sense, but I don't want to keep it any since it can either be either (`inbound` | `outbound` | null). So I feel `Nullable` makes a bit more sense here. WDYT?

apps/meteor/app/api/server/v1/voip/rooms.ts
murtaza98: Sure, its done :heavy_check_mark:

murtaza98:
> Pls check if voip/rooms endpoint can be filtered by the new prop too :)

Alright! Added a new parameter there too, allowing filter by call direction

apps/meteor/client/views/omnichannel/directory/calls/CallTable.tsx
tiagoevanp: Why direction can be a null value? 🤔

tiagoevanp: Nullable type definition doesn't exist on TS... 🤔

murtaza98: I was trying to add a fallback kind of thing here, hence I'd earlier added this extra typing to make the prop nullable. However, I now realise that its not required so I've removed it.

KevLehman: Remember that (maybe not for this case :) but for others) you can create Utility types for your use cases, for example:
`Nullable<T> = T | null` and then use `Nullable<IVoipRoom['direction']>`

apps/meteor/app/api/server/v1/voip/rooms.ts
```diff
@@ -89,10 +90,11 @@ API.v1.addRoute(
 				token: String,
 				agentId: Match.Maybe(String),
 				rid: Match.Maybe(String),
+				direction: VoipRoomDirectionValidator,
```
KevLehman, line 93: What if instead of having it as optional, you move the check to inside the `!rid` if? 👀 That way you don't have to check for the prop to not to be optional

KevLehman (approved):

KevLehman:
Conflicts 😬

murtaza98:
> Conflicts grimacing

Fixed it :+1:

KevLehman:
New conflicts 😬

murtaza98:
> New conflicts grimacing

Haha... lots of changes happening here which is great!! Fixed the conflicts btw

murtaza98:
Finally, the builds passed!! Had to restart them about 4 times :see_no_evil:

KevLehman (commented):
There's some changes to RDStation that I removed 👀 not sure if they should be there. If not, pls remove those changes

apps/meteor/server/startup/migrations/v269.ts
```diff
@@ -0,0 +1,20 @@
+import { VoipRoom } from '../../../app/models/server/raw';
+import { addMigration } from '../../lib/migrations';
+
+// Removes deprecated RDStation functionality from Omnichannel
```
murtaza98, line 4:
```suggestion
```

apps/meteor/server/startup/migrations/v269.ts
```diff
@@ -0,0 +1,20 @@
+import { VoipRoom } from '../../../app/models/server/raw';
```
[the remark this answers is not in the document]
murtaza98: I think you're referring to the comment which was carried from before migration here. If so, I've removed it now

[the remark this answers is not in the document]
KevLehman: Yups, the comment should be on 269. Not sure how it got here tho :eyes: if the migration numbers are different

KevLehman (dismissed):
Function wise, looks good. @tiagoevanp pls check FE changes here. I'll move the task to QA for now, but feel free to change status if you find something during your review

KevLehman (dismissed):

KevLehman:
This task won't be tested as the changes are not perceptible through UI or endpoints yet

murtaza98:
Hi @tiagoevanp @RocketChat/chat-engine Just a friendly reminder about the review of this PR :smile:

tiagoevanp (approved):

tiagoevanp (changes requested):
fix Alek suggestions!

apps/meteor/client/views/omnichannel/directory/calls/CallTable.tsx
```diff
@@ -54,6 +55,21 @@ const CallTable: FC = () => {
 	const query = useQuery(debouncedParams, debouncedSort, userIdLoggedIn);
 	const directoryRoute = useRoute('omnichannel-directory');
 
+	const resolveDirectionLabel = useCallback(
+		(direction: IVoipRoom['direction']) => {
+			if (direction === 'inbound') {
```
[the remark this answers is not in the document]
murtaza98: Now that looks much better :heart_eyes:  Thanks for the suggestion!!

aleksandernsilva (dismissed):

apps/meteor/client/views/omnichannel/directory/calls/CallTable.tsx
```diff
@@ -54,6 +55,17 @@ const CallTable: FC = () => {
 	const query = useQuery(debouncedParams, debouncedSort, userIdLoggedIn);
 	const directoryRoute = useRoute('omnichannel-directory');
 
+	const resolveDirectionLabel = useCallback(
+		(direction: IVoipRoom['direction']) => {
+			const labels: { [key: string]: 'Incoming' | 'Outgoing' } = {
```
aleksandernsilva, line 60: Actually, here you could use `as const`. Ex: 
```
const label = {
 ...
} as const
```

murtaza98: That's a really good suggestion. Thanks!!

aleksandernsilva (approved):

apps/meteor/server/startup/migrations/v270.ts
```diff
@@ -0,0 +1,21 @@
...
+
+import { addMigration } from '../../lib/migrations';
+
+addMigration({
+	version: 270,
+	async up() {
+		// mark all voip rooms as inbound which doesn't have any direction property set or has an invalid value
+		await VoipRoom.updateMany(
+			{
+				t: 'v',
```
murtaza98, line 11: I realised today that this migration query was not very specific, which was resulting in adding the direction prop to all rooms (even non-omnichannel related). This new query should fix that issue

KevLehman (dismissed):

ggazzo:
I can't see why we are skipping QA for this task

KevLehman:
> I can't see why we are skipping QA for this task

This can't be tested as it's BE mostly, and will be tested in another task

KevLehman (approved):

sampaiodiego (approved):

tiagoevanp (approved):

PR #25651, opened 2022-05-26
Regression: VoIp wrap up modal not opening after call disconnect

aleksandernsilva:

This PR fixes a bug preventing the wrap up call modal from being displayed after caller or agent ends the call.


tiagoevanp (approved):
LGTM

PR #25652, opened 2022-05-27
[NEW] [VideoConference] VideoConf Call Sidebar

dougfabris:

Kapture 2022-05-26 at 22 04 26


apps/meteor/client/providers/VideoConfPopupProvider.tsx
```diff
@@ -1,37 +1,74 @@
...
+				return (): void => {
+					customSound.pause('calling');
+					clearInterval(soundInterval);
+				};
+			}
+		}
+
+		return setPopUps([]);
+	}, [incomingCalls, contextValue, customSound, isRinging]);
+
```
ggazzo, line 55: I think this could be an hook somehow , and not sure if `incomingCalls` is mutable or not, this could be a potencial bug.

I didnt like the `return setPopUps([]);` part as well

ggazzo (changes requested):

apps/meteor/client/providers/VideoConfPopupProvider.tsx
dougfabris: solved in peer review!

ggazzo (approved):

PR #25658, opened 2022-05-27
Chore: Converting files from app/livechat folder from JS to TS

amolghode1981:

Converting files from apps/meteor/app/livechat/lib/ from JS to TS


apps/meteor/app/ui-utils/lib/MessageTypes.ts
```diff
@@ -9,7 +9,7 @@ export type MessageType = {
 	/* deprecated */
 	template?: (message: IMessage) => unknown;
 	message: TranslationKey;
-	data?: (message: IMessage) => Record<string, string>;
+	data?: (message: IMessage) => Record<string, string> | undefined;
```
murtaza98, line 12: I don't think the data() method is equipped to deal with an undefined value. I'd recommend instead returning an empty string corresponding to the parameter

apps/meteor/app/livechat/lib/messageTypes.ts
```diff
@@ -122,7 +126,10 @@ MessageTypes.registerType({
 	id: 'omnichannel_on_hold_chat_resumed',
 	system: true,
 	message: 'Omnichannel_on_hold_chat_resumed',
-	data(message) {
+	data(message: IOmnichannelSystemMessage) {
+		if (!message.comment) {
+			return;
```
murtaza98, line 131:
```suggestion
			return {  };
```
The translation string expects an object with a comment prop from here, so I'd not recommend returning undefined values. You may either return a empty blank string (like `{ comment: '' }`) or you can even add a custom message there (like `{ comment: 'No comment provided' }`)

apps/meteor/app/livechat/lib/messageTypes.ts
```diff
@@ -105,13 +105,17 @@ MessageTypes.registerType({
...
 });
 
 MessageTypes.registerType({
 	id: 'omnichannel_placed_chat_on_hold',
 	system: true,
 	message: 'Omnichannel_placed_chat_on_hold',
-	data(message) {
+	data(message: IOmnichannelSystemMessage) {
+		if (!message.comment) {
+			return;
```
murtaza98, line 117: same here

apps/meteor/app/ui-utils/lib/MessageTypes.ts
amolghode1981: Fixed.

apps/meteor/app/livechat/lib/messageTypes.ts
amolghode1981: Fixed.

apps/meteor/app/livechat/lib/messageTypes.ts
amolghode1981: Fixed.

murtaza98 (approved):
LGTM!

Issue #25460, opened 2022-05-10
Web access connection problems since upgrade 4.6.3 -> 4.7.0


Sh0uld:
### Description:

Web access is not available due to connection problems, but the Rocket.Chat App launches and is usable through Android apps (cannot test it with iOS). The web parts of the android apps (e.g. Admin Panel) doesn't load either.
The error occurred after the upgrade from 4.6.3 to 4.7.0 with the docker image.

### Steps to reproduce:

1. Have a running docker instance with Rocket.Chat version 4.6.3
2. Upgrade to Rocket.Chat version 4.7.0
3. Admire the connection/loading errors in a browser ;-)

### Expected behavior:

Rocket.Chat should greet me with a functional login screen.

### Actual behavior:

The login screen doesn't load completely, and a connection banner on the top of screen appears (see screenshot). This behavior occurs in Edge, Chrome and Firefox (no other browsers tested).

![Screenshot login screen](https://user-images.githubusercontent.com/6522458/167709091-2e71fae0-434b-4ed8-970f-18ed5f0bbaaf.png)

### Server Setup Information:

- Version of Rocket.Chat Server: 4.7.0
- Operating System: Ubuntu 20.04 LTS
- Deployment Method: docker
- Number of Running Instances: 1
- DB Replicaset Oplog: Enabled
- NodeJS Version: 14.18.3 - x64
- MongoDB Version: 5.0.8

### Client Setup Information

- Browser Version: Edge, Firefox and Chrome in current version.
- Operating System: Ubuntu 22.04 / Windows 10

### Additional context

I tried to roll back to Rocket.Chat version 4.6.3 (docker), but this version doesn't start anymore due to "migration 261 not found".

### Relevant logs:

Rocket.Chat Startup-Log:
```
(node:1) UnhandledPromiseRejectionWarning: MongoError: An equivalent index already exists with the same name but different options. Requested index: { v: 2, key: { _updatedAt: 1 }, name: "_updatedAt_1", expireAfterSeconds: 2592000 }, existing index: { v: 2, key: { _updatedAt: 1 }, name: "_updatedAt_1", ns: "parties.rocketchat_integration_history" }
    at MessageStream.messageHandler (/app/bundle/programs/server/npm/node_modules/meteor/npm-mongo/node_modules/mongodb/lib/cmap/connection.js:272:20)
    at MessageStream.emit (events.js:400:28)
    at MessageStream.emit (domain.js:475:12)
[...]

Issue #25513, opened 2022-05-14
4.7.0 and 4.7.1 - web clients do not work, mobile clients work


fc757:
NodeJS: 14.18.3
Mongodb: 5.0.8

When building 4.7.0 or 4.7.1 get this:

---

`linux-x64-83-glibc` exists; testing
Binary is fine; exiting

> gc-stats@1.4.0 install /tmp/bundle/programs/server/npm/node_modules/gc-stats
> node-pre-gyp install --fallback-to-build

node-pre-gyp ERR! install response status 403 Forbidden on https://node-binaries.s3.amazonaws.com/gcstats/v1.4.0/Release/node-v83-linux-x64.tar.gz
node-pre-gyp WARN Pre-built binaries not installable for gc-stats@1.4.0 and node@14.18.3 (node-v83 ABI, glibc) (falling back to source compile with node-gyp)
node-pre-gyp WARN Hit error response status 403 Forbidden on https://node-binaries.s3.amazonaws.com/gcstats/v1.4.0/Release/node-v83-linux-x64.tar.gz

---

When starting get this:

---

May 14 09:39:13 cetialphafive systemd[1]: Started The Rocket.Chat server.
May 14 09:39:18 cetialphafive node[226292]: (node:226292) UnhandledPromiseRejectionWarning: MongoError: An equivalent index already exists with the same name but different options. Requested index: { v: 2, key: { _updatedAt: 1 }, name: "_updatedAt_1", expireAfterSeconds: 2592000 }, existing index: { v: 2, key: { _updatedAt: 1 }, name: "_updatedAt_1" }
May 14 09:39:18 cetialphafive node[226292]:     at MessageStream.messageHandler (/opt/Rocket.Chat/bundle/programs/server/npm/node_modules/meteor/npm-mongo/node_modules/mongodb/lib/cmap/connection.js:272:20)
May 14 09:39:18 cetialphafive node[226292]:     at MessageStream.emit (events.js:400:28)
May 14 09:39:18 cetialphafive node[226292]:     at MessageStream.emit (domain.js:475:12)
May 14 09:39:18 cetialphafive node[226292]:     at processIncomingData (/opt/Rocket.Chat/bundle/programs/server/npm/node_modules/meteor/npm-mongo/node_modules/mongodb/lib/cmap/message_stream.js:144:12)
[...]

PR #25669, opened 2022-05-27
[FIX] Bumb metoer-node-stubs to version 1.2.3 to fix issue #25460

Sh0uld:

With meteor-node-stubs version 1.2.3 a bug was fixed, which occured in issue #25460 and probably #25513 (last one not tested).
For the issue in meteor see: https://github.com/meteor/meteor/issues/11974

## Issue(s)
#25460 

## Steps to test or reproduce

The bug can be reproduced by importing a specific mongodb which I can provide if needed.

## Further comments

I've checked some differences between a working mongodb and my (productive) mongodb, which is affected by #25460, but wasn't able to find the setting (or whatever) which causes the bug.

### DIFF
--- a/apps/meteor/package.json
+++ b/apps/meteor/package.json
@@ -293,7 +293,7 @@
 		"mailparser": "^3.4.0",
 		"marked": "^0.7.0",
 		"mem": "^8.1.1",
-		"meteor-node-stubs": "^1.2.1",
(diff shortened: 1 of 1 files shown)


ggazzo:
please @Sh0uld commit your yarn.lock

ggazzo (approved):

Issue #25456, opened 2022-05-10
4.7.0 regression | Files consisting of unicode characters are saved with a broken name.


IgorOhrimenko:
Files consisting of unicode characters are saved with a broken name.
## Reproduce issue
1. docker-compose up --detach
 <details>
 <summary>compose.yaml</summary>

```yaml
services:
  rocketchat:
    image: registry.rocket.chat/rocketchat/rocket.chat:4.7.0
    command: >
      bash -c
        "for i in `seq 1 30`; do
          node main.js &&
          s=$$? && break || s=$$?;
          echo \"Tried $$i times. Waiting 5 secs...\";
          sleep 5;
        done; (exit $$s)"
    restart: unless-stopped
    environment:
      - MONGO_URL=mongodb://mongo:27017/rocketchat
      - MONGO_OPLOG_URL=mongodb://mongo:27017/local
    depends_on:
      - mongo
    ports:
      - 3000:3000

  mongo:
    image: mongo:5.0
    restart: unless-stopped
    volumes:
      - ./data/db:/data/db
      - ./data/dump:/dump
    command: docker-entrypoint.sh mongod --oplogSize 128 --replSet rs0 --storageEngine=wiredTiger

  mongo-init-replica:
    image: mongo:5.0
    command: >
      bash -c
        "for i in `seq 1 30`; do
          mongo mongo/rocketchat --eval \"
            rs.initiate({
              _id: 'rs0',
              members: [ { _id: 0, host: 'localhost:27017' } ]})\" &&
          s=$$? && break || s=$$?;
          echo \"Tried $$i times. Waiting 5 secs...\";
          sleep 5;
        done; (exit $$s)"
    depends_on:
      - mongo
```
</details>

2. Try uplouad file with UTF characters any UTF caracter from https://www.w3.org/2001/06/utf-8-test/UTF-8-demo.html, example `Σὲ γνωρίζω ἀπὸ τὴν κόψη`
3. See broken filename
3.1. error on console
<details>

```
rocketchat_1          | FileStream {
rocketchat_1          |   _readableState: ReadableState {
rocketchat_1          |     objectMode: false,
rocketchat_1          |     highWaterMark: 16384,
rocketchat_1          |     buffer: BufferList { head: null, tail: null, length: 0 },
rocketchat_1          |     length: 0,
rocketchat_1          |     pipes: [],
rocketchat_1          |     flowing: null,
rocketchat_1          |     ended: false,
[...]

PR #25673, opened 2022-05-28
[FIX] Change form body parameter charset to UTF-8 to fix issue #25456

divinespear:

since [mscdex/busboy](https://github.com/mscdex/busboy) 1.5.0, new option named `defParamCharset` for form body parameter encoding is added with default value `latin1`, so unicode filenames are broken since 4.7.0.

Screenshot from 2022-05-28 16-26-06

## Issue(s)

#25456 

## Steps to test or reproduce

Just upload any files that has non-ascii filename.
(You can get some UTF-8 characters from https://www.w3.org/2001/06/utf-8-test/UTF-8-demo.html)


ggazzo (approved):

ankar84:
Please add that fix to 4.8

PR #25689, opened 2022-05-30
Regression: App event listeners broke Slackbridge integration and importers

d-gubert:

Some event listeners triggered by Apps were calling `Meteor.user()` in functions that could run outside of Meteor environment


apps/meteor/app/lib/server/functions/saveUser.js
```diff
@@ -420,7 +421,7 @@ export const saveUser = function (userId, userData) {
 		Apps.triggerEvent(AppEvents.IPostUserUpdated, {
 			user: userUpdated,
 			previousUser: oldUserData,
-			performedBy: Meteor.user(),
+			performedBy: safeGetMeteorUser(),
```
ggazzo, line 424: why not `Users.findOneById(userData._id)` ?

ggazzo (approved):

PR #25690, opened 2022-05-30
Chore: Translate admin helpers to TS

rique223:
Translated the admin helpers.js file to typescript, changed some types on the orchestrator to better use already existing types and changed the heap size of the typecheck script for it to stop overflowing.


packages/core-typings/src/Apps.ts
```diff
@@ -27,6 +27,27 @@ export type Tiers = {
...
+	seats: number;
+	maxSeats: number;
+	license: License;
+	startDate: string;
+	periodEnd: string;
+	endDate: string;
+	externallyManaged: boolean;
+	isSubscribedViaBundle: boolean;
+};
+
```
ggazzo, line 50: what do you think about keep as prefix like AppLicense and AppSubscriptionInfo ?

ggazzo (changes requested):

matheuslc (approved):
LGTM 🚀

packages/core-typings/src/Apps.ts
```diff
@@ -45,10 +66,11 @@ export type App = {
 	version: string;
 	price: number;
 	purchaseType: string;
-	pricingPlans: PricingPlan[];
+	pricingPlans: AppPricingPlan[];
 	iconFileContent: string;
 	installed?: boolean;
 	isEnterpriseOnly?: boolean;
+	isPurchased?: string;
```
graywolf336, line 73: Why is this a string?

packages/core-typings/src/Apps.ts
```diff
@@ -10,21 +10,42 @@ export type AppScreenshot = {
...
-	price: number;
+export type AppLicense = {
+	license: string;
+	version: number;
+	expireDate: string;
+};
+
+export type AppSubscriptionInfo = {
+	typeOf: string;
+	status: string;
```
graywolf336, line 38: This is an enum:
```
export enum PurchaseSubscriptionStatus {
    Trialing = 'trialing',
    Active = 'active',
    Cancelled = 'cancelled',
    Cancelling = 'cancelling',
    PastDue = 'pastDue',
}
```

packages/core-typings/src/Apps.ts
```diff
@@ -10,21 +10,42 @@ export type AppScreenshot = {
...
+	statusFromBilling: boolean;
+	isSeatBased: boolean;
+	seats: number;
+	maxSeats: number;
+	license: AppLicense;
+	startDate: string;
+	periodEnd: string;
+	endDate: string;
+	externallyManaged: boolean;
```
graywolf336, line 47:
```suggestion
```
This property will not be exposed in the near future.

packages/core-typings/src/Apps.ts
```diff
@@ -45,10 +66,11 @@ export type App = {
 	version: string;
```
graywolf336, line 66:
```
detailedDescription: IDetailedDescription;
detailedChangelog: IDetailedChangelog;
```
These are missing, there were defined above and got lost.

packages/core-typings/src/Apps.ts
```diff
@@ -59,11 +81,13 @@ export type App = {
 	marketplaceVersion: string;
 	latest: App;
 	status?: AppStatus;
+	subscriptionInfo: AppSubscriptionInfo;
 	licenseValidation?: {
 		errors: { [key: string]: string };
 		warnings: { [key: string]: string };
 	};
 	marketplace: unknown;
```
graywolf336, line 89: What is this property?

packages/core-typings/src/Apps.ts
```diff
@@ -10,21 +10,42 @@ export type AppScreenshot = {
...
+	maxSeats: number;
+	license: AppLicense;
+	startDate: string;
+	periodEnd: string;
+	endDate: string;
+	externallyManaged: boolean;
+	isSubscribedViaBundle: boolean;
 };
 
 export type App = {
```
graywolf336, line 51: Is this an AppOverview or AppInfo? AppOverview contains information about the app and the AppInfo contains information about the version of an app. This seems to be a mixture of everything and it honestly worries me.

packages/core-typings/src/Apps.ts
```diff
@@ -45,10 +66,11 @@ export type App = {
...
 	price: number;
 	purchaseType: string;
-	pricingPlans: PricingPlan[];
+	pricingPlans: AppPricingPlan[];
 	iconFileContent: string;
 	installed?: boolean;
 	isEnterpriseOnly?: boolean;
+	isPurchased?: string;
 	isSubscribed: boolean;
 	bundledIn: {
```
graywolf336, line 75: Also missing `bundledAppIds: string[]` if this is an AppOverview. If this is an AppInfo, then this property won't exist.

graywolf336 (dismissed):
I have a lot of questions about the `core-typings/src/Apps.ts`. Internally in Marketplace we have AppOverview and AppInfo. While they relate to the same app, they contain vastly different information. One is about the app as a whole (hence `Overview`) and the other is information about a version (hence `Info`). However, this typing is a mixture of both in addition to other information. Can someone explain what is going on?

ggazzo:
> 

well, the types were not checked internally, they were built based on the code/usage.

clearly you are the most knowledgeable person on the subject, what do you think of a fix commit?

packages/core-typings/src/Apps.ts
rique223: Oversight, will be solving this on the next commit.

packages/core-typings/src/Apps.ts
rique223: Solving this on the next commit.

packages/core-typings/src/Apps.ts
rique223: Solving it on the next commit.

packages/core-typings/src/Apps.ts
rique223: detailedDescription was defined on the spot. 
detailedChangelog wasn't being used and, consequently, not defined, I'll be defining it on the next commit for future endeavors.

packages/core-typings/src/Apps.ts
rique223: Not sure, we'll have to discover and correct it on a future PR.

packages/core-typings/src/Apps.ts
rique223: Indeed it looks weird. Let's have a discovery on this and solve it.

packages/core-typings/src/Apps.ts
rique223: Also, something to be discovered about.

ggazzo (changes requested):
Please check the translations, I'm afraid you are missing some

rique223:
> Please check the translations, I'm afraid you are missing some

Those are being translated on the "view" side of things, that's why I skipped it, I imagine that if I return a translated string from there it'll bug the translation function since it'll try to translate twice.

PR #25711, opened 2022-06-01
Chore: Update text from `Layout_Home_Body`

LucianoPierdona:

Update `Layout_Home_Body` default text and add `blocked` field in it


apps/meteor/server/startup/migrations/v267.ts
```diff
@@ -0,0 +1,22 @@
...
+		const newValue = `<p>~~~~ Default html example ~~~~</p>\n<strong>Welcome to (ENTER ORGANIZATION NAME HERE)</strong>\n\n<p>All general communications should be done through #general</p>\n<p>find more information <a href="INSERT LINK" target="_blank" rel="noopener">here</a></p>`;
+
+		await Settings.updateMany(
```
matheusbsilva137, line 10:
```suggestion
		await Settings.updateOne(
```
Since only one setting will be updated, we could use `updateOne` here instead of `updateMany`.

matheusbsilva137 (changes requested):
The changes applied to the `Layout_Home_Body` look fine. We still need to update the new `Layout_Custom_Body` setting, though
Maybe you could use ` Settings.getValueById('Layout_Home_Body')` before updating the `Layout_Home_Body` setting's value so that you know if `Layout_Custom_Body` should be true or false.

gabriellsh (commented):
BTW, you forgot to import the migration in the index :p. I'm testing it so I'll fix it.

apps/meteor/server/startup/migrations/v268.ts
```diff
@@ -0,0 +1,37 @@
...
+		}
+
+		await Settings.updateOne(
+			{
+				_id: 'Layout_Home_Body',
+				value: oldValue,
+			},
+			{
+				$set: {
+					value: newValue,
```
gabriellsh, line 32: We need to ALWAYS update the `packageValue` to be the new value. This ensures that when we click the `reset setting` button (red backwards arrow)  the text resets to the new value.

gabriellsh (dismissed):

matheusbsilva137 (approved):

gabriellsh (approved):

PR #25726, opened 2022-06-02
[FIX] Session with loginToken not being saved after login

albuquerquefabio:
## Issue(s)

1. 1 Login with your account
1.1. Check collection rocketchat_sessions and there we found 2 documents related to this session started.
2. Click on logout button
3. Login again
3.1. Check rocketchat_sessions again and there is 1 document related to this session started. 

## Further comments

Why do we need 2 documents?
1st sends the object resume with a username and password
2nd sends the object resume with token only after processed login with success.

*both documents are sent by Meteor.Account
**This token is an important key to filtering all opened sessions.

### DIFF
--- a/apps/meteor/app/statistics/server/lib/SAUMonitor.ts
+++ b/apps/meteor/app/statistics/server/lib/SAUMonitor.ts
@@ -138,7 +138,7 @@ export class SAUMonitorClass {
 		if (!data) {
 			return;
 		}
-		await Sessions.createOrUpdate(data);
+		await Sessions.insertOne({ ...data, createdAt: new Date() });
 	}
 
 	private async _finishSessionsFromDate(yesterday: Date, today: Date): Promise<void> {


apps/meteor/app/statistics/server/lib/SAUMonitor.ts
```diff
@@ -138,7 +138,7 @@ export class SAUMonitorClass {
 		if (!data) {
 			return;
 		}
-		await Sessions.createOrUpdate(data);
+		await Sessions.insertOne({ ...data, createdAt: new Date() });
```
debdutdeb, line 141:
```suggestion
		await Sessions.insertOne({ ...data, createdAt: now });
```

debdutdeb (changes requested):
Since no longer `updating` - would be better to rename the method to just `create`.

debdutdeb (approved):

PR #25731, opened 2022-06-02
[FIX] Broken Omnichannel>Agents page

tiagoevanp:


apps/meteor/client/views/omnichannel/agents/AgentsRoute.tsx
```diff
@@ -63,9 +63,9 @@ const AgentsRoute: FC = () => {
 	const context = useRouteParameter('context');
 	const id = useRouteParameter('id');
 
-	if (!id) {
-		throw new Error('Agent id is required');
-	}
+	// if (!id) {
+	// 	throw new Error('Agent id is required');
+	// }
```
KevLehman, line 68: 👀

tiagoevanp: ouch!

apps/meteor/client/components/FilterByText.tsx
```diff
@@ -27,7 +27,7 @@ const FilterByText: FC<FilterByTextProps> = ({ placeholder, onChange: setFilter,
 	}, []);
 
 	useEffect(() => {
-		setFilter({ text });
```
MartinSchoeler, line 30: Isn’t `setFilter` a required prop? It should never be undefined, if it is happening, we should fix the place that is not setting the prop

Or make it optional.

MartinSchoeler (changes requested):

apps/meteor/client/views/omnichannel/agents/AgentsRoute.tsx
```diff
@@ -180,8 +176,8 @@ const AgentsRoute: FC = () => {
 					<VerticalBar.Close onClick={handleVerticalBarCloseButtonClick} />
 				</VerticalBar.Header>
 
-				{context === 'edit' && <AgentEditWithData uid={id} reload={reload} />}
-				{context === 'info' && (
+				{context === 'edit' && id && <AgentEditWithData uid={id} reload={reload} />}
```
ggazzo, line 179:
```suggestion
				{context === 'edit' && (id ? <AgentEditWithData uid={id} reload={reload} />: <SomeErrorPage />}
```

tiagoevanp: ok

MartinSchoeler (changes requested):

murtaza98:
@tiagoevanp Just a suggestion to add a basic test case for this page to verify if this page is getting loaded or not - just so we can avoid this type of issue in future

tiagoevanp:
> 

Yes, thanks for the comment Murtaza. I know that we have different teams working on tests outside of our main team at rocket.chat... Probably this case will be covered by them.

MartinSchoeler (approved):

PR #25745, opened 2022-06-03
Chore: Implementing Service Middlewares

pierre-lehnen-rc:


pierre-lehnen-rc:
This implementation wouldn't work from service functions called from inside a micro service. (for example: when the Authorization micro service calls something from the License service which is running on the monolith).

PR #25750, opened 2022-06-03
[FIX] Access issue on chat.getThreadsList

murtaza98:

### DIFF
--- a/apps/meteor/app/api/server/v1/chat.js
+++ b/apps/meteor/app/api/server/v1/chat.js
@@ -524,6 +524,10 @@ API.v1.addRoute(
 	{
 		get() {
 			const { rid, type, text } = this.queryParams;
+			check(rid, String);
+			check(type, String);
+			check(text, Match.Maybe(String));
+
 			const { offset, count } = this.getPaginationItems();
 			const { sort, fields, query } = this.parseJsonQuery();
 
@@ -547,7 +551,7 @@ API.v1.addRoute(
 				msg: new RegExp(escapeRegExp(text), 'i'),
 			};
 
-			const threadQuery = { ...query, ...typeThread, rid, tcount: { $exists: true } };
+			const threadQuery = { ...query, ...typeThread, rid: room._id, tcount: { $exists: true } };
 			const cursor = Messages.find(threadQuery, {
 				sort: sort || { tlm: -1 },
 				skip: offset,


gabriellsh (changes requested):
Tests failing. I think it's related to the new checks

murtaza98:
> Tests failing. I think it's related to the new checks

Yups, indeed it was. It should be fixed now. Can you please check again :)

apps/meteor/app/api/server/v1/chat.js
```diff
@@ -524,6 +524,10 @@ API.v1.addRoute(
 	{
 		get() {
 			const { rid, type, text } = this.queryParams;
+			check(rid, String);
```
KevLehman, line 527: Since now you have this, the check on L534 is no longer needed

apps/meteor/app/api/server/v1/chat.js
```diff
@@ -547,7 +551,7 @@ API.v1.addRoute(
 				msg: new RegExp(escapeRegExp(text), 'i'),
 			};
 
-			const threadQuery = { ...query, ...typeThread, rid, tcount: { $exists: true } };
+			const threadQuery = { ...query, ...typeThread, rid: room._id, tcount: { $exists: true } };
```
KevLehman, line 554: I know is OOS, but, would be good to move this to the model 😬

KevLehman (changes requested):

apps/meteor/app/api/server/v1/chat.js
murtaza98: Technically you're right however I'd rather prefer using strings returned from the database rather than a user input while trying to generate a mongo query :grimacing:

apps/meteor/app/api/server/v1/chat.js
murtaza98: Yeah, I agree, not a good practice to build a query on the route itself. However since this will be treated as a security fix, I'd prefer to keep the changes as minimal here and pick this up on Eng day in a separate PR

apps/meteor/app/api/server/v1/chat.js
murtaza98: Oops! Sorry I misunderstood your comment earlier here :see_no_evil: I've removed that check now. Thanks!!

PR #25756, opened 2022-06-03
Chore: Migrate from meteor model to raw

KevLehman:


packages/core-typings/src/ILivechatVisitor.ts
```diff
@@ -20,6 +20,10 @@ export interface IVisitorEmail {
 	address: string;
 }
 
+interface ILivechatData {
+	[k: string]: any;
```
ggazzo, line 24: could you check if unknown works?
```suggestion
	[k: string]: unknown;
```

KevLehman: Typecheck didn't complain, so should work 👀

ggazzo (approved):

thassiov (dismissed):

apps/meteor/app/livechat/imports/server/rest/sms.js
```diff
@@ -53,8 +54,8 @@ const defineVisitor = (smsNumber, targetDepartment) => {
 		data.department = targetDepartment;
 	}
 
-	const id = Livechat.registerGuest(data);
-	return LivechatVisitors.findOneById(id);
+	const id = Promise.await(Livechat.registerGuest(data));
```
sampaiodiego, line 57: this can be changed to regular `await`

apps/meteor/app/livechat/server/api/lib/livechat.js
```diff
@@ -39,15 +39,17 @@ export function findDepartments(businessUnit) {
...
-	return LivechatVisitors.getVisitorByToken(token, {
-		fields: {
-			name: 1,
-			username: 1,
-			token: 1,
-			visitorEmails: 1,
-			department: 1,
-		},
-	});
+	return Promise.await(
```
sampaiodiego, line 42: most places I found using `findGuest` were already `async`, so changing this to `async` would be an easy "enough" change

apps/meteor/app/livechat/server/methods/registerGuest.js
```diff
@@ -32,14 +33,15 @@ Meteor.methods({
...
+			// TODO: refactor to use normal await
 			customFields.forEach((customField) => {
 				if (typeof customField !== 'object') {
 					return;
 				}
 
 				if (!customField.scope || customField.scope !== 'room') {
 					const { key, value, overwrite } = customField;
-					LivechatVisitors.updateLivechatDataByToken(token, key, value, overwrite);
+					Promise.await(LivechatVisitors.updateLivechatDataByToken(token, key, value, overwrite));
```
sampaiodiego, line 44:
```suggestion
					await LivechatVisitors.updateLivechatDataByToken(token, key, value, overwrite);
```

apps/meteor/server/startup/migrations/v260.ts
```diff
@@ -32,7 +32,7 @@ addMigration({
 			});
 
 			if (updates.length) {
-				Promise.await(VisitorsRaw.col.bulkWrite(updates));
+				Promise.await(LivechatVisitors.col.bulkWrite(updates));
```
sampaiodiego, line 35: migrations can be `async`
```suggestion
				await LivechatVisitors.col.bulkWrite(updates);
```

sampaiodiego (changes requested):

apps/meteor/app/livechat/server/methods/registerGuest.js
KevLehman: I kept this one as Promise.await to not refactor the foreach in this PR 😬

apps/meteor/server/startup/migrations/v260.ts
KevLehman: C'mon, this mig is from like 4 months ago :trollface:

sampaiodiego (approved):

sampaiodiego (approved):

PR #25764, opened 2022-06-05
Chore: ts rewrite users raw model

debdutdeb:


sampaiodiego:
are you still planning finishing this one? it is getting far behind and has some conflict code merged 🙈

apps/meteor/app/livechat/server/api/lib/users.ts
```diff
@@ -38,8 +48,18 @@ async function findUsers({ role, text, pagination: { offset, count, sort } }) {
...
 	};
 }
-export async function findAgents({ text, pagination: { offset, count, sort } }) {
-	return findUsers({
+export async function findAgents({
+	text,
+	pagination: { offset, count, sort },
+}: {
+	text: string | undefined;
+	pagination: IPaginationOptions & {
```
KevLehman, line 56: Create a helper type "WithSort<>" and pass IPaginationOptions, or create an extended type 👀 (so you don't repeat & { sort }...)

apps/meteor/ee/app/livechat-enterprise/server/business-hour/Helper.ts
```diff
@@ -10,7 +10,7 @@ const getAllAgentIdsWithoutDepartment = async (): Promise<string[]> => {
 	).map((dept: any) => dept.agentId);
 	const agentIdsWithoutDepartment = (
 		await Users.findUsersInRolesWithQuery(
-			'livechat-agent',
+			['livechat-agent'],
```
KevLehman, line 13: So, this is a "current" bug, or does the "findUsersInRolesWithQuery" accepts string | string[] ? 👀

apps/meteor/ee/app/livechat-enterprise/server/business-hour/Multiple.ts
```diff
@@ -200,7 +200,16 @@ export class MultipleBusinessHoursBehavior extends AbstractBusinessHourBehavior
 				return false;
 			}
 
-			const activeBusinessHoursForAgent = departmentsWithActiveBH.map(({ businessHourId }) => businessHourId);
+			const activeBusinessHoursForAgent = departmentsWithActiveBH.reduce(
```
KevLehman, line 203: Instead of reduce, you can chain a .filter to remove "empty" items from the array 👀

apps/meteor/ee/server/api/sessions.ts
```diff
@@ -105,7 +105,7 @@ API.v1.addRoute(
 
 				search.push(
 					...(await Users.findActiveByUsernameOrNameRegexWithExceptionsAndConditions<Pick<IUser, '_id'>>(
-						{ $regex: filter, $options: 'i' },
+						new RegExp(filter, 'i'),
```
KevLehman, line 108: Why? 👀

apps/meteor/ee/server/lib/registerServiceModels.ts
```diff
@@ -38,6 +38,7 @@ export function registerServiceModels(db: Db, trash?: Collection): void {
 	);
 	registerModel('ITeamModel', () => new TeamRaw(db));
 	registerModel('ITeamMemberModel', () => new TeamMemberRaw(db));
+	// @ts-ignore-error
```
KevLehman, line 41: image

apps/meteor/server/lib/findUsersOfRoom.ts
```diff
@@ -40,12 +40,12 @@ export function findUsersOfRoom({
 		...(limit > 0 && { limit }),
 	};
 
-	const searchFields = settings.get<string>('Accounts_SearchFields').trim().split(',');
+	const searchFields = settings.get<string>('Accounts_SearchFields').trim().split(',') as Array<keyof IUser>;
 
-	return Users.findPaginatedByActiveUsersExcept(filter, undefined, options, searchFields, [
+	return Users.findPaginatedByActiveUsersExcept(filter, [], options, searchFields, [
```
KevLehman, line 45: Do undefined and [] work the same? 👀

apps/meteor/ee/app/livechat-enterprise/server/business-hour/Helper.ts
debdutdeb: No, using non-list has been deprecated, don't know how long. I just switched all singles to lists.

apps/meteor/ee/app/livechat-enterprise/server/business-hour/Multiple.ts
debdutdeb: *BUT*, this way we just had to iterate once? 🙈 I am very critical that way lol.

apps/meteor/ee/server/api/sessions.ts
debdutdeb: Ah nvm I confused between different things.

debdutdeb: made sense to not mix types so much 😬  they're the same in this context any way

apps/meteor/ee/server/lib/registerServiceModels.ts
debdutdeb: 🙈 ops

apps/meteor/server/lib/findUsersOfRoom.ts
debdutdeb: *Yeah*

This one, or rather **these** ones I am not so sure about; it's already a bad union (old code baby), type and type[] ? then adding another falsy undefined. the previous implementation of that method did a null check and default set that to a list any way. So, I moved that default here. 

Open to other POVs though.

Again, we don't have standards for these (at least not that I know of), like when to use null/undefined vs when to use type-specific falsy inputs like empty lists or empty strings etc. The rule that **I** follow or like to follow is any serializable data, use the type specific falsy, for objects "alive" objects or function types, use undef/null, with exceptions if any.

debdutdeb: There are other places that kinda does this mix as well iirc

apps/meteor/ee/app/livechat-enterprise/server/business-hour/Multiple.ts
KevLehman: Yeah, this is a bit like "pick your poison". For this case, departments with active BH won't be an insane list that would hang the server 👀 most operations would have this list in the orders of 10-99 maybe. So, not sure how much load would an array of strings produce 🤔 

We can keep the reduce, just, don't want to sacrifice readability for optimization 👀 .

apps/meteor/ee/app/livechat-enterprise/server/business-hour/Helper.ts
KevLehman: naice

apps/meteor/ee/server/api/sessions.ts
KevLehman: But, is this function only used here? Cause changing the params would break other places if any 👀

apps/meteor/server/lib/findUsersOfRoom.ts
KevLehman: Not really sure 🤔 I mean, you're right that there's no standard, but what feels better? 👀 
@sampaiodiego , wanna put an standard here? :trollface:

debdutdeb: To me it's like english (or any language really), a parameter equals a question; for example in this case, `what are the username exceptions?` - the answer would be `none`, i.e. empty list, not `undefined`, i.e. "i don't know".

debdutdeb: My thought process, if it helps:

To me it's like english (or any language really), a parameter equals a question; for example in this case, "what are the username exceptions?" - the answer would be `no exceptions`, i.e. `empty list`, not `undefined`, i.e. `i'm not sure`.

it also enforces others to be more alert of **what** they're passing to a function.

debdutdeb: PS: i should mention the memory overhead, although I don't know how much, but from that perspective undefined is more "efficient" than empty lists or objects. But then again, would it matter *so much*? IDK 🙈

KevLehman:
@debdutdeb , is this still needed? Worth solving conflicts or redo the PR? 👀

apps/meteor/ee/server/api/sessions.ts
KevLehman: I'm with you on that one. But yeah, to keep the change pls find other places using the func and update accordingly

PR #25793, opened 2022-06-07
[FIX] Slackbridge Connect Error

LucianoPierdona:

This PR fixes an issue that was happening when an invalid token was passed on SlackBridge, basically the app crashes because the error was not being handled


apps/meteor/app/slackbridge/server/SlackAdapter.js
```diff
@@ -41,7 +41,19 @@ export default class SlackAdapter {
 		}
 		this.slackAPI = new SlackAPI(this.apiToken);
 		this.rtm = new RTMClient(this.apiToken);
-		this.rtm.start();
+
+		Promise.await(
```
ggazzo, line 45: is `Promise.await` needed?

LucianoPierdona: yes, if we don't use it, the `then()` and `catch()` doesn't get executed and we cannot handle the error.

ggazzo: but you are still throwing the error, so if the application was breaking here before the changes its still breaking, or is there something I'm missing?

apps/meteor/app/slackbridge/server/slackbridge.js
```diff
@@ -30,14 +30,21 @@ class SlackBridgeClass {
...
+					this.rocket.addSlack(slack);
+					this.slackAdapters.push(slack);
+
+					slack.connect(apiToken);
+				});
+			} catch (err) {
+				connLogger.error('error connecting to slack', err);
+
+				throw err;
+			}
```
ggazzo, line 47: changing the scope of your try/catch, your application becomes more "sensitive", before, it would try all tokens, even if some failed, now it stops at the first token that fails

apps/meteor/app/slackbridge/server/SlackAdapter.js
ggazzo: Actually throwing the error and Adding Promise.await it will breaks the application for sure (thats why you were forced to add the try/catch here https://github.com/RocketChat/Rocket.Chat/pull/25793/files#diff-034e87d82cdd01feb879443b1f59f27a70dbebb6c8b60bc1827eec093942347dR34

ggazzo: I think you can change the connect method to async and use `await` instead of `Promise.await` and then just handle the exception like that `slack.connect(apiToken).catch(e => connLogger.error('error connecting to slack', e)`

apps/meteor/app/slackbridge/server/slackbridge.js
LucianoPierdona: removed try/catch

apps/meteor/app/slackbridge/server/SlackAdapter.js
LucianoPierdona: changes done!

matheusbsilva137 (approved):
The requested chanegs have been applied. LGTM

matheusbsilva137 (approved):

PR #25795, opened 2022-06-07
[NEW] Create header for new homepage

guijun13:
## Further comments

TODO: 
- isAdmin permission


dougfabris:
@guijun13 With this new page, I think we can deprecate the old burger menu 😃

dougfabris (commented):
@guijun13 Having this new page, I think we can remove the old burger menu 😃

guijun13:
> @guijun13 Having this new page, I think we can remove the old burger menu smiley

Removed! Thanks, @dougfabris

gabriellsh (approved):

PR #25798, opened 2022-06-08
Chore: A MUCH better pull request template

debdutdeb:
## (Optional) A joke, a meme or a cute picture of an animal, if possible related to what you did
image


ggazzo:
In fact the pictures and jokes will end up in the changelog, I don't believe this is really useful.

The idea is good, but it would force us to reconsider how the  changelog is generated

PR #25806, opened 2022-06-08
[NEW] Send notification when login

albuquerquefabio:

image


apps/meteor/server/startup/migrations/v271-deviceManagement.ts
```diff
@@ -0,0 +1,117 @@
...
+    }
+    .rc-color {
+      color: #F5455C;
+    }`;
+		await Settings.updateOne(
+			{
+				_id: 'email_style',
+			},
+			{
+				$set: {
```
sampaiodiego, line 110: you should not overwrite the whole setting like this, doing so will discard/revert any change the sys admin might have done.

I saw the only change to the setting was a new `.rc-color` class.. so for this migration I'd suggest you to just add the new class to the existing value, this way no change will be lost during the migration.

apps/meteor/server/startup/migrations/v271-deviceManagement.ts
```diff
@@ -0,0 +1,117 @@
+import { addMigration } from '../../lib/migrations';
+import { Settings } from '../../../app/models/server/raw';
+
+addMigration({
+	version: 267,
```
sampaiodiego, line 5: the version is incorrect.. it should be `271` per file name.

apps/meteor/ee/server/lib/deviceManagement/session.ts
```diff
@@ -0,0 +1,121 @@
...
+		const user = await Users.findOneById<IUser>(userId, { projection: { name: 1, username: 1, emails: 1 } });
+		if (user?.emails?.length && !connection.loginToken) {
+			const {
+				name,
+				username,
+				emails: [{ address: email }],
+			} = user;
+			const { browser, os, device, cpu, app } = await uaParser(connection.httpHeaders['user-agent']);
+
+			let mailData = {};
```
sampaiodiego, line 51: I saw a lot of duplicated code when defining a new value for `mailData` in the `switch` below, instead I'd recommend to change to a `const` and provide the common values here and just provide the new value on the switch statement, something like this:
```suggestion
			const mailData = {
				name,
				username,
				ipInfo: connection.clientAddress,
			};
```

apps/meteor/ee/server/lib/deviceManagement/session.ts
```diff
@@ -0,0 +1,121 @@
...
+	if (uaString && UAParserDesktop.isDesktopApp(uaString)) {
+		rcAgent = UAParserDesktop.uaObject(uaString);
+	}
+
+	const ua = new UAParser(uaString);
+	return { ...ua.getResult(), ...rcAgent };
+};
+
+export const listenSessionLogin = async (): Promise<void> => {
+	sauEvents.on('accounts.login', async ({ userId, connection }) => {
```
sampaiodiego, line 39: even though this works, you're adding a "hidden" dependency to the SAU Monitor service, which is not ideal.

what you should do instead, is to create a new service (let's say a `DeviceManagementService`) and add a listener to the original `accounts.login` event, just like the SAU Monitor service does, as you can see below:

https://github.com/RocketChat/Rocket.Chat/blob/33f28392697fcf93eceb23982e3e9451799e93e8/apps/meteor/server/services/sauMonitor/service.ts#L13

the difference though is that it should be an EE service, pretty much like `LDAPEEService`

sampaiodiego (changes requested):

apps/meteor/server/startup/migrations/v271-deviceManagement.ts
albuquerquefabio: I added `apps/meteor/app/lib/server/startup/email.ts` where the content is also created.

sampaiodiego:
already merged into #25791

PR #25807, opened 2022-06-08
[NEW] [VideoConference] Provider apps

pierre-lehnen-rc:

- [X] Jitsi Basic functionality
- [ ] Jitsi Advanced features
- [X] Pexip basic functionality
- [ ] Pexip Advanced features
- [ ] BBB basic functionality


_templates/rc-app/new/app.json.ejs.t
```diff
@@ -0,0 +1,19 @@
+---
+to: apps/<%= name %>/app.json
+---
+
+{
+	"id": "",
+	"version": "0.0.1",
+	"requiredApiVersion": "^1.4.0",
```
graywolf336, line 8: If you're going to use `"@rocket.chat/apps-engine": "^1.19.0",` below in the `package.json`, then this should declare that version.

geekgonecrazy:
Idk that I like the app being inside the repo. Feels weird.  But I am super happy to see this coming!  It was a long time ago that we hacked that initial jitsi integration in from hotel lobby 😁😁

pierre-lehnen-rc:
> Idk that I like the app being inside the repo. Feels weird. 

We still need to decide where they'll be stored. I'm just using the monorepo for now to be able to load the same lint and ts rules from the main app.

But I do think this is the best place to put our apps.

PR #25821, opened 2022-06-09
Chore: Update Docker Compose with newer mongo and using wiredTiger

dudanogueira:
Update docker-compose in order to avoid deprecation warnings and use newer one.


debdutdeb:
As discussed, the updated files are in https://github.com/RocketChat/Docker.Official.Image; asked Fon to update our docs :)

PR #25822, opened 2022-06-09
[NEW] Hide channel actions + message box when the room is federated

MarcosSpessatto:

The goal of this PR is to disable the following items on federated rooms:
- Threads
- Discussions
- Call
- Files

And hide the following:
- Prune messages
- Livestream
- Pinned Messages

Also, hide the following items on the message box:
- Audio Message
- Plus icon
- Emojis
- Ability to upload files


apps/meteor/client/views/room/Header/ToolBox/ToolBox.tsx
```diff
@@ -32,22 +32,24 @@ const ToolBox = ({ className }: ToolBoxProps): ReactElement => {
...
-				[item.id]: item.renderOption || renderMenuOption,
-			};
-			return [
-				item.id,
-				{
-					label: { title: t(item.title), icon: item.icon },
-					action: (): void => {
-						openTabBar(item.id);
+		(isMobile ? actions : actions.slice(6))
+			.filter((item) => !item.disabled)
```
MarcosSpessatto, line 36: @ggazzo I decided to just filter here instead of showing it disabled since the options inside the Menu component seem not to support disabled options.

Any suggestion?

PR #25826, opened 2022-06-10
[NEW] [VideoConference] Livechat Support

pierre-lehnen-rc:


apps/meteor/server/services/video-conference/service.ts
```diff
@@ -293,15 +351,53 @@ export class VideoConfService extends ServiceClassInternal implements IVideoConf
...
+		});
+
+		const call = await this.getUnfiltered(callId);
+		if (!call) {
+			throw new Error('failed-to-create-livechat-call');
+		}
+
+		const joinUrl = await this.getUrl(call);
+		const messageId = await this.createLivechatMessage(rid, user, callId, joinUrl);
+		this.VideoConference.setMessageById(callId, 'started', messageId);
```
murtaza98, line 382:
```suggestion
		await this.VideoConference.setMessageById(callId, 'started', messageId);
```

murtaza98: since this func returns a promise

packages/livechat/src/components/Messages/MessageList/index.js
```diff
@@ -127,6 +131,17 @@ export class MessageList extends MemoizedComponent {
...
+			if (videoConfJoinBlock) {
+				if (ongoingCall && isCallOngoing(ongoingCall.callStatus)) {
+					const { url, callProvider, rid } = incomingCallAlert || {};
+					items.push(
+						<JoinCallButton callStatus={ongoingCall.callStatus} url={url} callProvider={callProvider} rid={rid} />,
+					);
+				}
+				continue;
+			}
+
```
murtaza98, line 144: Is this code really required? since LiveChat is totally equipped to deal with UIKit messages so the join call message which you're sending on the core for this video call should be rendered here as well. I tested it locally and here's the result. As you can see, the joincall link is visible to visitor

image

murtaza98: Btw, this current code seems to be failing due to that translation error

pierre-lehnen-rc: I tried not to modify the current behavior, but replacing it with the regular message is fine by me.
Main difference I can see is that the current join button disappears once the call is over and the videoconf message will remain there.

murtaza98: since the existing code isn't working due to that weird translation bug, I'd also say let's keep the regular message!!

pierre-lehnen-rc: That code with the translation error would still be used by the webrtc option though (I haven't changed anything on how webrtc works)

PR #25841, opened 2022-06-13
[FIX] Permission `view-all-teams` is not checked in the `teams.info` endpoint

LucianoPierdona:

Previously any authenticated user was able to access the `teams.info` endpoint, this PR updates this so only users with the `view-all-teams` permission or team members can access it.

### User 1
- Create a private team with only User 1 added to it.

### User 2
- Access the above team through `curl`
```
curl  -H "X-Auth-Token: {authToken}" \
      -H "X-User-Id: {userId}" \
      -H "Content-type: application/json" \
      http://localhost:3000/api/v1/teams.info?teamName=Team1
```

### Expected result
- If User 2 _doesn't have_ the `view-all-teams` permission, the response must be `unauthorized`;
- If User 2 _has_ the `view-all-teams` permission, the response must be succesful;
- For User 1, the response must be succesful;
- If the team is changed to public, the response must be succesful for User 1 and User 2 regardless of the `view-all-teams` permission.


apps/meteor/app/api/server/v1/teams.ts
```diff
@@ -579,6 +579,14 @@ API.v1.addRoute(
 				return API.v1.failure('Team not found');
 			}
 
+			const canViewInfo =
+				hasPermission(this.userId, 'view-all-teams', teamInfo.roomId) ||
```
KevLehman, line 583: If this is the same endpoint that normal users use to get the info about a team, this will break it for them.

LucianoPierdona: check updated to see if he is an admin or a team member

matheusbsilva137 (dismissed):

matheusbsilva137 (approved):

ggazzo (approved):

apps/meteor/app/api/server/v1/teams.ts
```diff
@@ -576,6 +576,15 @@ API.v1.addRoute(
 				return API.v1.failure('Team not found');
 			}
 
+			const canViewInfo =
+				(await Team.getMember(teamInfo._id, this.userId)) ||
+				(await Subscriptions.findOneByRoomIdAndUserId(teamInfo.roomId, this.userId)) ||
+				hasPermission(this.userId, 'view-all-teams');
```
matheusbsilva137, line 582: I think a better option here would be to use the `canAccessRoom` method instead of checking if the user is a member of the team and searching for the subscription directly. By doing so, we'd no longer need the `getMember` method.

matheusbsilva137 (changes requested):

apps/meteor/app/api/server/v1/teams.ts
LucianoPierdona: updated!

apps/meteor/app/api/server/v1/teams.ts
```diff
@@ -576,6 +576,18 @@ API.v1.addRoute(
 				return API.v1.failure('Team not found');
 			}
 
+			const room = Rooms.findOneById(teamInfo.roomId);
+
+			if (!teamInfo) {
+				return API.v1.failure('Room not found');
+			}
```
matheusbsilva137, line 583:
```suggestion
			if (!teamInfo) {
				return API.v1.failure('Room not found');
			}

			const room = Rooms.findOneById(teamInfo.roomId);
```
Since we're using using information from `teamInfo` to search for the room, we should check if `teamInfo` is not undefined _before_ searching for the room.

matheusbsilva137 (changes requested):

apps/meteor/app/api/server/v1/teams.ts
LucianoPierdona: updated!

apps/meteor/tests/end-to-end/api/25-teams.js
```diff
@@ -819,6 +819,77 @@ describe('[Teams]', () => {
...
+				.query({
+					teamName: publicTeam.name,
+				})
+				.expect('Content-Type', 'application/json')
+				.expect(200)
+				.expect((response) => {
+					expect(response.body).to.have.property('success', true);
+					expect(response.body).to.have.property('teamInfo');
+					expect(response.body.teamInfo).to.have.property('_id').and.to.be.eq(publicTeam._id);
+					expect(response.body.teamInfo).to.have.property('name').and.to.be.eq(publicTeam.name);
```
matheusbsilva137, line 836:
```suggestion
					expect(response.body.teamInfo).to.have.property('_id', publicTeam._id);
					expect(response.body.teamInfo).to.have.property('name', publicTeam.name);
```
You can check the property and its value all at once.

apps/meteor/tests/end-to-end/api/25-teams.js
```diff
@@ -819,6 +819,77 @@ describe('[Teams]', () => {
...
+				.query({
+					teamId: publicTeam._id,
+				})
+				.expect('Content-Type', 'application/json')
+				.expect(200)
+				.expect((response) => {
+					expect(response.body).to.have.property('success', true);
+					expect(response.body).to.have.property('teamInfo');
+					expect(response.body.teamInfo).to.have.property('_id').and.to.be.eq(publicTeam._id);
+					expect(response.body.teamInfo).to.have.property('name').and.to.be.eq(publicTeam.name);
```
matheusbsilva137, line 854:
```suggestion
					expect(response.body.teamInfo).to.have.property('_id', publicTeam._id);
					expect(response.body.teamInfo).to.have.property('name', publicTeam.name);
```

You can check the property and its value all at once.

matheusbsilva137 (changes requested):

apps/meteor/tests/end-to-end/api/25-teams.js
LucianoPierdona: Updated!

apps/meteor/tests/end-to-end/api/25-teams.js
LucianoPierdona: Updated!

matheusbsilva137 (approved):
LGTM. We just need a review from the chat-engine team for the new tests.

ggazzo (approved):

PR #25863, opened 2022-06-14
Chore: Remove old rest api code

ggazzo:


KevLehman (approved):
Not that my approval will unlock the process, but here i am :trollface:

PR #25891, opened 2022-06-15
[FIX] Settings not being overwritten to their default values

sampaiodiego:


geekgonecrazy (approved):
This is super helpful

PR #25896, opened 2022-06-16
[IMPROVE] VoIP configuration cleanup

cauefcr:

Some configuration fields were redundant, or not used, this PR fixes this.


apps/meteor/server/startup/migrations/v270.ts
```diff
@@ -0,0 +1,14 @@
+import { addMigration } from '../../lib/migrations';
+import { Settings as RawSettings } from '../../../app/models/server/raw';
+
+// Removes unused VoIP settings
+addMigration({
+	version: 270,
+	async up() {
+		await RawSettings.deleteMany({
+			_id: {
+				$in: ['VoIP_Server_Name', 'VoIP_Server_Host', 'VoIP_Server_Websocket_Path', 'VoIP_Server_Websocket_Port'],
```
murtaza98, line 10: Some of these settings are still being referenced within the codebase right now - like for example over [here](https://github.com/RocketChat/Rocket.Chat/blob/0d12cdec197013653e7e7063a7b21956bc736d97/apps/meteor/server/services/voip/lib/Helper.ts#L10)

PR #25900, opened 2022-06-17
[NEW] [VideoConference] Expire Old Calls

pierre-lehnen-rc:


apps/meteor/server/cron/videoConferences.ts
```diff
@@ -0,0 +1,29 @@
...
+
+export function videoConferencesCron(syncedCron: typeof SyncedCron): void {
+	settings.watch('VideoConf_Enabled', (value) => {
+		if (!value) {
+			return syncedCron.remove('VideoConferences');
+		}
+		syncedCron.add({
+			name: 'VideoConferences',
+			schedule(parser: any) {
+				return parser.cron('* */3 * * *');
```
sampaiodiego, line 24: this is running on every minute every 3 hours.. if you meant to run *once* every 3 hours, this is the correct cron:
```suggestion
				return parser.cron('0 */3 * * *');
```

apps/meteor/app/models/server/raw/VideoConference.ts
```diff
@@ -145,4 +140,25 @@ export class VideoConferenceRaw extends BaseRaw<VideoConference> {
 			},
 		});
 	}
+
+	public async expireOldVideoConferences(minDate: Date): Promise<void> {
+		await this.update(
```
sampaiodiego, line 145: `update` is deprecated, pls use `updateMany` or `updateOne`:
```suggestion
		await this.updateMany(
```

sampaiodiego (changes requested):

PR #25901, opened 2022-06-17
[FIX] [VideoConference] Changed how the calling popup is brought up

pierre-lehnen-rc:


dougfabris (approved):

apps/meteor/client/lib/VideoConfManager.ts
```diff
@@ -66,6 +66,8 @@ type VideoConfEvents = {
 	// The list of ringing incoming calls may have changed
 	'ringing/changed': void;
 
+	// The value os `isCalling` may have changed
```
graywolf336, line 69:
```suggestion
	// The value of `isCalling` may have changed
```

PR #25905, opened 2022-06-17
[NEW] remove user from matrix room

carlosrodrigues94:


apps/meteor/app/federation-v2/server/infrastructure/rocket-chat/callbacks/room.ts
```diff
@@ -0,0 +1,22 @@
...
+import { IFederationBridge } from '../../../domain/IFederationBridge';
+
+class RocketChatRoomCallbacksHandler {
+	private bridge: IFederationBridge;
+
+	public injectBridgeInstance(bridgeInstance: IFederationBridge): void {
+		this.bridge = bridgeInstance;
+	}
+
+	public async afterLeaveRoom(rocketUser: any, rocketRoom: any): Promise<void> {
```
MarcosSpessatto, line 11: I believe it would be better to have this logic inside the `RoomServiceSender`, so here we can just listen to the event itself, in this way all of our business logic will be in a single place, what do you think?

apps/meteor/app/federation-v2/server/infrastructure/matrix/Bridge.ts
```diff
@@ -132,6 +132,10 @@ export class MatrixBridge implements IFederationBridge {
 		`);
 	}
 
+	public async leaveRoom(externalRoomId: string): Promise<void> {
+		await this.bridgeInstance.getIntent('user-id').leave(externalRoomId);
```
MarcosSpessatto, line 136: We can't hardcode the userId here, We do need to provide an userId instead.

apps/meteor/app/federation-v2/server/infrastructure/rocket-chat/adapters/Settings.ts
```diff
@@ -30,9 +30,15 @@ export class RocketChatSettingsAdapter {
 	}
 
 	public getBridgePort(): number {
-		const [, , port] = this.getBridgeUrl().split(':');
+		const defaultPort = 3334;
```
MarcosSpessatto, line 33: I don't think we should hardcode a default port here. In my opinion, it would be better to find the root problem of having the port undefined.

apps/meteor/app/federation-v2/server/index.ts
```diff
@@ -29,7 +33,11 @@ export const federationRoomServiceSender = FederationFactory.buildRoomServiceSen
 
 export const runFederation = async (): Promise<void> => {
 	queueInstance.setHandler(federationEventsHandler.handleEvent.bind(federationEventsHandler), FEDERATION_PROCESSING_CONCURRENCY);
+
 	await federation.start();
+
+	rocketChatRoomCallbacksHandler.injectBridgeInstance(federation);
```
MarcosSpessatto, line 39: In my opinion, it would be better to have a method on the Factory to set up all the listeners and from them, we call the appropriate handler.
Something like: 

```
setupListener(roomServiceSender: FederationRoomServiceSender): void {
  RocketChatRoomCallbacksHandler.onRoomLeft(async (user: IUser, room: IRoom) => await roomServiceSender.onUserLeftRoom(FederationRoomSenderConverter.convert(roomServiceSender.onRoomLeft(FederationRoomSenderConverter.convert()))));
}
```
And so we could move the business logic and centralize it to the service. 

```
public async onRoomLeft(inputDto: Dto): Promise<void> {
 await this.bridge.leaveRoom(rocketRoom._id, rocketUser._id);
}
```

MarcosSpessatto (changes requested):

PR #25908, opened 2022-06-17
[FIX] update query to find active users on `Team.members`

LucianoPierdona:

This PR fixes an issue where previously if no query was passed through, the query to find active users was looking for all users instead of only team members


apps/meteor/server/services/team/service.ts
```diff
@@ -633,14 +633,19 @@ export class TeamService extends ServiceClassInternal implements ITeamService {
...
 		}
 
-		const users = await Users.findActive({ ...query }).toArray();
-		const userIds = users.map((m) => m._id);
-		const cursor = TeamMember.findMembersInfoByTeamId(teamId, count, offset, {
-			userId: { $in: userIds },
-		});
+		const cursor = this.TeamMembersModel.findMembersInfoByTeamId(teamId, count, offset);
+		const records = await cursor.toArray();
+		const teamUserIds = records.map((m) => m.userId);
```
matheusbsilva137, line 638:
```suggestion
		const teamMembers = await this.TeamMembersModel.findMembersInfoByTeamId(teamId, count, offset).toArray();
		const teamUserIds = teamMembers.map((m) => m.userId);
```
Maybe we could change the const names here to improve legibility

apps/meteor/server/services/team/service.ts
```diff
@@ -633,14 +633,19 @@ export class TeamService extends ServiceClassInternal implements ITeamService {
...
+
+		query._id = {
+			$in: teamUserIds,
+		};
```
matheusbsilva137, line 642: I think we should check if `query._id` is undefined here before assigning this new value so as to prioritize the user's query (if there is any)

matheusbsilva137 (changes requested):

apps/meteor/server/services/team/service.ts
LucianoPierdona: Updated!

LucianoPierdona:
Close PR due to fixes have been applied on other PR.

PR #25912, opened 2022-06-17
[FIX] add image format validation

filipemarins:
## Further comments

Add validation to prevent the 'heic' format to be uploaded on the room and user's avatar.
This format is not supported on the browsers and meteor.

Before:
Screen Shot 2022-06-17 at 14 13 49
Screen Shot 2022-06-17 at 14 14 13

After:
Screen Shot 2022-06-17 at 14 15 06
Screen Shot 2022-06-17 at 14 16 01


tassoevan (dismissed):

ggazzo (changes requested):
I still thinks this is something that the sever should handle @tassoevan wyt?

filipemarins:
> I still thinks this is something that the sever should handle @tassoevan wdyt?

I believe this is work for both sides because the browser will try to render this image and we will get an error, for now, we don't handle this kind of error and the user doesn't have proper feedback on what is happening. And make a request to server just to validate the format and return feedback to the client will be more expensive than treating on front.

guijun13 (approved):

guijun13 (dismissed):

gabriellsh (approved):

PR #25925, opened 2022-06-20
[FIX] toolbox menu behind thread component

filipemarins:
## Further comments

Add a container component to usePosition hook to prevent the menu be rendering outside the container.

Before:
Screen Shot 2022-06-20 at 08 19 08

After:
Screen Shot 2022-06-22 at 10 56 19


gabriellsh (dismissed):
image

This crashes the view for extend threads when the screen is smaller.

filipemarins:
> this is supposed to be a generic component, do not enter application details/context here instead do something like `<ToolboxDropdown reference={ref} {...rest} container={PUT THE ELEMENT HERE}>`

Done 👍

hugocostadev (approved):

PR #25934, opened 2022-06-20
[FIX] Importer fails to download files from URLs with query string params

pierre-lehnen-rc:


sampaiodiego (changes requested):
we cannot ignore the get params to download the file, it may be used to identify the correct file to be downloaded. for example both URLs:

```
http://artifacts.com/download?file=A.csv
http://artifacts.com/download?file=ABC.csv
```

pierre-lehnen-rc:
This PR will only change the name of the file that is downloaded to /tmp; It'll still use the full url to get the file contents.

sampaiodiego:
oops.. my bad.. that's right :D

sampaiodiego (approved):

PR #25947, opened 2022-06-21
[FIX] VOIP CallContext snapshot infinite loop

hugocostadev:

The application was crashing due to an error on the `useCallerInfo()` hook.
The error was: 
image
image

To prevent this issue to happen it was added a cached and out-of-scope snapshot variable to the hook using `useSyncExternalStore`

## Steps to test or reproduce

1. Open rocket.chat server
2. Enable Omnichannel 
3. Enable Voip
4. Refresh de page


apps/meteor/client/contexts/CallContext.ts
```diff
@@ -62,10 +62,16 @@ export const useIsCallEnabled = (): boolean => {
...
 export const useIsCallReady = (): boolean => {
-	const { ready } = useContext(CallContext);
+	const context = useContext(CallContext);
+
+	if (isCallContextReady(context)) {
+		callerInfo = context.voipClient.callerInfo;
+	}
 
-	return Boolean(ready);
+	return Boolean(context.ready);
```
tiagoevanp, line 74:
```suggestion
	return !!context.ready;
```

tiagoevanp (approved):

PR #25957, opened 2022-06-22
[FIX] Active users count on `@all` and `@here` 

LucianoPierdona:

this PR updates the old `roomMembersCount` to count active users instead of everyone

## Steps to test or reproduce

Open rocket.chat 
Create a channel/room with all the members
Go to Administration > Settings > General > Notifications.
update Max Room Members Before Disabling All Message Notifications to 2  (if it matches the above requirements, if not, please use the number of active users you have in the room).
send a message with one of the active users and check if you've received a notification with the other one.

Expected Result:

Active users should've received the notification (except the one that send it).


apps/meteor/app/lib/server/lib/sendNotificationsOnMessage.js
```diff
@@ -244,7 +244,12 @@ export async function sendMessageNotifications(message, room, usersInThread = []
 
 	// Don't fetch all users if room exceeds max members
 	const maxMembersForNotification = settings.get('Notifications_Max_Room_Members');
-	const roomMembersCount = Subscriptions.findByRoomId(room._id).count();
+	const roomMembersCount = await Subscriptions.model
+		.rawCollection()
+		.aggregate([{ $match: { rid: room._id } }, lookup, { $match: { 'receiver.active': true } }, { $count: 'count' }])
+		.toArray()
+		.then((res) => res?.[0]?.count || 0);
+
```
ggazzo, line 252: convert to SubscriptionRaw

ggazzo (changes requested):

apps/meteor/app/lib/server/lib/sendNotificationsOnMessage.js
ggazzo: And I think this should be a subscription method

gabriellsh (approved):

sampaiodiego (dismissed):
using a `$lookup` here is not ideal.. large rooms will have issues performing it..

to illustrate this, I ran both codes (the old and the one) against the top 10 rooms in one of our servers and this is the result:

the code:
```javascript
rooms.forEach(function(rid) {
	const now = Date.now();
	db.rocketchat_subscription.find({ rid }).count();
	print(rid + ": " + (Date.now() - now));
});

rooms.forEach(function(rid) {
	const now = Date.now();
	db.rocketchat_subscription.aggregate([
		{ $match: { rid } },
		{
			$lookup: {
				from: 'users',
				localField: 'u._id',
				foreignField: '_id',
				as: 'receiver',
			},
		},
		{ $match: { 'receiver.active': true } },
		{ $count: 'count' },
	]);
	print(rid + ": " + (Date.now() - now));
});
```

the results:

| id | members | `.count()` in ms | `.aggregate()` in ms |
| ---- | ----: | ----: | ----: |
| room1 | 38072 | 21 | 4191 |
| room2 | 36770 | 16 | 3833 |
| room3 | 10200 | 4 | 1102 |
| room4 | 3843 | 2 | 401 |
| room5 | 2225 | 2 | 245 |
| room6 | 1629 | 1 | 179 |
| room7 | 1454 | 1 | 160 |
| room8 | 1148 | 1 | 125 |
| room9 | 899 | 0 | 99 |
| room10 | 810 | 1 | 88 |

I also found out we already "archive" subscriptions of deactivated users, but for some reason we do that only for DMs (look at `Subscriptions.setArchivedByUsername`).. if we change that to all subscription types, we can filter for not archives.

sampaiodiego:
so, I remembered we have a property called `__rooms` in `IUser`, this fields stores every room ID the user is joined, except DMs.. since this field is indexed, it can be used for this purpose, no need to use `$lookup` then..

btw, my previous comment is still valid, this is just another solution that is easier to implement.. I think we should get rid of the `__rooms` field in the future.. in the meantime, we can use it for this.

LucianoPierdona:
> o

Query updated!

apps/meteor/app/models/server/models/Users.js
```diff
@@ -1645,6 +1645,10 @@ Find users to send a message by email if:
 			},
 		});
 	}
+
+	countRoomMembers(roomId) {
+		return this.find({ __rooms: roomId, active: true }).count();
```
gabriellsh, line 1650: shouldn't this be: `__rooms: { $in: [roomId] }`?

LucianoPierdona: For simple array of strings, it can be done like that:
https://www.mongodb.com/docs/manual/tutorial/query-arrays/

ggazzo (approved):

PR #25843, opened 2022-06-13
[NEW] Enable outbound calling for EE

amolghode1981:
## Proposed changes (including videos or screenshots)

Adding structure for supporting EE features for Voip client. This PR handles making outbound calls.

PR #25890, opened 2022-06-15
[NEW] EE Feature - Voip DialPad

tiagoevanp:
## Proposed changes (including videos or screenshots)

image

image

PR #25960, opened 2022-06-22
[NEW] Enable outbound calling for EE (#25843)

KevLehman:
* Clickup Task: https://app.clickup.com/t/22bmce2
Description: Creating EE structure for outbound dialing.
Refactored VoIPUser.ts and made necessary fields protected.
Created VoIPUser or EEVoipClient based on the licensing in SimpleVoipUser
Made necessary changes for using the voip client.

* Clickup Task: https://app.clickup.com/t/22bmce2
Description: Adding missing file.

* Clickup Task: https://app.clickup.com/t/22bmce2
Description: Connector changes for detecting the outbound call.

* Clickup Task: https://app.clickup.com/t/22bmce2
Description: Handling review comments.


murtaza98:
@KevLehman FYI: There are some conflicts here :grimacing:

murtaza98 (approved):
Reviewed and tested and everything looks good to me!

apps/meteor/client/sidebar/sections/components/OmnichannelMakeCall.tsx
```diff
@@ -0,0 +1,31 @@
...
+	const getColor = (): 'success' => 'success';
+
+	const voipCallIcon = {
+		title: getTooltip(),
+		color: getColor(),
+		icon: getIcon(),
+	} as const;
+
+	// TODO: move registration flow to context provider
+	const makeCall = useMutableCallback((): void => {
```
MartinSchoeler, line 23: Will we merge this PR without the `TODO` done? Or is this another "test" file? If so pls comment it

MartinSchoeler (dismissed):

apps/meteor/client/sidebar/sections/components/OmnichannelMakeCall.tsx
```diff
@@ -0,0 +1,31 @@
+import { Sidebar } from '@rocket.chat/fuselage';
+import { useMutableCallback } from '@rocket.chat/fuselage-hooks';
+import React, { ReactElement } from 'react';
+
+import { useOutboundDialer } from '../../../../ee/client/hooks/useOutboundDialer';
+
+export const OmnichannelMakeCall = (): ReactElement => {
```
aleksandernsilva, line 7: This component is not necessary as it will be replaced by the Dial Pad being implemented by @tiagoevanp

apps/meteor/client/sidebar/sections/OmnichannelSection.tsx
```diff
@@ -61,6 +64,7 @@ const OmnichannelSection = (props: typeof Box): ReactElement => {
 				{showOmnichannelQueueLink && <Sidebar.TopBar.Action icon='queue' title={t('Queue')} onClick={(): void => handleRoute('queue')} />}
 				{isCallEnabled && <OmnichannelCallToggle />}
 				<Sidebar.TopBar.Action {...availableIcon} onClick={handleAvailableStatusChange} />
+				{isCallEnabled && isCallReady && isEnterprise && <OmnichannelMakeCall />}
```
aleksandernsilva, line 67: Here you need to be careful because the hook `useHasLicence` returns `true | false | 'loading'`. 
Considering one of the values is a string (truthy) it might cause an issue here. One possible solution is to enforce the type: `const isEnterprise = useHasLicence('voip-enterprise') === true`

apps/meteor/client/providers/CallProvider/CallProvider.tsx
```diff
@@ -71,6 +78,57 @@ export const CallProvider: FC = ({ children }) => {
...
+				return '';
+			}
+			const { visitor } = await visitorEndpoint({
+				visitor: {
+					token: Random.id(),
+					phone: caller.callerId,
+					name: caller.callerName || caller.callerId,
+				},
+			});
+			const voipRoom = visitor && (await voipEndpoint({ token: visitor.token, agentId: user._id }));
```
aleksandernsilva, line 97: Is it possible that `visitor` is `null` or `undefined` here? If so it might be a problem for the following instructions

apps/meteor/ee/client/hooks/useOutboundDialer.ts
```diff
@@ -0,0 +1,40 @@
...
+import { EEVoipClient } from '../lib/voip/EEVoipClient';
+
+type UseOutboundDialerResult = {
+	outboundDialer?: EEVoipClient;
+	error?: Error | unknown;
+};
+
+const empty = {};
+
+export const useOutboundDialer = (): UseOutboundDialerResult => {
```
aleksandernsilva, line 16: This hook could be opmized by using another hook called `useHasLicence`. Ex.: 
```
const useOutboundDialer = (): UseOutboundDialerResult => {
    const isEE = useHasLicence('voip-enterprise');
    const voipClient = useCallClient()

    useEffect(() => () => voipClient?.clear())
    
    return isEE && voipClient
        ? { outboundDialer: voipClient } 
        : {}
}
```
The code above is just to give the idea and might not be complete.

aleksandernsilva: Something to also consider is whether or not is a good idea to restrict the access to the voipClient based on the licence. I feel like this restriction might make more sense elsewhere.

aleksandernsilva (changes requested):

apps/meteor/client/sidebar/sections/components/OmnichannelMakeCall.tsx
amolghode1981: @MartinSchoeler This file is a for demonstrating how to use makeCall function. Once the dialpad is in place, this file should be removed. I will add such comment to that file.

apps/meteor/client/sidebar/sections/OmnichannelSection.tsx
amolghode1981: Sure. I will fix this. Importantly, this line will be removed once the dialpad is wired. @tiagoevanp please take a note of this line.

apps/meteor/client/providers/CallProvider/CallProvider.tsx
amolghode1981: Fixed. Also remove the duplication of the same code from  CallContextValue.

apps/meteor/client/sidebar/sections/components/OmnichannelMakeCall.tsx
amolghode1981: Agreed. This is just to demonstrate the usage. Even the changes below will be removed once the dialpad is using this.

apps/meteor/ee/client/hooks/useOutboundDialer.ts
amolghode1981: @aleksandernsilva , the access to VoIP client is not restricted based on the license. If the system is in CE mode, https://github.com/RocketChat/Rocket.Chat/blob/870a3e0affb14abeb228e99e5f64cb6c5e37bc18/apps/meteor/client/lib/voip/SimpleVoipUser.ts#L29 would create the object of parent VoIPUser. VoIPUser does not provide outbound calling.  All the components common to EE and CE will use hook 'useVoipClient' to get the access to voipClient to receive calls.

Dialpad is suppose to use useOutboundDialer for getting access to EEVoipClient. As mentioned above this gets created in SimpleVoipUser when there is a proper license. Even though useOutboundDialer will be used by EE component, there is a possibility that it might get used by CE component (Either by mistake or with a hacker mind.)
In that case it may make sense to add a license check here. Wdyt?

amolghode1981: > 

@aleksandernsilva I do not find useHasLicence. So could not change this part of the code.

tiagoevanp (approved):

tiagoevanp:
https://github.com/RocketChat/Rocket.Chat/pull/25890 needs this one to be tested by QA!

tiagoevanp (approved):

ggazzo (changes requested):
I think would make more sense move useVoipClient to ee folder

tiagoevanp (approved):

PR #25963, opened 2022-06-22
Chore: Small fix on callProvider

tiagoevanp:


aleksandernsilva (approved):

apps/meteor/client/providers/CallProvider/CallProvider.tsx
```diff
@@ -54,10 +57,29 @@ export const CallProvider: FC = ({ children }) => {
...
+		async (data): Promise<void> => {
+			roomInfo &&
+				(await voipCloseRoomEndpoint({
+					rid: roomInfo.rid,
+					token: roomInfo.v.token || '',
+					options: { comment: data?.comment, tags: data?.tags },
+				}));
+			homeRoute.push({});
+
+			const queueAggregator = result?.voipClient?.getAggregator();
```
ggazzo, line 72: there is no need to `?` for `result?`and `data?` or am I wrong?

ggazzo (changes requested):

apps/meteor/client/providers/CallProvider/CallProvider.tsx
tiagoevanp: on result you are right! data is a different approach... Let me upgrade this PR, better things will come!

apps/meteor/client/lib/voip/VoIPUser.ts
```diff
@@ -750,7 +750,7 @@ export class VoIPUser extends Emitter<VoipEvents> {
 		return this.queueInfo;
 	}
 
-	getRegistrarState(): string | undefined {
+	getRegisterState(): string | undefined {
```
ggazzo, line 753: [Registrar](https://www.voip-info.org/sip-registrar-server/)

ggazzo (changes requested):

ggazzo (approved):

ggazzo (dismissed):

PR #25985, opened 2022-06-23
[FIX] Default BH not getting applied in-case any other BH is disabled

murtaza98:
## Steps to test or reproduce

Pre-requisite: 
- Have Multiple Business hours enabled
- Have exactly 2 BH setups and one of them should be linked to a department which has an agent

To reproduce this issue:
1. Login as an agent which is connected to a department (lets say Sales) which is again connected to a BH (let's call it ET)
2. Modify the Default Business hour time to make it active and make the other BH inactive
3. Try to enable the agent now and you should see an error saying that the BH is closed which is expected since this Sales department has BH closed right now (we closed it in step 2)
4. Now disable this other BH. And try to make this agent available again. 

Actual behaviour:
Agent is getting error saying BH is closed

Expected behaviour:
Agent should be able to make itself available since the default BH should be applied to him/her - considering the other BH (ET) is disabled

## Further comments

Clickup: https://app.clickup.com/t/2gabfqd


murtaza98:
Closing this in favour of https://github.com/RocketChat/Rocket.Chat/pull/26471 since there was some issue with the QA env of this PR

PR #25988, opened 2022-06-23
Regression: Add appId prop to slashcommand

tapiarafael:

Pass the appId when present to the slashcommand array. This avoid problems with contextual bar and modals not opening.


AllanPazRibeiro (dismissed):

apps/meteor/client/startup/slashCommands.ts
```diff
@@ -11,7 +11,7 @@ Tracker.autorun(async () => {
 	if (oldUserId === null && newUserId) {
 		APIClient.get('/v1/commands.list').then((result) => {
 			result.commands.forEach((command) => {
-				slashCommands.add(command.command);
+				slashCommands.add(command.command, undefined, undefined, undefined, undefined, undefined, undefined, command.appId);
```
d-gubert, line 14: Would it be possible to change the signature to accept an object instead of multiple parameters? This is looking weird, honestly 🙈

d-gubert (changes requested):

apps/meteor/client/startup/slashCommands.ts
tapiarafael: Yeah totally agree with you, it would require a big change to refactor all files that use the `slashCommand.add`, but I think that is worth it.

apps/meteor/app/utils/lib/slashCommand.ts
```diff
@@ -30,6 +42,7 @@ export const slashCommands = {
 			providesPreview: Boolean(providesPreview),
 			previewer,
 			previewCallback,
+			appId,
```
ggazzo, line 45: I didnt get why not only add this change here, why the whole 'refactor' was required?

ggazzo (dismissed):

apps/meteor/app/utils/lib/slashCommand.ts
ggazzo: I think you misunderstood my point.
I have nothing against passing `appId` as a property of options.

But as your pr is marked as a Regression(and before as fix), I couldn't understand how this changes helped to fix the problem introduced. now that you explained that it has nothing to do with it the issue, it was just aesthetic, I can ignore this part

PR #26002, opened 2022-06-24
[BREAK] Remove show message in main thread preference

dougfabris:


ggazzo (approved):

TBG-FR:
You mixed up two different use-cases... While people can understand the behaviour of sending some (first, and maybe important other) thread messages to the main chat, for some communities, some people would like to see all thread messages in main chat, wether or not the box was ticked by the OP when the message was send

Bring back that feature, the two can live together !

dougfabris:
@aceArt-GmbH @TBG-FR @rumato163 After receiving those requests about the preference removal we brought the complaints to the product team and we're planning to bring back the preference on our next release: #29302. 
Thanks for your contributions!

PR #26008, opened 2022-06-24
Regression: Webhook Integration Creation + string error toast msg

dudanogueira:
## Issue(s)
While creating an webhook integration, it was missing the **v1/**. Also, the toast was not properly displaying error messages

## Before:

image

image

## After:

image

## Steps to test or reproduce
Try creating an Integration Webhook in Admin > Integrations. 
The payload will be sent to: http://localhost:3000/api/integrations.create

Edit a user, and set it's email the same email as an existing user: the toast message will not display the correct message

### DIFF
--- a/apps/meteor/client/hooks/useEndpointAction.ts
+++ b/apps/meteor/client/hooks/useEndpointAction.ts
@@ -24,7 +24,7 @@ export const useEndpointAction = <TMethod extends Method, TPath extends PathFor<
 
 			return data;
 		} catch (error) {
-			dispatchToastMessage({ type: 'error', message: String(error) });
+			dispatchToastMessage({ type: 'error', message: String(error.error) });
 			throw error;
 		}
 	}, [dispatchToastMessage, params, sendData, successMessage]);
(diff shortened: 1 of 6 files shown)


dudanogueira:
Found more paces without the "/v1/", being:
users.2fa.enableEmail
and integrations.remove

apps/meteor/client/hooks/useEndpointAction.ts
```diff
@@ -24,7 +24,7 @@ export const useEndpointAction = <TMethod extends Method, TPath extends PathFor<
 
 			return data;
 		} catch (error) {
-			dispatchToastMessage({ type: 'error', message: String(error) });
+			dispatchToastMessage({ type: 'error', message: String(error.error) });
```
debdutdeb, line 27: idk why we need to use `String` here :eyes:

debdutdeb: Should be handled here https://github.com/RocketChat/Rocket.Chat/blob/0c7034f1bb0e77eb058fcf459c8b566baf94debc/apps/meteor/client/lib/utils/handleError.ts#L27

debdutdeb: @dudanogueira ping to check this 😬 would be good to ship with 5.x :)

debdutdeb:
```suggestion
			dispatchToastMessage({ type: 'error', message: error });
```

debdutdeb (approved):

ggazzo (approved):

Issue #25775, opened 2022-06-06
Omnichannel Analytics page error - Chart is not a constructor


jarismar:
### Description:

We recently upgraded to 4.8.0 (from 4.7.1) and noticed that Omnichannel Analytics page does not finish loading.

After investigation it seems chart.js dependency was upgraded which is causing this error, I believe the newer version of chartJS no longer defines a default export.

### Steps to reproduce:

1. Just visit omnichannel analytics page notice chart won't render and there is a Undefined message displayed

### Expected behavior:

Omnichannel analytics page should load without errors.

### Actual behavior:

![Omnichannel_Analytics_page_error](https://user-images.githubusercontent.com/1980614/172216323-b00262ee-f307-4ef4-8156-351c73d474e8.jpeg)

### Server Setup Information:
```
+--------------------------------------------+ 
|               SERVER RUNNING               | 
+--------------------------------------------+ 
|                                            | 
|  Rocket.Chat Version: 4.8.0                | 
|       NodeJS Version: 14.18.3 - x64        | 
|      MongoDB Version: 5.0.8                | 
|       MongoDB Engine: wiredTiger           | 
|             Platform: linux                | 
|         Process Port: 3000                 | 
|             Site URL: https://localhost  | 
|     ReplicaSet OpLog: Enabled              | 
|          Commit Hash: 0e38639928           | 
|        Commit Branch: HEAD                 | 
|                                            | 
+--------------------------------------------+ 
```

### Client Setup Information

- Desktop App or Browser Version: Version 102.0.5005.61 (Official Build) (x86_64)
- Operating System: Mac OSX

### Additional context

I think the dynamic import for chat.js needs to change from:

```
	const chartImport = await import('chart.js');
	const Chart = chartImport.default; // <=== error here chartImport.default is undefined
```

to

```
	const { Chart } = await import('chart.js');
```

ddkedr:
Still no response from RC team...

PR #26022, opened 2022-06-27
[FIX] Fixes Omnichannel Analytics page error (default export of chart.js was undefined)

belmeopmenieuwesim:
## Issue(s)

#25775

### DIFF
--- a/apps/meteor/app/livechat/client/lib/chartHandler.ts
+++ b/apps/meteor/app/livechat/client/lib/chartHandler.ts
@@ -151,8 +151,7 @@ export const drawLineChart = async (
 			fill: false,
 		});
 	});
-	const chartjs = await import('chart.js/auto');
-	const Chart = chartjs.default;
+	const { Chart } = await import('chart.js/auto');
 	return new Chart(chart, {
 		type: 'line',
 		data: {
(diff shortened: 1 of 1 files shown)


belmeopmenieuwesim:
The bug is still present on that PR. We need to get rid of using .default to grab the Chart export.

KevLehman:
maybe you need to install the deps again? how are you testing it? 👀 on develop, seems to be working fine, but let me know how you're testing it so we can work on a fix if needed
image
image

debdutdeb:
Technically should work 👀 looking at the docs chats.js/auto exports Chart as default not as named. Therefore this pr should not be working;

KevLehman:
yeah, per docs, you can do the import like ` { Chart } from 'chart.js'` when you're not using `/auto`. That means you have to add extra config to generate the graph, but since we were using already "light" config, auto was the choice.
And `/auto` has a `default` export for creating charts 🤗 

@belmeopmenieuwesim , let us know after you tested, and we can close this one if it works 🥳

belmeopmenieuwesim:
Alright awesome, seems to be fine indeed using `chart.js/auto`. Hopefully a new release is being pushed soon with the current fix in `develop` :D

PR #26024, opened 2022-06-27
Regression: Fix voip call wrap-up model not working

murtaza98:
## Issue(s)

image

### DIFF
--- a/apps/meteor/ee/app/livechat-enterprise/server/api/lib/tags.js
+++ b/apps/meteor/ee/app/livechat-enterprise/server/api/lib/tags.js
@@ -1,5 +1,5 @@
 import { escapeRegExp } from '@rocket.chat/string-helpers';
-import LivechatTag from '@rocket.chat/models';
+import { LivechatTag } from '@rocket.chat/models';
 
 import { hasPermissionAsync } from '../../../../../../app/authorization/server/functions/hasPermission';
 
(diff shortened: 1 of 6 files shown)


KevLehman (dismissed):

KevLehman (dismissed):

apps/meteor/ee/app/livechat-enterprise/server/api/lib/tags.js
```diff
@@ -1,5 +1,5 @@
 import { escapeRegExp } from '@rocket.chat/string-helpers';
-import LivechatTag from '@rocket.chat/models';
+import { LivechatTag } from '@rocket.chat/models';
```
ggazzo, line 2: What do you think about convert to TS ? this way this is never going to happen again

murtaza98: Yeah sounds good! Converted it to typescript. Please check

KevLehman (dismissed):

murtaza98:
Seems like some tests are breaking :see_no_evil: Let me check until then making it draft

murtaza98:
> Seems like some tests are breaking see_no_evil Let me check until then making it draft

Fixed those build! This should be ready for review now

KevLehman (approved):

PR #26037, opened 2022-06-27
Chore: Device management search with regex

albuquerquefabio:


apps/meteor/app/models/server/raw/Sessions.ts
```diff
@@ -985,31 +985,13 @@ export class SessionsRaw extends BaseRaw<ISession> {
 		return [
 			{ key: { createdAt: -1 } },
 			{ key: { loginAt: -1 } },
+			{ key: { searchTerm: 1 }, background: true },
```
debdutdeb, line 988: do we need this? since we're using regex for the query it won't be using the index any way.

debdutdeb (changes requested):

apps/meteor/app/models/server/raw/Sessions.ts
albuquerquefabio: In this case, we use it, because we are searching for just one object in the document. The problem is, when we use the operator `$or`.

debdutdeb: how about we make it a partial index then? `{ partialFilterExpression: { searchTerm: { $exists: true } }` since it's an optional field? 👀

apps/meteor/app/statistics/server/lib/SAUMonitor.ts
```diff
@@ -185,6 +187,26 @@ export class SAUMonitorClass {
 		// TODO missing an action to perform on dangling sessions (for example remove sessions not closed one month ago)
 	}
 
+	private async _getSearchTerm(session: Omit<ISession, '_id' | '_updatedAt' | 'createdAt'>): Promise<string | void> {
```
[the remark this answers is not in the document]
debdutdeb: It returns early so needs to have `void` there.

apps/meteor/app/statistics/server/lib/SAUMonitor.ts
```diff
@@ -185,6 +189,26 @@ export class SAUMonitorClass {
 		// TODO missing an action to perform on dangling sessions (for example remove sessions not closed one month ago)
 	}
 
+	private async _getSearchTerm(session: Omit<ISession, '_id' | '_updatedAt' | 'createdAt'>): Promise<string | undefined> {
+		const user = await Users.findOneById<IUser>(session.userId, {
```
sampaiodiego, line 193: this generic is useful to specify only the fields returned by the project, in this case:
```suggestion
		const user = await Users.findOneById<Pick<IUser, 'username' | 'name'>>(session.userId, {
```

apps/meteor/app/statistics/server/lib/SAUMonitor.ts
```diff
@@ -185,6 +189,26 @@ export class SAUMonitorClass {
...
+			projection: {
+				_id: 1,
```
sampaiodiego, line 195: `_id` is always returned, you don't need to specify
```suggestion
```

apps/meteor/app/statistics/server/lib/SAUMonitor.ts
```diff
@@ -185,6 +189,26 @@ export class SAUMonitorClass {
...
+			},
+		});
+
+		if (!user) {
+			return;
+		}
+
+		const searchTerm = `${session.device?.name || ''}${session.device?.type || ''}${session.device?.os.name || ''}${session.sessionId}${
+			user.name || ''
+		}${user.username || ''}`;
```
sampaiodiego, line 207: not sure if this is a concern or not, but storing the name and username like this will make it impossible to update in case the user changes his name or username, is that an issue?

let's my name when I registered was `Diego Sampaio`, so the `searchTerm` will have this information.. if later on I change my name to `Didi`, when searching for "didi" I'll find only the new records and if I search for "sampaio" only the old ones will be found.

if this limitation is not an issue, maybe we should just add a notice on the regarding it, so people don't get confused when try to find for something and the results are not exactly what they're expecting.

apps/meteor/server/models/raw/Sessions.ts
```diff
@@ -991,31 +991,13 @@ export class SessionsRaw extends BaseRaw<ISession> implements ISessionsModel {
 		return [
 			{ key: { createdAt: -1 } },
 			{ key: { loginAt: -1 } },
+			{ key: { searchTerm: 1 }, partialFilterExpression: { searchTerm: { $exists: true } }, background: true },
```
sampaiodiego, line 994: not sure what is the impact of `partialFilterExpression` here, since it mention the same field of the index.

sampaiodiego (changes requested):

d-gubert:
Please update your target branch with `develop` (or change the target branch to `develop`) since some changes in your PR have already been merged (the app settings one, for example)

apps/meteor/server/models/raw/Sessions.ts
albuquerquefabio: `searchTerm` can be undefined on old documents. In this case, `partialFilterExpression` just will add to the index if the object exists.

ggazzo (changes requested):

apps/meteor/server/startup/migrations/v272.ts
```diff
@@ -1,14 +1,12 @@
...
-		await Settings.deleteMany({
-			_id: {
-				$in: ['VoIP_Server_Host', 'VoIP_Server_Websocket_Port'],
-			},
-		});
+		await Promise.allSettled(
+			['instanceId_1_sessionId_1_year_1_month_1_day_1', 'instanceId_1_sessionId_1', 'type_1'].map((idx) => Sessions.col.dropIndex(idx)),
+		);
 	},
 });
```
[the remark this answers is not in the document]
albuquerquefabio: Fixed!

sampaiodiego:
already merged into #25791

PR #26042, opened 2022-06-28
Chore: move omni-jobs to services

murtaza98:

The goal here is to refactor 2 of our Omnichannel jobs (both of which run in EE) & move them inside an internal service. This should solve some issues with previous implementations where jobs were not running when a server gets restarted. 

Some other modifications in this PR also include
- Modified the scheduler to use a single generic collection 
- Improved transfer message
- Add E2E tests to test both the jobs

## Further comments

- [ ] Create migration to drop un-used collections
- [x] Improve transfer messages while transferring chat due to inactivity
- [x] Add e2e tests to cover both auto-onhold-chat-closed & auto-transfer-unanswered-chat feature


apps/meteor/ee/app/livechat-enterprise/server/jobs/AbstractOmniSchedulerClass.ts
```diff
@@ -0,0 +1,44 @@
...
+import { schedulerLogger } from '../lib/logger';
+
+const SCHEDULER_NAME = 'omnichannel_scheduler';
+
+export abstract class AbstractOmniSchedulerClass {
+	scheduler: Agenda;
+
+	abstract createJobDefinition(): void;
+
+	abstract createIndexes(collection: Collection): void;
```
KevLehman, line 14: What if we narrow down what this function can do? I think we can implement createIndexes right here and receive an array/object with the indexes to be created and options. That way, the function can only do what we want it to do :eyes:

apps/meteor/ee/app/livechat-enterprise/server/jobs/AbstractOmniSchedulerClass.ts
```diff
@@ -0,0 +1,44 @@
...
+			this.scheduler.stop();
+		});
+		process.on('SIGTERM', () => {
+			schedulerLogger.info(`SIGTERM received. Stopping ${SCHEDULER_NAME}...`);
+			this.scheduler.stop();
+		});
+
+		this.createJobDefinition();
+
+		this.createIndexes(db.collection(SCHEDULER_NAME));
```
KevLehman, line 42: Can we convert db.collection to a local property? That way if we ever need to extend this, we don't have to change this part :eyes: (by extend i mean: doing other things with the agenda collection)

apps/meteor/ee/app/livechat-enterprise/server/jobs/AutoCloseOnHoldScheduler.ts
```diff
@@ -0,0 +1,108 @@
...
+import moment from 'moment';
+
+import { Livechat } from '../../../../../app/livechat/server';
+import { schedulerLogger } from '../lib/logger';
+import { AbstractOmniSchedulerClass } from './AbstractOmniSchedulerClass';
+
+const JOB_NAME = 'omnichannel_auto_close_on_hold_scheduler';
+
+type JobData = {
+	roomId: string;
```
KevLehman, line 14:
```suggestion
	roomId: IRoom['id'];
```

apps/meteor/ee/app/livechat-enterprise/server/lib/AutoTransferChatScheduler.ts
```diff
@@ -25,30 +31,52 @@ class AutoTransferChatSchedulerClass {
 		this.scheduler = new Agenda({
 			mongo: (MongoInternals.defaultRemoteCollectionDriver().mongo as any).client.db(),
 			db: { collection: SCHEDULER_NAME },
+			defaultLockLifetime: 1000, // 1 minute
```
KevLehman, line 34: [not for this line]
Why this class doesn't extend from the abstract one? :eyes:

apps/meteor/ee/app/livechat-enterprise/server/jobs/AbstractOmniSchedulerClass.ts
```diff
@@ -0,0 +1,44 @@
+import Agenda from 'agenda';
+import { MongoInternals } from 'meteor/mongo';
+import { Collection } from 'mongodb';
+
+import { schedulerLogger } from '../lib/logger';
```
KevLehman, line 5: Can we convert the logger in part of the class? That way subclasses can just use "this.debug" and so on for logging

apps/meteor/ee/app/livechat-enterprise/server/jobs/AutoTransferChatScheduler.ts
```diff
@@ -0,0 +1,138 @@
...
+				userId: agent.agentId,
+				transferredBy: await this.getSchedularUser(),
+				transferredTo: agent,
+			});
+		}
+
+		return false;
+	}
+
+	private async executeJob({ attrs: { data } }: any = {}): Promise<void> {
```
KevLehman, line 121: You can type this any :eyes: 
I remember agenda provides the Job definition or JobData, and you can pass a generic to that type to get typed job attributes

KevLehman (changes requested):

apps/meteor/tests/e2e/page-objects/fragments/home-sidenav.ts
```diff
@@ -41,6 +43,27 @@ export class HomeSidenav {
...
+		const toggleButton = this.page.locator('#omnichannel-status-toggle');
+		expect(toggleButton).toBeVisible();
+
+		const currentStatus: 'Available' | 'Not Available' = (await toggleButton.getAttribute('title')) as any;
+		if (status === 'offline') {
+			if (currentStatus === 'Available') {
+				await toggleButton.click();
+			}
+		} else if (currentStatus === 'Not Available') {
+			console.log('Switching omnichannel status to online');
```
murtaza98, line 57:
```suggestion
```

apps/meteor/ee/server/local-services/omnichannel/OmniEEService.ts
```diff
@@ -0,0 +1,175 @@
...
+import moment from 'moment';
+import { TAPi18n } from 'meteor/rocketchat:tap-i18n';
+
+import { Livechat } from '../../../../app/livechat/server';
+import { forwardRoomToAgent } from '../../../../app/livechat/server/lib/Helper';
+import { RoutingManager } from '../../../../app/livechat/server/lib/RoutingManager';
+import { settings } from '../../../../app/settings/server';
+import { ServiceClassInternal } from '../../../../server/sdk/types/ServiceClass';
+import { schedulerLogger } from '../../../app/livechat-enterprise/server/lib/logger';
+import { OmniJobSchedularService } from '../../sdk';
```
KevLehman, line 13:
```suggestion
import { OmniJobSchedulerService } from '../../sdk';
```
👀

apps/meteor/ee/server/local-services/omnichannel/OmniEEService.ts
```diff
@@ -0,0 +1,175 @@
...
+	private logger = schedulerLogger;
+
+	private systemUser: IUser;
+
+	constructor() {
+		super();
+		this.logger.debug('OmniEEService started');
+	}
+
+	async monitorOnHoldRoomForAutoClose(roomId: string, timeout: number, comment: string): Promise<void> {
```
KevLehman, line 38: What if we have a "sub-service" for holding this logic? And then on constructor of this service, we initialize it. 
So, you can use those things in the form of "OmniEEService.scheduling.monitorxxxx" 🤔 

(It just passed my mind, but, would be better to keep things "separated")

apps/meteor/ee/server/local-services/omnichannel/OmniEEService.ts
```diff
@@ -0,0 +1,175 @@
...
+	}
+
+	async monitorUnansweredRoomForAutoTransfer(roomId: string, timeout: number): Promise<void> {
+		this.logger.debug(`monitorUnansweredRoomForAutoTransfer: ${roomId}`);
+		await this.cancelMonitorUnansweredRoomForAutoTransfer(roomId);
+
+		const when = moment(new Date()).add(timeout, 's').toDate();
+
+		const [job] = await Promise.all([
+			OmniJobSchedularService.scheduleJobAt<OmniAutoTransferUnansweredChatJobData>(OMNI_JOB_NAME.AUTO_TRANSFER_UNANSWERED_CHAT, when, {
```
KevLehman, line 88: If you're passsing the type as a param, why do you need the type coercion?

apps/meteor/ee/server/local-services/omnichannel/OmniEEService.ts
```diff
@@ -0,0 +1,175 @@
...
+
+	async cancelMonitorUnansweredRoomForAutoTransfer(roomId: string): Promise<void> {
+		this.logger.debug(`cancelMonitorUnansweredRoomForAutoTransfer: ${roomId}`);
+
+		const [, totalCancelledJobs] = await Promise.all([
+			LivechatRooms.unsetAutoTransferOngoingById(roomId),
+			OmniJobSchedularService.cancelJobByRoomId(OMNI_JOB_NAME.AUTO_TRANSFER_UNANSWERED_CHAT, roomId),
+		]);
+
+		this.logger.debug(
```
KevLehman, line 105: Some of those logs should be on the "OmniJobSchedulerService"

apps/meteor/ee/server/local-services/omnichannel/OmniJobSchedularService.ts
```diff
@@ -0,0 +1,93 @@
...
+	scheduler: Agenda;
+
+	logger = schedulerLogger;
+
+	db: Db;
+
+	constructor() {
+		super();
+
+		const { db } = MongoInternals.defaultRemoteCollectionDriver().mongo;
```
KevLehman, line 30: A service can receive the DB as param 👀 (like, Authorization service iirc)

KevLehman (dismissed):

apps/meteor/ee/server/local-services/omnichannel/OmniEEService.ts
murtaza98: You're right, it wasn't necessary. I've removed it now

apps/meteor/ee/server/local-services/omnichannel/OmniEEService.ts
murtaza98: Makes sense :+1: I've moved all the relevant scheduling & un-scheduling logs inside the scheduling service itself

apps/meteor/ee/server/local-services/omnichannel/OmniJobSchedularService.ts
murtaza98: Done :+1:

apps/meteor/ee/server/local-services/omnichannel/OmniEEService.ts
murtaza98: I like this idea and I tried to implement it here. Overall it went good but there is just a small issue with SDK due to which the methods using the new service SDK have to apply 2 promises like this. 
`await (await OmniEEService.getSchedulingSubService()).monitorUnansweredRoomForAutoTransfer(rid, autoTransferTimeout);`

One promise to get the inner class & then the other promise to get actually execute the method. This is happening because the proxy SDK used for services is always async - so all the methods contained within the service will be async.

I also tried to assign a public variable so we can access it as you mentioned above - `OmniEEService.scheduling.monitorxxx`  - however even this isn't working & it also requires an async operation.

@KevLehman  Let me know if you have any suggestions on how to improve this or if you'd rather prefer to revert this change & avoid sub-service in order to avoid multiple promises

KevLehman: Yeah, probably that's because 🤔 of how it's being done. 
What if we change scheduling service to serve purely as scheduling? So instead of a "sub service", create a separate service for "scheduling" things, without any extra logic, just agenda interaction. Like

```
ScheduleService.scheduleAction('OmniEEfdafddafd', () => {});
```

So, it's called from inside the service (OmniEEService) and all the Omni-logic for scheduling is managed there 👀 . That way,  you avoid the await await thing 👀 wdyt?

murtaza98: Isn't this what `OmniJobSchedulerService` does? :thinking:  It only has interactions with agenda lib & `OmniEEService` uses it to just schedule jobs

murtaza98: JFYI: I've moved back the sub-service-related logic :leftwards_arrow_with_hook:

KevLehman: If that's what it does, I'll suggest renaming the service 👀 so it's generic and can be used in other places
Instead of "OmniJob..." just "Scheduler" so ppl don't think it's omni specifici

Issue #25728, opened 2022-06-02
GET users.list broken for "normal" users


mb987654:
### Description:

When trying to perform a GET against /api/v1/users.list with no query parameters, the following response is returned for users who do not have a role that includes the "View Full Other User Info" permission:

{
    "success": false,
    "error": "Invalid attribute: emails.address [error-invalid-query]",
    "errorType": "error-invalid-query"
}

If a user has a role that includes the "View Full Other User Info" permission, then the error does not occur and a list of users is returned as expected.  

### Steps to reproduce:

1. perform a GET against /api/v1/users.list as a user that has no roles that include the "View Full Other User Info" permission

### Expected behavior:

a list of users is returned

### Actual behavior:

an error is returned:

{
    "success": false,
    "error": "Invalid attribute: emails.address [error-invalid-query]",
    "errorType": "error-invalid-query"
}

I did not have this problem with 4.7.0.

### Server Setup Information:

- Version of Rocket.Chat Server: 4.8.0
- Operating System: CentOS
- Deployment Method: docker-compose
- Number of Running Instances: 1
- DB Replicaset Oplog: Enabled
- NodeJS Version: 14.18.3 - x64
- MongoDB Version: 4.0.28

Gummikavalier:
RC 4.7.4 seems to suffer from this too.

gelpiu-developers:
Have anyone solved this? I'm having the exact same problem and I don't find anything in the documentation explaining why this is nos working now...

dbigham:
Any updates?

PR #26050, opened 2022-06-28
[FIX] Users without the `view-other-user-info` permission can't use the `users.list` endpoint

LucianoPierdona:

This PR fix the query when a normal users access `users.list`

## Issue(s)

Closes: #25728


ggazzo:
now this is a break change

apps/meteor/app/api/server/lib/users.ts
```diff
@@ -111,6 +112,10 @@ export function getNonEmptyQuery(query: Query): typeof _defaultQuery | (typeof _
 		],
 	};
 
+	if (!canSeeAllUserInfo) {
+		defaultQuery.$or = defaultQuery.$or.filter((or) => !or['emails.address']);
+	}
+
```
debdutdeb, line 118: Can you invert the logic here please? easier to read `if hasPermission then push elem to array` than `if not hasPermission then create new array by filtering out xyz field` kinda over-wording this but I hope you're getting the point 🙈

debdutdeb (changes requested):
The pr description is saying users need the `view-user-administration` permission, not seeing the check there (in the route options you can add `permissionsRequired: ['view-user-administration']`.

apps/meteor/app/api/server/lib/users.ts
LucianoPierdona: updated!

LucianoPierdona:
> The pr description is saying users need the `view-user-administration` permission, not seeing the check there (in the route options you can add `permissionsRequired: ['view-user-administration']`.

@debdutdeb my bad, I forgot to update the description but this permission is not needed anymore, looks like the endpoint is going to be available for all users. @matheusbsilva137 can confirm.

debdutdeb (approved):

nishant23122000:
It's working fine.

ggazzo (approved):

Issue #25954, opened 2022-06-22
Omnichannel can´t close from Agent


charlyrock:
### Description:

With an open omnichannel, it was previously possible to end the chat as an agent in the bar where the name is.
This is now no longer possible in the new version and you get the error message "undefined". 
This is of course a serious problem. If the requester closes the browser window without closing the omnichannel chat, the chat remains permanently open for the agent. The agent has no way to end the chat. 

### Steps to reproduce:

1. Open Omnichannel, write a message
2. Agent try to close the omnichannel
3. Problem: undefined 

### Expected behavior:

The Omnichannel will be closed, if the agent close the chat session.

### Actual behavior:

Agent can´t close the omnichannel chat to customer. 

### Server Setup Information:

- Version of Rocket.Chat Server: 4.8.1
- Operating System: Ubuntu
- Deployment Method: 
- Number of Running Instances: 1
- DB Replicaset Oplog: 
- NodeJS Version: 14.18.3
- MongoDB Version: 4.2.17

### Client Setup Information

- Desktop App or Browser Version: Desktop App, Chrome Browser
- Operating System: Windows 11

### Relevant logs:

![Screenshot 2022-06-22 085510](https://user-images.githubusercontent.com/67680741/174963868-6db381fa-d2da-4f43-a4dd-941911b83495.jpg)

didacroyo:
Hi, same problem. Downgraded from 4.8.1. to 4.8.0 and the problem is still there
In our previous version 4.4.2 there was no problem closing omnichannel sessions

SolidairesFinancesPubliques:
Hi, same problem here since 4.8.1.

saschafoerster:
Same here, we are on 4.8.1, docker installation.

saschafoerster:
@didacroyo Where do I need to change the settings? Thanks for finding out!

didacroyo:
@saschafoerster 

Admin > Settings > Omnichannel

![image](https://user-images.githubusercontent.com/35463401/176741958-cdea0bef-5554-4d09-895a-59e1fa67ebf7.png)

saschafoerster:
@didacroyo The workaround worked. 😅

PR #26057, opened 2022-06-29
[FIX] Unable to close chats when comments is disabled

murtaza98:

Fixes https://github.com/RocketChat/Rocket.Chat/issues/25954

### DIFF
--- a/apps/meteor/client/components/Omnichannel/modals/CloseChatModal.tsx
+++ b/apps/meteor/client/components/Omnichannel/modals/CloseChatModal.tsx
@@ -61,6 +61,10 @@ const CloseChatModal = ({
 		return cannotSendTag || cannotSendComment;
 	}, [comment, commentRequired, errors, tagRequired, tags]);
 
+	const onSubmitWithoutCommentsAndTags = useCallback((): void => {
+		onConfirm();
+	}, [onConfirm]);
+
 	useEffect(() => {
 		if (department?.requestTagBeforeClosingChat) {
 			setTagRequired(true);
@@ -113,7 +117,7 @@ const CloseChatModal = ({
 		<GenericModal
 			variant='warning'
 			title={t('Are_you_sure_you_want_to_close_this_chat')}
-			onConfirm={onConfirm}
+			onConfirm={onSubmitWithoutCommentsAndTags}
 			onCancel={onCancel}
 			onClose={onCancel}
 			confirmText={t('Confirm')}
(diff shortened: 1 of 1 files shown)


KevLehman:
This is not a regression per se, but this would be a really annoying issue 😬 to keep on 5.0. 

@ggazzo , agree on put this as "regression" (it will go through normal QA anyways, but since the fix is pretty self contained, we should be good)

murtaza98:
Yeah, I agree @KevLehman. And considering the release candidate deadline was extended we should have some time to get it reviewed and qa'd.

apps/meteor/client/components/Omnichannel/modals/CloseChatModal.tsx
```diff
@@ -113,7 +117,7 @@ const CloseChatModal = ({
 		<GenericModal
 			variant='warning'
 			title={t('Are_you_sure_you_want_to_close_this_chat')}
-			onConfirm={onConfirm}
+			onConfirm={onSubmitWithoutCommentsAndTags}
```
MartinSchoeler, line 120:
```suggestion
			onConfirm={(): void => onConfirm()}
```

apps/meteor/client/components/Omnichannel/modals/CloseChatModal.tsx
```diff
@@ -61,6 +61,10 @@ const CloseChatModal = ({
 		return cannotSendTag || cannotSendComment;
 	}, [comment, commentRequired, errors, tagRequired, tags]);
 
+	const onSubmitWithoutCommentsAndTags = useCallback((): void => {
+		onConfirm();
+	}, [onConfirm]);
+
```
MartinSchoeler, line 67:
```suggestion
	
```

MartinSchoeler (changes requested):
No need to add a new callback just for this, I think it would be better to do it inline

apps/meteor/client/components/Omnichannel/modals/CloseChatModal.tsx
```diff
@@ -61,6 +61,7 @@ const CloseChatModal = ({
 		return cannotSendTag || cannotSendComment;
 	}, [comment, commentRequired, errors, tagRequired, tags]);
 
+
```
murtaza98, line 64:
```suggestion
```

murtaza98:
> No need to add a new callback just for this, I think it would be better to do it inline

Done :heavy_check_mark: , please check

aleksandernsilva (approved):

PR #26068, opened 2022-06-29
[FIX] Remove `undefined` and `null` fields from `Messages` and improve type narrowing

LucianoPierdona:

This PR creates a new migration to remove `undefined` and `null` fields from `rocketchat_message`


apps/meteor/server/startup/migrations/v275.ts
```diff
@@ -0,0 +1,25 @@
+import { Messages } from '@rocket.chat/models';
+
+import { addMigration } from '../../lib/migrations';
+
+addMigration({
+	version: 275,
+	async up() {
+		const fieldsToUpdate = ['editedBy', 'editedAt', 'emoji', 'avatar', 'alias', 'customFields', 'groupable', 'attachments', 'reactions'];
+
+		fieldsToUpdate.map(async (field) => {
```
KevLehman, line 10: This could be a ton of queries being issues per field per message. Not sure how many of messages with this condition will be there. have you considered using `bulkWrite`?
Like `apps/meteor/server/startup/migrations/v260.ts`

KevLehman (changes requested):

apps/meteor/server/startup/migrations/v275.ts
LucianoPierdona: @KevLehman updated!

PR #26073, opened 2022-06-30
[Fix] Link not scrolling to message

filipemarins:
## Further comments

Fix: link message was not opening the correct message
Fix: key prop on map element (warning)


gabriellsh (changes requested):
There are some occasions that still don't go to the correct message.

To test this, I created a channel and flooded it with a bunch of messages, got one from the middle an quoted it. Then, refresh the page, wait for it to load and click at the time inside the quote. After the refresh, the first click in the quote always fails to scroll to the correct message.

gabriellsh (dismissed):
It's working now. Will release it for QA to test.

gabriellsh (approved):

PR #26084, opened 2022-06-30
Chore: Allow endpoints to optionally require authentication

pierre-lehnen-rc:

### DIFF
--- a/apps/meteor/ee/server/api/licenses.ts
+++ b/apps/meteor/ee/server/api/licenses.ts
@@ -71,7 +71,7 @@ API.v1.addRoute(
 
 API.v1.addRoute(
 	'licenses.isEnterprise',
-	{ authRequired: true },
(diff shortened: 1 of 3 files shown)


sampaiodiego (dismissed):

apps/meteor/ee/server/api/licenses.ts
```diff
@@ -71,7 +71,7 @@ API.v1.addRoute(
 
 API.v1.addRoute(
 	'licenses.isEnterprise',
-	{ authRequired: true },
```
KevLehman, line 74: Notice that even with this change in place, the `method.callAnon/license:isEnterprise` is being used 👀 is that expected?

KevLehman (approved):
The comment is not related to the functionallity of the PR, so approving with a comment :)

ggazzo (approved):

PR #26091, opened 2022-07-01
[FIX] Regression: Clicling on the call button on the dialpad now closes the dialpad

cauefcr:


KevLehman:
Already fixed in Alek's branch 😬 thanks man!!

PR #26092, opened 2022-07-01
Chore: Fix Omnichannel E2E tests not running

murtaza98:


sampaiodiego (changes requested):

apps/meteor/server/models/raw/LivechatVisitors.ts
```diff
@@ -287,7 +287,7 @@ export class LivechatVisitorsRaw extends BaseRaw<ILivechatVisitor> implements IL
 	}
 
 	removeById(_id: string): Promise<DeleteResult> {
-		return this.removeById(_id);
+		return this.deleteOne({ _id });
```
sampaiodiego, line 290: is this being used? I mean, glad you fixed since it is a infinite recursive loop, if this is being used somewhere how none found it before? 🙈

murtaza98: Yeah, I believe this is getting used on Livechat when visitor requests to delete their data :see_no_evil:

PR #26098, opened 2022-07-01
[BREAK] Remove support to old MongoDB versions

sampaiodiego:

As per MongoDB Lifecycle Schedules (https://www.mongodb.com/support-policy/lifecycles) we're removing official support to MongoDB versions **3.6 and 4.0** that have already reached end-of-life.

As MongoDB 4.2 was a "supported" version before Rocket.Chat 5.0, we'll continue supporting it, but will be flagged as deprecated. We recommend upgrading to MongoDB 4.4+.

Here are official docs on how to upgrade to some of the supported versions:

- https://www.mongodb.com/docs/manual/release-notes/4.2-upgrade-replica-set/
- https://www.mongodb.com/docs/v4.4/release-notes/4.4-upgrade-replica-set/
- https://www.mongodb.com/docs/manual/release-notes/5.0-upgrade-replica-set/


ggazzo (dismissed):

KevLehman (approved):

debdutdeb (changes requested):
this also need to be updated 

https://github.com/RocketChat/Rocket.Chat/blob/develop/.github/workflows/build_and_test.yml#L756

sampaiodiego:
good catch @debdutdeb .. thx 🙏

debdutdeb:
I am slightly confused here, are we removing 4.2 or deprecating? The patch shows deprecating but pr description is saying remoing :eyes:

KevLehman:
Why this tries to run the tests against 3.6, is a mistery I don't have the answer for
image

debdutdeb:
Nvm noticed Kev's edit

KevLehman:
> Nvm noticed Kev's edit

:trollface:

apps/meteor/server/startup/serverRunning.js
```diff
@@ -72,8 +72,8 @@ Meteor.startup(function () {
 			exitIfNotBypassed(process.env.BYPASS_NODEJS_VALIDATION);
 		}
 
-		if (!semver.satisfies(semver.coerce(mongoVersion), '>=3.6.0')) {
-			msg += ['', '', 'YOUR CURRENT MONGODB VERSION IS NOT SUPPORTED,', 'PLEASE UPGRADE TO VERSION 3.6 OR LATER'].join('\n');
+		if (!semver.satisfies(semver.coerce(mongoVersion), '>=4.2.0')) {
+			msg += ['', '', 'YOUR CURRENT MONGODB VERSION IS NOT SUPPORTED,', 'PLEASE UPGRADE TO VERSION 4.2 OR LATER'].join('\n');
```
debdutdeb, line 76:
```suggestion
			msg += ['', '', 'YOUR CURRENT MONGODB VERSION IS NOT SUPPORTED,', 'PLEASE UPGRADE TO VERSION 4.4 OR LATER'].join('\n');
```
maybe we should directly ask to use 4.4

debdutdeb: else basically they'll update to 4.2 and see the deprecation warning. better ask them to start with the supported version imo

sampaiodiego:
> Why this tries to run the tests against 3.6, is a mistery I don't have the answer for 

the pipeline is still mandatory on github branch config.. once this one gets validated I'll go ahead and change github's config to not enforce this pipeline to merge to `develop` anymore.

apps/meteor/server/startup/serverRunning.js
KevLehman: No, because 4.2 is the last "supported" version (even tho it's deprecated. So customers can live there until we drop support (as many would probably live with 3.6 for a while, hehe)

debdutdeb: oh no I commited already 🙈 

But, even though that is supported, the message basically asks them to start using 4.2, which is not good. I get the technicality of it but for new deployments this is problematic

debdutdeb: maybe we mention the fact that 4.2 is "deprecated" explicitely?

KevLehman: Well, good enough. That way less ppl will know 4.2 is still around 🙈

debdutdeb: should we go the "supported yet deprecated" way? i do kinda agree with your point too - since 4.2 is supported still.

debdutdeb: nvm it won't exit out - 4.4 message is good enough 🙊 approving

debdutdeb (approved):

apps/meteor/server/startup/serverRunning.js
```diff
@@ -82,11 +82,11 @@ Meteor.startup(function () {
 		showSuccessBox('SERVER RUNNING', msg);
 
 		// Deprecation
-		if (!semver.satisfies(semver.coerce(mongoVersion), '>=4.2.0')) {
+		if (!semver.satisfies(semver.coerce(mongoVersion), '>=4.4.0')) {
 			msg = [
 				`YOUR CURRENT MONGODB VERSION (${mongoVersion}) IS DEPRECATED.`,
 				'IT WILL NOT BE SUPPORTED ON ROCKET.CHAT VERSION 5.0.0 AND GREATER,',
```
KevLehman:
```suggestion
				'IT WILL NOT BE SUPPORTED ON ROCKET.CHAT VERSION 6.0.0 AND GREATER,',
```

KevLehman: cmon

sampaiodiego: faster than light babe

debdutdeb: 🙈

KevLehman: That's the power of caffeine

debdutdeb (approved):
can we keep this approval from being dismissed plox? 🥲

KevLehman (approved):

KevLehman:
> can we keep this approval from being dismissed plox? 🥲

You want me to dismiss it again don't you

debdutdeb:
> > can we keep this approval from being dismissed plox? 🥲
> 
> You want me to dismiss it again don't you

You wouldn't 😘

KevLehman:
> > > can we keep this approval from being dismissed plox? 🥲
> > 
> > 
> > You want me to dismiss it again don't you
> 
> You wouldn't 😘

Try me :trollface:

debdutdeb:
> > > > can we keep this approval from being dismissed plox? 🥲
> > > 
> > > 
> > > You want me to dismiss it again don't you
> > 
> > 
> > You wouldn't 😘
> 
> Try me :trollface:

you do that and i'll dm you all the spams every day for a month 😇

ggazzo (approved):

PR #26100, opened 2022-07-01
[BREAK] Upgrade to version 5.0 can be done only from version 4.x

sampaiodiego:

I've changed [migration 233](https://github.com/RocketChat/Rocket.Chat/pull/26100/files#diff-85bddcfda67538763abfda90ee7bc87935470be1f08422be3c0548b300777643) (the first one before version 4.0 to show an error in case it gets executed, because that means the server never ran 4.x.


apps/meteor/server/startup/migrations/v240.ts
```diff
@@ -1,10 +1,18 @@
-import { Settings } from '@rocket.chat/models';
-
 import { addMigration } from '../../lib/migrations';
 
+// this was the last migration on version 4.0, so if this needs to run,
```
debdutdeb, line 3: should this not be on the first migration added to 4.0.0? since if they are still in 3.18.7 they'll be on migration 232 if iirc and if they try to jump to 5.x the first attempt will be to load 233. with this patch it won't show this intended message but just 'migration not found'.

sampaiodiego: you're right! I got few minutes stuck trying to think on this logic, but I think you got it right.. let me to the changes

sampaiodiego: well, actually, it doesn't matter if it is 232 or 240, because not having either means the same, that server didn't use 4.0 .. the migration 240 would be found even if having to jump  from 232.. 

BUT even though this works, I still agree your logic makes more sense and am doing the changes =)

debdutdeb: > the migration 240 would be found even if having to jump from 232..
Interesting, i'm not aware of how migrations are run. So basically you're saying that even if 233 is not found it'll keep finding the next migration[s]? 

I thought if it can't find 233 (the next one) it'll just error out with 'migration not found' - not showing the error you are intending.

Either way one of the intended results (the more important one) will be the same, not being able to start unless at least they are at migration 240.

debdutdeb (approved):
234 through 240 can be removed as well.

ggazzo (approved):

KevLehman (approved):

PR #26102, opened 2022-07-02
Chore: Remove unused migrations

debdutdeb:

After giving it some thought:

- 234 through 240 are not going to be run anymore. Keeping them does not affect behavior of course, but this (removing) makes it easier to quickly glance at and understand what migrations are actually included in 5.x.y (especially in tag compare view or in general just checking the ref).

- Also changed the file name of 233 to be more explicit at what it does so to not confuse with actual "migrations" without having to open the file. 

- The redirect to the documentation page (go.rocket....) is not yet set up, jfyi.


pierre-lehnen-rc:
You forgot to remove them from the index

debdutdeb:
Screenshot 2022-07-02 at 11 06 08 PM
lord o lord 🥲

KevLehman (approved):

sampaiodiego (approved):

PR #26117, opened 2022-07-04
Regression: Users on new sessions are forced to re-configure 2fa

pierre-lehnen-rc:


ggazzo (approved):

ggazzo:
some automated test would be welcome here

PR #26130, opened 2022-07-05
[IMPROVE] System messages' consistency

matheusbsilva137:


apps/meteor/packages/rocketchat-i18n/i18n/en.i18n.json
```diff
@@ -3901,20 +3901,20 @@
...
   "Room_archivation_state_true": "Archived",
   "Room_archived": "Room archived",
   "room_changed_announcement": "Room announcement changed to: <em>__room_announcement__</em> by <em>__user_by__</em>",
-  "room_changed_avatar": "Room avatar changed by <em>__user_by__</em>",
+  "room_changed_avatar": "changed room avatar",
   "room_changed_description": "Room description changed to: <em>__room_description__</em> by <em>__user_by__</em>",
-  "room_changed_privacy": "Room type changed to: <em>__room_type__</em> by <em>__user_by__</em>",
-  "room_changed_topic": "Room topic changed to: <em>__room_topic__</em> by <em>__user_by__</em>",
+  "room_changed_privacy": "changed room to __room_type__",
+  "room_changed_topic": "Changed room topic to <em>__room_topic__</em>",
```
GleidsonDaniel, line 3914:
```suggestion
  "room_changed_topic": "changed room topic to <em>__room_topic__</em>",
```

apps/meteor/packages/rocketchat-i18n/i18n/en.i18n.json
```diff
@@ -7,8 +7,8 @@
   "__usersCount__people_will_be_invited": "__usersCount__ people will be invited",
   "__username__is_no_longer__role__defined_by__user_by_": "__username__ is no longer __role__ by __user_by__",
   "__username__was_set__role__by__user_by_": "__username__ was set __role__ by __user_by__",
-  "This_room_encryption_has_been_enabled_by__username_": "This room's encryption has been enabled by __username__",
-  "This_room_encryption_has_been_disabled_by__username_": "This room's encryption has been disabled by __username__",
+  "This_room_encryption_has_been_enabled_by__username_": "enabled E2E Encryption for this room",
+  "This_room_encryption_has_been_disabled_by__username_": "disabled E2E Encryption for this room",
```
sampaiodiego, line 11: if `__username__` is going to be removed from these translations the code needs to be updated as well.

ideally I'd say to create a new string instead of changing these.

apps/meteor/app/channel-settings/server/functions/saveRoomType.js
```diff
@@ -41,13 +41,13 @@ export const saveRoomType = function (rid, roomType, user, sendMessage = true) {
 	if (sendMessage) {
 		let message;
 		if (roomType === 'c') {
-			message = TAPi18n.__('Channel', {
+			message = TAPi18n.__('Public', {
 				lng: (user && user.language) || settings.get('Language') || 'en',
-			});
+			}).toLowerCase();
```
sampaiodiego, line 46: not sure, but maybe it is a better idea to have different translations for lowercase.. not sure how other languages behave to `toLowerCase()`

sampaiodiego (dismissed):
I see a lot of translations where a variable has been removed, but the code is still providing them. For those cases I'd recommend creating a new string, because other languages will be incorrect anyways. The code also needs to be updated.

apps/meteor/packages/rocketchat-i18n/i18n/en.i18n.json
diegolmello: And this is a breaking change on mobile apps. We need to keep supporting old servers.

apps/meteor/packages/rocketchat-i18n/i18n/en.i18n.json
matheusbsilva137: Fixed! Thanks for pointing this issue 🤗

apps/meteor/packages/rocketchat-i18n/i18n/en.i18n.json
matheusbsilva137: > if `__username__` is going to be removed from these translations the code needs to be updated as well.
> 
> ideally I'd say to create a new string instead of changing these.

Agree this is a better option. We were first thinking of keeping the code the same + changing the strings so as not to "lose" the translations we already have, but I see this way it'd be difficult to track the languages with outdated translations :/
So we'll create new strings and change the code to work only with the data from our new system messages. It's still going to be a breaking change on mobile apps anyways, right @diegolmello ?

diegolmello: > So we'll create new strings and change the code to work only with the data from our new system messages. It's still going to be a breaking change on mobile apps anyways, right @diegolmello ?

It won't if we sync the implementation.

apps/meteor/app/channel-settings/server/functions/saveRoomType.js
LucianoPierdona: updated!

apps/meteor/packages/rocketchat-i18n/i18n/en.i18n.json
LucianoPierdona: updated!

hugocostadev (dismissed):
Hi there, @LucianoPierdona could you please remove all font-styles from system-messages ? E.g: `<em></em>` and etc... 
 
There are a task to remove those, if you add will be confusing in some cases. Thanks

gabriellsh (dismissed):
LGTM! Let's wait for QA and we can merge.

apps/meteor/packages/rocketchat-i18n/i18n/en.i18n.json
```diff
@@ -3902,19 +3914,24 @@
...
   "Room_archivation_state_true": "Archived",
   "Room_archived": "Room archived",
   "room_changed_announcement": "Room announcement changed to: <em>__room_announcement__</em> by <em>__user_by__</em>",
   "room_changed_avatar": "Room avatar changed by <em>__user_by__</em>",
+  "room_avatar_changed": "changed room avatar",
   "room_changed_description": "Room description changed to: <em>__room_description__</em> by <em>__user_by__</em>",
   "room_changed_privacy": "Room type changed to: <em>__room_type__</em> by <em>__user_by__</em>",
   "room_changed_topic": "Room topic changed to: <em>__room_topic__</em> by <em>__user_by__</em>",
+  "room_changed_type": "changed room to __room_type__",
+  "room_changed_topic_to": "Changed room topic to __room_topic__",
```
GleidsonDaniel, line 3930:
```suggestion
  "room_changed_topic_to": "changed room topic to __room_topic__",
```

gabriellsh (approved):

tiagoevanp (approved):

PR #26132, opened 2022-07-05
Chore: Upgrade ESLint

tassoevan:

Upgrade ESLint (to 8.19.0) and its dependencies, dropping outdated rules.


d-gubert (dismissed):
All good on Apps-Engine side

apps/meteor/ee/client/.eslintrc.json
```diff
@@ -0,0 +1 @@
+../../client/.eslintrc.json
```
sampaiodiego, line 1: can we stop using symbolic links? they're way harder to keep, someone needs to remember they exists and need to be taken care of.

we could simply use `"extends": "../../client/.eslintrc.json"`

apps/meteor/client/.eslintrc.json
```diff
@@ -0,0 +1,257 @@
+{
+  "root": true,
```
sampaiodiego, line 2: why do we still need a eslint config specific for client? why not merge it with the root one?

sampaiodiego (changes requested):

apps/meteor/client/.eslintrc.json
tassoevan: My memory might be betraying me, but I think at the time some `eslint-react-plugin` rules conflicted with other ESLint rules (maybe due to ESLint version being outdated), as well as many of the configuration comes from the Fuselage monorepo. I wanted to review everything, but I thought this is a little bit out of the scope of this pull request.

sampaiodiego (approved):

sampaiodiego (approved):

ggazzo (approved):

PR #26135, opened 2022-07-05
Regression: Added missing call button in the contact info contextual bar

aleksandernsilva:

### DIFF
--- a/apps/meteor/client/views/omnichannel/directory/contacts/contextualBar/ContactInfo.js
+++ b/apps/meteor/client/views/omnichannel/directory/contacts/contextualBar/ContactInfo.js
@@ -1,4 +1,4 @@
-import { Box, Margins, ButtonGroup, Button, Icon } from '@rocket.chat/fuselage';
+import { Box, Margins, ButtonGroup, Button, Icon, Divider } from '@rocket.chat/fuselage';
 import { useMutableCallback } from '@rocket.chat/fuselage-hooks';
 import { useToastMessageDispatch, useCurrentRoute, useRoute, useTranslation } from '@rocket.chat/ui-contexts';
 import React, { useEffect, useMemo, useState } from 'react';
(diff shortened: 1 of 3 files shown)


apps/meteor/client/views/omnichannel/directory/contacts/contextualBar/ContactInfo.js
```diff
@@ -157,13 +163,20 @@ const ContactInfo = ({ id, rid, route }) => {
...
+				<ButtonGroup stretch flexWrap='wrap'>
+					{isCallReady && (
+						<>
+							<VoipInfoCallButton phoneNumber={phoneNumber} mi={0} flexBasis='0' />
+							{showContactHistory && <Divider width='100%' />}
+						</>
+					)}
+
+					{showContactHistory && (
+						<Button onClick={onChatHistory} mis={0} flexBasis='0'>
```
aleksandernsilva, line 175: This logic is necessary, so we can have both layouts depending on the number of buttons:
Screen Shot 2022-07-05 at 13 03 22
Screen Shot 2022-07-05 at 13 03 33

tiagoevanp: Wait, this divider is not aligned... can you fix this?

tiagoevanp (changes requested):

apps/meteor/client/views/omnichannel/directory/contacts/contextualBar/ContactInfo.js
aleksandernsilva: Yeah, this image is outdated. I reused an old print just to give the idea 😅 
Everything should be aligned.

tiagoevanp (approved):

ggazzo (changes requested):
already dm alek about how to improve

PR #26150, opened 2022-07-06
[IMPROVE] General federation improvements

MarcosSpessatto:

I know this changed a lot of files, but the main goal for this PR is not to change any behavior, the goals for the PR are:

- Refactor the code;
- Solve any tech debt;
- Simplify and reuse some parts of the code;
- Remove duplicated code;
- Remove all unsafe type castings;
- Solve all Eslint errors and warnings;
- Split too big files;
- Encapsulate the business logic in a better way, avoiding exposing and leaking internal logic to the unintended layers;
- Improve the actual test cases;
- Add more test cases, since a lot of cases were omitted during the release phase;
- Remove unsafe `Object.assign` statements and prefer to use the class `constructor` instead;


alansikora (dismissed):

apps/meteor/app/lib/server/methods/sendMessage.js
```diff
@@ -107,8 +108,9 @@ Meteor.methods({
 		}
 
 		try {
-			if (Promise.await(federationRoomServiceSender.isAFederatedRoom(message.rid))) {
-				return federationRoomServiceSender.sendMessageFromRocketChat(
+			const room = Rooms.findOneById(message.rid);
```
sampaiodiego, line 111: while suggestion to add a projection to return only the field `federated`, I realised a best place for this check is actually after like 78, where `room` is already defined.

apps/meteor/server/models/MatrixBridgedRoom.ts
```diff
@@ -0,0 +1,7 @@
+import { registerModel } from '@rocket.chat/models';
+
+import { trashCollection } from '../database/trash';
+import { db } from '../database/utils';
+import { MatrixBridgedRoomRaw } from './raw/MatrixBridgedRoom';
+
+registerModel('IMatrixBridgedRoomModel', new MatrixBridgedRoomRaw(db, trashCollection));
```
sampaiodiego, line 7: I guess we don't need to store removed data from this collection right?

```suggestion
registerModel('IMatrixBridgedRoomModel', new MatrixBridgedRoomRaw(db));
```

apps/meteor/server/models/MatrixBridgedUser.ts
```diff
@@ -0,0 +1,7 @@
+import { registerModel } from '@rocket.chat/models';
+
+import { trashCollection } from '../database/trash';
+import { db } from '../database/utils';
+import { MatrixBridgedUserRaw } from './raw/MatrixBridgedUser';
+
+registerModel('IMatrixBridgedUserModel', new MatrixBridgedUserRaw(db, trashCollection));
```
sampaiodiego, line 7:
```suggestion
registerModel('IMatrixBridgedUserModel', new MatrixBridgedUserRaw(db));
```

apps/meteor/server/models/raw/Users.js
```diff
@@ -1126,4 +1126,16 @@ export class UsersRaw extends BaseRaw {
...
+	removeRoomByRoomId(rid) {
+		return this.update(
+			{
+				__rooms: rid,
+			},
+			{
+				$pull: { __rooms: rid },
+			},
+			{ multi: true },
+		);
```
sampaiodiego, line 1139:
```suggestion
		return this.updateMany(
			{
				__rooms: rid,
			},
			{
				$pull: { __rooms: rid },
			},
		);
```

packages/core-typings/src/IUser.ts
```diff
@@ -153,7 +153,9 @@ export interface IRegisterUser extends IUser {
 	username: string;
 	name: string;
 }
+
 export const isRegisterUser = (user: IUser): user is IRegisterUser => user.username !== undefined && user.name !== undefined;
+export const isUserFederated = (user: Partial<IUser>): user is IUser => 'federated' in user && user.federated === true;
```
sampaiodiego, line 158: since this is not a type assertion, I'm not sure this should be here.. it is conceptually different from `isRegisterUser` that asserts a new type (`IRegisterUser`) if some validation is `true`.

sampaiodiego (changes requested):

tassoevan (dismissed):

apps/meteor/app/models/server/models/Users.js
```diff
@@ -999,6 +999,16 @@ export class Users extends Base {
 		);
 	}
 
+	findActiveFederated(options = {}) {
+		return this.find(
+			{
+				active: true,
+				federated: true,
```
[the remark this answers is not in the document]
MarcosSpessatto: I'd say it would be better to have a new one, since as you said this one is from the oldest federation and probably will be deprecated soon. What do you think?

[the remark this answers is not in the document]
sampaiodiego: I think it may be worse having two different flags for the same purpose..

apps/meteor/app/federation-v2/server/domain/FederatedRoom.ts
```diff
@@ -85,4 +87,120 @@ export class FederatedRoom {
...
+	}
+
+	public static createInstance(externalId: string, creator: FederatedUser, members: FederatedUser[]): DirectMessageFederatedRoom {
+		return new DirectMessageFederatedRoom({
+			externalId,
+			members,
+			internalReference: {
+				t: RoomType.DIRECT_MESSAGE,
+				u: creator.getInternalReference(),
+			} as unknown as IRoom,
```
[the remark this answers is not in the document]
MarcosSpessatto: Yes... it's the data we need to manipulate DM rooms, this is even not persisted into the DB.

apps/meteor/app/federation-v2/server/infrastructure/rocket-chat/adapters/Message.ts
```diff
@@ -3,7 +3,7 @@ import type { FederatedRoom } from '../../../domain/FederatedRoom';
 import type { FederatedUser } from '../../../domain/FederatedUser';
 
 export class RocketChatMessageAdapter {
-	public async sendMessage(user: FederatedUser, text: string, room: FederatedRoom): Promise<void> {
-		new Promise((resolve) => resolve(sendMessage(user.internalReference, { msg: text }, room.internalReference)));
+	public async sendMessage(user: FederatedUser, room: FederatedRoom, messageText: string): Promise<void> {
+		Promise.resolve(sendMessage(user.getInternalReference(), { msg: messageText }, room.getInternalReference()));
```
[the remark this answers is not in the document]
MarcosSpessatto: Yes, you are right. 😅

apps/meteor/app/lib/server/methods/sendMessage.js
MarcosSpessatto: Makes sense.

apps/meteor/server/models/MatrixBridgedRoom.ts
MarcosSpessatto: Yes

packages/core-typings/src/IUser.ts
MarcosSpessatto: I followed the idea of `isRoomFederated()`, https://github.com/RocketChat/Rocket.Chat/blob/develop/packages/core-typings/src/IRoom.ts#L90. Where do you think it should be?

apps/meteor/app/lib/server/functions/createRoom.ts
```diff
@@ -136,7 +136,7 @@ export const createRoom = function <T extends RoomType>(
 			try {
 				callbacks.run('federation.beforeAddUserAToRoom', { user: member, inviter: owner }, room);
 			} catch (error) {
-				throw new Meteor.Error((error as any)?.message);
+				continue;
```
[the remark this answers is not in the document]
MarcosSpessatto: The original idea was to behave like that, but this was leading to strange behavior since the room was created, but with that throw, the subscriptions were not being created, even for the room creator. So I've changed this to allow the subscription creation for the users who exist (in this case, the creator is 100% sure that exist), and I handle the federation stuff in the other callback.

apps/meteor/app/models/server/models/Users.js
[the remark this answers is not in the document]
alansikora: One is for the old, v1 federation, which will be deprecated. This one is for the Matrix federation.

tassoevan (dismissed):

apps/meteor/server/models/raw/MatrixBridgedRoom.ts
```diff
@@ -0,0 +1,39 @@
...
+import { BaseRaw } from './BaseRaw';
+
+export class MatrixBridgedRoomRaw extends BaseRaw<IMatrixBridgedRoom> implements IMatrixBridgedRoomModel {
+	constructor(db: Db, trash?: Collection<RocketChatRecordDeleted<IMatrixBridgedRoom>>) {
+		super(db, 'matrix_bridged_rooms', trash);
+	}
+
+	protected modelIndexes(): IndexDescription[] {
+		return [
+			{ key: { rid: 1, mri: 1 }, unique: true, sparse: true },
```
sampaiodiego, line 14: there is no need for this index since there is no specific `find` using both fields at the same time
```suggestion
```

apps/meteor/server/models/raw/MatrixBridgedRoom.ts
```diff
@@ -0,0 +1,39 @@
...
+			{ key: { rid: 1 }, unique: true, sparse: true },
+			{ key: { mri: 1 }, unique: true, sparse: true },
```
sampaiodiego, line 16: the other two shouldn't be `sparse`, as reading the code they will never be `undefined`.. but since they're already created, we can let it them as they are.

apps/meteor/server/models/raw/MatrixBridgedUser.ts
```diff
@@ -0,0 +1,53 @@
...
+import { BaseRaw } from './BaseRaw';
+
+export class MatrixBridgedUserRaw extends BaseRaw<IMatrixBridgedUser> implements IMatrixBridgedUserModel {
+	constructor(db: Db, trash?: Collection<RocketChatRecordDeleted<IMatrixBridgedUser>>) {
+		super(db, 'matrix_bridged_users', trash);
+	}
+
+	protected modelIndexes(): IndexDescription[] {
+		return [
+			{ key: { uid: 1, mui: 1 }, unique: true, sparse: true },
```
sampaiodiego, line 14: same comments as `MatrixBridgedRoom.ts`
```suggestion
```

apps/meteor/app/federation-v2/server/domain/FederatedRoom.ts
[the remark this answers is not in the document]
sampaiodiego: my point is, why is it an `IRoom` then? you should create another type that is a sub-set of `IRoom`, so you don't need to add type cast.

packages/core-typings/src/IUser.ts
sampaiodiego: ok, `IRoomFederated()` is fine since it is indeed asserting the type `IRoomFederated` ... you could create a `IUserFederated` here, where the `federated` property is not optional.. that would be a correct usage.

apps/meteor/app/federation-v2/server/infrastructure/rocket-chat/adapters/User.ts
```diff
@@ -36,58 +55,29 @@ export class RocketChatUserAdapter {
...
+		const user = await Users.findOneById(userId);
+		if (!user || !user.username) {
+			throw new Error(`User with internalId ${userId} not found`);
+		}
+		return user;
 	}
 
 	public async createFederatedUser(federatedUser: FederatedUser): Promise<void> {
-		const existingLocalUser = await Users.findOneByUsername(federatedUser.internalReference.username || '');
+		const existingLocalUser = await Users.findOneByUsername(federatedUser.getUsername() || '');
```
sampaiodiego, line 76: I guess we querying an empty string is not really intended right? if so, this is better if written like this:
```suggestion
		const existingLocalUser = federatedUser.getUsername() && await Users.findOneByUsername(federatedUser.getUsername());
```

apps/meteor/ee/app/federation-v2/server/infrastructure/rocket-chat/adapters/Room.ts
```diff
@@ -1,49 +1,8 @@
...
-		createDirectMessage(members, creatorId) as ICreatedRoom;
-	}
-
-	private createFederatedRoomEEInstance(externalRoomId: string, room: IRoom): FederatedRoomEE {
-		const federatedRoom = FederatedRoomEE.build();
-		federatedRoom.externalId = externalRoomId;
-		federatedRoom.internalReference = room;
-
-		return federatedRoom;
+		return Promise.resolve(createDirectMessage(members, creatorId));
```
sampaiodiego, line 6:
```suggestion
	createDirectMessage(members, creatorId);
```

apps/meteor/ee/app/federation-v2/server/infrastructure/rocket-chat/adapters/User.ts
```diff
@@ -1,24 +1,24 @@
...
 import { Users } from '@rocket.chat/models';
 
 import { RocketChatUserAdapter } from '../../../../../../../app/federation-v2/server/infrastructure/rocket-chat/adapters/User';
-import type { FederatedUserEE } from '../../../domain/FederatedUser';
 
 export class RocketChatUserAdapterEE extends RocketChatUserAdapter {
-	public async createLocalUser(federatedUser: FederatedUserEE): Promise<void> {
-		const existingLocalUser = await Users.findOneByUsername(federatedUser.internalReference.username || '');
+	public async createLocalUser(internalUser: IUser): Promise<void> {
+		const existingLocalUser = await Users.findOneByUsername(internalUser.username || '');
```
sampaiodiego, line 8:
```suggestion
		const existingLocalUser = internalUser.username && await Users.findOneByUsername(internalUser.username, { projection: { _id: 1 } });
```

apps/meteor/app/federation-v2/server/application/sender/RoomInternalHooksValidator.ts
```diff
@@ -0,0 +1,97 @@
...
+			usernames.some(
+				(username) =>
+					!FederatedUser.isOriginalFromTheProxyServer(this.bridge.extractHomeserverOrigin(username), this.internalHomeServerDomain),
+			) || internalUsers.filter((user) => !this.isAddingANewExternalUser(user)).some((user) => isUserFederated(user as IUser));
+		if (atLeastOneExternalUser) {
+			throw new Error('error-this-is-an-ee-feature');
+		}
+	}
+
+	private isAddingANewExternalUser(user: IUser | string): boolean {
```
sampaiodiego, line 94: if you change this function to a type assertion you can remove the type casting all over this file where you need to do like `internalUser as IUser`
```suggestion
	private isAddingANewExternalUser(user: IUser | string): user is string {
```

sampaiodiego (dismissed):

MarcosSpessatto:
> I don't have the knowledge to say it will in fact break both systems, but the isRemote flag is also a generic flag to identify "not local users".. introducing a new flag can cause other issues as well, like we need to make sure to filter both out..

> there is at least another place missing an update to consider the new flag, the statistics for federated users:
[Rocket.Chat/apps/meteor/app/statistics/server/lib/statistics.ts](https://github.com/RocketChat/Rocket.Chat/blob/8c2c016ac184f40f4e411eedcf288dfbc3b968d0/apps/meteor/app/statistics/server/lib/statistics.ts#L275)

> Line 275 in [8c2c016](https://github.com/RocketChat/Rocket.Chat/commit/8c2c016ac184f40f4e411eedcf288dfbc3b968d0)

 > statistics.federatedUsers = federationOverviewData.numberOfFederatedUsers; 
> I wonder if there are others as well

I understand your point and I agree with the statistics example you gave. Indeed both flags are being used for the same purpose, but I respectfully disagree we should use the same for either "old federation" and the new version of it, since if we are using both at the same time, changes happening in one feature will affect directly the other. Another example will be, if you are using the old federation and you enable the new one, some unpredictable things could happen since you will start to use data from one federation to another. Maybe @alansikora can help us on this as well.

alansikora:
It is exactly what @MarcosSpessatto said. We can't.

And the old federation is going to be removed anyway, so we should not bother about it. We will ensure a full cleanup when the time to remove the old federation comes.

tassoevan (approved):

casalsgh (approved):

PR #26154, opened 2022-07-06
Regression: Parse outbound phone number removing * putting + char

tiagoevanp:

### DIFF
--- a/apps/meteor/client/components/voip/room/VoipRoomForeword.tsx
+++ b/apps/meteor/client/components/voip/room/VoipRoomForeword.tsx
@@ -9,14 +9,14 @@ import { parseOutboundPhoneNumber } from '../../../../ee/client/lib/voip/parseOu
 export const VoipRoomForeword = ({ room }: { room: IVoipRoom }): ReactElement => {
 	const t = useTranslation();
 
-	const avatarUrl = getUserAvatarURL(room.name || room.fname);
+	const avatarUrl = getUserAvatarURL(room.name);
 
-	const roomName = room.name || room.fname;
+	const roomName = room.name;
 
 	return (
 		<Box is='div' flexGrow={1} display='flex' justifyContent='center' flexDirection='column'>
 			<Box display='flex' justifyContent='center' mbs='x24'>
-				<Avatar size='x48' title={room.name || room.fname} url={avatarUrl} />
+				<Avatar size='x48' title={room.name} url={avatarUrl} />
 			</Box>
 			<Box color='default' fontScale='h2' flexGrow={1} mb='x16'>
 				{t('You_have_joined_a_new_call_with')}


packages/core-typings/src/IRoom.ts
```diff
@@ -193,6 +193,7 @@ export interface IOmnichannelRoom extends IOmnichannelGenericRoom {
 
 export interface IVoipRoom extends IOmnichannelGenericRoom {
 	t: 'v';
+	name: string;
```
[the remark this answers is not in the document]
tiagoevanp: name is not optional for IVoipRoom... that's why I'm overriding types. I'm open to discuss about my point of view... 😃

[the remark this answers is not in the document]
KevLehman: Ohhhh, well, think that makes sense 👀

KevLehman:
Opening the Omnichannel Directory seems to not to work on this PR 🤔

apps/meteor/client/views/omnichannel/directory/calls/CallTable.tsx
```diff
@@ -147,8 +148,10 @@ const CallTable: FC = () => {
 			const duration = moment.duration(callDuration / 1000, 'seconds');
 			return (
 				<Table.Row key={_id} tabIndex={0} role='link' onClick={(): void => onRowClick(_id, v?.token)} action qa-user-id={_id}>
-					<Table.Cell withTruncatedText>{fname}</Table.Cell>
-					<Table.Cell withTruncatedText>{Array.isArray(v?.phone) ? v?.phone[0]?.phoneNumber : v?.phone}</Table.Cell>
+					<Table.Cell withTruncatedText>{parseOutboundPhoneNumber(fname)}</Table.Cell>
```
KevLehman, line 151: There could be rooms without `fname` out there. I can add a migration to fix them, but you can put a default value in here too (to avoid crashes)

tiagoevanp: well, fname always will be part of VoIP rooms...

KevLehman: Yep, but some old rooms (from the earlier stages of MVP) don't have that prop. If there's one of those on the wild, then the page may crash 👀 (for new rooms, this shouldn't be a problem)

KevLehman (approved):
Functional-wise, seems to be working 👀

aleksandernsilva (dismissed):

apps/meteor/client/components/voip/room/VoipRoomForeword.tsx
```diff
@@ -1,15 +1,18 @@
...
 
-export const VoipRoomForeword = ({ room }: { room: IRoom }): ReactElement => {
+export const VoipRoomForeword = ({ room }: { room: IVoipRoom }): ReactElement => {
 	const t = useTranslation();
 
 	const avatarUrl = getUserAvatarURL(room.name || room.fname);
 
+	const roomName = room.name || room.fname;
+
 	return (
```
ggazzo, line 16: you changed the IVoipRoom and added `name: string;` as required, but looking at the code, we dont look very confident about it(always using name || fname), could you check this for me please?

ggazzo (changes requested):

ggazzo (approved):

PR #26155, opened 2022-07-06
Regression: Contact manager edit/view not working

KevLehman:

Basically, the Contact Center was working, but not the right way. This PR fixes:
- Ability to select Contact Managers from dropdown
- Ability to validate Contact Edits without requesting data a ton of times
- Ability to remove Contact manager from a contact
- Ability to see Contacts and Contact Managers on Contact View
- Fix endpoints validation
- Add validators (ajv) to endpoint, thou not being used yet (since we hit a special endpoint)


apps/meteor/app/livechat/server/api/v1/contact.ts
```diff
@@ -0,0 +1,70 @@
...
+				token: String,
+				name: String,
+				email: Match.Maybe(String),
+				phone: Match.Maybe(String),
+				username: Match.Maybe(String),
+				customFields: Match.Maybe(Object),
+				contactManager: Match.Maybe({
+					username: String,
+				}),
+			});
```
murtaza98, line 24: This shouldn't be required now since you now have this AJV validator over [here](https://github.com/RocketChat/Rocket.Chat/blob/ab7d25c00a6dde6a42d70012ae03a48f55385905/packages/rest-typings/src/v1/omnichannel.ts#L873)

apps/meteor/app/livechat/server/api/v1/contact.ts
```diff
@@ -0,0 +1,70 @@
...
+
+			const contact = await Contacts.registerContact(this.bodyParams);
+
+			return API.v1.success({ contact });
+		},
+		async get() {
+			check(this.queryParams, {
+				contactId: String,
+			});
```
murtaza98, line 33: This shouldn't be required now since you now have this AJV validator over [here](https://github.com/RocketChat/Rocket.Chat/blob/ab7d25c00a6dde6a42d70012ae03a48f55385905/packages/rest-typings/src/v1/omnichannel.ts#L875)

apps/meteor/app/livechat/server/lib/Contacts.ts
```diff
@@ -46,31 +64,38 @@ export const Contacts = {
...
+				contactId = (await LivechatVisitors.insertOne(userData)).insertedId;
 			}
 		}
 
-		updateUser.$set.name = name;
-		updateUser.$set.phone = (phone && [{ phoneNumber: phone }]) || null;
-		updateUser.$set.visitorEmails = (visitorEmail && [{ address: visitorEmail }]) || null;
-
-		const allowedCF = LivechatCustomField.find({ scope: 'visitor' }, { fields: { _id: 1 } }).map(({ _id }) => _id);
+		const allowedCF: ILivechatCustomField['_id'][] = LivechatCustomField.find({ scope: 'visitor' }, { fields: { _id: 1 } }).map(
```
murtaza98, line 74:
```suggestion
		const allowedCF: ILivechatCustomField['_id'][] = LivechatCustomField.find({ scope: 'visitor' }, { projection: { _id: 1 } }).map(
```

murtaza98:
@KevLehman While testing this PR I noticed that there isn't a way to un-assign a contact manager once it has already been assigned to a contact. I've raised this PR for it. Please check and merge it here if all looks okay -> https://github.com/RocketChat/Rocket.Chat/pull/26167

ggazzo:
could you just point me where the regression was introduced?

apps/meteor/app/livechat/server/api/v1/contact.ts
KevLehman: Left them intentionally, I'm not using the AJV validators since the route has both GET/POST and both accept different paylaods. So will check later when this is possible to do with the current config.

KevLehman:
> > 
> 
> ok, it was found there, but not introduced... Looking the changes and descriptiob, I can not understand what are the multiple issues and I'm forced to disagree with `Regression` terminology, with that I would not merge this on the freeze period as well

Merge it when you want man :) you have the power.

ggazzo (approved):

ggazzo:
@KevLehman can you ask some help from  QA?

apps/meteor/client/views/omnichannel/directory/contacts/contextualBar/ContactNewEdit.js
```diff
@@ -73,6 +73,8 @@ function ContactNewEdit({ id, data, close }) {
 	const [emailError, setEmailError] = useState();
 	const [phoneError, setPhoneError] = useState();
 	const [customFieldsError, setCustomFieldsError] = useState([]);
+	const [userId, setUserId] = useState('');
+	const [currentUsername, setCurrentUsername] = useState(username);
```
tiagoevanp, line 77: Why you need a currentUsername when you have a form controlled username? I don't get it...

apps/meteor/client/views/omnichannel/directory/contacts/contextualBar/ContactNewEdit.js
```diff
@@ -213,7 +227,7 @@ function ContactNewEdit({ id, data, close }) {
 						setCustomFieldsError={setCustomFieldsError}
 					/>
 				)}
-				{ContactManager && <ContactManager value={username} handler={handleUsername} />}
+				{ContactManager && <ContactManager value={userId} handler={handleUsername} />}
```
tiagoevanp, line 230: the handler is for userName, but the value passed is userID? Why to change the file where userId is used?

tiagoevanp (changes requested):
Can you provide a description of what your PR is doing? At least a resume with principal points of fix...

apps/meteor/app/livechat/server/api/v1/contact.ts
murtaza98: Makes sense :+1:

apps/meteor/client/views/omnichannel/directory/contacts/contextualBar/ContactNewEdit.js
murtaza98: Makes sense. Removed this variable :heavy_check_mark:

apps/meteor/client/views/omnichannel/directory/contacts/contextualBar/ContactNewEdit.js
murtaza98: This part was updated, check you please check this again?

ggazzo (approved):

ggazzo (approved):

PR #26157, opened 2022-07-06
[FIX] undefined toast error

pierre-lehnen-rc:

### DIFF
--- a/apps/meteor/client/hooks/useEndpointData.ts
+++ b/apps/meteor/client/hooks/useEndpointData.ts
@@ -24,10 +24,12 @@ export const useEndpointData = <TPath extends PathFor<'GET'>>(
 			.then(resolve)
 			.catch((error) => {
 				console.error(error);
-				dispatchToastMessage({
-					type: 'error',
-					message: error,
-				});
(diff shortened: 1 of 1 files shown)


apps/meteor/client/hooks/useEndpointData.ts
```diff
@@ -24,10 +24,12 @@ export const useEndpointData = <TPath extends PathFor<'GET'>>(
 			.then(resolve)
 			.catch((error) => {
 				console.error(error);
-				dispatchToastMessage({
-					type: 'error',
-					message: error,
-				});
+				if (error && typeof error === 'string' && error !== 'undefined') {
```
gabriellsh, line 27: Error doesn't need to be string. It can be a normal error object. Maybe change to `(error !== undefined)`?.

gabriellsh (changes requested):

apps/meteor/client/hooks/useEndpointData.ts
pierre-lehnen-rc: Tbh I'd prefer we just remove this toast completely already. It shouldn't have been here in the first place.

PR #26158, opened 2022-07-06
Regression: Cannot logout when CallProvider is unregistered and mounted

KevLehman:

### DIFF
--- a/apps/meteor/client/providers/CallProvider/CallProvider.tsx
+++ b/apps/meteor/client/providers/CallProvider/CallProvider.tsx
@@ -149,7 +149,11 @@ export const CallProvider: FC = ({ children }) => {
 
 		setQueueAggregator(voipClient.getAggregator());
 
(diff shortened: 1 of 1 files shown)


apps/meteor/client/providers/CallProvider/CallProvider.tsx
```diff
@@ -149,8 +149,8 @@ export const CallProvider: FC = ({ children }) => {
 
 		setQueueAggregator(voipClient.getAggregator());
 
-		return (): void => voipClient.unregister();
-	}, [result]);
+		return (): void => (clientState === 'registered' ? voipClient.unregister() : undefined);
```
aleksandernsilva, line 152: I would probably use `clientState === 'registered' && voipClient.unregister()` as a one liner or even better do `if (clientState === 'registered') voipClient.unregister()`. I think it expresses better that the important is the fn call and not the returned value.

aleksandernsilva (changes requested):

PR #26161, opened 2022-07-06
[IMPROVE] Use the provider app's user to send videoconf messages

pierre-lehnen-rc:


d-gubert (approved):
All good on Apps-Engine side

PR #26162, opened 2022-07-06
Regression: Fix marketplace releases tab crash bug

rique223:

Fixed a bug where RC would crash because the marketplace releases tab was trying to display undefined data from manually installed apps. 
Demo gif:
app-releases-tab-crash-error


rique223:
> If they only provide one link, such as terms but not the privacy, I would vote we show whichever one we get back. Thoughts?

Makes sense, when I did this logic I believed both informations were mandatory. Will adapt it in a sec.

graywolf336 (dismissed):

geekgonecrazy (approved):

apps/meteor/client/views/admin/apps/AppSecurity.tsx
```diff
@@ -75,12 +75,16 @@ const AppSecurity: FC<AppSecurityProps> = ({ privacyPolicySummary, appPermission
...
 						<Box display='flex' flexDirection='column'>
-							<Box is='a' href={tosLink}>
-								Terms of use
-							</Box>
-							<Box is='a' href={privacyLink}>
-								Privacy policy
-							</Box>
+							{tosLink && (
+								<Box is='a' href={tosLink}>
+									Terms of use
```
ggazzo, line 80: could you replace this hard text ? use translations keys instead

ggazzo (changes requested):

apps/meteor/client/views/admin/apps/AppSecurity.tsx
rique223: Solved

ggazzo: solved where my friend?

PR #26172, opened 2022-07-07
Regression: Cannot open Menu in searched message.

gabriellsh:


apps/meteor/app/ui/client/views/app/lib/getCommonRoomEvents.js
```diff
@@ -327,7 +328,8 @@ export const getCommonRoomEvents = () => ({
 	'click .message-actions__menu'(e, template) {
 		const messageContext = messageArgs(this);
 		const { msg: message, u: user, context: ctx } = messageContext;
-		const room = Rooms.findOne({ _id: template.data.rid });
+		const rid = Session.get('openedRoom');
+		const room = Rooms.findOne({ _id: rid });
```
tassoevan, line 332: I don't quite get what is happening here... Why `template.data.rid` is not returning a valid value? Is there any chance of the value of `Session.get('openedRoom')` not being the right one at some moment? Why not use `message.rid`?

gabriellsh: `template.data` in this case comes from another template (the search template) where the rid is no available. I'd have to either use `Session.get` there or use `message.rid` in order to provide this through the template data. I think using message.rid is better though, thanks for the tip!

tassoevan (approved):

PR #26173, opened 2022-07-07
Regression: Inline code and copyonly tag styles

gabriellsh:


filipemarins (approved):
I think we should move this style to the component side later on.

PR #26178, opened 2022-07-07
[IMPROVE] Capability to search visitors by custom fields

cauefcr:

Users of the endpoints [api/v1/omnichannel/contact.search](https://developer.rocket.chat/reference/api/rest-api/endpoints/omnichannel/livechat-endpoints/livechat-contact/omnichannel-search-contact) and [/api/v1/livechat/visitors.search](https://developer.rocket.chat/reference/api/rest-api/endpoints/omnichannel/livechat-endpoints/visitor/search-for-visitors) are now able to search by custom fields in their objects. 
Capability of selecting if a custom field can be searched for is added in the Omnichannel pannel as a toggle for `searchable`,  the included JSON in the Accounts' Custom Field example has been updated to make it explicit for future configurations that the field has to be enabled as searchable for that to happen.

## Steps to test or reproduce

Create a new custom field on Omnichannel->custom field, selecting it as searchable.
Edit a LiveChat visitor conversation to include custom field created above.
On the contact center, you can now search for the contents of that custom field.

## Further comments

This branch is broken, [here](https://github.com/RocketChat/Rocket.Chat/pull/26312) is the new one


apps/meteor/app/lib/server/functions/validateCustomFields.js
```diff
@@ -44,5 +44,9 @@ export const validateCustomFields = function (fields) {
 				method: 'registerUser',
 			});
 		}
+
+		if (field.searchable && fieldValue !== true && fieldValue !== false) {
```
KevLehman, line 48: Will only boolean fields be searchable?

KevLehman: Btw, not sure 🤔 is this file used by omnichannel?

apps/meteor/app/models/server/models/Users.js
```diff
@@ -61,6 +61,10 @@ export class Users extends Base {
 		this.tryEnsureIndex({ language: 1 }, { sparse: true });
 		this.tryEnsureIndex({ 'active': 1, 'services.email2fa.enabled': 1 }, { sparse: true }); // used by statistics
 		this.tryEnsureIndex({ 'active': 1, 'services.totp.enabled': 1 }, { sparse: true }); // used by statistics
+		this.tryEnsureIndex({ 'customFields.$**': 1 });
+
+		// const collectionObj = this.model.rawCollection();
+		// this.findAndModify = Meteor.wrapAsync(collectionObj.findAndModify, collectionObj);
```
KevLehman, line 67: :oh-no:

apps/meteor/app/livechat/server/api/v1/contact.js
```diff
@@ -51,19 +51,23 @@ API.v1.addRoute(
...
-
-				if (!email && !phone) {
+				if (!email && !phone && !custom) {
 					throw new Meteor.Error('error-invalid-params');
 				}
-
+				const allowedCF = LivechatCustomField.find({ scope: 'visitor' }, { fields: { _id: 1 } }, { searchability: true }).map(
+					({ _id }) => _id,
+				);
+				const customObj = Object.fromEntries(Array.from(new URLSearchParams(custom)).filter(([k]) => allowedCF.includes(k)));
```
KevLehman, line 64: Can you share an example of what `custom` would look like? Wanna see why we need to manually parsing the searchparams 👀

KevLehman, line 64: If you do this the other way, you can avoid having to query _all_ custom fields for visitor 👀 
Like:
1. Parse received custom fields
2. Validate they're not something nefarious (string only)
3. Find Custom Fields that match the received ones
4. Use the `found` list to do the queries

apps/meteor/app/livechat/server/api/v1/contact.js
```diff
@@ -51,19 +51,23 @@ API.v1.addRoute(
...
+				const { email, phone, custom } = this.queryParams;
 
-				const { email, phone } = this.queryParams;
-
-				if (!email && !phone) {
+				if (!email && !phone && !custom) {
 					throw new Meteor.Error('error-invalid-params');
 				}
-
+				const allowedCF = LivechatCustomField.find({ scope: 'visitor' }, { fields: { _id: 1 } }, { searchability: true }).map(
```
KevLehman, line 61: Shouldn't `searchability` be on the query object? 👀

apps/meteor/client/components/MarkdownText.tsx
```diff
@@ -110,7 +110,7 @@ const MarkdownText: FC<Partial<MarkdownTextParams>> = ({
 		})();
 
 		return preserveHtml ? html : html && sanitizer(html, { ADD_ATTR: ['target'] });
-	}, [content, preserveHtml, sanitizer, markedOptions, parseEmoji]);
+	}, [content, preserveHtml, sanitizer, markedOptions, parseEmoji, variant]);
```
KevLehman, line 113: 🤔

apps/meteor/packages/rocketchat-i18n/i18n/en.i18n.json
```diff
@@ -68,7 +68,7 @@
...
   "Accounts_BlockedUsernameList": "Blocked Username List",
   "Accounts_BlockedUsernameList_Description": "Comma-separated list of blocked usernames (case-insensitive)",
-  "Accounts_CustomFields_Description": "Should be a valid JSON where keys are the field names containing a dictionary of field settings. Example:<br/><code>{\n \"role\": {\n  \"type\": \"select\",\n  \"defaultValue\": \"student\",\n  \"options\": [\"teacher\", \"student\"],\n  \"required\": true,\n  \"modifyRecordField\": {\n   \"array\": true,\n   \"field\": \"roles\"\n  }\n },\n \"twitter\": {\n  \"type\": \"text\",\n  \"required\": true,\n  \"minLength\": 2,\n  \"maxLength\": 10\n }\n}</code> ",
+  "Accounts_CustomFields_Description": "Should be a valid JSON where keys are the field names containing a dictionary of field settings. Example:```{ \"role\": {  \"type\": \"select\",  \"defaultValue\": \"student\",  \"options\": [\"teacher\", \"student\"],  \"required\": true,  \"modifyRecordField\": {   \"array\": true,   \"field\": \"roles\"  } }, \"twitter\": {  \"type\": \"text\",  \"required\": true,  \"minLength\": 2,  \"maxLength\": 10,  \"searchable\": true }}```",
```
KevLehman, line 71: I think 🤔 this is for other type of custom fields. Are we using these?

KevLehman (changes requested):
There's also some "custom fields" specially made for livechat that i don't see here 👀 (endpoints) could you pls check that? Thanks!

apps/meteor/app/models/server/models/Users.js
cauefcr: :oh-no:

apps/meteor/app/livechat/server/api/v1/contact.js
cauefcr: sure, we can make one with new URLSearchParams({field:1,other:"string"}).toString(), for example, that leads to `field=1&other=string`, and the quoted line way to get the object to put on the query.

apps/meteor/client/components/MarkdownText.tsx
cauefcr: typescript complained that it was being used without being depended on

apps/meteor/app/lib/server/functions/validateCustomFields.js
KevLehman: 1. Yeah, but fieldValue seems to hold the actual value of the field. So by checking fieldValue !== true & false, aren't you validating that the value of the filed _is_ a boolean, instead of the searchable prop? (which is on "field.searchable"?)
2. Should we modify the user api on this PR? 👀 i thought this was for visitors, and visitors are not users

apps/meteor/app/livechat/server/api/v1/contact.js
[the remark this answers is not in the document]
KevLehman: 🤔 thought this was from raw model, and all raw models need await 👀 can you confirm? Maybe you're looking at other usages of this from a meteor model (or i'm crazy, which could be)

apps/meteor/packages/rocketchat-i18n/i18n/en.i18n.json
cauefcr: damn, as much as i like searchability it seems that it's not a word, fixing it.

apps/meteor/client/components/MarkdownText.tsx
cauefcr: it does affect the CI, breaking on eslint[react-hooks/exhaustive-deps](https://github.com/facebook/react/issues/14920)

apps/meteor/app/livechat/server/api/v1/contact.js
cauefcr: it gets URL encoded, so it's end up as `custom=a%3D1%26b%3D1`, very ugly but works.

apps/meteor/app/livechat/server/api/v1/contact.js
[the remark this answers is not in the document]
cauefcr: it is a raw model, and find on raw models returns a cursor, not a promise, finally understood the new model registering code to check this, but yeah, no await needed.

one of the find codes on BaseRaw is below, but they all return FindCursor:
`	find(query?: Filter<T>): FindCursor<ResultFields<T, C>>;`

apps/meteor/app/lib/server/functions/validateCustomFields.js
cauefcr: Reverting changes, you're correct.

apps/meteor/client/components/MarkdownText.tsx
KevLehman: No but, can you try updating with latest of develop? I'm curious why this is not failing on any other place 👀

apps/meteor/app/livechat/server/api/v1/contact.js
```diff
@@ -51,19 +51,30 @@ API.v1.addRoute(
...
+				if (!email && !phone && !custom) {
 					throw new Meteor.Error('error-invalid-params');
 				}
+				const customObj = Object.fromEntries(
+					Array.from(new URLSearchParams(custom)).filter(([k, v]) => typeof k === 'string' && typeof v === 'string'),
+				);
+
+				const foundCF = Object.fromEntries(
+					LivechatCustomField.find(
+						{ scope: 'visitor', searchable: true, _id: RegExp(Object.keys(customObj).join('|')) },
```
KevLehman, line 67: Instead of regex, use `$in`
```suggestion
						{ scope: 'visitor', searchable: true, _id: { $in: customObj } },
```

KevLehman: I'm pretty sure this one is a raw model 👀 so you need to `await` for it

apps/meteor/server/models/raw/LivechatVisitors.ts
```diff
@@ -159,11 +160,16 @@ export class LivechatVisitorsRaw extends BaseRaw<ILivechatVisitor> implements IL
...
 	 * Find visitors by their email or phone or username or name
 	 */
-	findPaginatedVisitorsByEmailOrPhoneOrNameOrUsername(
+	async findPaginatedVisitorsByEmailOrPhoneOrNameOrUsernameOrCustomField(
 		emailOrPhone: string,
 		nameOrUsername: RegExp,
 		options: FindOptions<ILivechatVisitor>,
-	): FindPaginated<FindCursor<ILivechatVisitor>> {
+	): Promise<FindPaginated<FindCursor<ILivechatVisitor>>> {
+		const allowedCF = LivechatCustomField.find({ scope: 'visitor', searchable: true }, { projection: { fields: { _id: 1 } } }).map(
```
KevLehman, line 168: Pls await this one 👀 I'm pretty sure this returns something async

apps/meteor/server/models/raw/LivechatVisitors.ts
```diff
@@ -178,6 +184,11 @@ export class LivechatVisitorsRaw extends BaseRaw<ILivechatVisitor> implements IL
 				{
 					username: nameOrUsername,
 				},
+				...(await allowedCF.toArray().then((cf) =>
+					cf.map((c: string) => {
+						return { [`livechatData.${c}`]: filter };
+					}),
```
KevLehman, line 190: You don't need the wrap function, you can return the { []: filter } directly

KevLehman (changes requested):

apps/meteor/client/components/MarkdownText.tsx
cauefcr: done

PR #26194, opened 2022-07-08
Regression: Fix Omnichannel not working after meteor update

KevLehman:

Fixed things:
- Omnichannel Directory
- Omnichannel Current Chats
- Auto Selection Algo
- Load Balance Algo
- Manual Selection Algo
- Livechat New Conversations

Other fixed things:
- Warning on fields deprecation
- Warning on "remove" deprecation
- Remove findAndModify usage

### DIFF
--- a/apps/meteor/app/models/server/models/LivechatDepartmentAgents.js
+++ b/apps/meteor/app/models/server/models/LivechatDepartmentAgents.js
@@ -1,4 +1,3 @@
-import { Meteor } from 'meteor/meteor';
 import _ from 'underscore';
 
 import { Base } from './_Base';
@@ -14,9 +13,6 @@ export class LivechatDepartmentAgents extends Base {
 		this.tryEnsureIndex({ departmentEnabled: 1 });
 		this.tryEnsureIndex({ agentId: 1 });
 		this.tryEnsureIndex({ username: 1 });
-
-		const collectionObj = this.model.rawCollection();
-		this.findAndModify = Meteor.wrapAsync(collectionObj.findAndModify, collectionObj);
 	}
 
 	findByDepartmentId(departmentId) {
@@ -96,7 +92,7 @@ export class LivechatDepartmentAgents extends Base {
 
 		const collectionObj = this.model.rawCollection();
 
-		const agent = Promise.await(collectionObj.findAndModify(query, sort, update));
+		const agent = Promise.await(collectionObj.findOneAndUpdate(query, update, { sort, returnNewDocument: 'after' }));
 		if (agent && agent.value) {
 			return {
 				agentId: agent.value.agentId,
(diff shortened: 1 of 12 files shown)


apps/meteor/app/models/server/models/LivechatRooms.js
```diff
@@ -37,7 +38,7 @@ export class LivechatRooms extends Base {
 
 	findLivechat(filter = {}, offset = 0, limit = 20) {
 		const query = Object.assign(filter, { t: 'l' });
-		return this.find(query, { sort: { ts: -1 }, offset, limit });
+		return this.findPaginated(query, { sort: { ts: -1 }, offset, limit });
```
sampaiodiego, line 41: I don't think there is a `findPaginated` in `Base` class.. I actually moved all queries that needed to use `findPaginated` to raw model

apps/meteor/app/models/server/models/LivechatRooms.js
```diff
@@ -1,9 +1,10 @@
 import s from 'underscore.string';
 import _ from 'underscore';
+import { Settings } from '@rocket.chat/models';
 
 import { Base } from './_Base';
 import Rooms from './Rooms';
-import Settings from './Settings';
+// import Settings from './Settings';
```
sampaiodiego, line 7:
```suggestion
```

apps/meteor/app/models/server/models/LivechatDepartmentAgents.js
```diff
@@ -188,7 +184,7 @@ export class LivechatDepartmentAgents extends Base {
 			},
 		};
 
-		const bot = this.findAndModify(query, sort, update);
+		const bot = this.findOneAndUpdate(query, update, { sort, returnNewDocument: 'after' });
```
sampaiodiego, line 187: this should use `rawCollection()` or this method/operation should be moved to a raw model

sampaiodiego (changes requested):

apps/meteor/app/models/server/models/LivechatRooms.js
KevLehman: Weird, this worked 🤔 let me change

apps/meteor/app/models/server/models/LivechatRooms.js
```diff
@@ -269,9 +264,8 @@ export class LivechatRooms extends Base {
 			},
 		};
 
-		const livechatCount = Settings.findAndModify(query, null, update);
-
-		return livechatCount.value.value;
+		const livechatCount = await Settings.findOneAndUpdate(query, update, { returnDocument: 'after' });
+		return livechatCount.value;
```
sampaiodiego, line 268: the correct return value here should be the same as before.. but since the return is not used right now, it doesn't break anything 😬 
```suggestion
		return livechatCount.value.value;
```

sampaiodiego (approved):

PR #26213, opened 2022-07-11
Chore: Remove method meteor call on message action

filipemarins:
## Further comments

Remove method call on the message action component


apps/meteor/app/ui-utils/client/lib/MessageAction.ts
```diff
@@ -11,16 +11,16 @@ import { TranslationKey } from '@rocket.chat/ui-contexts';
...
+import { APIClient } from '../../../utils/client';
+
+const getMessage = async (msgId: string): Promise<any> => {
+	try {
+		const { message } = await APIClient.get('/v1/chat.getMessage', { msgId });
+		return message;
+	} catch {
+		return null;
+	}
+};
```
ggazzo, line 23: we were rejecting if there was an error, and where we are suppressing the error, is that correct?

ggazzo (changes requested):

yash-rajpal (approved):

yash-rajpal (approved):

PR #26216, opened 2022-07-11
Regression: Replace contact center icon

filipemarins:
## Further comments

Replace contact center icon:

Before:
Screen Shot 2022-07-11 at 12 33 47

After:
Screen Shot 2022-07-11 at 12 30 49


tassoevan (approved):

ggazzo:
I still think the icons should be different

PR #26224, opened 2022-07-11
Chore: Plan tag

gabriellsh:

Now we only have one plan tag for all plans \/
image

### DIFF
--- a/apps/meteor/client/components/PlanTag.tsx
+++ b/apps/meteor/client/components/PlanTag.tsx
@@ -1,32 +1,30 @@
 import { Box, Tag } from '@rocket.chat/fuselage';
 import { useSafely } from '@rocket.chat/fuselage-hooks';
-import { useMethod } from '@rocket.chat/ui-contexts';
+import { useEndpoint } from '@rocket.chat/ui-contexts';
 import React, { ReactElement, useEffect, useState } from 'react';
+import { useQuery } from 'react-query';
 
 import { ILicenseTag } from '../../ee/app/license/definitions/ILicenseTag';
 
-function PlanTag(): ReactElement {
-	const [plans, setPlans] = useSafely(
-		useState<
-			{
-				name: string;
-				color: string;
-			}[]
-		>([]),
-	);
+type PlanTagType = {
+	name: string;
+	color: string;
+};
 
-	const getTags = useMethod('license:getTags');
+function PlanTag(): ReactElement {
+	const [plans, setPlans] = useSafely(useState<PlanTagType[]>([]));
+
+	const isEnterpriseEdition = useEndpoint('GET', '/v1/licenses.isEnterprise');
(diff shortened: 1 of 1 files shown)


apps/meteor/client/components/PlanTag.tsx
```diff
@@ -1,32 +1,30 @@
...
 
-	const getTags = useMethod('license:getTags');
+	const isEnterpriseEdition = useEndpoint('GET', '/v1/licenses.isEnterprise');
+	const { data } = useQuery(['licenses.isEnterprise'], () => isEnterpriseEdition(), {
+		refetchOnWindowFocus: false,
+	});
 
 	useEffect(() => {
 		const developmentTag = process.env.NODE_ENV === 'development' ? { name: 'development', color: '#095ad2' } : null;
+		const enterpriseTag = data?.isEnterprise ? { name: 'enterprise', color: '#095ad2' } : null;
```
tassoevan, line 24: Is this color `primary-600`? Maybe use a reference instead of a hardcoded value.

apps/meteor/client/components/PlanTag.tsx
```diff
@@ -1,32 +1,30 @@
...
-				color: string;
-			}[]
-		>([]),
-	);
+	const [plans, setPlans] = useSafely(useState<PlanTagType[]>([]));
 
-	const getTags = useMethod('license:getTags');
+	const isEnterpriseEdition = useEndpoint('GET', '/v1/licenses.isEnterprise');
+	const { data } = useQuery(['licenses.isEnterprise'], () => isEnterpriseEdition(), {
+		refetchOnWindowFocus: false,
```
tassoevan, line 19: Is this really necessary? Maybe using a positive value for `staleTime` will do better.

apps/meteor/client/components/PlanTag.tsx
```diff
@@ -1,32 +1,30 @@
...
+		const enterpriseTag = data?.isEnterprise ? { name: 'enterprise', color: '#095ad2' } : null;
 
-		const fetchTags = async (): Promise<void> => {
-			const tags = await getTags();
-			setPlans([developmentTag, ...tags].filter(Boolean) as ILicenseTag[]);
-		};
-
-		fetchTags();
-	}, [getTags, setPlans]);
+		setPlans([developmentTag, enterpriseTag].filter(Boolean) as ILicenseTag[]);
```
tassoevan, line 26:
```suggestion
		setPlans([developmentTag, enterpriseTag].filter((tag): tag is ILicenseTag => Boolean(tag)));
```

This is such a common pattern that maybe we could make a helper function like this:

```ts
const isTruthy = <T>(x: T | null | undefined | 0 | false | ''): x is T => Boolean(x);
```

tassoevan (changes requested):

tassoevan (approved):

PR #26234, opened 2022-07-12
Regression: Don't open mdm feature modal on registration page

yash-rajpal:
## Issue(s)

MDM Feature Modal was opening on user registration page
image


apps/meteor/client/startup/banners.ts
```diff
@@ -15,14 +15,20 @@ const fetchInitialBanners = async (): Promise<void> => {
...
+						imperativeModal.open({
+							component: DeviceManagementFeatureModal,
+							props: {
+								close: imperativeModal.close,
+							},
+						});
+					}, 2000);
+					computation.stop();
+				}
+			});
```
ggazzo, line 31: would you mind mind return if username is empty?

ggazzo (changes requested):

apps/meteor/client/startup/banners.ts
yash-rajpal: Done!

ggazzo (approved):

PR #26236, opened 2022-07-12
Regression: User avatar not updating on new message template

filipemarins:
## Further comments

Fixes:
Message sequential after message thread preview
Avatar not updating after new message template

Passing an etag to user avatar component will force the re-render when avatar etag change.

Before (message sequential):
image

After (message sequential):
image


gabriellsh (changes requested):

apps/meteor/client/components/avatar/RoomAvatar.tsx
```diff
@@ -13,15 +15,19 @@ type RoomAvatarProps = {
...
 		t?: string;
+		uids?: string[];
 		avatarETag?: string;
 	};
 };
 
 const RoomAvatar = function RoomAvatar({ room, ...rest }: RoomAvatarProps): ReactElement {
 	const getRoomPathAvatar = useRoomAvatarPath();
-	const { url = getRoomPathAvatar(room), ...props } = rest;
+	const { avatarETag } = usePresence(room.uids?.filter((uid) => uid !== room.u?._id)[0]) || room;
```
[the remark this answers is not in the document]
filipemarins: Yes, this is the Room avatar, but this is the component used to handle the direct messages. I am getting the user id of the room, jus to get the user presence.

hugocostadev (changes requested):
Hi @filipemarins ,

I found some issues... 

1. Using the hook `usePresence` only trigger if the user is not offline, if the user is offline we found several inconsistence in user Avatar, including on uploading a new one

2. Because of the limitation of usePresence, I suggest creating a hook only for avatar that gets the `avatarEtag` somehow through another origin, and always that the avatar changes, the hook will trigger all places that the avatar should change.

3. With your current solution  (when the user is online) the avatars got updated but one place did not... View profile through a room still shows the previous avatar

4. Also, be careful when using array index as a `key` prop, sometimes the original order could change and you lose reference

apps/meteor/client/components/RoomAutoComplete/Avatar.tsx
```diff
@@ -1,16 +1,17 @@
+import { RoomType } from '@rocket.chat/apps-engine/definition/rooms';
 import { Options } from '@rocket.chat/fuselage';
 import React, { FC } from 'react';
 
 import RoomAvatar from '../avatar/RoomAvatar';
 
 type AvatarProps = {
 	value: string;
-	type: string;
+	type: RoomType;
```
gabriellsh, line 9: Please, use `IRoom['type']` from `@rocket.chat/core-typings`. This type is from apps-engine.

apps/meteor/client/components/avatar/DirectRoomAvatar.tsx
```diff
@@ -0,0 +1,18 @@
+import { useUserId } from '@rocket.chat/ui-contexts';
+import React, { memo, ReactElement, ComponentProps, useMemo } from 'react';
+
+import { useEndpointData } from '../../hooks/useEndpointData';
+import RoomAvatar from './RoomAvatar';
+
+const DirectRoomAvatar = ({ room, ...rest }: ComponentProps<typeof RoomAvatar>): ReactElement => {
+	const ownId = useUserId();
+	const otherId = room.uids?.find((uid) => uid !== ownId);
+	const { value: data } = useEndpointData(
```
gabriellsh, line 10: I think we are better of using the usePresence hook. Even though it only brings offline users, getting the info from the endpoint won't be reactive, and we'll have the same problem again.

apps/meteor/client/components/avatar/UserAvatar.tsx
```diff
@@ -12,7 +13,11 @@ type UserAvatarProps = Omit<BaseAvatarProps, 'url' | 'title'> & {
 
 const UserAvatar: FC<UserAvatarProps> = ({ username, etag, ...rest }) => {
 	const getUserAvatarPath = useUserAvatarPath();
-	const { url = getUserAvatarPath(username, etag), ...props } = rest;
+	const { value: data } = useEndpointData(
```
gabriellsh, line 16: same here

gabriellsh (changes requested):

ggazzo:
PARA DE ATUALIZAR SEU DOENTE

ggazzo:
E REVISA

ggazzo:
I'm closing this pr because it didn't fix the problem, and we already have a new solution in progress.

PR #26239, opened 2022-07-12
Regression: Removed CE watermark from VoipFooter

aleksandernsilva:

image (5)

### DIFF
--- a/apps/meteor/client/sidebar/footer/voip/index.tsx
+++ b/apps/meteor/client/sidebar/footer/voip/index.tsx
@@ -2,14 +2,13 @@ import type { VoIpCallerInfo } from '@rocket.chat/core-typings';
 import { useEndpoint, useTranslation } from '@rocket.chat/ui-contexts';
 import React, { ReactElement, useCallback, useMemo, useState } from 'react';
 
-import { useHasLicenseModule } from '../../../../ee/client/hooks/useHasLicenseModule';
 import { useVoipFooterMenu } from '../../../../ee/client/hooks/useVoipFooterMenu';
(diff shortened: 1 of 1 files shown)


apps/meteor/client/sidebar/footer/voip/index.tsx
```diff
@@ -104,8 +103,7 @@ export const VoipFooter = (): ReactElement | null => {
 			dispatchEvent={dispatchEvent}
 			openedRoomInfo={openedRoomInfo}
 			anonymousText={t('Anonymous')}
-			isEnterprise={isEE === true}
-			children={<SidebarFooterWatermark />}
+			isEnterprise={isEnterprise}
 			options={options}
```
ggazzo, line 107: I purposely put this watermark there, in my view it makes no sense to remove the watermark, I will ask for an approver from the pm of digital workplace @milton-rucks

engelgabriel: I don't think we should have this distinction

aleksandernsilva:
I have updated the description of this pull request to clarify the intent and scope of this change.
fyi @ggazzo

engelgabriel:
Now it makes more sense, I agree with the change.

tassoevan (approved):

milton-rucks (approved):

apps/meteor/client/sidebar/footer/voip/index.tsx
caskelly: I've checked with Renato and Milton and they're aligned.  This change is approved from Product.  An important note here - the overlay of the watermark is temporary when ringer is displayed.  The watermark is not being removed when VoIP is enabled.

ggazzo (approved):

PR #26241, opened 2022-07-12
Regression: Special characters on phone number

tiagoevanp:

PR Includes:
- Keep focus on phone input of dial pad
- Handle submit with "Enter" key
- Remove mask and mandatory "+" char
- Long press for "0"/"+" button


apps/meteor/ee/client/voip/modal/DialPad/hooks/useDialPad.tsx
```diff
@@ -41,18 +47,33 @@ export const useDialPad = ({ initialValue, errorMessage }: DialPadProps): DialPa
...
+		[clearErrors, setValue, value],
+	);
+
+	const handlePadButtonLongPressed = useCallback(
+		(digit: string | number): void => {
+			if (digit === 0) {
+				return setValue('PhoneInput', `${value}+`);
+			}
+
+			return setValue('PhoneInput', value + digit);
```
aleksandernsilva, line 75: Won't long pressing any digit other than 0 add something like `ABC`? It doesn't seem very useful.

apps/meteor/ee/client/voip/modal/DialPad/hooks/useDialPad.tsx
```diff
@@ -65,26 +86,41 @@ export const useDialPad = ({ initialValue, errorMessage }: DialPadProps): DialPa
...
+		},
+		[clearErrors, onChange],
+	);
+
 	useEffect(() => {
 		setError('PhoneInput', { message: errorMessage });
 	}, [setError, errorMessage]);
 
 	useEffect(() => {
 		setDisabled(!value);
```
aleksandernsilva, line 106: This may not work as intended if the value is falsy. Ex.: 0

apps/meteor/ee/client/voip/modal/DialPad/hooks/useLongPress.tsx
```diff
@@ -0,0 +1,63 @@
...
+export function useLongPress(onLongPress: () => void, options?: { onClick?: () => void; threshold?: number }): UseLongPressResult {
+	const action = useRef<'LONG_PRESS' | 'CLICK'>();
+	const isLongPress = useRef(false);
+	const timerRef = useRef<NodeJS.Timeout>();
+
+	const startPressTimer = useMutableCallback((): void => {
+		isLongPress.current = false;
+		timerRef.current = setTimeout(() => {
+			isLongPress.current = true;
+			action.current = 'LONG_PRESS';
```
aleksandernsilva, line 21: `isLongPress` and `action.current === 'LONG_PRESS'` seem to store the same information. Can we remove one of them?

apps/meteor/ee/client/voip/modal/DialPad/hooks/useLongPress.tsx
```diff
@@ -0,0 +1,63 @@
...
+		startPressTimer();
+	}, [startPressTimer]);
+
+	const handleOnTouchEnd = useCallback((): void => {
+		if (action.current === 'LONG_PRESS') {
+			return;
+		}
+
+		clearTimeout(timerRef.current);
+	}, []);
```
aleksandernsilva, line 54: I believe we should provide the option for the user to keep the original methods functionality. Meaning `options` should also be able to receive `onMouseDown`, `onMouseUp` etc. Each of them should be called at the end of the "handle" functions and receive the original `event` as parameter.

apps/meteor/ee/client/voip/modal/DialPad/hooks/useDialPad.tsx
```diff
@@ -41,18 +47,33 @@ export const useDialPad = ({ initialValue, errorMessage }: DialPadProps): DialPa
...
 			setValue('PhoneInput', value + digit);
 		},
+		[clearErrors, setValue, value],
+	);
+
+	const handlePadButtonLongPressed = useCallback(
+		(digit: string | number): void => {
+			if (digit === 0) {
+				return setValue('PhoneInput', `${value}+`);
+			}
```
aleksandernsilva, line 73: From what I see the `digit` should already be + in this scenario. Isn't this redundant?

aleksandernsilva (changes requested):

apps/meteor/ee/client/voip/modal/DialPad/hooks/useDialPad.tsx
tiagoevanp: I was developing same behavior that we found on competitors dial pad... only zero is important to us as a long press at this point.

apps/meteor/ee/client/voip/modal/DialPad/hooks/useDialPad.tsx
tiagoevanp: Tested value is a `string` type... are you sure?

apps/meteor/ee/client/voip/modal/DialPad/hooks/useDialPad.tsx
tiagoevanp: image
Here is digits... there is only 0

apps/meteor/ee/client/voip/modal/DialPad/hooks/useDialPad.tsx
aleksandernsilva: It seems not. You're correct. I was thinking using numbers. 🤦

apps/meteor/ee/client/voip/modal/DialPad/hooks/useLongPress.tsx
tiagoevanp: No, because of handleOnClick control... I think both are necessary on this case 😞

tiagoevanp: No, you are correct... thanks!

apps/meteor/ee/client/voip/modal/DialPad/PadButton.tsx
```diff
@@ -1,31 +1,42 @@
...
-		</Box>
-	</Button>
-);
+	children: PadDigit;
+	onClickPadButton: (digit: PadDigit) => void;
+	onLongPressPadButton: (digit: PadDigit) => void;
+}): ReactElement => {
+	const { onClick, onMouseDown, onMouseUp, onTouchStart, onTouchEnd } = useLongPress(() => onLongPressPadButton(children), {
+		onClick: () => onClickPadButton(children),
+	});
```
aleksandernsilva, line 18: - First I believe we can improve readability destructuring the children array, like so `const [digit, longDigit] = children`. 
- Second I think you should return only the appropriate digit, instead of the tuple. Ex.: `onClickPadButton(digit)` and `onLongPressPadButton(longDigit)`. This way we don't need to share this implementation detail to other parts of the code.

apps/meteor/ee/client/voip/modal/DialPad/hooks/useDialPad.tsx
```diff
@@ -41,17 +48,32 @@ export const useDialPad = ({ initialValue, errorMessage }: DialPadProps): DialPa
...
+		[clearErrors, setValue, value],
+	);
+
+	const handlePadButtonLongPressed = useCallback(
+		(digit: PadDigit): void => {
+			if (digit[0] === '0') {
+				return setValue('PhoneInput', value + digit[1]);
+			}
+
+			setValue('PhoneInput', value + digit[0]);
```
aleksandernsilva, line 76: After the changes I recommended earlier, here you can just ignore digits different than `0`.
```
if (longDigit !== '0') {
  return;
}

// Do what you have to do
```

aleksandernsilva (changes requested):

apps/meteor/ee/client/voip/modal/DialPad/hooks/useDialPad.tsx
tiagoevanp: ok!

apps/meteor/ee/client/voip/modal/DialPad/PadButton.tsx
tiagoevanp: ok!

aleksandernsilva (approved):

aleksandernsilva (approved):

KevLehman (approved):

PR #26242, opened 2022-07-13
Regression: Fix scheduler not working

tapiarafael:
Fix Agenda package to work with new Meteor version update

### DIFF
--- a/packages/agenda/src/Agenda.ts
+++ b/packages/agenda/src/Agenda.ts
@@ -597,7 +597,7 @@ export class Agenda extends EventEmitter {
 
 		// Don't try and access MongoDB if we've lost connection to it.
 		// @ts-ignore
-		const s = this._mdb.s || this._mdb.db.s;
+		const s = this._mdb.s.client || this._mdb.db.s.client;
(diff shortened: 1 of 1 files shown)


packages/agenda/src/Agenda.ts
```diff
@@ -597,7 +597,7 @@ export class Agenda extends EventEmitter {
 
 		// Don't try and access MongoDB if we've lost connection to it.
 		// @ts-ignore
-		const s = this._mdb.s || this._mdb.db.s;
+		const s = this._mdb.s.client || this._mdb.db.s.client;
```
AllanPazRibeiro, line 600: maybe this const could be change to 
```suggestion
		const client = this._mdb.s.client || this._mdb.db.s.client;
```

d-gubert: I think we'd need a bigger effort to understand and refactor the naming - the name is not clear now, but we need to be careful not to be misleading to our future selves 🙈 

Also keep in mind this is not _our_ code, we absorbed it from the original repo, so we don't have a deep understanding of it yet

d-gubert (approved):

PR #26251, opened 2022-07-13
Regression: Remove 4.0 version banner

hugocostadev:

Created a migration to disable and dismiss for all users the old 4.0 version banner.
It happened when a new admin user has been added.


filipemarins (approved):

sampaiodiego (changes requested):
I'm sorry, but where is this banner coming from?

I'd like to understand as this doesn't look like to be the best implementation for this purpose.

hugocostadev:
> I'm sorry, but where is this banner coming from?
> 
> I'd like to understand as this doesn't look like to be the best implementation for this purpose.

Hi @sampaiodiego, this banner was added in a previous migration ([v231](https://github.dev/RocketChat/Rocket.Chat/blob/4.8.1/apps/meteor/server/startup/migrations/v231.ts)) and the migration [v232](https://github.dev/RocketChat/Rocket.Chat/blob/4.8.1/apps/meteor/server/startup/migrations/v232.ts) tried to dismiss, but the problem is that the v232 migration just dismisses for admins if a user is added as admin after this migration execution the user is able to see the banner. This issue can be verified at:  https://unstable.rocket.chat/

apps/meteor/server/startup/migrations/v278.ts
```diff
@@ -0,0 +1,45 @@
...
+
+		const isEE = isEnterprise();
+
+		if (!isEE && (isCustomOAuthEnabled || LDAPEnabled || SAMLEnabled)) {
+			return;
+		}
+
+		const users = Promise.await(Users.find().toArray());
+
+		const banners = users.map((a) => Promise.await(Banner.getBannersForUser(a._id, BannerPlatform.Web))).flat();
```
sampaiodiego, line 29: we also should avoid queries in loops like whenever it is possible.. in this case, after fetching 500k+ users, a new query will be performed for every user, synchronously (due to `Promise.await`).. so let's say the query takes only 20ms to execute, for 500k users this loop would take at least 10.000 seconds.

apps/meteor/server/startup/migrations/v278.ts
```diff
@@ -0,0 +1,45 @@
...
+		// @ts-ignore
+		const authBanner = banners.find((b) => b.view.blocks[0].text.text === oldMsg || b.view.blocks[0].text.text === newMsg);
+
+		if (!authBanner) {
+			return;
+		}
+
+		Banner.disable(authBanner._id);
+
+		users.map((a) => Banner.dismiss(a._id, authBanner._id));
```
sampaiodiego, line 43: same as before, but in this case it is a write operation, that should take more than 20ms to execute.

sampaiodiego (changes requested):
I've explained what we should avoid and why.. 

so please, you can change everything starting from line 26, and perform only the `Banner` from my previous comment:

```ts
Banners.updateOne({
  'view.blocks.0.text.text': /authentication\-changes/,
  active: { $ne: false }
}, {
  $set: {
    active: false,
    inactivedAt: new Date(),
  },
})
```

hugocostadev:
> I've explained what we should avoid and why..
> 
> so please, you can change everything starting from line 26, and perform only the `Banner` from my previous comment:
> 
> ```ts
> Banners.updateOne({
>   'view.blocks.0.text.text': /authentication\-changes/,
>   active: { $ne: false }
> }, {
>   $set: {
>     active: false,
>     inactivedAt: new Date(),
>   },
> })
> ```

Perfect Diego, I got it! Thanks so much for your explanations, indeed this is a much cleaner and wise solution

sampaiodiego (approved):

PR #26263, opened 2022-07-14
Regression: Do not allow non-owners edit rooms or add/remove users on federated rooms

MarcosSpessatto:


alansikora (dismissed):

apps/meteor/client/views/room/hooks/useUserInfoActions/actions/useRemoveUserAction.tsx
```diff
@@ -16,9 +17,13 @@ import { getRoomDirectives } from '../../../lib/getRoomDirectives';
 export const useRemoveUserAction = (user: Pick<IUser, '_id' | 'username'>, rid: IRoom['_id'], reload?: () => void): Action | undefined => {
 	const t = useTranslation();
 	const room = useUserRoom(rid);
+	const currentUser = useUser();
 	const { _id: uid } = user;
 
-	const userCanRemove = usePermission('remove-user', rid);
+	const hasPermissionToRemove = usePermission('remove-user', rid);
+	const userCanRemove = isRoomFederated(room as Partial<IRoom>)
+		? Federation.canEdit(currentUser as IUser, room as IRoom)
```
sampaiodiego, line 25: did you have to add the type castings because they can be `null`/`undefined`? what if they're really `undefined`, then your code could break =\ you should properly validate the values instead of type casting
```suggestion
	const userCanRemove = isRoomFederated(room)
		? Federation.canEdit(currentUser, room)
```

apps/meteor/client/views/room/contextualBar/Info/RoomInfo/RoomInfoWithData.js
```diff
@@ -68,9 +72,13 @@ const RoomInfoWithData = ({ rid, openEditing, onClickBack, onEnterRoom, resetSta
...
 	);
 
-	const canDelete = usePermission(type === 'c' ? 'delete-c' : 'delete-p', rid);
-	const canEdit = usePermission('edit-room', rid);
-	const canConvertRoomToTeam = usePermission('create-team');
+	const isFederated = isRoomFederated(room);
+	const hasPermissionToDelete = usePermission(type === 'c' ? 'delete-c' : 'delete-p', rid);
+	const hasPermissionToEdit = usePermission('edit-room', rid);
+	const hasPermissionToConvertRoomToTeam = usePermission('create-team');
+	const canDelete = isFederated ? Federation.canEdit(user, room) : hasPermissionToDelete;
```
sampaiodiego, line 79: should this be something like `Federation.canRemove` ?

apps/meteor/client/views/room/contextualBar/Info/RoomInfo/RoomInfoWithData.js
```diff
@@ -68,9 +72,13 @@ const RoomInfoWithData = ({ rid, openEditing, onClickBack, onEnterRoom, resetSta
...
+	const canEdit = isFederated ? Federation.canEdit(user, room) : hasPermissionToEdit;
```
sampaiodiego, line 80: I think even for federated rooms we should check the permission, wdyt?

sampaiodiego (changes requested):
I added some comments, but I was thinking about the root cause of the issue.. maybe I need an explanation what the issue is 😬 

because in theory only the room owner or an admin will have the permission to edit a room already.. maybe in case of a federated room there will be no subscription (do they have subscriptions?), so the owner check would fail, but for admin it not fail.. 🤔

MarcosSpessatto:
> I added some comments, but I was thinking about the root cause of the issue.. maybe I need an explanation what the issue is 😬
> 
> because in theory only the room owner or an admin will have the permission to edit a room already.. maybe in case of a federated room there will be no subscription (do they have subscriptions?), so the owner check would fail, but for admin it not fail.. 🤔

Good point, so the first answer is: yes, they do have subscriptions, it's at the end of the day, a regular rocket.chat room, there's even no specific type to describe it, it's using the same as we use today (p, c, d, etc).

Also, in the case of the federated rooms, even the admins should now be able to edit that, I mean, for now, until we do not support any kind of permission/role from the Matrix side, we must allow only and only the owner to edit the things.

apps/meteor/client/views/room/contextualBar/Info/RoomInfo/RoomInfoWithData.js
MarcosSpessatto: Very good question, but yes, you are right. Probably in the near future when we'll start to have some sort of overlap between RC roles and Matrix roles, we need to revisit that, but for now, I agree 👍

ggazzo (approved):

PR #26272, opened 2022-07-14
fix: send message for a room without subscribe

carlosrodrigues94:

### DIFF
--- a/apps/meteor/app/api/server/v1/chat.js
+++ b/apps/meteor/app/api/server/v1/chat.js
@@ -217,6 +217,21 @@ API.v1.addRoute(
 				throw new Meteor.Error('error-invalid-params', 'The "message" parameter must be provided.');
 			}
 
+			const { rid: roomId } = this.bodyParams.message;
+
+			const subscriptionOptions = {
+				projection: {
+					blocked: 1,
+					blocker: 1,
+				},
+			};
+
+			const subscription = Subscriptions.findOneByRoomIdAndUserId(roomId, this.userId, subscriptionOptions);
+
+			if (!subscription) {
+				return API.v1.unauthorized('Test message for 401');
+			}
+
 			const sent = executeSendMessage(this.userId, this.bodyParams.message);
 			const [message] = normalizeMessagesForUser([sent], this.userId);
 


LucianoPierdona (approved):

matheusbsilva137 (approved):

apps/meteor/app/api/server/v1/chat.js
```diff
@@ -217,6 +217,14 @@ API.v1.addRoute(
 				throw new Meteor.Error('error-invalid-params', 'The "message" parameter must be provided.');
 			}
 
+			const { rid: roomId } = this.bodyParams.message;
```
ggazzo, line 220:
```suggestion
			const { rid } = this.bodyParams.message;
```

apps/meteor/app/api/server/v1/chat.js
```diff
@@ -217,6 +217,14 @@ API.v1.addRoute(
...
+
+			const subscription = Subscriptions.findOneByRoomIdAndUserId(roomId, this.userId);
```
ggazzo, line 222:
```suggestion
			const subscription = Subscriptions.findOneByRoomIdAndUserId(rid, this.userId);
```

apps/meteor/app/api/server/v1/chat.js
```diff
@@ -217,6 +217,14 @@ API.v1.addRoute(
...
+
+			if (!subscription) {
```
ggazzo, line 224: why not `canSendMessage` function instead?

ggazzo (changes requested):

apps/meteor/app/api/server/v1/chat.js
matheusbsilva137: `canSendMessage` or `canAccessRoom` would return `true` to all public rooms (even when the user is not a member of the room). Since in this case we'd like to block users from sending messages to rooms they are not part of, I guess none of these would fit 🤔

ggazzo: correct me if i'm wrong, but being able to send a message on a channel i don't belong ,yet, (but have permission) should cause an auto join. this is more like a feature than a bug in my opinion

maybe you have more information on what caused or caused this issue/confusion?

just an image to describe where this feature could be used:
image

ggazzo (changes requested):
even if this task makes sense (depending on the answer of who identified/raised the problem), I'd like to see an e2e test case showing how this behavior should work.

gabriellsh:
Converted to draft since we're not sure this is the correct Fix. This way we don't accidentaly merge it.

PR #26273, opened 2022-07-14
Regression: Search on Member List

tassoevan:


sampaiodiego (dismissed):
looks like there are other places calling `findPaginatedByActiveUsersExcept` without providing the `searchFields` param.. I can change them later, or if you can update this PR to fix all places it would be awesome.

tassoevan:
I really wanted to add a test ensure this changes on `dm.members`, but geez... It's so slow to write tests without mocking the models.

sampaiodiego (approved):

sampaiodiego (approved):

PR #26276, opened 2022-07-15
[FIX] Unable to send voice recording to Whatsapp

murtaza98:
## Issue(s)

Apparently, WhatsApp doesn't accept an mp3 file without a valid magic number in its headers and RC doesn't add those headers to the mp3 file that it generates :( . This PR fixes that issue and makes it possible for agents to send a voice note to WhatsApp users
## Further comments

PR: https://app.clickup.com/t/2rq7jm6


KevLehman (approved):

apps/meteor/app/ui/client/lib/recorderjs/audioEncoder.js
```diff
@@ -40,7 +40,10 @@ class AudioEncoder extends Emitter {
 	handleWorkerMessage = (event) => {
 		switch (event.data.command) {
 			case 'end': {
-				const blob = new Blob(event.data.buffer, { type: 'audio/mpeg' });
+				// prepend mp3 magic number to the buffer
+				const magicNoPrefix = new Int8Array([73, 68, 51, 3, 0, 0, 0, 0, 0, 0]);
```
KevLehman, line 44: Just a quick thing: can you check the buffer doesn't have the magic number before preppending? That way if the provider changes and preppends it, we won't have problems 👀

KevLehman (changes requested):

apps/meteor/app/ui/client/lib/recorderjs/audioEncoder.js
murtaza98: I don't think this check is necessary because we'll always have only audio-related data in `event.data.buffer` & not any headers

KevLehman (approved):

KevLehman (approved):

PR #26278, opened 2022-07-15
Regression: Fix app icons breaking UI

murtaza98:

### DIFF
--- a/apps/meteor/client/components/RoomIcon/OmnichannelRoomIcon/provider/OmnichannelRoomIconProvider.tsx
+++ b/apps/meteor/client/components/RoomIcon/OmnichannelRoomIcon/provider/OmnichannelRoomIconProvider.tsx
@@ -24,34 +24,46 @@ export const OmnichannelRoomIconProvider: FC = ({ children }) => {
 
 	return (
 		<OmnichannelRoomIconContext.Provider
-			value={useMemo(
-				() => ({
+			value={useMemo(() => {
+				let snapshot: AsyncState<string> = {
+					phase: AsyncStatePhase.LOADING,
+					value: undefined,
+					error: undefined,
+				};
+
+				return {
 					queryIcon: (
 						app: string,
 						iconName: string,
 					): [subscribe: (onStoreChange: () => void) => () => void, getSnapshot: () => AsyncState<string>] => [
-						(callback): (() => void) => OmnichannelRoomIcon.on(`${app}-${iconName}`, callback),
-						(): AsyncState<string> => {
-							const icon = OmnichannelRoomIcon.get(app, iconName);
+						(callback): (() => void) =>
+							OmnichannelRoomIcon.on(`${app}-${iconName}`, () => {
+								// First, let's update the snapshot
+								const icon = OmnichannelRoomIcon.get(app, iconName);
 
-							if (!icon) {
-								return {
-									phase: AsyncStatePhase.LOADING,
-									value: undefined,
-									error: undefined,
-								};
-							}
+								if (icon) {
+									snapshot = {
+										phase: AsyncStatePhase.RESOLVED,
+										value: icon,
+										error: undefined,
+									};
+								} else {
+									snapshot = {
+										phase: AsyncStatePhase.LOADING,
+										value: undefined,
+										error: undefined,
+									};
+								}
 
-							return {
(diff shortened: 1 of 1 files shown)


apps/meteor/client/components/RoomIcon/OmnichannelRoomIcon/lib/OmnichannelRoomIcon.ts
```diff
@@ -18,20 +18,26 @@ const OmnichannelRoomIcon = new (class extends Emitter {
...
-					NAMESPACE: 'http://www.w3.org/2000/svg',
-					USE_PROFILES: { svg: true, svgFilters: true },
-				})
-					.replace(`<svg`, `<symbol id="${appId}-${icon}"`)
-					.replace(`</svg>`, '</symbol>'),
-			);
-			this.emit('change');
-			this.emit(`${appId}-${icon}`);
-		});
+		APIClient.send(`/apps/public/${appId}/get-sidebar-icon?icon=${icon}`, 'GET')
```
murtaza98, line 21: Note: we cannot use .get() now since it is accepting only JSON response and fails if it gets a non-JSON response

KevLehman: Is this problem on APIClient side or in apps side? 👀 (I mean, where should this be fixed?)

tassoevan: It's a general issue of the REST API, as it was devised as just delivering JSON responses. `APIClient.get()` assumes the response is encoded as one.

But if this endpoint is supposed to respond **only** SVGs, which are text-based, it's fine to encode it as an JSON value, as strings are valid i.e. `JSON.parse("\"<svg>...</svg>\"") === "<svg>...</svg>"` (mind the escaped quotes).

KevLehman: So, we can leave the .send for now until we update the apps to encode the svg as JSON 👀 
wdyt of adding a note for it, @murtaza98 ? 👀

apps/meteor/client/components/RoomIcon/OmnichannelRoomIcon/provider/OmnichannelRoomIconProvider.tsx
```diff
@@ -24,34 +24,55 @@ export const OmnichannelRoomIconProvider: FC = ({ children }) => {
...
+						return {
+							phase: AsyncStatePhase.RESOLVED,
+							value: icon,
+							error: undefined,
+						};
+					}
+
+					return {
+						phase: AsyncStatePhase.LOADING,
+						value: undefined,
```
tiagoevanp, line 41: It's really necessary to send undefined as a value?

tiagoevanp (changes requested):

apps/meteor/client/components/RoomIcon/OmnichannelRoomIcon/provider/OmnichannelRoomIconProvider.tsx
tassoevan: It is:

```ts
export type AsyncState<T> =
	| { phase: AsyncStatePhase.LOADING; value: undefined; error: undefined }
	| { phase: AsyncStatePhase.LOADING; value: T; error: undefined }
	| { phase: AsyncStatePhase.LOADING; value: undefined; error: Error }
	| { phase: AsyncStatePhase.RESOLVED; value: T; error: undefined }
	| { phase: AsyncStatePhase.UPDATING; value: T; error: undefined }
	| { phase: AsyncStatePhase.REJECTED; value: undefined; error: Error };
```

apps/meteor/client/components/RoomIcon/OmnichannelRoomIcon/lib/OmnichannelRoomIcon.ts
murtaza98: > So, we can leave the .send for now until we update the apps to encode the svg as JSON eyes wdyt of adding a note for it, @murtaza98 ? eyes

Done :heavy_check_mark:

PR #26282, opened 2022-07-15
[FIX] Add Offline License Endpoint

LucianoPierdona:

This PR updates the endpoint to add a license


ggazzo (changes requested):
Could you convert the file to TS? in that way we can prevent even more problems

apps/meteor/client/views/admin/info/OfflineLicenseModal.tsx
```diff
@@ -35,8 +41,8 @@ const OfflineLicenseModal = ({ onClose, license, licenseStatus, ...props }) => {
 	const handleApplyLicense = useMutableCallback(async () => {
 		setIsUpdating(true);
 		setLastSetLicense(newLicense);
-		const data = await addLicense({ license: newLicense });
-		if (data.success) {
+		const data = (await addLicense({ license: newLicense })) as unknown as { success: true };
```
tassoevan, line 44: No need to check the response object because the `APIClient` implicitly consumes the `success` field and throws an error if `response.success !== true`. In any case you need to handle the request error, it's a better option to use `useEndpoint` hook directly.

tassoevan (changes requested):

tassoevan (approved):

ggazzo (dismissed):

hugocostadev (approved):

PR #26283, opened 2022-07-15
Regression: Matrix Federation regressions

MarcosSpessatto:


apps/meteor/app/federation-v2/server/application/RoomServiceSender.ts
```diff
@@ -148,18 +148,50 @@ export class FederationRoomServiceSender {
...
+		if (!internalRoom.federated) {
 			return;
 		}
+		const tryingToAddNewFederatedUser = typeof internalUser === 'string';
+		if (tryingToAddNewFederatedUser) {
+			throw new Error('error-this-is-an-ee-feature');
+		}
 
-		if (!internalRoom.federated) {
+		const invitee = await this.rocketUserAdapter.getFederatedUserByInternalId((internalUser as IUser)._id);
```
sampaiodiego, line 160: the function signature says `internalUser` can be either `IUser | string`, so why are you type casting as `IUser`? should it be typed as `IUser` only on the function signature?

apps/meteor/server/methods/createDirectMessage.js
```diff
@@ -89,6 +90,11 @@ export function createDirectMessage(usernames, userId, excludeSelf = false) {
 	if (excludeSelf && hasPermission(this.userId, 'view-room-administration')) {
 		options.subscriptionExtra = { open: true };
 	}
+	try {
+		callbacks.run('federation.beforeCreateDirectMessage', roomUsers);
+	} catch (error) {
+		throw new Meteor.Error(error?.message);
```
sampaiodiego, line 96: any benefit in throwing a `Meteor.Error` here? just would like to understand if this try/catch is really needed

sampaiodiego (changes requested):

apps/meteor/app/federation-v2/server/application/RoomServiceSender.ts
MarcosSpessatto: Yes, you are right, I forgot to move the validation to the top of the function, but I have already fixed that on my last commit. When the user is a string it should not be possible to add the user.

https://github.com/RocketChat/Rocket.Chat/pull/26283/commits/9200b3d67ea841f6f42042af4324b01fccc30861#diff-818e7b29c0e27fb8420e16fc4c4b9f9ae4cae69e6c4c89f011018da429006edcR155

apps/meteor/server/methods/createDirectMessage.js
MarcosSpessatto: The idea was to be able to send the right error to the front-end, I tried with `Error` only, but it looked like it was not working as expected, any idea?

apps/meteor/app/federation-v2/client/Federation.ts
```diff
@@ -1,4 +1,5 @@
-import { ValueOf } from '@rocket.chat/core-typings';
+import { IRoom, IUser, ValueOf } from '@rocket.chat/core-typings';
+import { RoomType } from '@rocket.chat/apps-engine/definition/rooms';
```
gabriellsh, line 2: I'm not sure this is the best place to get RoomType 🤔

apps/meteor/app/federation-v2/client/Federation.ts
```diff
@@ -9,6 +10,15 @@ const allowedActionsInFederatedRooms: ValueOf<typeof RoomMemberActions>[] = [
 	RoomMemberActions.LEAVE,
 ];
 
-export const actionAllowed = (action: ValueOf<typeof RoomMemberActions>): boolean => {
-	return allowedActionsInFederatedRooms.includes(action);
+export const actionAllowed = (room: Partial<IRoom>, action: ValueOf<typeof RoomMemberActions>): boolean => {
+	return room.t === RoomType.DIRECT_MESSAGE && action === RoomMemberActions.REMOVE_USER
```
gabriellsh, line 14: Maybe change `RoomType` for `isDirectMessagesRoom` from `core-typings`?

apps/meteor/app/federation-v2/server/infrastructure/rocket-chat/Federation.ts
```diff
@@ -14,8 +15,10 @@ export class Federation {
 		return room.federated === true;
 	}
 
-	public static federationActionAllowed(action: ValueOf<typeof RoomMemberActions>): boolean {
-		return allowedActionsInFederatedRooms.includes(action);
+	public static actionAllowed(room: IRoom, action: ValueOf<typeof RoomMemberActions>): boolean {
+		return room.t === RoomType.DIRECT_MESSAGE && action === RoomMemberActions.REMOVE_USER
```
gabriellsh, line 19: Same here.

apps/meteor/app/federation-v2/server/infrastructure/rocket-chat/Federation.ts
```diff
@@ -1,3 +1,4 @@
+import { RoomType } from '@rocket.chat/apps-engine/definition/rooms';
```
gabriellsh, line 1: Same here.

apps/meteor/app/ui-message/client/messageBox/messageBox.html
```diff
@@ -24,7 +24,7 @@
 							{{> icon block="rc-input__icon-svg" icon="emoji"}}
 						</span>
 					{{else}}
-						<div class="rc-message-box__icon"></div>
+						<div class="rc-message-box__federation_icon"></div>
```
gabriellsh, line 27: Shouldn't the class be `rc-message-box__icon rc-message-box__federation_icon`?

gabriellsh (changes requested):

apps/meteor/app/federation-v2/server/application/RoomServiceSender.ts
```diff
@@ -148,18 +148,50 @@ export class FederationRoomServiceSender {
...
+		const inviter = await this.rocketUserAdapter.getFederatedUserByInternalId((internalInviter as IUser)._id);
```
sampaiodiego, line 161: ok, with the new validation if the param is not a string you can remove the type casting:
```suggestion
		const invitee = await this.rocketUserAdapter.getFederatedUserByInternalId(internalUser._id);
		const inviter = await this.rocketUserAdapter.getFederatedUserByInternalId(internalInviter._id);
```

sampaiodiego (changes requested):

apps/meteor/app/federation-v2/server/application/RoomServiceSender.ts
MarcosSpessatto: for sure

apps/meteor/app/ui-message/client/messageBox/messageBox.html
MarcosSpessatto: I've created this new class because the old one was showing the cursor as a pointer, which was leading to weird behavior ( cursor as pointer where there is nothing (emoji picker)).

apps/meteor/app/federation-v2/server/application/RoomServiceSender.ts
MarcosSpessatto: > btw, I've found another similar type castings in this very same file (that are not part of this PR), if you could take a look on them as well, it would be good 😬

About the type castings, I'll address that in another PR.

PR #26286, opened 2022-07-16
[IMPROVE] General Showing Undefined

nishant23122000:

When the user wants to leave or hide the channel, in the warning box it shows undefined for the channel **general**.

Screenshot 2022-07-16 at 6 51 53 PM

## Steps to test or reproduce
1. Go to the general channel
2. open channel info section
3. try to hide or leave Chanel.
4. it will show undefined.


debdutdeb (changes requested):
Here's what I am thinking, every room has a full name or `fname` or `displayName` - you can't create one without it. The only exception being `general` since it is created internally that bypasses all checks. Which makes me think that a much better solution to this problem is to introduce a name to the general channel instead of adding a conditional essentially, which is needed for just a single case.

Issue #19110, opened 2020-10-02
Quotes in E2E chats are missing


rusty1281:
### Description:

There seem to be missing quotes when chatting in an encrypted channel 

### Steps to reproduce:

01. click the "Quote" icon (opens a quote above the chat window, that works)
02. write a response
03. post

### Expected behavior:

As in a normal channel, expected response with a nested quote

### Actual behavior:

Considering it is a private encrypted chat, no screen, but the fact is that the result is only response gets posted, like a regular reply

### Server Setup Information:

- Version of Rocket.Chat Server: 3.7
- Operating System: Linux
- Deployment Method: docker
- Number of Running Instances: 1 
- DB Replicaset Oplog: 
- NodeJS Version: v12.18.4
- MongoDB Version: 4.2.3

### Client Setup Information

- Desktop App or Browser Version: 2.17.9
- Operating System: macOS Catalina

ankar84:
Confirm that issue

baltazorbest:
Hello. Any news about fix?

rusty1281:
True, but the regular quote is still not presenting itself.

olafmertens:
I can confirm the bug. quoting doesn't work in our private channels.

fanecovi:
Hello,

I can confirm this too.
In the preview, when you write a message it is visible. Probably it is not posted, as normally a link to the original message is used, what is not working in this case. Maybe there need to be a change to repeat it instead of link it? Even if the data in the database then grows more then with a link. This could be the price to pay for using E2EE.

On the other hand, in other situations, like threads, it is also working fine. And I guess, that Rocket.Chat is there also working with links to messages and not with copies. Maybe it is worth, to check, if it can not made work for quotes in the same way too?

From my experience point of view, quotes are important. If you just have one answer to a previous message, a thread is to much. Threads are good for small discussions or topic separations, where a discussion room is too much. But quotes are perfect for just an answer or two. Often you mix topics in a chat, especially in a chat with more people.

So I guess it could be worth to check, if it is not possible to improve the quotes (=make it working) for E2EE rooms.

Thanks!

/fanecovi

Platt123:
It's been a while, but the issue still exists in 3.14
Afaik it was always there since e2e was introduced

Guile17:
Did anyone get it fixed? Just wondering since I'd like to start using E2E.

xcrone:
any update about this?

undallesoyeah:
[...]

Issue #16148, opened 2020-01-07
Quotes not being shown in group chats (E2E)


stukev:
### Description:

Whenever anyone in our group chats uses the quote functionality to quote someone else's message it only displays the actual answer typed by the user, not the original quote. Tested in web client as well as the desktop client for Ubuntu. I noticed on Ubuntu that I see the formatting for the quote appear in the desktop notification, which makes me think it's a render issue rather than a more complicated bug.

### Steps to reproduce:

1. Create a group chat (we use E2E)
2. Add multiple people to the conversation.
3. Use the quote button to quote what someone else said.
4. Add some text underneath the quote and send the message: https://i.imgur.com/t2lR8bT.png
5. Only what you typed will be shown in the group chat, the quote is missing: https://i.imgur.com/mt2Qi8S.png

### Expected behavior:

I expect the quote to be shown the same way it does in 1:1 conversation with a user directly. It works without any problems there.

### Actual behavior:

Only what you typed will be shown in the group chat, the quote is missing: https://i.imgur.com/mt2Qi8S.png

### Server Setup Information:

- Version of Rocket.Chat Server: 2.3.1
- Operating System: Ubuntu 19.04
- Deployment Method: tar
- Number of Running Instances: 1
- DB Replicaset Oplog: Enabled
- NodeJS Version: v8.15.1
- MongoDB Version: 3.6.14

### Client Setup Information

- Desktop App or Browser Version: Tested with latest Firefox, Opera, Chrome, bug is replicable on all versions.
- Operating System: Windows, Linux and Android

### Additional context

As noted above, on Ubuntu I can see Desktop notifications when somebody tried to send a quoted message to group chat, I can see the markup in the notification but it doesn't get rendered in the chat itself.

no-nice-username:
Bump for the thread. The issue still exists, but is closed for no reason.

dougfabris:
Hi there, thanks for the contribution! 🚀 💯 

Closing this PR because it's stale now and it was already fixed here: #26303.

---

**Questions? Help needed? Feature Requests?** 

- Join our [Open Server](https://open.rocket.chat/) in the #support channel and feel free to raise a question
- Join our [Community Forum](https://forums.rocket.chat/) and search/create a post there

Issue #24721, opened 2022-03-07
Quoting Messages and Link Previews don't work when channels are E2E enabled and encryption is on


himpierre:
### Description:

After enabling E2E quoting and link previews of messages do not work anymore. 

### Steps to reproduce:

1. Enable and activate E2E 
2. Goto E2E enabled channel
3. Quote a message, enter text, send message
4. Quoted message doesn't show. 

5. Enter Link in message window and send it
6. Only the link appears, no preview of the link

### Expected behavior:

Quoting and Link Previews work in E2E enabled channels too.

### Actual behavior:

They don't.

### Server Setup Information:

- Version of Rocket.Chat Server: 4.4.2 / 4.5.0
- Operating System: Linux
- Deployment Method: docker
- Number of Running Instances: 4
- DB Replicaset Oplog: Sure
- NodeJS Version: 14
- MongoDB Version: 4.4

### Client Setup Information

- Desktop App or Browser Version: Both
- Operating System: Linux and MacOS

lovekyleman:
Also have this issue

KamilSucharski:
It's inconvenient - having to choose between security and functionality. Please fix this.

fcharlaix-opendsi:
I can confirm the issue, is annoying to losses such functionality for security

undallesoyeah:
It would be very cool to get an update on this open security issue. :)

tobiasreinhard:
Now its almost mid of September 2022 and it's still not working. This is a core feature which was working without E2E, please fix this asap.

PR #26303, opened 2022-07-19
[IMPROVE] Quotes on E2EE Messages

yash-rajpal:

This PR provides basic quoting to E2EE messages, it doesn't support multiple chaining of quotes messages.

Screenshot 2022-07-19 at 9 42 19 PM

## Issue(s)

#19110 
#16148 
## Further comments

This implementation has a few limitations on how quoting will work in E2EE messages. Quotes are handled as attachments and generally quotes are added by the server when it receives the message with some message-link in its content. This approach is not possible for E2EE messages as server receives the encrypted message and cannot get the links. So quotes on E2EE messages have to be handled on Client side.

Here we use the API `chat.getMessage` to get the quoted message so we cannot see quoted messages of rooms of which user is not a part of.

We can also not chain multiple quote messages as server has no information on existing quote attachments on a E2EE message (only client has that info), though we can handle this on client side meteor collections, but not sure if that is a good idea.


apps/meteor/app/e2e/client/rocketchat.e2e.js
```diff
@@ -402,6 +411,64 @@ class E2E extends Emitter {
...
+
+			if (!getQuotedMessage?.success || !quotedMessage) {
+				return;
+			}
+
+			const decryptedQuoteMessage = await this.decryptMessage(quotedMessage);
+
+			message.attachments = message.attachments || [];
+
+			const quoteAttachment = {
```
gabriellsh, line 447: I think we do something similar in backend for regular messages, right? Is there any possibility we could implement a function to generate an attachment from a message and use it in both places? We could event unit test this function after...

gabriellsh (changes requested):
I'd like to see e2e-ui tests for this too.

gabriellsh (approved):

apps/meteor/app/e2e/client/rocketchat.e2e.ts
```diff
@@ -370,38 +406,86 @@ class E2E extends Emitter {
...
 		}
 
-		return {
+		const decryptedMessage: IE2EEMessage = {
 			...message,
 			msg: data.text,
 			e2e: 'done',
 		};
+
+		console.log('before quote', decryptedMessage);
```
sampaiodiego, line 415: oopsie.. can someone pls remove these? 🙈

yash-rajpal: Removed here - https://github.com/RocketChat/Rocket.Chat/pull/27092/commits/2d4f247bce23cf3e9728caab348665eb58a3bebb

Thanks!

KamilSucharski:
@yash-rajpal You linked this issue for this PR https://github.com/RocketChat/Rocket.Chat/issues/24721 , but it also requires working attachment previews in E2E to be closed. Is that included, or did you only modify quoting?

PR #26307, opened 2022-07-19
[FIX] - Incoming SMSs no longer clash with ongoing livechat conversations by the same visitor

cauefcr:

There was a data race in the defineVisitor function, causing new guests to be created even if a registered guest with that number already existed, also made sure that the open room being searched on is the correct source type, so the clash is not possible anymore.


murtaza98 (dismissed):
LGTM!

apps/meteor/app/models/server/models/LivechatRooms.js
```diff
@@ -288,13 +288,14 @@ export class LivechatRooms extends Base {
...
 
-	findOneOpenByVisitorTokenAndDepartmentId(visitorToken, departmentId, options) {
+	findOneOpenByVisitorTokenAndDepartmentIdAndSource(visitorToken, departmentId, source, options) {
 		const query = {
 			't': 'l',
 			'open': true,
 			'v.token': visitorToken,
 			departmentId,
 		};
+		if (source) query['source.type'] = source;
```
KevLehman, line 298: We don't use short ifs 🙈 (I like them, but it's on the styleguide haha)
```suggestion
		if (source) {
			query['source.type'] = source;
		}
```

KevLehman (changes requested):

apps/meteor/app/models/server/models/LivechatRooms.js
cauefcr: fixed!

KevLehman (changes requested):
Will a conversation coming from Widget or from an app collide with an already open room? 👀 if so, we can use this PR to change that behavior to 🤗

KevLehman (approved):
Better let's try that one outside here 👀

PR #26309, opened 2022-07-19
Regression: Options overlapping input in Users Autocomplete

gabriellsh:


ggazzo:
please, ask for some QA I'd tried here and I could not see any changes

PR #26320, opened 2022-07-20
[FIX] incorrect error toast messages

debdutdeb:
## Issue(s)

https://open.rocket.chat/channel/rocketchat-dev?msg=NHaMAkZDYssEEBZW4

## Further comments

`useEndpoint` throws an object that doesn't implement `toString` which is causing `String(error)` to resolve to `[object Object]` even though the object contains the `error` property. On the other hand `handleError` checks for both `message` and `error` properties along with doing other things (and now defaults to a string conversion) so there is no need to check type on runtime and cast for each instance manually.


debdutdeb:
ignore the very incorrect branch name 🙈

tassoevan (changes requested):
I screwed up big time in this one.

However, I don't like the idea of slapping `any` at the caught errors... I believe a better approach for it is introducing a new function that takes an `unknown`-type error and uses `dispatchToastMessage` under the hood to figure the "stringification" out.

ggazzo (approved):

murtaza98 (approved):

d-gubert (approved):
Looking good on the Apps-Engine side

PR #26325, opened 2022-07-20
Chore: bump fuselage packages

dougfabris:


ggazzo (changes requested):
bump to v0.31.16

packages/livechat/package.json
```diff
@@ -86,7 +86,7 @@
 	},
 	"dependencies": {
 		"@rocket.chat/sdk": "^1.0.0-alpha.42",
-		"@rocket.chat/ui-kit": "next",
+		"@rocket.chat/ui-kit": "0.31.15",
```
murtaza98, line 89: Please don't bump any Fuselage package in livechat and livechat isn't compatible with those new versions. We are facing issues like this due to these upgrades https://github.com/RocketChat/Rocket.Chat/pull/26327

dougfabris: Can you add the recommended version, please?

murtaza98 (approved):

PR #26329, opened 2022-07-21
Regression: Prevent unhandled Error with invalid emails MDM

ggazzo:


apps/meteor/ee/server/lib/deviceManagement/session.ts
```diff
@@ -92,10 +95,7 @@ export const listenSessionLogin = async (): Promise<void> => {
 					data: mailData,
 				});
 			} catch ({ message }) {
-				throw new Meteor.Error('error-email-send-failed', `Error trying to send email: ${message}`, {
-					method: 'listenSessionLogin',
-					message,
-				});
+				logger.error(`Failed to send email to ${email}: ${message}`);
```
sampaiodiego, line 98: in this case I think it is better to log the whole error obj:
```ts
} catch (err) {
    logger.error({ msg: `Failed to send email to ${email}`, err });
}
```

sampaiodiego (changes requested):

PR #25789, opened 2022-06-07
[FIX] Correct IMAP configuration for email inbox

cauefcr:
## Proposed changes (including videos or screenshots)

The primary change here has been to make the library try and reconnect after some time, up to a certain configured number of times, on a few different error classes.

## Steps to test or reproduce

Enable Omnichannel, select something as livechat routing strategy.
Administration -> Users -> click your username -> make sure your e-mail is verified.
Administration -> Email Inbox -> NEW -> fill everything up -> Save
Test the email by clicking on the arrow icon, and checking if it arrives on your email.
If it does, send an e-mail to the configured e-mail from another e-mail account.
The e-mail should arrive as a new visitor with a little envelope icon on the sidebar.
Click on the new visitor -> Reply via email -> type something and press enter
The response should arrive on the visitor's email. 

## Further comments

Email can be finicky to get working, so here a few specific tricks to get it to run:
gmail: enable 2-factor-authentication, enable an app password called 'Rocket.Chat', and use that as the password on the inbox.
outlook: really dislikes our SMTP SSL connection, SSL must be disabled to work with outlook/office365.

This list is compatible with our SMTP library https://nodemailer.com/smtp/well-known/ and provides working configurations for known hosts, comparing it with this list https://www.arclab.com/en/kb/email/list-of-smtp-and-imap-servers-mailserver-list.html makes it easier to test different email services.

P.S.: Other e-mail providers may need app passwords like gmail does.

PR #26363, opened 2022-07-26
[FIX] Onhold auto chat resume feature not working for email channel

murtaza98:
## Issue(s)

If an omnichannel chat is placed on hold & it receives a new message from the guest then the system should automatically resume this chat from on hold. However, for the email channel, the chat wasn't getting resumed automatically, which should be fixed by this PR


KevLehman (approved):

murtaza98:
Skipping QA for this item since it can be tested together with https://github.com/RocketChat/Rocket.Chat/pull/25789

PR #26372, opened 2022-07-27
[FIX] Slash commands description as undefined

carlosrodrigues94:


apps/meteor/app/utils/lib/slashCommand.ts
```diff
@@ -30,14 +31,16 @@ export const slashCommands = {
...
 		previewCallback,
 		appId,
+		description = '',
 	}: ISlashCommandAddParams<T>): void {
 		this.commands[command] = {
 			command,
 			callback,
 			params: options.params,
-			description: options.description,
+			description: options.description || description,
```
LucianoPierdona, line 40: instead of creating a new `description` field that seems like isn't being used anywhere, could you update it to be `options.description || ''` or define it as default on the initialization of the `options` field?

apps/meteor/app/utils/lib/slashCommand.ts
```diff
@@ -30,14 +31,16 @@ export const slashCommands = {
...
 			permission: options.permission,
 			clientOnly: options.clientOnly || false,
+
```
LucianoPierdona, line 43: Remove this empty line.

LucianoPierdona (changes requested):

apps/meteor/app/utils/lib/slashCommand.ts
carlosrodrigues94: I believe that this will not be possible because, depending of the context, description comes from options or not

LucianoPierdona (approved):

PR #26377, opened 2022-07-27
Chore: Migrate omni-chat forwarding to use API instead of meteor method

murtaza98:

- Use `livechat/room.forward` endpoint to forward omnichannel chats instead of using meteor method "livechat:transfer"

## Further comments

Pending tasks:

- [x] Migrate UI to use this endpoint instead of meteor method
- [x] Tests


apps/meteor/tests/data/livechat/department.ts
```diff
@@ -0,0 +1,61 @@
...
+						showOnRegistration: true,
+						showOnOfflineForm: true,
+						name: `new department ${Date.now()}`,
+						description: 'created from api',
+					}, initialAgents],
+					id: 'id',
+					msg: 'method',
+				}),
+			})
+			.end((err: any, res: any) => {
```
KevLehman, line 28: Hey, i made this file to typescript. Made a "helper" type for responses, that should help you here 😬

apps/meteor/tests/end-to-end/api/livechat/00-rooms.ts
```diff
@@ -240,4 +243,128 @@ describe('LIVECHAT - rooms', function () {
...
+				.expect((res: Response) => {
+					expect(res.body).to.have.property('success', true);
+				});
+
+			const latestRoom = await getLivechatRoomInfo(newRoom._id);
+
+			expect(latestRoom).to.have.property('departmentId');
+			expect(latestRoom.departmentId).to.be.equal(forwardToDepartment._id);
+
+			expect(latestRoom).to.have.property('lastMessage');
```
KevLehman, line 362: Pls type `getLivechatRoomInfo` to avoid using `as any` a ton 😬

KevLehman (changes requested):

apps/meteor/app/livechat/server/api/v1/room.js
```diff
@@ -199,10 +199,20 @@ API.v1.addRoute('livechat/room.survey', {
...
+
+			Meteor.runAsUser(this.userId, () => {
+				result = Meteor.call('livechat:transfer', this.bodyParams);
+			});
+
+			if (result) {
+				return API.v1.success();
+			}
+
+			return API.v1.failure();
```
KevLehman, line 215: Is the UI prepared for this change? Otherwise, this would be a breaking change

murtaza98: Actually, this endpoint wasn't getting used on the UI before - we were still using a meteor method over there. Hence I've also made the transition on UI to use this endpoint instead of the method

KevLehman (approved):

ggazzo (dismissed):

ggazzo (approved):

PR #26385, opened 2022-07-27
[FIX] Blank screen after requesting transcript

MartinSchoeler:


KevLehman (approved):

aleksandernsilva (dismissed):

ggazzo (changes requested):
is there any way to cover this case with some test?

aleksandernsilva (approved):

ggazzo (approved):

PR #26394, opened 2022-07-28
[FIX] Not allowed error in discussion room with a private parent channel

filipemarins:
## Further comments

If a user is added to a discussion that was created under a private group as its parent channel and that user does not have access to the parent private group, the user keeps getting a "not-allowed" error banner whenever they open the discussion.
Sometimes the error `too-many-request` is thrown (429).

The `ParentRoomWithEndpointData` component was calling the `rooms.info` as a fallback when the parent component didn't have the `Subscription`. This fallback was fine before when the room component wasn't await for `Subscription`. Nowadays this fallback doesn't make sense because the room already waits for `Subscription` to load, the current fallback is only being used when the user doesn't have permission to view the info in the parent room.


gabriellsh:
> the current fallback is only being used when the user doesn't have permission to view the info in the parent room.

Actually, if the parent room is a public channel, the user does have permission to fetch the info through the endpoint. We should only change the fetch to not dispatch an error message.

gabriellsh (dismissed):

gabriellsh (approved):

PR #26400, opened 2022-07-28
fix: [ENTERPRISE] Guest users can join more than maxRoomsPerGuest rooms

carlosrodrigues94:

 - Add `licenses.maxGuestUsers` and `licenses.maxRoomsPerGuest` endpoints to enable EE workspaces to get `maxGuestUsers` and `maxRoomsPerGuest` info from their licenses;
 - Properly block guest users from joining more than the `maxRoomsPerGuest` amount of rooms defined in the workspace license.

## Further comments

TC-70


apps/meteor/ee/server/startup/maxRoomsPerGuest.ts
```diff
@@ -0,0 +1,20 @@
+import { Meteor } from 'meteor/meteor';
+import { TAPi18n } from 'meteor/rocketchat:tap-i18n';
+import type { IUser } from '@rocket.chat/core-typings';
+
+import { callbacks } from '../../../lib/callbacks';
+import { getMaxRoomsPerGuest } from '../../app/license/server/license';
+import { Subscriptions } from '../../../app/models/server';
+
+callbacks.add(
```
gabriellsh, line 9: I'm not sure, but I thought we were deprecating callbacks? @matheusbsilva137 Can you confirm?

matheusbsilva137: Hey Henriques! I've talked to @pierre-lehnen-rc and we're still using callbacks in this case even though they're deprecated

LucianoPierdona (approved):

sampaiodiego (changes requested):
hey @matheusbsilva137 , can you please assure we have tests for this validation? thx

LucianoPierdona (dismissed):

hugocostadev (approved):

LucianoPierdona (dismissed):

LucianoPierdona (dismissed):

apps/meteor/ee/server/api/licenses.ts
```diff
@@ -70,6 +78,31 @@ API.v1.addRoute(
...
+	{ authRequired: true },
+	{
+		get() {
+			const maxRoomsPerGuest = getMaxRoomsPerGuest() || null;
+
+			return API.v1.success({ maxRoomsPerGuest });
+		},
+	},
+);
+
```
sampaiodiego, line 105: any special reason for this information to be publicly available (to any logged in user)? I don't see value making this information public and consider it somewhat confidential.

sampaiodiego (changes requested):

apps/meteor/ee/server/api/licenses.ts
matheusbsilva137: Not exactly :thinking: I decided to create it so that I could use it in the end-to-end tests. Do you think this should be protected by a permission or removed?

matheusbsilva137: Fixed! I've removed the endpoints and replaced them in the tests by a constant

ggazzo (dismissed):

matheusbsilva137 (approved):

sampaiodiego (dismissed):

matheusbsilva137 (approved):

PR #26402, opened 2022-07-28
Chore: create setAvatarFromService endpoint

felipe-rod123:

Created the 'users.setAvatarFromService' endpoint for the `apps/meteor/client/hooks/useUpdateAvatar.ts` file, and added Ajv validations.


apps/meteor/tests/data/interactions.js
```diff
@@ -1,2 +1,3 @@
 export const targetUser = 'rocket.cat';
 export const imgURL = './public/images/logo/1024x1024.png';
+export const avatarURL = 'apps/meteor/public/images/logo/1024x1024.png';
```
ggazzo, line 3:
```suggestion
export const avatarURL = BASE_URL +  '/images/logo/1024x1024.png';
```

apps/meteor/tests/data/interactions.js
```diff
@@ -1,2 +1,3 @@
 export const targetUser = 'rocket.cat';
```
ggazzo, line 1:
```suggestion
import { BASE_URL } from '../e2e/config/constants'

export const targetUser = 'rocket.cat';
```

ggazzo (approved):

apps/meteor/app/api/server/v1/users.ts
```diff
@@ -226,6 +227,43 @@ API.v1.addRoute(
 	},
 );
 
+API.v1.addRoute(
+	'users.setAvatarFromService',
```
sampaiodiego, line 231: there is no need to create a new endpoint.. we already have the endpoint `users.setAvatar`, if it is missing this functionality we should just improve it.

apps/meteor/server/methods/setAvatarFromService.ts
```diff
@@ -29,7 +31,8 @@ Meteor.methods({
 		let user;
 
 		if (userId && userId !== Meteor.userId()) {
-			if (!hasPermission(Meteor.userId(), 'edit-other-user-avatar')) {
+			// eslint-disable-next-line @typescript-eslint/no-non-null-assertion
```
sampaiodiego, line 34: instead of having to disable this rule, you can make your code complaint. one way to achieve that is assigning `Meteor.userId()` value to a variable and checking/using that variable, so something like:
```ts
const userId = Meteor.userId();
if (!userId) {
  throw new Error('generic error');
}

// now you can use userId here, no need to disable a TS lint rule
if (!hasPermission(userId, 'edit-other-user-avatar')) {
  // throw ...
}
```

apps/meteor/tests/end-to-end/api/01-users.js
```diff
@@ -751,6 +751,91 @@ describe('[Users]', function () {
...
+		let userCredentials;
+		before(async () => {
+			user = await createUser();
+			userCredentials = await login(user.username, password);
+		});
+		before((done) => {
+			updateSetting('Accounts_AllowUserAvatarChange', true).then(() => {
+				updatePermission('edit-other-user-avatar', ['admin', 'user']).then(done);
+			});
+		});
```
sampaiodiego, line 765: please use `async/await` (just like the lines above):
```suggestion
		before(async () => {
			await updateSetting('Accounts_AllowUserAvatarChange', true);
			await updatePermission('edit-other-user-avatar', ['admin', 'user']);
		});
```

sampaiodiego (changes requested):

PR #26419, opened 2022-07-29
Chore: Permissions check per endpoint/method

KevLehman:
Note: most files changed are conversions to use the implemented permissions param :eyes: so don't worry to see 40 files.

The idea:
- There are some routes that have multiple handlers per HTTP method. What happens in those routes is that we need to check for permissions on each one of the handlers, which is a bit of boilerplate, and increases the number of code branches. With this change, endpoints will be able to say which permissions apply to each one of the methods, as well as the "operation" that they want the permissions to be.
- For example, an endpoint could be accessed by any user with 'view-l-room' or 'view-m-room'. With the current implementation, that cannot be done. With the new changes, devs can pass a `operation: hasAll | hasAny` prop to signal how permissions should be evaluated, if inclusively or exclusively.

:)


apps/meteor/app/livechat/imports/server/rest/rooms.js
```diff
@@ -19,7 +19,7 @@ const validateDateParams = (property, date) => {
 
 API.v1.addRoute(
 	'livechat/rooms',
-	{ authRequired: true },
+	{ authRequired: true, permissionsRequired: { GET: { permissions: ['view-livechat-rooms', 'view-l-room'], operation: 'hasAny' } } },
```
sampaiodiego, line 22: you didn't remove the checks from endpoint code

apps/meteor/app/livechat/imports/server/rest/users.js
```diff
@@ -20,41 +29,42 @@ API.v1.addRoute(
 			const { text } = this.queryParams;
 
 			if (this.urlParams.type === 'agent') {
+				if (!(await hasAtLeastOnePermission(this.userId, ['transfer-livechat-guest', 'edit-omnichannel-contact']))) {
```
sampaiodiego, line 32: either `await` for the `async` version or don't use `await`
```suggestion
				if (!(await hasAtLeastOnePermissionAsync(this.userId, ['transfer-livechat-guest', 'edit-omnichannel-contact']))) {
```

apps/meteor/app/api/server/api.js
```diff
@@ -52,6 +52,25 @@ const getRequestIP = (req) => {
 	return forwardedFor[forwardedFor.length - httpForwardedCount];
 };
 
+const checkPermisisonsForInvocation = (userId, permissionsPayload, requestMethod) => {
```
sampaiodiego, line 55: (picky alert)

we might need a better definition on these, but I think we should use more named functions.. 

maybe the definition could be: use arrow functions only when creating local/scoped functions, like a function inside another function. wdyt?

apps/meteor/app/api/server/api.js
```diff
@@ -311,21 +330,47 @@ export class APIClass extends Restivus {
 		return routeActions.map((action) => this.getFullRouteName(route, action, apiVersion));
 	}
 
+	checkPermissions(options) {
```
sampaiodiego, line 333: (pick alert 2)

this method could be a function in file's scope right?

I'd also recommend not changing the `options` param, but instead having  the `permissionRequired` as the return value.

having this function in TS and having a unit test would also be good

sampaiodiego (changes requested):
picky comments can be addressed in a future engineering day =)

apps/meteor/app/api/server/api.js
sampaiodiego: yep, I think we're missing a standard.. I'm worried about two things specifically:

- hoist variables and how that affects gc
- better stack traces

apps/meteor/app/livechat/imports/server/rest/users.js
```diff
@@ -1,17 +1,26 @@
...
 	'livechat/users/:type',
-	{ authRequired: true },
 	{
-		get() {
+		authRequired: true,
+		permissionsRequired: {
+			GET: {
+				permissions: ['manage-livechat-agents'],
+				operation: 'hasAll',
+			},
```
murtaza98, line 18: I believe there's a small issue here which changes the permission checks for this endpoint. If we take an example of agents (i.e. `:type=agent`), right now any user with **one of these 3 permission** is allowed to query this endpoint

- 'manage-livechat-agents'
- 'transfer-livechat-guest'
- 'edit-omnichannel-contact'

But with this new logic, the user will always need to have 'manage-livechat-agents' role & users with who don't have this role but do have the other 2 roles won't be able to access this endpoint

murtaza98, line 18: I believe there's a small issue here which changes the permission checks for this endpoint. If we take an example of agents (i.e. `:type=agent`), right now any user with **one of these 3 permission** is allowed to query this endpoint

- 'manage-livechat-agents'
- 'transfer-livechat-guest'
- 'edit-omnichannel-contact'

But with this new logic, the user will always need to have 'manage-livechat-agents' role & users who don't have this role but do have the other 2 roles won't be able to access this endpoint

murtaza98: FYI: fixed this here https://github.com/RocketChat/Rocket.Chat/pull/26552

murtaza98: FYI: fixed this here https://github.com/RocketChat/Rocket.Chat/pull/26552

murtaza98:
Hey @KevLehman This looks really good :tada: I've left just a few suggestions over here https://github.com/RocketChat/Rocket.Chat/pull/26552 - they are mostly removing some permission checks which aren't required now & addressing the issue I mentioned above

murtaza98 (dismissed):
LGTM :tada:

sampaiodiego (approved):

murtaza98 (approved):

KevLehman:
@sampaiodiego @murtaza98 , you'll have to re-approve this one since there's a conflict :cry:

KevLehman:
btw, can we skip QA on this one? :eyes:

sampaiodiego (approved):

sampaiodiego:
since the endpoints changed are all omnichannel, I'll let you guys decide if you want to skip QA.

murtaza98 (approved):

apps/meteor/app/api/server/api.js
```diff
@@ -465,6 +462,8 @@ export class APIClass extends Restivus {
 
 						result = DDP._CurrentInvocation.withValue(invocation, () => Promise.await(originalAction.apply(this))) || API.v1.success();
 
+						console.log('------------------------------------------------------');
+						console.log(result);
```
murtaza98, line 466:
```suggestion
```

murtaza98 (approved):

sampaiodiego (approved):

murtaza98 (approved):

PR #26420, opened 2022-07-29
Chore: auto approve lingohub

ggazzo:


sampaiodiego:
we discussed about this and we think it is better to set a codeowner for i18n instead of merging automatically.

sampaiodiego:
@dougfabris help me explain this to @ggazzo pls

ggazzo:
image

ggazzo:
I thought we had this tool called lingohub to do the validations

Issue #26361, opened 2022-07-26
[bug] Chats holds to load history for some time until any action since 5.0 version


ankar84:
### Description:

Since 5.0 I consider that some chats are loading history for a long time (some even infinite). 
But on any action, history is loaded immediately.

### Steps to reproduce:

1. Search user you don't have recent talk (for easy reproduce, that could be any chats in fact)
2. Click to start conversation with that user
3. Chat history is loading even if there is no history and it as a brand new DM

### Expected behavior:

Chats open fast like it was on a pre 5.0 versions

### Actual behavior:

![image](https://user-images.githubusercontent.com/4023037/180988323-edebc8a6-a08a-4ac8-a93a-df0d8d328498.png)
From opens server
![load-history01](https://user-images.githubusercontent.com/4023037/180991158-a3c13434-9769-4dc9-ab32-86c870939a75.gif)

### Server Setup Information:

- Version of Rocket.Chat Server: 5.0.0
- Operating System: OL8
- Deployment Method: docker
- Number of Running Instances: 3
- DB Replicaset Oplog: Enabled
- NodeJS Version: v14.19.3
- MongoDB Version: 4.4.12 / wiredTiger (oplog Enabled)

### Client Setup Information

- Desktop App or Browser Version: Chrome 103
- Operating System: Windows 10

### Additional context

I see it on open.rocket.chat as well

### Relevant logs:

No logs in moment of loading...

stachi:
Can confirm this with 5.0.0-rc12 (which is identical to final) with manual deployment. Tested with latest Firefox ESR. It happens very randomly.

JeroenAdam:
Yes, a large negative impact on performance since upgrade from 4.8.2
Rocketchat v5.0.0 - Mongo 5.0.9
I tried dropping/creating the index navigation.token on messages, it is not a solution

engelgabriel:
Are there any errors on the browser console?

JeroenAdam:
No errors in console.
Maybe this screenshot can help.
![image](https://user-images.githubusercontent.com/16437621/181545924-fa297c0b-4778-4753-a624-2fd9f2364990.png)

milton-rucks:
Task already included in the current sprint as a high priority - the team is working on that

ankar84:
could be fixed in https://github.com/RocketChat/Rocket.Chat/pull/26425

chotaire:
[...]

PR #26425, opened 2022-07-30
[FIX] Chats holds to load history for some time

filipemarins:
## Further comments

Fix the conditional of wait until wrapper exists function, the `querySelector` returns an `Element` when the selector is found and `null` when isn't and doesn't have the property `length`. So the element was not being returned when it was already in the dom.


ggazzo (changes requested):
I think we have some issues opened related to that issue, could you mention them with 'closes'?


btw, could you create some test reproducing the steps to get stuck in this situation?

thanks

apps/meteor/app/ui-utils/client/lib/RoomHistoryManager.js
```diff
@@ -42,7 +42,7 @@ export const normalizeThreadMessage = ({ ...message }) => {
 
 export const waitUntilWrapperExists = async (selector = '.messages-box .wrapper') => {
 	const element = document.querySelector(selector);
-	if (element?.length) {
+	if (element) {
```
gabriellsh, line 45: I believe this will reintroduce the previous bug of not scrolling to the correct message when following a URL.

gabriellsh (changes requested):

apps/meteor/app/ui-utils/client/lib/RoomHistoryManager.js
gabriellsh: maybe we need to test `typeof element`? I remember when first resolving the issue I mentioned, `document.querySelector` was bringing an "empty" element that was not falsy.

filipemarins:
> I think we have some issues opened related to that issue, could you mention them with 'closes'?
> 
> btw, could you create some test reproducing the steps to get stuck in this situation?
> 
> thanks

Already have a [mention](https://github.com/RocketChat/Rocket.Chat/issues/26361#issuecomment-1200611347) for it.

I added a unit test to this method, I had to use the `jsdom` library inside the test to mock the `MutationObserver` class. The `jsdom-global` doesn't export the `MutationObserver` class.

gabriellsh (approved):
LGTM! I tested the old behaviour related to the message URLs and it doesn't seem to be happening. We should keep an eye open for it though.

PR #26435, opened 2022-08-01
Chore: Upgrade Fuselage packages to next dist-tag

dougfabris:


ggazzo (approved):

murtaza98:
Hi @dougfabris @ggazzo A friendly reminder about not updating the Livechat UiKit package when bumping all fuselage packages  :pray: Our Livechat isn't compatible with these latest UiKit versions and this causes problems like Quick reply buttons not working on Livechat. We do have a tech debt on our squad for this but until it is fixed, we need to make sure we're always using that old UiKit version on Livechat

dougfabris:
> Hi @dougfabris @ggazzo A friendly reminder about not updating the Livechat UiKit package when bumping all fuselage packages 🙏 Our Livechat isn't compatible with these latest UiKit versions and this causes problems like Quick reply buttons not working on Livechat. We do have a tech debt on our squad for this but until it is fixed, we need to make sure we're always using that old UiKit version on Livechat

Oh, man! Indeed, I forgot about it

murtaza98:
Thanks for understanding @dougfabris :heart: Hopefully we'd be able to get rid of this tech debt very soon so we don't have to be this careful while bumping the packages :smile:

PR #26465, opened 2022-08-03
Chore: Remove settings Fibers usage

ggazzo:


apps/meteor/app/settings/server/raw.js
```diff
@@ -1,11 +1,11 @@
-import Settings from '../../models/server/models/Settings';
+import { Settings } from '@rocket.chat/models';
 
 const cache = new Map();
 
 export const setValue = (_id, value) => cache.set(_id, value);
 
 const setFromDB = async (_id) => {
-	const setting = Settings.findOneById(_id, { fields: { value: 1 } });
+	const setting = await Settings.findOneById(_id, { fields: { value: 1 } });
```
sampaiodiego, line 8:
```suggestion
	const setting = await Settings.findOneById(_id, { projection: { value: 1 } });
```

apps/meteor/app/ui-master/server/index.js
```diff
@@ -124,16 +125,11 @@ Meteor.startup(() => {
...
-		const colors = Settings.find({ _id: /theme-color-rc/i }, { fields: { value: 1, editor: 1 } })
-			.fetch()
-			.filter((color) => color && color.value);
-
-		if (!colors) {
-			return;
-		}
+		const colors = await Settings.find({ _id: /theme-color-rc/i }, { projection: { value: 1, editor: 1 } }).toArray();
 		const css = colors
+			.filter((color) => color && color.value)
```
sampaiodiego, line 132: is it better filter colors without values or let it go to CSS with an empty value?

because if `color` has no `value`, with this `filter` there will be no `--color-name: ` into the final CSS

apps/meteor/app/settings/server/SettingsRegistry.ts
```diff
@@ -83,21 +83,21 @@ const compareSettings = compareSettingsIgnoringKeys([
 ]);
 
 export class SettingsRegistry {
-	private model: typeof SettingsModel;
+	private model: typeof Settings;
```
sampaiodiego, line 86:
```suggestion
	private model: ISettingsModel;
```

apps/meteor/app/livechat/server/methods/saveIntegration.ts
```diff
@@ -1,12 +1,13 @@
+import { Settings } from '@rocket.chat/models';
```
sampaiodiego, line 1: should all updates in this file have an `await`?

sampaiodiego, line 1: should all updates in this file have an `await`?

apps/meteor/app/cloud/server/functions/unregisterWorkspace.ts
```diff
@@ -0,0 +1,20 @@
...
+
+import { retrieveRegistrationStatus } from './retrieveRegistrationStatus';
+
+export async function unregisterWorkspace() {
+	const { workspaceRegistered } = retrieveRegistrationStatus();
+	if (!workspaceRegistered) {
+		return true;
+	}
+
+	Settings.updateValueById('Cloud_Workspace_Id', undefined);
```
sampaiodiego, line 11: why did you change to `undefined`?

sampaiodiego: maybe `await Promise.all( ... )`?

sampaiodiego (changes requested):

apps/meteor/app/ui-master/server/index.js
ggazzo: valid question, but since this is the original filter, I will kindly ignore, since its not the PR scope ;)

apps/meteor/app/cloud/server/functions/unregisterWorkspace.ts
```diff
@@ -0,0 +1,22 @@
...
+	}
+
+	await Promise.all([
+		Settings.updateValueById('Cloud_Workspace_Id', undefined),
+		Settings.updateValueById('Cloud_Workspace_Name', undefined),
+		Settings.updateValueById('Cloud_Workspace_Client_Id', undefined),
+		Settings.updateValueById('Cloud_Workspace_Client_Secret', undefined),
+		Settings.updateValueById('Cloud_Workspace_Client_Secret_Expires_At', undefined),
+		Settings.updateValueById('Cloud_Workspace_PublicKey', undefined),
+		Settings.updateValueById('Cloud_Workspace_Registration_Client_Uri', undefined),
```
sampaiodiego, line 18:
```suggestion
		Settings.updateValueById('Cloud_Workspace_Id', null),
		Settings.updateValueById('Cloud_Workspace_Name', null),
		Settings.updateValueById('Cloud_Workspace_Client_Id', null),
		Settings.updateValueById('Cloud_Workspace_Client_Secret', null),
		Settings.updateValueById('Cloud_Workspace_Client_Secret_Expires_At', null),
		Settings.updateValueById('Cloud_Workspace_PublicKey', null),
		Settings.updateValueById('Cloud_Workspace_Registration_Client_Uri', null),
```

sampaiodiego (changes requested):

apps/meteor/app/cloud/server/functions/unregisterWorkspace.ts
ggazzo: this is the ONLY PLACE where we use null instead of undefined, type definitions predict undefined, what is the gain/benefit of using null somewhere?

sampaiodiego: I actually found just a SINGLE PLACE using `undefined` 🤔 the is in TS, so it can't anything else.. but clearly there are more usage of `null`..

I don't understand why you trust more on the type definition than on this function.. the point is, `undefined` is ignored and the record is not updated.. so the correct is `null`, the type definition must be changed to reflect this.

sampaiodiego (approved):

PR #26467, opened 2022-08-04
Chore: Exit process on `unhandledRejection` on CI

sampaiodiego:

### DIFF
--- a/apps/meteor/app/voip/server/startup.ts
+++ b/apps/meteor/app/voip/server/startup.ts
@@ -1,9 +1,5 @@
-import debounce from 'lodash.debounce';
-
 import { settings } from '../../settings/server';
 import { Voip } from '../../../server/sdk';
-
-const debouncedRefresh = debounce(Voip.refresh, 1000);
 
 settings.watch('VoIP_Enabled', (value: boolean) => {
 	return value ? Voip.init() : Voip.stop();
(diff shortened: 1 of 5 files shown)


apps/meteor/app/voip/server/startup.ts
```diff
@@ -14,6 +14,8 @@ settings.changeMultiple(
 	(_values) => {
 		// Here, if 4 settings are changed at once, we're getting 4 diff callbacks. The good part is that all callbacks are fired almost instantly
 		// So to avoid stopping/starting voip too often, we debounce the call and restart 1 second after the last setting has reached us.
-		return debouncedRefresh();
+		if (settings.get('VoIP_Enabled')) {
+			return debouncedRefresh();
+		}
```
ggazzo, line 19: changeMultiple is already debounced, so there is no need for the line 6

ggazzo: image

ggazzo: image

sampaiodiego: the debounce was higher than the settings' default (1000ms while settings' is 500ms), but I agree 500ms should be enough, so removed.

this comment from your image was not being used as documentation for editor's intellisense, so I changed that as well

sampaiodiego:
btw, with the fixes introduced by this PR I think we should make `test-ee` CI step mandatory to merge.

ggazzo (approved):

apps/meteor/app/voip/server/startup.ts
ggazzo: image
its possible to change the default if 1s is the exactly value

PR #26481, opened 2022-08-04
[FIX] Save edited tags for omnichannel departments

tiagoevanp:

A simple workaround for the main problem, when editing a department must be possible to save addition or remotion of tags.


apps/meteor/client/views/omnichannel/departments/EditDepartment.js
```diff
@@ -59,6 +59,7 @@ function EditDepartment({ data, id, title, reload, allowedToForwardData }) {
 	const { department } = data || { department: {} };
 
 	const [[tags, tagsText], setTagsState] = useState(() => [department?.chatClosingTags ?? [], '']);
+	const [newTagsState, setNewTagsState] = useState(false);
```
aleksandernsilva, line 62: Wouldn't it be better if you named this const as `hasTagsChanged` or `hasNewTags`? I think having state in the name doesn't add much meaning to the variable.

aleksandernsilva (changes requested):

apps/meteor/client/views/omnichannel/departments/EditDepartment.js
tiagoevanp: I don't think this is really necessary at this point... Probably this will be removed on rewrite, at the moment is just a workaround 😬

ggazzo (changes requested):
I think comparing the old x new value would be the best solution 

btw dont forget e2e tests!

KevLehman (dismissed):
Function-wise seems to be working :eyes:

aleksandernsilva (approved):

apps/meteor/client/views/omnichannel/departments/EditDepartment.js
```diff
@@ -58,7 +58,9 @@ function EditDepartment({ data, id, title, reload, allowedToForwardData }) {
 
 	const { department } = data || { department: {} };
 
-	const [[tags, tagsText], setTagsState] = useState(() => [department?.chatClosingTags ?? [], '']);
+	const initialTags = useMemo(() => department?.chatClosingTags ?? [], [department?.chatClosingTags]);
```
ggazzo, line 61:
```suggestion
	const [initialTags] = useState(() => department?.chatClosingTags ?? [], [department?.chatClosingTags]);
```

aleksandernsilva (dismissed):

ggazzo (approved):

PR #26495, opened 2022-08-05
[FIX] Allow normal user to open apps contextual bar

tapiarafael:

Fix the bug where normal users cannot open an app contextual bar.
The request made by the contextual bar to get the app information, which was for admin only, was removed since the response was not being used.


apps/meteor/client/views/room/hooks/useAppsContextualBar.ts
```diff
@@ -24,17 +23,12 @@ export const useAppsContextualBar = (): AppsContextualBarData | undefined => {
...
-			setAppInfo(app);
-		}
-
 		if (viewId) {
 			setPayload(getUserInteractionPayloadByViewId(viewId) as IUIKitContextualBarInteraction);
 		}
 
 		if (payload?.appId) {
-			getAppData(payload.appId);
+			setAppInfo({ id: payload.appId } as any);
```
dougfabris, line 31: we should avoid any!! 
as you're not going to hold all app data you should fix the `appInfo` type

dougfabris (changes requested):

ggazzo (approved):

PR #26416, opened 2022-07-29
[NEW] Surface featured apps endpoint 

rique223:
## Proposed changes (including videos or screenshots)

Created the /featured endpoints on the rest.js file. Also created the necessary typings to use together with it.

PR #26514, opened 2022-08-08
[NEW] Sections layout and featured apps for marketplace

rique223:
### [MKP??+ - Featured Apps](https://app.clickup.com/t/2t1w9x3)
Implemented a new layout for marketplace's apps list page. Now the page has an All apps section and a "featured apps" section in which pre-selected and pre-categorized apps are dynamically shown on the top area of the page to make them stand out.
Demo gif:
featured_apps

ClickUp Task link: https://app.clickup.com/t/2t1w9x3


apps/meteor/client/views/admin/apps/AppsList.tsx
```diff
@@ -1,202 +1,26 @@
...
-					</States>
-				</Box>
-			)}
-		</>
-	);
-};
+} & ComponentProps<typeof Box>;
+
+const AppsList = ({ apps, title, isMarketplace, ...props }: AppsListProps): ReactElement => (
+	<Fragment key={props.key}>
```
ggazzo, line 14: I'm quite sure prop.key will throw you a console.warning

apps/meteor/client/views/admin/apps/AppsPage.tsx
```diff
@@ -58,16 +58,29 @@ const AppsPage = ({ isMarketplace }: AppsPageProps): ReactElement => {
 					)}
 				</ButtonGroup>
 			</Page.Header>
-			<Tabs bg={colors.n100}>
-				<Tabs.Item onClick={(): void => marketplaceRoute.push({ context: '' })} selected={isMarketplace}>
+			<Tabs bg={colors.n100} borderBlockEnd='4px solid #CBCED1' marginInline='x24'>
```
ggazzo, line 61: dont just, dont  pass styles to components, if you want, create this divider inside the fuselage lib and than use as some kind of variation

apps/meteor/client/views/admin/apps/AppsPageContent.tsx
```diff
@@ -0,0 +1,213 @@
...
+						<StatesTitle>{t('No_installed_app_matches')}</StatesTitle>
+						{appsResult?.value?.shouldShowSearchText ? (
+							<StatesSubtitle>
+								<span>
+									{t('No_app_matches_for')} <strong>"{text}"</strong>
+								</span>
+							</StatesSubtitle>
+						) : (
+							''
+						)}
```
ggazzo, line 170: 🤔 is that really necessary?

ggazzo (dismissed):
its not a good practice put a bunch of <Box ....props> around you application code.
its recommended to create a 'custom component' like "FeaturedAppSectionItem" and them there you are allowed to use some styling

apps/meteor/client/views/admin/apps/AppsList.tsx
rique223: Solved

apps/meteor/client/views/admin/apps/AppsPageContent.tsx
rique223: Solved

apps/meteor/client/views/admin/apps/AppsPage.tsx
rique223: Will open a task to solve this since waiting for fuselage would unnecessarily stall this PR.

apps/meteor/app/apps/client/orchestrator.ts
```diff
@@ -107,15 +107,10 @@ class AppClientOrchestrator {
...
+			const { latest, ...rest } = app;
 			return {
 				...latest,
-				price,
-				pricingPlans,
-				purchaseType,
-				isEnterpriseOnly,
-				modifiedAt,
-				bundledIn,
+				...rest,
```
d-gubert, line 113: IIRC `latest` has some properties with the same name as the ones in the top-level `app` object, so spreading the `rest` here will likely override those if done this way.

I also like extracting exactly the info we need from the original `app` object so that we don't "overshare" unnecessary data, but that is not objectively better (unlike the above), just my two cents :P

apps/meteor/app/apps/server/communication/rest.js
```diff
@@ -362,7 +362,7 @@ export class AppsRestApi {
 
 					let result;
 					try {
-						result = HTTP.get(`${baseUrl}/v1/apps/featured`, {
+						result = HTTP.get(`${baseUrl}/v1/featured-apps`, {
```
d-gubert, line 365: This is the URL in the Marketplace API, not the Rocket.Chat server. Please double check if the Marketplace API has introduced this new endpoint, as my test with cURL (`curl https://marketplace.rocket.chat/v1/featured-apps`) hints it doesn't exist

d-gubert (changes requested):

apps/meteor/app/apps/server/communication/rest.js
rique223: This endpoint is still in the marketplace staging, it will be released shortly.

rique223: This is the staging URL if you want to test it: https://marketplace.staging.cloud.rocket.chat/v1/featured-apps

apps/meteor/app/apps/client/orchestrator.ts
rique223: Got it, will solve this shortly. I think we should take the refactoring of this app object to the engineering day. Right now it is a simple spread but it can become a big problem in the future.

apps/meteor/client/views/admin/apps/BundleChips.tsx
```diff
@@ -22,7 +22,7 @@ const BundleChips = ({ bundledIn, isIconOnly }: BundleChipsProps): ReactElement
 	return (
 		<>
 			{bundledIn.map((bundle) => (
-				<>
+				<Fragment key={bundle.bundleId}>
```
juliajforesti, line 25: Comment for line 41:
try using our typography tokens from [fuselage](http://localhost:6006/?path=/docs/layout-box-props--font-sizes) to keep the standards.
In this case would be **c2**

apps/meteor/app/apps/client/orchestrator.ts
rique223: Solved

apps/meteor/client/views/admin/apps/BundleChips.tsx
rique223: Solved

apps/meteor/client/views/admin/apps/AppsPage.tsx
```diff
@@ -58,16 +58,29 @@ const AppsPage = ({ isMarketplace }: AppsPageProps): ReactElement => {
 					)}
 				</ButtonGroup>
 			</Page.Header>
-			<Tabs bg={colors.n100}>
```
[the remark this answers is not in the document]
rique223: Solved

apps/meteor/client/views/admin/apps/AppsPageContent.tsx
```diff
@@ -0,0 +1,209 @@
...
+			<AllAppsSection appsResult={appsResult} isMarketplace={isMarketplace} />
+
+			{appsResult.phase === AsyncStatePhase.RESOLVED && Boolean(appsResult.value.count) && (
+				<Pagination
+					current={current}
+					itemsPerPage={itemsPerPage}
+					count={appsResult.value.total}
+					onSetItemsPerPage={onSetItemsPerPage}
+					onSetCurrent={onSetCurrent}
+					borderBlockStart={`2px solid ${colors.n300}`}
```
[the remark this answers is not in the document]
rique223: Solved

apps/meteor/client/views/admin/apps/AppsPageContent.tsx
```diff
@@ -0,0 +1,209 @@
...
+								<StatesSuggestionListItem>{t('Search_for_a_more_specific_term')}</StatesSuggestionListItem>
+								<StatesSuggestionListItem>{t('Check_if_the_spelling_is_correct')}</StatesSuggestionListItem>
+							</StatesSuggestionList>
+						</StatesSuggestion>
+					</States>
+				</Box>
+			)}
+
+			{noInstalledAppMatches && (
+				<Box mbs='x20'>
```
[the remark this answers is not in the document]
rique223: Solved

apps/meteor/client/views/admin/apps/AppsList.tsx
```diff
@@ -1,202 +1,26 @@
...
-				</Box>
-			)}
-		</>
-	);
-};
+} & ComponentProps<typeof Box>;
+
+const AppsList = ({ apps, title, isMarketplace, ...props }: AppsListProps): ReactElement => (
+	<>
+		<Box is='h3' fontSize='h3' fontWeight={700} lineHeight='x28' color='default' mbe='x20'>
```
ggazzo, line 15: instead of `fontSize='h3' fontWeight={700} lineHeight='x28'` I'm sure that exists some fontScale to be used

rique223: Solved

packages/core-typings/src/AppOverview.ts
```diff
@@ -0,0 +1,115 @@
...
+};
+
+export type bundledIn = {
+	bundleId: string;
+	bundleName: string;
+	apps: App[];
+	addonTierId: string;
+};
+
+export type Latest = {
```
matheuslc, line 35: @rique223 as I understood, the "latest" version could be different from workspaces. It depends on which engine version they are using. So, what do you think about renaming this object to CurrentVersion? This way we do not assume we are always using the Latest version of an app.

rique223: Solved

apps/meteor/packages/rocketchat-i18n/i18n/en.i18n.json
```diff
@@ -3082,6 +3082,8 @@
   "Marketplace_app_last_updated": "Last updated __lastUpdated__",
   "Marketplace_view_marketplace": "View Marketplace",
   "Marketplace_error": "Cannot connect to internet or your workspace may be an offline install.",
+  "marketplace_featured_section_featured": "Featured apps",
+  "marketplace_featured_section_social": "New arrivals",
```
graywolf336, line 3086:
```suggestion
  "marketplace_featured_section_community_featured": "Featured Community Apps",
  "marketplace_featured_section_community_supported": "Community Supported Apps",
  "marketplace_featured_section_enterprise": "Featured Enterprise Apps",
  "marketplace_featured_section_featured": "Featured Apps",
  "marketplace_featured_section_most_popular": "Most Popular Apps",
  "marketplace_featured_section_new_arrivals": "New Arrivals",
  "marketplace_featured_section_popular_this_month": "Apps Popular this Month",
  "marketplace_featured_section_recommended": "Recommended Apps",
  "marketplace_featured_section_social": "Social Apps",
  "marketplace_featured_section_trending": "Trending Apps",
  "marketplace_featured_section_omnichannel": "Omnichannel Apps",
  "marketplace_featured_section_video_conferencing": "Video Conferencing Apps",
```

graywolf336 (changes requested):
Let's add these sections :)

graywolf336 (dismissed):
Looks good to me and I've seen it running, from the Marketplace perspective it looks good and the APIs area in production to be used now.

d-gubert (dismissed):

apps/meteor/client/views/admin/apps/components/RadioDropDown/RadioDropDown.tsx
```diff
@@ -9,7 +9,7 @@ import DropDownListWrapper from '../DropDownListWrapper';
 import RadioButtonList from '../RadioButtonList';
 import RadioDropDownAnchor from './RadioDownAnchor';
 
-const RadioDropDown: FC<RadioDropDownProps & Partial<ComponentProps<typeof Select>>> = ({ group, onSelected, ...props }) => {
+const RadioDropDown = ({ group, onSelected, ...props }: RadioDropDownProps & Partial<ComponentProps<typeof Select>>): ReactElement => {
```
juliajforesti, line 12: I would suggest not to type with all Select props, but to define specifically the ones that are being used. Or is there a reason for forwarding all "rest" props? It increases the chance of unwanted/unforeseen props being passed.

apps/meteor/client/views/admin/apps/FeaturedAppsSections.tsx
```diff
@@ -0,0 +1,47 @@
...
+		!isFiltered;
+
+	if (shouldShowFeaturedSections && isMarketplace) {
+		return (
+			<>
+				{featuredApps.value.sections.map((section) => (
+					<AppsList
+						key={section.slug}
+						apps={normalizeFeaturedApps(section.apps, appsResult.value.items)}
+						title={t(section.i18nLabel as TranslationKey)}
```
juliajforesti, line 35: useTranslation has the method `has` that checks if this key exists.
```suggestion
title={t.has(section.i18nLabel) ? t(section.i18nLabel) : section.i18nLabel}
```

apps/meteor/client/views/admin/apps/AppsList.tsx
```diff
@@ -1,202 +1,26 @@
...
-								{t('Reload_page')}
-							</StatesAction>
-						</StatesActions>
-					</States>
-				</Box>
-			)}
-		</>
-	);
-};
+} & ComponentProps<typeof Box>;
```
juliajforesti, line 11: `ComponentProps<typeof Box>` is a whole lot of props... do you really want to add all that?

apps/meteor/client/views/admin/apps/FeaturedAppsSections.tsx
rique223: Solved

apps/meteor/client/views/admin/apps/AppsList.tsx
rique223: Solved

apps/meteor/client/views/admin/apps/components/RadioDropDown/RadioDropDown.tsx
rique223: The idea was for RadioDropDown to be a direct abstraction of Select that's why I used all props of Select so that both components can be used interchangeably.

apps/meteor/client/views/admin/apps/FeaturedAppsSections.tsx
```diff
@@ -0,0 +1,47 @@
...
+	if (shouldShowFeaturedSections && isMarketplace) {
+		return (
+			<>
+				{featuredApps.value.sections.map((section) => (
+					<AppsList
+						key={section.slug}
+						apps={normalizeFeaturedApps(section.apps, appsResult.value.items)}
+						title={t.has(section.i18nLabel) ? t(section.i18nLabel) : section.i18nLabel}
+						isMarketplace={isMarketplace}
+						mbe='x36'
```
juliajforesti, line 37: is this prop necessary? 
would be better if you could put the margin on the parent component, not pass it as a prop

rique223: Solved

d-gubert (dismissed):
All good on Apps side :)

packages/rest-typings/src/apps/index.ts
```diff
@@ -96,9 +96,9 @@ export type AppsEndpoints = {
 		};
 	};
 
-	'/apps/featured': {
+	'/apps/featured-apps': {
```
pierre-lehnen-rc, line 99: Isn't this a breaking change?

graywolf336: No sir. That route was added in https://github.com/RocketChat/Rocket.Chat/pull/26416 and was added specifically for this new feature. The Marketplace API doesn't have the old route after we tested in staging and realized it wouldn't work.

ggazzo (dismissed):

ggazzo (approved):

graywolf336 (approved):

d-gubert (approved):

PR #26534, opened 2022-08-10
Chore: Fix UiKit dependency issue for Livechat

murtaza98:


dougfabris (approved):
I'm sorry about it mate, I'll work to always remember it!

PR #26539, opened 2022-08-10
[FIX] New messages in threads are only marked as read after a message is sent in the main room

matheusbsilva137:

 - Update rooms' `lm` fields when messages in threads are sent.

## Steps to test or reproduce

1. Go to **Administration > Message** and enable "Show Read Receipts" and "Detailed Read Receipts";
2. Login with two users and access the same room with both of them;
3. As "User 1", send a message in an existing thread;
4. Access this room as "User 2". **_Expected behavior:_** the message sent in step 3 should be marked as read by "User 2" (even though the user hasn't opened the thread yet);
5. Open the thread as "User 2" and send a new message in the thread. **_Expected behavior:_** the new message should be instantly mark as read by both users.

## Further comments

By updating rooms' `lm` field when messages in threads are sent, the condition in [this line](https://github.com/RocketChat/Rocket.Chat/blob/9d818a7af44ecac02982085634cc2e0b2bdc095e/apps/meteor/imports/message-read-receipt/server/lib/ReadReceipt.js#L46) may not be met, which allows new read receipts to be created (since the new thread messages will be found in `findUnreadMessagesByRoomAndDate`).


sampaiodiego (changes requested):
@matheusbsilva137 I think the issue as actually a different one. The steps we should do to reproduce are:

- User1 sends a message in an old thread
- User2 reads the channel only (doesn't open the thread)
- Message is marked as read **inside the thread**

now that the feature to show messages from a thread in the channel is gone, it might be easier to fix this. a possible fix would be changing `findUnreadMessagesByRoomAndDate` to ignore messages from a thread. and create a new method to mark thread messages as read (storing the read receipts).

it would be very important to also had tests for this feature, since it is not enabled by default, we don't usually test how other changes affects it.

apps/meteor/server/methods/readThreads.js
```diff
@@ -28,6 +29,7 @@ Meteor.methods({
 			throw new Meteor.Error('error-not-allowed', 'Not allowed', { method: 'getThreadMessages' });
 		}
 
-		return readThread({ userId: user._id, rid: thread.rid, tmid });
+		readThread({ userId: user._id, rid: thread.rid, tmid });
+		Promise.await(markRoomAsRead(thread.rid, user._id));
```
sampaiodiego, line 33: marking the room as read might create undesired side effects, like updating the sidebar item..

I'd recommend to call `ReadReceipt.markMessagesAsRead` directly, but this will also generate weird behaviors, like adding read receipts to threads that are older but not read..

sampaiodiego (changes requested):

gabriellsh:
Closed in favor of https://github.com/RocketChat/Rocket.Chat/pull/26611

PR #26556, opened 2022-08-12
Chore: create max chats by agent test

souzaramon:


weslley543 (approved):
Will message sending now be done in tests? It's fine with me to be like that. Just make sure that all tests will also be done

PR #26558, opened 2022-08-12
[FIX] SMS service check

KevLehman:


cauefcr (dismissed):
LGTM, i don't see how this could be a bad change.

murtaza98 (approved):

cauefcr (approved):
I liked the logging added, it'll make much easier for admins to have an idea of what's going on on their SMS integration.

LGTM.

PR #26571, opened 2022-08-15
Regression: Update custom homepage content behavior

gabriellsh:
Custom homepage content should now be always visible below the default cards (in case there is any content there).
For enterprise, a new setting was created `Show custom content only`, which when enabled should hide the default cards.


hugocostadev (dismissed):

apps/meteor/ee/app/settings/server/settings.ts
```diff
@@ -16,7 +16,7 @@ export function changeSettingValue(record: ISetting): SettingValue {
 	}
 
 	if (!record.modules?.length) {
-		return record.invalidValue;
+		return record.value;
```
sampaiodiego, line 19: can you please explain this change? `invalidValue` should be the correct value to use here.

apps/meteor/app/lib/server/startup/settings.ts
```diff
@@ -1555,8 +1555,11 @@ settingsRegistry.addGroup('Layout', function () {
 			type: 'boolean',
 			public: true,
 		});
-		this.add('Layout_Custom_Body', false, {
+		this.add('Layout_Custom_Body_Only', false, {
```
sampaiodiego, line 1558: is a new setting really needed? this is not *renaming* the old setting, it is creating a new one.

if a new setting is really required, the old one should be removed.

sampaiodiego (changes requested):

apps/meteor/ee/app/settings/server/settings.ts
tassoevan: @sampaiodiego It was not the behavior before @ggazzo refactor of the settings API. Also, it implies all enterprise settings *MUST HAVE* a list of modules they are contained. Is it right? If so, which module fits here?

sampaiodiego: oh, I see, this check is just in case an enterprise setting doesn't have a list of modules.. since there is an `isEnterprise()` validation before, I'm ok with this change then.. 

initially it wouldn't make sense have an enterprise license without modules defined, because there is no enterprise license without a module, but I can see it may be a thing.

apps/meteor/app/lib/server/startup/settings.ts
hugocostadev: Hi @sampaiodiego , yes the setting is really needed, the migration was not created because the 'old' setting was not released yet, it was only merged at develop branch couple of days ago.

sampaiodiego: I see, thx @hugocostadev .. we'll need to be more careful on things like these once we start rolling upgrades even before they got released, but for now, I think we're good 👍

sampaiodiego (dismissed):

tassoevan (dismissed):

ggazzo:
is this a chore? or more like a regression? changing the name just in case

hugocostadev (dismissed):

hugocostadev (approved):

ggazzo (approved):

PR #26572, opened 2022-08-15
Chore: Fail-fast on callbacks

KevLehman:
## Further comments

The idea of this PR is to skip some omnichannel-only callbacks when the room/message being passed is not from Omnichannel. This will allow other messages to perform a little bit better.


apps/meteor/ee/app/livechat-enterprise/server/hooks/handleNextAgentPreferredEvents.js
```diff
@@ -25,6 +25,10 @@ const checkDefaultAgentOnNewRoom = (defaultAgent, defaultGuest) => {
 		return defaultAgent;
 	}
 
+	if (!lastChattedAgentPreferred) {
+		return defaultAgent;
+	}
+
```
murtaza98, line 31: This changes the current logic when "Contact manager preferred routing" is on but "last preferred agent routing" is off. Based on the current logic, it won't transfer the chat to the contactManager.

Maybe you can add a check for both settings here, like

```
if (!lastChattedAgentPreferred && !contactManagerPreferred) {
		return defaultAgent;
}
```

KevLehman: :thinking: I think I'll leave this one where it was before, since seems it affects some functionality

murtaza98 (approved):

KevLehman:
@sampaiodiego , from our side, no changes would change "functionality", so we should be good skipping QA for this one after you review it :eyes:

apps/meteor/app/models/server/models/Subscriptions.js
```diff
@@ -28,9 +28,9 @@ export class Subscriptions extends Base {
 		this.tryEnsureIndex({ desktopNotifications: 1 }, { sparse: 1 });
 		this.tryEnsureIndex({ mobilePushNotifications: 1 }, { sparse: 1 });
 		this.tryEnsureIndex({ emailNotifications: 1 }, { sparse: 1 });
+		this.tryEnsureIndex({ audioNotifications: 1 }, { sparse: 1 });
```
sampaiodiego, line 31: not sure where this came from 🤔

apps/meteor/app/models/server/models/Subscriptions.js
```diff
@@ -12,8 +12,8 @@ export class Subscriptions extends Base {
 	constructor(...args) {
 		super(...args);
 
-		this.tryEnsureIndex({ rid: 1 });
```
sampaiodiego, line 15: how are we going to remove this index later on? 👀

apps/meteor/app/lib/server/lib/sendNotificationsOnMessage.js
```diff
@@ -372,8 +372,11 @@ export async function sendAllNotifications(message, room) {
 			}),
 		)
 			.then((users) => {
+				const subscriptions = Subscriptions.findByRoomIdAndUserIds(room._id, users, {
+					fields: { 'u._id': 1 },
+				});
```
sampaiodiego, line 377: you need all fields here
```suggestion
				const subscriptions = Subscriptions.findByRoomIdAndUserIds(room._id, users);
```

sampaiodiego (changes requested):

apps/meteor/app/models/server/models/Subscriptions.js
KevLehman: Hmmm, copied this from the old PR. I kinda remember this prop was used on "notifications", but cannot find any ref to it anymore

apps/meteor/app/models/server/models/Subscriptions.js
KevLehman: I have the migration, but migrations are a sin now, so not sure if I should sin :trollface:

sampaiodiego: we got rid of this already 😬

apps/meteor/app/models/server/models/Subscriptions.js
```diff
@@ -12,8 +12,8 @@ export class Subscriptions extends Base {
...
 		this.tryEnsureIndex({ rid: 1, ls: 1 });
+		this.tryEnsureIndex({ 'userHighlights.0': 1, 'rid': 1 });
```
sampaiodiego, line 16: what if you revert the changes to indexes and we do them in another PR?

KevLehman: Fair :eyes:

apps/meteor/app/models/server/models/Subscriptions.js
```diff
@@ -30,7 +30,6 @@ export class Subscriptions extends Base {
 		this.tryEnsureIndex({ emailNotifications: 1 }, { sparse: 1 });
 		this.tryEnsureIndex({ autoTranslate: 1 }, { sparse: 1 });
 		this.tryEnsureIndex({ autoTranslateLanguage: 1 }, { sparse: 1 });
-		this.tryEnsureIndex({ 'userHighlights.0': 1 }, { sparse: 1 });
```
sampaiodiego, line 33: still removing an index 🙈

sampaiodiego (changes requested):

sampaiodiego (approved):

apps/meteor/app/models/server/models/Subscriptions.js
KevLehman: Wait, i removed that one lol. Thanks for fixing it.

PR #26574, opened 2022-08-15
Regression: Team name validation failing always.

gabriellsh:


LucianoPierdona (approved):

ggazzo (changes requested):
could you create a test for this case? just to be sure we never gonna repeat this mistake again :x

ggazzo (approved):

ggazzo:
I heard @souzaramon is covering this test case so we are good to go

PR #26575, opened 2022-08-16
[FIX] Agents (with user status offline & omni-status as available) not able to take or forward chat

murtaza98:
## Issue(s)

Issue mentioned [here](https://app.clickup.com/t/2nnw9b0?comment=1123321986)

## Further comments

Note: this PR was built on top of https://github.com/RocketChat/Rocket.Chat/pull/26147 to fix the issue raised by QA


KevLehman (approved):

KevLehman:
Lol, no, this is not a regression. Changed the title. 

The other 2 points, @murtaza98 :trollface:

apps/meteor/tests/e2e/utils/test.ts
```diff
@@ -71,7 +71,7 @@ export const test = baseTest.extend<BaseTest>({
 				return request.put(BASE_API_URL + prefix + uri, { headers, data });
 			},
 			delete(uri: string, prefix = API_PREFIX) {
-				return request.delete(BASE_API_URL + prefix + uri, { headers });
+				return request.delete(BASE_URL + prefix + uri, { headers });
```
sampaiodiego, line 74: `BASE_API_URL` is actually set like this:
```
export const BASE_API_URL = BASE_URL + API_PREFIX;
```

so I guess you should either remove `BASE_API_URL` , or keep using it and remove `prefix` from here..

btw, this needs to be changed for all operations

sampaiodiego (changes requested):

ggazzo (dismissed):
we have a special test waiting the agent number 2 get online to transfer 

```
	await test.step('Expect to connect as agent 2', async () => {
			agent2 = await createAuxContext(browser, 'user2-session.json'); <---- HERE
		

```

please create a second test trying to assign to a offline agent as well 

thanks

apps/meteor/tests/e2e/utils/test.ts
murtaza98: So I went with the first approach & removed `BASE_API_URL` altogether. Also modified it for the other methods :smile:

apps/meteor/client/components/Omnichannel/modals/ForwardChatModal.tsx
```diff
@@ -43,7 +44,26 @@ const ForwardChatModal = ({
...
+					roles: {
+						$ne: 'bot',
+					},
+				},
+				{
+					roles: 'bot',
+				},
+			],
+		}),
+		statusLivechat: 'available',
```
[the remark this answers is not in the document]
murtaza98: > here we are considering statusLivechat only here, but probably there are many places where a offline, but active agent should appears

@ggazzo Sorry, I didn't get this part :see_no_evil: 

Also, any chance we can take this task separately to remove the dependency of `UserAutoComplete` component within the forwarding chat modal? Because the release cut is quite close & I don't think I'd be able to replace this entire behaviour in time :sweat_smile: (We don't even have any endpoints already present for this right now :grimacing: )

murtaza98:
> we have a special test waiting the agent number 2 get online to transfer
> 
> ```
> 	await test.step('Expect to connect as agent 2', async () => {
> 			agent2 = await createAuxContext(browser, 'user2-session.json'); <---- HERE
> 		
> ```
> 
> please create a second test trying to assign to a offline agent as well
> 
> thanks

Added this test case as well here :heavy_check_mark:

KevLehman (approved):

tassoevan (approved):

PR #26623, opened 2022-08-19
[FIX] Check caller room membership before sending message

debdutdeb:
## Issue(s)

Fixes: 26266


apps/meteor/tests/end-to-end/api/05-chat.js
```diff
@@ -512,6 +512,7 @@ describe('[Chat]', function () {
 	});
 
 	describe('/chat.sendMessage', () => {
+		// FIXME do here
```
debdutdeb, line 515:
```suggestion
```

sampaiodiego:
any reason this is still in draft?

while it makes sense, did you confirm this change with the product team?

PR #26634, opened 2022-08-19
Chore: removing useEndpointData from license api

hugocostadev:

The deprecated hook `useEndpointData` was removed and changed to `useMethod` + `useQuery`separated in a hook.

In `useQuery` options we added an option to keep the cache as `Infinity` until the queries are invalidated or the page refreshes, also the query keys were changed to consolidate the keys and manage the cache in the best way possible to avoid unnecessary API calls. 

In settings changed or user action that impact the License status was added a method to invalidate the cache for license query keys

E.g: `queryClient.invalidateQueries(['licenses']);`


apps/meteor/client/hooks/useIsEnterprise.ts
```diff
@@ -1,8 +1,13 @@
 import { useEndpoint } from '@rocket.chat/ui-contexts';
 import { useQuery } from '@tanstack/react-query';
 
-export const useIsEnterprise = (): boolean => {
+export const useIsEnterprise = (): { isEnterprise: boolean; isLoading: boolean } => {
```
gabriellsh, line 4: Please, don't hide the return of `useQuery`. Always return all the properties. There's a helper type from `@tanstack/react-query` called `UseQueryResult`, that you can use to wrap the type that comes in `data`. See `apps/meteor/client/hooks/useRoomInfoEndpoint.ts` for more context.

apps/meteor/client/hooks/useLicense.ts
```diff
@@ -0,0 +1,32 @@
+import { CloudRegistrationStatus, ILicense } from '@rocket.chat/core-typings';
+import { useEndpoint } from '@rocket.chat/ui-contexts';
+import { useQuery } from '@tanstack/react-query';
+
+export const useLicense = (): {
```
gabriellsh, line 5: As we talked on a DM, move this to 2 separate hooks.

apps/meteor/client/hooks/useLicense.ts
```diff
@@ -0,0 +1,32 @@
...
+	licenses: ILicense[] | undefined;
+	registrationStatus: CloudRegistrationStatus | undefined;
+	isError: boolean;
+	isLoading: boolean;
+	isSuccess: boolean;
+} => {
+	const getRegistrationStatus = useEndpoint('GET', '/v1/cloud.registrationStatus');
+	const getLicenses = useEndpoint('GET', '/v1/licenses.get');
+
+	const { data: registrationStatusData } = useQuery(['licenses', 'getRegistrationStatus'], () => getRegistrationStatus(), {
```
gabriellsh, line 15: `registrationStatus` and `licenses` are somewhat related, but at the same time independent from each other. Can you change this key? Thanks.

gabriellsh (changes requested):

apps/meteor/client/hooks/useIsEnterprise.ts
hugocostadev: Done!

apps/meteor/client/hooks/useLicense.ts
hugocostadev: Done!

apps/meteor/client/hooks/useLicense.ts
hugocostadev: Done!

gabriellsh (approved):

PR #26685, opened 2022-08-25
[FIX] Units filter not working on current chats panel

murtaza98:

### DIFF
--- a/apps/meteor/ee/app/models/server/raw/LivechatRooms.js
+++ b/apps/meteor/ee/app/models/server/raw/LivechatRooms.js
@@ -13,6 +13,7 @@ const applyRestrictions = (method) =>
 overwriteClassOnLicense('livechat-enterprise', LivechatRoomsRaw, {
 	find: applyRestrictions('find'),
 	update: applyRestrictions('update'),
+	findPaginated: applyRestrictions('findPaginated'),
 	remove: applyRestrictions('remove'),
 	updateDepartmentAncestorsById(originalFn, _id, departmentAncestors) {
 		const query = {


apps/meteor/ee/app/models/server/raw/LivechatRooms.ts
```diff
@@ -0,0 +1,78 @@
...
+									departmentAncestors: { $ne: unitId },
+								},
+							],
+						},
+					],
+				},
+			],
+		};
+		const update = { $set: { departmentAncestors: [unitId] } };
+		queriesLogger.error({ msg: `LivechatRoomsRaw.associateRoomsWithDepartmentToUnit - association step`, query, update });
```
KevLehman, line 52: Do we need these logs?

apps/meteor/ee/app/models/server/raw/LivechatRooms.ts
```diff
@@ -0,0 +1,78 @@
...
+		const associationResult = await this.update(query, update, { multi: true });
+		queriesLogger.error({ msg: `LivechatRoomsRaw.associateRoomsWithDepartmentToUnit - association step`, result: associationResult });
+
+		const queryToDisassociateOldRoomsConnectedToUnit = {
+			departmentAncestors: unitId,
+			departmentId: { $nin: departments },
+		};
+		const updateToDisassociateRooms = { $unset: { departmentAncestors: 1 } };
+		queriesLogger.error({
```
KevLehman, line 61: :eyes:

apps/meteor/ee/app/models/server/raw/LivechatRooms.ts
```diff
@@ -0,0 +1,78 @@
...
+		});
+		const disassociationResult = await this.update(queryToDisassociateOldRoomsConnectedToUnit, updateToDisassociateRooms, { multi: true });
+		queriesLogger.error({ msg: `LivechatRoomsRaw.associateRoomsWithDepartmentToUnit - disassociation step`, result: disassociationResult });
+	},
+	async removeUnitAssociationFromRooms(_, unitId: string) {
+		const query = {
+			departmentAncestors: unitId,
+		};
+		const update = { $unset: { departmentAncestors: 1 } };
+		queriesLogger.debug({ msg: `LivechatRoomsRaw.removeUnitAssociationFromRooms`, query, update });
```
KevLehman, line 74: :grimacing:

apps/meteor/ee/app/models/server/raw/LivechatRooms.ts
```diff
@@ -0,0 +1,78 @@
...
+		const result = await this.update(query, update, { multi: true });
```
KevLehman, line 75: I thought raw models had "this.update" deprecated in favor of updateMany :thinking:

apps/meteor/ee/app/models/server/raw/LivechatRooms.ts
```diff
@@ -0,0 +1,78 @@
+import { LivechatRoomsRaw } from '../../../../../server/models/raw/LivechatRooms';
+import { queriesLogger } from '../../../livechat-enterprise/server/lib/logger';
+import { addQueryRestrictionsToRoomsModel } from '../../../livechat-enterprise/server/lib/query.helper';
+import { overwriteClassOnLicense } from '../../../license/server';
+
+const applyRestrictions = (method: 'find' | 'update' | 'findPaginated') =>
+	function (originalFn: any, originalQuery: object, ...args: any) {
```
KevLehman, line 7: Not sure, but, can you try something around `Parameters<LivechatRoomsRaw[TMethod]>`?

KevLehman (changes requested):

apps/meteor/ee/app/models/server/raw/LivechatRooms.ts
murtaza98: Oops! They were supposed to be just debug logs :grimacing: Changed them back to debug :heavy_check_mark:

apps/meteor/ee/app/models/server/raw/LivechatRooms.ts
murtaza98: This one was already debug log :eyes:

apps/meteor/ee/app/models/server/raw/LivechatRooms.ts
murtaza98: Yeah good point, let me change that :+1:

apps/meteor/ee/app/models/server/raw/LivechatRooms.ts
murtaza98: This was getting way too complex due to the types of override class functions, so I decided to extend the Raw LivechatRoomsRaw class & then re-register it simply. It looks much simpler now, and the base models' types are also better preserved. Do let me know if you have any suggestions about this :smile:

apps/meteor/ee/app/license/server/getStatistics.ts
```diff
@@ -31,39 +31,42 @@ export async function getStatistics(): Promise<ENTERPRISE_STATISTICS> {
...
-	);
 
-	// Number of canned responses
-	statsPms.push(
-		CannedResponse.col.count().then((count) => {
-			statistics.cannedResponses = count;
-			return true;
-		}),
-	);
+	if (hasLicense('livechat-enterprise')) {
```
KevLehman, line 35: Can you extract this into its own function and just call it when license is there? (I assume this is since models are inside an onlicense now, so it's required)

apps/meteor/ee/app/livechat-enterprise/server/hooks/afterForwardChatToDepartment.js
```diff
@@ -13,7 +15,7 @@ callbacks.add(
 			cbLogger.debug('Skipping callback. No room found');
 			return options;
 		}
-		LivechatRooms.unsetPredictedVisitorAbandonmentByRoomId(room._id);
+		LegacyLivechatRoomsModel.unsetPredictedVisitorAbandonmentByRoomId(room._id);
```
KevLehman, line 18: Can you migrate this function to the raw model?

apps/meteor/ee/app/models/server/models/LivechatUnit.js
```diff
@@ -105,6 +106,8 @@ export class LivechatUnit extends LivechatDepartment {
 			);
 		});
 
+		Promise.await(LivechatRooms.associateRoomsWithDepartmentToUnit(departmentsToSave, _id));
+
 		return _.extend(record, { _id });
```
KevLehman, line 111:
```suggestion
		return { _id, ...record }
```

KevLehman (changes requested):

apps/meteor/ee/app/models/server/models/LivechatUnit.js
murtaza98: Sorry but I had to revert this back :disappointed:  because this is causing some weird issues in some endpoints defined here because some of the helper files are still in js and the typescript isn't able to correctly resolve its types causing issues -> `apps/meteor/ee/app/livechat-enterprise/server/api/units.ts`. Will check this again when we attempt to covert LivechatUnit model to typescript

apps/meteor/ee/app/license/server/getStatistics.ts
murtaza98: Sure, its done :heavy_check_mark:

apps/meteor/ee/app/livechat-enterprise/server/hooks/afterForwardChatToDepartment.js
murtaza98: Done :smile:

apps/meteor/ee/app/models/server/models/LivechatUnit.js
KevLehman: Oh, if it was a "typings" problems, you can use Object.assign, and that should work 👀

packages/core-typings/src/IRoom.ts
```diff
@@ -192,6 +192,9 @@ export interface IOmnichannelGenericRoom extends Omit<IRoom, 'default' | 'featur
 
 export interface IOmnichannelRoom extends IOmnichannelGenericRoom {
 	t: 'l';
+	omnichannel?: {
+		predictedVisitorAbandonmentAt: Date;
```
KevLehman, line 196: Shouldn't this be optional too? 👀 not sure if we add other props to "omnichannel" key.

apps/meteor/ee/app/license/server/getStatistics.ts
```diff
@@ -31,6 +31,25 @@ export async function getStatistics(): Promise<ENTERPRISE_STATISTICS> {
...
 		}),
 	);
+
+	if (hasLicense('livechat-enterprise')) {
+		const eeModelsStats = await getStatsFromEEModels();
+		statistics = {
+			...statistics,
+			...eeModelsStats,
+		};
+	}
```
KevLehman, line 41: Move this check to inside of "getStatsFromEEModels" func

apps/meteor/ee/app/license/server/getStatistics.ts
```diff
@@ -31,6 +31,25 @@ export async function getStatistics(): Promise<ENTERPRISE_STATISTICS> {
...
+
+	await Promise.all(statsPms).catch(log);
+	return statistics;
+}
+
+// These models are only available on EE license so don't import them inside CE license as it will break the build
+async function getStatsFromEEModels(): Promise<ENTERPRISE_STATISTICS> {
```
KevLehman, line 48:
```suggestion
async function getEEStatistics(): Promise<ENTERPRISE_STATISTICS> {
```

apps/meteor/ee/app/license/server/getStatistics.ts
```diff
@@ -18,7 +18,7 @@ type ENTERPRISE_STATISTICS = {
 export async function getStatistics(): Promise<ENTERPRISE_STATISTICS> {
 	const statsPms: Array<Promise<any>> = [];
 
-	const statistics: ENTERPRISE_STATISTICS = {} as any;
+	let statistics: ENTERPRISE_STATISTICS = {} as any;
```
KevLehman, line 21: Every time you type any, a ts dev dies

KevLehman (dismissed):

packages/core-typings/src/IRoom.ts
murtaza98: I checked this and didn't find any other prop being set inside "omnichannel", so I think we should be safe here :smiley:

apps/meteor/ee/app/license/server/getStatistics.ts
murtaza98: Lol :joy:

apps/meteor/ee/app/models/server/models/LivechatUnit.js
murtaza98: This should be fixed now :smile:

apps/meteor/ee/app/license/server/getStatistics.ts
murtaza98: Done :heavy_check_mark:

apps/meteor/ee/app/license/server/getStatistics.ts
murtaza98: Fixed those "any" typings on this file :tada:

packages/core-typings/src/IStats.ts
```diff
@@ -133,10 +133,10 @@ export interface IStats {
 		modules: string[];
 		tags: string[];
 		seatRequests: number;
-		livechatTags: number;
-		cannedResponses: number;
-		priorities: number;
-		businessUnits: number;
+		livechatTags?: number;
```
KevLehman, line 136: Hey @cauefcr , you worked with stats collector recently. Do you think changing these props to be possibly undefined would affect how they are collected?

KevLehman (approved):
I'll approve it 👀 so we can start QAing, but let's wait on confirmation about stats collector b4 merging it

packages/core-typings/src/IRoom.ts
```diff
@@ -189,6 +189,9 @@ export interface IOmnichannelGenericRoom extends Omit<IRoom, 'default' | 'featur
 
 export interface IOmnichannelRoom extends IOmnichannelGenericRoom {
 	t: 'l';
+	omnichannel?: {
```
ggazzo, line 192: @sampaiodiego

ggazzo: I got lost here, are we adding an optional property `omnichannel` to an `omnichannel` room? what is the advantage of it? force everyone to do "Optional chaining" everywhere?
like `room.omnichannel?.predictedVisitorAbandonmentAt` ?


right now that we've managed to type pretty much everywhere, and we know we're dealing with omnichannel rooms in the code, are we going to add optional structures everywhere?

ggazzo: I would ask to stay "shallow"  as the other 100 props we have for omnichannel and other room typews

ggazzo (changes requested):

packages/core-typings/src/IRoom.ts
murtaza98: Yeah, I agree, the nested chaining is obviously not required here, especially since this "omnichannel" prop only has a single field "predictedVisitorAbandonmentAt". But since I wasn't using this prop anywhere in this fix, I didn't modify the logic there, rather just updated our types. However, if you guys are okay with including those migrations here do let me know and I can update this PR accordingly - or if you want me to create a separate PR for it, that works too :smiley: 

Wrt the type is optional, that is actually intentional as this prop will only be set under specific cases even in omnichannel rooms so it's not a generic field which all omnichannel rooms would have, hence it's marked optional.

KevLehman: I'd prefer doing that migration on other PR.

As a side note: `omnichannel.xxx` is a really old prop, not being added by the PR. Someone decided to put an omnichannel key with keys inside. All other omni props are top level. This is a design mistake and it indeed looks ugly, but migrating things should be OOS for this PR.

ggazzo (changes requested):

apps/meteor/ee/server/models/startup.ts
```diff
@@ -1,5 +1,10 @@
...
-import './LivechatUnitMonitors';
+import { onLicense } from '../../app/license/server/license';
+
+onLicense('livechat-enterprise', () => {
+	require('./CannedResponse');
+	require('./LivechatPriority');
+	require('./LivechatTag');
+	require('./LivechatUnit');
+	require('./LivechatUnitMonitors');
+	require('./LivechatRooms');
```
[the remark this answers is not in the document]
murtaza98: Sure, changed it to import :heavy_check_mark:

KevLehman (approved):

ggazzo (approved):

PR #26687, opened 2022-08-25
[IMPROVE] Results of user auto complete

pierre-lehnen-rc:
When searching for users on autocomplete, list first the users that the requester has DMs with;
If the requester doesn't have permission to view outside their assigned rooms, search for users that have rooms in common instead of returning an empty list.


ggazzo (changes requested):
please check my comments above, feel free to dismiss if makes no sense

apps/meteor/app/api/server/v1/users.ts
```diff
@@ -815,11 +817,32 @@ API.v1.addRoute(
...
+				exceptions: [],
+				conditions: {},
+				term: '',
+			};
+
+			try {
+				selector = JSON.parse(selectorRaw);
+
+				if (selector) {
+					selector = clean(selector) as typeof selector;
```
sampaiodiego, line 832: from what I can see, `clean` will remove every key starting with `$`, but on `isValidQuery` called below you're allowing `$or` and `$and`, which won't ever happen.

so you either want:
- not call `clean` here (my suggestion)
- remove `isValidQuery` call since there will be no special selector

apps/meteor/tests/end-to-end/api/01-users.js
```diff
@@ -65,6 +65,31 @@ function deleteTestUser(user) {
...
+		request
+			.post(api('channels.create'))
+			.set(userCredentials)
+			.send({
+				name,
+			})
+			.then((res) => resolve(res.body.channel._id))
+			.catch(reject);
+	});
+}
```
sampaiodiego, line 79: you can use async/await here:

```suggestion
async function createChannel(userCredentials, name) {
	const res = await request
		.post(api('channels.create'))
		.set(userCredentials)
		.send({
			name,
		});
	return res.body.channel._id;
}
```

apps/meteor/tests/end-to-end/api/01-users.js
```diff
@@ -65,6 +65,31 @@ function deleteTestUser(user) {
...
+
+function joinChannel(userCredentials, roomId) {
+	return new Promise((resolve) => {
```
sampaiodiego, line 82: same as above

apps/meteor/tests/end-to-end/api/01-users.js
```diff
@@ -3105,33 +3154,51 @@ describe('[Users]', function () {
...
+						.set(userCredentials)
+						.expect('Content-Type', 'application/json')
+						.expect(200)
+						.expect((res) => {
+							expect(res.body).to.have.property('success', true);
+							expect(res.body).to.have.property('items').and.to.be.an('array').with.lengthOf(1);
+						})
+						.end(done);
+				});
+			});
```
sampaiodiego, line 3171: pls prefer async/await:
```suggestion
			it('should return users that are subscribed to the same rooms as the requester', async () => {
				await joinChannel(user2Credentials, roomId);
				request
					.get(api('users.autocomplete?selector={}'))
					.set(userCredentials)
					.expect('Content-Type', 'application/json')
					.expect(200)
					.expect((res) => {
						expect(res.body).to.have.property('success', true);
						expect(res.body).to.have.property('items').and.to.be.an('array').with.lengthOf(1);
					});
			});
```

apps/meteor/tests/end-to-end/api/01-users.js
```diff
@@ -3105,33 +3154,51 @@ describe('[Users]', function () {
...
 		});
-		it('should return an error when the required parameter "selector" is not provided', (done) => {
-			updatePermission('view-outside-room', ['admin', 'user']).then(() => {
+
+		describe('[with permission]', () => {
+			it('should return an error when the required parameter "selector" is not provided', (done) => {
+				updatePermission('view-outside-room', ['admin', 'user']).then(() => {
```
sampaiodiego, line 3176: pls prefer using async/await here as well

sampaiodiego (changes requested):

apps/meteor/app/api/server/v1/users.ts
pierre-lehnen-rc: There is a bug in `clean` so this is actually not happening. I'm adding extra tests to ensure the outcome is the expected one.

sampaiodiego (approved):

sampaiodiego (approved):

PR #26689, opened 2022-08-25
[FIX] Avatars of other chats disappear when they located near chat with broken avatar

filipemarins:
## Steps to test or reproduce

1. Have one user or chat with a broken avatar on the sidebar (You can reproduce it by blocking an avatar URL on the network tab in the browser)
2. Grap any chat below the chat with a broken avatar and send a message. (It will force reorder the chat position on the sidebar)
3. Look at other chats near that chat with a broken avatar - their avatars disappear

https://user-images.githubusercontent.com/9275105/186748058-b82c8b81-7782-4acf-81af-3873ff2a7a01.mov

## Further comments

During the list re-render on the sidebar, React was have some issues rendering the correct element for the room avatar component, passing a key prop to a component will allow React engine track better dom changes.


gabriellsh (approved):

ankar84:
Thanks a lot!!!

PR #26692, opened 2022-08-25
[FIX] Omnichannel inquiries being updated even if not needed

KevLehman:
… needed


sampaiodiego (approved):

KevLehman:
How's this increasing UI coverage is a mystery.

PR #26697, opened 2022-08-26
[NEW] URI allowed schemes

Chirtep:

1. This feature by itself simply adds the ability to expand the scheme of working protocols (schemes), but does not add a new single scheme by default, which means no affect on security of app.

2. Feature allows the admin to add certain schemes in the settings (section "messages - markdown"). The admin himself is responsible for his server and must understand what he is doing and what it can lead to.

image


apps/meteor/client/components/MarkdownText.tsx
```diff
@@ -109,7 +110,14 @@ const MarkdownText: FC<Partial<MarkdownTextParams>> = ({
 			}
 		})();
 
-		return preserveHtml ? html : html && sanitizer(html, { ADD_ATTR: ['target'] });
+		const URIschemes = new RegExp(
+			'^(?:(?:(?:f|ht)tps?|mailto|tel|callto|cid|xmpp|' +
+				`${settings.get('Markdown_URIAllowedSchemes')}` +
```
filipemarins, line 115:
```suggestion
				`${settings.get('Markdown_SupportSchemesForLink').split(',').join('|')}` +
```

I don't think we should create a new preference just for this case... We could use the `Markdown_SupportSchemesForLink` setting.

filipemarins (changes requested):
Could you please also sign the [Contributor License Agreement](https://cla-assistant.io/RocketChat/Rocket.Chat?pullRequest=26697)?

gabriellsh:
Hi @Chirtep ! Thanks for your contribution!

Unfortunately, this will no longer be useful since we now only offer proper support to our own parser (https://github.com/RocketChat/fuselage/tree/develop/packages/message-parser).

Also, the place you modified only affects a few spare places that use the MarkdownText component. This meaning most important places (such as the message) would not change in any way.

Our new parser is way more permissive in regards of link protocols, and we're actively working to make it better. I'll be closing this for now, but again, thanks a lot for the effort!

gabriellsh:
I made a very similar but yet different fix here: https://github.com/RocketChat/Rocket.Chat/pull/27245

PR #26704, opened 2022-08-26
Chore: Convert AutoCompleteAgent to tsx

juliajforesti:


apps/meteor/client/hooks/lists/useRecordList.ts
```diff
@@ -4,7 +4,7 @@ import { useEffect, useState } from 'react';
 import { AsyncStatePhase } from '../../lib/asyncState';
 import { RecordList } from '../../lib/lists/RecordList';
 
-type RecordListValue<T> = {
+export type RecordListValue<T> = {
```
MartinSchoeler, line 7: Are we importing this in any place?

apps/meteor/client/components/AutoCompleteAgent.tsx
```diff
@@ -1,14 +1,24 @@
...
 
 import { useRecordList } from '../hooks/lists/useRecordList';
 import { AsyncStatePhase } from '../lib/asyncState';
 import { useAgentsList } from './Omnichannel/hooks/useAgentsList';
 
-const AutoCompleteAgent = (props) => {
-	const { value, onChange = () => {}, haveAll = false, haveNoAgentsSelectedOption = false } = props;
-	const [agentsFilter, setAgentsFilter] = useState('');
+type AutoCompleteAgentProps = {
+	value: string | number | readonly string[];
```
MartinSchoeler, line 10: Is there a moment where `value` can be a number?

MartinSchoeler: it does not make much sense since `onChange: (value: string) => void;`

apps/meteor/client/components/AutoCompleteAgent.tsx
```diff
@@ -42,10 +48,12 @@ const AutoCompleteAgent = (props) => {
 			onChange={onChange}
 			flexShrink={0}
 			filter={agentsFilter}
-			setFilter={setAgentsFilter}
+			setFilter={setAgentsFilter as (value: string | number | undefined) => void}
```
MartinSchoeler, line 51: Again, can this really be a number?

MartinSchoeler (changes requested):

MartinSchoeler (approved):

PR #26706, opened 2022-08-26
Chore: Remove & Test cannedResponse meteor templates

MartinSchoeler:


apps/meteor/tests/e2e/omnichannel-canned-responses-sidebar.spec.ts
```diff
@@ -0,0 +1,63 @@
...
+		});
+
+		await test.step('Expect to have 1 omnichannel assigned to agent 1', async () => {
+			await agent.poHomeChannel.sidenav.openChat(newUser.name);
+		});
+
+		await test.step('Expect to be able to open canned responses sidebar and creation', async () => {
+			await agent.poHomeChannel.content.btnCannedResponses.click();
+			await agent.poHomeChannel.content.btnNewCannedResponse.click();
+		});
```
aleksandernsilva, line 61: Shouldn't we have at least one assertion here to confirm if the screen is in the expected state?

aleksandernsilva (changes requested):

apps/meteor/tests/e2e/omnichannel-canned-responses-sidebar.spec.ts
MartinSchoeler: The idea is to create the framework of the test so the coverage can detect the changes

ggazzo (approved):

PR #26714, opened 2022-08-26
Chore: first non-aggressive CSS removal

ggazzo:


MartinSchoeler (approved):
Couldn't find any obvious styling issues on the omnichannel side. LGTM

PR #26720, opened 2022-08-29
Regression: Empty custom-fields filter on Current Chats causing issues

murtaza98:
## Further comments

Clickup: https://app.clickup.com/t/2r65yjw


apps/meteor/app/livechat/imports/server/rest/rooms.js
```diff
@@ -43,7 +44,11 @@ API.v1.addRoute(
 			}
 
 			if (customFields) {
-				customFields = JSON.parse(customFields);
+				const parsedCustomFields = JSON.parse(customFields);
```
KevLehman: Wrap JSON parse in a try/catch

KevLehman (approved):

PR #26724, opened 2022-08-29
Test/improve reliability

souzaramon:


weslley543 (commented):
In therms of code is ok. But needs to verify why in the ci is failing !

PR #26727, opened 2022-08-29
Chore: Marketplace initial e2e coverage

rique223:

Created playwright e2e tests to test the functionalities of the marketplace search bar


tiagoevanp (changes requested):
Use role and name, avoid data-qa-id usage.

PR #26730, opened 2022-08-29
Chore: Add E2E tests to missing omnichannel endpoints

KevLehman:


apps/meteor/tests/end-to-end/api/livechat/01-agents.ts
```diff
@@ -323,10 +323,53 @@ describe('LIVECHAT - Agents', function () {
...
+		});
+		it('should fail when token is valid but rid isnt', async () => {
+			const visitor = await createVisitor();
+			await request.get(api(`livechat/agent.info/invalid-rid/${visitor.token}`)).expect(400);
+		});
+		it('should fail when room is not being served by any agent', async () => {
+			const visitor = await createVisitor();
+			const room = await createLivechatRoom(visitor.token);
+			await request.get(api(`livechat/agent.info/${room._id}/${visitor.token}`)).expect(400);
+		});
```
murtaza98, line 339: This particular test seems to be failing on the build. I think it is because the routing algo is automatically assigning the room to an agent however this particular test case is expecting that room to be not served by any agent. So maybe, try to switch the routing algo here to manual to avoid agent assignment :thinking:

murtaza98 (changes requested):

murtaza98 (approved):

sampaiodiego (approved):

PR #26754, opened 2022-08-30
[IMPROVE] Read threads when marking a room as read

### LINKED ISSUE #22802: Mark read in chat menu doesn't affect unread threads
### Description:

**Mark read** option in chat menu doesn't mark chat as read if there is unread messages in threads

### Steps to reproduce:

1. Get unread message in thread in some chat
2. Try to mark that chat read
3. Chat still unread and with count in blue circle

### Expected behavior:

Chat become mark as read

### Actual behavior:

Here is a chat with unread message in thread
image
Try to mark room read
image
Even if user open chat and read all main chat messages there still count for unread thread
image

### Server Setup Information:

- Version of Rocket.Chat Server: 3.16.3
- Operating System: CentOS7
- Deployment Method: docker
- NodeJS Version: 12.22.1
- MongoDB Version: 4.2.14

### Client Setup Information

- Desktop App or Browser Version: 3.2.3 and Chrome 92

### Additional context

I understand that thread unread messages it's a bit different thing then main chat messages.
But users need a button to mark chat as read despite of threads.
Maybe some option in users preferences like "Mark threads read too on mark chat as read" or something like that.

Now it's not a but, but it's not complete work of **Mark read** button, which must be improved.

### Relevant logs:

No

LucianoPierdona:

This PR changes the functionality of `subscription.read` to add read threads when endpoint is called

## Issue(s)

Closes #22802


debdutdeb (dismissed):
This is a bit weird. When I look at `markRoomAsRead` the behavior it insists on is "not marking a room as read if there are unread threads". But then you're explicitely unsetting all unread threads for a seemingly similar method (`readMessages` ~== `markRoomAsRead`). Seems inconsistent. 

IMO the way should be to update the subscription fields all at once, no reason to update the same set of documents twice. I don't see the point of having a separate `markAllThreadsAsReadByRidAndUid`-like function either, since nowhere in rocketchat we have a way to **just** mark threads as read today.

As you're taking away the distinction, you can update the model directly to reflect that, and remove any legacy code :)

LucianoPierdona:
> i

@debdutdeb 
Updated `markRoomAsRead` and removed unused code.

debdutdeb:
Couple notes for the reviewer:
- Primary goal as per the title of the pr is to unset thread unread counts when `subscriptions.read` is called.
- From the users' side, we only want two actions to actually do that, 1) "mark room as read" button in the burger menu from the sidebar room entry, 2) `Shift + Esc` keybind.
- Both direct "mark as read" actions and open room event uses the same backend function.

To cover both we could either add another argument like `readThreads bool` or decouple both logic. It felt like decoupling makes it easier to understand. But open to other reasoning.

apps/meteor/tests/end-to-end/api/10-subscriptions.js
```diff
@@ -208,6 +211,110 @@ describe('[Subscriptions]', function () {
...
+						threadId = res.body.message._id;
+					});
+			});
+
+			after((done) => {
+				deleteUser(user).then(done);
+			});
+
+			beforeEach(async () => {
+			});
```
debdutdeb, line 242:
```suggestion
```

apps/meteor/app/api/server/v1/subscriptions.ts
```diff
@@ -77,9 +79,15 @@ API.v1.addRoute(
...
 	{
-		post() {
-			Meteor.call('readMessages', this.bodyParams.rid);
-
+		async post() {
+			const { rid, readThreads } = this.bodyParams;
+			check(rid, String);
+			check(readThreads, Match.Maybe(Boolean));
+			try {
+				await readMessages(rid, this.userId, readThreads ?? true);
```
debdutdeb, line 87:
```suggestion
				await readMessages(rid, this.userId, readThreads === undefined ? true : readThreads);
```

apps/meteor/app/api/server/v1/subscriptions.ts
```diff
@@ -77,9 +79,15 @@ API.v1.addRoute(
...
 	{
-		post() {
-			Meteor.call('readMessages', this.bodyParams.rid);
-
+		async post() {
+			const { rid, readThreads } = this.bodyParams;
+			check(rid, String);
+			check(readThreads, Match.Maybe(Boolean));
+			try {
+				await readMessages(rid, this.userId, readThreads === undefined ? true : readThreads);
```
pierre-lehnen-rc, line 87:
```suggestion
				await readMessages(rid, this.userId, readThreads ?? true);
```

pierre-lehnen-rc (approved):
Small suggestion but otherwise approved.

pierre-lehnen-rc (changes requested):
Actually, the readMessages method was removed so this is a breaking change.
Either keep the method as deprecated or change the PR to [BREAK]

LucianoPierdona:
> Actually, the readMessages method was removed so this is a breaking change. Either keep the method as deprecated or change the PR to [BREAK]

Updated PR title!

debdutdeb:
/cc @RocketChat/documentation-team 
respective document needs to be updated. 

changes:
- subscriptions.read by default reads ALL messages, i.e. including all the threads in a room [breaking]
- additional parameter - `readThreads` can be set to `false` to mimick the old behaviour.

Actually, I'm now considering defauling the api behaviour to **not** reading all threads. That way at least the endpoint behavior will be consistent.

apps/meteor/app/api/server/v1/subscriptions.ts
```diff
@@ -77,9 +79,15 @@ API.v1.addRoute(
...
-
+		async post() {
+			const { rid, readThreads } = this.bodyParams;
+			check(rid, String);
+			check(readThreads, Match.Maybe(Boolean));
+			try {
+				await readMessages(rid, this.userId, readThreads ?? true);
+			} catch (err: unknown) {
+				return API.v1.failure(err instanceof Error ? err.message : String(err));
+			}
```
ggazzo, line 90: no need for try catch here

ggazzo (changes requested):

ggazzo (changes requested):
if this is a break change, it will not be added in milestone 5.2.0, only on 6.0.0

ggazzo (changes requested):
could you show me where the `setMessagesAsReadByRoomIdAndUserId` is implemented?

btw I dont think is wise change the behavior of `setAsReadByRoomIdAndUserId` as far as I know we are using it in multiple places, and could introduce some problems

I would create something like `setAsReadByRoomIdAndUserIdIncludingThreads` instead 
@sampaiodiego could you help us here?

engelgabriel:
I agree with @ggazzo 

I don't think that changing the default behaviour of the API is wise, because it may break a lot of other places where it is not expected to clear all the unread counter....

We should create a new API for that.

apps/meteor/server/models/raw/Subscriptions.ts
```diff
@@ -113,9 +113,16 @@ export class SubscriptionsRaw extends BaseRaw<ISubscription> implements ISubscri
...
+				$unset: {
+					tunread: 1,
+					tunreadUser: 1,
+					tunreadGroup: 1,
+				} as const,
+			}),
 			$set: {
 				open: true,
-				alert,
+				alert: false,
```
sampaiodiego, line 125: always updating `alert` to `false` may let the room in a weird state, having unread threads but not showing the room as unread, like in the image below:
image

sampaiodiego (changes requested):

apps/meteor/server/models/raw/Subscriptions.ts
LucianoPierdona: updated!

ggazzo (dismissed):

PR #26756, opened 2022-08-31
Regression: Visitor being overwritten on call end

aleksandernsilva:

This PR adds a check to the `createRoom` method, responsible for creating VoIP rooms. It checks whether the visitor already exists before creating a new one, if one is found it uses it instead of overwriting existing visitors.

### DIFF
--- a/apps/meteor/client/providers/CallProvider/CallProvider.tsx
+++ b/apps/meteor/client/providers/CallProvider/CallProvider.tsx
@@ -12,6 +12,7 @@ import {
 	isVoipEventCallAbandoned,
 	UserState,
 	ICallDetails,
+	ILivechatVisitor,
 } from '@rocket.chat/core-typings';
 import { useMutableCallback } from '@rocket.chat/fuselage-hooks';
 import {
@@ -77,6 +78,7 @@ export const CallProvider: FC = ({ children }) => {
 	const visitorEndpoint = useEndpoint('POST', '/v1/livechat/visitor');
 	const voipEndpoint = useEndpoint('GET', '/v1/voip/room');
 	const voipCloseRoomEndpoint = useEndpoint('POST', '/v1/voip/room.close');
+	const getContactBy = useEndpoint('GET', '/v1/omnichannel/contact.search');
 	const setModal = useSetModal();
 	const t = useTranslation();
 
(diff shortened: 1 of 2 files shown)


apps/meteor/client/providers/CallProvider/CallProvider.tsx
```diff
@@ -161,20 +163,36 @@ export const CallProvider: FC = ({ children }) => {
...
 	}, []);
 
+	const findOrCreateVisitor = useCallback(
+		async (caller: ICallerInfo): Promise<ILivechatVisitor> => {
+			const phone = parseOutboundPhoneNumber(caller.callerId);
+
+			const { contact } = await getContactBy({ phone });
+
+			if (contact) {
+				return contact as unknown as ILivechatVisitor;
```
aleksandernsilva, line 173: The type returned by `useEndpoint` is not compatible with the same type on `core-typings`. If any of you know how to solve this, please let me know.

KevLehman (dismissed):
Approving since it works and we need it on QA :grimacing: 
Pls FE folks, do your magic here.

apps/meteor/client/providers/CallProvider/CallProvider.tsx
MartinSchoeler, line 173: This is something that is not exclusive to this PR, I seen this happen in other places, we need to investigate if this is an inconsistency issue between our types.

This is a work that will go over the scope of this PR so we can do this separately

MartinSchoeler (approved):

apps/meteor/client/providers/CallProvider/CallProvider.tsx
yash-rajpal: Yes, this is because of Types Serialization, when we send some data over HTTP requests, the objects and types are converted to strings, in this case `Date` type is converted to `string` type which creates this problem.

apps/meteor/client/providers/CallProvider/CallProvider.tsx
```diff
@@ -161,20 +163,36 @@ export const CallProvider: FC = ({ children }) => {
 		roomCoordinator.openRouteLink('v', { rid });
 	}, []);
 
+	const findOrCreateVisitor = useCallback(
+		async (caller: ICallerInfo): Promise<ILivechatVisitor> => {
```
yash-rajpal, line 167:
```suggestion
		async (caller: ICallerInfo): Promise<Serialized<ILivechatVisitor>> => {
```

This should solve all the problems and we can remove all type assertions.

yash-rajpal (approved):

MartinSchoeler (approved):

KevLehman (approved):

PR #26759, opened 2022-08-31
Regression: Select settings options not visible on Apps Setting panel

murtaza98:
## Issue(s)

image

### DIFF
--- a/apps/meteor/client/views/admin/apps/AppSetting.tsx
+++ b/apps/meteor/client/views/admin/apps/AppSetting.tsx
@@ -90,7 +90,7 @@ function AppSetting({ appSetting, onChange, value, ...props }: AppSettingProps):
 			value={value}
 			onChangeValue={onChange}
 			_id={id}
-			{...{ values: translatedValues }}
+			{...(translatedValues && { values: translatedValues })}
 			{...props}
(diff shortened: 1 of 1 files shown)


apps/meteor/client/views/admin/apps/AppSetting.tsx
```diff
@@ -90,7 +90,7 @@ function AppSetting({ appSetting, onChange, value, ...props }: AppSettingProps):
 			value={value}
 			onChangeValue={onChange}
 			_id={id}
-			{...{ translatedValues }}
+			{...{ values: translatedValues }}
```
gabriellsh, line 93: Not sure if your fix is valid but syntax wise, this is the same as `value={translatedValues}`. So no need to do a spread.

gabriellsh (changes requested):

apps/meteor/client/views/admin/apps/AppSetting.tsx
murtaza98: Yups, makes sense :+1: Modified the spread condition to only add that field if it's not null. Can you please check it again?

gabriellsh (approved):

PR #26770, opened 2022-08-31
Chore: Fix docker latest tag push

debdutdeb:

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -43,7 +43,7 @@ jobs:
         run: |
           LATEST_RELEASE="$(
             git -c 'versionsort.suffix=-' ls-remote -t --exit-code --refs --sort=-v:refname "https://github.com/$GITHUB_REPOSITORY" '*' |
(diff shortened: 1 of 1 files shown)


.github/workflows/build_and_test.yml
```diff
@@ -39,15 +38,6 @@ jobs:
           echo "RELEASE: ${RELEASE}"
           echo "::set-output name=release::${RELEASE}"
 
-      - id: latest
-        run: |
-          LATEST_RELEASE="$(
-            git -c 'versionsort.suffix=-' ls-remote -t --exit-code --refs --sort=-v:refname "https://github.com/$GITHUB_REPOSITORY" '*' |
-              sed -En '1!q;s/^[[:xdigit:]]+[[:space:]]+refs\/tags\/(.+)/\1/gp'
```
sampaiodiego, line 46: the fix could be changing only this line to
```
sed -En 's/^[[:xdigit:]]+[[:space:]]+refs\/tags\/([0-9]+.[0-9]+.[0-9]+)$/\1/gp' | head -n 1
```

I like this idea more as it won't rely on a rocket.chat service to get this information and it is also more fork friendly

sampaiodiego (changes requested):

sampaiodiego (approved):

PR #26776, opened 2022-09-01
[FIX] Livechat trigger messages covering all the website

tiagoevanp:


packages/livechat/src/widget.js
```diff
@@ -81,7 +81,7 @@ const updateWidgetStyle = (isOpened) => {
 	}
 
 	if (isOpened) {
-		widget.style.left = smallScreen ? '0' : 'auto';
+		widget.style.left = smallScreen && widget.dataset.state !== 'triggered' ? '0' : 'auto';
```
MartinSchoeler, line 84: Could you save this value to avoid repetition? `cons isTriggered = widget.dataset.state === 'triggered'`

packages/livechat/package.json
```diff
@@ -112,5 +112,8 @@
 		"updateFiles": [
 			"package.json"
 		]
+	},
+	"volta": {
```
MartinSchoeler, line 116: Are these changes required for this to work or just QoL Improvements? If they are we could do this on a different PR :)

MartinSchoeler (changes requested):

packages/livechat/webpack.config.js
```diff
@@ -177,6 +177,7 @@ module.exports = (env, argv) => [
 			clientLogLevel: 'trace',
 			overlay: true,
 			stats: 'normal',
+			writeToDisk: true,
```
MartinSchoeler, line 180: Same question about QoL here too

MartinSchoeler (changes requested):

packages/livechat/package.json
tiagoevanp: 😢 no...

packages/livechat/webpack.config.js
tiagoevanp: 😢 no...

aleksandernsilva (approved):

MartinSchoeler (approved):

PR #26780, opened 2022-09-01
[FIX] Reactions not breaking line.

gabriellsh:

Before:
image

After:
image


ggazzo (changes requested):
this shouldn't be an option, it should be implemented in the component, And that way it's horrible, try again

PR #26793, opened 2022-09-02
Chore: Add RocketChatDesktop function to open video calls when using Electron

jeanfbrito:


jeanfbrito:
I will need help to fix the TS type part because of the RocketChatDesktop on window. How I fix this in a proper manner?
image

pierre-lehnen-rc (approved):

dougfabris (approved):

PR #26795, opened 2022-09-02
[FIX] Full name not applied even if `useRealName` is true after reacting 

try-catch-stack:
Names are not stored along with the reactions when `setReaction` method is called. This is way usernames were displayed in the Reactions even when `useRealName` is set true. See attached video for more info.

https://user-images.githubusercontent.com/77742477/188212000-b81ffa32-5dca-4ea9-95ac-568f86a92536.mp4

https://user-images.githubusercontent.com/804994/187958874-470ec04e-a409-4b68-97f6-d2370b9e5329.mp4

## Steps to test or reproduce

* Make sure `useRealName` is turned on for the server.
* See reactions on a message. It'll show full names of the users.
* Now react to that message and open reactions again. It'll show usernames instead of Full Names.


apps/meteor/app/reactions/server/setReaction.js
```diff
@@ -3,16 +3,40 @@ import { TAPi18n } from 'meteor/rocketchat:tap-i18n';
...
+	message.reactions[reaction].usernames.splice(index, 1);
+	if (settings.get('UI_Use_Real_Name') && message.reactions[reaction].names) {
+		if (!message.reactions[reaction].names) {
+			message.reactions[reaction].names = message.reactions[reaction].usernames.map((username) =>
+				getNameByUsername(username, message, reaction),
+			);
+		} else {
+			message.reactions[reaction].names.splice(index, 1);
+		}
+	}
```
debdutdeb, line 39: You can avoid looping through twice with uncertain iterations by changing it from `reactions: {usernames: [], names: []}` to `reactions: [{username: '', name: ''}, ...]` and even sorting it in the backend.

Another problem with this is duplicate names. What if a user changes their name?

You're also in-place migrating? Should be a separate migration for this, not to be part of the code here.

The inner `if (true)` will never run.

apps/meteor/app/reactions/server/setReaction.js
```diff
@@ -84,6 +107,13 @@ async function setReaction(room, user, message, reaction, shouldReact) {
 			};
 		}
 		message.reactions[reaction].usernames.push(user.username);
+		if (settings.get('UI_Use_Real_Name') && !message.reactions[reaction].names) {
+			message.reactions[reaction].names = message.reactions[reaction].usernames.map((username) =>
+				getNameByUsername(username, message, reaction),
+			);
+		} else if (message.reactions[reaction].names) {
+			message.reactions[reaction].names.push(user.name);
+		}
```
debdutdeb, line 116: Conditionally pushing is a bad idea here. You're inserting incomplete data, further creating the need to modify it on runtime. Why is username a must field but names is conditional?

debdutdeb (changes requested):
You're introducing an unnecessary db call here, since frontend will check for the setting anyway. 

I'd suggest ajusting the frontend appropriately.

apps/meteor/app/reactions/server/setReaction.js
```diff
@@ -3,16 +3,40 @@ import { TAPi18n } from 'meteor/rocketchat:tap-i18n';
...
+	Users.findUsersByUsernames(message.reactions[reaction].usernames, {
+		fields: {
+			username: 1,
+			name: 1,
+		},
+	}).forEach((data) => {
+		names.set(data.username, data.name);
+	});
+	return names.get(username) || username;
+};
```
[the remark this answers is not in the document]
try-catch-stack: We only have the user object for the `user` who is reacting, right? The idea was to add names for users who have already reacted as well, using their usernames.

apps/meteor/app/reactions/server/setReaction.js
try-catch-stack: I thought `names` may not be an expected field when `UI_Use_Real_Name` is false but we can make it a must field as well if you say.

apps/meteor/app/reactions/server/setReaction.js
try-catch-stack: Yes it would make sense to use `[":reaction:"]: [{username: '', name: ''}, ...]` but I followed the existing type of [reaction](reaction) and the [`normalizeMessageForUser`](https://github.com/RocketChat/Rocket.Chat/blob/bf915717dbb1d0f4ba46df71f682fbb815ddcbe2/apps/meteor/app/utils/server/lib/normalizeMessagesForUser.ts#L74) function.

Is there a better way to achieve this?

apps/meteor/app/reactions/server/setReaction.js
try-catch-stack: Also names array is optional in the current [type](https://github.com/RocketChat/Rocket.Chat/blob/bf915717dbb1d0f4ba46df71f682fbb815ddcbe2/packages/core-typings/src/IMessage/IMessage.ts#L166).

apps/meteor/app/reactions/server/setReaction.js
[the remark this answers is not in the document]
debdutdeb: That should be a migration, not part of the running code. It's a one-off task. Or even we could just leave it empty and let the frontend fall back to username for old reactions. Why? Because the migration can be quite costly.

apps/meteor/app/reactions/server/setReaction.js
debdutdeb: > I thought names may not be an expected field when UI_Use_Real_Name is false
We should still add that or else we incur the cost of having to resolve those once the setting is updated.

apps/meteor/app/reactions/server/setReaction.js
debdutdeb: Update the type, and everywhere the structure is being used. 

I am also thinking we probably, instead of `[":reaction:"]: [{username: '', name: ''}, ...]`, use 
```js
[":reaction:"]: {
    [_id]: {
        username: '',
        name: ''
    },
    .
    .
    .
},
.
.
.
```

This way we won't have to iterate at all.

debdutdeb: I don't think we currently update the reaction usernames (consequently the names) once a user updates either. I'm thinking we could add a subscription for this, for each user, to keep track of the messages they've reacted to, and then update with a single query. Still would be expensive. 

@sampaiodiego wdyt?

debdutdeb: NB: Looked at how currently mentions are handled after those changes, and it directly uses the messages model.

dougfabris:

Closing this PR because it's stale now and it was already fixed in another one.

---

Issue #25992, opened 2022-06-23
Agent Activity Collection not being populated after 4.7.0


paulobernardoaf:
### Description:

After the 4.7.0 update, the collection `rocketchat_livechat_agent_activity` is not being populated. Don't know if that is the intended behavior from 4.7.0 and beyond but the only commit related to this collection that i've found is #25123.

### Steps to reproduce:

1. With a Rocket.Chat in version 4.7.0 or later
2. Configure Omnichannel
3. Become online as an agent
4. Verify the `rocketchat_livechat_agent_activity` collection

### Expected behavior:

Prior to 4.7.0 that collection was populated as usual when the agent's status changed.

### Actual behavior:

The collection is not populated.

### Server Setup Information:

- Version of Rocket.Chat Server: 4.8.1
- Operating System: Linux
- Deployment Method: Docker
- Number of Running Instances: 1
- DB Replicaset Oplog: true
- NodeJS Version: v14.18.3
- MongoDB Version: 4.2.21

### Client Setup Information

- Desktop App or Browser Version: Desktop App / Chrome
- Operating System: Ubuntu 21.10

KevLehman:
Should be fixed on recent releases.

PR #26797, opened 2022-09-03
[FIX] Agent Activity Collection not being populated after 4.7.0

mahyarkarimi:
[FIX] Agent Activity Collection not being populated after 4.7.0 #25992

Fixed method LivechatAgentActivity.createOrUpdate by adding upsert option.

## Issue(s)

Fixes https://github.com/RocketChat/Rocket.Chat/issues/25992


debdutdeb (approved):
@KevLehman this fixes the linked issue, yes, but, looking at `$setOnInsert` do you think it should also add `lastStartedAt` field in? Makes sense considering the collection is supposed to track agent activity and when an entry is created that should indicate the agent's last active moment? 

I am unaware of omnichannel code so this is just an observation. Please correct me if I'm wrong :)

murtaza98:
Thank you for the contribution!! I see here that the change has already been applied to the develop branch, so I'm going to close this PR

PR #26809, opened 2022-09-05
[FIX] Prevent user to access discussions from private rooms

LucianoPierdona:

This PR prevents users who aren't member of a private room to access it's discussions.


apps/meteor/app/lib/server/methods/getMessages.ts
```diff
@@ -21,6 +21,20 @@ Meteor.methods({
 			throw new Meteor.Error('error-not-allowed', 'Not allowed', { method: 'getSingleMessage' });
 		}
 
+		const prids = msgs.reduce<string[]>((prids, m) => {
+			const room: IRoom = Rooms.findById(m.rid);
```
sampaiodiego, line 25: please move this find to outside the loop/reduce. we should avoid multiple round trips to db as much as possible.

you already have the list of all room ids you need to query, the `rids` variable from line 19. you can use to perform a single find, returning only the `prid` field, something like:

```
const rooms = Rooms.findByIds(rids, { fields: { prid: 1 } }).fetch();
```

sampaiodiego (changes requested):

apps/meteor/app/lib/server/methods/getMessages.ts
LucianoPierdona: updated! Sorry for the miss on that from my side

debdutdeb:
I want to confirm exactly what kind of behavior are we looking for here? For discussions, anyone _should_ be able to access them despite it's parent room's class. What matters, is the users' membership. And this is currently being checked already. 

Secondly, the `getMessages` meteor method is nowhere called in our codebase. So, I don't know how this is changing anything.

PR #26811, opened 2022-09-05
Chore: get coverage from integration tests

weslley543:


souzaramon (approved):
lgtm

weslley543:
> @weslley543 I gave it a try excluding some API tests to see if the coverage changed but it was not changing apparently.. "Codecov Report" was also showing only `0.47%` coverage.. not sure this is accurate. how can we make sure it is collecting the correct metrics?
> 
> also `nyc` commands are very slow on my local env, for some reason.. it takes almost 5 minutes to "start", only after that time it runs the "mocha" command, for example.

Hello Diego ! 
I'm checking the points you raised. I believe the delay is due to the amount of files that need to be read by the library. I'll check more for sure and let you know. About this coverage I also found it strange and I'm checking too

PR #26824, opened 2022-09-06
[IMPROVE] Include `syncAvatars` on `ldap.syncNow`

LucianoPierdona:

This PR includes a new call for the method `syncAvatars` when `ldap.syncNow` is called


apps/meteor/ee/server/api/ldap.ts
```diff
@@ -24,7 +24,7 @@ API.v1.addRoute(
 				throw new Error('LDAP_disabled');
 			}
 
-			await LDAPEE.sync();
+			await Promise.all([LDAPEE.sync(), LDAPEE.syncAvatars()]);
```
pierre-lehnen-rc, line 27:
```suggestion
			await LDAPEE.sync();
			await LDAPEE.syncAvatars();
```

Running them at the same time would cause some race conditions and keep multiple connections active. It would also not sync the avatars of new users until the next full sync.

pierre-lehnen-rc (changes requested):

apps/meteor/ee/server/api/ldap.ts
LucianoPierdona: updated!

carlosrodrigues94 (approved):

pierre-lehnen-rc (approved):

PR #26829, opened 2022-09-07
Chore: Omnichannel endpoints to ts

KevLehman:
## Further comments

- Converts almost all Omnichannel endpoint definitions (CE/EE) to typescript with AJV as param validation tool.
- Removes "Meteor" dependency on endpoint where possible

JIRA be happy: OC-66


apps/meteor/app/api/server/api.js
```diff
@@ -425,7 +425,20 @@ export class APIClass extends Restivus {
 								typeof _options.validateParams === 'function' ? _options.validateParams : _options.validateParams[requestMethod];
 
 							if (validatorFunc && !validatorFunc(requestMethod === 'GET' ? this.queryParams : this.bodyParams)) {
-								throw new Meteor.Error('invalid-params', validatorFunc.errors?.map((error) => error.message).join('\n '));
+								console.log(this.queryParams, JSON.stringify(validatorFunc.errors, null, 2));
```
murtaza98, line 428:
```suggestion
```

apps/meteor/app/livechat/imports/server/rest/rooms.ts
```diff
@@ -1,54 +1,53 @@
...
 				return API.v1.unauthorized();
 			}
 
+			let parsedCf: { [key: string]: string } | undefined = undefined;
 			if (customFields) {
 				try {
-					const parsedCustomFields = JSON.parse(customFields);
-					check(parsedCustomFields, Object);
+					const parsedCustomFields = JSON.parse(customFields) as { [key: string]: string };
+					if (typeof parsedCustomFields !== 'object' && parsedCustomFields !== null && !Array.isArray(parsedCustomFields)) {
```
murtaza98, line 45: Didn't get the logic over here. Shouldn't it just be 
```
if (typeof parsedCustomFields !== 'object') {
	throw new Error('Invalid Custom fields')
}
```

apps/meteor/app/livechat/imports/server/rest/upload.ts
```diff
@@ -9,24 +9,14 @@ import { FileUpload } from '../../../../file-upload/server';
...
-	} catch (e) {
-		maxFileSize = await Settings.findOneById('FileUpload_MaxFileSize').packageValue;
-	}
-});
-
 API.v1.addRoute('livechat/upload/:rid', {
 	async post() {
 		if (!this.request.headers['x-visitor-token']) {
 			return API.v1.unauthorized();
 		}
```
murtaza98, line 17:
```suggestion
		if (typeof visitorToken !== 'string') {
			return API.v1.failure('Invalid visitor token');
		}
		const visitor = await LivechatVisitors.getVisitorByToken(visitorToken as string, {});
```

Enforcing the type check here to prevent any injection

apps/meteor/app/livechat/server/api/v1/pageVisited.ts
```diff
@@ -0,0 +1,35 @@
...
+				token: String,
+				rid: Match.Maybe(String),
+				pageInfo: Match.ObjectIncluding({
+					change: String,
+					title: String,
+					location: Match.ObjectIncluding({
+						href: String,
+					}),
+				}),
+			});
```
murtaza98, line 22: Won't be needing this ;)

apps/meteor/app/livechat/server/api/v1/message.ts
```diff
@@ -0,0 +1,293 @@
+import { Meteor } from 'meteor/meteor';
```
murtaza98, line 1: We're importing Meteor in this file just for its Meteor Wrapper, so can we change it to `Error` to remove this dependency altogether?

apps/meteor/ee/app/livechat-enterprise/server/api/agents.ts
```diff
@@ -45,28 +50,28 @@ API.v1.addRoute(
...
-	{ authRequired: true, permissionsRequired: ['view-livechat-manager'] },
+	{ authRequired: true, permissionsRequired: ['view-livechat-manager'], validateParams: isLivechatAnalyticsAgentsTotalServiceTimeProps },
 	{
 		async get() {
 			const { offset, count } = this.getPaginationItems();
-			let { start, end } = this.requestParams();
+			const { start, end } = this.requestParams();
 
 			check(start, String);
 			check(end, String);
```
murtaza98, line 60: Won't be needed this :)

murtaza98 (commented):
@KevLehman Left a few comments here. Please check

apps/meteor/app/livechat/imports/server/rest/rooms.ts
KevLehman: That extra ifs are because of this:
image

Both params will be parsed as they're valid JSON, but they'll make the endpoint fail since it's not what we're expecting.

apps/meteor/app/livechat/imports/server/rest/upload.ts
KevLehman: In which cases the type of the header value will be something different from string, since we're not parsing it? 👀

sampaiodiego (approved):
this adds a lot of `@ts-expect-error` .. any expectation on when to actually this them?

KevLehman:
They will go when we start migrating that big Livechat service into an actual service 🙈 no ETA for that yet, but happening soon

MartinSchoeler (approved):
Checked the places where the Client Files were modified and it seems to be working OK, the FE code is OK as well (mostly typing or fallback changes)

apps/meteor/app/livechat/server/api/v1/agent.ts
```diff
@@ -0,0 +1,60 @@
+import { isGETAgentNextToken } from '@rocket.chat/rest-typings';
+
+import { API } from '../../../../api/server';
+import { findRoom, findGuest, findAgent, findOpenRoom } from '../lib/livechat';
+import { Livechat } from '../../lib/Livechat';
+
+API.v1.addRoute('livechat/agent.info/:rid/:token', {
```
cauefcr, line 7: shouldn't this endpoint be validated too?

apps/meteor/ee/app/api-enterprise/server/canned-responses.ts
```diff
@@ -1,12 +1,12 @@
...
-import { Match, check } from 'meteor/check';
+import { isPOSTCannedResponsesProps, isDELETECannedResponsesProps, isCannedResponsesProps } from '@rocket.chat/rest-typings';
 
 import { API } from '../../../../app/api/server';
 import { findAllCannedResponses, findAllCannedResponsesFilter, findOneCannedResponse } from './lib/canned-responses';
 
 API.v1.addRoute(
 	'canned-responses.get',
-	{ authRequired: true },
+	{ authRequired: true, permissionsRequired: ['view-canned-responses'] },
```
cauefcr, line 9: same on this one, any reason for not verifying?

cauefcr (approved):

PR #26842, opened 2022-09-09
[FIX] LDAP "Sync Roles" option doesn't work for custom roles

LucianoPierdona:

Previously the sync of custom fields on LDAP wasn't working because the check was assuming that the `name` of the role is the same value from `_id`, which is the same for initial roles, but different for custom.

## Further comments

TC-38


matheusbsilva137 (approved):

apps/meteor/ee/server/lib/ldap/Manager.ts
```diff
@@ -223,11 +224,18 @@ export class LDAPEEManager extends LDAPManager {
 			const userFields = ensureArray<string>(fieldMap[ldapField]);
 
 			for await (const userField of userFields) {
-				const [roleId] = userField.split(/\.(.+)/);
-				allowedRoles.push(roleId);
+				const [roleName] = userField.split(/\.(.+)/);
+
+				const role = roles.find((role) => role.name === roleName);
```
pierre-lehnen-rc, line 229:
```suggestion
				const [roleIrOrName] = userField.split(/\.(.+)/);

				const role = roles.find((role) => role._id === roleIrOrName) ?? roles.find((role) => role.name === roleIrOrName);
```

Accept both ids and names in the LDAP setting, prioritizing IDs.
This way it'll ensure that any environment currently configuring with IDs will not need to be reconfigured.

pierre-lehnen-rc (changes requested):

apps/meteor/ee/server/lib/ldap/Manager.ts
LucianoPierdona: updated!

pierre-lehnen-rc (approved):

PR #26858, opened 2022-09-13
[FIX] `channels.convertToTeam` endpoint doesn't work when only the `channelName` param is provided

matheusbsilva137:

- Fix the `channels.convertToTeam` endpoint not working when the `channelName` is sent as parameter.

## Steps to test or reproduce

Example request (using _**cURL**_):
```
curl 'http://localhost:3000/api/v1/channels.convertToTeam' \
  -H 'X-Auth-Token: GNNn5trk37eeK2sZvamOHRy1PMRlz_rfMUCwCE6UBlO' \
  -H 'X-User-Id: HZcEBodsm3gFxg2Jt' \
  --data-raw '{"channelName":"channel-example"}'
```


LucianoPierdona (approved):

LucianoPierdona (dismissed):

packages/rest-typings/src/v1/channels/ChannelsConvertToTeamProps.ts
```diff
@@ -22,6 +22,15 @@ const channelsConvertToTeamPropsSchema = {
...
 		},
+		{
+			type: 'object',
+			properties: {
+				channelId: { type: 'string' },
+				channelName: { type: 'string' },
+			},
+			required: ['channelId', 'channelName'],
+			additionalProperties: false,
+		},
```
debdutdeb, line 33: this is wrong. this will allow passing BOTH at the same time.

apps/meteor/app/api/server/v1/channels.ts
```diff
@@ -471,8 +465,8 @@ API.v1.addRoute(
 
 			const room = findChannelByIdOrName({
 				params: {
-					roomId: channelId,
-					roomName: channelName,
+					...(channelId && { roomId: channelId }),
+					...(channelName && { roomName: channelName }),
```
debdutdeb, line 469:
```suggestion
					roomId: channelId,
					roomName: channelName,
```
The function will have to check each field any way. No need to do it again here.

debdutdeb (changes requested):
I see the problem. All you need to do is instead of checking for membership, check for value truthness, i.e. `params.roomId ? findbyid : findbyname`

packages/rest-typings/src/v1/channels/ChannelsConvertToTeamProps.ts
debdutdeb: to add: this doesn't make sense because 
1. passing the id and the name of the same room is redundant
2. passing the id of one room and name of another .. endpoint at its current stage will just process the id, so the name is unnecessary. and we definitely don't want to process each item individually, that's just bad design.

apps/meteor/app/api/server/v1/channels.ts
matheusbsilva137: This endpoint is not working when only the `channelName` param is provided, so I added this change to fix it. But now I see a better approach would be to change the validation in `findChannelByIdOrName`. Tks!

packages/rest-typings/src/v1/channels/ChannelsConvertToTeamProps.ts
matheusbsilva137: This has been requested by the mobile team, but I've started a conversation with the architecture team and we decided to remove this change. Thanks for pointing this!

apps/meteor/tests/end-to-end/api/02-channels.js
```diff
@@ -1938,6 +1938,62 @@ describe('[Channels]', function () {
 			});
 		});
 
+		it(`should successfully convert a channel to a team when the channel's name is sent as parameter`, (done) => {
+			updatePermission('create-team', ['admin']).then(() => {
+				updatePermission('edit-room', ['admin']).then(() => {
```
sampaiodiego, line 1943: there is no need to update permissions.. looking the previous test, this is the current state already.

apps/meteor/tests/end-to-end/api/02-channels.js
```diff
@@ -1938,6 +1938,62 @@ describe('[Channels]', function () {
...
+					request
+						.post(api('teams.convertToChannel'))
+						.set(credentials)
+						.send({ teamName: this.newChannel.name })
+						.expect(200)
+						.expect((res) => {
+							expect(res.body).to.have.a.property('success', true);
+						})
+						.then(() => {
+							request
```
sampaiodiego, line 1953: this Promise nesting and the previous one with the permissions update is wrong in so many levels 🙈 I see this has been the "standard" on this particular file, but we shouldn't propagate bad code even if it is the current "standard"..

this is conceptually wrong since the `Promise` API was designed to actually mitigate callback hell.. so a correct usage of the `Promise` API should look something like:

```js
updatePermission()
    .then(() => updatePermission())
    .then(() => request
        .post(api('teams.convertToChannel'))
        .set(credentials)
        // ....
    ).then(() => request
        .post(api('teams.convertToTeam'))
        .set(credentials)
        // ....
    )
```

as you can see, everything is done in a **single** level of nesting.

**BUT**, you should also not do this, since `async/await` is available for tests and it is the recommended way to write sequential requests, so please refactor at least the new tests to use `async/await`.

sampaiodiego (changes requested):

apps/meteor/app/api/server/v1/channels.ts
```diff
@@ -471,8 +471,7 @@ API.v1.addRoute(
 
 			const room = findChannelByIdOrName({
 				params: {
-					roomId: channelId,
-					roomName: channelName,
+					...(channelId ? { roomId: channelId } : { roomName: channelName }),
```
debdutdeb, line 474: Do we have to do it like this?

**In this particular context**, `roomId in params` is the same as `params.roomId ?` - so no matter what you do in the calling function, the called **still** has to check for both values any way.

This conditional and destructuring just seems useless at that point. And IMO is bad. We should keep the code simple. And inspire others (people who are gonna read your code) to do so too.

on the called, you can simply change the param type from the union to a simple all-partial and chain the truthness checks `roomId ? .. : roomName ? ... : undefined`

it's late for me, hopefully i was able to explain myself.

debdutdeb (changes requested):

apps/meteor/app/api/server/v1/channels.ts
matheusbsilva137: Hey Debdut, I'll first explain what was the issue on the way we were _calling_ this function (that is, with `params: { roomId: channelId, roomName: channelName }`): this would cause one of the fields in the `params` object to be `undefined`. So checking `'roomId' in params` in `findChannelByIdOrName` would _always_ return `true` (since this field is in the object, even though its value is `undefined`).

Still, I see we could change the params type to `params: { roomId?: string; roomName?: string }` and check for `params.roomId` instead of `'roomId' in params` (is that what you mean?). If we do so, I think we should also bring back the error check we had before (so as to throw an error when both params are undefined).

debdutdeb: > throw an error when both params are undefined

Already done by the route function. So, you can safely ignore that part.

debdutdeb: https://github.com/RocketChat/Rocket.Chat/blob/aaebe90bb2a1bbd72c0c2b0c33bd38849b9058ef/apps/meteor/app/api/server/v1/channels.ts#L464-L466

sampaiodiego (approved):

sampaiodiego (approved):

PR #26859, opened 2022-09-13
[NEW] Matrix federation events coverage expansion (adding support for 3 extra events)

MarcosSpessatto:

FED-17
FED-25
FED-27
FED-28
FED-30
The goal of this PR is to add support for more events on Matrix Federation feature. The new supported events are:

- Mentions
- User avatar sync
- Quotes

Also, the PR adds a new matrix federation object into the statistics collector.


alansikora (dismissed):

alansikora (dismissed):

pierre-lehnen-rc (changes requested):

apps/meteor/server/startup/migrations/v282.ts
```diff
@@ -0,0 +1,19 @@
...
+addMigration({
+	version: 282,
+	async up() {
+		await Statistics.updateMany(
+			{},
+			{
+				$set: {
+					matrixFederation: await getMatrixFederationStatistics(),
+				},
+				$unset: { matrixBridgeEnabled: 1 },
```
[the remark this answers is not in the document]
MarcosSpessatto: Got it, I didn't know about that. Thanks for letting me know.

pierre-lehnen-rc (dismissed):

gabriellsh (approved):

apps/meteor/server/startup/migrations/v282.ts
```diff
@@ -0,0 +1,18 @@
...
+		await Statistics.updateMany(
+			{},
+			{
+				$set: {
+					matrixFederation: await getMatrixFederationStatistics(),
+				},
+			},
+		);
+	},
+});
```
ggazzo, line 18: We strongly recommend avoid migrations, if something happens (during the release candidates or any other deployment) we cannot only revert to the old version.

Sometimes its not possible to continue without migrations, but I'm afraid this is not the case. 
Looks 100% useless to me

@sampaiodiego fyi

ggazzo (changes requested):

apps/meteor/server/startup/migrations/v282.ts
MarcosSpessatto: Makes sense, honestly, it was to replace one value with another, but since I've removed the `$unset` in the previous change requested, it became completely useless now. 👍

PR #26863, opened 2022-09-13
[IMPROVE] Changed dial pad appearance to match original design

aleksandernsilva:
OC-61

Before:
Screen Shot 2022-09-13 at 15 00 34

After:
Screen Shot 2022-09-13 at 14 59 23


aleksandernsilva:
OC-61

apps/meteor/ee/client/voip/modal/DialPad/PadButton.tsx
```diff
@@ -1,9 +1,26 @@
...
+	height: 64px;
+	padding: 8px;
+	margin: 8px;
+	border: none;
+	display: flex;
+	flex-direction: column;
+	align-items: center;
+
+	&:hover {
+		background-color: #e4e7ea;
```
tiagoevanp, line 20: I don't known if this is the proper way to handle with color... I say that regarding on how this icon will stay fixed with this color and not tied with our design system... I know that we have a custom situation here... but in the future, it will be important to follow a pattern-name or stay the same from eternity?

apps/meteor/ee/client/voip/modal/DialPad/PadButton.tsx
```diff
@@ -20,20 +37,17 @@ const PadButton = ({
...
+			className={padButtonStyle}
 			onClick={onClick}
 			onMouseDown={onMouseDown}
 			onMouseUp={onMouseUp}
 			onTouchStart={onTouchStart}
 			onTouchEnd={onTouchEnd}
 		>
-			<Box fontSize='h2'>{firstDigit}</Box>
-			<Box fontSize='c1' color='info'>
+			<Box is='p' fontSize='h2' lineHeight='32px'>
```
tiagoevanp, line 47: There is a real necessity of is='p'? We commonly use Box (as div) with a children text inside of it without this concern...

tiagoevanp (changes requested):

apps/meteor/ee/client/voip/modal/DialPad/PadButton.tsx
aleksandernsilva: @tiagoevanp When possible I like to use more meaningful tags. In this case since we are dealing with text I choose to use p. I don't think there's a problem with either element for this specific case.

apps/meteor/ee/client/voip/modal/DialPad/PadButton.tsx
aleksandernsilva: Makes sense. I'll check if we have a variable for this color.

apps/meteor/ee/client/voip/modal/DialPad/PadButton.tsx
tiagoevanp: I agree with you on improve the meaningful using tags, but in this specific case, you're using inside of a button component, right? So, what about use `span` tags instead `p` tag? I think is more semantically correct once `span` is a [phrasing content](https://developer.mozilla.org/en-US/docs/Web/Guide/HTML/Content_categories#phrasing_content)

tiagoevanp (changes requested):

tiagoevanp (approved):

PR #26866, opened 2022-09-14
[NEW] REST API endpoint `/v1/rooms.delete`

felipe-rod123:


packages/rest-typings/src/v1/teams/index.ts
```diff
@@ -182,4 +182,8 @@ export type TeamsEndpoints = {
 			room: IRoom;
 		};
 	};
+
+	'/v1/teams.eraseRoom': {
```
ggazzo, line 186: move to room

ggazzo: /v1/room.erase

ggazzo (changes requested):

ggazzo (changes requested):
pleas check comments bellow

ggazzo (approved):

ggazzo:
> any special reason to not follow the same pattern as `POST users.delete` ?

since we have no ~good~pattern defined I decided to follow the world convention (RestFul) and apply the http methods usage properly, maybe we could keep both (one because our ancient patterns and the right one) what do you think ?

sampaiodiego:
my suggestion is to use the same pattern we use on other endpoints, so for this one it would be `POST rooms.delete { roomId: 'ID-TO-DELETE' }` .. I know it is not the best pattern, but it is the one we use, so it is somewhat predictable if you're already familiar with Rocket.Chat API.

sampaiodiego (dismissed):

sampaiodiego (approved):

PR #26875, opened 2022-09-14
feat: Allow Incoming Webhooks to override destination channel

LucianoPierdona:

This PR adds a new setting for the incoming webhooks integration, where you can specify if the integration will accept to overwrite the channel defined in the configurations.

- Added field `overrideDestinationChannelEnabled` to incoming webhooks integration with default value being `true`, to not break old integrations.

### With setting enabled
1. Setup an incoming webhook integration in RC with `overrideDestinationChannelEnabled` being set as `true`
2. Add the sending user to other channels/rooms apart from the room specified in the "Post to Channel" field in the incoming integration settings.

Request example:
```
{
        "text": "Example test",
	"channel": ["general-2"]
}
```

3. Expected result: should send the message to the channel specified in the body request.

### With setting disabled
1. Setup an incoming webhook integration in RC with `overrideDestinationChannelEnabled` being set as `false`
2. Add the sending user to other channels/rooms apart from the room specified in the "Post to Channel" field in the incoming integration settings.

Request example:
```
{
        "text": "Example test",
	"channel": ["general-2"]
}
```

3. Expected result: should throw an error saying that the webhook integration doesn't support this action.

## Further comments

[TC-47]

[TC-47]: https://rocketchat.atlassian.net/browse/TC-47?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/meteor/tests/end-to-end/api/07-incoming-integrations.js
```diff
@@ -165,6 +175,30 @@ describe('[Incoming Integrations]', function () {
...
+				.get(api('channels.messages'))
+				.set(credentials)
+				.query({
+					roomId: channel._id,
+				})
+				.expect('Content-Type', 'application/json')
+				.expect(200)
+				.expect((res) => {
+					expect(res.body).to.have.property('success', true);
+					expect(res.body).to.have.property('messages', []);
```
matheusbsilva137, line 198:
```suggestion
					expect(res.body).to.have.property('messages').and.to.be.an('array');
					expect(res.body.messages).to.eql([]);
```
You need deep equality here, this should work

matheusbsilva137 (changes requested):

apps/meteor/tests/end-to-end/api/07-incoming-integrations.js
LucianoPierdona: updated!

matheusbsilva137 (changes requested):

apps/meteor/app/integrations/server/api/api.js
```diff
@@ -270,10 +270,22 @@ function executeIntegrationRest() {
...
+
+		this.bodyParams.channel = channels
+			.map((channel) => {
+				if (defaultValues.channel.find((c) => c === channel)) {
+					return channel;
+				}
+
+				return null;
+			})
+			.filter(Boolean);
```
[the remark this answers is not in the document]
LucianoPierdona: Updated!

apps/meteor/app/integrations/server/api/api.js
```diff
@@ -270,10 +270,11 @@ function executeIntegrationRest() {
 	this.bodyParams.bot = { i: this.integration._id };
 
 	try {
-		const message = processWebhookMessage(this.bodyParams, this.user, defaultValues);
-		if (_.isEmpty(message)) {
-			return API.v1.failure('unknown-error');
-		}
```
matheusbsilva137, line 276: Couldn't we keep this check anyways?

matheusbsilva137 (changes requested):

apps/meteor/tests/end-to-end/api/07-incoming-integrations.js
```diff
@@ -165,6 +175,31 @@ describe('[Incoming Integrations]', function () {
...
+				.set(credentials)
+				.query({
+					roomId: channel._id,
+				})
+				.expect('Content-Type', 'application/json')
+				.expect(200)
+				.expect((res) => {
+					expect(res.body).to.have.property('success', true);
+					expect(res.body).to.have.property('messages').and.to.be.an('array');
+					expect(res.body.messages).to.eql([]);
```
matheusbsilva137, line 199: I'd suggest adding a new test to ensure that the message sent here has been "redirected" to the correct channel (since this is the behavior we chose to follow).

matheusbsilva137 (changes requested):

apps/meteor/app/integrations/server/api/api.js
matheusbsilva137: you mean the channel in `bodyParams`? I see that `_.isEmpty()` would still return `true` if there's no default channel.

LucianoPierdona: re-added condition

apps/meteor/tests/end-to-end/api/07-incoming-integrations.js
LucianoPierdona: Added test!

matheusbsilva137 (approved):

sampaiodiego (changes requested):
not sure if ignoring the payload is the best approach here.

what happens to third party code that currently uses either `channel` or `roomId` and update their rocket.chat, their params will be silently ignored, causing all kinds of issues without no previous warning.

IMO we should first display and return a warning saying the providing those params is now deprecated, and in the next major version, drop them throwing an error in case they're provided.

I'm open to discuss about this though =)

LucianoPierdona:
> not sure if ignoring the payload is the best approach here.
> 
> what happens to third party code that currently uses either `channel` or `roomId` and update their rocket.chat, their params will be silently ignored, causing all kinds of issues without no previous warning.
> 
> IMO we should first display and return a warning saying the providing those params is now deprecated, and in the next major version, drop them throwing an error in case they're provided.
> 
> I'm open to discuss about this though =)

@sampaiodiego I've done one update, is this what you expected? or should I let the user go through with the flow and just return the usual response but with the additional warning?

rodrigok:
Not sure this is really an issue. Will double check

LucianoPierdona:
Removed from `6.0.0` for internal discussions

LucianoPierdona:
Hi,

After discussions with @rodrigok and @milton-rucks, the scope of this PR was changed to allow the user to specify the channel he wants to send the message to if the `overrideDestinationChannelEnabled` setting is enabled. Otherwise, we should send an error message saying that the webhook doesn't support this update.

The default value for `overrideDestinationChannelEnabled` is `false` for newer integrations. I'm creating a migration for the old ones to set the value as `true`.

apps/meteor/client/views/admin/integrations/IncomingWebhookForm.js
```diff
@@ -22,6 +22,7 @@ export default function IncomingWebhookForm({ formValues, formHandlers, extraDat
 		handleAvatar,
 		handleEmoji,
 		handleScriptEnabled,
+		handleOverrideDestinationEnabled,
```
matheusbsilva137, line 25:
```suggestion
		handleOverrideDestinationChannelEnabled,
```
There's a typo here, we should use the exact same name you defined the prop, otherwise this handle function won't be found. This was causing the toggle button not to work, since this handle function is undefined

apps/meteor/client/views/admin/integrations/IncomingWebhookForm.js
```diff
@@ -155,6 +156,17 @@ export default function IncomingWebhookForm({ formValues, formHandlers, extraDat
...
+					{useMemo(
+						() => (
+							<Field>
+								<Field.Label display='flex' justifyContent='space-between' w='full'>
+									{t('Override_Destination_Channel')}
+									<ToggleSwitch checked={overrideDestinationChannelEnabled} onChange={handleOverrideDestinationEnabled} />
+								</Field.Label>
+							</Field>
+						),
+						[t, overrideDestinationChannelEnabled, handleOverrideDestinationEnabled],
```
matheusbsilva137, line 168:
```suggestion
						[t, overrideDestinationChannelEnabled, handleOverrideDestinationChannelEnabled],
```
Same as stated above

matheusbsilva137 (changes requested):

matheusbsilva137 (dismissed):

gabriellsh (dismissed):

gabriellsh (dismissed):
frontend LGTM

matheusbsilva137 (approved):

PR #26880, opened 2022-09-15
[FIX] `MongoInvalidArgumentError` on overwriting existing setting

debdutdeb:


KevLehman (dismissed):

sampaiodiego:
I've found another case when providing a value via env var would not be applied if the setting props also changed.. I'm working on unit tests to cover both cases

sampaiodiego (approved):

PR #26895, opened 2022-09-19
[FIX] `Not allowed` error when attempting to change status with custom message disabled

### LINKED ISSUE #26820: Status is not updating
When the status "Online" is clicked, the status does not update. The same is true with other statuses.

Video reference: https://watch.screencastify.com/v/Oz6Wzlgdf3fFs3agOYUp

debdutdeb:
## Issue(s)

Fixes #26820


debdutdeb:
TODO: add test

PR #26909, opened 2022-09-20
[FIX] Adjusted livechat fallbacks to take null values into account

aleksandernsilva:
OC-209

This pull request fixes:
-  Fallbacks from a few livechat methods to take into account `null` values and fallback to the correct default value instead of breaking the widget. 
- `defaultAgent` not being cleared from cache when a new department is set
- default guest information not being displayed


tiagoevanp (approved):

ggazzo (dismissed):

tassoevan (approved):
Nullish coalescing (e.g. `a ?? b`) could be better, but `||` is fine as we're expecting objects. Just keep it in mind when dealing with falsy primitives.

MartinSchoeler (approved):

MartinSchoeler (approved):

tassoevan (approved):

PR #26912, opened 2022-09-21
Chore: fix regressions for omnichannel due room refactor

ggazzo:
This PR fixes the following:
- [x] Opening a VoIP room causes a crash (invalid room type)
- [x] Closing a room causes an error (room not found)
- [x] Contextual bar not opening automatically when the room is opened (cannot read tabBar from null)


MartinSchoeler (changes requested):
Screen Shot 2022-09-22 at 1 49 13 PM

If you try to open a livechat that has been closed via `current-chats` the channel will appear empty

MartinSchoeler (approved):

MartinSchoeler:
> Screen Shot 2022-09-22 at 1 49 13 PM
> 
> If you try to open a livechat that has been closed via `current-chats` the channel will appear empty

Okay, since the changes are already on develop, this issue is not locked to the scope of this PR, we can work out a fix in another PR

PR #26949, opened 2022-09-26
Regression: return home after close chat

MartinSchoeler:
## Further comments

OC-259


aleksandernsilva (approved):

ggazzo (changes requested):
Could you explain to me why sending it home is something useful?  there is no information that helps the agent there.  if he's in current chats it's even worse.  I would very kindly ask you to reconsider

I actually did it on purpose and found it much better.  I know it changes behavior, no that doesn't mean it got worse

Feel free to dismiss my review but I would like to see you considering my suggestion

cauefcr (approved):
Reviewed and tested, it works perfectly here.

aleksandernsilva (approved):

ggazzo (changes requested):
what happens if the room is closed by other client or other user tab?
since we're so focused on consistency and behavior, I suppose it's critical if someone else (or yourself in another tab) closes and doesn't get an update

MartinSchoeler:
Closing this, giving preference to https://github.com/RocketChat/Rocket.Chat/pull/26988

PR #26974, opened 2022-09-28
Chore: Tests for EE dashboards

KevLehman:
## Further comments

Jira: OC-65


apps/meteor/app/livechat/server/hooks/processRoomAbandonment.js
```diff
@@ -64,7 +64,11 @@ callbacks.add(
 		if (!closedByAgent || !wasTheLastMessageSentByAgent) {
 			return;
 		}
-		const agentLastMessage = Messages.findAgentLastMessageByVisitorLastMessageTs(room._id, room.v.lastMessageTs);
+		// There's a possibility of a visitor not sending any message b4 agent, and this will cause
```
murtaza98, line 67: Wouldn't the false condition here change the logic? From the variable name, it seems to be a "message from an agent", however, if this condition is false, then we're now picking the room's last message which isn't guaranteed to be a message from visitor

apps/meteor/tests/data/livechat/business-hours.ts
```diff
@@ -0,0 +1,11 @@
+import { ILivechatBusinessHour } from '@rocket.chat/core-typings';
+import { credentials, methodCall, request } from '../api-data';
+
+export const saveBusinessHour = async (businessHour: ILivechatBusinessHour) => {
+	await request
+		.post(methodCall('livechat:saveBusinessHour'))
+		.set(credentials)
+		.send({ message: JSON.stringify({ params: [businessHour], msg: 'method', method: 'livechat:saveBusinessHour', id: '101' }) })
+		.expect(200);
```
murtaza98, line 9: One thing I found recently was that meteor methods return 200 status code even if the method throws an error :sweat: So the only way to assure that the method was successful would be to parse the response body and check if the body has an error field

murtaza98, line 9:
```suggestion
		.expect(200);
		.expect((res: Response) => {
			expect(res.body).to.have.property('success', true);
			const parsedBody = JSON.parse(res.body.message);
			expect(parsedBody).to.have.property('error');
			expect(parsedBody.error).to.have.property('error', 'error-not-allowed');
		});
```

Here's a sample snippet from my other PR where I was parsing the error message from a meteor method. Hope it helps

apps/meteor/app/livechat/server/hooks/processRoomAbandonment.js
KevLehman: You brought an interesting point. And I noticed this actually doesnt work (not the line, but the whole logic for selecting the last message). It _assumes_ the first message on a room will be _always_ a visitor message. Which makes sense because that's what livechat does (routing chat and do events only on first message)

However, if the message _after_ the visitor one is an app (like, what happens with CRMs) the query will return the message from the app, not from the agent.

So, I'll take some extra time to fix this behavior, or adapt the test to what currently happens and create a task for what may happen 👀

cauefcr (approved):
Good tests, ran fine on my machine, changes requested were taken into account, approving.

murtaza98 (approved):

ggazzo (approved):

Issue #26677, opened 2022-08-25
Problem with room avatars


Lumiond:
### Description:

Room avatar disappears and after this I can't change it (avatar field always empty, see screencaps).
![image](https://user-images.githubusercontent.com/51132430/186594453-11e9b8d7-68a4-4ce4-b765-5007d22f57e0.png)

### Steps to reproduce:

1. Select room and click on "room information" button
2. Click edit
3. Click "Upload" and select avatar. Click save.
4. Reload chat without cache (option from app)
5. Repeat 1-3.

### Expected behavior:

New avatar is set

### Actual behavior:

Avatar disappears. After this I can't set new avatar.

### Server Setup Information:

- Version of Rocket.Chat Server: 5.0.3
- Operating System: ubuntu lts 20.04
- Deployment Method: docker
- Number of Running Instances: 1
- NodeJS Version: v14.19.3
- MongoDB Version: 5.0.6 / wiredTiger (oplog Enabled)

### Client Setup Information

- Desktop App
- Operating System: windows 10 pro

### Additional context

Also I have noticed that if I select "default avatar" and press save I get this error: "TypeError: Cannot convert undefined or null to object".
**We use s3 compatible storage for files.**

Issue #22903, opened 2021-08-11
Uploaded channel avatar photo doesn't appear


csimpi:
### Description:

When I upload an avatar for a channel, the uploaded photo appears on the upload section but nowhere else.

### Steps to reproduce:

1. Go to Room Information page
2. Click on Edit button
3. Upload a new avatar
4. Save

### Expected behavior:

The new avatar should appear, the default avatar should be replaced

### Actual behavior:

The default avatar stays everywhere.

### Server Setup Information:

- Version of Rocket.Chat Server: 3.17.0
- Operating System: Linux 
- Deployment Method:  tar
- Number of Running Instances: 1
- DB Replicaset Oplog: 
- NodeJS Version:  12.22.1
- MongoDB Version: 4.4.3 / wiredTiger  (oplog Enabled)

### Client Setup Information

- Desktop App or Browser Version: chrome newest
- Operating System: Linux Debian 20.4

ankar84:
Related to https://github.com/RocketChat/Rocket.Chat/issues/22307
Could you check if avatar set in mobile client when you set it in web?

csimpi:
> Related to #22307
> Could you check if avatar set in mobile client when you set it in web?

Yes, it sets in right away on mobile, when I set it on web. On web, nothing happens.

ankar84:
> Yes, it sets in right away on mobile, when I set it on web. On web, nothing happens.

I have exactly the same experience.
So, thinking logically - avatar is set properly  if we see it in mobile client. 
But the real problem is in web client - it's now updating avatar right after it was set. Or not show properly.

debdutdeb:
Thanks Peter ! Thanks Anton 👋🏼 

I'll try and get this fixed ASAP. Should be an easy one.

ankar84:
> Thanks Peter ! Thanks Anton 👋🏼
> 
> I'll try and get this fixed (pass to the product team) ASAP. Should be an easy one.

Still not fixed in 3.18.1

pierreozoux:
I have a slightly similar, but maybe a bit different.

It doesn't reproduce consistently, but If I try 2 times, usually, I can reproduce.
We use an s3 compatible api as an object store (minio) and the logs look good at the upload:
Server:
```
[...]

Issue #15470, opened 2019-09-30
Avatar upload to S3 bucket only works when initial avatar is set


splattner:
### Description:

RocketChat is configured to use AmazonS3 (but not with AWS S3 Service) as FileUpload storage type. Setting a new user avatar (upload or gravatar) works when the current avatar is the initial avatar (1st letter of name). When the avatar is set and a user uploads a new avatar, the new avatar is not stored in the S3 bucket and a broken image is shown. MongoDB entry is fine but object does not exist in S3 bucket. Reverting to initial avatar, does remove the MongoDB entry.

### Steps to reproduce:

1. Go to Account / Profile
2. Set a profile picture (upload one or use gravatar) and save the changes
3. Profile picture shows a broken Image -> NOK
4. Change to initial avatar and save changes
5. Profile Picture shows initial avatar
6. Set a Profile Picture (upload one or use gravatar) and save the changes
7. Profile picture shows the new image corrrectly. -> OK

### Expected behaviour:

the new user avatar should be store correctly no matter what previous avatar was used

### Actual behavior:

MongoDB entry with initial avatar set:

```
db.rocketchat_avatars.find({name: "splattner"})
# No Result
```

MongoDB Entry when changing from initial avatar to a new profile picture. S3 Object exists

```
db.rocketchat_avatars.find({name: "splattner"})
{ "_id" : "YTb345cKGtLrBcGQZ", "userId" : "oEEnfd378rh8H83is", "type" : "image/jpeg", "size" : 5273, "AmazonS3" : { "path" : "4a4C7dfyuCXH4yygw/avatars/oEEnfd378rh8H83is" }, "store" : "AmazonS3:Avatars", "_updatedAt" : ISODate("2019-09-30T12:03:08.529Z"), "instanceId" : "jgvePrpWxRKB3nWvH", "name" : "splattner", "complete" : true, "etag" : "PiNKXiEnMweZzCSZm", "path" : "/ufs/AmazonS3:Avatars/YTb345cKGtLrBcGQZ/undefined", "progress" : 1, "token" : "f8Bbdb292a", "uploadedAt" : ISODate("2019-09-30T12:03:08.642Z"), "uploading" : false, "url" : "https://chat.****/ufs/AmazonS3:Avatars/YTb345cKGtLrBcGQZ/undefined" }
```

MongoDB after changing profile picture again. S3 Object does not exist
```
db.rocketchat_avatars.find({name: "splattner"})
[...]

PR #26985, opened 2022-09-30
[FIX] Room avatar deleted after updating

hrenard:

Delete current room's avatar before inserting the new one and not after. On store like S3, if we delete after the insertion then the object isn't available anymore on the storage.

## Issue(s)

#22903
#15470
#26677

## Steps to test or reproduce

1. Enable S3 storage.
2. Set a room's avatar and save.
3. Set a new avatar and save.

The room has an empty avatar because S3 returns a 404.

## Further comments

This is consistent with the way it's done in `setUserAvatar`:
https://github.com/RocketChat/Rocket.Chat/blob/0e6bc45d74d94f93810fce8697dfc4aec36f741d/apps/meteor/app/lib/server/functions/setUserAvatar.ts#L100-L119


gabriellsh:
Hi @hrenard ! Thanks for contributing with your solution!

I think you changed the logic a little too much, though. Just deleting before upload should suffice.
I've opened a new PR with the ideal solution here #27060. Please, feel free to comment if you have any more suggestions. (Added you as co-author).

PR #27020, opened 2022-10-06
[FIX] Uncaught error by listening stream-notify-room without subscription

tiagoevanp:

fix this:
image


aleksandernsilva (approved):

KevLehman:
@sampaiodiego you're more familiar with this code, can you help us reviewing? 🙏

apps/meteor/server/modules/notifications/notifications.module.ts
```diff
@@ -173,9 +181,9 @@ export class NotificationsModule {
 			if (!this.userId) {
 				return false;
 			}
+			const canAccess = await Authorization.canAccessRoom(room, { _id: this.userId }, extraData);
```
sampaiodiego, line 184: since you have only `rid`, you can use `Authorization.canAccessRoomId` , so you don't  need to add the query added on line 164.

sampaiodiego (changes requested):

sampaiodiego:
can you also please create a test for this fix?

tassoevan (approved):

PR #27037, opened 2022-10-10
[FIX] Ajusted form validation to disallow saving existing emails

aleksandernsilva:

This pull request revisits the user info form, fixing a bug that allowed the form to be saved with an already existing e-mail. Taking advantage of this opportunity this PR also refactors the file to typescript and adds new libs like `useForm` and `useQuery`.


cauefcr (approved):
Code changes look good, i've tested the code changes, it does prevent creation of new users with the same e-mail, the js->tsx change is also great for new contact page.

MartinSchoeler (dismissed):
Can we add some tests?

MartinSchoeler (approved):

apps/meteor/client/sidebar/sections/OmnichannelSection.tsx
```diff
@@ -41,7 +41,12 @@ const OmnichannelSection = (props: typeof Box): ReactElement => {
 				{isCallEnabled && <OmnichannelCallToggle />}
 				<OmnichannelLivechatToggle />
 				{hasPermissionToSeeContactCenter && (
-					<Sidebar.TopBar.Action data-tooltip={t('Contact_Center')} icon='address-book' onClick={(): void => handleRoute('directory')} />
+					<Sidebar.TopBar.Action
+						data-tooltip={t('Contact_Center')}
+						aria-label={t('Contact_Center')}
```
ggazzo, line 46: the prop `title`  is not enough for this case?

aleksandernsilva: @ggazzo We had problems before on a few browsers where both, the native tooltip and custom tooltip were displayed at the same time when using `title`. That's why we're using `data-tooltip` here. The `aria-label` helps with the accessibility that was missing.

ggazzo (approved):

PR #27044, opened 2022-10-10
[IMPROVE] Change status/price badge to fuselage tag

rique223:
## Jira task: [MKP-131](https://rocketchat.atlassian.net/browse/MKP-131?atlOrigin=eyJpIjoiODRjOTk4OGYyZDAwNDIzNmI5N2EyZGRhNTUxOGY2NGEiLCJwIjoiaiJ9)
Changed the custom component that showed prices and statuses to a fuselage Tag component for maintainability reasons since after this PR every general design change to the tag component will be automatically reflected in the marketplace UI.
Demo image (App List):
image

Demo Image (App price):
image


apps/meteor/client/views/admin/apps/PriceDisplay.tsx
```diff
@@ -46,13 +46,15 @@ const PriceDisplay: FC<PriceDisplayProps> = ({ purchaseType, pricingPlans, price
...
-			{showType && (
-				<Box color='default' withTruncatedText>
-					{t(type as TranslationKey)}
-				</Box>
-			)}
-			<Box withTruncatedText>{!showType && type === 'Free' ? t(type) : formattedPrice}</Box>
+		<Box mis='x8'>
+			<Tag small>
+				{showType && (
+					<Box color='default' withTruncatedText>
```
juliajforesti, line 52: is this inside Box necessary?

rique223: Currently yes. The fuselage Tag has a rendering problem/implementation particularity that makes it render any props passed to it directly in the dom which, basically, ignores almost anything given to it outside of its pre-determined props, and because of that I had to use the wrapper Box so that I could implement this margin.

rique223: This is currently being solved here: https://github.com/RocketChat/fuselage/pull/865

dougfabris (approved):

PR #27046, opened 2022-10-11
Chore: Migrate EE LivechatRooms model to typescript

murtaza98:
## Issue(s)

OC-376


apps/meteor/ee/app/livechat-enterprise/server/lib/VisitorInactivityMonitor.js
```diff
@@ -28,8 +28,8 @@ export class VisitorInactivityMonitor {
 		SyncedCron.add({
 			name: this._name,
 			schedule: (parser) => parser.cron(everyMinute),
-			job: () => {
-				this.handleAbandonedRooms();
+			job: async () => {
```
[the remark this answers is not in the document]
murtaza98: Couldn't find anything related to that within their docs :see_no_evil: So let me avoid making it async

apps/meteor/ee/server/models/raw/LivechatRooms.ts
```diff
@@ -10,14 +10,93 @@ declare module '@rocket.chat/model-typings' {
...
+	}
+
+	unsetAllOnHoldFieldsByRoomId(roomId: string): Promise<UpdateResult> {
+		return this.updateOne(
+			{
+				_id: roomId,
+			},
+			{
+				$unset: {
+					'omnichannel.predictedVisitorAbandonmentAt': 1,
```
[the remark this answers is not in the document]
murtaza98: Technically yes, however, the goal of this func was to unset both onhold and visitor abandonment fields in the same DB call. But it seems like the function name isn't super clear here, so let me change that now

apps/meteor/ee/app/livechat-enterprise/server/lib/VisitorInactivityMonitor.js
```diff
@@ -90,28 +90,31 @@ export class VisitorInactivityMonitor {
 		const guest = visitor.name || visitor.username;
 		const comment = TAPi18n.__('Omnichannel_On_Hold_due_to_inactivity', { guest, timeout });
 
-		LivechatEnterprise.placeRoomOnHold(room, comment, this.user) &&
-			(await LivechatRooms.unsetPredictedVisitorAbandonmentByRoomId(room._id));
+		await Promise.all([
```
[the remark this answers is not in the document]
murtaza98: That's a good point, let me change it

apps/meteor/ee/app/livechat-enterprise/server/lib/Helper.js
```diff
@@ -235,11 +235,17 @@ export const updateInquiryQueuePriority = (roomId, priority) => {
...
 		LivechatRoomsRaw.findOpenRoomsByPriorityId(priorityId).forEach((room) => {
 			updateInquiryQueuePriority(room._id);
 		}),
 	);
+	if (result.some((result) => result.status === 'rejected')) {
+		logger.error(
+			'Error removing priority from rooms',
+			result.filter((result) => result.status === 'rejected'),
+		);
+	}
```
KevLehman, line 248: Filter first then print, to save one loop :eyes:

apps/meteor/ee/app/livechat-enterprise/server/lib/VisitorInactivityMonitor.js
```diff
@@ -90,31 +91,46 @@ export class VisitorInactivityMonitor {
...
 			LivechatEnterprise.placeRoomOnHold(room, comment, this.user),
 			LivechatRooms.unsetPredictedVisitorAbandonmentByRoomId(room._id),
 		]);
+		if (result.some((r) => r.status === 'rejected')) {
+			logger.error(
+				'Error placing room on hold',
+				result.filter((r) => r.status === 'rejected').map((r) => r.reason),
+			);
+			throw new Error(`Error placing room on hold: ${result.find((r) => r.status === 'rejected').reason}`);
+		}
```
KevLehman, line 104: Filter, then print to reduce one loop

apps/meteor/ee/app/livechat-enterprise/server/methods/removePriority.js
```diff
@@ -11,6 +11,6 @@ Meteor.methods({
 			});
 		}
 
-		return LivechatEnterprise.removePriority(id);
+		Promise.await(LivechatEnterprise.removePriority(id));
```
KevLehman, line 14: methods can be async too :eyes:

apps/meteor/ee/app/livechat-enterprise/server/lib/VisitorInactivityMonitor.js
```diff
@@ -90,31 +91,46 @@ export class VisitorInactivityMonitor {
...
 						break;
 					}
 				}
 			}),
 		);
+		const failed = result.filter((r) => r.status === 'rejected');
+		if (failed.length) {
+			logger.error(
+				'Failed to handle abandoned omnichannel rooms',
+				failed.map((r) => r.reason),
```
KevLehman, line 130: Will this print something good or something with a weird format? Should we join them? Or stringify?

apps/meteor/ee/server/models/raw/LivechatRooms.ts
```diff
@@ -10,14 +10,93 @@ declare module '@rocket.chat/model-typings' {
...
+		unsetPriorityByIdFromAllOpenRooms(priorityId: string): Promise<void>;
+		findOpenRoomsByPriorityId(priorityId: string): FindCursor<IOmnichannelRoom>;
 	}
 }
 
+// Note: Expect a circular dependency error here 😓
 export class LivechatRoomsRawEE extends LivechatRoomsRaw implements ILivechatRoomsModel {
-	async findAbandonedOpenRooms(date: Date): Promise<FindCursor<IOmnichannelRoom>> {
+	async unsetAllPredictedVisitorAbandonment(): Promise<void> {
+		await this.updateMany(
```
KevLehman, line 29: Do you really need to await here? Can't the parent function await this one?

KevLehman (changes requested):

apps/meteor/ee/server/models/raw/LivechatRooms.ts
murtaza98: Indeed they can :smile: But I think I earlier added an await here because the return type of this function wasn't the same as the return type of updateMany. 

On second thought, I think I can modify the return type within a promise handler so we can avoid await here

apps/meteor/ee/app/livechat-enterprise/server/lib/VisitorInactivityMonitor.js
murtaza98: It kind of depends on the error being thrown within the promise. So if it's a normal error like for eg: `throw new Error('Some error msg')`), then the `reason` prop would only have the error string, i.e. `Some error msg`. However, if it's a custom error wrapper (like `Meteor.Error`), then it will have its own object.

In any case, I think the logger which we're using should be able to parse then so don't think we'd explicitly need to stringify them or process them in any special way

cauefcr (dismissed):
Good updates, ran e2e api tests with various configs, read all the code, and found no issues with this change.

murtaza98:
Hey @cauefcr Lost your previous approval here :sweat_smile: Mind checking it again?

KevLehman (approved):

apps/meteor/ee/app/livechat-enterprise/server/lib/Helper.js
```diff
@@ -234,12 +234,18 @@ export const updateInquiryQueuePriority = (roomId, priority) => {
...
-	});
+export const removePriorityFromRooms = async (priorityId) => {
+	const result = await Promise.allSettled(
+		LivechatRoomsRaw.findOpenRoomsByPriorityId(priorityId).forEach((room) => {
+			updateInquiryQueuePriority(room._id);
+		}),
+	);
+	const rejected = result.filter((r) => r.status === 'rejected').map((r) => r.reason);
+	if (rejected.length) {
+		logger.error(`Error while removing priority from ${rejected.length} rooms. Detail: `, rejected);
```
KevLehman, line 245: If a ton of rooms fail, this log will be pretty weird

apps/meteor/ee/app/livechat-enterprise/server/lib/VisitorInactivityMonitor.js
```diff
@@ -90,28 +91,41 @@ export class VisitorInactivityMonitor {
...
 
-		LivechatEnterprise.placeRoomOnHold(room, comment, this.user) &&
-			(await LivechatRooms.unsetPredictedVisitorAbandonmentByRoomId(room._id));
+		const result = await Promise.allSettled([
+			LivechatEnterprise.placeRoomOnHold(room, comment, this.user),
+			LivechatRooms.unsetPredictedVisitorAbandonmentByRoomId(room._id),
+		]);
+		const rejected = result.filter((r) => r.status === 'rejected').map((r) => r.reason);
+		if (rejected.length) {
+			logger.error('Error placing room on hold. Detail: ', rejected);
```
KevLehman, line 100: Same here

apps/meteor/ee/app/livechat-enterprise/server/lib/VisitorInactivityMonitor.js
```diff
@@ -90,28 +91,41 @@ export class VisitorInactivityMonitor {
...
+						break;
+					}
 				}
-			}
-		});
+			}),
+		);
+		const rejected = result.filter((r) => r.status === 'rejected').map((r) => r.reason);
+		if (rejected.length) {
+			logger.error('Failed to handle abandoned omnichannel rooms. Detail: ', rejected);
```
KevLehman, line 126: same here

KevLehman (changes requested):
My idea could be:
1. Print errors on a map so one log per room is printed.
2. Try to print only a generic error and put the actual error array on a debug log to not clog the error output

wdyt?

KevLehman (approved):

cauefcr (approved):
I believe the changes made addressed the concerns raised, approved.

PR #27052, opened 2022-10-11
[FIX] Actionable messages are now sent to integrations webhook

cauefcr:

Actionable messages, with buttons and similar, (saved as blocks internally) were not being sent to external integrations, so CRMs and the like had no visibility on what was the actionable message sent, this PR adds a yaml version of the actions, as well as a proper json object with the same data on the webhook output, so that our clients can have more visibility in their process

## Steps to test or reproduce

 - Enable external integrations
 - Create a webhook for such integration (a simple echo server works fine)
 - Send a message with actionable items (like whatsapp templates)
 - Note that the message above is readable on the webhook

## Further comments

OC-170


apps/meteor/app/livechat/server/hooks/sendToCRM.js
```diff
@@ -64,7 +64,8 @@ function sendToCRM(type, room, includeMessages = true) {
 			const msg = {
 				_id: message._id,
 				username: message.u.username,
-				msg: message.msg,
+				msg: message.msg || JSON.stringify(message.blocks),
```
murtaza98, line 67:
```suggestion
				msg: message.msg || (blocks?.length && JSON.stringify(message.blocks)),
```

Can you please verify if the message actually has "blocks"?  Because I think if it's a media message, then it won't have blocks and msg field set

cauefcr: You're right in the sense that if the message has media it won't have blocks, but if it has media, it will get sent too, like this example (from a webhook):
```
{"_id":"hmK99JeWFmfXFshPc","label":"testupload","createdAt":"2022-10-13T17:59:47.692Z","lastMessageAt":"2022-10-13T18:00:22.924Z","visitor":{"_id":"63485210f4d4a4223b85e4c9","token":"9fc6c1651f0f60a38673b4369a616da590330c0381c9b45fc656842f70ce8690","name":"testupload","username":"guest-2","email":[{"address":"asdasd@example.com"}],"phone":null},"agent":{"_id":"8N28t2QTHTHpgZ6oq","username":"botty","name":"a.bot","email":"caue.felchar+dfbot@rocket.chat"},"type":"Message","messages":[{"u":{"_id":"63485210f4d4a4223b85e4c9","username":"guest-2","name":"testupload"},"_id":"u4M6fD9zWKB5fRNGB","username":"guest-2","ts":"2022-10-13T18:02:21.776Z","file":{"_id":"XoxbLEW9QrXgm6Z2d","name":"image.png","type":"image/png"},"attachments":[{"ts":"1970-01-01T00:00:00.000Z","title":"image.png","title_link":"/file-upload/XoxbLEW9QrXgm6Z2d/image.png","title_link_download":true,"image_dimensions":{"width":1913,"height":861},"image_preview":"/9j/2wBDAAYEBQYFBAYGBQYHBwYIChAKCgkJChQODwwQFxQYGBcUFhYaHSUfGhsjHBYWICwgIyYnKSopGR8tMC0oMCUoKSj/2wBDAQcHBwoIChMKChMoGhYaKCgoKCgoKCgoKCgoKCgoKCgoKCgoKCgoKCgoKCgoKCgoKCgoKCgoKCgoKCgoKCgoKCj/wAARCAAPACADASIAAhEBAxEB/8QAGAAAAgMAAAAAAAAAAAAAAAAAAggAAwX/xAAhEAABAwQDAAMAAAAAAAAAAAABAgMRAAQGEgUxURMhQf/EABQBAQAAAAAAAAAAAAAAAAAAAAL/xAAaEQEBAQEAAwAAAAAAAAAAAAABAgARITFB/9oADAMBAAIRAxEAPwDAxHK8guMssLV/kHnG3HIKfw02NqwNAFtjYd0mGCn5M44oSdS8KdQbWoSB0TAp2NcOuBIezEptpAlbQA9qxm3ZUJCU/flFcMKdZUlS4nyowsNoSjXoRM0JaPHXNmfm/9k=","image_url":"/file-upload/XoxbLEW9QrXgm6Z2d/image.png","image_type":"image/png","image_size":213663,"type":"file"}],"fileUpload":{"publicFilePath":"http://localhost:3000/file-upload/XoxbLEW9QrXgm6Z2d/image.png?token=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJpYXQiOjE2NjU2ODQxNDEsIm5iZiI6MTY2NTY4NDE0MSwiZXhwIjoxNjY1Njg3NzQxLCJhdWQiOiJSb2NrZXRDaGF0IiwiY29udGV4dCI6eyJ1c2VySWQiOiI2MzQ4NTIxMGY0ZDRhNDIyM2I4NWU0YzkiLCJmaWxlSWQiOiJYb3hiTEVXOVFyWGdtNloyZCJ9fQ.RLM_dfq5noD9d2Q2VOYVwQ8p_56rFRCJOmPd8vqcPe8","type":"image/png","size":213663}}]}
```

you can see why in [apps/meteor/app/livechat/server/hooks/sendToCRM.js#L86](https://github.com/RocketChat/Rocket.Chat/blob/ff8ee18d3bdb25ee1150106847f23f50172af580/apps/meteor/app/livechat/server/hooks/sendToCRM.js#L86)

KevLehman: I think he's referring it more to what will `JSON.stringify(null) or JSON.stringify(undefined)` will produce 👀

cauefcr: it stays with nothing in the msg field (you can see it in the example above) if the message only has an upload.

KevLehman: And won't that fail? Like, the thing was failing cause msg was empty 👀 won't an empty message cause the same?

murtaza98: Yes, indeed my concern was handling the null values but I realised that stringify doesn't throw errors in-case of null and undefined. So we should be safe then :relieved: 
image

murtaza98 (approved):

KevLehman (approved):

PR #27054, opened 2022-10-11
Chore: create addOAuthApp endpoint

felipe-rod123:

Created the 'oauth-apps.addOAuthApp' endpoint for the `apps/meteor/client/views/admin/oauthApps/OAuthAddApp.tsx` file, and added Ajv validation.


sampaiodiego (approved):
missing tests though

sampaiodiego (changes requested):
the return type of the new endpoint is not compatible with the old method. it should return the created application

debdutdeb:
Should this set of endpoints not check for `manage-oauth-apps` permission?

debdutdeb:
Hm it's being checked in the meteor method. A bit confusing considering the new `permissionRequired` option being part of addRoute now :)

sampaiodiego (approved):

PR #27056, opened 2022-10-11
[IMPROVE] Move app list install logic to app menu

rique223:
## Jira task: [MKP-136](https://rocketchat.atlassian.net/browse/MKP-136?atlOrigin=eyJpIjoiM2M0ODhkZDc0YjQ2NGE5OGE4YTBkZDJhY2M0NGVmZDgiLCJwIjoiaiJ9)
Moved the install/buy/subscribe logic from the CTA buttons that existed in the app list to the app menu. Now the user can install apps by clicking in the kebab menu in the right of the app list entry. Also solved a little undefined bug happening in the app permissions modal.
Demo gif:
app_menu


apps/meteor/client/views/admin/apps/AppMenu.js
```diff
@@ -1,11 +1,35 @@
...
 
+import { Apps } from '../../../../app/apps/client/orchestrator';
 import WarningModal from '../../../components/WarningModal';
+import AppPermissionsReviewModal from './AppPermissionsReviewModal';
 import CloudLoginModal from './CloudLoginModal';
 import IframeModal from './IframeModal';
-import { appEnabledStatuses, warnStatusChange, handleAPIError } from './helpers';
+import { appEnabledStatuses, warnStatusChange, handleAPIError, appButtonProps, handleInstallError } from './helpers';
+
+const installApp = async ({ id, name, version, permissionsGranted }) => {
```
juliajforesti, line 12: Both `installApp` and `actions` are duplicated in AppStatus and AppMenu, doing the same. Isn't it worth exporting them or moving them to another file?

juliajforesti (changes requested):

apps/meteor/client/views/admin/apps/AppMenu.js
rique223: Solved

juliajforesti (approved):

PR #27059, opened 2022-10-13
Chore: Bump Apps-Engine version

d-gubert:
Bump Apps-Engine version to latest stable


tassoevan (approved):
I've some reservations about using dist-tags instead of version numbers, but let's see if `latest` is more stable.

PR #27070, opened 2022-10-14
[NEW] Matrix federation Users' typing event

MarcosSpessatto:

This PR introduces the ability to enable the typing indicator, it's only possible to do that, by enabling the push ephemeral events in the bridge configuration file (yaml). 

There is no option to enable it manually, strictly only in the configuration file.

Jira: FED-47


apps/meteor/app/federation-v2/server/infrastructure/rocket-chat/adapters/Settings.ts
```diff
@@ -202,8 +234,30 @@ export class RocketChatSettingsAdapter {
...
 					i18nDescription: 'Federation_Matrix_registration_file_desc',
+					alert: 'Federation_Matrix_registration_file_Alert',
 				});
 			});
 		});
 	}
+
+	private getRegistrationFileFromHomeserver(): Record<string, any> | undefined {
+		try {
+			const registrationYaml = Assets.getText('federation/registration.yaml');
```
MarcosSpessatto, line 245: @RocketChat/chat-engine The purpose here is to read a file from a place in the File System in order to enable/disable the setting. We are doing that to force the user to strictly enable it using the Matrix homeserver config file.

I'm following @alansikora suggestion here to read the file from the `private` folder and use the `Assets.getText()` method. I'm not sure if this is the best way to go, appreciate any suggestion. Thanks.

apps/meteor/server/modules/notifications/notifications.module.ts
```diff
@@ -250,6 +251,17 @@ export class NotificationsModule {
 				return false;
 			}
 
+			if (
+				e === 'typing' ||
+				(e === 'user-activity' && Array.isArray(_activity) && (_activity.length === 0 || _activity.includes('user-typing')))
+			) {
+				api.broadcast('user.typing', {
```
MarcosSpessatto, line 258: @RocketChat/chat-engine I've created this event to notify when the user is typing, but I'm not sure also if the best approach. Any suggestions on how to improve this? Thanks.

alansikora (approved):

apps/meteor/app/federation-v2/server/infrastructure/rocket-chat/adapters/Settings.ts
sampaiodiego: this would mean people would have to build rocket.chat from source after creating this file, is that what you want to be done?

MarcosSpessatto: For sure, it's not. So, what is the proper way to do that?

The final desire here is to be able to enable/disable a setting only and only through this file (it's the same file used to configure the Matrix Bridge).

Is there any other (and suggested) folder to place this file? Also, should we use native NodeJS `fs` instead of `Assets.getText()`?

MarcosSpessatto: @sampaiodiego I've changed it to use `fs` reading from a folder in the meteor root project.

apps/meteor/server/modules/notifications/notifications.module.ts
pierre-lehnen-rc: The event itself is fine, but calling it from inside `allowWrite` is not ideal.

You should use `on` to listen to the event directly. This way you'll also only trigger it for the rooms that you are expecting it to.

```js
notifications.streamRoom.on(`${rid}/user-activity`, (username, activity) => {
  if (Array.isArray(activity) && (!activity.length || activity.includes('user-typing')) {
    api.broadcast('user.typing', {
      user: { username },
      isTyping: activity.includes('user-typing'),
      roomId: rid,
    });
  }
});
```

Issue #23051, opened 2021-08-27
Use Real Names


ddv88:
This problem has not been fixed for more than 2 years..
Maybe you should remove the setting with names altogether?
Let the full name always be used by default. If the account does not have a full name only then use the username.

<details>
  <summary>SHOW</summary>

![Rocket Chat - httpsrocket-chat unictech r2u](https://user-images.githubusercontent.com/8138405/131099231-e9f94b29-70ee-4c1e-bd13-6791572d2746.png) ![Rocket Chat - httpsrocket-chat unictech ru](https://user-images.githubusercontent.com/8138405/131099038-8b97dd35-1ae0-4d16-849f-b061456debf3.png)

</details>

```
Deployment
Version
3.17.2
Apps Engine Version
1.27.1
Node Version
v12.22.1
Database Migration
228 (август 27, 2021 10:40)
MongoDB
4.0.26 / mmapv1 (oplog Enabled)
Commit Details
HEAD: (fe0e5e435)
Branch: HEAD
```

debdutdeb:
Hi 

The "use real name" setting seems to be working, is it not? If it's turned on, and some one is mentioned in a room, the full name is showed instead of username.

Or are you talking about the quoted messages?

> This problem has not been fixed for more than 2 years..

Can you please put a hyperlink to that issue here? I have a meeting with the product team today, I'll see if I can bring it up then.

One final thing, please try to **always** fill the bug template. I know sometimes it might seem unnecessary, but that's not the point. It's about making it a habit, a practice. If you don't (no matter the necessity), someone else will see and follow without really understanding why you did what you did. I'm just trying to make this issue list as uniform and presentable as possible. 

Thank you.

lebkayn:
one place comes to my mind: on starting a new direct message with another user. on top of it it says "You have joined a new direct message with user_name", not using the real name

ddv88:
Hi!

> Or are you talking about the quoted messages?

Yep. Quotes, Threads, Resend messages etc.
I saw, that on public test server there no issue with this. I thought that problem is on our server, but now i ran absolutely clean install in Docker and the problem is still exist.

> Can you please put a hyperlink to that issue here?
I have not started a thread before, but I have seen already created topics for the same problem.

> I'm just trying to make this issue list as uniform and presentable as possible.
I've got it. Sorry for this

dmonn:
[...]

Issue #21480, opened 2021-04-07
Real name resolving in channel member selection dialog


Gummikavalier:
### Description:

Many other real name resolving issues have been and are being resolved, so I'll also report the specific issue with channel member selection dialog.

This issue is inconvenient on systems that get illegible provisioned usernames from SAML and Oauth endpoints and rely mainly on the real name to tell people apart.

### Steps to reproduce:

1. Enable `Use Real Name` under `Administrator -> Layout -> User Interface`
2. Create a user with an illegible username (here `099fnd2ee`) and add a real name to the user (here `Test Account`)
3. Invite the user to a channel

### Expected behavior:

After selecting the name in the channel add user dialog, real name is shown in the box after the selection.

To be exact, for the security reasons (several users with the same name), both values could be shown during and after making the selection.

### Actual behavior:

After selecting the real name from the list, it transforms into a username in the box with already selected users, making over viewing especially bigger selection impossible just before confirming the selections.

Selecting one user by their real name:
![Screenshot from 2021-04-07 17-31-33](https://user-images.githubusercontent.com/6958942/113883921-42352f80-97c7-11eb-90ae-76c427302cda.png)

After the selection the real name transforms into a username:
![Screenshot from 2021-04-07 17-31-50](https://user-images.githubusercontent.com/6958942/113883936-45302000-97c7-11eb-82ab-8ec17ce3ce11.png)

The report after the user has been added also shows the username instead of the real name. User can find out whether the addition of correct person succeeded only by cross referencing the name and the account elsewhere (search for instance).
![Screenshot from 2021-04-07 17-32-11](https://user-images.githubusercontent.com/6958942/113883950-48c3a700-97c7-11eb-980c-938b22d21de3.png)

### Server Setup Information:

- Version of Rocket.Chat Server:  3.12.3
- Operating System: CentOS7
- Deployment Method: tar
- Number of Running Instances: 20
- DB Replicaset Oplog: Yes
- NodeJS Version: 12.18.4
- MongoDB Version: 4.0

### Client Setup Information

- Desktop App or Browser Version: Any web browser

Gummikavalier:
Closing as fixed.

Issue #21129, opened 2021-03-15
User info shows username instead of full name


billynoah:
### Description:

The “Full Name” that’s shown when clicking “View Profile” on a user in a channel shows only their *username* instead of their *full name.* The correct full name is shown in the admin panel user info but not on the user front end.

I have tested or verified this in both the desktop app and the web ui.

### Steps to reproduce:

1. Click on a username in a channel
2. Click "See full profile"
3. A panel opens called "User Info".  Under "Full Name" it shows the username instead of the full name.

### Expected behavior:

The "*Full Name*" section should show the user's *full name*.

### Actual behavior:

The "*Full Name*" section wrongly shows the user's *username*.

![Screen Shot 2021-03-15 at 4 12 18 PM](https://user-images.githubusercontent.com/3309911/111222429-32ed0880-85b2-11eb-86a4-f87d31536394.png)

And here is a view of the same user from the admin panel which shows the correct *full name*:

![Screen Shot 2021-03-15 at 4 14 46 PM](https://user-images.githubusercontent.com/3309911/111223635-cbd05380-85b3-11eb-99fc-4adf00c8b812.png)

### Server Setup Information:

- Rocketchat Version 3.12.1
- Ubuntu 20.04
- Installed via snaps
- Desktop App v3.1.1 / All Browsers

### Version of Rocket.Chat Server

- Rocketchat Version 3.12.1

### Client Setup Information

- All browsers
- OS X / Windows 10

Darshilp326:
@billynoah Hello! You need to turn ON the setting 'UI_Use_Real_Name' for displaying real names instead of usernames.

billynoah:
@Darshilp326 - Where do I find this setting?

shrinish123:
@Darshilp326 after looking at the code even I felt that there has to a setting to show real names, but not able to find it.

Darshilp326:
@billynoah @shrinish123 Administration->Layout->User Interface

Darshilp326:
@billynoah But sadly code is implemented by using this setting :disappointed:

billynoah:
I see what you mean - in /client/views/room/contextualBar/UserInfo/index.js:

    name: showRealNames ? name : username

That really seems like the wrong use of that setting to me.  I'd propose we either need another setting or simply change the code to:

    name: name

It's nonsensical to display the username under "full name" since that's clearly not what it is.

Darshilp326:
@dougfabris Can you look into this issue?

[...]

PR #27072, opened 2022-10-14
fix: Display Avatar Preference not hiding avatars on messages

### LINKED ISSUE #26741: Display Avatars switch:  unable to disable showing avatars in messages
User Preference:
Display Avatars: Hides avatars on the channels, showing only the users' usernames.

This setting since version 5 does not disable the viewing of user's avatars in messages.

nishant23122000:
## Issue(s)
Closes #26741

## Steps to test or reproduce

- Go to the user account
- Go to preferences
- In Message Section use the toggle button to disable or enable the **Display avatar**

You will able to toggle it, but if it's enabled, messages are rendering along with the user avatar.


nishant23122000:
This same issue happens with a The hide username also. before that, I'm Waiting for the reviewer to check whether it's the right approach to do or not.

dudanogueira:
Hi! There is at least 3 issues reagarding the real name issue areound:
https://github.com/RocketChat/Rocket.Chat/issues/23051
https://github.com/RocketChat/Rocket.Chat/issues/21480
https://github.com/RocketChat/Rocket.Chat/issues/21129

Indeed it needs more discussion. 
Thanks for this PR!! I'll put it with our team.

apps/meteor/client/views/room/MessageList/components/Message.tsx
```diff
@@ -29,6 +30,7 @@ const Message: FC<{
 		actions: { openUserCard },
 	} = useMessageActions();
 
+	const showUserAvatar = !!useUserPreference<boolean>('displayAvatars');
```
debdutdeb, line 33:
```suggestion
	const showUserAvatar = useUserPreference<boolean>('displayAvatars');
```
No need to convert.

debdutdeb (changes requested):

apps/meteor/client/components/message/variants/RoomMessage.tsx
```diff
@@ -41,6 +41,7 @@ const RoomMessage = ({ message, sequential, all, mention, unread, context, ignor
 	const chat = useChat();
 	const messageRef = useRef(null);
 
+	const showUserAvatar = useUserPreference<boolean>('displayAvatars');
```
ggazzo, line 44: would you mind passing that value via prop?
I'm afraid this hook might add a bit of burden to rendering as it's the same for all list items, maybe you could use it in MessageList.tsx

ggazzo (changes requested):
please check the comment and apply to all cases

apps/meteor/client/components/message/variants/RoomMessage.tsx
nishant23122000: Sure @ggazzo

nishant23122000: Please check now.

ggazzo (approved):

PR #27074, opened 2022-10-14
[BREAK] Move read-receipts to EE and threads check marks improvements

matheusbsilva137:

- Improvement: add read receipts for messages in threads only when the thread is open;
- Move read-receipts feature to EE (under the `message-read-receipt` module). The `chat.getMessageReadReceipts` endpoint and `getReadReceipts` Meteor method are now EE only.

## Further comments

[TC-73]

[TC-73]: https://rocketchat.atlassian.net/browse/TC-73?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/meteor/server/services/reads/service.ts
```diff
@@ -0,0 +1,40 @@
...
+
+		const threadMessage = Messages.findOneById(tmid, { projection: { ts: 1, tlm: 1, rid: 1 } });
+		if (!threadMessage || !threadMessage.tlm) {
+			return;
+		}
+
+		ReadReceipt.storeThreadMessagesReadReceipts(tmid, userId, read?.ls || threadMessage.ts);
+
+		// doesn't mark as read if not all room members have read the thread
+		const subscriptions = await Subscriptions.findByRoomId(threadMessage.rid, {
```
sampaiodiego, line 25: this query should filter out archived subscriptions.. please remember to create a test case for this scenario

apps/meteor/server/services/reads/service.ts
```diff
@@ -0,0 +1,40 @@
...
+import { Messages, Subscriptions } from '../../../app/models/server';
+import { ReadReceipt } from '../../../imports/message-read-receipt/server/lib/ReadReceipt';
+
+export class ReadsService extends ServiceClassInternal implements IReadsService {
+	protected name = 'reads';
+
+	async readThread(userId: string, tmid: string): Promise<void> {
+		const read = await Reads.findOneByUserIdAndThreadId(userId, tmid);
+
+		await Reads.updateReadTimestampByUserIdAndThreadId(userId, tmid);
```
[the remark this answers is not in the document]
matheusbsilva137: Fixed!

apps/meteor/server/services/reads/service.ts
matheusbsilva137: Fixed! I'll now add some tests for this

outofrange:
Since I just stumbled over this, and haven't found any other info about `Move read-receipts feature to EE` - does this mean read notifications on chat messages (a very useful feature) will be removed from the upcoming Community Edition?

LucianoPierdona (dismissed):

hugocostadev (dismissed):

milton-rucks:
> Since I just stumbled over this, and haven't found any other info about `Move read-receipts feature to EE` - does this mean read notifications, those little check marks to the right of each message, a very useful feature, will be removed from the upcoming Community Edition?

Hi @outofrange 

Yes, it’s correct.
We are improving the read receipts on threads and moving the feature to EE.
It has been announced since 2020 as part of the Enterprise offer through our old pricing page, but due to a mistake, it continued to be available on CE.
The official communication will come together with the 6.0 launch.

apps/meteor/ee/server/local-services/reads/service.ts
```diff
@@ -0,0 +1,48 @@
...
+		const threadMessage = Messages.findOneById(tmid, { projection: { ts: 1, tlm: 1, rid: 1 } });
+		if (!threadMessage || !threadMessage.tlm) {
+			return;
+		}
+
+		await Reads.updateReadTimestampByUserIdAndThreadId(userId, tmid);
+		ReadReceipt.storeThreadMessagesReadReceipts(tmid, userId, read?.ls || threadMessage.ts);
+
+		// doesn't mark as read if not all room members have read the thread
+		const subscriptions = await Subscriptions.findByRoomId(threadMessage.rid, {
```
rodrigok, line 25: You should retrieve the subscriptions only if they are less then `MAX_ROOM_SIZE_CHECK_INDIVIDUAL_READ_RECEIPTS`. So you can do a count first and retrieve them later, you can exclude the archived ones in the query as well

rodrigok (changes requested):

LucianoPierdona (dismissed):

apps/meteor/server/models/raw/Reads.ts
```diff
@@ -0,0 +1,62 @@
...
+	}
+
+	getMinimumLastSeenByThreadId(tmid: IMessage['_id']): Promise<Reads | null> {
+		return this.findOne(
+			{
+				tmid,
+			},
+			{
+				sort: {
+					ls: 1,
```
rodrigok, line 27: A sort without an index is always a bad idea ;)

apps/meteor/tests/end-to-end/api/05-chat.js
```diff
@@ -1346,39 +1346,6 @@ describe('[Chat]', function () {
 		});
 	});
 
-	describe('[/chat.getMessageReadReceipts]', () => {
```
rodrigok, line 1349: Why were those testes removed?

rodrigok (changes requested):

apps/meteor/tests/end-to-end/api/05-chat.js
LucianoPierdona: It's referenced here to add a new file to cover both EE and CE, I believe:
https://github.com/RocketChat/Rocket.Chat/pull/27074/commits/a31375aeeeb33afe1c05aa22cd1f4c757848722d

Since this PR is supposed to go to the 6.0, I don't think it's viable to do these tests today

apps/meteor/server/models/raw/Reads.ts
matheusbsilva137: Index added. I had one index there already, but I guess it's best to keep this as a separate one, right?

apps/meteor/tests/end-to-end/api/05-chat.js
matheusbsilva137: Yeah, not only I think this should be in a separate file, but also this needs to be updated since this endpoint is now EE-only. I plan to get the tests I worked on #26611 (old PR related to this issue), but these also need to be changed since this feature has been moved to EE
I'm planning to add this as a regression later so as not to hold this PR

apps/meteor/ee/server/local-services/reads/service.ts
```diff
@@ -0,0 +1,50 @@
+import { Reads, Subscriptions } from '@rocket.chat/models';
+import type { ISubscription } from '@rocket.chat/core-typings';
+import { ServiceClassInternal } from '@rocket.chat/core-services';
+
+import type { IReadsService } from '../../sdk/types/IReadsService';
+import { Messages } from '../../../../app/models/server';
+import { ReadReceipt } from '../../lib/message-read-receipt/ReadReceipt';
+import { MAX_ROOM_SIZE_CHECK_INDIVIDUAL_READ_RECEIPTS } from '../../lib/constants';
+
+export class ReadsService extends ServiceClassInternal implements IReadsService {
```
KevLehman, line 10: A bit picky, but could we name this `ReadReceipts`? 
I know `Reads` makes sense when you have the context, but otherwise it's not very descriptive

apps/meteor/ee/app/message-read-receipt/server/hooks/afterReadMessages.ts
```diff
@@ -0,0 +1,24 @@
+import type { IUser, IRoom, IMessage } from '@rocket.chat/core-typings';
+
+import { Reads } from '../../../../server/sdk';
```
KevLehman, line 3: Can't you import it from @rocket.chat/core-services?

apps/meteor/ee/app/message-read-receipt/server/index.js
```diff
@@ -0,0 +1,5 @@
+import { onLicense } from '../../license/server';
+
+onLicense('message-read-receipt', () => {
+	require('./hooks');
```
KevLehman, line 4: You can use `await import('....')` to not use require 👀

apps/meteor/ee/server/api/chat.ts
```diff
@@ -0,0 +1,33 @@
...
+
+			try {
+				return API.v1.success({
+					receipts: await Meteor.call('getReadReceipts', { messageId }),
+				});
+			} catch (error: any) {
+				return API.v1.failure({
+					error: error.message,
+				});
+			}
```
KevLehman, line 30: Since the error is just being returned, do you really need the try/catch or can we rely on the endpoint's try/catch?

KevLehman (changes requested):

apps/meteor/tests/end-to-end/api/05-chat.js
matheusbsilva137: Thanks @LucianoPierdona for adding this :hugs:

apps/meteor/ee/server/local-services/reads/service.ts
matheusbsilva137: I do agree that `Reads` is not a meaningful name, but I also think that calling it `ReadReceipts` could be misleading :thinking: since it manages the `Reads` collection (which currently works like subscriptions to threads, but we plan to expand it)

apps/meteor/ee/app/message-read-receipt/server/hooks/afterReadMessages.ts
matheusbsilva137: Done! Thanks for suggesting this, though (I had forgotten to add this service to the `core-services` package :grimacing: )

apps/meteor/ee/app/message-read-receipt/server/index.js
matheusbsilva137: Done! Also converted this file to TS

apps/meteor/ee/server/api/chat.ts
matheusbsilva137: Hmmm I see we're doing something similar (the try/catch strategy) in some other endpoints. Not sure about what's the ideal behavior here. Maybe @sampaiodiego can help us here

apps/meteor/tests/end-to-end/api/05-chat.js
```diff
@@ -1347,8 +1347,14 @@ describe('[Chat]', function () {
 	});
 
 	describe('[/chat.getMessageReadReceipts]', () => {
+		const isEnterprise = typeof process.env.IS_EE === 'string' ? process.env.IS_EE === 'true' : !!process.env.IS_EE;
 		describe('when execute successfully', () => {
 			it("should return the statusCode 200 and 'receipts' property and should be equal an array", (done) => {
+				if (!isEnterprise) {
+					done();
```
rodrigok, line 1354: Would be better to skip it, so we know it's a test that didn't run rather than run successfully

apps/meteor/ee/server/local-services/reads/service.ts
KevLehman: Just out of curiosity, why we needed another collection for thread subscriptions? 👀

apps/meteor/tests/end-to-end/api/05-chat.js
matheusbsilva137: Agree, done!

apps/meteor/app/threads/server/methods/getThreadMessages.js
```diff
@@ -28,18 +29,24 @@ Meteor.methods({
 
 		const user = Meteor.user();
 		const room = Rooms.findOneById(thread.rid);
+		callbacks.run('beforeReadMessages', thread.rid, user._id);
```
sampaiodiego, line 32: this should be moved to right before the call to `readThread(..)`, after all checks are done

apps/meteor/ee/server/local-services/reads/service.ts
matheusbsilva137: That's because we decided to treat read receipts for messages in threads just as if threads were "separate rooms". So in order to control which messages have been read, we gotta know whether everyone in the room opened the thread and the messages they read.
For messages in the main room, we use subscriptions' `ls` field (`lastSeen`), so that we can get the minimum value of this field among room members and mark all older messages as read. We needed something similar for threads.

apps/meteor/ee/server/api/chat.ts
sampaiodiego: this try/catch is redundant.. as @KevLehman mentioned we already have that for every REST endpoint. as you can see [here](https://github.com/RocketChat/Rocket.Chat/blob/2bea8dc006351de264e1424534f0299098c304fd/apps/meteor/app/api/server/api.js#L419)

apps/meteor/server/methods/readThreads.js
```diff
@@ -22,12 +23,15 @@ Meteor.methods({
 		}
 
 		const user = Meteor.user();
+		callbacks.run('beforeReadMessages', thread.rid, user._id);
```
sampaiodiego, line 26: same as the other one, this should be moved to right before the call to `readThread()` function

packages/core-typings/src/IRoom.ts
```diff
@@ -74,6 +74,7 @@ export interface IRoom extends IRocketChatRecord {
 	usernames?: string[];
 	ts?: Date;
 
+	closedAt?: Date;
```
sampaiodiego, line 77: why was this moved from `IOmnichannelGenericRoom` to `IRoom`?

sampaiodiego (changes requested):

apps/meteor/ee/server/local-services/reads/service.ts
sampaiodiego: what do you guys think of calling the service as `MessageReads`? a bit more meaningful I guess 😬

packages/core-typings/src/IRoom.ts
matheusbsilva137: That's because I converted [this file](https://github.com/RocketChat/Rocket.Chat/blob/1e5a0650d9afd8c2649765241e14f58a40443f62/apps/meteor/imports/message-read-receipt/server/hooks.js#L14) to TS and I'd have the `"Property closedAt does not exist on type IRoom"` error if I didn't do so. Also, since any room can be closed and have this property, I guess it makes sense to move it to `IRoom`. wdyt?

apps/meteor/server/methods/readThreads.js
matheusbsilva137: Thankss, fixed!

apps/meteor/ee/server/api/chat.ts
matheusbsilva137: great, fixed!

packages/core-typings/src/IRoom.ts
sampaiodiego: not really, only omnichannel rooms can be closed.. or am I missing something? how do you close a room that is not an omnichannel room?

apps/meteor/app/threads/server/methods/getThreadMessages.js
matheusbsilva137: tks, done!

apps/meteor/ee/server/local-services/reads/service.ts
matheusbsilva137: hmmm sound better to me :thinking: it's currently just "threads' reads", but it's better for us not to be so specific, I guess
I'll update the service name

matheusbsilva137: Done! Updated the service name to `MessageReads` and the collection name to `message_reads`

packages/core-typings/src/IRoom.ts
matheusbsilva137: You're right, I guess I confused it with the `closedAt` field in subscriptions :grimacing: 
That's fixed.

sampaiodiego (approved):

rodrigok (approved):

ggazzo (approved):

KevLehman (approved):

LucianoPierdona (approved):

ae5960e8-a6fc-491f-b252-898ecf59af95:
Please see this forum post regarding this change:
https://forums.rocket.chat/t/update-6-0-0-read-receipts-complaint/16164/4

Issue #26723, opened 2022-08-29
Omnichannel transfer comment is not wrapped to multiple lines


krenuser:
### Description:

When transferring omnichannel conversation to another agent with a comment, text is not wrapping to new line, going out of the line with "...".

### Steps to reproduce:

1. Agent A: Receive omnichannel conversation.
2. Transfer it to agent B, leaving long text comment.
3. See on agent B's chat non-wrapped text (rest of it is not visible).

### Expected behavior:

Transfer comment text is wrapped to multiple lines.

### Actual behavior:

Transfer comment text is rendered in one line hidden behind the end of the line.

![image](https://user-images.githubusercontent.com/32645945/187180012-bcc24a73-469b-41e2-be4e-b396c495e2d4.png)

### Server Setup Information:

- Version of Rocket.Chat Server: 5.0.4
- Operating System: Ubuntu 20.04
- Deployment Method: docker
- Number of Running Instances: 1
- DB Replicaset Oplog: Enabled
- NodeJS Version: v14.19.3
- MongoDB Version: 5.0.11 / wiredTiger

### Client Setup Information

- Desktop App or Browser Version: Desktop 3.8.7
- Operating System: Windows 10

### Additional context

─
### Relevant logs:

─

dudanogueira:
Hi! Nice catch :)

As a dirty workaround, while work on fixing this .... :see_no_evil: 
you can add this at ADMIN > GENERAL > LAYOUT > CUSTOM CSS:

div.rcx-message-system__body{
  white-space: pre-wrap !important;
}

It may affect other elements around, but at least will make  the comments readable.

Thanks for pointing out. We really appreciate it.

vorbrodt:
Is this being worked on? Otherwise, I would like to be assigned to it.

sandeep13122002:
is anyone working on this issue?

dudanogueira:
@vorbrodt and @sandeep13122002, thanks for the interest!!

Please, instead of asking if there is anyone already working on the issue, you can go ahead and fix it.

If you have any doubts about how to achieve that, you ask at https://open.rocket.chat/channel/hacktoberfest

Thanks!

nishant23122000:
Hui @dudanogueira Can you please go through this PR https://github.com/RocketChat/fuselage/pull/891#issuecomment-1313759737 and put your thoughts? Thanks

anikdhabal:
[...]

PR #27080, opened 2022-10-16
[IMPROVE] Omnichannel Transfer Comment Not Wrapped To Multiple Mines

nishant23122000:
## Issue(s)
#26723

## Steps to test or reproduce
mentioned in the issue

## Further comments

Added separate CSS for system messages inside livechat.css. but I feel like the more great approach is to directly modify this inside Fuselage SystemMessage Component.

### DIFF
--- a/apps/meteor/app/livechat/client/stylesheets/livechat.css
+++ b/apps/meteor/app/livechat/client/stylesheets/livechat.css
@@ -150,6 +150,7 @@
 .rcx-message-system__block {
 	align-items: flex-start !important;
 }
+
 .rcx-message-system__body {
 	white-space: pre-wrap !important;
 }
--- a/apps/meteor/app/livechat/client/stylesheets/livechat.css
+++ b/apps/meteor/app/livechat/client/stylesheets/livechat.css
@@ -146,3 +146,10 @@
 .contact-chat-history-search-empty {
 	padding-left: 24px;
 }
+
+.rcx-message-system__block {
+	align-items: flex-start !important;
+}
+.rcx-message-system__body {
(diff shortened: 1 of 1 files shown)


apps/meteor/app/livechat/client/stylesheets/livechat.css
```diff
@@ -146,3 +146,11 @@
...
 	padding-left: 24px;
 }
+
+.rcx-message-system__block {
+	align-items: flex-start !important;
+}
+
+.rcx-message-system__body {
+	white-space: pre-wrap !important;
+}
```
ggazzo, line 156: this is not the place where the changes should be placed, please look for the message component

ggazzo (changes requested):

apps/meteor/app/livechat/client/stylesheets/livechat.css
nishant23122000: Can you check now again?

nishant23122000: Yes, that's true, I tried it with @rocket.chat/css-in-js, but it's not working with it. 

any CSS file? where you handle Message Component custom styling?

dougfabris: take a look at the Fuselage's component!

nishant23122000: https://rocketchat.github.io/fuselage/fuselage/master/?path=/story/message-messagesystem--default

not find any specific for this ...

PR #27083, opened 2022-10-17
[NEW] [EE] Omnichannel Priorities & SLAs

KevLehman:
This PR introduces a new way of assigning priorities to Omnichannel rooms. Previously, users could set "priorities" to rooms and they were a "fire and forget" approach.
Now, priorities are more integrated into the UI, allowing agents to have more visibility of what's going on.

The old feature known as "Priorities" (Prior 6.0) is now renamed "Service Level Agreements". They will work exactly the same as the previous one, so if an organization had Priorities set, they will now become SLAs!

## Further comments

Sorting rooms by Priorities
image

Assigning Priorities to rooms from Contextual Bar
image

Assigning Priorities to rooms from Sidebar
image

Viewing Priority of current rooms from Current chats panel
image

Viewing Priority from Room's sidebar
image


apps/meteor/client/sidebar/RoomList/SideBarItemTemplateWithData.tsx
```diff
@@ -134,13 +138,18 @@ function SideBarItemTemplateWithData({
...
-			<Badge {...({ style: { flexShrink: 0 } } as any)} variant={variant}>
-				{unread + tunread?.length}
-			</Badge>
-		) : null;
+
+	const badges = (
+		<InlineGroup gap={8}>
+			{showBadge &&
+				isUnread && ( // TODO: Remove any
+					<Badge {...({ style: { display: 'inline-flex', flexShrink: 0 } } as any)} variant={variant}>
```
filipemarins, line 146: Does `any` still required?

MartinSchoeler (changes requested):
https://user-images.githubusercontent.com/20868078/218504386-39a6bfb3-66c9-488c-bf2b-d0a60f38fc1b.mov

I've also found a small issue with the current chats page, the initial icon for the rooms is incorrect, but as soon as I update the page (click on sort for example) it turns into the correct icon

MartinSchoeler (changes requested):
Captura de Tela 2023-02-13 às 12 49 00
 
 
 We probably need to define a max width here...

MartinSchoeler (changes requested):
Same as the comment above for fields, long names for priorities extend the sidebar
Captura de Tela 2023-02-13 às 12 50 22

apps/meteor/client/sidebar/RoomList/SideBarItemTemplateWithData.tsx
aleksandernsilva: Unfortunately yes.

MartinSchoeler (dismissed):

MartinSchoeler (dismissed):

aleksandernsilva (approved):

MartinSchoeler (dismissed):

tsconfig.base.server.json
```diff
@@ -2,7 +2,7 @@
 	"extends": "./tsconfig.base.json",
 	"compilerOptions": {
 		"target": "es2020",
-		"lib": ["es2020"],
+		"lib": ["es2020", "dom", "dom.iterable"],
```
sampaiodiego, line 5: IMO the "server" base tsconfig should not include `dom` and `dom.iterable` by default.. if it is required by package, it should be included in that package only.

see https://github.com/RocketChat/Rocket.Chat/blob/222fcc4f859970b3a76e9fedc0dd1817ba2b75a5/packages/cas-validate/tsconfig.json#L4

apps/meteor/server/startup/migrations/v289.ts
```diff
@@ -0,0 +1,87 @@
...
+	constructor(db: Db) {
+		super(db, 'livechat_priority');
+	}
+}
+
+// Migration to migrate old priorities to new SLAs, plus for new priorities feature
+addMigration({
+	version: 289,
+	async up() {
+		const isEE = isEnterprise();
```
sampaiodiego, line 28: it is not safe to check this during a migration.. it is not guaranteed a license is applied at this point.. usually licenses coming from Cloud take longer to be applied.

so either remove any check for `isEE` on this migration, or change it to run at some point later.

apps/meteor/server/startup/migrations/v289.ts
```diff
@@ -0,0 +1,87 @@
...
+import { db } from '../../database/utils';
+
+interface IOldLivechatPriority extends IRocketChatRecord {
+	_id: string;
+	name: string;
+	description: string;
+	dueTimeInMinutes: number;
+}
+
+class OldLivechatPriorityRaw extends BaseRaw<IOldLivechatPriority> {
```
sampaiodiego, line 18: you don't need to create a model for a non-existent collection.. instead you can use the MongoDB driver directly by doing, for example:

```typescript
const { mongo } = MongoInternals.defaultRemoteCollectionDriver();
const messages = mongo.db.collection('rocketchat_livechat_priority');

File: apps/meteor/server/startup/migrations/v289.ts [hunk 5]
```diff
@@ -0,0 +1,87 @@
...
+		await LivechatRooms.updateMany({ priorityId: { $exists: true } }, { $rename: { priorityId: 'slaId' } });
+		if (currentPriorities.length) {
+			// Since we updated the typings of the model
+			await OmnichannelServiceLevelAgreements.insertMany(currentPriorities);
+		}
+
+		// migrate old priority history messages to new sla history messages
+		await Messages.updateMany(
+			// intentionally using any since this is a legacy type which we've removed
+			{ t: 'livechat_priority_history' as any },
```
sampaiodiego, line 61: can this query update large amount of messages in large deployments? if too many, it is safer to change this to be done in batches.

apps/meteor/server/models/raw/LivechatInquiry.ts
```diff
@@ -91,4 +95,87 @@ export class LivechatInquiryRaw extends BaseRaw<ILivechatInquiryRecord> implemen
...
+	async getCurrentSortedQueueAsync({
+		inquiryId,
+		department,
+		queueSortBy,
+	}: {
+		inquiryId?: string;
+		department: string;
+		queueSortBy: OmnichannelSortingMechanismSettingType;
+	}): Promise<(Pick<ILivechatInquiryRecord, '_id' | 'rid' | 'name' | 'ts' | 'status' | 'department'> & { position: number })[]> {
+		const filter: UpdateFilter<ILivechatInquiryRecord>[] = [
```
sampaiodiego, line 108: I guess this is just `Filter<ILivechatInquiryRecord>[]` right?

apps/meteor/ee/app/livechat-enterprise/server/hooks/beforeRoutingChat.js
```diff
@@ -59,7 +59,11 @@ const beforeRouteChat = async (inquiry, agent) => {
 	saveQueueInquiry(inquiry);
 
 	if (settings.get('Omnichannel_calculate_dispatch_service_queue_statistics')) {
-		const [inq] = await LivechatInquiry.getCurrentSortedQueueAsync({ _id, department });
+		const [inq] = await LivechatInquiry.getCurrentSortedQueueAsync({
+			_id,
```
sampaiodiego, line 63: looks like `getCurrentSortedQueueAsync` doesn't accept an `_id` param 🤔  do you mean `inquiryId` instead?

apps/meteor/ee/app/livechat-enterprise/server/lib/Helper.js
```diff
@@ -301,7 +259,11 @@ export const getLivechatQueueInfo = async (room) => {
 		return null;
 	}
 
-	const [inq] = await LivechatInquiry.getCurrentSortedQueueAsync({ _id, department });
+	const [inq] = await LivechatInquiry.getCurrentSortedQueueAsync({
+		_id,
```
sampaiodiego, line 263: looks like `getCurrentSortedQueueAsync` doesn't accept an `_id` param 🤔  do you mean `inquiryId` instead?

apps/meteor/ee/app/livechat-enterprise/server/api/lib/inquiries.ts
```diff
@@ -1,25 +1,22 @@
...
-	const priorityData = priority && (await LivechatPriority.findOneByIdOrName(priority));
-	if (!priorityData) {
-		throw new Error('error-invalid-priority');
+	const slaData = sla && (await OmnichannelServiceLevelAgreements.findOneByIdOrName(sla));
+	if (!slaData) {
+		throw new Error('error-invalid-sla');
 	}
 
-	LivechatEnterprise.updateRoomPriority(roomId, await Users.findOneById(userId, { projection: { username: 1 } }), priorityData);
+	const user = await Users.findOneById(userId);
```
sampaiodiego, line 16: if you just want to check if a record exists, pls use a `projection` to fetch only the `_id` field.

apps/meteor/ee/app/livechat-enterprise/server/api/lib/priorities.ts
```diff
@@ -46,6 +39,78 @@ export async function findPriorities({ text, pagination: { offset, count, sort }
...
+	return createdResult.value;
 }
+
+export const updateRoomPriority = async (rid: string, user: IUser, priorityId: string): Promise<void> => {
+	const room = await LivechatRooms.findOneById(rid, { projection: { _id: 1 } });
+	if (!room) {
+		throw new Error('error-room-does-not-exist');
+	}
+
+	const priority = await LivechatPriority.findOneById(priorityId);
```
sampaiodiego, line 85: if you just want to check if a record exists, pls use a `projection` to fetch only the `_id` field, just like on the line above.

apps/meteor/ee/app/livechat-enterprise/server/hooks/beforeNewInquiry.ts
```diff
@@ -0,0 +1,67 @@
...
+		});
+		if (!priority) {
+			throw new Meteor.Error('error-invalid-priority', 'Invalid priority', {
+				function: 'livechat.beforeInquiry',
+			});
+		}
+	}
+
+	const now = new Date();
+	const ts = new Date(now.getTime());
```
sampaiodiego, line 46: any special reason for doing this? I guess this is exactly the same, isn't it?
```suggestion
	const ts = new Date();
```

apps/meteor/ee/app/livechat-enterprise/server/hooks/onSaveVisitorInfo.ts
```diff
@@ -0,0 +1,65 @@
...
+import { removeRoomSLA, updateRoomSLA } from '../api/lib/sla';
+import { cbLogger } from '../lib/logger';
+
+const updateSLA = async (room: IOmnichannelRoom, user: IUser, slaId?: string) => {
+	if (!slaId) {
+		await removeRoomSLA(room._id, user);
+		return;
+	}
+
+	const sla = await OmnichannelServiceLevelAgreements.findOneById(slaId);
```
sampaiodiego, line 15: if you just want to check if a record exists, pls use a `projection` to fetch only the `_id` field.

apps/meteor/ee/app/livechat-enterprise/server/lib/Helper.js
```diff
@@ -199,65 +205,17 @@ export const updateQueueInactivityTimeout = () => {
...
-
-	const { _id: priorityId } = priority;
-	LivechatRooms.findOpenByPriorityId(priorityId).forEach((room) => {
-		updateInquiryQueuePriority(room._id, priority);
+	const { _id: slaId } = sla;
+	const promises = [];
+	await LivechatRoomsRaw.findOpenBySlaId(slaId).forEach((room) => {
+		promises.push(updateInquiryQueueSla(room._id, sla));
 	});
+	await Promise.allSettled(promises.length ? promises : []);
```
sampaiodiego, line 218: this is the same right?
```suggestion
	await Promise.allSettled(promises);
```

apps/meteor/ee/app/livechat-enterprise/server/priorities.ts
```diff
@@ -0,0 +1,39 @@
...
+	},
+	{
+		i18n: 'Highest',
+		sortItem: LivechatPriorityWeight.HIGHEST,
+		dirty: false,
+	},
+];
+
+export const createDefaultPriorities = async (): Promise<void> => {
+	const priorities = await LivechatPriority.find().toArray();
```
sampaiodiego, line 34: there is no need to fetch ALL data just to check if a single record exists.. imagine how much data will be transferred and parsed just to be thrown away.. 

you have two options here:

- you can use `.countDocuments()` to count and check if the result is equal to zero
- you can query using `findOne()` without a filter using `projection: { _id: 1 }` and check if the result is not `null`.. this is the preferred method

apps/meteor/ee/server/models/raw/LivechatPriority.ts
```diff
@@ -24,4 +24,12 @@ export class LivechatPriorityRaw extends BaseRaw<ILivechatPriority> implements I
 
 		return this.findOne(query, options);
 	}
+
+	async canResetPriorities(): Promise<boolean> {
+		return Boolean(await this.findOne({ dirty: true }));
```
sampaiodiego, line 29: if you just want to check if a record exists, pls use a `projection` to fetch only the `_id` field.

sampaiodiego (changes requested):

apps/meteor/server/startup/migrations/v289.ts
KevLehman: Priorities was a not-so-used feature only available on EE, so we should be good here 👀

apps/meteor/server/startup/migrations/v289.ts
KevLehman: hmm, we copied the idea from the 278 migration 🤔

apps/meteor/ee/app/livechat-enterprise/server/hooks/beforeNewInquiry.ts
KevLehman: They had another use before the changes, so they're leftovers

apps/meteor/ee/app/livechat-enterprise/server/lib/Helper.js
KevLehman: I was probably on drugs these day

apps/meteor/ee/app/livechat-enterprise/server/priorities.ts
KevLehman: In normal cases I'd say yes. But for this case, LivechatPriority is expected to have 5 and only 5 elements, so we went this way.

We can change to a countDocuments too, but yeah, not so concerning here given the size of the collection

sampaiodiego:
lint issues?

tsconfig.base.server.json
KevLehman: So, found _why_ but not the real _why_. Seems like, for some reason, on this PR everything touching `core-typings` (which requires dom & dom.iterable) needs to have those options on build too. Still trying to find why it happens, or why it only happens on this CI tho

apps/meteor/ee/app/livechat-enterprise/server/api/lib/priorities.ts
murtaza98: In this case, we're not just checking if record exists, but also passing this variable to further where other properties within it will be used. Hence we don't need to apply projection in this case.

murtaza98 (dismissed):

murtaza98 (dismissed):

aleksandernsilva (dismissed):

filipemarins (dismissed):

apps/meteor/ee/app/livechat-enterprise/server/api/lib/priorities.ts
```diff
@@ -46,6 +39,78 @@ export async function findPriorities({ text, pagination: { offset, count, sort }
...
+			$unset: { name: 1 },
+		}) || {
+			// Trim value before inserting
+			$set: { name: data.name?.trim(), dirty: true },
+		}),
+	};
+
+	if (data.name) {
+		// If we want to enforce translated duplicates we need to change this
+		const priority = await LivechatPriority.findOne({ name: new RegExp(`^${escapeRegExp(data.name.trim())}$`, 'i') });
```
KevLehman, line 61: just fetch name

apps/meteor/ee/app/livechat-enterprise/server/hooks/onSaveVisitorInfo.ts
```diff
@@ -0,0 +1,65 @@
...
+import { OmnichannelServiceLevelAgreements } from '@rocket.chat/models';
+
+import { callbacks } from '../../../../../lib/callbacks';
+import { removePriorityFromRoom, updateRoomPriority } from '../api/lib/priorities';
+import { removeRoomSLA, updateRoomSLA } from '../api/lib/sla';
+import { cbLogger } from '../lib/logger';
+
+const updateSLA = async (room: IOmnichannelRoom, user: IUser, slaId?: string) => {
+	if (!slaId) {
+		await removeRoomSLA(room._id, user);
```
KevLehman, line 11: Return right away

apps/meteor/ee/app/livechat-enterprise/server/hooks/onSaveVisitorInfo.ts
```diff
@@ -0,0 +1,65 @@
...
+	if (!sla) {
+		throw new Error(`SLA not found with id: ${slaId}`);
+	}
+
+	await updateRoomSLA(room._id, user, sla);
+};
+
+const updatePriority = async (room: IOmnichannelRoom, user: IUser, priorityId?: string) => {
+	if (!priorityId) {
+		await removePriorityFromRoom(room._id, user);
```
KevLehman, line 25: Return right away

apps/meteor/ee/app/livechat-enterprise/server/api/lib/priorities.ts
sampaiodiego: this query should also be moved to the model class:

```ts
LivechatPriority.findOneByName(new RegExp(`^${escapeRegExp(data.name.trim())}$`, 'i'), options);
```

apps/meteor/ee/app/livechat-enterprise/server/api/lib/priorities.ts
```diff
@@ -46,6 +39,78 @@ export async function findPriorities({ text, pagination: { offset, count, sort }
...
+		if (priority && priority._id !== _id) {
+			throw new Error('error-duplicate-priority-name');
+		}
+	}
+
+	const createdResult = await LivechatPriority.findOneAndUpdate(query, update, {
```
sampaiodiego, line 67: can you create a model method for this action? just like `setSlaForRoom` and `findNextAndLock` .. ideally we should not use collection method outside the model. 

also, bringing the var definitions closer to the function call helps readability (in this case `query` and `update` and defined far away from here, where they're being used)

apps/meteor/app/authorization/server/functions/upsertPermissions.ts
```diff
@@ -117,7 +117,7 @@ export const upsertPermissions = async (): Promise<void> => {
 			_id: 'on-hold-others-livechat-room',
 			roles: ['livechat-manager', 'livechat-monitor', 'admin'],
 		},
-		{ _id: 'save-others-livechat-room-info', roles: ['livechat-manager', 'livechat-monitor'] },
+		{ _id: 'save-others-livechat-room-info', roles: ['livechat-manager', 'livechat-monitor', 'admin'] },
```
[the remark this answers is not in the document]
KevLehman: There's already a migration that will upsert permissions, so we should be good 👍🏽

apps/meteor/ee/app/livechat-enterprise/server/index.ts
```diff
@@ -34,17 +32,19 @@ import './lib/routing/LoadRotation';
...
+import { createDefaultPriorities } from './priorities';
 
-onLicense('livechat-enterprise', () => {
+onLicense('livechat-enterprise', async () => {
 	require('./api');
 	require('./hooks');
-	require('./startup');
-	const { createPermissions } = require('./permissions');
-	const { createSettings } = require('./settings');
+	await import('./startup');
```
sampaiodiego, line 40: it seems `./startup` file is being imported twice.. here and on line 15 ..

tsconfig.base.server.json
sampaiodiego: I'll give it a try without it ;)

sampaiodiego (changes requested):

apps/meteor/ee/app/livechat-enterprise/server/api/lib/priorities.ts
```diff
@@ -46,6 +39,78 @@ export async function findPriorities({ text, pagination: { offset, count, sort }
...
+			$unset: { name: 1 },
+		}) || {
+			// Trim value before inserting
+			$set: { name: data.name?.trim(), dirty: true },
+		}),
+	};
+
+	if (data.name) {
+		// If we want to enforce translated duplicates we need to change this
+		const priority = await LivechatPriority.findOne({ name: new RegExp(`^${escapeRegExp(data.name.trim())}$`, 'i') }, { name: 1 });
```
sampaiodiego, line 61: 🙈 
```suggestion
		const priority = await LivechatPriority.findOne({ name: new RegExp(`^${escapeRegExp(data.name.trim())}$`, 'i') }, { projection: { name: 1 } });
```

sampaiodiego (dismissed):

sampaiodiego (dismissed):

MartinSchoeler (dismissed):

PR #27089, opened 2022-10-18
Chore: Apply correct fuselage color tokens

ggazzo:
before we didn't have a well-defined semantic palette and with that several used the wrong values, this pr tries to adjust the vast majority, but still some are missing


apps/meteor/client/sidebar/footer/voip/VoipFooter.tsx
```diff
@@ -107,7 +107,7 @@ export const VoipFooter = ({
 								data-tooltip={holdTitle}
 								disabled={!isEnterprise}
 								icon={paused ? 'pause' : 'pause-unfilled'}
-								color={paused ? 'neutral-500' : 'info'}
+								color={paused ? 'disabled' : 'info'}
```
gabriellsh, line 110: here too.

ggazzo: info is deprecated but still works good catch

gabriellsh (approved):

PR #27095, opened 2022-10-18
[IMPROVE] Adapt apps management to use service

tapiarafael:

Move the communication folder to `app` folder and adapt app management to use service instead of the orchestrator

## Further comments

[Task AE-4](https://rocketchat.atlassian.net/browse/AE-4?atlOrigin=eyJpIjoiMDIyNjViMzA3NmUxNDUwNGFjNWM5MzRlMWEzMTA2ZjQiLCJwIjoiaiJ9)


apps/meteor/app/apps/server/communication/methods.ts
```diff
@@ -93,7 +81,7 @@ export class AppMethods {
 
 				Settings.updateValueById('Apps_Framework_enabled', false);
 
-				Promise.await(waitToUnload(instance.orch));
+				Promise.await(waitToUnload());
```
thassiov, line 84: Are those `Promise.await`s to be removed in the next PR or we will keep them for now? Because they are tied to meteor. I don't remember what the decision was about this particular case

tapiarafael: That's a good question. We must avoid using Meteor in the `service` folder specifically. But in this case, this file is in the core folder, so maybe it's ok. @d-gubert, what do you think about it?

tapiarafael: I gave a look at this file, and this `Promise.await` is inside of `Meteor.methods`, so I think that is ok to keep this for now since it will still depend on Meteor

thassiov (approved):

PR #27114, opened 2022-10-20
[NEW] Sync real names from federated users

MarcosSpessatto:

The purpose of this PR is to add the Sync for Users' real names. 

Jira: FED-12


MarcosSpessatto:
@RocketChat/chat-engine the build is broken, but I believe it's a cache issue since there is no `typecheck` error executing this locally, how can I proceed with this?

alansikora (approved):

pierre-lehnen-rc (approved):

pierre-lehnen-rc:
> @RocketChat/chat-engine the build is broken, but I believe it's a cache issue since there is no `typecheck` error executing this locally, how can I proceed with this?

I think the `@rocket.chat/presence` package is hitting a cache that it shouldn't, maybe because it is importing files from `apps/meteor`. So your changes to EventSignature are causing the service that was compiled without them to be invalid.

PR #27117, opened 2022-10-20
[FIX] Delete first message when initializing thread returns black screen

filipemarins:
## Issue(s)

Error deleting the first message when initializing a thread returns a black screen. (the thread should have only the main message)

## Steps to test or reproduce

1. Write a message in a group
2. Click "Reply In Thread" in message actions
3. Click "Delete" in the message actions for the first message within the thread view

## Further comments

The thread template was rendering the message component without checking if the data exist. 

Also added a new behavior to close the threads when doesn't exist messages in the thread anymore.

### DIFF
--- a/apps/meteor/app/threads/client/flextab/thread.ts
+++ b/apps/meteor/app/threads/client/flextab/thread.ts
@@ -42,6 +42,7 @@ type ThreadTemplateInstance = Blaze.TemplateInstance<{
 		sendToChannel: boolean;
 		jump?: string | null;
 	}>;
+	closeThread: () => void;
 	loadMore: () => Promise<void>;
 	atBottom?: boolean;
 	sendToBottom: () => void;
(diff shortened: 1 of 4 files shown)


guijun13 (approved):

apps/meteor/app/threads/client/flextab/thread.ts
```diff
@@ -141,14 +142,7 @@ Template.thread.helpers({
...
 				if (keyCode === keyCodes.ESCAPE && !result && !input?.value.trim()) {
-					const {
-						route,
-						params: { context, tab, ...params },
-					} = FlowRouter.current();
-					if (!route || !route.name) {
-						throw new Error('FlowRouter.current().route.name is undefined');
-					}
-					FlowRouter.go(route.name, params);
+					instance.closeThread();
```
ggazzo, line 145: could be possible by any way to cover this case with some test?

ggazzo (changes requested):

gabriellsh (changes requested):
Added tests are failing

ggazzo (approved):

PR #27148, opened 2022-10-28
[FIX] User email address does not change when provisioned by OAuth

### LINKED ISSUE #26868: User email address does not change when provisioned by Oauth
### Description:

Email address changes happen quite often when the users' marital status changes.

The issue is that while the email address of the user changes at the Oauth provider and gets transferred into RC in the database, the new email address does no become the main email address in RC. Only the old one is visible in the directory and any other UI elements.

### Steps to reproduce:

1. Configure custom Oauth. Set merge on and use Username as the key field.
2. Login as Oauth provisioned user normally. Then log out.
3. Change the email address of a privisioned account at the Oauth source end
4. Login again as the user
5. Check the email address in the UX

### Expected behavior:

Email address should have changed to a new one.

### Actual behavior:

The new email address has not replaced the old one.

### Server Setup Information:

- Version of Rocket.Chat Server: 5.1.2
- Operating System: RHEL8
- Deployment Method: tar
- DB Replicaset Oplog: Yes
- NodeJS Version: 14.19.3
- MongoDB Version: 4.4

### Additional context

We checked that the new email address got merged successfully in the database under user's data here:
```
	"services" : {
		"userauth" : {
                       "email" : "new email address here"
```

Yet the old email address still occupies the main email address entry here:
```
	"emails" : [
		{
			"address" : "old address here"
```

If this is a bug and not by design, it might be easiest to spot the cause for it by comparing the code between LDAP and Oauth authentication data handling. When using LDAP instead of Oauth the email addresses update properly.

carlosrodrigues94:
With this change, every time that you log in on rocket chat, your email from provider it will be updated on rocket.chat database

## Issue(s)

Fix #26868

## Steps to test or reproduce
- Configure a OAuth provider for login
- Login
- Change the email your account on provider and login again
* You can also change the previous email on the database to make the provider email on database different of the provider email, ex:  Provider email = carlos.google@gmail.com, after login  carlos.google2@gmail.com (Only on Rocketchat database).

## Further comments

TC-195


apps/meteor/app/custom-oauth/server/custom_oauth_server.js
```diff
@@ -362,6 +362,7 @@ export class CustomOAuth {
 				const update = {
 					$set: {
 						name: serviceData.name,
+						email: serviceData.email,
```
pierre-lehnen-rc, line 365: `email` is not an attribute on IUser, the email addresses are stored in the `emails` list.
It should also only be changed if the keyfield is set to username.

```suggestion
						...(this.keyField === 'username' && serviceData.email && { emails: [{ address: serviceData.email, verified: true }] }),
```

pierre-lehnen-rc (changes requested):
You'll also need to change the conditional in line 348 so it doesn't skip the whole thing if only the email is different.

matheusbsilva137 (dismissed):

apps/meteor/app/custom-oauth/server/custom_oauth_server.js
```diff
@@ -349,7 +349,8 @@ export class CustomOAuth {
 					user.services &&
 					user.services[serviceName] &&
 					user.services[serviceName].id === serviceData.id &&
-					user.name === serviceData.name
+					user.name === serviceData.name &&
+					user.email === serviceData.email
```
pierre-lehnen-rc, line 353:
```suggestion
					(this.keyField === 'email' || user.emails?.find(({address}) => address === serviceData.email))
```
`user.email` doesn't exist.

pierre-lehnen-rc (changes requested):

pierre-lehnen-rc (approved):

PR #27151, opened 2022-10-31
[FIX] Hide system messages setting not being respected.

gabriellsh:

There was a query missing the parameters in the client.

Also added a few tests to help reduce the risk of this happening again.

## Further comments

TC-141


apps/meteor/client/views/room/MessageList/components/MessageSystem.tsx
```diff
@@ -49,6 +49,7 @@ export const MessageSystem: FC<{ message: IMessage }> = ({ message }) => {
 			isSelected={isSelected}
 			data-qa-selected={isSelected}
 			data-qa='system-message'
+			data-qa-id={message.t}
```
ggazzo, line 52: data-qa-id should follow the same id rule (no duplicated one)

ggazzo (approved):

hugocostadev (approved):

PR #27158, opened 2022-11-01
[FIX] Disable/Enable Federation callbacks when the setting is turned on/off + avoid to create DMs with local users only

MarcosSpessatto:

This PR fixes 2 problems:

- The callbacks/hooks were not being added/removed whenever the setting to enable/disable the feature change, requiring a restart to take effect;
- The hook for DM creation was not validating if there is at least one external user to create external rooms, making even local DMs only being created on the proxy home server.

### DIFF
--- a/apps/meteor/app/federation-v2/server/application/AbstractFederationService.ts
+++ b/apps/meteor/app/federation-v2/server/application/AbstractFederationService.ts
@@ -18,7 +18,7 @@ export abstract class FederationService {
 		this.internalHomeServerDomain = this.internalSettingsAdapter.getHomeServerDomain();
 	}
 
-	protected async createFederatedUser(
+	protected async createFederatedUserInternallyOnly(
 		externalUserId: string,
 		username: string,
 		existsOnlyOnProxyServer = false,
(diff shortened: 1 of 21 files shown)


apps/meteor/app/federation-v2/server/index.ts
```diff
@@ -71,6 +71,17 @@ const federationMessageServiceSender = FederationFactory.buildMessageServiceSend
 
 let cancelSettingsObserver: () => void;
 
+const onFederationEnabledStatusChanged = async (isFederationEnabled: boolean): Promise<void> => {
+	federationBridge.onFederationAvailabilityChanged(isFederationEnabled);
+	if (isFederationEnabled) {
+		federationBridge.logFederationStartupInfo('Running Federation V2');
+		FederationFactory.setupActions(federationRoomServiceSender, federationMessageServiceSender);
+		require('./infrastructure/rocket-chat/slash-commands');
```
pierre-lehnen-rc, line 79: Replace this and the other requires with an await import - some of our tools don't deal well with requires.

```suggestion
		await import('./infrastructure/rocket-chat/slash-commands');
```

PR #16617, opened 2020-02-17
[IMPROVE] Send files over rest api

ggazzo:

Issue #16542, opened 2020-02-10
[BUG] File uploads in Rocket Chat are incredibly slow, but there is a simple solution for it.


recklessnl:
### Description:
See this thread for a lot more information: https://forums.rocket.chat/t/file-uploads-are-super-slow-because-theyre-sent-in-many-small-post-requests-wtf/4600

In Rocket chat, apparently the files are uploaded via many small sequential POST requests. For example, a 100KiB file (dd’d from /dev/zero) takes four requests to upload! And a 6.05mb jpeg took about 160 requets to upload. And each request takes around 300ms before the next one will be sent, causing the upload to take a stupidly long time. See this gif: https://owyf.qonq.gq/ 18

Why does it do this? It’s incredibly inefficient and causes file uploads to be super slow, especially for those with high latency.

Rocket.chat uses this for uploads https://github.com/jalik/meteor-jalik-ufs#uploading-files 8 and the uploader can be configured with custom chunk sizes as well as disabling adaptive mode which will make chunks large enough to upload at 80% of bandwidth up to a certain chunk size limit.

And this is the only use of UploadFS.Uploader i’ve found in rocket chat codebase using github search https://github.com/RocketChat/Rocket.Chat/blob/d53e264b83ccc971236123853e4b68c9eb6beeb4/app/file-upload/lib/FileUploadBase.js#L56-L67 2 and it doesn’t specify anything so looking in the code, the default chunk size is 32KiB and default max size for adaptive is 4K(i?)B (4 * 1024 * 1000). So I think I can just disable adaptive thing as it doesn’t seem to help at all and set the chunk size to whatever is the highest upload limit for the gateways (i.e. 1MB for nginx or 50MB for cloudflare) and it’ll use just one request if the file is smaller, or split it into chunks if larger, which is actually useful in this case.

**The solution would be to allow admins to set a much higher chunk size, so that upload speed can be MUCH increased.** A user (that I quoted the above from) in the forums already did it to great success. The problem is that every time snaps updates, I lose this edit. **This should be a setting in the Administration panel.**

### Steps to reproduce:
Upload any larger file on a LAN connection and observe the incredibly slow upload speeds.

### Server Setup Information:

- Version of Rocket.Chat Server: 2.4.2
- Operating System: Ubuntu
- Deployment Method: snap
- Number of Running Instances: 1 
- DB Replicaset Oplog: Enabled
- NodeJS Version: v8.17.0
- MongoDB Version:  3.6.14

[...]

PR #27105, opened 2022-10-19
[FIX] File upload receiving whole file to apply limits

sampaiodiego:
## Issue(s)

ARCH-56

PR #27168, opened 2022-11-02
feat: Support for chunked upload in rooms.upload/:rid

### LINKED ISSUE #25732: [BOUNTY] Can Not Upload Files Larger Than CDN or Proxy Single Request Limits without Chunking
Description

Since the change to send files via the REST API https://github.com/RocketChat/Rocket.Chat/pull/16617, RocketChat instances hosted behind a CDN can not upload files larger than the limit for a single request (100MB for Cloudflare)

Problem has been recognized in posts here: https://github.com/RocketChat/Rocket.Chat/pull/16617#issuecomment-591119408, https://forums.rocket.chat/t/file-uploads-are-super-slow-because-theyre-sent-in-many-small-post-requests-wtf/4600/24, and https://forums.rocket.chat/t/file-uploads-are-super-slow-because-theyre-sent-in-many-small-post-requests-wtf/4600/26

While the previous uploading library https://github.com/jalik/meteor-jalik-ufs#uploading-files had issues, it did chunk uploads.  If the upload chunk size was configurable, rather than a single POST request it would fix this issue and allow administrators to tune the setup to their network.

Steps to reproduce

Host RocketChat instance behind Cloudflare, and upload a file over 100MB

Server Setup Information

Version of Rocket.Chat Server: 3.14.3
Operating System: Ubuntu
Deployment Method: Manual install
Number of Running Instances: 4
DB Replicaset Oplog: Enabled
NodeJS Version: v12.18.4
MongoDB Version: 4.0.25

Client Setup Information 

Desktop App or Browser Version: Firefox version 100
Operating System: Any

Additional context

Relevant logs

Ousret:

Simply put, reinstate the chunked upload capability (workaround) since the last major change. This does not revert any commit.
This change does not change the actual behavior, instead, it offers two parameters that will enable admins to tweak it.

This fix applies to both single instances and multiple.

rocket patch-params

rocket patch-upload

rocket patch-upload-range

## Issue(s)

Close #25732
Related #16542 #16617

## Steps to test or reproduce

Since the change to send files via the REST API #16617, RocketChat instances hosted behind a CDN can not upload files larger than the limit for a single request.

## Further comments

The main criticism about this fix is that to provide a minimum changeset, I had to compose.
- Chunks must be sent in order (no parallelism allowed)
- Tmp file is re-read entirely in memory (last post request may be longer)


sampaiodiego:
thanks @Ousret for your contribution.. the main problem with this approach is that multiple requests can land in different rocket.chat instances, causing the upload to fail. this was on of the main problems with the previous approach, where we'd have to route the chunks to the instance that has received the first chunk in order to complete the upload.

Ousret:
Ok. I understand a bit more. There was no mention of this issue in previous issues.
Why is there no volume shared across containers or hosts? Is this why you used GridFs in the first place?

It would be up to the admin to choose the best-suited configuration.

Maybe, on first-through, add a parameter server-only, _tmp path shared_ across instances. And enforce a lock on the file.

Ousret:
Let's find an ideal compromise on that feature.
I have updated the branch to support chunk upload through GridFs, so that, when selected (in params upload) will work as expected.
Nevertheless, when filesystem is selected, it will be up to the admin to ensure that proper volume (if needed) sharing through whatever (eg. cephfs, glusterfs, swarm volumes, etc..). 

WDYT?

rocket patch-params-2

Ousret:
https://github.com/RocketChat/Rocket.Chat/pull/27105 just caused conflicts.

Ousret:
Fixed and tested locally.
I noted a minor decrease in performance (backend processing).
@sampaiodiego

Ousret:
Can someone approve the CI run?

cross-hello:
Not related with formal developer, just delegates self, תודה אתה.

----------------------------------------

*From: *TAHRI Ahmed R. ***@***.***>
*To: *RocketChat/Rocket.Chat ***@***.***>
*CC: *Subscribed ***@***.***>
*Date: *Nov 12, 2022 11:41:51
*Subject: *Re: [RocketChat/Rocket.Chat] [FIX] Support for chunked upload 
in rooms.upload/:rid (PR #27168)

cross-hello:
@sampaiodiego Please revisit if the code is suitable.
We want the feature : )

Ousret:
ping @RocketChat/chat-engine

Ousret:
PR #27457 just broke this PR.

```
import { Uploading } from '../../../../../app/ui/client/lib/fileUpload';
import { Upload } from '../../../../lib/chats/Upload';
```

sampaiodiego:
hi @Ousret , I've been discussing about this with @rodrigok and we're considering merging this, but we still have a few concerns about the implementation. We're wondering if the implementation should be simpler and always save the chunks using gridfs, this way it will covers all deployments with no extra config. wdyt?

Ousret:
Hi @sampaiodiego 

> We're wondering if the implementation should be simpler and always save the chunks using gridfs, this way it will covers all deployments with no extra config. wdyt?

I think that is a viable option, nevertheless, could be weird to see GridFS used when explicitly left aside. It works for me.
I'll leave the final say to you.

p.s. fixed the merge conflict.

Ousret:
@sampaiodiego @rodrigok did you reach a decision about excluding (or not) the fs as the chunk receiver?

Ousret:
friendly ping @sampaiodiego

rodrigok:
@Ousret, thanks for your patience, end of the year is always a crazy time.

FileSystem is, for sure, a bad idea for this purpose. It requires access to the same disk for all the instances, which leads to confusion for less experienced users. I suggest you remove this possibility for this PR and keep it on GridFS for the temp files. GridFS is not intended for the final place per scale purposes, but for a temporary location, it should be fine. We can have S3 as an option later.

In other to have this incorporated we will need the implementation on our mobile application https://github.com/RocketChat/Rocket.Chat.ReactNative. Please check and let us know when you can make the implementation.

I'll go over the code and review it ASAP, thanks 😄

Ousret:
Hi,

I understand. No pressure.

As you decided to exclude fs for the chunk temporary storage, it is done.
The PR now reflects that position.

I am checking out the react native application as I speak. 
Meanwhile, this PR is ready for review.

rodrigok:
@Ousret thanks, will review it ASAP

Ousret:
gentle ping

Ousret:
ping @sampaiodiego @rodrigok 
pr is fixed for the latest develop.

PR #27191, opened 2022-11-04
Chore: Restrict `ChatMessages` API

tassoevan:

It prepares `ChatMessages` to be splitted into smaller components to handle the interaction of a list of messages and the message composer.


ggazzo (approved):

MartinSchoeler (approved):
Everything ok in the Omnichannel part

PR #27198, opened 2022-11-05
Chore: Rewrite AutoCompleteDepartment to TypeScript

tiagoevanp:

Department field inside of current chats is one of the places who uses AutoCompleteDeparment:
image

## Further comments

OC-420


apps/meteor/ee/client/omnichannel/cannedResponses/components/cannedResponseForm.tsx
```diff
@@ -3,7 +3,7 @@ import { Box, Field, TextInput } from '@rocket.chat/fuselage';
 import { useTranslation } from '@rocket.chat/ui-contexts';
 import React, { FC } from 'react';
 
-import AutoCompleteDepartment from '../../../../../client/components/AutoCompleteDepartment';
+import { AutoCompleteDepartment } from '../../../../../client/components/AutoCompleteDepartment';
```
aleksandernsilva, line 6: Have you checked if the folder `client/components` have an index file? If so, it makes sense importing from there, if not, maybe use this opportunity to create one?

aleksandernsilva (changes requested):

apps/meteor/client/components/AutoCompleteDepartment.tsx
```diff
@@ -1,18 +1,29 @@
...
+};
 
+export const AutoCompleteDepartment = ({
+	value,
+	excludeDepartmentId,
+	onlyMyDepartments,
+	onChange,
+	haveAll,
+	haveNone,
+}: AutoCompleteDepartmentProps): ReactElement | null /* null added to satisfy register type condition */ => {
```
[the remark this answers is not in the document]
tiagoevanp: I think it is, because there is no null return as conditional at the component... but, as register.ts has a dynamic import there... it makes sense to have this null for default exports

apps/meteor/ee/client/omnichannel/cannedResponses/components/cannedResponseForm.tsx
tiagoevanp: Well, I have spoken with Tasso about it, and he told me this folder deserves a future attention on changing files to specific modules and other places... so, doing a barrel here will not be helpful for this future task at all. What I mean is, stay as it is for now...

apps/meteor/client/components/AutoCompleteDepartment.tsx
```diff
@@ -0,0 +1,90 @@
...
+				if (a.name < b.name) {
+					return -1;
+				}
+
+				return 0;
+			}),
+		[departmentsItems],
+	);
+
+	const department = useMemo(() => sortedByName.find((dep) => dep._id === findValue)?.value, [sortedByName, findValue]);
```
aleksandernsilva, line 68: I think grouping the `findValue` variable with the sorting makes more sense here, since no other code needs it.
```
const department = useMemo(() => {
    const findValue = typeof value === 'string' ? value : value.value || '');
    return sortedByName.find((dep) => dep._id === findValue)?.value, [sortedByName, findValue]);
});
```

aleksandernsilva (changes requested):

apps/meteor/client/components/AutoCompleteDepartment.tsx
tiagoevanp: ouch... in fact

apps/meteor/client/components/AutoCompleteDepartment.tsx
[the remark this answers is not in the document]
tiagoevanp: well, removed...

aleksandernsilva (approved):

apps/meteor/client/components/AutoCompleteDepartment.tsx
```diff
@@ -76,7 +76,9 @@ export const AutoCompleteDepartment = ({
 			filter={departmentsFilter}
 			// Workaround for setFilter weird typing
 			setFilter={setDepartmentsFilter as (value: string | number | undefined) => void}
-			options={sortedByName.map((dep) => ({ value: dep._id, label: dep.name }))}
+			// TODO: Fix typing on fuselage
+			// Workaround for options wrong typing
```
KevLehman, line 80: image

aleksandernsilva (approved):

aleksandernsilva (approved):

apps/meteor/client/views/omnichannel/currentChats/FilterByText.tsx
```diff
@@ -5,7 +5,7 @@ import moment from 'moment';
 import React, { Dispatch, FC, SetStateAction, useEffect } from 'react';
 
 import AutoCompleteAgent from '../../../components/AutoCompleteAgent';
-import AutoCompleteDepartment from '../../../components/AutoCompleteDepartment';
+import { AutoCompleteDepartment } from '../../../components/AutoCompleteDepartment';
```
ggazzo, line 8: why are we exporting a `memo(AutoCompleteDepartment)` and using the named export?

ggazzo (changes requested):

apps/meteor/client/views/omnichannel/currentChats/FilterByText.tsx
tiagoevanp: We have a pattern to use named export instead of default export, right? However, on `apps/meteor/ee/client/omnichannel/additionalForms/register.ts` we have a dynamic import, forcing me to sustain the default export too...

tassoevan: > We have a pattern to use named export instead of default export, right?

Wrong. https://developer.rocket.chat/fuselage/conventions#react

> Most of time we prefer using named exports, but [it] provides a more readable code when using HOCs like `memo` and `forwardRef` and it ties nicely to `lazy`.

ggazzo: believe me just import from the default (and memoized)

tiagoevanp: hmmmmmmm, ok... updating until the EOD...

tiagoevanp: sorry if I misunderstood how we deal with our code regarding our definitions. 😬
Thanks for your help on it ☺️

ggazzo (approved):

aleksandernsilva (approved):

PR #27204, opened 2022-11-07
[FIX] Disabling not supported features in federated channels

MarcosSpessatto:


alansikora (approved):

yash-rajpal (changes requested):
Shouldn't we also block "Enable E2E" room action for federated rooms?

image

yash-rajpal (approved):

PR #27212, opened 2022-11-08
fix: Failed login counter not reset after successful login

### LINKED ISSUE #26828: User blocked = "Die Anmeldung wurde für diesen Benutzer vorübergehend gesperrt"
### Description:
See also #21158
Same Problem = "Login has been temporarily blocked for this User" = "Die Anmeldung wurde für diesen Benutzer vorübergehend gesperrt"

Unfortunately, it happens several times a year that a user is blocked. Even if username and password were entered correctly (because stored in the password list)!

The user enters username, password and TOTP and then the message "Login has been temporarily blocked for this User" is displayed.

At the moment I solve the problem by playing around with 
- Change user
- Change email address
- activate/deactivate "Verify User"
- enable/disable 2FA

At some point the user works again. Unfortunately I can't exactly reproduce the process because I'm just "playing around". 

Currently I use RC 5.1.0
Application engine version 1.34.0
Node version v14.20.0
Database migration 281 (September 6, 2022 11:10)
MongoDB 5.0.11 / wiredTiger (oplog Enabled)

Whereas all this information is secondary as the problem has been there for years.
And all details are also described in #21158

A solution approach as described by @benji1000 in [#21158#issuecomment-856884323 ](https://github.com/RocketChat/Rocket.Chat/issues/21158#issuecomment-856884323) would be to aim for.
image

LucianoPierdona:
## Issue(s)

Closes #26828
## Steps to test or reproduce

- Enable failed log in block thing
- enter your password wrong the defined number of times
- wait for the defined number of minutes to be able to attempt again
- enter the wrong password again
- notice you're now blocked without trying x number of times as set in the setting.

## Further comments

[TC-183]

[TC-183]: https://rocketchat.atlassian.net/browse/TC-183?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


debdutdeb (changes requested):
This won't fix the issue. The reproducible steps are not exactly correct either 🙈 

First of all there is no "counter" exactly, using that word creates a confusion. 

Steps should be somewhat like:
1. Enable failed log in block thing
2. enter your password wrong the defined number of times
3. wait for the defined number of minutes to be able to attempt again
4. enter the wrong password again
5. notice you're now blocked without trying x number of times as set in the setting.

Now looking at the code,

Once you log in, the "pseudo counter" is "reset" correctly because the successful login event is tracked, and we only count the failed attempts **after** the successful login.

The problem lies in the check where even after the user waited for, let's say 1 minute, he/she enters the wrong password **again**, this means the last successful login event data we have in our database is still maybe the one done 1 day ago. So the count will be quite a big number. So we're not exiting where we probably should have.

Further down when we check if the attempt is valid or not by calculating the time difference , we're checking against the last failed attempt which may or may not be >`the defined wait time once blocked`. Thus, getting blocked again.

You can see this, iiuc, by let's say you set 1 minute for the blocked period, then wait 1 minute for each attempt (purposefully enter a wrong pass) even after you were blocked **at least once** - you will not be blocked again, unless, you attempt within that 1 minute.

---

Next couple comments in general, 

You are removing all failed attempts by a user or ip, what's the point of having them there as separate documents then in the first place?

Unless the events are being used some place else in our code, I think we should introduce an *actual* counter. I.e. having a single document for each ip - user combination with a count of the number of failed attempts. 

Or maybe we can use our existing sessions collection for that somehow. I'm not sure about this one.

We should probably take this opportunity to rewrite this module instead of trying to patch wounds.

apps/meteor/app/authentication/server/lib/restrictLoginAttempts.ts
```diff
@@ -78,9 +79,17 @@ export const isValidLoginAttemptByIp = async (ip: string): Promise<boolean> => {
...
+
+	if (lastBlockAt && lastBlockAt < new Date() && attemptsUntilBlock && failedAttemptsSinceLastBlock < attemptsUntilBlock) {
+		return true;
+	}
+
 	const minutesUntilUnblock = settings.get('Block_Multiple_Failed_Logins_Time_To_Unblock_By_Ip_In_Minutes') as number;
-	const willBeBlockedUntil = addMinutesToADate(new Date(lastAttemptAt), minutesUntilUnblock);
+	const willBeBlockedUntil = addMinutesToADate(new Date(), minutesUntilUnblock);
 	const isValid = moment(new Date()).isSameOrAfter(willBeBlockedUntil);
+	await saveBlockedLogin(login, willBeBlockedUntil);
```
gabriellsh, line 88: I think this could be made into it's own function and used in `isValidAttemptByUser`, since the same code is duplicated there.

apps/meteor/app/authentication/server/lib/restrictLoginAttempts.ts
```diff
@@ -121,9 +130,16 @@ export const isValidAttemptByUser = async (login: ILoginAttempt): Promise<boolea
...
+	const failedAttemptsSinceLastBlock = await ServerEvents.countFailedAttemptsByUsernameSince(user.username, new Date(lastBlockAt));
+	if (lastBlockAt && lastBlockAt < new Date() && attemptsUntilBlock && failedAttemptsSinceLastBlock < attemptsUntilBlock) {
+		return true;
+	}
+
 	const minutesUntilUnblock = settings.get('Block_Multiple_Failed_Logins_Time_To_Unblock_By_User_In_Minutes') as number;
-	const willBeBlockedUntil = addMinutesToADate(new Date(lastAttemptAt), minutesUntilUnblock);
+	const willBeBlockedUntil = addMinutesToADate(new Date(), minutesUntilUnblock);
 	const isValid = moment(new Date()).isSameOrAfter(willBeBlockedUntil);
+	await saveBlockedLogin(login, willBeBlockedUntil);
```
gabriellsh, line 139: Use said function commented before here.

apps/meteor/app/authentication/server/lib/restrictLoginAttempts.ts
```diff
@@ -121,9 +130,16 @@ export const isValidAttemptByUser = async (login: ILoginAttempt): Promise<boolea
 		return true;
 	}
 
+	const lastBlockAt = (await ServerEvents.findLastBlockByUsername(user.username as string))?.ts;
+	const failedAttemptsSinceLastBlock = await ServerEvents.countFailedAttemptsByUsernameSince(user.username, new Date(lastBlockAt));
+	if (lastBlockAt && lastBlockAt < new Date() && attemptsUntilBlock && failedAttemptsSinceLastBlock < attemptsUntilBlock) {
```
gabriellsh: `lastBlockAt < new Date()`

Please, correct me if I'm wrong, but isn't this going to be ALWAYS true, since the lastBlockAt is also a `Date()` that cannot be in the future? a `new Date()` would always be bigger than a Date set before...

gabriellsh (changes requested):

gabriellsh:
OOOPS! Requested changes on a previous commit 🙈 I'll review again

matheusbsilva137 (changes requested):

matheusbsilva137 (approved):
Great improvements! :slightly_smiling_face:

debdutdeb:
Will check out this patch ASAP. Still travelling :) should be done in a couple hours. Thanks.

matheusbsilva137 (dismissed):

debdutdeb (dismissed):
`blocked-at-*` are not exactly "events", and I'm afraid will cause an eventual semantic exhaustion.

I'm continuing the discussion internally in the backend channel.

apps/meteor/app/authentication/server/lib/restrictLoginAttempts.ts
```diff
@@ -79,14 +84,14 @@ export const isValidLoginAttemptByIp = async (ip: string): Promise<boolean> => {
 	}
 
 	const minutesUntilUnblock = settings.get('Block_Multiple_Failed_Logins_Time_To_Unblock_By_Ip_In_Minutes') as number;
```
KevLehman, line 86:
```suggestion
	const minutesUntilUnblock = settings.get<number>('Block_Multiple_Failed_Logins_Time_To_Unblock_By_Ip_In_Minutes');
```

KevLehman (changes requested):

matheusbsilva137 (approved):

loagencydev:
If this in anyway fixes the issue, then I believe it must also be merged pretty soon. we're locked out of our accounts, and even admin cannot log into the system

gabriellsh:
Closed in favor of #30018

PR #27213, opened 2022-11-08
[FIX] User auto complete breaks on enter key press

yash-rajpal:
## Issue(s)

Fixes [TC-252](https://rocketchat.atlassian.net/jira/software/c/projects/TC/boards/36?modal=detail&selectedIssue=TC-169&assignee=62b4acf799eafa602d8a0cc8)


debdutdeb:
Just a note, should this not be handled on fuselage side to accomodate enter press alongside mouse down? iiuc this will create an inconsistency where clicking shows the real name in the selection list, whereas pressing enter would the username

yash-rajpal:
You are right, we are discussing this internally.

The fuselage component used doesn't handle keyboards events, so we are discussing either to implement this or handle this some other way.

apps/meteor/client/components/UserAutoCompleteMultiple/UserAutoCompleteMultipleFederated.tsx
```diff
@@ -55,33 +55,55 @@ const UserAutoCompleteMultipleFederated = ({
...
-		const cachedOption = options.find(([curVal]) => curVal === value)?.[1];
-		if (!cachedOption) {
-			throw new Error('UserAutoCompleteMultiple - onAddSelected - failed to cache option');
-		}
-		setSelectedCache({ ...selectedCache, [value]: cachedOption });
-	};
+	const onAddSelected = useCallback(
+		([value]: OptionType, shouldRemoveUser = false): void => {
+			setFilter('');
+			if (shouldRemoveUser) {
```
gabriellsh, line 63: This doesn't really make sense inside this function. Instead of passing an additional prop, I suggest making a new function for removing from cache.

apps/meteor/client/components/UserAutoCompleteMultiple/UserAutoCompleteMultipleFederated.tsx
```diff
@@ -55,33 +55,55 @@ const UserAutoCompleteMultipleFederated = ({
 		{ keepPreviousData: true },
 	);
 
-	const options = data || [];
+	const options = useMemo(() => data || [], [data]);
```
gabriellsh, line 58: Does this make a difference? From what I know, useQuery return is already cached...

apps/meteor/client/components/UserAutoCompleteMultiple/UserAutoCompleteMultipleFederated.tsx
```diff
@@ -55,33 +55,55 @@ const UserAutoCompleteMultipleFederated = ({
...
+			}
+			setSelectedCache({ ...selectedCache, [value]: cachedOption });
+		},
+		[selectedCache, setSelectedCache, options],
+	);
+
+	const handleOnChange = useCallback(
+		(usernames: string[]) => {
+			onChange(usernames);
+			const selectedUsernames = Object.entries(selectedCache).map(([, item]) => item.username);
```
gabriellsh, line 82: instead of using selectedCache here, you could use the `value` prop which comes from outside de component. Since it's required in the type definition, should be safe enough.

Also this would make the changes to onAddSelected not needed anymore.

gabriellsh (changes requested):

gabriellsh (changes requested):

apps/meteor/client/components/UserAutoCompleteMultiple/UserAutoCompleteMultipleFederated.tsx
yash-rajpal: Yes, actually this function was earlier used at other places, so followed this pattern, forgot to remove it.

apps/meteor/client/components/UserAutoCompleteMultiple/UserAutoCompleteMultipleFederated.tsx
yash-rajpal: You are right but this was to fix a warning from eslint.

gabriellsh (approved):
LGTM! Nice work!

PR #27223, opened 2022-11-09
Chore: Make it easier to run micro services

sampaiodiego:

You can now run a microservices environment with `yarn turbo run ms`

## Issue(s)

ARCH-45


ggazzo (approved):

KevLehman (approved):
Not required, but like this 🥳

PR #27233, opened 2022-11-10
[FIX] Shrug missing backslash character

hugocostadev:
## Proposed changes (including videos or screenshots)

Previously the escape character `\` was escaping everything.

To avoid unexpected text formatting errors it was restricted to only escape curtains chars like: * _ ~ ` # . 
Doing so the Shrug had to change in the code to display the correct ASCII Art

## Issue(s)

closes https://github.com/RocketChat/Rocket.Chat/issues/26907
closes https://github.com/RocketChat/Rocket.Chat/issues/26800

## Steps to reproduce:

1. Create a message using `/shrug`.

### Expected behavior:

The kaomoji being rendered looks as follows: ` ¯\_(ツ)_/¯ `

### Actual behavior:

The kaomoji is missing an arm: ` ¯_(ツ)_/¯ `

image

## Further comments

The Shrug issue will be fixed after this Pull Request and https://github.com/RocketChat/fuselage/pull/899 are released

TC-99

PR #27234, opened 2022-11-10
[FIX] Emails as usernames dont work on mentions.

### LINKED ISSUE #26869: The new message template cannot resolve pointer usernames that contain @ character
### Description:

The new message template cannot "resolve" user names in the message pointers should they contain @ character in the target's username.

Especially the systems that use emails as their main username attribute suffer from the issue. Also those Oauth enabled instances that get the user names in format such as `username@example.com`.

In this issue example the username is called `099fnd2ee@example.com`. The real name is `Test Account`.

### Steps to reproduce:

1. Set RC to user real names by default under `Layout -> User Interface -> Use Real Name`
2.  Create account with username that contains @ character. Set the name and email separately to something else.
3. Send a message to the user that has @ character in their username

### Expected behavior:

The pointer string should show the real name (Test Account) and also have it highlighted.

### Actual behavior:

The username is shown instead of the real name and it is not highlighted.

Screenshot from 2022-09-14 15-25-14

### Server Setup Information:

- Version of Rocket.Chat Server: 5.1.2
- Operating System: RHEL8
- Deployment Method: tar
- DB Replicaset Oplog: Yes
- NodeJS Version: 14.19.3
- MongoDB Version: 4.4

### Client Setup Information

- Desktop App or Browser Version: Any browser

gabriellsh:
## Issue(s)

Fixes #26869

TC-254

### DIFF
--- a/apps/meteor/ee/server/services/package.json
+++ b/apps/meteor/ee/server/services/package.json
@@ -21,7 +21,7 @@
 		"@rocket.chat/apps-engine": "^1.32.0",
 		"@rocket.chat/core-typings": "workspace:^",
 		"@rocket.chat/emitter": "0.31.21",
-		"@rocket.chat/message-parser": "0.31.21",
+		"@rocket.chat/message-parser": "next",
 		"@rocket.chat/model-typings": "workspace:^",
 		"@rocket.chat/models": "workspace:^",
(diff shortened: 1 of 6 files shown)


ggazzo (approved):

gabriellsh:
Closed because #27233 also updates the same package, and was already merged.

PR #27251, opened 2022-11-12
[FIX] Disabling admin actions for federated users

MarcosSpessatto:

This PR disables some admin actions for federated users + count exclusively federated AND remote users only as federated users.

Jira: FED-51

## Further comments

[FED-51]: https://rocketchat.atlassian.net/browse/FED-51?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


alansikora (approved):

apps/meteor/app/models/server/models/Users.js
```diff
@@ -1006,6 +1006,7 @@ export class Users extends Base {
 			{
 				active: true,
 				federated: true,
+				username: { $regex: /.*:.*/ },
```
sampaiodiego, line 1009: do we really need to match the username as well? I mean, isn't `federated: true` enough?

performing a regex is not the most efficient thing

sampaiodiego (changes requested):

apps/meteor/app/lib/server/functions/deleteUser.ts
```diff
@@ -15,19 +16,15 @@ import { LivechatUnitMonitors } from '../../../../ee/app/models/server';
...
 	}
 
-	if (user.federation) {
-		const existingSubscriptions = Subscriptions.find({ 'u._id': user._id }).count();
-
-		if (existingSubscriptions > 0) {
-			throw new Meteor.Error('FEDERATION_Error_user_is_federated_on_rooms');
-		}
+	if (isUserFederated(user)) {
+		throw new Meteor.Error('error-user-is-federated', 'Cannot delete federated user');
```
pierre-lehnen-rc, line 27: Is this right? Feels like a weird rule.

Once a federated user is created it'll stay on the server forever? what about GDPR?

apps/meteor/app/models/server/models/Users.js
MarcosSpessatto: No, revisiting this again, it's not necessary, tks.

apps/meteor/app/lib/server/functions/deleteUser.ts
MarcosSpessatto: From what we have until now (we don't have anything integrated with Matrix regarding users' deletion) and for the scope of this PR, yes. But for sure this is not correct, we'll handle the user's deletion in this [task](https://rocketchat.atlassian.net/browse/FED-74). Thanks for the heads up about GDPR.

apps/meteor/client/views/admin/users/AdminUserInfoWithData.tsx
```diff
@@ -107,7 +108,7 @@ const AdminUserInfoWithData = ({ uid, onReload }: AdminUserInfoWithDataProps): R
 					isAdmin={data?.user.roles.includes('admin')}
 					userId={data?.user._id}
 					username={user.username}
-					isAFederatedUser={data?.user.federated}
+					isAFederatedUser={isUserFederated(data?.user as unknown as IUser)}
```
hugocostadev, line 111: Try to be consistent in naming properties and methods and don't use articles like "a and an", instead be direct like `isFederatedUser`

hugocostadev: And why it was necessary to use `as unknown` since the method expects IUser as a param?

packages/core-typings/src/IUser.ts
```diff
@@ -164,7 +164,8 @@ export interface IRegisterUser extends IUser {
 }
 
 export const isRegisterUser = (user: IUser): user is IRegisterUser => user.username !== undefined && user.name !== undefined;
-export const isUserFederated = (user: Partial<IUser>): user is IUser => 'federated' in user && user.federated === true;
+export const isUserFederated = (user: Partial<IUser>): user is IUser =>
+	'federated' in user && user.federated === true && Boolean(user.username?.includes(':'));
```
hugocostadev, line 168: Is it really necessary to check by `:`?  checking by `federated` being true is not enough?

apps/meteor/ee/client/views/admin/users/SeatsCapUsage/SeatsCapUsage.tsx
```diff
@@ -26,7 +26,7 @@ const SeatsCapUsage = ({ limit, members }: SeatsCapUsageProps): ReactElement =>
 				fontScale='c1'
 				mb='x8'
 			>
-				<div>{t('Seats_Available', { seatsLeft })}</div>
+				<div data-qa-id='seats-available'>{t('Seats_Available', { seatsLeft })}</div>
```
hugocostadev, line 29: Why include this `data-qa-id` if you are not using it? 
Also, check the e2e Wiki for learning more about selector conventions: https://github.com/RocketChat/Rocket.Chat/wiki/E2E-Testing#recommended-selector-usage

hugocostadev (changes requested):
Check the suggestions and considerations, if you don't agree or have doubts/considerations let's chat

apps/meteor/client/views/admin/users/AdminUserInfoWithData.tsx
MarcosSpessatto: Because the `data.user` is not a `IUser` type, there are some properties incompatible with `IUser` (e.g `createdAt` as string instead of Date)

apps/meteor/ee/client/views/admin/users/SeatsCapUsage/SeatsCapUsage.tsx
MarcosSpessatto: I'm not using it in this PR, but I'll use that in another PR when I create the E2E tests for the feature. I know the idea is to make the E2E tests in the same PR we are creating the feature, but this is blocked at least until we merge the PR for the federation E2E tests configuration. For the next release for sure, things will happen in the expected flow.

MarcosSpessatto: Removed data-qa btw.

packages/core-typings/src/IUser.ts
MarcosSpessatto: No, it's not, I'll remove it, thanks for pointing it out.

hugocostadev (approved):

sampaiodiego (approved):

PR #27263, opened 2022-11-14
[NEW] Endpoints for room prioritizing

cauefcr:

New endpoints: 
```
'/v1/livechat/room/:rid/priority': {
    POST: (params: POSTLivechatRoomPriorityParams) => { roomId: string; priorityId: string };
    DELETE: () => { success: boolean };
};
```

## Further comments

OC-402


apps/meteor/app/livechat/imports/server/rest/rooms.ts
```diff
@@ -70,3 +71,34 @@ API.v1.addRoute(
...
+		permissionsRequired: {
+			POST: { permissions: ['view-l-room'], operation: 'hasAny' },
+			DELETE: { permissions: ['view-l-room'], operation: 'hasAny' },
+		},
+	},
+	{
+		async post() {
+			const { rid } = this.urlParams;
+			const { priorityId } = this.bodyParams;
+			const result = LivechatRooms.setPriorityByRoomId(rid, priorityId);
```
murtaza98, line 89: maybe missing `await` here :eyes:

apps/meteor/app/livechat/imports/server/rest/rooms.ts
```diff
@@ -70,3 +71,34 @@ API.v1.addRoute(
...
+			if (!result) {
+				return API.v1.failure({ error: 'Error setting priority' });
+			}
+			return API.v1.success({ roomId: rid, priorityId });
+		},
+		async delete() {
+			const { rid } = this.urlParams;
+			const result = LivechatRooms.unsetPriorityByRoomId(rid);
```
murtaza98, line 97: and here too - missing `await` :eyes:

KevLehman (changes requested):
There's a file that decides which fields should be propagated to client when requesting it. Please add the priorityWeight there too 👀 

apps/meteor/client/startup/mergeSubscriptionsAndRooms.ts
apps/meteor/server/modules/watchers/publishFields.ts

apps/meteor/app/livechat/imports/server/rest/rooms.ts
cauefcr: it is, thanks for spotting it :see_no_evil:

apps/meteor/app/livechat/imports/server/rest/rooms.ts
cauefcr: That's what i was missing, thanks!

apps/meteor/app/livechat/imports/server/rest/rooms.ts
```diff
@@ -70,3 +71,34 @@ API.v1.addRoute(
...
 );
+
+API.v1.addRoute(
+	'livechat/rooms/:rid/priority',
+	{
+		authRequired: true,
+		validateParams: { POST: isPOSTLivechatRoomPriorityParams },
+		permissionsRequired: {
+			POST: { permissions: ['view-l-room'], operation: 'hasAny' },
+			DELETE: { permissions: ['view-l-room'], operation: 'hasAny' },
```
murtaza98, line 82: Within the tests below you're using another permission named `manage-livechat-priorities`, however, here within the definitions, I don't see that permission. TBH, I'm not really sure about the permissions of this priority feature, so I wanted to confirm if this isn't an issue

cauefcr: well, this feature is about prioritizing an existing room with existing priorities, so it's more of a managing rooms permission, not managing-livechat-priorities, but good point nevertheless.

murtaza98: Oh, I got it now. Thanks for the explanation!

murtaza98 (approved):

PR #27275, opened 2022-11-16
add support persian font

masoudit:


sampaiodiego:
@masoudit you probably opened the PR to the wrong repo :)

PR #27280, opened 2022-11-16
[NEW] Incompatible Apps

matheuslc:
This PR is introducing incompatible apps to the marketplace. 

Right now, if an app or an app version is incompatible with the current version of Rocket.Chat the user is using, we do not return them to the marketplace. We are changing that to allow users to see the apps and updates they are missing when they do not update their Rocket.Chat version.

The incompatibility is decided by looking at the Apps Engine version of the Rocket.Chat instance is running and matching with the versions of Apps Engine an app supports.

Take a look at this diagram to understand how we match versions.

versions_matching

## Issue(s)

[MKP-138](https://rocketchat.atlassian.net/browse/MKP-138)
[MKP-133](https://rocketchat.atlassian.net/browse/MKP-133)
[MKP-133](https://rocketchat.atlassian.net/browse/MKP-135)
[MKP-106](https://rocketchat.atlassian.net/browse/MKP-106)

## Steps to test or reproduce
In order to provide backward compatibility with previous versions of Rocket.Chat, we check which version of apps engine is a Rocket.Chat user using to either return the incompatibles or not. After this Incompatible feature got merged into production, we will update the minimum version on the marketplace API.

That said, to test, you need to override the Apps Engine version you are going to send to the marketplace. Change the value of appsEngineVersionForMarketplace at https://github.com/RocketChat/Rocket.Chat/blob/develop/apps/meteor/app/apps/server/communication/rest.js. For instance, change it to 1.37.0.

Now, if you look at marketplace apps list, probably all apps are incompatible and you can test the whole feature.

## Further comments

Some extra tasks included:
* Migrated AppStatus from JavaScript to TypeScript
* Added TooltipOnHover component to reuse some code
* Fixed marketplace app details page version (it was showing the installed version number)


d-gubert:
Didn't get the chance to review yet, just want to send kudos on the description of the PR ❤️

apps/meteor/client/views/admin/apps/AppDetailsPage/AppDetailsPageHeader.tsx
```diff
@@ -25,16 +39,36 @@ const AppDetailsPageHeader = ({ app }: { app: App }): ReactElement => {
...
 					</Box>
 					<Box is='span'> | </Box>
-					<Box mi='x16'>{t('Version_version', { version })}</Box>
+
+					<Box marginInlineStart={'16px'} marginInlineEnd={'4px'}>
+						{t('Version_version', { version: versioni18nKey(app) })}
+					</Box>
+
+					<Box is={'span'} marginInlineEnd={'16px'} marginInlineStart={'4px'} marginBlockStart={'4px'}>
+						{versionIncompatible === true && (
```
juliajforesti, line 58: Prefer using `marginInlineEnd={16}` or `marginInlineEnd='x16' `

apps/meteor/client/views/admin/apps/AppDetailsPage/tabs/AppStatus/TooltipOnHover.tsx
```diff
@@ -0,0 +1,29 @@
+import { Box, PositionAnimated, AnimatedVisibility } from '@rocket.chat/fuselage';
+import React, { RefObject, useRef, useState, ReactElement } from 'react';
+
+type TooltipOnHoverProps = {
+	element: ReactElement;
+	tooltip: ReactElement;
+};
+
+export const TooltipOnHover = ({ element, tooltip }: TooltipOnHoverProps): ReactElement => {
+	const ref = useRef<Element>();
```
juliajforesti, line 10: type it as:
`const ref: RefObject<Element> = useRef(null);`
to avoid the assertion on line 20.

apps/meteor/client/views/admin/apps/AppDetailsPage/tabs/AppStatus/TooltipOnHover.tsx
```diff
@@ -0,0 +1,29 @@
...
+	const [isHovered, setIsHovered] = useState(false);
+
+	return (
+		<>
+			<Box is={'span'} ref={ref} onMouseEnter={(): void => setIsHovered(true)} onMouseLeave={(): void => setIsHovered(false)}>
+				{element}
+			</Box>
+
+			<PositionAnimated
+				anchor={ref as RefObject<Element>}
```
juliajforesti, line 20: no need for this assertion if you type the ref correctly above :)

apps/meteor/client/views/admin/apps/AppMenu.js
```diff
@@ -213,14 +233,27 @@ function AppMenu({ app, ...props }) {
 		showAppPermissionsReviewModal();
 	}, [action, app.id, app.purchaseType, cancelAction, checkUserLoggedIn, isAppPurchased, setModal, showAppPermissionsReviewModal]);
 
+	const incompatibleIconName = (app, action) => {
+		if (app.versionIncompatible) {
+			return 'warning';
+		}
+
+		switch (action) {
```
juliajforesti, line 241: why switch case, instead of if? there are only 2 options...

juliajforesti (changes requested):

apps/meteor/client/views/admin/apps/AppDetailsPage/tabs/AppStatus/TooltipOnHover.tsx
matheuslc: Much better! Many thanks, @juliajforesti 👊 I've fixed it ✅

apps/meteor/client/views/admin/apps/AppMenu.js
matheuslc: That's true 🤔 I've removed it in favor of an early return. I've also removed the second parameter as it seems I wasn't using it so much.

Thanks 👊

apps/meteor/client/views/admin/apps/AppDetailsPage/AppDetailsPageHeader.tsx
rique223: Solved

rique223 (approved):
Great work on this Matheus! LGTM.

apps/meteor/app/apps/server/communication/rest.js
```diff
@@ -65,6 +66,22 @@ export class AppsRestApi {
 
 		this.api.addRoute('actionButtons', ...actionButtonsHandler(this));
 
+		this.api.addRoute(
+			'incompatibleModal',
+			{ authRequired: false },
```
graywolf336, line 71: What's the reason for making this not auth required?

apps/meteor/client/views/admin/apps/AppDetailsPage/AppDetailsPageHeader.tsx
```diff
@@ -1,18 +1,32 @@
...
 import BundleChips from '../BundleChips';
+import { appIncompatibleStatusProps } from '../helpers';
 import AppStatus from './tabs/AppStatus';
+import { TooltipOnHover } from './tabs/AppStatus/TooltipOnHover';
+
+const versioni18nKey = (app: App): string => {
+	const { version, marketplaceVersion, marketplace } = app;
+	const marketplaceCast = marketplace as boolean;
+
+	if (marketplaceCast) {
```
graywolf336, line 18: Why not check if `typeof marketplace === 'boolean' instead of casting? Casting it with typescript doesn't validate it is actually a boolean, it only convinces typescript it is a boolean and then you check if it is falsey.

matheuslc: Make sense! Fixed ✅ Thanks 👊

apps/meteor/app/apps/server/communication/rest.js
matheuslc: @graywolf336 We don't ask for the token on this modal. Should we do that? This modal doesn't execute any action, it's just read-only.

graywolf336: So, even users who are not Rocket.Chat users should be able to use this endpoint?

matheuslc: Do you mean the marketplace endpoint? Right now, you must provide a workspace id, but there's no need for the token. We just render the app info version and name into the modal, but for more safety, I'm considering adding this token as mandatory. What do you think? 🤔

graywolf336: No sir, I was talking about the endpoint in the Rocket.Chat version. By having `{ authRequired: false }`, anyone can call that endpoint even if they are not a user of that Rocket.Chat server. Which seeing the bugs and "security" issues that come in, this will be one that is brought up and I want to make sure we have a proper response if that is the intended functionality. I think it should be `{ authRequired: true }` so that the person must be a user of that rocket.chat instance. The auth required at this endpoint is on rocket.chat's side, not Marketplace's.

The token to Marketplace side shouldn't matter, in fact most users will not have access to that setting and shouldn't have access to it because then they could grab that and act on the behalf of the admins since that token is an authorized token.

d-gubert (approved):
LGTM! Cool to see this going live :)

apps/meteor/client/views/admin/apps/AppDetailsPage/AppDetailsPageHeader.tsx
```diff
@@ -25,16 +37,36 @@ const AppDetailsPageHeader = ({ app }: { app: App }): ReactElement => {
...
 						{t('By_author', { author: author?.name })}
 					</Box>
 					<Box is='span'> | </Box>
-					<Box mi='x16'>{t('Version_version', { version })}</Box>
+
+					<Box marginInlineStart='16px' marginInlineEnd='4px'>
+						{t('Version_version', { version: versioni18nKey(app) })}
+					</Box>
+
+					<Box is={'span'} marginInlineEnd='x16' marginInlineStart='x4' marginBlockStart='x4'>
```
dougfabris, line 55:
```suggestion
					<Box is='span' marginInlineEnd='x16' marginInlineStart='x4' marginBlockStart='x4'>
```

apps/meteor/client/views/admin/apps/AppDetailsPage/AppDetailsPageHeader.tsx
```diff
@@ -25,16 +37,36 @@ const AppDetailsPageHeader = ({ app }: { app: App }): ReactElement => {
...
+						{versionIncompatible === true && (
+							<>
+								<TooltipOnHover
```
dougfabris, line 58: You can remove this component and just add `title` prop in the component you want to display the tooltip

apps/meteor/client/views/admin/apps/AppDetailsPage/AppDetailsPageHeader.tsx
```diff
@@ -25,16 +37,36 @@ const AppDetailsPageHeader = ({ app }: { app: App }): ReactElement => {
...
 					{(installed || isSubscribed) && <AppMenu app={app} mis='x8' />}
 				</Box>
 				<Box display='flex' flexDirection='row' color='hint' alignItems='center'>
 					<Box fontScale='p2m' mie='x16'>
 						{t('By_author', { author: author?.name })}
 					</Box>
 					<Box is='span'> | </Box>
-					<Box mi='x16'>{t('Version_version', { version })}</Box>
+
+					<Box marginInlineStart='16px' marginInlineEnd='4px'>
```
juliajforesti, line 51: same as comment below here :)

apps/meteor/client/views/admin/apps/BundleChips.tsx
```diff
@@ -15,29 +16,20 @@ type BundleChipsProps = {
...
-					<Box ref={bundleRef} onMouseEnter={(): void => setIsHovered(true)} onMouseLeave={(): void => setIsHovered(false)}>
-						<Tag variant='primary'>{bundle.bundleName}</Tag>
-					</Box>
-					<PositionAnimated
-						anchor={bundleRef as RefObject<Element>}
-						placement='top-middle'
-						margin={8}
-						visible={isHovered ? AnimatedVisibility.VISIBLE : AnimatedVisibility.HIDDEN}
-					>
+				<TooltipOnHover
```
dougfabris, line 22: replace using just the `Tag` with `title` prop

dougfabris (changes requested):

apps/meteor/app/apps/server/communication/rest.js
matheuslc: aaah got it @graywolf336!! Thank you so much for shedding a light on that 👊 Agreed with you! The rationale behind the { authRequired: false } was a misunderstanding about that. I changed it to { authRequired: true } now ✅

ggazzo (approved):

matheuslc:
@d-gubert @rique223 We've sent some changes :( could you please review again to re-approve? Thanks!

rique223 (approved):

d-gubert (approved):

PR #27281, opened 2022-11-16
Chore: Remove role requirement to use change streams

### LINKED ISSUE #20017: Change stream does not need db admin priviledge
### Description:

Currently:
 - it is stated in the log that change stream requires db admin
 - it is stated [here](https://docs.rocket.chat/installation/manual-installation/mongo-versions) in the doc that it requires clusterMonitor

but these requirements are too much, readWrite should actually be enough.

The code defining this is here:
https://github.com/RocketChat/Rocket.Chat/blob/master/app/models/server/models/_oplogHandle.ts#L31
With this line, it actually needs clusterStatus action, this is true, but it is only needed to detect engine and version. Where as chage stream is available with readWrite role only.

To have changeStream, Actions needed are `find` and `changeStream`
https://docs.mongodb.com/manual/changeStreams/#access-control

And they are available in the read role (which is embeded in the readwrite role)
https://docs.mongodb.com/manual/reference/built-in-roles/#read

### Steps to reproduce:

1. Have a rocketchat with readWrite and oplog user with read roles 
2. start rocket chat

### Expected behavior:

It should setup and use change stream.

### Actual behavior:

Log is saying:
`Change Stream is available for your installation, give admin permissions to your database user to use this improved version`

And it uses oplog which is currently [broken](https://github.com/RocketChat/Rocket.Chat/issues/19905).

### Server Setup Information:

- Version of Rocket.Chat Server: 3.9.3
- Deployment Method: k8s
- DB Replicaset Oplog: yes
- MongoDB Version: 4.2.3-4

### Workaround

https://github.com/RocketChat/docs/pull/1853

sampaiodiego:

We used to require the `clusterMonitor` role to be able to use Change Streams because we performed a `serverStatus` command, to be able to check if the requirements for change streams are met (like using wired tiger for example). But since our minimum MongoDB supported version is now 4.2, and MMAPv1 [was removed on this version](https://www.mongodb.com/docs/v4.2/core/storage-engines/), there is no need to check that.

The new approach is just try to use change streams, if it fails for any reason, fallback to oplog.

If oplog fails as well, we're now halting the process, since it wasn't able to establish a connection for reading real time data, something that is crucial for the application. An additional check was added to make sure `$MONGO_OPLOG_URL` is pointing to the `local` database.

## Issue(s)

Closes #20017


apps/meteor/server/database/DatabaseWatcher.ts
```diff
@@ -34,69 +35,57 @@ export class DatabaseWatcher extends EventEmitter {
...
 		this._oplogHandle = _oplogHandle;
 		this.metrics = metrics;
+		this.logger = new LoggerClass('DatabaseWatcher');
 	}
 
 	async watch(): Promise<void> {
 		if (useMeteorOplog) {
+			// TODO remove this when updating to Meteor 2.8
+			this.logger.warn(
+				'Using USE_NATIVE_OPLOG=true is currently desencouraged due to known performance issues. Please use IGNORE_CHANGE_STREAM=true instead.',
```
geekgonecrazy, line 59:
```suggestion
				'Using USE_NATIVE_OPLOG=true is currently discouraged due to known performance issues. Please use IGNORE_CHANGE_STREAM=true instead.',
```

geekgonecrazy:
Just for context before this.. frequently people who start using permissions with mongo but missed this permission start encountering reactivity issues that look like Rocket.Chat is buggy.  

But then as soon as they added it things started working fine.  One common one was connectivity services, but also anything that depended on change stream to populate setting cache.  

The permission is only needed for something informational that provides little value.  So requiring is costing way more than we lose by removing it.

IMO for customer experience this is a must for the release.  If a release wasn't coming I'd argue for a patch that's how impactful it is.

PR #27283, opened 2022-11-16
Chore: Restrict `ChatMessages` API - Phase 2

tassoevan:

It split `ChatMessages` service class into a collection of smaller objects, in a composition pattern fashion. Additionally, it loosens its dependency on the DOM tree and assumes almost all methods are asynchronous.

## Further comments

- The naming convention is a little wanky; it might change.
- Initially this API was redesigned as "reactivity friendly" -- i.e. queries over state should have a pair of functions `get` and `subscribe` -- but the effort in refactoring some complex mutations took place in the schedule. The next phase might address this issue.
- The `flows` are the most dependent and complex behaviors. A flow might depend on another flow, trigger UI effects, and handle exceptions by themselves.


ggazzo:
👀

ggazzo:
🕐

tassoevan:
> 🕐

✋

Issue #27257, opened 2022-11-14
Problem with displaying non-FQDN links in the new message template


shevchukma:
### Description:

If you use the new message template, links with hostname are not displayed as links, but as text.
![image](https://user-images.githubusercontent.com/104189570/201660710-32f05e11-9e3d-48ac-acfb-7f6d17f174ba.png)
In the old message template, non-FQDN links display fine.
![image](https://user-images.githubusercontent.com/104189570/201660905-f396ccf3-8fc4-4034-ba4e-96f746e80979.png)

### Steps to reproduce:

1. Go to "My Account -> Preference -> Global".
2. Disable "Use legacy message template" setting.
3. Write a message with the link http://site
4. Link does not become active

### Expected behavior:

The new message template converts non-FQDN links to clickable ones.

### Actual behavior:

The new message template leaves non-FQDN links as text.

### Server Setup Information:

- Version of Rocket.Chat Server: 5.3.0
- Operating System: Oracle Linux 8.6
- Deployment Method: docker
- Number of Running Instances: 1
- DB Replicaset Oplog: no
- NodeJS Version: 14.19.3
- MongoDB Version: 5.0.13

### Client Setup Information

- Desktop App or Browser Version: Chromium 103
- Operating System: Debian 11

### Additional context

We will try to avoid using non-FQDN links, but for compatibility reasons, please return support for them.

ankar84:
I guess @hugocostadev and @gabriellsh already working on that problem.

gabriellsh:
Yep, here's a merged PR for it: https://github.com/RocketChat/fuselage/pull/901

This will only fix when using the `[title](link)` pattern though. We have some limitations that we're currently trying to overcome regarding our parser, but we didn't want to leave this issue unattended, so we did what's possible for now.

Issue #27141, opened 2022-10-27
AutoLinker Phone overly aggressive


nmagedman:
### Description:
The AutoLinker Phone code matches any plus sign (`+`) followed by one or more digits, regardless whether it is a valid phone number.

### Steps to reproduce:

1. Go to `/admin/Settings/Message` > AutoLinker and ensure that **Enabled** and **AutoLinker Phone** are both enabled.
2. In a chat, enter the text `1+1=2`

### Expected behavior:

The chat message should not contain any hyperlinks

### Actual behavior:

The `+1` is hyperlinked to `tel:1` even though it is not a valid phone number.

<img width="289" alt="Screen Shot 2022-10-27 at 11 15 25" src="https://user-images.githubusercontent.com/1042298/198229614-de598a4d-1a3e-4f49-b181-c7327d99dcd7.png">

Here's another sample:
![AutoLinker Number](https://user-images.githubusercontent.com/1042298/198229285-e2b8e5cc-4e03-4215-b1d6-8c64d516c3bb.png)

### Server Setup Information:

- Version of Rocket.Chat Server: 5.1.2
- Operating System: Linux
- Deployment Method: tar
- Number of Running Instances: 5
- DB Replicaset Oplog: 
- NodeJS Version: v14.19.3
- MongoDB Version: 4.2.15 / wiredTiger (oplog Enabled)

### Client Setup Information

- Desktop App or Browser Version: Chrome 106.0.5249.119
- Operating System: MacOS

### Additional context
* This problem has actually been around for a long time.  It appears to have started over a year ago when we upgraded from RC 3.12.1 to 3.16.3

    #### Most recent correct behavior
    ```
    > db.rocketchat_message.find({msg:/\+\d/, md:{$exists:false}}).sort({$natural:-1}).pretty()
    ```
    Message includes `28.5/(84.5+28.5)` which was left as-is.  If I type this same text now, it links the `+28` to `tel:28`

    #### First sighting of the problem
    ```
    > db.rocketchat_message.find({msg:/\+\d/, ts:{$gt:ISODate("2021-07-31T22:48:53.089Z")}}).pretty()
    ```
    Message includes `+3.5%` which is linked as `tel:3`

    We had upgraded RC during the hour in between these two messages.

* This feature is not currently disableable due to bug #27135

hugocostadev:
Hi, @nmagedman thank you for your time! We will address this internally and improve in our message-parser package this behavior

Issue #27065, opened 2022-10-14
Markdown Support Schemes for Link option doesn't work with new parser template


ankar84:
### Description:

Markdown Support Schemes for Link option works perfect with legacy parser template, but not work with new message parser.

### Steps to reproduce:

1. Go to Admin UI - Message - Markdown Support Schemes for Link and add `notes` scheme protocol.
2. Write in any chat this line `[notes link](notes://Server/C3257116002CAD60/0/CCAF6BE2824A1F49432588D2001FA73E)`
3. With legacy message parser that line will parse as clickable link but with new parser it will just text.

### Expected behavior:

That Markdown Support Schemes for Link option must work same with new parser as it work in legacy parser.

### Actual behavior:

Admin UI - Message - Markdown
![IMG_20221014_083627.jpg](https://user-images.githubusercontent.com/4023037/195742096-01f309af-ea5b-446e-8e1b-c23bc4a753df.jpg)

But with new parser notes and e1c protocols not parse as a links
![1](https://user-images.githubusercontent.com/4023037/148885772-5bed5fc5-1676-4bcc-9f1f-af9fdac6ad9b.png)
![2](https://user-images.githubusercontent.com/4023037/148885813-37d2ec6a-2abe-4013-9c32-6b4a0522d590.png)

### Server Setup Information:

- Version of Rocket.Chat Server: 4.8.6
- Operating System: Oracle Linux 8.5
- Deployment Method: docker
- Number of Running Instances: 20
- DB Replicaset Oplog: Enabled 
- NodeJS Version: v14.18.3
- MongoDB Version: 4.4.16

### Client Setup Information

- Desktop App or Browser Version: 3.8.11
- Operating System: Windows 10

### Additional context

This is a new parser issue. And that is really critical for us.
@hugocostadev asked me to create a separate issue about that problem

ankar84:
5.3.0 still not fixed

hugocostadev:
Hi @ankar84 your new parser is too permissive, you can use any protocol and domain.

I think that  your problem is not because of the setting not being honored (that indeed it is not), I tested here and figure out that your issue it's because of that you are using a File Path instead of a URI inside the Link markdown `[Link Label](URI_HERE)`

Several settings only work with the legacy parser, we will clean up several settings to avoid confusion 

But I understand your use case and we will discuss this internally if make sense to add this behavior to our message-parser.

ankar84:
> Hi @ankar84 

[...]

PR #27284, opened 2022-11-16
[FIX] Message Parser version upgrade

hugocostadev:

This version includes some fixes to the message-parser package about the phone number and link highlighting 

## Issue(s)

 
fix https://github.com/RocketChat/Rocket.Chat/issues/27257
fix https://github.com/RocketChat/Rocket.Chat/issues/27141
fix https://github.com/RocketChat/Rocket.Chat/issues/27065

## Further comments

TC-255
TC-250


cuonghuunguyen:
can we have the option to allow all link schema? We use a lot of schema on our side

hugocostadev:
> can we have the option to allow all link schema? We use a lot of schema on our side

Hi @cuonghuunguyen could you elaborate, please?

Currently we accept all protocol and domain schemas,  now with this PR we are enabling file path schema with this pattern: `[Label](FILE_PATH)`

cuonghuunguyen:
do you mean without the option Allowed schemes for link?

hugocostadev:
> do you mean without the option Allowed schemes for link?

Yes, currently these options work only with the Legacy parser, is the next releases we will make this clear and soon we will remove those options together with the legacy parser

gabriellsh (approved):

PR #27291, opened 2022-11-17
[NEW] Create stream for priority updates

KevLehman:
## Further comments

OC-413

Note: Collection was made "registered" on CE, even when the collection is EE only.
There was some issues in how to register EE only handlers to stream-hub service, and move the complete registration was easier than any other option.
No logic is moved to CE, only collection registration.

This will also fix an issue with the 2 migrations created by the priorities feature, which were assuming LivechatPriority collection will be there (as well as SLA).


murtaza98 (approved):

KevLehman:
Closing for now

PR #27315, opened 2022-11-22
[FIX] Add reaction name in reaction list of a message right after user react it

liv1n9:

Add user's full name to reaction's name list of message before sending it back to client via `afterSetReaction` so that user can see his/her name in reaction list right after reacting.

## Issue(s)
Currently after reacting, user can only see his/her username in reaction list.

## Steps to test or reproduce
1. User react a message
2. User click on message's options and choose to see reaction list
3. User only see his/her username in reaction list instead of his/her full name (requires refreshing client to reload message so that their full names can be loaded)

## Further comments

Only add name to reaction's name list after calling `setReaction` so that name list won't be saved to database


debdutdeb:
We had a short discussion about this internally, and not sure if we want to do it like this or use the normalizeMessage function.

liv1n9:
> 

So i will close this PR for now

PR #27323, opened 2022-11-23
[BREAK] update LivechatPriority data representation at disk, adds individual priority PUT

cauefcr:
## Further comments

OC-299


apps/meteor/ee/app/livechat-enterprise/server/api/lib/priorities.ts
```diff
@@ -37,11 +37,22 @@ export async function findPriority({
...
+	_id: string,
+	data: Pick<ILivechatPriority & { reset: boolean }, 'name' | 'reset'>,
+): Promise<ILivechatPriority> {
 	const query = {
 		_id,
 	};
-
-	const created = await LivechatPriority.findOneAndUpdate(query, { $set: { ...data, dirty: true } }, { returnDocument: 'after' });
+	if (data.reset) {
+		data.name = '$.defaultValue';
```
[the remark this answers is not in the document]
cauefcr: no, it is not, afaik they're translated on the frontend

apps/meteor/ee/app/livechat-enterprise/server/api/lib/priorities.ts
```diff
@@ -37,11 +37,19 @@ export async function findPriority({
...
+export async function updatePriority(
+	_id: string,
+	data: Pick<ILivechatPriority & { reset: boolean }, 'name' | 'reset'>,
+): Promise<ILivechatPriority> {
 	const query = {
 		_id,
 	};
-
-	const created = await LivechatPriority.findOneAndUpdate(query, { $set: { ...data, dirty: true } }, { returnDocument: 'after' });
+	const update: { $set: Record<string, boolean>; $unset?: Record<string, 1 | ''> } = { $set: { dirty: !data.reset } };
```
KevLehman, line 47: Can you use `Pick<UpdateOperation<ILivechatPriority>, '$set'>` 👀

apps/meteor/ee/app/livechat-enterprise/server/api/lib/priorities.ts
```diff
@@ -37,11 +37,19 @@ export async function findPriority({
...
+	if (data.reset) {
+		update.$unset = { name: 1 };
+	}
+	const created = await LivechatPriority.findOneAndUpdate(query, update, {
+		returnDocument: 'after',
+	});
 	return created.value as ILivechatPriority;
```
KevLehman, line 54: 🤔 this shouldn't be needed.

apps/meteor/ee/app/models/server/models/LivechatPriority.js
```diff
@@ -7,9 +7,8 @@ export class LivechatPriority extends Base {
 	constructor() {
 		super('livechat_priority');
 
-		this.tryEnsureIndex({ name: 1 }, { unique: true });
+		this.tryEnsureIndex({ name: 1 });
```
KevLehman, line 10: Please tell alek how to remove the unique index on their DB. Don't think we need a migration since we're not in develop yet

apps/meteor/tests/end-to-end/api/livechat/12-priorites.ts
```diff
@@ -276,6 +257,40 @@ import { IS_EE } from '../../../e2e/config/constants';
...
+				})
+				.expect('Content-Type', 'application/json')
+				.expect(200);
+			expect(response.body).to.have.property('success', true);
+			expect(response.body.priorities).to.be.an('array');
+			expect(response.body.priorities).to.have.length.greaterThan(0);
+			expect(response.body.priorities[0]).to.have.property('_id');
+			expect(response.body.priorities[0]).to.have.property('i18n', priority.i18n);
+		});
+		it('should reset a single priority with a reset:true PUT parameter', async () => {
```
KevLehman, line 275: Create a test trying to "reset" a priority sending reset as false

packages/rest-typings/src/v1/omnichannel.ts
```diff
@@ -2776,19 +2776,28 @@ export const isGETDashboardsAgentStatusParams = ajv.compile<GETDashboardsAgentSt
...
-};
+type PUTLivechatPriority =
+	| {
+			name?: string;
+			reset: boolean;
+	  }
+	| {
+			name: string;
+			reset?: boolean;
+	  };
```
KevLehman, line 2787: 🤔 wouldn't be better
```
{ name: string } | { reset: true }

File: apps/meteor/server/startup/migrations/v283.ts [hunk 7]
```diff
@@ -12,5 +12,11 @@ addMigration({
 			// ignore
 			console.log(error);
 		}
+		try {
+			await LivechatPriority.col.dropIndex('name_1');
+		} catch (error) {
+			// ignore
+			console.log(error);
+		}
```
KevLehman, line 20: Don't think we need this, since the change is not in develop (it won't even run on our dev machines since migration 283 should have already run)

cauefcr: name has to stop being a unique field, or a lack of name will break the index

KevLehman: Yeah, What I'm saying is that we don't need this index removal since the index is not created outside Alek, Yours and my machine 👀 

The only reason would be if QA envs have the index created, like if they reuse the DB, but not sure of it. Can we confirm that? 👀

apps/meteor/ee/app/livechat-enterprise/server/api/lib/priorities.ts
cauefcr: UpdateOperation does not exist here, but UpdateFilter works.

apps/meteor/ee/app/models/server/models/LivechatPriority.js
cauefcr: alright, removing this change

apps/meteor/ee/app/livechat-enterprise/server/api/lib/priorities.ts
cauefcr: you're right, that was a latent bug.

packages/rest-typings/src/v1/omnichannel.ts
cauefcr: not a good idea, it breaks the type checking on the endpoint, went with {name?: string; reset?: boolean;}

apps/meteor/ee/app/livechat-enterprise/server/api/lib/priorities.ts
```diff
@@ -37,11 +37,25 @@ export async function findPriority({
 	};
 }
 
-export async function updatePriority(_id: string, data: Pick<ILivechatPriority, 'name'>): Promise<ILivechatPriority> {
+export async function updatePriority(
+	_id: string,
+	data: Pick<ILivechatPriority & { reset: boolean }, 'name' | 'reset'>,
```
KevLehman, line 42: `data: Pick<ILivechatPriority & { reset: boolean }, 'name' | 'reset'>`
change for
`data: Pick<ILivechatPriority, 'name'> & { reset: boolean }`

apps/meteor/ee/app/livechat-enterprise/server/api/lib/priorities.ts
```diff
@@ -37,11 +37,25 @@ export async function findPriority({
...
+		}),
+	};
+
+	const created = await LivechatPriority.findOneAndUpdate(query, update, {
+		returnDocument: 'after',
+	});
 
-	const created = await LivechatPriority.findOneAndUpdate(query, { $set: { ...data, dirty: true } }, { returnDocument: 'after' });
-	return created.value as ILivechatPriority;
+	return created.value;
```
KevLehman, line 60: You can put an `if (!created) throw Error('invalid-priority')` or smth :eyes:

apps/meteor/ee/app/livechat-enterprise/server/api/priorities.ts
```diff
@@ -54,12 +54,16 @@ API.v1.addRoute(
...
-			const { name } = this.requestParams();
+			const { name, reset } = this.requestParams();
 
-			await updatePriority(priorityId, {
+			const ok = await updatePriority(priorityId, {
 				name,
+				reset: Boolean(reset),
 			});
 
+			if (!ok) {
```
KevLehman, line 64: Move this check to inside `updatePriority`

packages/rest-typings/src/v1/omnichannel.ts
```diff
@@ -2790,19 +2790,19 @@ export const isGETDashboardsAgentStatusParams = ajv.compile<GETDashboardsAgentSt
...
+type PUTLivechatPriority = { name?: string; reset?: boolean };
 
 const PUTLivechatPrioritySchema = {
 	type: 'object',
 	properties: {
 		name: {
 			type: 'string',
 		},
+		reset: {
+			type: 'boolean',
```
KevLehman, line 2802: Pls use what we discussed here about unions on AJV

KevLehman (changes requested):
Pls check comments

apps/meteor/ee/app/livechat-enterprise/server/api/priorities.ts
```diff
@@ -54,12 +54,16 @@ API.v1.addRoute(
 		},
 		async put() {
 			const { priorityId } = this.urlParams;
-			const { name } = this.requestParams();
+			const { name, reset } = this.requestParams();
 
-			await updatePriority(priorityId, {
+			const ok = await updatePriority(priorityId, {
 				name,
+				reset: Boolean(reset),
```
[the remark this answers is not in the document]
cauefcr: so undefined becomes false

apps/meteor/ee/app/livechat-enterprise/server/priorities.ts
```diff
@@ -43,6 +33,6 @@ export const createDefaultPriorities = async (): Promise<void> => {
 	const priorities = await LivechatPriority.find().toArray();
 
 	if (!priorities.length) {
-		await LivechatPriority.insertMany(defaultPriorities);
+		console.error(await LivechatPriority.insertMany(defaultPriorities));
```
KevLehman, line 36: Won't this print the insert everytime?

packages/rest-typings/src/v1/omnichannel.ts
```diff
@@ -2807,19 +2807,33 @@ export const isGETDashboardsAgentStatusParams = ajv.compile<GETDashboardsAgentSt
 	GETLivechatAnalyticsDashboardsAgentStatusParamsSchema,
 );
 
-type PUTLivechatPriority = {
-	name: string;
-};
+type Without<T, U> = { [P in Exclude<keyof T, keyof U>]?: never };
+type XOR<T, U> = T | U extends object ? (Without<T, U> & U) | (Without<U, T> & T) : T | U;
 
+type PUTLivechatPriority = XOR<{ name: string }, { reset: boolean }>;
```
KevLehman, line 2813: 🤔 do we need this type?

KevLehman (changes requested):
Just 2 smol things, the rest looks good

apps/meteor/ee/app/livechat-enterprise/server/priorities.ts
cauefcr: oof, forgot the debug print

packages/rest-typings/src/v1/omnichannel.ts
cauefcr: yes, it's the only proper way i found to type these parameters.

KevLehman: What's the problem with `{ name } | { reset }` and using type guards at the endpoint to know which of the types to use?

cauefcr: the de-structuring  was the problem, as we talked.

KevLehman (approved):

aleksandernsilva (approved):

PR #27329, opened 2022-11-23
[FIX] Remove unnecessary permissions check on app permissions review modal

rique223:
## Jira task: [MKP-166](https://rocketchat.atlassian.net/browse/MKP-166?atlOrigin=eyJpIjoiYzdjNzVhMThkOThjNGEzMmIyYTE2YmY2ZDA1MzA4YjEiLCJwIjoiaiJ9)
Removed an unnecessary check that verified if the app permissions list was an array and threw an error if it wasn't. This check has become unnecessary ever since the app permission review modal changed to show a default list of permissions in case the app happened to not have one and that's what caused those "random" undefined toasts that showed every once in a while.
Demo gif:
undefined_toast

## Steps to test or reproduce

1. Access the admin section and then the apps option
2. Install any app that has a 27 item(as of this PR) list of permissions
## Further comments

[MKP-166]: https://rocketchat.atlassian.net/browse/MKP-166?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


tiagoevanp (changes requested):
Rewrite the file to TS asserting expected types, maybe this could avoid future errors regarding what is expected from an object property... 🤔

PR #27360, opened 2022-11-25
[FIX] Allow multiple remplacement of {username} on external avatar url

ralmn:
Allow multiple remplacement of `{username}` on external avatar url


ggazzo:
why this is a fix?

ralmn:
Our webservice to retrieve the photos of our employees asks twice for the username in parameter which is currently not possible.

PR #27377, opened 2022-11-28
Chore: Deprecate meteor methods used by SLAs and replace with APIs

murtaza98:
## Issue(s)

[OC-426](https://rocketchat.atlassian.net/browse/OC-426)


apps/meteor/ee/app/livechat-enterprise/server/methods/saveSLA.js
```diff
@@ -2,15 +2,19 @@ import { Meteor } from 'meteor/meteor';
 
 import { hasPermission } from '../../../../../app/authorization';
 import { LivechatEnterprise } from '../lib/LivechatEnterprise';
+import { methodDeprecationLogger } from '../../../../../app/lib/server/lib/deprecationWarningLogger';
 
 Meteor.methods({
 	'livechat:saveSLA'(_id, slaData) {
+		methodDeprecationLogger.warn(
```
KevLehman, line 9: Same, since this will be merged in 6, you can remove the method and references

apps/meteor/ee/client/omnichannel/slaPolicies/RemoveSlaButton.tsx
```diff
@@ -11,18 +11,14 @@ type Props = {
...
 	const t = useTranslation();
 	const slaRoute = useRoute('omnichannel-sla-policies');
 
 	const handleRemoveClick = useMutableCallback(async () => {
-		try {
-			await removeSLA(_id);
-		} catch (error) {
-			console.log(error);
-		}
+		await removeSLA();
```
KevLehman, line 21: Wasn't the try/catch required here? Does useEndpont and useMethod handle errors differently?

apps/meteor/ee/client/omnichannel/slaPolicies/SlaEdit.tsx
```diff
@@ -53,16 +54,16 @@ function SlaEdit({ data, isNew, slaId, reload, ...props }: SlaEditProps): ReactE
 	const handleSave = useMutableCallback(async () => {
 		const { name, description, dueTimeInMinutes } = getValues();
 
-		if (!isValid) {
+		if (!isValid || !name || dueTimeInMinutes === undefined) {
```
KevLehman, line 57: Shouldn't this validations be handled in the form instead of the onSave? @MartinSchoeler maybe you can help here :eyes:

apps/meteor/tests/data/livechat/priorities.ts
```diff
@@ -1,17 +1,17 @@
 import faker from '@faker-js/faker';
 import { IOmnichannelServiceLevelAgreements } from '@rocket.chat/core-typings';
-import { credentials, methodCall, request } from '../api-data';
+import { api, credentials, methodCall, request } from '../api-data';
 import { DummyResponse } from './utils';
 
-export const saveSLA = (): Promise<IOmnichannelServiceLevelAgreements> => {
+export const saveSLAWithDeprecatedMeteorMethod = (): Promise<IOmnichannelServiceLevelAgreements> => {
```
KevLehman, line 6: If you remove the method, remove this helper too

apps/meteor/tests/end-to-end/api/livechat/12-priorites.ts
```diff
@@ -48,9 +49,42 @@ import { IS_EE } from '../../../e2e/config/constants';
 			expect(current).to.have.property('dueTimeInMinutes', sla.dueTimeInMinutes);
 			await deleteSLA(sla._id);
 		});
+		// POST
+		it('should return an "unauthorized error" when the user does not have the necessary permission', async () => {
```
KevLehman, line 53: Can you separate these on different describes? one for [GET] livechat/sla and the other for [POST]

KevLehman (dismissed):
Pls check comments

apps/meteor/ee/client/omnichannel/slaPolicies/RemoveSlaButton.tsx
murtaza98: Since we're now using the `useEndpoint()` here, we don't have to manually handle the API errors, because those hooks have their own mechanisms for error handling.

apps/meteor/ee/client/omnichannel/slaPolicies/SlaEdit.tsx
murtaza98: I believe there's already a validation over there however I did this just to avoid typescript errors, because, below typescript was still resolving it to be undefined.

apps/meteor/ee/app/livechat-enterprise/server/lib/LivechatEnterprise.js
```diff
@@ -133,18 +133,17 @@ export const LivechatEnterprise = {
...
-	saveSLA(_id, slaData) {
+	async saveSLA(_id, slaData) {
 		check(_id, Match.Maybe(String));
 
 		check(slaData, {
 			name: String,
 			description: Match.Optional(String),
-			dueTimeInMinutes: String,
+			dueTimeInMinutes: Number,
 		});
```
KevLehman, line 143: This should be already validated by the endpoint, so we can skip double validation here 👀

KevLehman (dismissed):

apps/meteor/ee/app/livechat-enterprise/server/lib/LivechatEnterprise.js
murtaza98: Ouch, I forgot about that. I'd kept it before because the meteor method was calling this same function so these validations would be used by it, however forgot to remove them while removing those methods. Let me do it now

Issue #967, opened 2015-10-02
Emoticons broken with a question mark at the end


ninja-:
Testcase:
1. type ":D?"
2. expected is a rendered emoticon
3. there is no emoticon rendered unlike ":D" and ":D!"

S4KH:
Hi guys, I am new, is there any local test users created beforehand?

gmsecrieru:
Based on [this comment](https://github.com/Ranks/emojione/issues/137#issuecomment-147127026) it looks like this is an expected behavior and your PR will not be merged. 

Closing as this is not an issue with RocketChat itself.

PR #27385, opened 2022-11-28
[FIX] “All Categories” dropdown menu shaking

felipe-rod123:

- “All Categories” dropdown menu is shaking when opened. 
- To solve this, I created a custom style inside another `<div>` in the `apps/meteor/client/views/admin/apps/components/DropDownListWrapper.tsx` component file.
- I also created a custom hook to check for screen width changes, and apply the correct style.

Update-Dropdown

## Steps to test or reproduce

1. Go to the workspace admin panel
2. Click on the “apps” menu
3. Click on the filter “All Categories”.

Old-Dropdown

## Further comments
Other suggestions to fix this error without CSS are welcome!


rique223 (approved):

dougfabris:
@felipe-rod123 Closing this one as the core of the problem was solved here #967

PR #27405, opened 2022-11-29
[IMPROVE] SAML and OAuth role sync to support id and name

LucianoPierdona:

This PR allows the sync of roles by either `name` or `id`.

## Further comments

[TC-278]

[TC-278]: https://rocketchat.atlassian.net/browse/TC-278?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/meteor/ee/server/configuration/saml.ts
```diff
@@ -21,7 +23,27 @@ onLicense('saml-enterprise', () => {
...
+
+			const roles = ensureArray<string>(value)
+				.map((role) => {
+					const savedRole = savedRoles.find((r) => r._id === role) ?? savedRoles.find((r) => r.name === role);
+
+					if (!savedRole) return '';
+
+					return savedRole.name;
+				})
+				.filter(Boolean);
```
matheusbsilva137, line 44:
```suggestion
			const roles = ensureArray<string>(value)
				.filter((role) => {
					const savedRole = savedRoles.find((r) => r._id === role) ?? savedRoles.find((r) => r.name === role);
					return savedRole;
				});
```
I think you could do this with a single filter.

apps/meteor/ee/server/lib/oauth/Manager.ts
```diff
@@ -50,9 +50,31 @@ export class OAuthEEManager {
...
+
+			const allowedRoles = rolesToSync
+				.map((role) => {
+					const savedRole = roles.find((r) => r._id === role) ?? roles.find((r) => r.name === role);
+
+					if (!savedRole) return '';
+
+					return savedRole._id;
+				})
+				.filter(Boolean);
```
matheusbsilva137, line 73: Same as the other comment. Maybe this could be replaced by a single filter.

matheusbsilva137 (changes requested):

apps/meteor/ee/server/configuration/saml.ts
LucianoPierdona: This filter does not fix the purpose of the task, which is to add support for the user to input both the id or the name in the role field.

apps/meteor/ee/server/lib/oauth/Manager.ts
```diff
@@ -50,9 +50,12 @@ export class OAuthEEManager {
 				user.roles = [];
 			}
 
+			const savedRoles = Roles.findInIdsOrNames(rolesToSync).toArray();
+			const allowedRoles = savedRoles.map((role) => role._id);
```
matheusbsilva137, line 54:
```suggestion
			const savedRoles = Promise.await(Roles.findInIdsOrNames(rolesToSync));
			const allowedRoles = Promise.await(savedRoles.toArray()).map((role) => role._id);
```
I think we're just missing some awaits here

matheusbsilva137 (changes requested):

apps/meteor/ee/server/lib/oauth/Manager.ts
```diff
@@ -50,9 +50,12 @@ export class OAuthEEManager {
 				user.roles = [];
 			}
 
+			const savedRoles = Promise.await(Roles.findInIdsOrNames(rolesToSync));
+			const allowedRoles = Promise.await(savedRoles.toArray()).map((role) => role._id);
+
 			Promise.await(
 				syncUserRoles(user._id, rolesFromSSO, {
```
matheusbsilva137, line 57:
```suggestion
			const rolesIdsFromSSO = Promise.await(Roles.findInIdsOrNames(rolesFromSSO).toArray()).map((role) => role._id);
			const allowedRoles = Promise.await(Roles.findInIdsOrNames(rolesToSync).toArray()).map((role) => role._id);

			Promise.await(
				syncUserRoles(user._id, rolesIdsFromSSO, {
```
We should send the roles' ids as params to the `syncUserRoles` function. This works

matheusbsilva137 (changes requested):
With the changes I suggested, this is already working for updates (when an existing user is synced on login, for example). However, roles are not being mapped correctly on user creation
We should change the `afterValidateNewOAuthUser` callback implementation to correctly map roles in this case

apps/meteor/ee/server/lib/oauth/Manager.ts
LucianoPierdona: Updated!

matheusbsilva137 (approved):

PR #27443, opened 2022-12-03
Regression: fix small issues before release

ggazzo:
is_uploading missing translations
register username form bug
login form error not reseting properly


ggazzo:
there's no way this pull request could be marked for 5.4.1 because it was released in the release candidate, precisely to allow us to make the release without any known bugs

PR #27453, opened 2022-12-05
fix: Emoji avatar for new message template

yash-rajpal:

image

## Issue(s)

TC-105


yash-rajpal:
Closing in favor of #28416

PR #27471, opened 2022-12-06
[IMPROVE] New marketplace routes

rique223:
## Jira task: [MKP-151](https://rocketchat.atlassian.net/browse/MKP-151?atlOrigin=eyJpIjoiZWRiYTAzMGJiMDQyNGFlNThmMzk5ZjUwMzk2NGM4Y2IiLCJwIjoiaiJ9)
Small refactor of the marketplace routes and bug fixes:
1. Changed the all context to explore;
2. Added a redirect for people that try and access marketplace from the old route(Bookmars for example)
3. Fixed a bug in which the apps engine disabled fallback would flicker because it would go back and fort between the fallback and the explore page;
4. Added the proper tab prop to the AppRow redirect function;

## Further comments

[MKP-151]: https://rocketchat.atlassian.net/browse/MKP-151?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/meteor/client/views/marketplace/AppDetailsPage/AppDetailsPageHeader.tsx
```diff
@@ -48,7 +48,7 @@ const AppDetailsPageHeader = ({ app }: { app: App }): ReactElement => {
 					</Box>
 					<Box is='span'> | </Box>
 
-					<Box mi='x16' marginInlineEnd='x4'>
+					<Box mi='x16' marginInlineEnd='x16'>
```
juliajforesti, line 51:
```suggestion
					<Box mi='x16'>
```
`mi='x16' is already setting the margin inline **start** and **end** to `x16`

apps/meteor/client/views/marketplace/MarketplaceRouter.tsx
```diff
@@ -7,13 +7,20 @@ import PageSkeleton from '../../components/PageSkeleton';
...
 
 	useEffect(() => {
-		if (!currentContext) {
-			marketplaceRoute.replace({ context: 'all', page: 'list' });
-		}
-	}, [currentContext, marketplaceRoute]);
+		const initialize = async () => {
+			// The currentContext === 'all' verification is for users who bookmarked
+			// the old marketplace
+			if ((!currentContext && (await isAppsEngineEnabled())) || currentContext === 'all') {
```
juliajforesti, line 17: Why is the verification of `isAppsEngineEnabled` needed for one condition (`!currentContext`) and not needed for the second one (`currentContext === 'all'`) ?

Also, is it really necessary to include those users who bookmarked? This pendency will be here forever..?

juliajforesti (changes requested):

apps/meteor/client/views/marketplace/MarketplaceRouter.tsx
rique223: First part: Oversight on my side;
Second part: I believe it is necessary since marketplace doesn´t really have fallbacks for wrong routes as of yet and I don´t think is a nice experience for our users to browse blank pages even if they are easy to get out of, wdyt?

juliajforesti: Maybe add a TO DO comment, to remove this once the fallback is implemented?

rique223: Absolutely, Ill add this and then solve these changes, thanks for the input Julia.

tiagoevanp (approved):

PR #27481, opened 2022-12-07
[BREAK] Freeze Custom Roles

LucianoPierdona:

This PR moves role creation and update to enterprise folder

--
- Deprecated method `authorization:saveRole`
- Moved `/v1/roles.create/` and `/v1/roles.update` to enterprise edition
- Changed message for delete a custom role in community edition

## Further comments

[TC-282]

[TC-282]: https://rocketchat.atlassian.net/browse/TC-282?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


matheusbsilva137 (dismissed):

matheusbsilva137 (approved):

gabriellsh (approved):

gabriellsh (approved):

rodrigok (approved):

rodrigok:
@LucianoPierdona great PR. Just need to adjust the description since this removes the method rather than deprecating it

danel1:
@rodrigok @gabriellsh Are you seriously removing the whole permissions / roles management for all "free" customers....?

rodrigok:
Hi @danel1 

From version 6.0 the Community Edition won't have the ability to create new Custom Roles. Existing Custom Roles will continue working but won't be able to be updated (change name or description), if deleted, they can't be recreated. So the effective change here is the movement of the API to create Custom Roles to the Enterprise Edition.

The ability to set permissions on native Roles, and the Custom Roles that already existed before the upgrade, will continue to be available in Community Edition.

In fact, the only change here is related to Custom Roles, not to the rest of the roles/permission system.

ulope:
This also breaks the option to enable or disable the `Users must use Two Factor Authentication` setting. 

But sine that setting is broken anyway I guess who the fuck cares.

PR #27483, opened 2022-12-07
Chore: Add missing indexes to SLA collection

KevLehman:
## Further comments

OC-451


murtaza98 (approved):

apps/meteor/ee/server/models/raw/ServiceLevelAgreements.ts
```diff
@@ -9,6 +9,13 @@ export class ServiceLevelAgreements extends BaseRaw<IOmnichannelServiceLevelAgre
 		super(db, 'omnichannel_service_level_agreements');
 	}
 
+	protected modelIndexes(): IndexDescription[] {
+		return [
+			{ key: { name: 1 }, unique: true },
+			{ key: { dueTimeInMinutes: 1 }, unique: true },
```
murtaza98, line 15: I'm assuming since this SLA feature is going to be deprecated as is, we're not creating any special error handlers for these indexes :eyes:

KevLehman: Which special error handlers would you create? 👀

PR #27531, opened 2022-12-13
[NEW] Implement Queue service

KevLehman:
## Further comments

OC-496
OC-497
OC-498

### DIFF
--- a/apps/meteor/server/sdk/types/IQueueWorkerService.ts
+++ b/apps/meteor/server/sdk/types/IQueueWorkerService.ts
@@ -1,3 +1,10 @@
+export type HealthAggResult = {
+	total: number;
+	type: string;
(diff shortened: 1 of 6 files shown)


apps/meteor/server/sdk/types/IQueueWorkerService.ts
```diff
@@ -1,3 +1,10 @@
+export type HealthAggResult = {
+	total: number;
+	type: string;
+	status: 'Rejected' | 'In progress';
```
murtaza98, line 4: Can we make this enum pls? Just to avoid typo's

KevLehman: typos would produce a type error anyways, but i can make it 👀

PR #27539, opened 2022-12-14
[FIX] Unread Message count not displayed for new messages in Omni-Rooms

murtaza98:
## Issue(s)

[OC-411](https://rocketchat.atlassian.net/browse/OC-411)
## Further comments

- [x] Add tests to verify if Omni-rooms subscription's `unread` prop is getting updated on a new message

[OC-411]: https://rocketchat.atlassian.net/browse/OC-411?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/meteor/tests/end-to-end/api/livechat/00-rooms.ts
```diff
@@ -1446,4 +1447,30 @@ describe('LIVECHAT - rooms', function () {
 			expect(response.body.filters.find((f: IOmnichannelRoom['source']) => f.type === 'api')).to.not.be.undefined;
 		});
 	});
+
+	describe('it should mark room as unread when a new message is sent', () => {
+		let room: IOmnichannelRoom;
+		let visitor: ILivechatVisitor;
+		let totalMessagesSent = 2;
```
murtaza98, line 1454:
```suggestion
		let totalMessagesSent = 0;
```

KevLehman (approved):

murtaza98:
Hi @RocketChat/chat-engine, Just a quick reminder - this PR is waiting for review. If you have a sec, can you check it out and let us know what you think? Thanks in advance :hugs:

apps/meteor/app/lib/server/lib/notifyUsersOnMessage.js
```diff
@@ -106,7 +107,8 @@ export async function updateUsersSubscriptions(message, room) {
 		}
 
 		// this shouldn't run only if has group mentions because it will already exclude mentioned users from the query
-		if (!toAll && !toHere && unreadCount === 'all_messages') {
+		// and this should always run if it's a omnichannel room
+		if (!toAll && !toHere && (unreadCount === 'all_messages' || isOmnichannelRoom(room))) {
```
sampaiodiego, line 111: so with this change, omnichannel rooms will always overwrite the `Unread_Count` setting value, therefore there will be no way of changing the behavior via configuration.. is that ok? what about creating a new setting `Unread_Count_Omni` to control this behavior? just an idea to be considered.

apps/meteor/tests/end-to-end/api/livechat/00-rooms.ts
```diff
@@ -1446,4 +1447,30 @@ describe('LIVECHAT - rooms', function () {
...
 		});
 	});
+
+	describe('it should mark room as unread when a new message is sent', () => {
+		let room: IOmnichannelRoom;
+		let visitor: ILivechatVisitor;
+		let totalMessagesSent = 0;
+		let departmentWithAgent: Awaited<ReturnType<typeof createDepartmentWithAnOnlineAgent>>;
+
+		before(async () => {
```
sampaiodiego, line 1457: no big deal, but instead of using `before` I feel this should be part of the test it self =)

I mean, you should use `before` to prepare something to be used in the next tests cases (`it`), but this before looks *very specific*, that's why.

or maybe only the `updateSetting` should be in the `before` statement.. or even you can create an `after` statement to clean everything up (like removing the department, room, etc)

sampaiodiego (changes requested):

apps/meteor/app/lib/server/lib/notifyUsersOnMessage.js
murtaza98: That's a good point! I think we should make it configurable just like we're doing for rest of the rooms. Updated the PR and added this new setting. Please check :smiley:

apps/meteor/tests/end-to-end/api/livechat/00-rooms.ts
murtaza98: Makes sense. Moved that bunch of operations to a separate `it` hook

sampaiodiego (approved):

PR #27543, opened 2022-12-14
[NEW] Upload service

KevLehman:
## Further comments

This is taken from AppsEngine teamwork. Idea is to have this merged into develop faster as Omnichannel needs this same service :) and we don't want to fix conflicts


apps/meteor/server/services/upload/service.ts
```diff
@@ -0,0 +1,36 @@
...
+		});
+		return uploadedFile;
+	}
+
+	async sendFileMessage({ roomId, file, userId, message }: ISendFileMessageParams): Promise<IMessage | undefined> {
+		let msg;
+		Meteor.runAsUser(userId, () => {
+			msg = Meteor.call('sendFileMessage', roomId, null, file, message);
+		});
+		return msg;
```
sampaiodiego, line 30:
```suggestion
		return Meteor.runAsUser(userId, () => Meteor.call('sendFileMessage', roomId, null, file, message));
```

apps/meteor/server/services/upload/service.ts
```diff
@@ -0,0 +1,36 @@
...
+export class UploadService extends ServiceClassInternal implements IUploadService {
+	protected name = 'upload';
+
+	async uploadFile({ buffer, details, userId }: IUploadFileParams): Promise<IUpload> {
+		let uploadedFile: IUpload = {} as IUpload;
+		Meteor.runAsUser(userId, () => {
+			const fileStore = FileUpload.getStore('Uploads');
+			uploadedFile = fileStore.insertSync(details, buffer);
+		});
+		return uploadedFile;
```
sampaiodiego, line 22: this doesn't need to be a `runAsUser`.. the confusion might be because `insertSync` requires a Fiber, but just regular `insert` can be used as well

```suggestion
		const fileStore = FileUpload.getStore('Uploads');
		return await fileStore.insert(details, buffer);
```

sampaiodiego (changes requested):

apps/meteor/server/services/upload/service.ts
```diff
@@ -0,0 +1,39 @@
...
+		return fileStore.insert(details, buffer);
+	}
+
+	async sendFileMessage({ roomId, file, userId, message }: ISendFileMessageParams): Promise<IMessage | undefined> {
+		return Meteor.runAsUser(userId, () => Meteor.call('sendFileMessage', roomId, null, file, message));
+	}
+
+	async sendFileLivechatMessage({ roomId, visitorToken, file, message }: ISendFileLivechatMessageParams): Promise<IMessage> {
+		return Meteor.call('sendFileLivechatMessage', roomId, visitorToken, file, message);
+	}
```
KevLehman, line 27: Don't want to sound like one of those service purists, but, should we have this "sendmessage" actions on this service? It's easier yeah, but 🤔 @sampaiodiego

apps/meteor/server/services/upload/service.ts
```diff
@@ -0,0 +1,39 @@
...
+	IUploadFileParams,
+	IUploadService,
+} from '../../sdk/types/IUploadService';
+import { FileUpload } from '../../../app/file-upload/server';
+
+export class UploadService extends ServiceClassInternal implements IUploadService {
+	protected name = 'upload';
+
+	async uploadFile({ buffer, details }: IUploadFileParams): Promise<IUpload> {
+		const fileStore = FileUpload.getStore('Uploads');
```
KevLehman, line 17: If this is an "uploads" service, we may have this as a property instead of a string. So we call `this.store`

apps/meteor/server/services/upload/service.ts
sampaiodiego: I think so..

just not sure if this service should also save the message, or just somehow associate the file with the message.

apps/meteor/server/services/upload/service.ts
sampaiodiego: once we move everything to this service, yes, agreed. right now the "stores" are created async, which then requires us from using `getStore` every time.. it is not something expensive to call `getStore` but I agree it makes sense to have it as a local variable inside the service

sampaiodiego (approved):

apps/meteor/server/services/upload/service.ts
KevLehman: To not remove your approval, I'll get this done for the future 👀

apps/meteor/server/services/upload/service.ts
KevLehman: I'd say it would be good for the future to have a "messaging" service and do those actions there. 🤔 I'll resolve this one, but will create tasks for updating this in the future.
Maybe using this call to proxy messages to reduce changes

PR #27551, opened 2022-12-15
[FIX] Anonymous user buttons

### LINKED ISSUE #27149: anonymous message writing does not work
### Description:

Anonymous message writing doesn't work, because the 'Sign in to start talking' and 'Or talk as anonymous' buttons dont do anything.

### Steps to reproduce:
Enable anonymous read/write under settings/accounts
Access server while logged out
Navigate to a public and default room
Click the 'Sign in to start talking' and 'Or talk as anonymous' buttons

### Expected behavior:
'Sign in to start talking' would prompt you to log in
'Or talk as anonymous' button would give you a prompt to create an anonymous login token

### Actual behavior:
The buttons do nothing

### Server Setup Information:

- Version of Rocket.Chat Server: 5.2.0
- Operating System: Debian 10 Buster 64 bit
- Deployment Method: Tar
- NodeJS Version: 14.20.1 - x64
- MongoDB Version: 6.0.2

### Client Setup Information

- Desktop App or Browser Version: Firefox 92 (64-bit)
- Operating System: Ubuntu 20.04.3 LTS

### Additional context
Consider the following file: https://github.com/RocketChat/Rocket.Chat/blob/develop/apps/meteor/client/views/room/components/body/composer/ComposerAnonymous.tsx

From the git history it kindof looks like while re-organizing someone commented out the click handlers for the two buttons in question because they cause a compilation failure, and never fixed/reimplemented them, so they in fact do nothing

### Relevant logs:
If you really want to see logs I can provide them, however I currently have a custom version deployed to my server since I tried re-building with the code un-commented (to be clear, i did NOT have a custom version deployed when I first noticed this issue), and then made various efforts to fix it myself (it is not as simple as just un-commenting the code, and I am not a typescript/react/javascript whatever guy so I was unable to figure it out after several hours).  It takes like an hour to redeploy everything and get you a version of the logs from your own production release and I currently believe I already mostly identified the issue so I will not bother for now.

hugocostadev:

When the server accepts an anonymous user to join and chat without login, two buttons appears: one for Login and the other to send messages without login.

The component was rewritten to use the fuselage Design System and native hooks to perform actions.

image

## Issue(s)

closes #27149 

fixes https://forums.rocket.chat/t/sign-in-to-start-talking-and-or-talk-as-anonymous-buttons-dont-work/15358 

## Further comments

TC-318


apps/meteor/tests/e2e/anonymous-user.spec.ts
```diff
@@ -0,0 +1,51 @@
...
+import { setSettingValueById } from './utils/setSettingValueById';
+import { expect, test } from './utils/test';
+
+test.describe('anonymous-user', () => {
+	let poHomeChannel: HomeChannel;
+
+	test.beforeAll(async ({ api }) => {
+		expect((await setSettingValueById(api, 'Accounts_AllowAnonymousRead', true)).status()).toBe(200);
+		expect((await setSettingValueById(api, 'Accounts_AllowAnonymousWrite', true)).status()).toBe(200);
+	});
```
ggazzo, line 13:
```suggestion
	test.beforeAll(async ({ api }) => {
		expect((await setSettingValueById(api, 'Accounts_AllowAnonymousRead', true)).status()).toBe(200);
		expect((await setSettingValueById(api, 'Accounts_AllowAnonymousWrite', true)).status()).toBe(200);
	});
	test.afterAll(async ({ api }) => {
		expect((await setSettingValueById(api, 'Accounts_AllowAnonymousRead', false)).status()).toBe(200);
		expect((await setSettingValueById(api, 'Accounts_AllowAnonymousWrite', false)).status()).toBe(200);
	});
```

gabriellsh (dismissed):

apps/meteor/client/hooks/useAnonymousUser.ts
```diff
@@ -0,0 +1,9 @@
+import { useMethod } from '@rocket.chat/ui-contexts';
+import { useMutation } from '@tanstack/react-query';
+
+export const useAnonymousUser = () => {
```
tassoevan, line 4: The hook's name is not very informative. Why not `useRegisterAnonymousUserMutation`? It adds a verb and it declares the return type is not a simple callback.

tassoevan, line 4: The hook's name is not very informative. Why not `useRegisterAnonymousUserMutation`? It adds a verb and it declares the return type is not a simple callback.

tassoevan (changes requested):

apps/meteor/client/hooks/useAnonymousUser.ts
hugocostadev: done

tassoevan (approved):

PR #27552, opened 2022-12-15
Chore: Update tsconfig base target to es2020

sampaiodiego:

The base files are not split into two other files, a "client" and a "server" file. Both extends the new base one. Packages should extend either "client" or "server" ones:

```mermaid
graph TD;
    base[tsconfig.base.json]-->client[tsconfig.base.client.json];
    base-->server[tsconfig.base.server.json];
    server-->s1["@rocket.chat/models"];
    server-->s2["@rocket.chat/ddp-streamer"];
    server-->s3[...];
    client-->c1["@rocket.chat/ui-theming"];
    client-->c2["@rocket.chat/fuselage-ui-kit"];
    client-->c3[...];
```

Important changes:

- The client config `target` have been bumped to `es2015` (it was `es5` before)
- The server config `target` have been bumped to `es2020`
- The default value for `jsx` is `react-jsx`
- Raised `"noUnusedLocals": true` to a base config, so had to remove a bunch of imports to `React` that were not being used


ggazzo (changes requested):
create a client base, sever base and base base :)

ggazzo (approved):

PR #27559, opened 2022-12-15
[FIX] Allow federated rooms with the same name per servers

MarcosSpessatto:

The main purpose for this PR is to allow the room creation with the same name, but from different servers, for example:
`room1`  from `matrix.org`
`room1`  from `my-matrix.server`
`room1` from `my-matrix2.server`
...

We are still enforcing to have only one room with a specific name per server, for example:
`room1` from `my-local-server.com`
`room1` from `my-local-server.com` <= This is not allowed.


alansikora (approved):

apps/meteor/server/models/raw/Rooms.js
```diff
@@ -309,6 +329,7 @@ export class RoomsRaw extends BaseRaw {
 				$in: groupsToAccept,
 			},
 			name: nameRegex,
+			federated: { $exists: false },
```
pierre-lehnen-rc, line 332: In my local instance all my private groups had `federated: false`; This change made me unable to add any of them to a team.
I don't know why that flag exist on them, but I assume it was used by some other feature at some point.

pierre-lehnen-rc (changes requested):
I stopped reviewing when I found a possible bug, will review the rest later.

apps/meteor/server/models/raw/Rooms.js
MarcosSpessatto: It's indeed a bug, the front end was always sending the federated property, even if it was false. But besides that, I should also add the validation to cover that case. It was fixed in the next commit. Thanks for noticing that.

hugocostadev (approved):

pierre-lehnen-rc (approved):

PR #27561, opened 2022-12-15
[FIX] IPostMessageReacted return field named isReacted instead isRemoved

### LINKED ISSUE #27497: [BUG] IPostMessageReacted returns field named `isRemoved` instead of `isReacted`
### Description:
- AppsEngine have `IPostMessageReacted`, **Handler for after a message get a reaction or a reaction got removed** . has method `executePostMessageReacted` which takes one of the parameter of type `IMessageReactionContext` which includes **isReacted** field to know whether the emoji has been added or deleted. But everytime if we react with emoji in messages instead of having a `boolean` value in **isReacted** field, we are getting `undefined` regardless the operation we are making ( add or delete ).

- Reason for having value as `undefined` everytime is from RocketChat, we are returning the field named `isRemoved` instead of `isReacted`. [Have a look over here](https://github.com/RocketChat/Rocket.Chat/blob/b12e64c661b2ecfa09e07ccab1cdcd38b79cd439/apps/meteor/app/apps/server/bridges/listeners.js#L88). But Expected field should have name `isReacted` as per the [docs](https://rocketchat.github.io/Rocket.Chat.Apps-engine/interfaces/messages_imessagereactioncontext.imessagereactioncontext.html).

### Steps to reproduce:

1. Implement `IPostMessageReacted` in RocketChat App.
2. log `isReacted` field in console.
3. Deploy the app and React any message with emoji and check console.

`OR` 

- Set up this [playground](https://github.com/Nabhag8848/RC-Apps-Snippets/tree/playground) app to avoid the step 1 and follow rest of steps.

### Expected behavior:

- Instead of having `undefined` in `isReacted` we are expecting it to be `boolean` value as to know the emoji operation within message (add or delete).

### Actual behavior:
- As we can see, getting isReacted field as `undefined`
isReactedField undefined

### Note: 
- [x] willing to work on this.

Nabhag8848:
## Proposed changes

- AppsEngine have `IPostMessageReacted`, Handler for after a message get a reaction or a reaction got removed . has method `executePostMessageReacted` which takes one of the parameter of type `IMessageReactionContext` which includes isReacted field to know whether the emoji has been added or deleted. But *Previously* everytime if we react with emoji in messages instead of having a `boolean` value in isReacted field, we were getting `undefined`. Reason for having value as `undefined` everytime previously was from RocketChat, we are returning the field named isRemoved instead of isReacted,  which is fixed.

## Issue(s)

- closes #27497


apps/meteor/app/apps/server/bridges/listeners.js
```diff
@@ -85,7 +85,7 @@ export class AppListenerBridge {
 						message: msg,
 						user: this.orch.getConverters().get('users').convertToApp(userReacted),
 						reaction,
-						isRemoved,
+						isReacted: isRemoved,
```
tapiarafael, line 88: I suggest changing the name of the variable being destructured in the parameters array, to `isReacted` to maintain the code cleaner

tapiarafael (changes requested):

apps/meteor/app/apps/server/bridges/listeners.js
Nabhag8848: Thanks @tapiarafael. will make the required changes.

Nabhag8848: Hey @tapiarafael, made the neccessary changes as per suggestion you gave.

tapiarafael (approved):
LGTM

debdutdeb (approved):

PR #27562, opened 2022-12-16
[IMPROVE] Update room.saveInfo endpoint to save priority info

murtaza98:

Changes required to save priority info along with other room info on this screen:
image

## Issue(s)

[OC-515](https://rocketchat.atlassian.net/browse/OC-515)


apps/meteor/tests/end-to-end/api/livechat/00-rooms.ts
```diff
@@ -1442,8 +1443,66 @@ describe('LIVECHAT - rooms', function () {
 				.expect('Content-Type', 'application/json')
 				.expect(400);
 		});
+		(IS_EE ? it : it.skip)('should update room priority', async () => {
```
aleksandernsilva, line 1446: Can't you just use `test.skip` at the beginning of the file, like we do in the FE?

aleksandernsilva (approved):

apps/meteor/tests/end-to-end/api/livechat/00-rooms.ts
murtaza98: I don't think so, that seems to be something that playwright offers however couldn't find something similar on mocha

KevLehman: Yeah, mocha is not that good. 
On the other hand, making it a global handler would be good: `eeDescribe` or `eeIt` (naming could be better obvs)

but that's something OOS for this task

KevLehman (approved):

PR #27568, opened 2022-12-16
[NEW] PDF Worker service

KevLehman:
## Further comments

OC-500


ee/apps/pdf-worker/src/templates/transcriptTemplate.tsx
```diff
@@ -0,0 +1,291 @@
+import React from 'react';
```
KevLehman, line 1: Ignore most of this file, since it's a dummy template to check if processing works

ee/apps/pdf-worker/src/templates/transcriptTemplate.tsx
```diff
@@ -0,0 +1,291 @@
...
+					</View>
+				</div>
+			</View>
+		</Page>
+	</Document>
+);
+
+export default async (data: any): Promise<NodeJS.ReadableStream> => {
+	return ReactPDF.renderToStream(<OmnichannelTranscript {...{ data }} />);
+};
```
KevLehman, line 291: This is the only interesting part, which is the function the service will eventually call on the template. This is the part that actually renders the file into an stream, then we consume the stream and upload the buffer :)

apps/meteor/server/models/raw/LivechatRooms.js
```diff
@@ -1288,4 +1288,39 @@ export class LivechatRoomsRaw extends BaseRaw {
 			},
 		]);
 	}
+
+	// These 3 methods shouldn't be here :( but current EE model has a meteor dependency
+	// And refactoring it could take time
+	setTranscriptRequestedPdfById(rid) {
```
murtaza98, line 1294: Can you re-use this one? At least based on the imports it doesn't seem to have meteor dependency 
https://github.com/RocketChat/Rocket.Chat/blob/7440e030f1551515b4ec29ff5fd8a6eed894d4a1/apps/meteor/ee/server/models/raw/LivechatRooms.ts#L26

apps/meteor/server/services/messages/service.ts
```diff
@@ -0,0 +1,25 @@
...
+import type { IMessageService } from '../../sdk/types/IMessageService';
+import { createDirectMessage } from '../../methods/createDirectMessage';
+import { executeSendMessage } from '../../../app/lib/server/methods/sendMessage';
+
+export class MessageService extends ServiceClassInternal implements IMessageService {
+	protected name = 'message';
+
+	async createDirectMessage({ to, from }: { to: string; from: string }): Promise<{ rid: string }> {
+		const toUser = await Users.findOneByUsername(to);
+		const fromUser = await Users.findOneByUsername(from);
```
murtaza98, line 14:
```suggestion
		const [toUser, fromUser] = await Promise.all([ Users.findOneByUsername(to), await Users.findOneByUsername(from) ]);
```

apps/meteor/server/services/translation/service.ts
```diff
@@ -0,0 +1,32 @@
...
+	}
+
+	// Use translateText when you already know the language, or want to translate to a predefined language
+	translateText(text: string, targetLanguage: string): Promise<string> {
+		return Promise.resolve(TAPi18n.__(text, { lng: targetLanguage }));
+	}
+
+	// Use translate when you want to translate to the user's language, or server's as a fallback
+	async translate(text: string, user: IUser): Promise<string> {
+		const language = user.language || (await this.getServerLanguage()) || 'en';
```
murtaza98, line 22:
```suggestion
		const language = user.language || (await this.getServerLanguage());
```

Since `this.getServerLanguage()` already has a default fallback

apps/meteor/server/services/translation/service.ts
```diff
@@ -0,0 +1,32 @@
...
+import type { IUser } from '@rocket.chat/core-typings';
+
+import { ServiceClassInternal } from '../../sdk/types/ServiceClass';
+import type { ITranslationService } from '../../sdk/types/ITranslationService';
+
+export class TranslationService extends ServiceClassInternal implements ITranslationService {
+	protected name = 'translation';
+
+	private async getServerLanguage(): Promise<string> {
+		return ((await Settings.findOneById('Language'))?.value as string) || 'en';
```
murtaza98, line 12: Can we cache this somewhere because if the room has 100 messages, then this will be called 100 times? Not sure if the `mem` package is now compatible with services (I hope it is :crossed_fingers: )

ee/apps/omnichannel-transcript/package.json
```diff
@@ -19,6 +19,7 @@
 		"@rocket.chat/emitter": "0.31.22",
 		"@rocket.chat/model-typings": "workspace:^",
 		"@rocket.chat/models": "workspace:^",
+		"@rocket.chat/pdf-worker2": "workspace:^",
```
murtaza98, line 22:
```suggestion
		"@rocket.chat/pdf-worker": "workspace:^",
```

murtaza98: This would also require renaming other files :see_no_evil: but I don't think the suffix makes sense here

ee/apps/omnichannel-transcript/src/OmnichannelTranscript.ts
```diff
@@ -1,16 +1,252 @@
...
+	}
+
+	private async getFiles(
+		userId: string,
+		messages: IMessage[],
+	): Promise<(Pick<IMessage, '_id' | 'ts' | 'u' | 'msg'> & { files: ({ name?: string; buffer: Buffer | null } | undefined)[] })[]> {
+		return Promise.all(
+			messages.map(async (message: IMessage) => {
+				if (!message.attachments || !message.attachments.length) {
+					// If there's no attachment and no message, what was sent? lol
```
murtaza98, line 90: AFAIK, there's a third case where UiKit blocks are sent, like for example when you interact with a chatbot that has quick reply buttons with messages. In this case, its data would be stored inside `blocks` property of message. [Here's](https://github.com/RocketChat/Apps.WhatsApp/blob/634aee83ec5f35add30bf5ff600e4f5f820cffe6/helpers/message.ts#L539) a parser for these block from one of the apps

apps/meteor/ee/app/livechat-enterprise/server/api/transcript.ts
```diff
@@ -0,0 +1,38 @@
+import { LivechatRooms } from '@rocket.chat/models';
+
+import { API } from '../../../../../app/api/server';
+import { canAccessRoomAsync } from '../../../../../app/authorization/server/functions/canAccessRoom';
+import { OmnichannelTranscript } from '../../../../../server/sdk';
+
+// This is a public route for testing purposes, this should be updated before merging into develop
```
murtaza98, line 7:
```suggestion
```
Still required?

apps/meteor/ee/app/livechat-enterprise/server/permissions.ts
```diff
@@ -22,5 +23,6 @@ export const createPermissions = async (): Promise<void> => {
 		Permissions.create('manage-livechat-canned-responses', [adminRole, livechatManagerRole, livechatMonitorRole]),
 		Permissions.create('spy-voip-calls', [adminRole, livechatManagerRole, livechatMonitorRole]),
 		Permissions.create('outbound-voip-calls', [adminRole, livechatManagerRole]),
+		Permissions.create('request-pdf-transcript', [adminRole, livechatManagerRole, livechatMonitorRole, livechatAgentRole]),
```
murtaza98, line 26: Missing translation string for this permission

murtaza98 (changes requested):
@KevLehman FYI: Left a few comments here. Overall it looks great! :tada:

apps/meteor/server/models/raw/LivechatRooms.js
KevLehman: "but current EE model has a meteor dependency"

It has one, on the query helper, on the units call 😬 and we haven't refactored it yet. So while we don't do it, i can't use that model (and refactoring is OOS for now)

ee/apps/omnichannel-transcript/package.json
KevLehman: Ohhhh, this is a "legacy" thing lol. The pdf-worker was the microservice, when i made it a package, i put it as "pdf-worker2" and seems i didn't change lol

apps/meteor/server/services/translation/service.ts
KevLehman: This will be called once per conversation 👀 to get the message we send to users, but, caching should be good anyways so 🥳 thanks

ee/apps/omnichannel-transcript/src/OmnichannelTranscript.ts
KevLehman: Good point, but I'm not sure either if UIKit blocks are part of the scope 🤔 
Let me see the parser and if the thing is not "too complex", i'll add it. Otherwise, I'll try to filter out those messages

ee/apps/omnichannel-transcript/src/OmnichannelTranscript.ts
```diff
@@ -1,16 +1,272 @@
...
+			agent: room.servedBy && (await Users.findOneAgentById(room.servedBy._id, { projection: { _id: 1, name: 1, username: 1 } })),
+			closedAt: room.closedAt,
+			messages: await this.getFiles(details.userId, messages),
+		};
+
+		try {
+			await this.doRender({ template, data, details });
+		} catch (error) {
+			await this.pdfFailed({ details, e: error as Error });
+		}
```
murtaza98, line 168:
```suggestion
		try {
			const room = await LivechatRooms.findOneById(details.rid);
			if (!room) {
				throw new Error('room-not-found');
			}
			const messages = await this.getMessagesFromRoom({ rid: room._id });

			const data = {
				visitor:
					room.v && (await LivechatVisitors.findOneById(room.v._id, { projection: { _id: 1, name: 1, username: 1, visitorEmails: 1 } })),
				agent: room.servedBy && (await Users.findOneAgentById(room.servedBy._id, { projection: { _id: 1, name: 1, username: 1 } })),
				closedAt: room.closedAt,
				messages: await this.getFiles(details.userId, messages),
			};

			await this.doRender({ template, data, details });
		} catch (error) {
			await this.pdfFailed({ details, e: error as Error });
		} finally {
			this.currentJobNumber--;
		}
```

There could be a chance that one of the databases calls before the try-catch and after `this.currentJobNumber++;` might fail, however in that case, the `currentJobNumber` won't get decremented because it is outside of try/catch loop. 
So maybe we can make use of try/catch/finally block here like this :top:

ee/apps/omnichannel-transcript/src/OmnichannelTranscript.ts
```diff
@@ -1,16 +1,272 @@
...
+						size: outBuff.length,
+					},
+				})
+				.then((file) => this.pdfComplete({ details, file }))
+				.catch((e) => this.pdfFailed({ details, e }));
+		});
+	}
+
+	private async pdfFailed({ details, e }: { details: WorkDetailsWithSource; e: Error }): Promise<void> {
+		this.currentJobNumber--;
```
murtaza98, line 201:
```suggestion
```

Won't be needed if you decide to go this route -> https://github.com/RocketChat/Rocket.Chat/pull/27568/files#r1057216219

ee/apps/omnichannel-transcript/src/OmnichannelTranscript.ts
```diff
@@ -1,16 +1,272 @@
...
+		this.log.log(`Transcript for room ${details.rid} by user ${details.userId} - Sending error message to user`);
+		await this.messageService.sendMessage({
+			fromId: 'rocket.cat',
+			rid,
+			msg: `${await this.translationService.translate('pdf_error_message', user)}: ${e.message}`,
+		});
+	}
+
+	private async pdfComplete({ details, file }: { details: WorkDetailsWithSource; file: IUpload }): Promise<void> {
+		this.currentJobNumber--;
```
murtaza98, line 226:
```suggestion
```

Won't be needed if you decide to go this route -> https://github.com/RocketChat/Rocket.Chat/pull/27568/files#r1057216219

apps/meteor/server/services/messages/service.ts
```diff
@@ -0,0 +1,24 @@
...
+export class MessageService extends ServiceClassInternal implements IMessageService {
+	protected name = 'message';
+
+	async createDirectMessage({ to, from }: { to: string; from: string }): Promise<{ rid: string }> {
+		const [toUser, fromUser] = await Promise.all([Users.findOneByUsername(to), Users.findOneByUsername(from)]);
+
+		if (!toUser || !fromUser) {
+			throw new Error('error-invalid-user');
+		}
+		return createDirectMessage([to], from);
```
murtaza98, line 18: @KevLehman Can you please re-check this? I feel there is some bug here. While testing I found that `to` and `from` params were both `userIds` and not `usernames`

murtaza98, line 18:
```suggestion
		const [toUser, fromUser] = await Promise.all([
			Users.findOneById(to, {
				projection: {
					username: 1,
				},
			}),
			Users.findOneById(from, {
				projection: {
					username: 1,
				},
			}),
		]);

		if (!toUser || !fromUser) {
			throw new Error('error-invalid-user');
		}
		return createDirectMessage([toUser.username], from);
```

Here's the one that worked for me

ee/apps/omnichannel-transcript/src/OmnichannelTranscript.ts
KevLehman: :notbad: let me do the change

apps/meteor/server/services/messages/service.ts
KevLehman: Ohhh, you're right. Since i use users with same id/username i didn't catch this. But indeed, it should be findById

murtaza98 (approved):

PR #27600, opened 2022-12-21
Chore: FileItem hover color

dougfabris:
### before
Screen Shot 2022-12-21 at 13 52 55

### after
Screen Shot 2022-12-21 at 13 43 48


apps/meteor/client/views/room/contextualBar/RoomFiles/components/FileItem.js
```diff
@@ -10,7 +9,7 @@ import MenuItem from './MenuItem';
 const hoverClass = css`
 	&:hover {
 		cursor: pointer;
-		background-color: ${colors.n100};
+		background-color: ${Palette.surface['surface-tint'].toString()};
```
juliajforesti, line 12:
```suggestion
		background-color: ${Palette.surface['surface-tint']};
```

juliajforesti (changes requested):

apps/meteor/client/views/room/contextualBar/RoomFiles/components/FileItem.js
juliajforesti: yes it will:
ts will complain if you are passing that as a `color` or `bg` Box prop.
in this case it's a css in js property value :)

dougfabris: you're completely right!

juliajforesti (approved):

PR #27603, opened 2022-12-21
[FIX] Repeated WS emission on 'stream-apps' for each HA environments

tiagoevanp:

This is related to a performance issue regarding api calls in `command/added` action.

## Further comments

AE-51


apps/meteor/app/apps/server/communication/websockets.ts
```diff
@@ -109,27 +97,6 @@ export class AppServerListener {
 		}
 
 		await this.orch.getManager()!.removeLocal(appId);
-		this.clientStreamer.emitWithoutBroadcast(AppEvents.APP_REMOVED, appId);
-	}
-
-	async onCommandAdded(command: string): Promise<void> {
-		this.clientStreamer.emitWithoutBroadcast(AppEvents.COMMAND_ADDED, command);
```
sampaiodiego, line 116: if things like are not sent to clients anymore I think it will change the current behavior, not sure if this is acceptable though.. I'll wait for @RocketChat/apps comments

PR #27616, opened 2022-12-22
[FIX] Unable to use SigV4 with S3 Uploads

debdutdeb:
`putObject` requires a file-stream (Stream) when using SigV4; So, a passthrough stream object was causing the upload to fail.

This patch switches over to the upload method which handles the stream correctly.

## Issue(s)

[SUP-110](https://rocketchat.atlassian.net/browse/SUP-110)


debdutdeb:
**Do not merge**

PR #27619, opened 2022-12-22
chore: Update eslint-config package manifest

d-gubert:

Updated the `eslint-config` package manifest to include `dependencies` instead of `devDependencies` - this way, when the package is installed in another project, dependencies are installed along with it. The way it is currently on `develop`, you have to run `npm install` manually for all dependencies if you want to use the package

## Further comments

It might mess a bit with how `yarn` keeps things organized? Not sure if the impact is significant though


packages/eslint-config/package.json
```diff
@@ -1,8 +1,9 @@
...
 	"name": "@rocket.chat/eslint-config",
-	"version": "0.5.0",
-	"description": "Rocket.Chat's JS ESLint config",
-	"devDependencies": {
+	"version": "0.5.2",
+	"description": "Rocket.Chat's JS/TS ESLint config",
+	"dependencies": {
+		"@babel/core": "^7.20.7",
 		"@babel/eslint-parser": "^7.19.1",
 		"@types/prettier": "^2.6.3",
```
tassoevan, line 8: `@types/*` dependencies must be dev-dependencies, but in this case (`@types/prettier`) it not even be needed here.

tassoevan (changes requested):

sampaiodiego (dismissed):

tassoevan (approved):

sampaiodiego (approved):

pierre-lehnen-rc (approved):

PR #27624, opened 2022-12-23
[FIX] Toast notification ignoring 403 errors

murtaza98:
## Issue(s)

[OC-528](https://rocketchat.atlassian.net/browse/OC-528)
## Further comments

[OC-528]: https://rocketchat.atlassian.net/browse/OC-528?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/meteor/packages/rocketchat-i18n/i18n/en.i18n.json
```diff
@@ -4904,6 +4904,7 @@
   "unarchive-room": "Unarchive Room",
   "unarchive-room_description": "Permission to unarchive channels",
   "Unassigned": "Unassigned",
+  "unauthorized": "Not authorized",
```
KevLehman, line 4907: didn't we have translations for this one b4? i kinda remember seeing it lol

apps/meteor/app/utils/client/lib/RestApiClient.ts
```diff
@@ -32,7 +32,7 @@ APIClient.use(async function (request, next) {
 	try {
 		return await next(...request);
 	} catch (error) {
-		if (!(error instanceof Response) || error.status !== 400) {
+		if (!(error instanceof Response)) {
```
KevLehman, line 35: Wouldn't this cause problems with 401s? If the error is swallowed, I think there would be no automatic logout on 401 👀

KevLehman (changes requested):

apps/meteor/packages/rocketchat-i18n/i18n/en.i18n.json
murtaza98: Yeah, we already had another similar translation string i.e. `error-not-authorized`, however, I decided not to use it because there were a lot of test cases like this which were depending on that exact error message. So it seemed a bit simple to just add a new translation, then go around updating all those tests :sweat_smile: 

https://github.com/RocketChat/Rocket.Chat/blob/feature/omnichannel-priorities/apps/meteor/tests/end-to-end/api/00-miscellaneous.js#L576

apps/meteor/app/utils/client/lib/RestApiClient.ts
murtaza98: Any suggestions on how to test this?

murtaza98: Based on this code, it seems like 401 errors will also result in an error being thrown since the promise eventually gets rejected. The only difference between the previous behavior and the new behavior would be that the `e` param would now contain the raw error message returned by the server like in the example of 401, it would be 
```
{
	status: 'error',
	message: 'You must be logged in to do this.',
}
```

Previously it was an instance of `Response` object from the browser 

https://github.com/RocketChat/Rocket.Chat/blob/57b12f3ba07c36383f9e53635744afe32e924904/apps/meteor/app/utils/client/lib/RestApiClient.ts#L46

aleksandernsilva (approved):

PR #27633, opened 2022-12-24
[NEW] Ability to set roles on federated rooms

MarcosSpessatto:

FED-70


apps/meteor/app/federation-v2/server/application/sender/RoomServiceSender.ts
```diff
@@ -263,4 +266,199 @@ export class FederationRoomServiceSender extends FederationService {
...
+
+		const federatedTargetUser = await this.internalUserAdapter.getFederatedUserByInternalId(internalTargetUserId);
+		if (!federatedTargetUser) {
+			return;
+		}
+
+		const userRoomRoles = await this.internalRoomAdapter.getInternalRoomRolesByUserId(internalRoomId, internalUserId);
+		const myself = federatedUser.getInternalId() === federatedTargetUser.getInternalId();
+		if (!userRoomRoles?.includes(ROCKET_CHAT_FEDERATION_ROLES.OWNER) && !myself) {
+			throw new Error('You are not allowed to change the owner');
```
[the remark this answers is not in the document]
MarcosSpessatto: hmm, good question, I think it would be better, will do that.

apps/meteor/app/federation-v2/server/Federation.ts
```diff
@@ -1,19 +1,40 @@
...
+			return false;
+		}
+		if (!userId) {
+			return true;
+		}
+
+		const userSubscription = Promise.await(Subscriptions.findOneByRoomIdAndUserId(room._id, userId));
+		if (!userSubscription) {
+			return true;
+		}
```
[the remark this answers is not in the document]
MarcosSpessatto: That's more of a concern because there are a lot of unknowns calls for this method, and since I'm adding this new feature, I thought it would be better to keep as-was and then add the validation if and only if we have all the data we need for federation.

apps/meteor/app/lib/server/methods/blockUser.ts
```diff
@@ -16,7 +16,7 @@ Meteor.methods({
 
 		const room = Rooms.findOne({ _id: rid });
 
-		if (!roomCoordinator.getRoomDirectives(room.t)?.allowMemberAction(room, RoomMemberActions.BLOCK)) {
+		if (!roomCoordinator.getRoomDirectives(room.t)?.allowMemberAction(room, RoomMemberActions.BLOCK, Meteor.userId() || undefined)) {
```
pierre-lehnen-rc, line 19: Meteor.userId will never be null in this case, since it's checked first. You could assign its value to a constant and use only the constant in this method so that typescript will understand it is not nullable here.

apps/meteor/client/views/room/lib/getRoomDirectives.ts
```diff
@@ -13,18 +13,18 @@ type getRoomDirectiesType = {
 	roomCanRemove: boolean;
 };
 
-export const getRoomDirectives = (room: IRoom): getRoomDirectiesType => {
+export const getRoomDirectives = (room: IRoom, showingUserId: IUser['_id'], userSubscription?: ISubscription): getRoomDirectiesType => {
```
tassoevan, line 16: 1. I don't like the idea of having functions with 3+ positional parameters instead of using an object for named parameters, it might hurt on refactors and code reviews;
2. If you have to inject so many dependencies in a function, might be better to introduce some companion helpers, like a React hook:

```ts
const useRoomDirectives = () => {
  const room = useRoom(); // consumes `RoomContext`
  const subscription = useRoomSubscription(); // consumes `RoomContext`
  const uid = useUserId(); // consumes `UserContext`
  return useCallback(() => getRoomDirectives({ room, uid, subscription }));
};
```

This example illustrates how we use React contexts as dependency containers.

tassoevan (changes requested):

apps/meteor/client/views/room/lib/getRoomDirectives.ts
MarcosSpessatto: Given the time (release constraints) I preferred to go with option 1, and I'll refactor that. Thanks for the suggestion.

pierre-lehnen-rc (approved):

tassoevan (approved):

PR #27644, opened 2022-12-26
[NEW] Permission to bypass message editing and removing limits

LucianoPierdona:

This PR adds a new permission called `bypass-time-limit-edit-and-delete`, to bypass the time limit when editing or deleting a message, the purpose of this PR is to remove the limitation for bots that cannot change the message after the specified time.

## Steps to test or reproduce

- Add a Jira integration with RC such that it creates threads in RC, and when the status of the Jira task is updated, the first message in the thread in RC gets updated via REST API to correspond to the Jira status.
- In RC, go to Administration -> Message -> Block Message Editing After X Minutes and set the value to 10
- Update the status of the Jira task after 10 minutes
## Further comments

[TC-329]

[TC-329]: https://rocketchat.atlassian.net/browse/TC-329?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/meteor/app/authorization/server/functions/upsertPermissions.ts
```diff
@@ -225,6 +225,7 @@ export const upsertPermissions = async (): Promise<void> => {
 		{ _id: 'view-import-operations', roles: ['admin'] },
 		{ _id: 'clear-oembed-cache', roles: ['admin'] },
 		{ _id: 'videoconf-ring-users', roles: ['admin', 'owner', 'moderator', 'user'] },
+		{ _id: 'bypass-time-limit-edit-and-delete', roles: ['bot'] },
```
matheusbsilva137, line 228:
```suggestion
		{ _id: 'bypass-time-limit-edit-and-delete', roles: ['admin'] },
```
Maybe this permission should be assigned only to admins by default so that no unexpected behavior is introduced to users when updating to the next rc version.

matheusbsilva137 (changes requested):
The `messageActionDefault.ts` file and the `canDeleteMessage` function in the `apps/meteor/client/lib/chats/data.ts` file should also be updated so that the action buttons are displayed in the UI when a user has the bypass permission.

apps/meteor/client/lib/chats/data.ts
```diff
@@ -210,10 +212,11 @@ export const createDataAPI = ({ rid, tmid }: { rid: IRoom['_id']; tmid: IMessage
 		}
 
 		const blockDeleteInMinutes = settings.get('Message_AllowDeleting_BlockDeleteInMinutes') as number | undefined;
+		const bypassBlockTimeLimit = hasPermission('bypass-time-limit-edit-and-delete');
 		const elapsedMinutes = moment().diff(message.ts, 'minutes');
 		const onTimeForDelete = !blockDeleteInMinutes || !elapsedMinutes || elapsedMinutes <= blockDeleteInMinutes;
 
-		return deleteAllowed && onTimeForDelete;
+		return !bypassBlockTimeLimit && deleteAllowed && onTimeForDelete;
```
matheusbsilva137, line 219:
```suggestion
		return deleteAllowed && onTimeForDelete;
```
This would not allow deleting a message when the bypass permission in active.

apps/meteor/client/lib/chats/data.ts
```diff
@@ -210,10 +212,11 @@ export const createDataAPI = ({ rid, tmid }: { rid: IRoom['_id']; tmid: IMessage
 		}
 
 		const blockDeleteInMinutes = settings.get('Message_AllowDeleting_BlockDeleteInMinutes') as number | undefined;
+		const bypassBlockTimeLimit = hasPermission('bypass-time-limit-edit-and-delete');
 		const elapsedMinutes = moment().diff(message.ts, 'minutes');
 		const onTimeForDelete = !blockDeleteInMinutes || !elapsedMinutes || elapsedMinutes <= blockDeleteInMinutes;
```
matheusbsilva137, line 217:
```suggestion
		const bypassBlockTimeLimit = hasPermission('bypass-time-limit-edit-and-delete');
		const elapsedMinutes = moment().diff(message.ts, 'minutes');
		const onTimeForDelete = bypassBlockTimeLimit || !blockDeleteInMinutes || !elapsedMinutes || elapsedMinutes <= blockDeleteInMinutes;
```
Try checking the permission here instead

matheusbsilva137 (changes requested):

matheusbsilva137 (dismissed):

gabriellsh (approved):
Frontend ok

ankar84:
wait that fix a lot!
thank you!

apps/meteor/tests/end-to-end/api/24-methods.js
```diff
@@ -1623,6 +1623,40 @@ describe('Meteor.methods', function () {
...
+								.expect('Content-Type', 'application/json')
+								.expect(200)
+								.expect((res) => {
+									expect(res.body).to.have.property('message').that.is.an('object');
+									expect(res.body.message.msg).to.equal('https://github.com updated with bypass');
+								});
+						})
+						.end(done);
+				});
+		});
```
sampaiodiego, line 1658: pls use async/await , the code looks so much better:
```suggestion
		it('should update a message when bypass time limits permission is enabled', async (done) => {
			await Promise.all([
				updatePermission('bypass-time-limit-edit-and-delete', ['admin']),
				updateSetting('Message_AllowEditing_BlockEditInMinutes', 0.01),
			]);

			await request
				.post(methodCall('updateMessage'))
				.set(credentials)
				.send({
					message: JSON.stringify({
						method: 'updateMessage',
						params: [{ _id: messageId, rid, msg: 'https://github.com updated with bypass' }],
						id: 'id',
						msg: 'method',
					}),
				})
				.expect('Content-Type', 'application/json')
				.expect(200)
				.expect((res) => {
					expect(res.body).to.have.a.property('success', true);
					expect(res.body).to.have.a.property('message').that.is.a('string');
				});

			await request
				.get(api(`chat.getMessage?msgId=${messageId}`))
				.set(credentials)
				.expect('Content-Type', 'application/json')
				.expect(200)
				.expect((res) => {
					expect(res.body).to.have.property('message').that.is.an('object');
					expect(res.body.message.msg).to.equal('https://github.com updated with bypass');
				});
		});
```

sampaiodiego (changes requested):
just a comment on code style on tests.. everything else seems good

hugocostadev (approved):

rodrigok (dismissed):

rodrigok (dismissed):

rodrigok:
@hugocostadev could you review this again?

hugocostadev (approved):

hugocostadev (approved):

rodrigok (approved):

PR #27647, opened 2022-12-27
[BREAK] Remove less theming

ggazzo:
[ARCH-79]

Well, this pull request removes the settings formerly used to customize the theme.

Some reasons for this change:

- after using the fuselage components the available colors were insufficient
- colors are no longer used by components
- color management via setting was unsatisfactory
- We were forced to use the less library to compile a 10-line css

So please, after these changes your customizations WILL DISAPPEAR, You will need to reapply the colors and fonts using `css var` through the `custom css` setting

Because we are removing your customizations, this change is considered a breaking change.

We know this can be frustrating and annoying. But this standardization allowed us things like Dark theme, and very soon we will bring a tool to help with customization

## Further comments

[ARCH-79]: https://rocketchat.atlassian.net/browse/ARCH-79?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/meteor/app/theme/server/server.ts
```diff
@@ -1,19 +1,25 @@
...
-import { Logger } from '../../logger/server';
 import { addStyle } from '../../ui-master/server/inject';
-import { Theme } from './Theme';
 
-const logger = new Logger('rocketchat:theme');
-
-export const theme = new Theme({ logger });
+settings.watch('theme-custom-css', (value) => {
+	if (!value || typeof value !== 'string') {
+		addStyle('theme-custom-css', '');
```
ggazzo, line 12:
```suggestion
		addStyle('css-theme', '');
```

tassoevan (approved):

Issue #27394, opened 2022-11-28
Livechat: line breaks do not appear


Floflobel:
### Description:

In the livechat client, the line break does not appear either when sending a message or when receiving a message.

### Steps to reproduce:

1. Send a message with line breaks
2. Look in the livechat, no line breaks are displayed
3. Look in Rocketchat, line breaks are displayed

### Expected behavior:

Show line breaks in the livechat

### Actual behavior:

Livechat:
![mki7mlugYX](https://user-images.githubusercontent.com/13889750/204393718-704e11a9-0376-4e7c-a002-b7028f1a5d7b.png)

Livechat:
![g0EFNkpTKE](https://user-images.githubusercontent.com/13889750/204393748-3b613b29-4b1e-4187-8fed-1beea0aa72cc.png)

Rocketchat:
![2geIABrmAB](https://user-images.githubusercontent.com/13889750/204393765-729022c7-164b-4946-b45f-4528e0155b75.png)

### Server Setup Information:

- Version of Rocket.Chat Server: 5.3.3
- Operating System: Ubuntu 20.04.5 LTS
- Deployment Method: snap
- Number of Running Instances: 1

### Client Setup Information

- Desktop App or Browser Version: Firefox 107.0
- Operating System:  Linux

Aashray446:
Hi, Can I work on this feature?

Aashray446:
@dudanogueira Can you please review my PR?  :-)

dudanogueira:
Hi! thanks for the PR! Our team will review!

albetros24:
hi can I still work on this issue @dudanogueira  Sir

Aashray446:
@albetros24, thanks for your interest, Actually I was a little busy with other tasks so couldn't focus on it much. I will also be working on it soon. But as you are interested you can try solving this. I will be very happy to help you, you can check out my PR to for getting some more information about the issue or ping me on rocket.chat server.

PR #27662, opened 2022-12-28
[FIX] line breaks don't appear in livechat #27394

Aashray446:


Aashray446:
Previously, live chat was using the Markdown-it default parser which was ignoring any extra line breaks. The parser rules can't be modified, so I changed the to in-line parsing which recognizes the new line as a soft or hard break. Thus it was easy to replace the renderer rules to change those to br and the whole HTML rendered will be wrapped up in a P tag.

filipemarins (changes requested):
it is breaking some markdown styles like header `# h1 header`.

Aashray446:
OK, I will solve it asap. 
Thank you for the review :smile:

PR #27678, opened 2023-01-02
Chore: move apps to ee directory

thassiov:

Protects the code under EE license.


geekgonecrazy:
IMO this is a breaking change and we should be transparent and say its a breaking change. 

It doesn't break functionality for customers.  But for any contributors that might be used to being able or planning to contribute to this code they will likely be unable or unwilling to contribute to it any more. Along lines of changing the API would be a breaking change.

Also still not a fan of this being moved.

graywolf336:
100% agree with Aaron, I am not a fan of this change. It appears to serve no purpose other than restrict the license and makes contributing more difficult.

debdutdeb:
Just tripling down on what Aaron and Bradley said already - let's please explain the purpose of this big change in this PR as I've gotten **a lot** (more than 10 for sure) confused questions around this very recently, and I'm not the best person to answer or explain. Best if it comes from the owners.

PR #27690, opened 2023-01-04
Chore: Move service shutdown logic to each service

sampaiodiego:
## Issue(s)

ARCH-120


apps/meteor/ee/server/lib/EnterpriseCheck.ts
```diff
@@ -0,0 +1,58 @@
...
+			if (!currentService) {
+				return true;
+			}
+
+			const { nodes } = currentService;
+
+			const firstNode = nodes.sort().shift();
+
+			// if the first node is the current node and there are others nodes running the same service, then it should shutdown
+			return firstNode === this.broker.nodeID && nodes.length > 0;
```
geekgonecrazy, line 37: So in rare cases.. this might depend on outside orchestrator to pop a node back online?

sampaiodiego: no, that's what the `sort` tries to solve.. if 2 nodes online performing the check at the same time, the "first node" will be the same for both, that means only one of them will shut down.

ggazzo (approved):

apps/meteor/ee/server/lib/EnterpriseCheck.ts
geekgonecrazy: Got it!  That makes sense 👍

PR #27692, opened 2023-01-05
[IMPROVE] Added priority sorting to subscriptions

aleksandernsilva:

### DIFF
--- a/apps/meteor/client/sidebar/hooks/useQueryOptions.ts
+++ b/apps/meteor/client/sidebar/hooks/useQueryOptions.ts
@@ -1,7 +1,8 @@
+import { OmnichannelSortingMechanismSettingType as OmniSortType } from '@rocket.chat/core-typings';
 import { useUserPreference, useSetting } from '@rocket.chat/ui-contexts';
 import { useMemo } from 'react';
 
-export const useQueryOptions = (): {
+type QueryOptions = {
 	sort:
 		| {
 				lm?: number | undefined;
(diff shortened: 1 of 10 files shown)


MartinSchoeler (changes requested):
Screen Shot 2023-01-11 at 2 52 44 PM

Chats that have never received a priority sit on top of the prioritized ones (`chat3`), if I add a priority and then remove, it works as intended (`chat1`)

apps/meteor/client/sidebar/hooks/useQueryOptions.ts
```diff
@@ -13,20 +14,35 @@ export const useQueryOptions = (): {
...
 		  };
-} => {
-	const sortBy = useUserPreference('sidebarSortby');
-	const showRealName = useSetting('UI_Use_Real_Name');
+};
+
+const getSortOption = (sortingMechanism: OmniSortType, sortBy: string, showRealName: boolean) =>
+	({
+		[OmniSortType.Priority]: { priorityWeight: 1, estimatedServiceTimeAt: 1, ts: 1 },
+		[OmniSortType.SLAs]: { estimatedServiceTimeAt: 1, priorityWeight: 1, ts: 1 },
```
murtaza98, line 28: For priorities and SLAs, can you please use this helper function instead to get the sort options? Just so we have a single definition of it
https://github.com/RocketChat/Rocket.Chat/blob/5b6a6962f9754136900ad431a86005faa3558525/apps/meteor/app/livechat/lib/inquiries.ts#L9

MartinSchoeler (approved):

MartinSchoeler:
Okay I made a mistake on where the sorting happens, it's only on queued chats 
Screen Shot 2023-01-23 at 3 57 11 PM

PR #27699, opened 2023-01-06
[FIX] Missing placeholders from encrypted channel preview messages now with expected behavior

Educg550:

Referent to [TC-259]

Clipboard screenshot
Note that now the file uses `isE2EEMessage()` function and the `useTranslation` internal hook.

## Steps to test or reproduce
In any encrypted channel:

1.     Send a few encrypted messages
2.     Create a new thread
3.     Send messages inside the thread, with the “Send to Channel” option activated.
4.     Reset the user’s E2E password
5.     Log in and enter the Room used in the previous steps without saving the new E2E password.

You'll note that now the preview messages created by the “Send to Channel” option really show up the correct message, instead of its encrypted hash code.

[TC-259]: https://rocketchat.atlassian.net/browse/TC-259?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/meteor/client/components/message/variants/threadPreview/ThreadMessagePreviewBody.tsx
```diff
@@ -10,7 +12,13 @@ type ThreadMessagePreviewBodyProps = {
...
 
 const ThreadMessagePreviewBody = ({ message }: ThreadMessagePreviewBodyProps): ReactElement | null => {
-	const parsedMessage = parseMessageTextToAstMarkdown(message, { colors: true, emoticons: true });
+	const t = useTranslation();
+	const isEncryptedMessage = isE2EEMessage(message);
+
+	const parsedMessage = parseMessageTextToAstMarkdown(
+		message.e2e === 'done' || !isEncryptedMessage ? message : { ...message, msg: t('E2E_message_encrypted_placeholder') },
+		{ colors: true, emoticons: true },
+	);
```
hugocostadev, line 21: Instead of using the parsed message of a static message (`E2E_message_encrypted_placeholder` ) that will always result in the same thing I suggest you check if it's an e2e and if it's encrypted outside the function and if it is, return the string, no the parsed Message. You can change the return type of the component if needed

hugocostadev (changes requested):
I realized that when typing the e2e password to decrypt the messages the Thread Preview has a delay to show the original message and sometimes does not show it, and other time is flickering. 

Also, I realized that the first message and the title thread in never decrypted, check the pic below:
image

PR #27701, opened 2023-01-08
[FIX] Add a Open button and a onClick listener inside the ThreadMetrics.tsx

anshgoyalevil:

- Previously, whenever a chat was going on in a thread, the 3rd person was not able decide where to click to check what messages are inside that thread.
- There was a button named "Reply", which means that the 3rd person wants to reply on that thread.
- What if the 3rd person don't want to reply, and was just roaming around to check the discussion going inside the thread.
- Thus, I added an "Open" button, which makes it easy for the user to click it and open the thread. The functionality is same as the "Reply" button, but now, the user wouldn't search around the way to open the thread.
- Also, added the onClick listener inside the 'Thread' icon besides the message counter. This will further improve the user experience.

## Here is what is looks now.
Now

Previously, there was no "Open" button.

## Steps to test or reproduce
I have tested the UI of web client on Desktop PC (1920x1080 res) 

## Further comments

I decided to add this fix because when I was first roaming inside the chat interface, I was so much confused, like where to click to open the discussion going inside the threads. The thread counter was making me more anxious, as I don't wanted to click "Reply" button to open the thread, because, I was not doing any reply, but was just reading the discussion going inside that thread.
Like me, many more people would be confused, thus, added a button to open the thread.

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,7 +36,7 @@ You can follow these instructions to setup a dev environment:
 ### Starting Rocket.Chat
 
 ```
-yarn dsv
+yarn dev
 ```
 
 After initialized, you can access the server at http://localhost:3000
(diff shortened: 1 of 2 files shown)


ggazzo:
I appreciate the contribution, but there is no real difference between the reply and open button, other than the label.

PR #27716, opened 2023-01-10
[FIX] Unintended "Reply on thread" button showing on thread's first message

Educg550:

Referent to TC-128

#### Before:
image

#### After:
image

## Steps to test or reproduce

Open a new channel, then:
1. Send any message
2. Click "Reply in thread" button and send a message in thread to that message
3. You'll see that the button "Reply in thread" won't show anymore for the top message, avoiding "thread nesting" and limiting it to only 1 thread, as it should be
4. Also any other messages sent in that thread won't be allowed to create another inside it

## Further comments

As flagged in comment on that issue, seems that there was a discussion whether the thread should open again or have a new created one when clicking on such unintended button. That said, it's conclusive to say that there was no point on opening a new thread for the top message when you're already on one, same for opening new threads for further messages.


hugocostadev:
Hi @Educg550 we have a method called  `isThreadMainMessage`, maybe you can use it to add the correct context and differentiate in other way in the MessageList without creating additional property   
image

apps/meteor/client/components/message/toolbox/Toolbox.tsx
```diff
@@ -54,7 +59,15 @@ const Toolbox = ({ message }: ToolboxProps): ReactElement | null => {
 		);
 		const menuActions = await MessageAction.getButtons({ message, room, user, subscription, settings: mapSettings, chat }, context, 'menu');
 
-		return { message: messageActions, menu: menuActions };
+		const filteredActions =
```
Educg550, line 62: This filtering has some delay and can show up as a blinking icon to the final user. That's why in my opinion, we should filter the icon by label on the `MessageAction.ts` file

apps/meteor/client/components/message/toolbox/Toolbox.tsx
```diff
@@ -14,21 +15,25 @@ import { useRoom } from '../../../views/room/contexts/RoomContext';
...
-	if (isThreadMessage(message) || message.renderedOnThread) {
+	if (isThreadMessage(message) || isThreadFirstMessage(message)) {
 		return 'threads';
 	}
 	return 'message';
 };
 
 type ToolboxProps = {
-	message: IMessage;
+	message: IThreadMessage | IThreadMainMessage;
```
Educg550, line 36: To avoid these kind of uncertain typings, I'd recommend to also keep the `renderedOnThread` prop on the `IMessage` interface, since both of these types extend from it.

apps/meteor/client/components/message/toolbox/Toolbox.tsx
```diff
@@ -14,21 +14,25 @@ import { useRoom } from '../../../views/room/contexts/RoomContext';
 import { useToolboxContext } from '../../../views/room/contexts/ToolboxContext';
 import MessageActionMenu from './MessageActionMenu';
 
-const getMessageContext = (message: IMessage, room: IRoom): MessageActionContext => {
+const isThreadFirstMessage = (message: IThreadMessage | IThreadMainMessage): boolean => {
+	return message.renderedOnThread;
+};
```
hugocostadev, line 19: use short handed syntax

apps/meteor/client/components/message/toolbox/Toolbox.tsx
hugocostadev: No, ToolBox has to accept the IMessage type, all other types (IThreadMessage | IThreadMainMessage) extend the original IMessage interface, if you keep like that anyone that is reading this component will think that the ToolBox component is used just for the Thread Messages. We always need to be clear in our implementation.

apps/meteor/client/components/message/toolbox/Toolbox.tsx
hugocostadev: I'm checking this behavior, maybe the root cause is not the filtering

hugocostadev (changes requested):

apps/meteor/client/components/message/toolbox/Toolbox.tsx
```diff
@@ -14,21 +14,25 @@ import { useRoom } from '../../../views/room/contexts/RoomContext';
...
+
+const getMessageContext = (message: IThreadMessage | IThreadMainMessage, room: IRoom): MessageActionContext => {
```
hugocostadev, line 21: The message should be IMessage type

apps/meteor/client/components/message/ToolboxHolder.tsx
```diff
@@ -7,7 +7,7 @@ import { useIsVisible } from '../../views/room/hooks/useIsVisible';
 import Toolbox from './toolbox/Toolbox';
 
 type ToolboxHolderProps = {
-	message: IMessage;
+	message: IThreadMessage | IThreadMainMessage;
```
hugocostadev, line 10: This should be IMessage type

hugocostadev (approved):

hugocostadev (changes requested):

hugocostadev (approved):

tassoevan (approved):

PR #27718, opened 2023-01-10
[FIX] asyncawaitify app converters

thassiov:


apps/meteor/ee/app/apps/bridges/listeners.js
```diff
@@ -129,24 +129,24 @@ export class AppListenerBridge {
 	}
 
 	async roomEvent(inte, room, ...payload) {
-		const rm = this.orch.getConverters().get('rooms').convertRoom(room);
+		const rm = await this.orch.getConverters().get('rooms').convertRoom(room);
 
-		const params = (() => {
+		const params = (async () => {
```
d-gubert, line 134: `params` can't be a Promise when we send it to the Apps-Engine, we need an `await` here

d-gubert (changes requested):

tapiarafael (approved):

PR #27750, opened 2023-01-13
[NEW] Add files component to chat transcript template

filipemarins:

Add messages and files component;
Add translations to the template;
Add a README file;

## Issue(s)

oc-520 oc-521 OC-618

## Further comments

Screenshot 2023-01-13 at 12 54 34


packages/pdf-worker/src/strategies/ChatTranscript.ts
```diff
@@ -0,0 +1,74 @@
...
+import type { IStrategy } from '../types/IStrategy';
+import exportChatTranscript from '../templates/ChatTranscript';
+import type { ChatTranscriptData, PDFMessage } from '../templates/ChatTranscript';
+
+export class ChatTranscript implements IStrategy {
+	private isNewDay(current: PDFMessage, previous: PDFMessage | undefined): boolean {
+		return !previous || !moment(current.ts).isSame(previous.ts, 'day');
+	}
+
+	private parserMessages(messages: PDFMessage[], dateFormat: unknown, timeAndDateFormat: unknown): unknown[] {
```
aleksandernsilva, line 13: Why does this method have so many unknowns?

filipemarins: This is because the data come from an external source and are uncertain. The `omnichannel-transcript` app doesn't share their types, so what we can do right now is set it as unknown...

aleksandernsilva (approved):

KevLehman (approved):

PR #27327, opened 2022-11-23
[FIX] Matrix Federation send message when multiple users are involved was not working

MarcosSpessatto:
## Proposed changes (including videos or screenshots)

FED-53
FED-61

It also allows us to use the latest version of Synapse.

PR #27282, opened 2022-11-16
Chore: Federation E2E tests for messaging + admin

MarcosSpessatto:
## Proposed changes (including videos or screenshots)

Depends on https://github.com/RocketChat/Rocket.Chat/pull/27115

Jira: FED-41

PR #27763, opened 2023-01-16
[NEW] Matrix federation events coverage expansion + bug fixes

MarcosSpessatto:

- Disables not supported features in federated channels (#27204)
- Add support for the whole Matrix Markdown spec (#27725)
- Allow federated rooms with the same name per different servers when they have a contact between them (#27559)
-  Matrix Federation send message when multiple users are involved was not working (#27327)
- Allow the room owner to set roles on federated rooms, following a specific and different approach as normal rooms (available on docs) (https://github.com/RocketChat/Rocket.Chat/pull/27633)
- E2E Tests (PS: they don't run on CI) (#27578 #27282 #27115)
- Disables admin actions disallowed for federated users/rooms (https://github.com/RocketChat/Rocket.Chat/pull/27251)

FED-81


apps/meteor/app/lib/server/functions/deleteUser.ts
```diff
@@ -15,19 +16,15 @@ import { LivechatUnitMonitors } from '../../../../ee/app/models/server';
...
 	}
 
-	if (user.federation) {
-		const existingSubscriptions = Subscriptions.find({ 'u._id': user._id }).count();
-
-		if (existingSubscriptions > 0) {
-			throw new Meteor.Error('FEDERATION_Error_user_is_federated_on_rooms');
-		}
+	if (isUserFederated(user)) {
+		throw new Meteor.Error('error-user-is-federated', 'Cannot delete federated user');
```
pierre-lehnen-rc, line 27: I remember raising some concerns about this on the original PR but don't recall what came of it. Why is this even a requirement?

apps/meteor/app/mentions/server/server.js
```diff
@@ -28,7 +28,20 @@ export class MentionQueries {
 	}
 
 	getChannels(channels) {
-		return Rooms.find({ name: { $in: [...new Set(channels)] }, t: { $in: ['c', 'p'] } }, { fields: { _id: 1, name: 1 } }).fetch();
+		return Rooms.find(
+			{
+				$and: [
+					{
```
pierre-lehnen-rc, line 34: This first `$and` is not doing anything.

apps/meteor/app/models/server/models/Rooms.js
```diff
@@ -344,13 +344,24 @@ export class Rooms extends Base {
...
+	findOneByNameAndType(name, type, options, includeFederatedRooms = false) {
 		const query = {
-			name,
 			t: type,
 			teamId: {
 				$exists: false,
 			},
+			...(includeFederatedRooms
+				? {
+						$and: [
```
pierre-lehnen-rc, line 355: This first `$and` is not doing anything

apps/meteor/app/models/server/models/Rooms.js
```diff
@@ -344,13 +344,24 @@ export class Rooms extends Base {
...
+							{
+								$or: [
+									{ $and: [{ $or: [{ federated: { $exists: false } }, { federated: false }], name }] },
+									{ federated: true, fname: name },
+								],
+							},
+						],
+				  }
+				: { $and: [{ $or: [{ federated: { $exists: false } }, { federated: false }], name }] }),
```
pierre-lehnen-rc, line 364: This `$and` is not doing anything

apps/meteor/app/ui-message/client/popup/messagePopupConfig.js
```diff
@@ -382,6 +383,7 @@ Template.messagePopupConfig.helpers({
 					.find(
 						{
 							name: exp,
+							federated: { $exists: false },
```
pierre-lehnen-rc, line 386: shouldn't this also check for `federation: false` ?

pierre-lehnen-rc (changes requested):

apps/meteor/app/lib/server/functions/deleteUser.ts
MarcosSpessatto: Yes, I remember, we've agreed we should discuss that internally and find a solution to this, having in mind GDPR rules, but it was not discussed yet, would you suggest removing this validation instead?

apps/meteor/app/mentions/server/server.js
MarcosSpessatto: Thanks, didn't notice that 😞

apps/meteor/app/lib/server/functions/deleteUser.ts
MarcosSpessatto: @pierre-lehnen-rc I'm reverting this change for now.

pierre-lehnen-rc (approved):

hugocostadev (approved):

pierre-lehnen-rc (approved):

PR #27765, opened 2023-01-16
Chore: Add surface-light background to AppRow

rique223:

In light of the recent changes to the fuselage color system, the AppRow went from having a default background color of white to not having a color. Because of this, I've added the bg='surface-light' prop to the Wrapper box of the component.
Before:
image

After: 
image


apps/meteor/client/views/admin/apps/AppsList/AppRow.tsx
```diff
@@ -76,6 +76,7 @@ const AppRow = (props: AppRowProps): ReactElement => {
 			pi='x16'
 			borderRadius='x4'
 			className={hoverClass}
+			bg='surface-light'
```
dougfabris, line 79:
```suggestion
			bg='light'
```

dougfabris (changes requested):

apps/meteor/client/views/admin/apps/AppsList/AppRow.tsx
rique223: Solved

dougfabris (approved):

PR #27639, opened 2022-12-26
[BREAK] Update Meteor to 2.9 and NodeJS to 14.21.1

sampaiodiego:

PR #27767, opened 2023-01-16
Regression: Add support for 2FA errors to `Meteor.callAsync`

sampaiodiego:

Since #27639 , when `callMethod` was changed to use `Meteor.callAsync` instead of `Meteor.call`, 2FA flow was not working.


apps/meteor/client/lib/2fa/process2faReturn.ts
```diff
@@ -76,3 +76,53 @@ export function process2faReturn({
...
+	promise: Promise<unknown>;
+	onCode: (code: string, method: string) => void;
+	emailOrUsername: string | null | undefined;
+}): Promise<unknown> {
+	return new Promise<unknown>((resolve, reject) => {
+		promise
+			// if the promise is resolved, we don't need to do anything
+			.then((result) => {
+				resolve(result);
+			})
```
ggazzo, line 94:
```suggestion
			.then(resolve)
```

apps/meteor/client/lib/2fa/process2faReturn.ts
```diff
@@ -76,3 +76,51 @@ export function process2faReturn({
...
+						},
+						onClose: (): void => {
+							imperativeModal.close();
+							reject(new Meteor.Error('totp-canceled'));
+						},
+					},
+				});
+			});
+	});
+}
```
ggazzo, line 126: I think you dont need to wrap the whole thing in a promise, just wrapping imperativeModal would do the trick

apps/meteor/client/lib/2fa/process2faReturn.ts
```diff
@@ -76,3 +76,51 @@ export function process2faReturn({
...
+					props: {
+						...props,
+						onConfirm: (code: string, method: string): void => {
+							imperativeModal.close();
+
+							// once we have the code, we resolve the promise with the result of the `onCode` callback
+							resolve(onCode(method === 'password' ? SHA256(code) : code, method));
+						},
+						onClose: (): void => {
+							imperativeModal.close();
```
ggazzo, line 119:
```suggestion
export async function process2faAsyncReturn({
	promise,
	onCode,
	emailOrUsername,
}: {
	promise: Promise<unknown>;
	onCode: (code: string, method: string) => void;
	emailOrUsername: string | null | undefined;
}): Promise<string> {
	return promise
		// if the promise is rejected, we need to check if it's a 2fa error
		.catch((error) => {
			// if it's not a 2fa error, we reject the promise
			if (!isTotpRequiredError(error) || !hasRequiredTwoFactorMethod(error)) {
				throw error;
			}

			const props = {
				method: error.details.method,
				emailOrUsername: emailOrUsername || error.details.emailOrUsername || Meteor.user()?.username,
			};

			assertModalProps(props);

			return new Promise<string>((resolve, reject) => {

			imperativeModal.open({
				component: TwoFactorModal,
				props: {
					...props,
					onConfirm: (code: string, method: string): void => {
						imperativeModal.close();

						// once we have the code, we resolve the promise with the result of the `onCode` callback
						resolve(onCode(method === 'password' ? SHA256(code) : code, method));
					},
					onClose: (): void => {
						imperativeModal.close();
						reject(new Meteor.Error('totp-canceled'));
					},
				},
			});
		});
	});
}
```

ggazzo (approved):

PR #27770, opened 2023-01-16
feat(matrix): Matrix search UI

gabriellsh:
### Depends on #28330

TC-315


apps/meteor/client/sidebar/header/MatrixFederationSearch/FederatedRoomList.tsx
```diff
@@ -0,0 +1,58 @@
...
+		{ getNextPageParam: (lastPage) => lastPage.nextPageToken },
+	);
+
+	const { mutate: onClickJoin, isLoading: isLoadingMutation } = useMutation(
+		['federation/joinExternalPublicRoom'],
+		async (id: string) => {
+			return joinExternalPublicRoom({ externalRoomId: id as `!${string}:${string}` });
+		},
+		{ onSuccess: () => setModal(null) },
+	);
```
hugocostadev, line 35: Maybe create two different hooks, one for the query and other for the Mutation, what do you think? 
That way these hooks can be used elsewhere

hugocostadev (changes requested):

hugocostadev:
Feedback:

1. The margin-bottom seems to be smaller in the design
2. the button should have some paddings according to prototype 

image

hugocostadev:
The user icon is not aligned correctly:

image

The scrollbar overlaps the Join button
image

The modal should have 664px and it has 608px
image

MarcosSpessatto:
> Is it necessary to repeat the channel's name in the server indicator? I think that we could remove it.
I'd say this is necessary, since it's not a repetition only, the second information is what Matrix call canonical_alias, if you create the channel from Element, you can define the canonical_alias for the channel which can different from the name, once it's unique per server you can have a canonical_alias different from the room name.

hugocostadev (approved):

ggazzo (approved):

Issue #27442, opened 2022-12-03
showing variable name instead of text


sourav-k7:
### Steps to reproduce:

1. create new account
2. try to login
3. on email confirmation page

### Actual behavior:

![image](https://user-images.githubusercontent.com/70101213/205430177-f0391d2c-28ac-461b-ae22-02d4c5e4d31e.png)

### Server Setup Information:

- Version of Rocket.Chat Server: 
- Operating System: 
- Deployment Method: 
- Number of Running Instances: 
- DB Replicaset Oplog: 
- NodeJS Version: 
- MongoDB Version:

### Client Setup Information

- Desktop App or Browser Version:
- Operating System:

raghavpatidar:
Can I work on this issue?
Please Assign it to me.

dvnkln:
Running into the same issue. There are also some spaces missing and the footer including privacy policy etc. doesn't get replaced by our custom text anymore.

![Unbenannt](https://user-images.githubusercontent.com/33687762/206641627-1ce7abe9-cf5a-467b-95f9-d84b35669f66.png)

**Server Setup Information:**

- Version of Rocket.Chat Server: 5.4.0
- Operating System: deb11
- Deployment Method: manual
- Number of Running Instances: 1
- DB Replicaset Oplog: Enabled
- NodeJS Version: 14.19.3 - x64
- MongoDB Version: 4.4.17

dvnkln:
Still happening with `5.4.1` even though the changelogs hints at a fix.

hugocostadev:
fixed by: https://github.com/RocketChat/Rocket.Chat/pull/28156

---
**Questions? Help needed? Feature Requests?** 

- Join our [Open Server](https://open.rocket.chat/) in the #support channel and feel free to raise a question
- Join our [Community Forum](https://forums.rocket.chat/) and search/create a post there
- Feature Request: Open an issue in [Feature Request](https://github.com/RocketChat/feature-requests) repository

PR #27771, opened 2023-01-17
[FIX] Fixed variables not being translated as expected sometimes

Thiagof99:

In some forms, some of the placeholders and text wasn't being translated properly. Instead, the variable name was showing up in its place, here is an example of this behaviour in the register screen:
image
image
after finding the problem I maneged to solve it by changing the translation imports. Here it is how it looks after the fix
image

## Issue(s)

#27442
## Further comments

The bugs that poped up to me were slightly diferent form some of the ones described at the issue, so I fixed everything i could find.


Thiagof99:
@hugocostadev please review

Thiagof99:
@hugocostadev please review

Thiagof99:
@dudanogueira please review

Thiagof99:
@rodrigok please review

Thiagof99:
@sampaiodiego please review

Thiagof99:
@engelgabriel please review

Thiagof99:
@ggazzo please review

PR #27781, opened 2023-01-17
feat: Custom roles and Public&Private apps stats

totoi690:
## Problem statement
We want to collect new metrics to measure the impact of the custom roles freeze in CE and have a better understanding about the use of apps

## Changes
The metrics below are being implemented on the core and ALSO on cloud side

- *Custom roles*: Number of custom roles in the workspace
- *Apps*: Public / Private per app

## Test plan
Check if the new stats is being received on the cloud side

TC-375


apps/meteor/server/models/raw/Roles.ts
```diff
@@ -157,6 +157,15 @@ export class RolesRaw extends BaseRaw<IRole> implements IRolesModel {
 		return this.find(query, options || {});
 	}
 
+	findCustomUserRoles(options?: FindOptions<IRole> | undefined): FindCursor<IRole> {
```
debdutdeb, line 160:
```suggestion
	findCustomUserRoles(options?: FindOptions<IRole>): FindCursor<IRole> {
```

apps/meteor/server/models/raw/Roles.ts
```diff
@@ -157,6 +157,15 @@ export class RolesRaw extends BaseRaw<IRole> implements IRolesModel {
...
+		const query = {
+			scope: 'Users' as IRole['scope'],
```
debdutdeb, line 162:
```suggestion
		const query: Filter<IRole> = {
			scope: 'Users',
```

debdutdeb (dismissed):

apps/meteor/server/models/raw/Roles.ts
debdutdeb: It's better to declare the type of the variable here and let ts check if the value makes sense, instead of casting. Casting means overriding a type (not changing) (in other words "shut up typescript, I know this type better than you") - which should be a last resort IMO. More so if you look at the scope field, it already has `Users` - you just need to help the typechecker a little bit in this case to understand the exact type name.

debdutdeb: Check [this](https://www.typescriptlang.org/play?#code/C4TwDgpgBAYg9nKBeKBvAUFLUCWATALigGdgAnHAOwHMBuTbAQ2oiMoFcBbAIwjKgA+JclTroAvvXQBjOJVJQAZgmRoGWfEQDkxOJwj4tAGnVRmrKACYAjBKkB6eyQjQAFnADuUYK+h4IwBDSgXje4G6MxFAABubRZlEAVuwKjFAcPHzojlAQAG4QlN7u7NSuUB7QnIwgUK6MBVAARFSBlP54TbjA3oi8sAgA2lrmWgC66KCQUAAqyogoUxBwikoItFDZTthQAHoA-OhbdXzQlcKMZD3ctYEANnei3lEe9T0A1lShK1B4jMBpACSFUYlGAx3mll6wjgZAgxwAFJ92lAfixKHxGI8AF5PHzQJb7ACUMjkCkhRHgCzUO00UB0egMeGMpnMRBsdiOsnkPTgrko1IwtMI9N0+kMJh2bKUWOI0EiUHYlHeAo8RQVVOGozGUBy7i80kiwCedwCURSv0QPlEUR8-wqfjkWh6GOkEGIxEuODutQ8oPB4iOS1mfIFqiWP1DcHoOR2ByD4VmkPD4R+kJj2yw8aAA)

apps/meteor/server/models/raw/Roles.ts
```diff
@@ -179,6 +179,15 @@ export class RolesRaw extends BaseRaw<IRole> implements IRolesModel {
 		return this.find(query, options || {});
 	}
 
+	findCustomUserRoles(options?: FindOptions<IRole>): FindCursor<IRole> {
+		const query: Filter<IRole> = {
+			scope: 'Users',
```
matheusbsilva137, line 184: I'm wondering if we should use only "Users" scoped roles for the stats. Maybe we'd like to count both ("Users" and "Rooms") scopes or split them in two different stats. Let's check that.

matheusbsilva137 (changes requested):

apps/meteor/server/models/raw/Roles.ts
```diff
@@ -179,6 +179,14 @@ export class RolesRaw extends BaseRaw<IRole> implements IRolesModel {
 		return this.find(query, options || {});
 	}
 
+	findCustomRoles(options?: FindOptions<IRole>): FindCursor<IRole> {
+		const query: Filter<IRole> = {
+			protected: false,
```
matheusbsilva137, line 184: I see there's usually no huge amount of data stored in this collection, but I still wonder if we should add an index here. wdyt @sampaiodiego ?
Other than that, this looks fine to me

hugocostadev (approved):

LucianoPierdona (approved):

apps/meteor/server/models/raw/Roles.ts
LucianoPierdona: This shouldn't affect the QA, I'm approving this one and we can go back on this one for the external review

pierre-lehnen-rc: we should add one, but it's not a blocking requirement for this PR

pierre-lehnen-rc (approved):

PR #27798, opened 2023-01-20
[NEW] New API endpoint for getting the reported messages

Dnouv:

A new API endpoint to fetch all of the reported messages within a certain timestamp.

Note: The `api/server/v1/moderation.ts` has some errors.


debdutdeb (changes requested):
please create raw models instead of meteor ones; take a look at `apps/meteor/server/models/raw`; 

I think we're also reducing the use of underscore;

please also don't create meteor methods;

overall good job 👏🏼

apps/meteor/server/startup/migrations/v283.ts
```diff
@@ -0,0 +1,9 @@
+import { addMigration } from '../../lib/migrations';
+import { upsertPermissions } from '../../../app/authorization/server/functions/upsertPermissions';
+
+addMigration({
+	version: 279,
```
debdutdeb, line 5: The version should be the same number as the file name

apps/meteor/app/api/server/v1/moderation.ts
```diff
@@ -0,0 +1,37 @@
...
+			const { count = 20, offset = 0 } = this.getPaginationItems();
+
+			const { cursor, totalCount } = oldest
+				? Reports.findReportsBetweenDates(latest ? new Date(latest) : new Date(), new Date(oldest), offset, count)
+				: Reports.findReportsBeforeDate(latest ? new Date(latest) : new Date(), offset, count);
+
+			const [reports, total] = await Promise.all([cursor.toArray(), totalCount]);
+
+			if (!reports) {
+				return API.v1.failure('No reports found');
```
debdutdeb, line 26: No report doesn't mean it "failed". Return early with the empty list.

apps/meteor/server/models/raw/Reports.ts
```diff
@@ -18,4 +18,96 @@ export class ReportsRaw extends BaseRaw<IReport> implements IReportsModel {
...
+				ts: -1,
+			},
+			skip: offset,
+			limit: count,
+		});
+	}
+
+	findReportsBeforeDate(latest: Date, offset = 0, count = 20): FindPaginated<FindCursor<IReport>> {
+		const query = {
+			_archived: {
```
debdutdeb, line 97: hidden is a better property name IMO.

apps/meteor/tests/end-to-end/api/27-moderation.ts
```diff
@@ -0,0 +1,75 @@
...
+	describe('[/moderation.history]', () => {
+		it('should return an array of reports', (done) => {
+			request
+				.get(api('moderation.history'))
+				.set(credentials)
+				.expect('Content-Type', 'application/json')
+				.expect(200)
+				.expect((res) => {
+					expect(res.body).to.have.property('success', true);
+					expect(res.body).to.have.property('reports');
```
debdutdeb, line 21: Also verify if it actually is an array.

debdutdeb (changes requested):

debdutdeb:
I'll review in a couple of hours.

Dnouv:
Thank you! I have fixed it as per the comments.

PR #27809, opened 2023-01-23
Chore: Changed Magic numbers to proper name and imported 'BIO_TEXT…

jain-arihant-2002:
…_MAX_LENGTH' from constraints file (issue#24968)


ggazzo (changes requested):
if you believe this actually an improvement, please at least share the const between both functions. 

btw its not a magic number (maybe missing some comment around explaining, feel free to add) its a limitation from mongo DB since we use $text.

btw: this not a improve in rocket.chat terms, its only a `chore` the final users dont use this variable so they dont need to be informed about the change as an improvement.

jain-arihant-2002:
Updated the pr title and understood your point.

Issue #27811, opened 2023-01-23
cryptographically weak pseudo-random number used


WYGIN:
### Steps to reproduce:

1. go to this [link](https://github.com/RocketChat/Rocket.Chat/blob/16eacbf94fa8514e2d0f96012b1748de3fa1d220/apps/meteor/app/lib/server/lib/PasswordPolicyClass.js#L163) 
2.  on line 163 you will find password is generated by maths.random which is cryptographically weak.
3. [about it](https://cwe.mitre.org/data/definitions/338.html) 

### Expected behavior:

Using cryptographically strong password.

### Server Setup Information:

- Version of Rocket.Chat Server: 
- Operating System: 
- Deployment Method: 
- Number of Running Instances: 
- DB Replicaset Oplog: 
- NodeJS Version: 
- MongoDB Version:

### Client Setup Information

- Desktop App or Browser Version:
- Operating System:

xxxmadxxx:
Hi @WYGIN,

Please, report to us at https://hackerone.com/rocket_chat so we can have our security team working on this.

Thanks!

WYGIN:
@xxxmadxxx thanks for suggestion but i already made a PR on the issue so if everything looks perfect you can check the PR & merge it

PR #27812, opened 2023-01-23
Chore: cryptographically secure pseudo-random number generator

WYGIN:

1. removed Maths.random() to generate password
2. Added Password generator to generate cryptographically secure password including numbers, special characters & not repeating the same character again

## Issue(s)

#27811


WYGIN:
This is my first PR please review quickly.

KevLehman (commented):
Hey @sampaiodiego , for me, this looks good. However, I'm not familiar with password policies, so maybe you can take a look.

Don't think there's nothing worrysome, apart from the `generate-password` dep 🤔 which I haven't used b4 (so don't know the code)

PR #27830, opened 2023-01-24
[FIX] Display Avatars switch in messages #26741

anikdhabal:
Fix Display Avatar Switch

- Add userpreferences


dougfabris:
@anikdhabal Thanks for your help! We're going to follow with an older PR #27072 that solves the same issue!

PR #27858, opened 2023-01-26
[NEW] Notify admins via rocket.cat when a user requests to use an app

matheuslc:
This PR adds a new feature on top of the app requests feature to notify all admins whenever a user requests to use an app. The notification happens via rocket.cat user.  

## Issue(s)

[MKP-197](https://rocketchat.atlassian.net/browse/MKP-197)
[MKP-224](https://rocketchat.atlassian.net/browse/MKP-224)

### Requirements
* You need a workspace connected to the cloud
* You need at least one admin user
* You need at least one non-admin user (end-user)

### Requesting an app
Logged as an end user, access the marketplace, and request an app. 

Example:

Screenshot 2023-01-26 at 14 20 42

### Notification
As soon as the user sends the request, admins will receive a notification.

Screenshot 2023-01-26 at 15 54 52

## Further comments

[MKP-197]: https://rocketchat.atlassian.net/browse/MKP-197?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[MKP-224]: https://rocketchat.atlassian.net/browse/MKP-224?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/meteor/app/apps/server/communication/rest.js
```diff
@@ -794,6 +796,39 @@ export class AppsRestApi {
...
 		);
 
+		this.api.addRoute(
+			'notify-admins',
+			{ authRequired: true },
+			{
+				async post() {
+					const { appId, appName, message } = this.bodyParams;
+					const workspaceUrl = settings.get('Site_Url');
+					const learnMore = `${workspaceUrl}marketplace/explore/info/${appId}`;
```
graywolf336, line 806: Sometimes the `Site_Url` will end with a slash and sometimes it won't. I highly recommend ensuring it has one and putting it there if it doesn't. However, my information could be wrong, so I'd like input from others :)

matheuslc: @graywolf336 we can ensure that using a regex that matches only at the end of the string. What do you think?

It would be something like that:
```javascript

const example = "http://localhost/"
const regex = new RegExp('\\/$', 'gm')
const result = str.replace(regex, subst);

console.log('Result:', result)
```

This way we ensure the site url doesn't have a / at the end of the url and we can add them.

matheuslc: Digging more into the code, I've found out that they assume Site_Url always doesn't have the / at the end 🤔 But we may guarantee that through my suggestion above

graywolf336: Yeah, I'd rather error on the side of caution than just making an assumption. However, someone more experienced can provide more context and thoughts.

rique223 (approved):
Nothing new on the Rocket.Chat front.

PR #27860, opened 2023-01-26
feat: [ENTERPRISE] Automatic deletion of LDAP synced users already deleted in AD

heitortanoue:
## Issue Description
Feature that automatically deletes LDAP-synced users that were already deleted in the AD.

## Steps to Reproduce
1. Setup LDAP integration with RC
2. Sync your AD users to RC and confirm they all sync fine.
3. Delete a user in your AD

## Expected Behavior
Since the synced user has been deleted in AD and no longer exists there, the user should automatically be deleted in RC at the next sync job.

## Actual Behavior
The synced user has to be manually deleted in RC. So if you have a lot of users deleted in the AD, you will have to manually delete each of them in RC.

## Test Plan
1. Setup LDAP integration with RC;
2. Turn on Background Sync on the admin panel (Admin -> LDAP -> Enterprise -> Background Sync);
3. On the same pane, turn on the options `Background Sync Import New Users` and `Background Sync Remove Deleted Users`;
4. Sync your AD users to RC and confirm they all sync fine;
5. Delete a user in your AD;
6. Sync the LDAP again.

**IMPORTANT:** in order to avoid mass deletions by mistake, we only delete RC users when at least one user could be found via LDAP.

TC-383


LucianoPierdona (dismissed):

LucianoPierdona (dismissed):

apps/meteor/ee/server/lib/ldap/Manager.ts
```diff
@@ -553,18 +555,37 @@ export class LDAPEEManager extends LDAPManager {
...
+				this.updateExistingUser(ldapUser, user, converter);
+			} else if (removeDeletedUsers && !ldapUser) {
+				await this.removeDeletedUser(user);
 			}
 		}
 	}
 
+	private static updateExistingUser(ldapUser: ILDAPEntry, user: IUser, converter: LDAPDataConverter): void {
+		const userData = this.mapUserData(ldapUser, user.username);
+		converter.addUser(userData);
```
pierre-lehnen-rc, line 582:
```suggestion
		converter.addUserSync(userData);
```

pierre-lehnen-rc (changes requested):

pierre-lehnen-rc:
This change will make every user that connected from LDAP be deleted, even if they also logged in through other services. This will also delete DMs with them and assign new owners to groups they own. Is this really what we want?

We should at least add a warning to the setting description.

apps/meteor/ee/server/lib/ldap/Manager.ts
```diff
@@ -553,18 +555,37 @@ export class LDAPEEManager extends LDAPManager {
...
+
 		for await (const user of users) {
 			const ldapUser = await this.findLDAPUser(ldap, user);
 
-			if (ldapUser) {
-				const userData = this.mapUserData(ldapUser, user.username);
-				converter.addUserSync(userData);
+			if (updateExistingUsers && ldapUser) {
+				this.updateExistingUser(ldapUser, user, converter);
+			} else if (removeDeletedUsers && !ldapUser) {
```
pierre-lehnen-rc, line 574: ldapUser being undefined doesn't mean the user was deleted. If the ldap search returns multiple entries, `findLDAPUser` will also return undefined. There are some situations where LDAP may return multiple entries for the same user - we don't support this situation for login, but we should probably only delete users when they really weren't found.

pierre-lehnen-rc (changes requested):

apps/meteor/app/lib/server/functions/getRoomsWithSingleOwner.ts
```diff
@@ -64,11 +64,11 @@ export async function getSubscribedRoomsForUserWithDetails(
 				}
 
 				// If there's no subscriber available to be the new owner and it's not a public room, we can remove it.
-				if (!roomData.shouldChangeOwner && roomData.t !== 'c') {
+				if (!roomData.shouldChangeOwner) {
 					roomData.shouldBeRemoved = true;
 				}
 			}
-		} else if (roomData.t !== 'c') {
```
matheusbsilva137, line 71: This is probably by design. Should we really change this behavior?

matheusbsilva137 (changes requested):

apps/meteor/app/lib/server/functions/getRoomsWithSingleOwner.ts
```diff
@@ -64,11 +64,11 @@ export async function getSubscribedRoomsForUserWithDetails(
 				}
 
 				// If there's no subscriber available to be the new owner and it's not a public room, we can remove it.
-				if (!roomData.shouldChangeOwner && roomData.t !== 'c') {
+				if (!roomData.shouldChangeOwner) {
```
pierre-lehnen-rc, line 67:
```suggestion
				if (!roomData.shouldChangeOwner && roomData.t !== 'c') {
```
Changing this would be a change in behavior, out of the scope of this PR.

apps/meteor/app/lib/server/functions/getRoomsWithSingleOwner.ts
pierre-lehnen-rc: Reopening this comment to confirm that it is by design and that the behavior should not be changed in the scope of this PR.

pierre-lehnen-rc (changes requested):

pierre-lehnen-rc (commented):
I'm concerned about this feature. An accidental misconfiguration in the ldap settings could cause all users to be removed from rocket.chat (along with channels in some situations). I think we should at least have some safety checks like ensuring that the LDAP search is not coming back empty before we start removing users.

apps/meteor/app/lib/server/functions/getRoomsWithSingleOwner.ts
matheusbsilva137: Fixed, @totoi690 will add it to another PR

apps/meteor/app/lib/server/functions/getRoomsWithSingleOwner.ts
matheusbsilva137: This will be chenged in another PR

pierre-lehnen-rc (changes requested):
We've decided not to add this feature to Rocket.Chat. Instead, we should add a separate button on the LDAP page where admins can click to delete such users, but never in an automatic manner.
This new button should also warn the admin how many users are going to be deleted, with an option to abort the process.

PR #27753, opened 2023-01-13
[BREAK] Move Multiple Department Creation to EE & [NEW] Department Archive

MartinSchoeler:
## Proposed changes (including videos or screenshots)

[BREAK] **Move Multiple Department Creation to EE**: Limit max department creation to 1 in Community edition, existing departments will still function as before.

[NEW] 📁 **Department archival**: Adds the ability to archive departments, archived departments are disabled and cannot be edited, but will still show in historical data (current chats, directory, chat info, etc...). This offers an alternative to department deletion.

Captura de Tela 2023-02-03 às 11 44 42
Captura de Tela 2023-02-03 às 11 45 15
Captura de Tela 2023-02-09 às 19 49 15
Captura de Tela 2023-02-09 às 19 49 22
Captura de Tela 2023-02-09 às 19 49 29
Captura de Tela 2023-02-09 às 19 49 54

OC-542
OC-543
OC-609
OC-610
OC-611
OC-659

PR #27861, opened 2023-01-26
[NEW] Department Archivation

MartinSchoeler:
OC-609
OC-610
OC-611

Requires https://github.com/RocketChat/Rocket.Chat/pull/27753


apps/meteor/app/livechat/imports/server/rest/departments.ts
```diff
@@ -147,12 +149,87 @@ API.v1.addRoute(
 	},
 );
 
+API.v1.addRoute(
+	'livechat/department/archived',
+	{
+		authRequired: true,
+		// validateParams: { GET: isGETLivechatDepartmentProps},
```
KevLehman, line 156: 🤔

apps/meteor/app/livechat/imports/server/rest/departments.ts
```diff
@@ -147,12 +149,87 @@ API.v1.addRoute(
 	},
 );
 
+API.v1.addRoute(
+	'livechat/department/archived',
```
KevLehman, line 153: Since this is a get, rename to `livechat/departments/archived`.

BTW, do we need an endpoint for this? Not sure if we can have this as part of current `find` endpoint with just an extra param (so it can search archived deps too)

apps/meteor/app/livechat/imports/server/rest/departments.ts
```diff
@@ -147,12 +149,87 @@ API.v1.addRoute(
...
+	},
+);
+
+API.v1.addRoute(
+	'livechat/department/:_id/archive',
+	{
+		authRequired: true,
+		permissionsRequired: {
+			POST: { permissions: ['manage-livechat-departments'], operation: 'hasAny' },
+		},
```
KevLehman, line 191: When there's only one permission, and one endpoint, you can use the short form of the setting :awesome:
`permissionsRequired: ['manage-livechat-departments'],`

apps/meteor/app/livechat/imports/server/rest/departments.ts
```diff
@@ -147,12 +149,87 @@ API.v1.addRoute(
...
+	},
+	{
+		async post() {
+			check(this.urlParams, {
```
KevLehman, line 195: URL params are "safe" in sense they'll always be a string 🤔 

(If they aren't, for some reason, it would break before reaching here anyways so we can remove the checks for `_id`)

apps/meteor/app/livechat/imports/server/rest/departments.ts
```diff
@@ -147,12 +149,87 @@ API.v1.addRoute(
...
+				_id: String,
+			});
+
+			if (Livechat.archiveOrUnarchiveDepartment(this.urlParams._id, true)) {
```
KevLehman, line 199: 🤔 Would say it's better to have 2 methods, one for archive and other for unarchive

apps/meteor/ee/app/livechat-enterprise/server/lib/LivechatEnterprise.js
```diff
@@ -228,6 +228,12 @@ export const LivechatEnterprise = {
 				});
 			}
 		}
+		console.log(department?.archived, departmentData.enabled);
```
KevLehman, line 231: :trollface:

apps/meteor/ee/app/livechat-enterprise/server/lib/LivechatEnterprise.js
```diff
@@ -228,6 +228,12 @@ export const LivechatEnterprise = {
...
+		if (department?.archived && departmentData.enabled) {
```
KevLehman, line 232: Is enabling the only action that we'll prevent on archived deps? Can we do other stuff like changing agents, names, settings, etc 👀 

Otherwise, would be good to just block everything apart from removing agents 🤔

apps/meteor/server/models/raw/LivechatDepartment.ts
```diff
@@ -114,4 +114,12 @@ export class LivechatDepartmentRaw extends BaseRaw<ILivechatDepartmentRecord> im
 
 		return this.updateMany(query, update);
 	}
+
+	archiveOrUnarchiveDepartment(_id: string, archive: boolean): Promise<Document | UpdateResult> {
+		if (archive) {
+			this.updateOne({ _id }, { $set: { enabled: false } });
```
KevLehman, line 120: missing await? :oh-no:

KevLehman (changes requested):
Great work martin! 🥳

apps/meteor/ee/app/livechat-enterprise/server/lib/LivechatEnterprise.js
MartinSchoeler: I've done it in this way since there are a couple of operations that should be available for archived departments (Like removing agents, removing from units/tags, etc...)

PR #27867, opened 2023-01-27
[BREAK] Remove message view mode from User Preferences

hugocostadev:

Since the view mode setting in `User preferences -> Messages -> View Mode` does not work with the new messages template it will be removed

It was removed from every file and it was created a migration to remove from user obj in DB

## Further comments

TC-367


sampaiodiego (commented):
This is a `[BREAK]` change right?

hugocostadev:
> This is a `[BREAK]` change right?

I'm not sure since this feature is not working in the new message-parser and the old parser (Legacy Message Template) is going to be removed... That PR will remove should be `[BREAK]`, right?  WDYT?

hugocostadev:
> > > This is a `[BREAK]` change right?
> > 
> > 
> > I'm not sure since this feature is not working in the new message-parser and the old parser (Legacy Message Template) is going to be removed... That PR will remove should be `[BREAK]`, right? WDYT?
> 
> it is not working due to a bug (from the user perspective).. but actually the feature (to control the message rendering size) is being removed (it was already silently removed actually, since the new message components do not support it), which means it is a breaking change.

Got it! Thanks a lot for you explanation, I'll edit it here! ;D

apps/meteor/server/startup/migrations/v283.ts
```diff
@@ -0,0 +1,21 @@
+import { Settings } from '@rocket.chat/models';
+
+import { addMigration } from '../../lib/migrations';
+import { Users } from '../../../app/models/server';
+
+addMigration({
+	version: 283,
+	up() {
+		Users.update(
```
gabriellsh, line 9: I'm not sure if it is worth to have a migration for the whole userbase (open has about 300k users for example). @sampaiodiego wdyt?

yash-rajpal (approved):

apps/meteor/server/startup/migrations/v283.ts
rodrigok: I'd prefer to cleanup the data. But I agree it may be heavy since we don't have an index for this field.

yash-rajpal (dismissed):

rodrigok:
@hugocostadev can you solve the conflict?

yash-rajpal (approved):

rodrigok (approved):

rodrigok:
@yash-rajpal can you review this PR again?

yash-rajpal (approved):

PR #27868, opened 2023-01-27
yarn lock package added

Yash-Sajwan24:
Added yarn lock to the project.
This will allow anyone else on your team, your deployments, your CI/continuous integration, and anyone else who runs npm install in your package source to get the exact same dependency tree that you were developing on.


ggazzo (changes requested):
we dont use npm thats why we dont have package-lock

ggazzo:
thanks for the explanation how `lock` works btw

Yash-Sajwan24:
Thanks for the info @ggazzo  . Just one small doubt I make my subsequent pull request now it will contain this package so will my pull request be rejected?

PR #27877, opened 2023-01-27
[NEW] Bridge to get all users crated by app and also delete them

albuquerquefabio:

We're working on [MS Teams bridge](https://github.com/RocketChat/Apps.teams.bridge), to make it usable we have created a new method for the apps engine to [create new users](https://github.com/RocketChat/Rocket.Chat.Apps-engine/pull/536) and we'll also need to delete all when we uninstall the app. Therefore, we created this bridge to cover the action of deleting users on the app side.


debdutdeb:
Should we depend on the app developer to do this or have the engine take care of this, like the other stuff- persistence, its own user etc? 

https://github.com/RocketChat/Rocket.Chat.Apps-engine/blob/d017731aff04890c40750edeb5ad007b25457619/src/server/AppManager.ts#L549-L564

I live the general idea of having the capability of deleting a user, but more on the generic path than just `deleteUsersCreatedByApp`, like `deleteUser`?

albuquerquefabio:
> Should we depend on the app developer to do this or have the engine take care of this, like the other stuff- persistence, its own user etc?
> 
> https://github.com/RocketChat/Rocket.Chat.Apps-engine/blob/d017731aff04890c40750edeb5ad007b25457619/src/server/AppManager.ts#L549-L564
> 
> I like the general idea of having the capability of deleting a user, but more on the generic path than just `deleteUsersCreatedByApp`, like `deleteUser`? not sure

So, we are preparing the Rocket.Chat and Apps Engine to support the App MS Teams bridge, these updates were necessary.

Let me explain some points:
- We need to allow the apps engine to create only bot users;
- We want to put the responsibility on the dev to call the method to exclude your bots;
- When you uninstall the app it will run `remove(user: IUser & { id: string }, appId: string)` where appId is just to mount a logs notification, and we call a method `deleteUser` which will unsubscribe the user from many places;
- `getUserCreatedByApp` is necessary to bring all users by appId and type, the next step is to run a loop and get each item to execute `deleteUser` step.
- I believe that we can improve the name of the method, but we should explain that this method only deletes the bot or app user

AllanPazRibeiro (dismissed):
LGTM

AllanPazRibeiro (approved):
LGTM

PR #27875, opened 2023-01-27
improve: Use workspace token when interacting with Marketplace

AlenDavid:
[IMPROVE] Use workspace token instead of user token when calling marketplace.

PR #27820, opened 2023-01-24
[IMPROVE] Registration Experience

csuadev:
## Proposed changes (including videos or screenshots)

If you didn't register when you initially setup your server.  Its now easier to register.

Disconnecting and reconnecting your workspace from the cloud is now also easier

PR #27896, opened 2023-01-30
improve: Remove Cloud Login restrictions from Marketplace

PedroRorato:

Along side improving the connectivity services flow we no longer need to require Users to login to their Cloud Account in order to buy apps


apps/meteor/client/views/admin/apps/AppsPage/AppsPage.tsx
```diff
@@ -25,49 +17,13 @@ const AppsPage = ({ isMarketplace }: AppsPageProps): ReactElement => {
...
-							{isLoggedInCloud === undefined ? (
-								<Skeleton width='x80' />
-							) : (
-								<>
-									<Icon name='download' /> {t('Login')}
-								</>
-							)}
-						</Button>
-					)}
-					{Boolean(isDevelopmentMode) && (
```
geekgonecrazy, line 64: removing this button I believe is out of scope here.  This is still something that needs to remain

PedroRorato: ok

geekgonecrazy:
Merging this down to the feature branch.  Needs #27875 and #27820 in order to be testable so can't be merged on its own.

PR #27900, opened 2023-01-31
[Fix] add URL preview

EduardoGurgel:
add URL preview fix
image
image


art1505:
@ggazzo , @sampaiodiego , @rodrigok

PR #27925, opened 2023-02-01
Chore: Notify Marketplace on App Install

graywolf336:

Notifies Rocket.Chat's Marketplace of an app install action. This correlates with the app request flow.

## Steps to test or reproduce
Mostly backend work. So, installing or updating an app shouldn't cause any errors as all errors are silenced intentionally.

## Further comments
As mentioned above, the errors are silenced/ignored. This is a secondary route for app requests and we have a backup flow for if this method fails.


apps/meteor/app/apps/server/marketplace/appInstall.ts
```diff
@@ -0,0 +1,50 @@
...
+
+export async function notifyAppInstall(marketplaceBaseUrl: string, action: installAction, appInfo: IAppInfo): Promise<void> {
+	let headers = {
+		Authorization: '',
+	};
+
+	try {
+		const token = await getWorkspaceAccessToken();
+		headers = {
+			Authorization: `Bearer ${token}`,
```
matheuslc, line 17: @graywolf336 if we got an error getting the token, I wonder if we should continue. 🤔 The siteUrl makes sense, as we can send it empty, but for the token, I don't think so 🤔

matheuslc (approved):
Awesome! 👏

tiagoevanp (dismissed):

tiagoevanp (approved):

matheuslc (approved):

PR #27933, opened 2023-02-01
feat: Presence cap UI

juliajforesti:


juliajforesti:
@sampaiodiego there are a few endpoints which the response brings the list of users with its current status (e.g. `users.presence`). We need to decide how to approach this :)

ggazzo (approved):

PR #27938, opened 2023-02-02
[IMPROVE] Add interpreter to message md

filipemarins:
## Further comments

Add an md interpreter to the message list component on react template instead on render the message.msg only.
It will allow adding more styles to pdf messages and rendering emoji images in near future.
Add bold, italic, strike, ul, and ol elements.

Screenshot 2023-02-02 at 13 18 33


packages/pdf-worker/src/templates/ChatTranscript/markup/elements/InlineElements.tsx
```diff
@@ -0,0 +1,47 @@
...
+				case 'EMOJI':
+					return <EmojiSpan key={index} {...child} />;
+
+				case 'MENTION_USER':
+				case 'MENTION_CHANNEL':
+					return <Text key={index}>{child.value?.value}</Text>;
+
+				default:
+					return null;
+			}
```
aleksandernsilva, line 42: I noticed a similar switch in a few of the files. Is it possible to center this logic into a component e reuse the same component on all of these files?

aleksandernsilva (changes requested):

aleksandernsilva (approved):

PR #27940, opened 2023-02-02
refactor: Omnichannel analytics conversion to typescript

FabioCavaleti:

In this PR I converted the omnichannel analytics files to typescript

## Further comments

OC-631


apps/meteor/client/views/omnichannel/analytics/AnalyticsPage.tsx
```diff
@@ -31,19 +40,19 @@ const useOptions = (type) => {
 const AnalyticsPage = () => {
 	const t = useTranslation();
 	const [type, setType] = useState('Conversations');
-	const [department, setDepartment] = useState(null);
-	const [dateRange, setDateRange] = useState({ start: null, end: null });
-	const [chartName, setChartName] = useState();
+	const [department, setDepartment] = useState<DepartmentType>('');
```
KevLehman, line 43: Can this default change affect something? From null to empty?

apps/meteor/client/views/omnichannel/analytics/AnalyticsPage.tsx
```diff
@@ -90,12 +104,30 @@ const AnalyticsPage = () => {
 							w='66%'
 							h='100%'
 							chartName={chartName}
-							departmentId={department?.value}
+							departmentId={
+								(
+									department as {
```
KevLehman, line 109: The fact that you need to do this means there's something wrong with the department type or some validation missing

Can it be {} | null instead of the 3-way union it is?

apps/meteor/client/views/omnichannel/analytics/AnalyticsPage.tsx
```diff
@@ -9,7 +10,15 @@ import DateRangePicker from './DateRangePicker';
...
 import Overview from './Overview';
 
-const useOptions = (type) => {
+type DepartmentType =
+	| string
+	| {
+			value: string;
+			label: string;
+	  }
+	| undefined;
```
KevLehman, line 19:
```suggestion
type DepartmentType =
	| null
	| {
			value: string;
			label: string;
	  };
```

apps/meteor/client/views/omnichannel/analytics/DateRangePicker.tsx
```diff
@@ -30,7 +42,7 @@ const DateRangePicker = ({ onChange = () => {}, ...props }) => {
 			end: range.end,
 		};
 		setRange(rangeObj);
-		onChange(rangeObj);
+		onChange?.(rangeObj);
```
KevLehman, line 45: Is onChange really optional?

apps/meteor/client/views/omnichannel/analytics/InterchangeableChart.tsx
```diff
@@ -51,9 +66,16 @@ const InterchangeableChart = ({ departmentId, dateRange, chartName, ...props })
 			if (!result?.chartLabel || !result?.dataLabels || !result?.dataPoints) {
 				throw new Error('Error! fetching chart data. Details: livechat:getAnalyticsChartData => Missing Data');
 			}
-			context.current = await drawLineChart(canvas.current, context.current, [result.chartLabel], result.dataLabels, [result.dataPoints], {
-				tooltipCallbacks,
-			});
+			context.current = await drawLineChart(
+				canvas.current as HTMLCanvasElement,
```
KevLehman, line 70: You can check if canvas.current is valid before doing this, so you don't typecast a null to this element

apps/meteor/client/views/omnichannel/analytics/InterchangeableChart.tsx
```diff
@@ -51,9 +66,16 @@ const InterchangeableChart = ({ departmentId, dateRange, chartName, ...props })
...
+				context.current as ChartType<'line', number, string>,
```
KevLehman, line 71: Same with this

apps/meteor/client/views/omnichannel/analytics/InterchangeableChart.tsx
```diff
@@ -70,7 +92,7 @@ const InterchangeableChart = ({ departmentId, dateRange, chartName, ...props })
 		});
 	}, [chartName, departmentId, draw, end, start, t, loadData]);
 
-	return <Chart border='none' pi='none' ref={canvas} {...props} />;
+	return <Chart /* border='none' pi='none'*/ ref={canvas} {...props} />;
```
KevLehman, line 95: Is the comment needed?

KevLehman (changes requested):

apps/meteor/client/views/omnichannel/analytics/AnalyticsPage.tsx
FabioCavaleti: Hey Kevin I remember I checked that the application is handling these values as string types. And when I changed for the default empty string I tested and I didn't see anything weird or different. Do you think this is enough to assume that this change isn't affecting anything?

apps/meteor/client/views/omnichannel/analytics/DateRangePicker.tsx
```diff
@@ -1,24 +1,36 @@
...
+		end: formatToDateInput(new Date(date.getFullYear(), date.getMonth() - monthsToSubtractFromToday + 1, 0)),
+	};
+};
+
+const getWeekRange = (daysToSubtractFromStart: number, daysToSubtractFromEnd: number): { start: string; end: string } => {
+	const date = new Date();
+	return {
+		start: formatToDateInput(new Date(date.getFullYear(), date.getMonth(), date.getDate() - daysToSubtractFromStart)),
+		end: formatToDateInput(new Date(date.getFullYear(), date.getMonth(), date.getDate() - daysToSubtractFromEnd)),
+	};
```
KevLehman, line 26: What was the problem with moment? 👀 any FE guideline to not to use it?

apps/meteor/client/views/omnichannel/analytics/DateRangePicker.tsx
```diff
@@ -39,12 +51,12 @@ const DateRangePicker = ({ onChange = () => {}, ...props }) => {
 			start: range.start,
 		};
 		setRange(rangeObj);
-		onChange(rangeObj);
+		onChange?.(rangeObj);
```
KevLehman, line 54: onChange doesn't look optional, is this chaining still needed?

apps/meteor/client/views/omnichannel/analytics/Overview.tsx
```diff
@@ -62,7 +63,7 @@ const Overview = ({ type, dateRange, departmentId }) => {
 							pb='x8'
 							flexBasis='100%'
 							key={i}
-							title={title ? t(title) : <Skeleton width='x60' />}
+							title={title ? t(title as TranslationKey) : <Skeleton width='x60' />}
```
KevLehman, line 66: Can we properly type this to avoid the typecast? Maybe we can pass a generic to the useState or typing the conversationsInitialData var.

wdyt @aleksandernsilva ?

KevLehman: Also, on L58, I see there's a default value for a map's first element. Is there any special case in which that default will be triggered? 👀

KevLehman (changes requested):

apps/meteor/client/views/omnichannel/analytics/DateRangePicker.tsx
FabioCavaleti: I was having problems with the type of date when I was using moment, so I search for examples about how to deal with dates on the repo and they are doing it this way. Do you think that I should turn back and use moment?

apps/meteor/client/views/omnichannel/analytics/DateRangePicker.tsx
FabioCavaleti: Nope, I forget to change haha. Sorry and thank you

KevLehman (dismissed):

filipemarins (dismissed):

PR #27941, opened 2023-02-02
[FIX] "Jump to first unread" banner does not always appear when the first unread message is outside of the visible screen

Educg550:
[TC-173]

### Steps to test or reproduce

1. Open a few channels and send messages until the first sent is outside the screen
2. The banner should be functional and always appear when you enter the channel for the first time and didn't read the content yet

## Further comments

Behavior with the banner buttons were fixed as well, since they weren't causing visible reactions to the final user

[TC-173]: https://rocketchat.atlassian.net/browse/TC-173?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


hugocostadev (dismissed):

hugocostadev (approved):

CitizenDev:
Issue still remains

PR #27942, opened 2023-02-02
[FIX] Invalid regex on search fields crashes UI

totoi690:
## Problem to be solved
`Admin -> Workspace -> Users`  entering `?` or another invalid regular expression into the search blank crashes the UI instantly.
Other Admin pages with Regex searches have shown the same error:
1. Custom Emoji
2. Custom Sounds
3. Custom User Status
4. Integrations

## Acceptance criteria
These search fields should accept valid RegEx and regular text, but showing an error message when an invalid input is given.

TC-418 TC-384


gabriellsh (dismissed):
Is there a possibility we could use
`import { escapeRegExp } from '@rocket.chat/string-helpers';`
instead of testing the regex? I think this would ensure no regex would break ever...

hugocostadev:
I agree with @gabriellsh , there is a setting that you can change the accepted chars in the username, because of that we don't have to restrict the search for certain chars.

hugocostadev (dismissed):

hugocostadev (approved):

PR #27962, opened 2023-02-07
Regression: run clientMessageRecieved hooks on thread main message

yash-rajpal:
## Issue(s)

TC-405


gabriellsh (changes requested):
If you send a message in the thread, it still shows the placeholder.

tassoevan (approved):

PR #27967, opened 2023-02-07
[FIX] Use fibers on async calls

thassiov:

### DIFF
--- a/apps/meteor/ee/app/apps/bridges/internal.ts
+++ b/apps/meteor/ee/app/apps/bridges/internal.ts
@@ -2,6 +2,7 @@ import { InternalBridge } from '@rocket.chat/apps-engine/server/bridges/Internal
 import type { ISetting } from '@rocket.chat/apps-engine/definition/settings';
 import type { ISubscription } from '@rocket.chat/core-typings';
 import { Settings, Subscriptions } from '@rocket.chat/models';
+import Future from 'fibers/future';
 
 import type { AppServerOrchestrator } from '../orchestrator';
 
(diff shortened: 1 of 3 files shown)


apps/meteor/ee/app/apps/bridges/livechat.ts
```diff
@@ -24,7 +25,7 @@ export class AppLivechatBridge extends LivechatBridge {
 	}
 
 	protected isOnline(departmentId?: string): boolean {
-		return Livechat.online(departmentId);
+		return Future.fromPromise(Livechat.online(departmentId)).wait() as boolean;
```
KevLehman, line 28: Promise.await was not enough? :oh-no:

apps/meteor/ee/app/apps/bridges/internal.ts
```diff
@@ -16,11 +17,13 @@ export class AppInternalBridge extends InternalBridge {
 			return [];
 		}
 
-		const records = Subscriptions.findByRoomIdWhenUsernameExists(roomId, {
-			projection: {
-				'u.username': 1,
-			},
-		}).toArray();
+		const records = Future.fromPromise(
```
tapiarafael, line 20: Why can't we use the normal `async/await` here?

apps/meteor/ee/app/apps/bridges/livechat.ts
d-gubert: `Promise.await` won't be available in the Microservice scope, so we need to go deeper

apps/meteor/ee/app/apps/bridges/internal.ts
d-gubert: Because it would be a breaking change as it's available to apps through here https://github.com/RocketChat/Rocket.Chat.Apps-engine/blob/alpha/src/server/rooms/Room.ts#L29

d-gubert (approved):

PR #27969, opened 2023-02-08
[BREAK] Unify monolith and microservices intercommunication on same technology and limit to Enterprise only 

sampaiodiego:

This change removes the old DDP Matrix communication by Moleculer based breaker over TCP, simplifying code maintenance, allowing better reliability, performance and monitoring.

It requires a license to enable communication among servers as the rest of the Moleculer implementation.

Community Edition will loose ephemeral events as Typing indication and others when running multiple instances.

## Further comments

It's still missing to implement the ability to disable matrix broadcast via existing setting.


apps/meteor/app/api/server/v1/instances.ts
```diff
@@ -17,18 +15,43 @@ API.v1.addRoute(
...
+				// 		connection,
+				// 	};
+				// }),
 			});
 		},
 	},
 );
+
+// TODO why?
+// function getConnection(address) {
```
rodrigok, line 36: I believe we used this endpoint for debug only, so it was returning all the relevant information about the instances and the connection status

ggazzo (approved):

PR #27970, opened 2023-02-08
Regression: `Admin UI -> Rooms` Search filter not working

totoi690:
## Issue Description
In Rocket.Chat it is possible to create channels with UTF-8 and non-Latin characters (Cyrillic or Greek for example) but the admin interface cannot find them.
The channel with Latin characters finds it well.

## Steps to Reproduce
1. Enable `UTF8 Names Validation` on `Admin UI → General`
2. Enable `Allow Special Characters in Room Names` on `Admin UI → Layout → User Interface`
3. Create a channel and private room with Cyrillic names, like `тестовый канал` and `тестовый чат` or Greek `Ελληνικά`
4. Go to Admin UI - Rooms and try to find that room and channel

## Expected Behavior
Non-Latin search should work in the `Admin → Rooms ` area and any other search field in Rocket.Chat interface even when the option `Allow Special Characters in Room Names` is enabled.

TC-385, TC-444, TC-499


apps/meteor/app/api/server/v1/rooms.js
```diff
@@ -333,14 +333,15 @@ API.v1.addRoute(
 		get() {
 			const { offset, count } = this.getPaginationItems();
 			const { sort } = this.parseJsonQuery();
-			const { types, filter } = this.requestParams();
+			const { types, filter, useFname } = this.requestParams();
```
gabriellsh, line 336: This here is something I have some questions about.

Should we expect this value to come from the request, or should we get it directly from the settings? (even if we expect it to come from the request, I think we should get the setting and overwrite the value if it is sent through a parameter)

gabriellsh (changes requested):
We can work on this together if you have questions. Please dm me if you need.

apps/meteor/server/models/raw/Rooms.js
```diff
@@ -75,13 +75,16 @@ export class RoomsRaw extends BaseRaw {
 					},
 			  };
 
+		const fnameQuery = useFname ? [{ fname: nameRegex }] : [];
```
LucianoPierdona, line 78: shouldn't it be like this?
```suggestion
		const fnameQuery = useFname ? { fname: nameRegex } : {};
```

apps/meteor/server/models/raw/Rooms.js
```diff
@@ -128,10 +131,13 @@ export class RoomsRaw extends BaseRaw {
 
 		const onlyTeamsCondition = onlyTeams ? { $and: [{ teamMain: { $exists: true } }, { teamMain: true }] } : {};
 
+		const fnameQuery = useFname ? [{ fname: nameRegex }] : [];
```
LucianoPierdona, line 134: same here
```suggestion
		const fnameQuery = useFname ? { fname: nameRegex } : {};
```

LucianoPierdona (changes requested):
Small points, otherwise backend changes looks good

LucianoPierdona (approved):

hugocostadev (dismissed):

apps/meteor/server/models/raw/Rooms.js
```diff
@@ -62,7 +62,7 @@ export class RoomsRaw extends BaseRaw {
 		return statistic;
 	}
 
-	findByNameContainingAndTypes(name, types, discussion = false, teams = false, showOnlyTeams = false, options = {}) {
+	findByNameOrFnameContainingAndTypes(name, types, discussion = false, teams = false, showOnlyTeams = false, options = {}) {
```
rodrigok, line 65: When you change a function name you need to change all the places that uses it

rodrigok (changes requested):

apps/meteor/server/models/raw/Rooms.js
gabriellsh: I accidentaly forgot to add this file in the commit. Thanks!

hugocostadev (approved):

LucianoPierdona (approved):

gabriellsh (approved):

tassoevan (approved):

PR #27977, opened 2023-02-08
[FIX] Clicking on view full icon navigates user to Home Screen

filipemarins:
## Further comments

Check if the data came from `room.findOne` is truly undefined to prevent redirect while `useReactiveQuery` still fetching.

### DIFF
--- a/apps/meteor/client/views/room/Header/Omnichannel/QuickActions/hooks/useQuickActions.tsx
+++ b/apps/meteor/client/views/room/Header/Omnichannel/QuickActions/hooks/useQuickActions.tsx
@@ -203,13 +203,15 @@ export const useQuickActions = (
 						  }
 						: { transcriptEmail: { sendToVisitor: false } }),
 				});
+				homeRoute.push();
+				RoomManager.close(room.t + rid);
 				closeModal();
 				dispatchToastMessage({ type: 'success', message: t('Chat_closed_successfully') });
 			} catch (error) {
 				dispatchToastMessage({ type: 'error', message: error });
 			}
 		},
-		[closeChat, closeModal, dispatchToastMessage, rid, t],
+		[closeChat, closeModal, dispatchToastMessage, homeRoute, room.t, rid, t],
 	);
 
 	const returnChatToQueueMutation = useReturnChatToQueueMutation({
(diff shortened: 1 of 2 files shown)


apps/meteor/client/views/room/providers/RoomProvider.tsx
```diff
@@ -102,7 +102,7 @@ const RoomProvider = ({ rid, children }: RoomProviderProps): ReactElement => {
 	const api = useMemo(() => ({}), []);
 
 	if (!pseudoRoom) {
-		return roomQuery.isSuccess ? <RoomNotFound /> : <RoomSkeleton />;
+		return roomQuery.isSuccess && roomQuery.data === undefined ? <RoomNotFound /> : <RoomSkeleton />;
```
aleksandernsilva, line 105: Shouldn't you also consider the `null` value? Maybe `!roomQuery.data` to cover all falsy values.

aleksandernsilva (changes requested):

apps/meteor/client/views/room/providers/RoomProvider.tsx
filipemarins: Actually the `roomQuery.data`  is returning `null` while is still try fetching the data from `room.findOne`, it will only return `undefined` if there is no room.

I can't just do `!roomQuery.data` here because `!null` is `true` and will show the `RoomNotFound` while the data is not fully checked yet.

apps/meteor/client/views/room/Header/Omnichannel/QuickActions/hooks/useQuickActions.tsx
```diff
@@ -203,13 +203,15 @@ export const useQuickActions = (
 						  }
 						: { transcriptEmail: { sendToVisitor: false } }),
 				});
+				homeRoute.push();
+				RoomManager.close(room.t + rid);
```
KevLehman, line 207: not really sure how this works, but, is the `l{someroomid}` format valid? no `/` in the concatenation? 👀

filipemarins: maybe `RoomManager.close(`${room.t}${rid}`);`?

KevLehman: No no, i was referring to, is that `t + id` what the .close function expects? If that's the case, pls ignore me lol

filipemarins: Yeah, `t + id` is what is expected 😢

apps/meteor/client/views/room/providers/RoomProvider.tsx
aleksandernsilva: Got it

KevLehman:
Well, if @aleksandernsilva is good, I'm good

aleksandernsilva (approved):

PR #27979, opened 2023-02-08
[IMPROVE]  Add mobile version for transcript dropdown

filipemarins:
## Issue(s)

oc-646

## Further comments

Add mobile version for transcript dropdown

Screenshot 2023-02-08 at 16 30 43


ggazzo (changes requested):
I dont get it `Dropdown` already contains that logic (if desktop/mobile) I cant understand why you are implementing the same thing again, just to show a "empty list"(2 items) using 50% of the screen?

filipemarins:
> I dont get it `Dropdown` already contains that logic (if desktop/mobile) I cant understand why you are implementing the same thing again, just to show a "empty list"(2 items) using 50% of the screen?

Unfortunately, the dropdown component has different media query breakpoints for mobile `(min-width: 500px)` and we are facing some position issues in different browsers (edge and safari) for screens less than `775px` so the idea was to keep the same behavior as the toolbox dropdown menu.
Screenshot 2023-02-08 at 16 30 34

PR #27986, opened 2023-02-09
[IMPROVE] Add styles for inline and multi-line code

filipemarins:

oc-648

## Further comments

Add styles for inline and multi-line code;
Fix storybook public directory.

Screenshot 2023-02-09 at 15 10 24


packages/pdf-worker/src/strategies/ChatTranscript.ts
```diff
@@ -6,26 +6,23 @@ import exportChatTranscript from '../templates/ChatTranscript';
 import type { ChatTranscriptData, PDFMessage } from '../templates/ChatTranscript';
 
 export class ChatTranscript implements IStrategy {
-	private isNewDay(current: PDFMessage, previous: PDFMessage | undefined, timezone: string): boolean {
-		return !previous || !moment(current.ts).tz(timezone).isSame(previous.ts, 'day');
+	private isNewDay(current: PDFMessage, previous: PDFMessage | undefined): boolean {
+		return !previous || !moment(current.ts).isSame(previous.ts, 'day');
```
KevLehman, line 10: Why are you removing the timezone param?

KevLehman (changes requested):

packages/pdf-worker/src/strategies/ChatTranscript.ts
filipemarins: It was throwing an error on the storybook, but the issue was on my side, thanks for catch it :)

KevLehman (approved):

PR #27987, opened 2023-02-09
Chore: e2e changes tests

ggazzo:
basically I chased down all the flaky tests and arranged each one in its own way.

Selectors that collided
Quick actions that required loading
semantic problems

I also removed the retries, he is the first great facilitator for us to have these intermittent tests, run 'x' times and pass, okay.

At another time, everything will not be fine and we have a result that we cannot trust.


apps/meteor/tests/e2e/create-discussion.spec.ts
```diff
@@ -20,7 +20,7 @@ test.describe.serial('create-discussion', () => {
 
 		await poHomeDiscussion.sidenav.openNewByLabel('Discussion');
 		await poHomeDiscussion.inputChannelName.type('general');
-		await page.keyboard.press('Enter');
+		await page.locator('role=modal >> role=listbox >> role=option[name=general]').click();
```
ggazzo, line 23:
```suggestion
		await page.locator('role=listbox >> role=option[name=general]').click();
```

rodrigok (approved):

PR #28012, opened 2023-02-13
[BREAK] Remove Chatpal

ggazzo:
as Chatpal is no longer maintained it makes no sense that we maintain compatibility within Rocket.Chat, luckily some solution with Rocket.Chat.Apps can add some capabilities again


tassoevan (approved):

geekgonecrazy:
you mention apps?  Can we add search providers via apps now??  That would be amazing..

PR #28031, opened 2023-02-14
[FIX] Custom fields does not appear on the registration form

murtaza98:
## Issue(s)

https://rocketchat.atlassian.net/browse/OC-540


apps/meteor/app/livechat/server/api/v1/config.ts
```diff
@@ -18,17 +19,17 @@ API.v1.addRoute(
 				return API.v1.success({ config: { enabled: false } });
 			}
 
-			const { token, department, businessUnit } = this.queryParams;
+			const { token, department, businessUnit } = this.queryParams as { token?: string; department?: string; businessUnit?: string };
```
KevLehman, line 22: Since this is on TS, shouldn't the type come from rest typings?

apps/meteor/app/livechat/server/api/v1/config.ts
```diff
@@ -18,17 +19,17 @@ API.v1.addRoute(
...
 
 			const config = await cachedSettings({ businessUnit });
 
 			const status = Livechat.online(department);
-			const guest = token && (await Livechat.findGuest(token));
+			const guest: ILivechatVisitor | null = token ? await Livechat.findGuest(token) : null;
```
KevLehman, line 27: I kinda remember live chat suffered from this because it doesn't trigger the default values, can we confirm this works?

apps/meteor/app/livechat/server/api/v1/config.ts
```diff
@@ -18,17 +19,17 @@ API.v1.addRoute(
...
 
-			const room = guest && findOpenRoom(token);
+			const room: IOmnichannelRoom | undefined = guest && token ? findOpenRoom(token) : undefined;
```
KevLehman, line 29: If there's a guest, there's a token so this feels a bit redundant

KevLehman (dismissed):

apps/meteor/app/livechat/server/api/v1/config.ts
murtaza98: I know, but I had to add token there to make typescript happy!!

apps/meteor/app/livechat/server/api/v1/config.ts
murtaza98: Sorry I didn't get this? Can you pls elaborate further?

apps/meteor/app/livechat/server/api/v1/config.ts
murtaza98: I'm welcome to any suggestions about making typescript infer that its not null at this point :)

apps/meteor/app/livechat/server/api/v1/config.ts
murtaza98: Lol! I totally forgot about that :joy: Let me check and remove this

apps/meteor/app/livechat/server/api/v1/config.ts
KevLehman: maybe using `guest.token` instead of `token`? 👀

apps/meteor/app/livechat/server/api/v1/config.ts
murtaza98: Checked it and this won't affect it

KevLehman (dismissed):

apps/meteor/tests/end-to-end/api/livechat/11-livechat.ts
```diff
@@ -75,6 +77,33 @@ describe('LIVECHAT - Utils', function () {
...
+				label: customFieldName,
+				defaultValue: 'test_default_address',
+				scope: 'visitor',
+				visibility: 'visible',
+				regexp: '',
+				public: true,
+				required: false,
+
+				options: '',
+			} as unknown as ILivechatCustomField & { field: string });
```
sampaiodiego, line 94: I've found `createCustomField` being used in 5 different places, and all of them have this type cast.. it doesn't feel correct to me..

why don't you fix `createCustomField` signature instead of having to using this type cast in every place?

sampaiodiego (changes requested):

apps/meteor/app/livechat/server/api/v1/config.ts
```diff
@@ -23,12 +24,12 @@ API.v1.addRoute(
 			const config = await cachedSettings({ businessUnit });
 
 			const status = Livechat.online(department);
-			const guest = token && (await Livechat.findGuest(token));
+			const guest: ILivechatVisitor | null = token ? await Livechat.findGuest(token) : null;
```
sampaiodiego, line 27: declaring the type for `guest` is not wrong, but since TS can infer it correctly, removing it improves code readability.. wdyt?
```suggestion
			const guest = token ? await Livechat.findGuest(token) : null;
```

apps/meteor/app/livechat/server/api/v1/config.ts
```diff
@@ -23,12 +24,12 @@ API.v1.addRoute(
...
 
-			const room = guest && findOpenRoom(token);
+			const room: IOmnichannelRoom | undefined = guest ? findOpenRoom(guest.token) : undefined;
```
sampaiodiego, line 29: same here, TS can infer `room` type correctly:
```suggestion
			const room = guest ? findOpenRoom(guest.token) : undefined;
```

sampaiodiego (approved):

PR #28037, opened 2023-02-14
[BREAK] Limit `underscore` usage and remove `underscore.string` (even from integration script sandboxes)

tassoevan:

[Underscore.js](https://underscorejs.org/) is quite outdated and we're not using it in a way that take advantage of modularization, bloating the client bundle. This PR aims to limit its usage across the application.

break: if you are used to have scripts, and they are using underscore, please pay attention probably you need to update them


rodrigok:
@tassoevan why is this a breaking PR?

tassoevan:
According to @pierre-lehnen-rc, removing `underscore.string` is a breaking change for integration scripts. Look at `apps/meteor/app/integrations/server/api/api.js`.

apps/meteor/app/2fa/client/TOTPGoogle.js
```diff
@@ -25,8 +24,8 @@ const loginWithGoogleAndTOTP = function (options, code, callback) {
 	// has the proper email address after the OAuth conversation.
 
 	if (typeof Accounts._options.restrictCreationByEmailDomain === 'string') {
-		options = _.extend({}, options || {});
-		options.loginUrlParameters = _.extend({}, options.loginUrlParameters || {});
+		options = Object.assign({}, options || {});
```
KevLehman, line 27: Do we need the assign? Can we use spread? 👀

apps/meteor/app/custom-sounds/server/methods/insertOrUpdateSound.js
```diff
@@ -1,19 +1,19 @@
 import { Meteor } from 'meteor/meteor';
-import s from 'underscore.string';
 import { check } from 'meteor/check';
 import { CustomSounds } from '@rocket.chat/models';
 import { api } from '@rocket.chat/core-services';
 
 import { hasPermission } from '../../../authorization/server';
 import { RocketChatFileCustomSoundsInstance } from '../startup/custom-sounds';
+import { trim } from '../../../../lib/utils/stringUtils';
```
KevLehman, line 8: Would be cool to have this on string-helpers package

apps/meteor/app/custom-sounds/server/methods/insertOrUpdateSound.js
```diff
@@ -1,19 +1,19 @@
...
 
 Meteor.methods({
 	async insertOrUpdateSound(soundData) {
 		if (!hasPermission(this.userId, 'manage-sounds')) {
 			throw new Meteor.Error('not_authorized');
 		}
 
-		if (!s.trim(soundData.name)) {
+		if (!trim(soundData.name)) {
```
KevLehman, line 16: soundData.name?.trim() can't be used?

apps/meteor/app/lib/server/methods/getChannelHistory.ts
```diff
@@ -39,12 +39,12 @@ Meteor.methods({
...
 
 		// Ensure latest is always defined.
-		if (_.isUndefined(latest)) {
+		if (latest === undefined) {
 			latest = new Date();
 		}
 
 		// Verify oldest is a date if it exists
-		if (!_.isUndefined(oldest) && !_.isDate(oldest)) {
+		if (oldest !== undefined && !_.isDate(oldest)) {
```
KevLehman, line 47: Can we use Date.parse to check?

rodrigok:
> According to @pierre-lehnen-rc, removing `underscore.string` is a breaking change for integration scripts. Look at `apps/meteor/app/integrations/server/api/api.js`.

Can you make it clear at the title and description then?

apps/meteor/app/2fa/client/TOTPGoogle.js
tassoevan: Maybe. I didn't like to take the risk because object spread and `Object.assign()` work different when there are property setters involved.

apps/meteor/app/custom-sounds/server/methods/insertOrUpdateSound.js
tassoevan: This `trim` implementation also handles the case of incorrect value types (e.g. `123`), coercing as needed. There is some risk.

apps/meteor/app/lib/server/methods/getChannelHistory.ts
tassoevan: `_.isDate()` is a type guard that uses `Object.prototype.toString` to validate the object type, while `Date.parse` will try to coerce a string whenever possible. Very different.

PR #28046, opened 2023-02-15
Regression: Emoji picker styling

yash-rajpal:
### Before
image

### After
image


hugocostadev (approved):

sampaiodiego:
what if we have a e2e test to prevent a future regression breaking the emoji picker?

yash-rajpal:
> what if we have an e2e test to prevent a future regression breaking the emoji picker?

We already have an e2e test for testing emoji picker, it is in `emojis.spec.ts`.

sampaiodiego:
right, so it needs to be improved to be able to catch other regressions like this

yash-rajpal:
> Why didn't it fail with the regressions then? 👀

Because end-to-end tests use selectors to check if element is present in DOM as expected. In this case, all the emojis are present on DOM, just due to some missing css, they are not visible to the user, that is why we were not able to catch this regression using these tests.

I'll try to improve this for such cases.

PR #28048, opened 2023-02-15
Regression: Avoid rendering unsupported media on PDFs & update quote styling

KevLehman:

- Fix faulty logic when deciding which attachment to render based on the mime type of it

- Replace media query syntax `(width < 767px)` with `(max-width: 767px)` since the first one isn't supported on safari
- Fix a bug when has bold text in multiple rows.

## Further comments

Add quotes style:
Screenshot 2023-02-16 at 22 31 58

Before:
Screenshot 2023-02-17 at 01 21 37
After:
Screenshot 2023-02-17 at 01 24 37

Since the library doesn't have support for inline styles and on the `Text` component margin, padding, and border won't work so we needed to simplify the inline code style component:
before:
Screenshot 2023-02-17 at 01 25 16

after:
Screenshot 2023-02-17 at 01 15 42

OC-674 OC-666 OC-646 OC-675 OC-694


.vscode/launch.json
```diff
@@ -0,0 +1,14 @@
+{
```
murtaza98, line 1: Pushed this by mistake? ;-)

murtaza98 (approved):
LGTM!

aleksandernsilva (approved):

murtaza98 (dismissed):

aleksandernsilva (approved):

apps/meteor/app/theme/client/imports/general/base_old.css
```diff
@@ -1815,7 +1815,7 @@
 	}
 }
 
-@media (width <= 780px) {
+@media (max-width: 767px) {
```
ggazzo, line 1818: I don't understand, why are these changes relevant?

apps/meteor/client/sidebar/RoomList/RoomList.tsx
```diff
@@ -104,7 +104,7 @@ const RoomList = (): ReactElement => {
 			padding-block-start: 12px;
 		}
 
-		@media (width <= 400px) {
+		@media (max-width: 400px) {
```
ggazzo, line 107: same

ggazzo (changes requested):

apps/meteor/app/theme/client/imports/general/base_old.css
KevLehman: Not fully related to the PR title, they're a small fix for other regression.
Filipe can explain better, but AFAIK, Safari don't support the `w < x` instruction so some breakpoints were not being applied.

filipemarins: You are right Kevin. For this specific one, I also change the breakpoint value to match the fuselage value and fix some position issues.

murtaza98 (approved):

tassoevan (approved):

PR #28050, opened 2023-02-15
Regression: Fix improper usage of useEndpointData

KevLehman:
## Further comments

OC-627


MartinSchoeler (changes requested):
Maybe we could update this to use tanstack's `useQuery`?

As we have deprecated use endpoint data

apps/meteor/client/views/omnichannel/directory/chats/ChatTable.tsx
```diff
@@ -74,7 +74,7 @@ const ChatTable: FC<{ setChatReload: Dispatch<SetStateAction<any>> }> = ({ setCh
 		}),
 	);
 
-	const { value: data, reload } = useEndpointData('/v1/livechat/rooms', query as any); // TODO: Check the typing for the livechat/rooms endpoint as it seems wrong
+	const { value: data, reload } = useEndpointData('/v1/livechat/rooms', { params: query }); // TODO: Check the typing for the livechat/rooms endpoint as it seems wrong
```
ggazzo, line 77: useEndpointData is deprecated, why not already move to useQuery? ❤️

ggazzo (changes requested):

murtaza98:
Thanks for the review! Converted useEndpointData to tanstack's query. Pls check :)

apps/meteor/client/views/omnichannel/directory/chats/ChatTable.tsx
```diff
@@ -167,12 +161,26 @@ const ChatTable: FC<{ setChatReload: Dispatch<SetStateAction<any>> }> = ({ setCh
 		[onRowClick],
 	);
 
+	const result = useCurrentChats(query);
+
+	useEffect(() => {
+		setChatReload?.(() => result.refetch);
```
MartinSchoeler, line 167: Instead of using the refetch we should invalidate the query on reload https://tanstack.com/query/v4/docs/react/guides/query-invalidation

apps/meteor/client/views/omnichannel/directory/chats/ChatTable.tsx
```diff
@@ -167,12 +161,26 @@ const ChatTable: FC<{ setChatReload: Dispatch<SetStateAction<any>> }> = ({ setCh
...
+	}, [result.refetch, setChatReload]);
+
+	if (result.isLoading) {
+		return <GenericTableLoadingTable headerCells={6} />;
+	}
+	if (result.error) {
+		console.error(result.error);
```
MartinSchoeler, line 174: do we need to log the errors here?

MartinSchoeler (changes requested):

apps/meteor/client/views/omnichannel/directory/chats/ChatTable.tsx
murtaza98: Not needed. Leftovers from local testing. Removed it

MartinSchoeler (dismissed):
Just pushed some changes, FYI when you use `invalidateQuery` you don't need an effect since the query client has no direct dependency on the used `useQuery`

apps/meteor/client/views/omnichannel/directory/OmnichannelDirectoryPage.tsx
```diff
@@ -30,7 +31,7 @@ const OmnichannelDirectoryPage = (): ReactElement => {
 
 	const handleTabClick = useCallback((tab) => (): void => directoryRoute.push({ tab }), [directoryRoute]);
 
-	const [chatReload, setChatReload] = useState();
+	const chatReload = () => queryClient.invalidateQueries({ queryKey: ['current-chats'] });
```
KevLehman, line 34: @MartinSchoeler , seems the refactor associated with this changed the "reactivity" of the view.

Previously, it was getting "refetched" everytime the page opened, now, it's getting stale until refresh, so we need a way to know that this query needs to be invalidated

murtaza98: Noticed that the `staleTime` is set to infinity which would be the reason y you had to refresh the whole page to see the latest value. I believe we need to change it to something reasonable like 10 seconds. WDYT?
https://github.com/RocketChat/Rocket.Chat/blob/de67df97dc15666b9db8627bf7bb514e6032cb83/apps/meteor/client/views/omnichannel/currentChats/hooks/useCurrentChats.ts#L13

KevLehman: I think that works. We probably need a stream that let these queries now when to refresh.
One that gets triggered when a room is taken/closed/updated etc so these areas that interact with rooms here may refetch.

Do you know if we have something similar? Otherwise, reducing stale time should work 👀 for now

MartinSchoeler (approved):

PR #28052, opened 2023-02-15
Chore: Change status label from Installed to Enabled

d-gubert:

Change app status label to "Enabled" instead of "Installed" - this is to avoid user confusion when checking what apps are counting towards the enabled limit


yash-rajpal (approved):

apps/meteor/client/views/marketplace/helpers.ts
```diff
@@ -7,7 +7,7 @@ import { Utilities } from '../../../app/apps/lib/misc/Utilities';
 import { t } from '../../../app/utils/client';
 import { dispatchToastMessage } from '../../lib/toast';
 
-export const appEnabledStatuses = [AppStatus.AUTO_ENABLED, AppStatus.MANUALLY_ENABLED, AppStatus.INITIALIZED];
```
rique223, line 10: Removing this will probably bring back this bug: https://github.com/RocketChat/Rocket.Chat/pull/27330

rique223 (changes requested):
Has this been discussed with design and product? The change from enabled to installed was purposeful for one of our initiatives.

apps/meteor/client/views/marketplace/helpers.ts
d-gubert: Good catch. However now we cannot consider `INITIALIZED` as "Enabled" anymore. Do you have a suggestion on how to move forward with this?

rique223: We could create an "initialized" status for the UI to better represent the actual apps engine status. Wdyt?

rique223: We could create an "initialized" status for the UI to better represent the actual apps engine status. Wdyt?

d-gubert: Makes sense to me

d-gubert:
Too many conflicts, I'm closing this and opening a new one

PR #28059, opened 2023-02-15
Chore: Move apps management to ee

tapiarafael:

We are moving apps-engine code to the ee folder but keeping bridges and converters in the old folder to maintain under open source license.


apps/meteor/ee/app/license/server/license.ts
```diff
@@ -49,7 +49,7 @@ class LicenseClass {
 		 * We added a dynamic import here to avoid this issue
```
d-gubert, line 49:
```suggestion
		 * We added a dynamic import here to avoid this issue
		 * @TODO as soon as the Apps-Engine service is available, use it instead of this dynamic import
```

d-gubert (approved):

PR #28075, opened 2023-02-16
Chore: QoL improvements to Omnichannel E2E tests

KevLehman:
## Further comments

- Closes dangling contexts after use
- Run non-block api calls in parallel
- Refactor suites to be made in steps instead of one by one
- Updates selectors to be more specific
- Updates tests to dismiss generated notifications so they don't affect following suites


ggazzo (approved):

ggazzo:
is that a feature? to be marked as "feat: omnichannel" ?

ggazzo:
image

KevLehman:
> is that a feature? to be marked as "feat: omnichannel" ?

We don't have a "team only" tag. We use the tag to group things that we did so we can find them later :)

ggazzo:
> > is that a feature? to be marked as "feat: omnichannel" ?
> 
> We don't have a "team only" tag. We use the tag to group things that we did so we can find them later :)

we can create since a bunch of `feat: squad` were created 🤔

Issue #27956, opened 2023-02-06
Each launch job of the Retention Policy changes the state of the room in the database


shevchukma:
### Description:

This is not a bug, but strange behavior. Every time the task to delete old messages is run, the "_updatedAt" field in the rocketchat_room collection changes.
Is it necessary? Even if no message has been deleted, this field is used to set the task start time. This is done in EVERY document in the rocketchat_room collection. We have hundreds of thousands of documents in this collection. Running a Retention Policy will cause a heavy load on the database.

### Steps to reproduce:

1. Enable Retention Policy.
2. For clarity, set "Use Advanced Retention Policy Cron" to every 5 minutes 
`*/5 * * * *`
3. After 5 minutes, query the database to verify that the "_updatedAt" value has changed in each room:
  `db.rocketchat_room.find({ }, {"_updatedAt": 1 }).pretty()`

### Expected behavior:

The "_updatedAt" value in the "rocketchat_room" collection does not change each time the Retention Policy job runs.

### Actual behavior:

Each time the Retention Policy job runs, the value of "_updatedAt" in the "rocketchat_room" collection changes.

### Server Setup Information:

- Version of Rocket.Chat Server: 5.4.1
- Operating System: Oracle Linux Server release 8.6
- Deployment Method: docker
- Number of Running Instances: 20
- DB Replicaset Oplog: yes
- NodeJS Version: v14.19.3
- MongoDB Version: 5.0.14 

### Client Setup Information

- Desktop App or Browser Version: Chrome 108
- Operating System: Windows 10

### Additional context

By submitting this issue, I am concerned not only with the load on the database. We still have the problem that after running the retention policy job, many chats move up in the contact list. They are sorted by the "_updatedAt" field. I don't quite understand the conditions under which this happens, but I think if the field "_updatedAt" is not updated in such situations, we will get rid of this problem too.

iMiKED:
I can confirm - after deleting old messages on server side, contacts moving to top of CL by self.
This is very strange and unsuspected behaviour.

PR #28081, opened 2023-02-16
fix: Added check if count is truthy on `cleanRoomHistory`

LucianoPierdona:

This PR adds a check when the room is updated inside the function `removeByIdPinnedTimestampLimitAndUsers` which is called on `cleanRoomHistory`; It has been added because even when the room wasn't being updated, the `_updatedAt` was changed, and caused move chats to the top of the list even when nothing has changed there.

## Issue(s)

#27956

## Steps to test or reproduce

- Enable `Retention Policy`.
- For clarity, set "Use Advanced Retention Policy Cron" to every 5 minutes
`*/5 * * * *`
- After 5 minutes, query the database to verify that the `_updatedAt` value has changed in each room:
`db.rocketchat_room.find({ }, {"_updatedAt": 1 }).pretty()`

## Further comments

[TC-488]

[TC-488]: https://rocketchat.atlassian.net/browse/TC-488?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


sampaiodiego:
while I don't have to time this properly, what do you think about creating some tests?

it doesn't need to use the scheduled "history clean" that you need to wait some minutes for it to run, you can trigger the "history clean" manually and assert the room data has not changed.

LucianoPierdona:
@sampaiodiego sure, I can do that

matheusbsilva137 (dismissed):

KevLehman (approved):

sampaiodiego (approved):

PR #28096, opened 2023-02-17
feat: adds verification when inviting external matrixIds

lmauromb:
Jira: FED-78

This PR adds the ability to verify external matrix-ids when the user tries to invite them on a `federated` channel (matrix room).


apps/meteor/app/federation-v2/server/infrastructure/matrix/Bridge.ts
```diff
@@ -146,6 +146,33 @@ export class MatrixBridge implements IFederationBridge {
...
 
+	public async verifyInviteeId(externalInviteeId: string): Promise<string> {
+		const [userId, homeserverUrl] = externalInviteeId.replace('@', '').split(':');
+		try {
+			const response = await fetch(`https://${homeserverUrl}/_matrix/client/v3/register/available?username=${userId}`);
+
+			if (response.status === 400) {
+				const responseBody = await response.json();
+
+				if (responseBody.errcode === 'M_USER_IN_USE') {
```
MarcosSpessatto, line 157: Wouldn't be better to have these comparisons as constants, (e.g `const userExists = responseBody.errcode === 'M_USER_IN_USE'`)? Instead of having code comments explaining them?

apps/meteor/app/federation-v2/server/infrastructure/matrix/Bridge.ts
```diff
@@ -146,6 +146,33 @@ export class MatrixBridge implements IFederationBridge {
 		}
 	}
 
+	public async verifyInviteeId(externalInviteeId: string): Promise<string> {
```
MarcosSpessatto, line 149: I believe it would be better to have an enum defining all the possibilities rather than having an arbitrary string, what do you think?

MarcosSpessatto (changes requested):

MarcosSpessatto:
Also, would it be possible to change the base branch to `feat/federation-6.1`? In that way, we can group everything together in this branch for the final 6.1 release.

apps/meteor/app/federation-v2/server/infrastructure/matrix/Bridge.ts
lmauromb: added the constant here: [apps/meteor/app/federation-v2/server/infrastructure/helpers/MatrixIdVerificationTypes.ts](https://github.com/RocketChat/Rocket.Chat/pull/28096/files#diff-58880fcef31d139da50a34bbfc71cd6db076bfcdb96dd47f55523d60249abb5d)

apps/meteor/app/federation-v2/server/infrastructure/matrix/Bridge.ts
lmauromb: Using enums now!

[apps/meteor/app/federation-v2/server/infrastructure/matrix/Bridge.ts](https://github.com/RocketChat/Rocket.Chat/pull/28096/files#diff-d16dc61cc516cc222f5b2e375e0e74a58bee770a4e1e190cb025d2970518d0b9)

apps/meteor/app/federation-v2/server/infrastructure/matrix/Bridge.ts
```diff
@@ -146,6 +146,33 @@ export class MatrixBridge implements IFederationBridge {
...
+		const [userId, homeserverUrl] = externalInviteeId.replace('@', '').split(':');
```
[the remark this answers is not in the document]
lmauromb: created this file: [apps/meteor/app/federation-v2/server/infrastructure/helpers/MatrixIdStringTools.ts](https://github.com/RocketChat/Rocket.Chat/pull/28096/files#diff-fcfe1f2ec2ec8a999533ccc48bd042fba7173b6783b8432d6e34d565d24de632) which is composed of the methods you shared on the comment plus the new one,

I added a deprecation warning on the method definition on [RoomReceiver](https://github.com/RocketChat/Rocket.Chat/pull/28096/files#diff-2c534580de1d3b6721d2e2937210b941740828d68da09ff6c27d6faa003c3e5b) so we can gradually move to export them, what do you think? if you want I can work on a Refactor but in subsequent PR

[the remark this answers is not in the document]
MarcosSpessatto: Awesome, thanks!

MarcosSpessatto (approved):

apps/meteor/ee/server/api/federation/rooms.ts
```diff
@@ -92,3 +93,20 @@ API.v1.addRoute(
 		},
 	},
 );
+
+API.v1.addRoute(
```
MarcosSpessatto, line 97: Is this an EE feature only? 🤔

apps/meteor/ee/server/api/federation/rooms.ts
```diff
@@ -92,3 +93,20 @@ API.v1.addRoute(
...
+	'federation/verifyeMatrixId',
```
MarcosSpessatto, line 98: I think there's a typo here 😬

apps/meteor/client/views/room/contextualBar/RoomMembers/AddUsers/ValidateMatrixInvitedUsers/ValidateMatrixIdModal.tsx
```diff
@@ -0,0 +1,64 @@
+import { Modal, Button, Box, Icon } from '@rocket.chat/fuselage';
+import { useTranslation } from '@rocket.chat/ui-contexts';
+import type { ReactElement } from 'react';
+import React from 'react';
+// import { Virtuoso } from 'react-virtuoso';
```
MarcosSpessatto, line 5: Can we remove this?

apps/meteor/ee/server/api/federation/rooms.ts
lmauromb: fixed! now supports both editions!

apps/meteor/server/services/federation/service.ts
```diff
@@ -317,4 +317,8 @@ export class FederationService extends AbstractBaseFederationService implements
 			FederationRoomSenderConverter.toCreateDirectMessageRoomDto(internalInviterId, internalRoomId, externalInviteeId),
 		);
 	}
+
+	public async verifyMatrixId(matrixId: string): Promise<string> {
```
MarcosSpessatto, line 321: Wouldn't be better to put this "common" method between CE and EE in the abstract class, so we can use the same for both?

MarcosSpessatto (changes requested):

MarcosSpessatto (approved):

apps/meteor/client/views/room/contextualBar/RoomMembers/AddUsers/AddUsersWithData.tsx
```diff
@@ -32,6 +35,16 @@ const AddUsersWithData = ({ rid, onClickBack, reload }: AddUsersWithDataProps):
 	const { users } = values as AddUsersInitialProps;
 	const { handleUsers } = handlers;
 
+	const matrixIdVerificationMap = new Map();
```
gabriellsh, line 38: This will create a new map every render. It needs to be saved in a state.

apps/meteor/client/views/room/contextualBar/RoomMembers/AddUsers/AddUsersWithData.tsx
```diff
@@ -32,6 +35,16 @@ const AddUsersWithData = ({ rid, onClickBack, reload }: AddUsersWithDataProps):
...
+	const dispatchValidationForMatrixId = useMatrixIdValidation({
+		matrixIdVerifiedStatus: matrixIdVerificationMap,
+		onClickBack,
+		rid,
+		reload,
+		users,
+	} as useMatrixIdValidationProps);
+	const dispatchVerifyEndpoint = useEndpoint('GET', '/v1/federation/verifyMatrixId', undefined);
```
gabriellsh, line 46: No need to pass undefined. If it's empty, it's already undefined.

apps/meteor/client/views/room/contextualBar/RoomMembers/AddUsers/AddUsersWithData.tsx
```diff
@@ -52,13 +65,44 @@ const AddUsersWithData = ({ rid, onClickBack, reload }: AddUsersWithDataProps):
...
+			const matrixIdsVerificationPromises = matrixIds.map((matrixId) =>
+				dispatchVerifyEndpoint({ matrixId } as FederationVerifyMatrixIdProps),
+			);
+
+			const matrixIdsVerificationPromiseResponse = await Promise.allSettled(matrixIdsVerificationPromises);
+			const matrixIdsVerificationFulfilledResults = matrixIdsVerificationPromiseResponse.filter(
+				(res) => res.status === 'fulfilled',
+			) as PromiseFulfilledResult<{ result: string }>[];
+
+			for (let i = 0; i < matrixIds.length; i++) {
```
gabriellsh, line 82: maybe use `.map(() => ...)` for this? Also, you're using a `filter` for `matrixIdsVerificationFulfilledResults` which can cause the array to be smaller than `matrixIds`. This by itself can cause 2 problems:
1 - `matrixIdsVerificationFulfilledResults[i];` can be undefined if `matrixIds` is bigger;
2 - the ID array position won't match with the response array.

apps/meteor/client/views/room/contextualBar/RoomMembers/AddUsers/ValidateMatrixInvitedUsers/ValidateMatrixIdModal.tsx
```diff
@@ -0,0 +1,63 @@
...
+type ValidateMatrixIdModalProps = {
+	_matrixIdVerifiedStatus: Map<string, string>;
+	onClose: () => void;
+	onConfirm?: () => void;
+};
+
+const ValidateMatrixIdModal = ({ onClose, onConfirm, _matrixIdVerifiedStatus }: ValidateMatrixIdModalProps): ReactElement => {
+	const t = useTranslation();
+
+	const verificationStatusAsIcon = (verificationStatus: string) => {
```
gabriellsh, line 15: You can move this function out of the Component scope, since it doesnt use any variables defined within it.

I recommend making this function return only the Icon `name` string property instead of returning the element.

apps/meteor/client/views/room/contextualBar/RoomMembers/AddUsers/AddUsersWithData.tsx
```diff
@@ -52,13 +65,44 @@ const AddUsersWithData = ({ rid, onClickBack, reload }: AddUsersWithDataProps):
 			dispatchToastMessage({ type: 'error', message: error as Error });
 		}
 	});
+
+	const handleMatrixValidation = useMutableCallback(async () => {
```
gabriellsh, line 69: I think it would make more sense to move this whole function to the `useMatrixIdValidation` hook.

apps/meteor/client/views/room/contextualBar/RoomMembers/AddUsers/ValidateMatrixInvitedUsers/useMatrixIdValidation.tsx
```diff
@@ -0,0 +1,48 @@
...
+export type useMatrixIdValidationProps = {
+	rid: IRoom['_id'];
+	matrixIdVerifiedStatus: Map<string, string>;
+	saveAction: () => void;
+	onClickBack: () => void;
+	reload: () => void;
+	users: string[];
+};
+
+export const useMatrixIdValidation = ({
```
gabriellsh, line 17: If the previous comments are respected, I suggest this gets renamed to `useAddMatrixUsers`

gabriellsh (changes requested):

MarcosSpessatto (dismissed):

sampaiodiego (dismissed):

apps/meteor/app/api/server/v1/federation.ts
```diff
@@ -0,0 +1,24 @@
+import { isFederationVerifyMatrixIdProps } from '@rocket.chat/rest-typings';
+import { Federation, FederationEE } from '@rocket.chat/core-services';
+
+import { isEnterprise } from '../../../../ee/app/license/server';
+import { API } from '../api';
+
+API.v1.addRoute(
+	'federation/verifyMatrixId',
```
KevLehman, line 8: What about `matrixId.verify` to make it similar to the others

apps/meteor/client/views/room/contextualBar/RoomMembers/AddUsers/ValidateMatrixInvitedUsers/ValidateMatrixIdModal.tsx
```diff
@@ -0,0 +1,63 @@
...
+import type { ComponentProps, ReactElement } from 'react';
+import React from 'react';
+
+type ValidateMatrixIdModalProps = {
+	matrixIdVerifiedStatus: Map<string, string>;
+	onClose: () => void;
+	onConfirm?: () => void;
+};
+
+const verificationStatusAsIcon = (verificationStatus: string) => {
```
KevLehman, line 12: Could verification status have other value or be an empty string? That case this would return undef

apps/meteor/client/views/room/contextualBar/RoomMembers/AddUsers/ValidateMatrixInvitedUsers/useAddMatrixUsers.tsx
```diff
@@ -0,0 +1,66 @@
...
+	users: string[];
+};
+
+export const useAddMatrixUsers = ({ rid, reload, onClickBack, handleUsers, users }: useAddMatrixUsersProps) => {
+	const t = useTranslation();
+	const setModal = useSetModal();
+	const saveAction = useMethod('addUsersToRoom');
+	const dispatchToastMessage = useToastMessageDispatch();
+	const handleClose = useMutableCallback(() => setModal(null));
+	const dispatchVerifyEndpoint = useEndpoint('GET', '/v1/federation/verifyMatrixId', undefined);
```
KevLehman, line 22: For gets, not sure if useQuery is the convention.

apps/meteor/client/views/room/contextualBar/RoomMembers/AddUsers/ValidateMatrixInvitedUsers/useAddMatrixUsers.tsx
```diff
@@ -0,0 +1,66 @@
...
+			const matrixIds = users.filter((user) => user.startsWith('@'));
+
+			const matrixIdsVerificationPromises = matrixIds.map((matrixId) => dispatchVerifyEndpoint({ matrixId }));
+
+			const matrixIdsVerificationPromiseResponse = await Promise.allSettled(matrixIdsVerificationPromises);
+			const matrixIdsVerificationFulfilledResults = matrixIdsVerificationPromiseResponse.filter(
+				(res) => res.status === 'fulfilled',
+			) as PromiseFulfilledResult<{ result: string }>[];
+
+			for (let i = 0; i < matrixIds.length; i++) {
```
KevLehman, line 49: Maybe a map is better here? You can get the current index of a map from the parameters of the callback

apps/meteor/server/services/federation/infrastructure/matrix/Bridge.ts
```diff
@@ -166,6 +169,29 @@ export class MatrixBridge implements IFederationBridge {
 		}
 	}
 
+	public async verifyInviteeId(externalInviteeId: string): Promise<VerificationStatus> {
+		const [userId, homeserverUrl] = extractUserIdAndHomeserverFromMatrixId(externalInviteeId);
+		try {
+			const response = await fetch(`https://${homeserverUrl}/_matrix/client/v3/register/available?username=${userId}`);
```
KevLehman, line 175: username can be passed as params instead of the URL

KevLehman: fetch(url, { params: { username } })

apps/meteor/server/services/federation/infrastructure/matrix/Bridge.ts
```diff
@@ -166,6 +169,29 @@ export class MatrixBridge implements IFederationBridge {
...
+
+			if (response.status === HttpStatusCodes.BAD_REQUEST) {
```
KevLehman, line 177: Check if response is ok before using status

KevLehman (changes requested):

apps/meteor/client/views/room/contextualBar/RoomMembers/AddUsers/ValidateMatrixInvitedUsers/ValidateMatrixIdModal.tsx
lmauromb: it shouldn't, since the value comes from this function

https://github.com/RocketChat/Rocket.Chat/blob/89c62679a8286ead3d433a5b16041a8f47139afb/apps/meteor/server/services/federation/infrastructure/matrix/Bridge.ts#L172-L193

which can only return those 3 values

apps/meteor/server/services/federation/infrastructure/matrix/Bridge.ts
lmauromb: If I understood, following node-fetch [example](https://github.com/node-fetch/node-fetch#handling-client-and-server-errors), using `response.ok` is the same as `response.status >= 200 && response.status < 300` which in this case will only make a redundant check since the only values we care are `200` and `400`.  

The way we're using the endpoint is inverted in a sense, 400 means VERIFIED for this method and 200 means UNVERIFIED, any other status or error will return UNABLE_TO_VERIFY

apps/meteor/client/views/room/contextualBar/RoomMembers/AddUsers/ValidateMatrixInvitedUsers/useAddMatrixUsers.tsx
lmauromb: that would not work, since the purpose of the loop is to fill the HashMap `matrixIdVerificationMap` and the `.map` will expect a return value after the call.

Basically I have two arrays which I wan to store into the HashMap, one contains the keys and the other the values; they share the index position and that is what I'm using to setup the HashMap

KevLehman: Then you can go with something like:

```
matrixIds.forEach((id, idx) => {
	const { value: { result } } = matrixIdsVerificationFulfilledResults[idx];
	matrixIdVerificationMap.set(id, result);
});
```

wouldn't that work? 👀

lmauromb: nice!

KevLehman (approved):

KevLehman (dismissed):

ggazzo (changes requested):
please change the endpoint to receive a list of users instead fetching one by one

ggazzo (approved):

PR #28107, opened 2023-02-20
Regression: Re-add app signature migration to fix the faulty previous one

d-gubert:

Previously merged migration files ended up in a faulty migration


sampaiodiego:
I might be missing something, but the code in both migration seems to be the same, so why would the new one not have the same result?

sampaiodiego (approved):

PR #28117, opened 2023-02-22
Regression: Fix omnichannel chat forwarding not working

murtaza98:
Requires https://github.com/RocketChat/fuselage/pull/988

## Issue(s)

https://rocketchat.atlassian.net/browse/OC-689


MartinSchoeler (changes requested):
This only fixes the symptoms, not the actual problem, we need to refactor the `useDepartmentList` hook instead, let me take a stab at it

MartinSchoeler (approved):

apps/meteor/client/components/AutoCompleteDepartmentMultiple.tsx
```diff
@@ -1,34 +1,45 @@
...
 
 	const { phase: departmentsPhase, items: departmentsItems, itemCount: departmentsTotal } = useRecordList(departmentsList);
 
 	const sortedByName = departmentsItems.sort((a, b) => {
-		if (a.name > b.name) {
+		if (a.label > b.label) {
 			return 1;
 		}
-		if (a.name < b.name) {
+		if (a.label < b.label) {
```
murtaza98, line 42: Can't we skip sorting the data on UI? I see that we're already passing the sort param to the API so you should have the data already sorted at this point

apps/meteor/client/components/AutoCompleteDepartment.tsx
```diff
@@ -51,29 +51,26 @@ const AutoCompleteDepartment = ({
 	const sortedByName = useMemo(
 		() =>
 			departmentsItems.sort((a, b) => {
-				const rankA = 'name' in a ? a.name : '';
-				const rankB = 'name' in b ? b.name : '';
+				const rankA = 'name' in a ? a.label : '';
+				const rankB = 'name' in b ? b.label : '';
```
murtaza98, line 55: The same goes here. I feel the data is already sorted by the backend based on the name field

apps/meteor/client/components/Omnichannel/hooks/useDepartmentsList.ts
```diff
@@ -75,28 +63,26 @@ export const useDepartmentsList = (
...
 				.map(({ _id, name, _updatedAt, ...department }): DepartmentListItem => {
 					return {
-						...department,
 						_id,
-						name: department.archived ? `${name} [${t('Archived')}]` : name,
 						label: department.archived ? `${name} [${t('Archived')}]` : name,
-						value: { value: _id, label: name },
-						...(_updatedAt && { _updatedAt: new Date(_updatedAt) }),
+						value: _id,
+						_updatedAt: new Date(_updatedAt || ''),
```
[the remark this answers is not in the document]
MartinSchoeler: That was a mistake that was happening, the updatedAt is a required property (that happened due the lack of TS)

apps/meteor/client/components/AutoCompleteDepartmentMultiple.tsx
MartinSchoeler: Indeed! That only shows how old is this hook, pushing changes

KevLehman (approved):

cauefcr (approved):
Tested the requested changes and everything works as expected, code also looks good, approving.

aleksandernsilva (approved):

PR #28132, opened 2023-02-23
Regression: "Enable unlimited apps" button on installation modal doesn't do anything

AllanPazRibeiro:
This PR fixes all buttons on the "Enable unlimited apps" modal, and the problems were:

- Close button: This button was just a 'placebo', now it's calling the `setModal(null)` properly inside a useCallback, which I noticed is a good practice and used on most of our modals.
- Contact sellers: This one was concatenating with the server's current path, so it was pointing to a non-existent URL like: http://localhost:3000/app/admin/go.rocket.chat /contact-sellers. In order to fix it I change it to an HTTPS URL having the target `_blank`.
- Start trial: Last but not least, this one was having a very similar problem to the previous one, but the redirect was faster than some requests so the application was not been able to check the permissions to show the page and at the same time the modal was not been closed. In order to fix it, I replace the Flowrouter to a more native approach (window), so now the page is refreshing and getting all the permissions needed.

## Issue(s)
MKP-246

## Further comments

MKP-246
MKP-255
MKP-256
MKP-257

[MKP-246]: https://rocketchat.atlassian.net/browse/MKP-246?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/meteor/client/views/marketplace/UnlimitedAppsUpsellModal.tsx
```diff
@@ -1,17 +1,30 @@
...
+	onClose: () => void;
+};
+
+const UnlimitedAppsUpsellModal = ({ onClose }: UnlimitedAppsUpsellModalProps) => {
 	const t = useTranslation();
 	const cloudWorkspaceHadTrial = useSetting('Cloud_Workspace_Had_Trial') as boolean;
 	const urls = {
 		goFullyFeaturedRegistered: 'admin/upgrade/go-fully-featured-registered',
-		talkToSales: 'go.rocket.chat/i/contact-sales',
+		talkToSales: 'https://www.rocket.chat/sales-contact',
```
d-gubert, line 15: We need to keep the `go.rocket.chat` URL here, it helps us visualize access
```suggestion
		talkToSales: 'https://go.rocket.chat/i/contact-sales',
```

apps/meteor/client/views/marketplace/UnlimitedAppsUpsellModal.tsx
```diff
@@ -1,17 +1,30 @@
...
 	};
 
+	const goFullyFeaturedRegistered = useCallback(() => {
+		window.open(urls.goFullyFeaturedRegistered, '_self');
```
d-gubert, line 19: We need to use the internal routing system for that instead, try following the example in `apps/meteor/client/views/invite/SecretURLPage.tsx`

d-gubert (approved):

d-gubert (changes requested):
Approved by mistake 😬 Previous comments must be solved

rique223 (dismissed):

d-gubert (approved):

rique223 (approved):

PR #28145, opened 2023-02-24
Regression: New messages not reflected in the chat window for queued chats.

murtaza98:

Close room coordinator when chat is returned to the queue, also removes chat from the queue collection when closing chat.


apps/meteor/app/livechat/server/hooks/saveLastMessageToInquiry.ts
```diff
@@ -8,12 +8,7 @@ import { RoutingManager } from '../lib/RoutingManager';
 callbacks.add(
 	'afterSaveMessage',
 	(message, room) => {
-		if (!isOmnichannelRoom(room)) {
-			return message;
-		}
-
-		// skip callback if message was edited
-		if (isEditedMessage(message)) {
+		if (!isOmnichannelRoom(room) || isEditedMessage(message) || message.t) {
```
KevLehman, line 11: Can you explain me why this was needed? 👀 I mean, to combine the ifs, as the only new condition i see is "message.t"

murtaza98: Previously, the transfer system message was getting stored within the lastMessage prop which is y it was showing the message text as `undefined`. This new condition is added so it can filter out messages like that which sidebar is not capable of rendering

KevLehman (dismissed):

KevLehman:
Hey @murtaza98 , I feel this fixes another regression, but not the one in 601. That one seems to be related to how the UI is subscribed to the messages and probably not happening after the move to queue and taken back. Can you take a look with this scenario?

aleksandernsilva (approved):

PR #28162, opened 2023-02-24
Regression: Support generating pdfs from queue

KevLehman:

This removes a code validation that prevented PDF files from being generated for rooms that were closed directly from the queue.

After discussing with product, this is a behavior we want and so we're removing the code preventing it and updating the translation keys.

## Further comments

OC-696


cauefcr (approved):
tested the changes requested, and it works as expected, approving.

sampaiodiego (approved):
great to see there was a test case for this scenario that had to be updated 🤗

filipemarins (approved):

PR #28163, opened 2023-02-24
Regression: Jump to message not working

gabriellsh:
## Steps to test or reproduce

- Scroll up and get a message link (kebab menu) from a message that is hidden before loading more messages.
- Paste it in the url bar

Sometimes there is a race condition and the room is opened but you don't go to the desired message. When scrolling manually to a point where the message is loaded, it then scrolls and highlights the message as it should.
## Further comments

TC-486


apps/meteor/client/views/room/components/body/RoomBody.tsx
```diff
@@ -203,6 +205,8 @@ const RoomBody = (): ReactElement => {
 
 	const retentionPolicy = useRetentionPolicy(room);
 
+	useJumpToMessage(msgParameter);
```
hugocostadev, line 208: What if the message it not loaded and it's not on the view in the first load of the room?

hugocostadev (changes requested):

tassoevan (approved):

PR #28165, opened 2023-02-24
Regression: Remove unnecessary messages from Email transcript

KevLehman:
## Further comments

OC-693


sampaiodiego (commented):
I mean, they *can* be relevante to the conversation..

KevLehman:
Well, until recently rocket.cat was banned from posting on omnichannel rooms... and now his only available message is "Here's your PDF" so... 
image

murtaza98:
I do agree with Diego - excluding messages from `rocket.cat` seems a bit extreme if you just want to hide the transcript message. In some cases, workspaces use chatbots to interact with visitors and direct them to the right agent. Often, the chatbot's agent is `rocket.cat`. If this logic is applied, then all the messages from the chatbot will be left out of the PDF generated.

A possible solution that I was thinking of yesterday was to add a custom field to those PDF transcript messages sent within the room. Then we can filter out those messages anywhere we don't need it, like over here while sending this email. WDYT?

KevLehman:
I'm not a fan of using a custom field for filtering this tbh. But I do wanna know, why d why those apps message are using rocket.cat instead of an app's user? 👀 

The option I was looking for is to avoid getting messages after the closing message, which will work and it's a more straightforward approach than putting another field onto messages (which will have to have and index and all of the stuff 😬 ) wdyt?

So it will be something like:
- Get the closing message ts
- Filter all messages sent after closing

murtaza98:
> * Get the closing message ts
> * Filter all messages sent after closing

This looks awesome!  :rocket:  I'm totally on-board going this route




> But I do wanna know, why d why those apps message are using rocket.cat instead of an app's user?

Those app messages can be from any bot user - it's configurable via app's setting. What I meant earlier was, people might re-use this `rocket.cat` bot in this situation since its already a bot user which is present on all servers. So in this situation, the earlier filter would cause issues.

cauefcr (dismissed):
locally tested the changes requested, everything works as expected, no more transcript generated message on the e-mail, code also looks good.

cauefcr (approved):
Re-tested the functionality, everything is still working after the typing, code also looks better now, with a typescript function migration.

apps/meteor/server/models/raw/Messages.ts
```diff
@@ -243,6 +253,70 @@ export class MessagesRaw extends BaseRaw<IMessage> implements IMessagesModel {
...
+		};
+
+		if (types.length > 0) {
+			query.t = { $nin: types };
+		}
+
+		return this.find(query, options);
+	}
+
+	findVisibleByRoomIdNotContainingTypesAndUsers(
```
murtaza98, line 288: Do we still need this method?

KevLehman: Not for this PR, but I have other place where i need this same thing so i'll leave it for now :awesome:

apps/meteor/app/livechat/server/lib/LivechatTyped.ts
```diff
@@ -177,6 +175,128 @@ class LivechatClass {
...
+			throw new Error('error-invalid-room');
+		}
+
+		// allow to only user to send transcripts from their own chats
+		if (room.t !== 'l' || !room.v || room.v.token !== token) {
+			throw new Error('error-invalid-room');
+		}
+
+		const showAgentInfo = settings.get<string>('Livechat_show_agent_info');
+		const closingMessage = await Messages.findLivechatClosingMessage(rid, { projection: { ts: 1 } });
```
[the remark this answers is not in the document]
KevLehman: Thanks to god no :kek:

sampaiodiego (approved):

PR #28169, opened 2023-02-24
[FIX] Notify apps engine event through ms api

thassiov:

The Apps-Engine must drop the current Streamer utilization and start using the Service APIs to correctly communicate between instances and client-server.

## Further comments

Problem statement

The Apps-Engine framework has always relied on Streamers to handle communication of real time events between instances in a cluster (High Availability) and server-client. 

However, the current Microservices architecture doesn’t allow for the same Streamer lib to be used - this renders all event communication in the Apps-Engine useless when the workspace is deployed in Microservices mode. 

This problem causes inconsistent states between instances of Rocket.Chat that are scaled horizontally in a microservice deployment and also cuts real time communication of app changes to clients. The only workaround in this situation is to restart all instances after a change in apps happened so that every one has these changes applied.

On the user perspective, though, apps become essentially useless in an MS deployment.

AE-106


apps/meteor/server/modules/listeners/listeners.module.ts
```diff
@@ -1,11 +1,13 @@
...
-import type { IUser, IRoom, VideoConference } from '@rocket.chat/core-typings';
+import type { AppStatus } from '@rocket.chat/apps-engine/definition/AppStatus';
+import type { IUser, IRoom, VideoConference, ISetting } from '@rocket.chat/core-typings';
 import { parse } from '@rocket.chat/message-parser';
 import type { IServiceClass } from '@rocket.chat/core-services';
 import { EnterpriseSettings } from '@rocket.chat/core-services';
 
 import type { NotificationsModule } from '../notifications/notifications.module';
 import { settings } from '../../../app/settings/server/cached';
+import { AppEvents } from '../../../ee/server/apps/communication';
```
ggazzo, line 10:
```suggestion
import { AppEvents } from '../../../ee/server/apps/communication/websockets';
```

apps/meteor/server/services/apps-engine/service.ts
```diff
@@ -1,18 +1,95 @@
 import { ServiceClassInternal } from '@rocket.chat/core-services';
 import type { IAppsEngineService } from '@rocket.chat/core-services';
+import type { AppStatus } from '@rocket.chat/apps-engine/definition/AppStatus';
+import { AppStatusUtils } from '@rocket.chat/apps-engine/definition/AppStatus';
+import type { ISetting } from '@rocket.chat/core-typings';
 
 import { Apps, AppEvents } from '../../../ee/server/apps/orchestrator';
+import { AppEvents as AppLifeCycleEvents } from '../../../ee/server/apps/communication';
```
ggazzo, line 8:
```suggestion
import { AppEvents as AppLifeCycleEvents } from '../../../ee/server/apps/communication/websockets';
```

ggazzo:
I give up, there's something wrong with the code and I'm not going to solve it myself, but your previous typescript problem is as a git to you ;)

probably a throwing an error if is empty should be enough but I dont have sufficient skills to this by myself 
image
image

ggazzo:
btw stay away from `enums`

AllanPazRibeiro (approved):
LGTM

PR #28180, opened 2023-02-27
Regression: broken message exhibition when edited

tassoevan:

Sometimes a message will have the fields `editedBy` and `editedAt` present but set to `null`. This PR handle this situation.


apps/meteor/app/federation-v2/server/infrastructure/rocket-chat/hooks/index.ts
```diff
@@ -160,7 +160,7 @@ export class FederationHooks {
 				) {
 					return message;
 				}
-				if (!isEditedMessage(message)) {
+				if (!isEditedMessage(message) || !message.editedBy) {
```
sampaiodiego, line 163: any reason to not add the check `!message.editedBy` into `isEditedMessage` function?

sampaiodiego (changes requested):

apps/meteor/app/federation-v2/server/infrastructure/rocket-chat/hooks/index.ts
tassoevan: Yes, three failing Federation unit tests when the type guard is more stricter.

ggazzo (dismissed):

sampaiodiego (approved):

ggazzo (approved):

PR #28191, opened 2023-02-27
feat: Replace "Hide Usernames" by the new "Messages Layout" preference

matheusbsilva137:

 - Replace "Hide Usernames" preference and "Use Real Name" setting by the **new "Messages Layout" preference** (with the "Show only username", "Show only full name" and "Show both full names and username" options);

## Steps to test or reproduce

Login and go to **My Account > Preferences > Messages** and choose one of the three options provided. Messages lists should be updated accordingly (threads' and users' lists should also honor the preference's value).
The default value for this setting can be chosen in **Workspace > Settings > Accounts > Default User Preferences > Default Messages Layout**.

## Further comments

The new setting's inital value will be assigned according to the "Use Real Name" setting' value AND the "Hide Usernames" preference's default value:

| Use Real Name | Default "Hide Usernames" | Default Messages Layout |
|---|---|---|
| true | true | Show only full name |
| true | false | Show both full name and username |
| false | false | Show only username |

TC-157


apps/meteor/app/api/server/v1/misc.ts
```diff
@@ -257,7 +258,7 @@ API.v1.addRoute(
 					const user = this.getUserFromParams();
 
 					// Respect the server's choice for using their real names or not
-					if (user.name && settings.get('UI_Use_Real_Name')) {
+					if (user.name && getMessagesLayoutPreference() !== 'username') {
```
LucianoPierdona, line 261: Can this check be moved into a helper function? I see that it's being checked more than 15 times the same thing, and I believe it would be better to move this into a separate function to reduce future maintenance.

`export const useRealName = (uid?: string) => getMessagesLayoutPreference(uid) !== 'username'`

apps/meteor/client/hooks/useUserDisplayName.ts
```diff
@@ -1,10 +1,13 @@
...
+import { useUserPreference, useSetting } from '@rocket.chat/ui-contexts';
 
 import { getUserDisplayName } from '../lib/getUserDisplayName';
 
 export const useUserDisplayName = ({ name, username }: Pick<IUser, 'name' | 'username'>): string | undefined => {
-	const useRealName = useSetting('UI_Use_Real_Name');
+	const messagesLayoutPreference = useUserPreference('messagesLayout');
+	const defaultMessagesLayout = useSetting('Accounts_Default_User_Preferences_messagesLayout');
+	const useRealName =
+		messagesLayoutPreference !== 'default' ? messagesLayoutPreference !== 'username' : defaultMessagesLayout !== 'username';
```
LucianoPierdona, line 10: This one is also interesting to move into a helper function, it's the same code being used 9 times along the app.

_note: Might need to think on a different name for this function, or rename the other I suggested_
```
export const useRealName = () => {
	const messagesLayoutPreference = useUserPreference('messagesLayout');
	const defaultMessagesLayout = useSetting('Accounts_Default_User_Preferences_messagesLayout');
	return messagesLayoutPreference !== 'default' ? messagesLayoutPreference !== 'username' : defaultMessagesLayout !== 'username';
}
```

LucianoPierdona (changes requested):

ggazzo (changes requested):
the preference already takes in consideration the default value from the setting so there is no need to test the default value

apps/meteor/client/hooks/useUserDisplayName.ts
matheusbsilva137: Fixed! I had to change the way `useUserPreference` works so that it never returns the string `default`.

apps/meteor/app/api/server/v1/misc.ts
matheusbsilva137: Fixed!

LucianoPierdona (approved):

gabriellsh (dismissed):

gabriellsh:
We'll hold on this one since it's changing specification quite often. Once all is set for good, we can move on.

gabriellsh:
Closed as per linked task cancellation status.

PR #28198, opened 2023-02-27
Regression: Fix app status filter not working in private apps page

rique223:
## Jira task: [MKP-266](https://rocketchat.atlassian.net/browse/MKP-266?atlOrigin=eyJpIjoiMzlhNmRhZDNmOGZkNDBmZGEyY2FkNTQ5ZTVlZjUzNzkiLCJwIjoiaiJ9)
Fixed a bug in which the app status filter didn't work because of the new 'enabled*' and 'disabled*' app status.
Demo gif:
private-apps-status-filter

## Further comments

MKP-266

[MKP-266]: https://rocketchat.atlassian.net/browse/MKP-266?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


PedroRorato (approved):

apps/meteor/client/views/marketplace/helpers/filterAppsByEnabled.ts
```diff
@@ -1,5 +1,10 @@
 import { appStatusSpanProps } from '../helpers';
 import type { App } from '../types';
 
-export const filterAppsByEnabled = (app: App): boolean =>
-	appStatusSpanProps(app)?.label === 'Enabled' || appStatusSpanProps(app)?.label === 'Trial period';
+export const filterAppsByEnabled = (app: App): boolean => {
+	const appStatus = appStatusSpanProps(app)?.label;
```
ggazzo, line 5:
```suggestion
	const appStatus = appStatusSpanProps(app)?.label ?? '';
```

ggazzo (changes requested):
please check the suggestions

apps/meteor/client/views/marketplace/helpers/filterAppsByEnabled.ts
rique223: Wouldnt this be the same as using the || operator down below but with more code?

d-gubert: I'm with @rique223 on this one, I don't think adding the if statement there would improve the clarity of the code here

ggazzo: Could you enlighten me on how clearly is testing if the array contains '" is it relevant for the whole code? you are testing something that will return false

A rule of thumb always return `early` if code clarity matters to you

ggazzo (approved):
I'll pretend I didn't see that you just ignored my suggestion and asked my review again expecting a different opinion

apps/meteor/client/views/marketplace/helpers/filterAppsByEnabled.ts
rique223: The way I see this code the "" is a fallback state that will never happen naturally but instead only if a problem happens to the app status, be it being undefined or simply not existing, so I do an || operation that will guarantee that this filter uses the empty string as input and consequently returns false in cases that fall out of expected states of the app status. The if does the same thing with extra symbols and the difference in clarity seemed negligible in contrast to the loss of cleanliness of the code. But my vision about all of this may be wrong and I wanted to understand if there were real advantages to one approach over the other and that's why I asked that question. Also, I didn't ignore your suggestion, I mistakenly assumed that this discussion was solved due to the time that it took to receive an answer on the question above.

Issue #28115, opened 2023-02-22
sound status goes out of window on popout mode


GauravBurande:
### Description:

When hovering over the bell icon in rocket.chat's popout mode the sound status goes out the window area

### Steps to reproduce:

1. Access: https://open.rocket.chat/livechat?mode=popout
2. Click on rocket.chat button at the bottom right
3. Click on expand chat
4. hover over bell icon

### Expected behavior:

the sound status should be visible in the window area

### Actual behavior:

the sound status which shows on hovering over bell icon goes out of the tab window

![rocket chat bot](https://user-images.githubusercontent.com/100035026/220558426-79094fdb-53a1-46d6-b9ff-96d418d0c363.png)

### Server Setup Information:

- Version of Rocket.Chat Server: 
- Operating System: 
- Deployment Method: 
- Number of Running Instances: 
- DB Replicaset Oplog: 
- NodeJS Version: 
- MongoDB Version:

### Client Setup Information

- Desktop App or Browser Version: browser
- Operating System: windows

dudanogueira:
Thanks!! I was able to reproduce this.

XshubhamX:
@dudanogueira Can I work on this?

**Possible Solution**
<img width="909" alt="image" src="https://user-images.githubusercontent.com/63532716/220760073-17635b39-40ee-4d94-acea-231c59c81ab5.png">
This is happening because white spaces are overflowing, a simple  and reliable fix would be to just add below styling to the div.
```
div {
  white-space: normal;
}
```

JayashTripathy:
@GauravBurande hey can you please assign me this issue ?
 
what i can fix is -
1.  I think its better if we move that tooltip div(which displays sound status) inside of its parent immediate button(classname ="header__action__Xiqxc") and style according to that button .
![image](https://user-images.githubusercontent.com/76092296/220897484-72fc3507-a0de-428a-905a-a61e8e5d533d.png)

2. the over flowing tooltip div(classname="tooltip__1MCkK ") can be reposition from bottom to left side and it'll work without overflowing

dudanogueira:
Sure thing!

I was not aware of that popout mode, and looks like it's not documented (or I was not able to find it, anyway)

priyanshu-kun:
should padding-left and padding-right in .header__Ma8Sh class get the jobs done?

Nilmohon:
Please assign me to work with this issue

GauravBurande:
[...]

PR #28205, opened 2023-02-28
fix: Fixes bell icon hover status

AdityaSingh-02:

I just made few changes in styles and it fixes the issue #28115 
image

## Issue(s)

This Pull request closes issue #28115  , the issue was while we were hovering over the notification icon the content was going outside the window
## Steps to test or reproduce
Goto http://localhost:3000/livechat?mode=popout
Then on top right corner look for bell icon and hover over it then you will see the content is inside the window and not going outside the window


ggazzo (changes requested):
rocket.chat has several languages modifying the percentage will not work

KevLehman:
Gazzo's right on this one. Can you try a more safe approach, that can work with multiple languages?
Your Language setting (workspace) defines which one do you use, try with Arabic or other languages with long text 🤗

AdityaSingh-02:
Sure I'll try to fix it.✌️

AdityaSingh-02:
hey @ggazzo i came across an approach where i added margin and it solves the issue and i have tested it on multiple languages.. if you like it i can make an pull request over it

margin

Issue #26980, opened 2022-09-29
Cannot modify custom emoji without re-uploading it


paulchen:
### Description:

When editing a custom emoji without uploading a new file, the error message "No file uploaded" appears when clicking the "Save" button and the changes will not be saved.

### Steps to reproduce:

1. Add a custom emoji in case you don't already have one.
2. Select a custom emoji for editing.
3. Edit its name or aliases.
4. Click the "Save" button.

### Expected behavior:

The changes are saved without an error.

### Actual behavior:

An error message is shown and the changes will not be saved:

![image](https://user-images.githubusercontent.com/404840/193107547-b48f6baf-4918-42ba-83ab-6fdb65e1402a.png)

### Server Setup Information:

- Version of Rocket.Chat Server: 5.1.4
- Operating System: Debian GNU/Linux
- Deployment Method: Docker
- Number of Running Instances: 1
- DB Replicaset Oplog: ?
- NodeJS Version: 14.19.3
- MongoDB Version: 5.0.12

### Client Setup Information

- Desktop App or Browser Version: Chromium 106.0.5249.61 (Offizieller Build) Arch Linux (64-Bit), Firefox 105.0.1 (64-Bit)
- Operating System: Arch Linux

florisvangeel:
https://user-images.githubusercontent.com/1267925/193622341-fb8e8183-6d51-4816-bf02-a80c37452b69.mp4

Confirmed this does not work in 5.1.4

kartikbh6614:
firstly create an emojis you want ADD then edit to them with help of typescript then get to server and add to them emojis or single symbol list

codewithmitesh:
Is anyone solving this issue? If not, @dudanogueira  would you kindly assign this to me?

Cob0ld:
can't tell if its related but we can't update our custom emojis ... it will let you save the new emoji but always false back to the older image. this is happens even if i delete the emoji and recreate it later.

dudanogueira:
there is indeed something going off here. Let's try to make it into 6.1.0!

Gardezi1:
Hey guys can i pick up this issue. Seems like nobody has started working on this issue

XpertWebApp:
this issue is fixed now , please check

https://user-images.githubusercontent.com/68045060/221577178-b52c5ce2-8ecc-4c28-9e0a-08633be8c388.mp4

paulchen:
> this issue is fixed now , please check

[...]

PR #28220, opened 2023-03-01
Custom emoji re-uploading file error no longer appears

XpertWebApp:
[FIX] Modify custom emoji without re-uploading it #26980

## Modify custom emoji

The file upload issue is now resolved issue file is not found by default when page is first rendered and only file preview is received. So, file upload is showing error but now working properly. 
please check further comments
Thank you

https://user-images.githubusercontent.com/68045060/222134438-711dfc35-3910-4f24-88eb-2046472a9f6d.mp4

## Issue(s)

#26980

## Steps to test or reproduce
http://localhost:3000/admin/emoji-custom

1. Add a new custom emoji
2. click on the emoji name 
3. change aliases name 
4. Click on the save button, now the error is gone and the data updated

## Further comments
This problem is solved but one problem which I have observed is when I update the data. The file and file name of the file has not been updated from the database. They create problem that while naming the emoji file doesn't show. And when we change the emoji file then next time edit it does not show the updated emoji file


hugocostadev:
Please, remove the file `.eslintcache`

PR #28234, opened 2023-03-01
[BREAK] Remove support to deprecated typing event

sampaiodiego:


ggazzo (approved):

geekgonecrazy:
Maybe in release notes we should be clear that the other one that replaced it a while back {rid}/user-activity with type user-typing should be used

PR #28239, opened 2023-03-02
[FIX] Escape HTML chars to avoid hyperlink injection

tapiarafael:
Escape HTML chars from the name field to avoid hyperlink injection in email.

## Further comments

[PVUL-53](https://rocketchat.atlassian.net/browse/PVUL-53?atlOrigin=eyJpIjoiODVlYmZlMGZkMTA4NGI0NzlhNGE1YTM4NjI2OThmZWYiLCJwIjoiaiJ9)

[PVUL-53]: https://rocketchat.atlassian.net/browse/PVUL-53?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/meteor/server/methods/registerUser.js
```diff
@@ -94,7 +95,7 @@ Meteor.methods({
 			throw new Meteor.Error(e.message);
 		}
 
-		Users.setName(userId, trim(formData.name));
+		Users.setName(userId, escapeHtml(trim(formData.name)));
```
ggazzo, line 98: so the idea here is saving a different name that I actually set? save an original name with a escaped value?
it doesn’t look good to me, the places where we render the name should handle that, its not the easiest way but the right one to do

maybe if its the case we should not allow such characters 

Anyway you are only handling the user registration but not later if the user changes the name using the api

ggazzo (changes requested):

PR #28240, opened 2023-03-02
fix: Events listeners were not being removed when services (local) were destroyed

MarcosSpessatto:

Whenever you register an event listener using the `onEvent` method, they are not removed, even calling the `destroyService()` method.
This happens on services registered with `LocalBroker` only, services relying on `NetworkBroker` were working fine since the removal of these events is handled by Moleculer.

Also, I've added the `stopped` lifecycle method call within `LocalBroker` when the service is a local one.

Jira: FED-89


packages/core-services/src/types/ServiceClass.ts
```diff
@@ -61,8 +62,15 @@ export abstract class ServiceClass implements IServiceClass {
...
-		return this.events.eventNames() as unknown as Array<keyof EventSignatures>;
+	getEvents(): { eventName: keyof EventSignatures; listeners: { (...args: any[]): void }[] }[] {
+		return this.events.eventNames().map((eventName) => ({
+			eventName: eventName as unknown as keyof EventSignatures,
+			listeners: this.events.rawListeners(eventName) as { (...args: any[]): void }[],
+		}));
+	}
+
+	removeAllListeners(): void {
+		this.events.removeAllListeners();
```
MarcosSpessatto, line 73: @sampaiodiego I've added this method to remove all the listeners at once when the instance is destroyed, do you believe this would be unnecessary maybe?

pierre-lehnen-rc (approved):

ggazzo (approved):

sampaiodiego (approved):

ggazzo (approved):

PR #28246, opened 2023-03-02
Regression: Adjusted departments upsell modal to check for licence and modules

aleksandernsilva:

This PR fixes a regression on the departments upsell modal, for the rare cases where the workspace has an enterprise licence but has no active modules. This changes improves the check to also consider modules.

## Further comments

OC-712


tiagoevanp (approved):

KevLehman (approved):
Can you add context on why this is needed for 6.0?

PR #28252, opened 2023-03-03
refactor: omnichannel on-hold feature

murtaza98:
## Issue(s)

Also fixes: Resume on-hold chat not working with max-chat's allowed per agent config (OC-529)
## Further comments

Note: this only fixes the issue in https://rocketchat.atlassian.net/browse/OC-529 which was caused by Subscription collection not having `onHold` prop. I also some further refactoring of the on-hold feature which can be found here -> https://github.com/RocketChat/Rocket.Chat/pull/28306


aleksandernsilva (dismissed):

apps/meteor/client/views/room/components/body/composer/ComposerOmnichannel/hooks/useResumeChatOnHoldMutation.ts
```diff
@@ -8,11 +8,18 @@ export const useResumeChatOnHoldMutation = (
...
+
 	const queryClient = useQueryClient();
 
 	return useMutation(
 		async (rid) => {
-			await resumeChatOnHold(rid, { clientAction: true });
+			try {
+				await resumeChatOnHold(rid, { clientAction: true });
+			} catch (error) {
+				console.error(error);
```
murtaza98, line 20:
```suggestion
```
oops

apps/meteor/ee/app/livechat-enterprise/server/hooks/checkAgentBeforeTakeInquiry.ts
```diff
@@ -66,6 +66,7 @@ const validateMaxChats = async ({
 	}
 
 	const { queueInfo: { chats = 0 } = {} } = user;
+	console.log('chats', chats, maxNumberSimultaneousChat);
```
murtaza98, line 69:
```suggestion
```

apps/meteor/client/views/room/components/body/composer/ComposerOmnichannel/hooks/useResumeChatOnHoldMutation.ts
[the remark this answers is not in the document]
murtaza98: Ouch! forgot to removes these logs, glad you caught it :hugs:

cauefcr (dismissed):
Simple and good changes, approving.

sampaiodiego (dismissed):

apps/meteor/ee/app/livechat-enterprise/server/lib/AutoTransferChatScheduler.ts
```diff
@@ -76,35 +77,39 @@ class AutoTransferChatSchedulerClass {
...
-		if (agent) {
-			return forwardRoomToAgent(room, {
-				userId: agent.agentId,
-				transferredBy: await this.getSchedulerUser(),
-				transferredTo: agent,
-				scope: 'autoTransferUnansweredChatsToAgent',
-				comment: timeoutDuration,
-			});
+		if (!agent) {
+			logger.error(`No agent found to transfer room ${room._id} which hasn't been answered in ${timeoutDuration} seconds`);
```
KevLehman, line 90: Use a translatable message (not sure if that word exists lol)

KevLehman (changes requested):

apps/meteor/ee/app/livechat-enterprise/server/lib/AutoTransferChatScheduler.ts
murtaza98: That's an error log for an async job, so it doesn't need translation (atleast for now until we support translation at log level!)

apps/meteor/ee/app/livechat-enterprise/server/api/rooms.ts
```diff
@@ -1,56 +1,107 @@
...
+				throw new Error('error-invalid-room');
+			}
+
+			if (!room.onHold) {
+				throw new Error('error-room-not-on-hold');
+			}
+
+			if (!room.open) {
+				throw new Error('This_conversation_is_already_closed');
+			}
```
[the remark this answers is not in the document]
murtaza98: done

apps/meteor/ee/app/livechat-enterprise/server/lib/logger.ts
```diff
@@ -7,3 +7,4 @@ export const queueLogger = logger.section('Queue');
 export const helperLogger = logger.section('Helper');
 export const cbLogger = logger.section('Callbacks');
 export const bhLogger = logger.section('Business-Hours');
+export const ohLogger = logger.section('On-Hold');
```
KevLehman, line 10: :kek:

KevLehman (dismissed):

apps/meteor/app/livechat/server/lib/RoutingManager.js
```diff
@@ -106,7 +106,7 @@ export const RoutingManager = {
 		const user = await Users.findOneById(agent.agentId);
 		const room = await LivechatRooms.findOneById(rid);
 
-		await Message.saveSystemMessage('command', rid, 'connected', user);
+		await Promise.all([Message.saveSystemMessage('command', rid, 'connected', user), Message.saveSystemMessage('uj', rid, '', user)]);
```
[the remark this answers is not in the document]
murtaza98: Sorry I didn't get what you were referring to here Kevin. Are you referring to the user join system message? If so then we had to move it here from `forwardRoomToAgent` method because of the issue mentioned by Harmeet here -> https://rocketchat.atlassian.net/browse/OC-529?focusedCommentId=15118

apps/meteor/app/livechat/server/lib/RoutingManager.js
```diff
@@ -176,7 +176,7 @@ export const RoutingManager = {
 			return room;
 		}
 
-		if (room.servedBy && room.servedBy._id === agent.agentId && !room.onHold) {
+		if (room.servedBy && room.servedBy._id === agent.agentId) {
```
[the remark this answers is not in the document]
murtaza98: Yes, I've tried to simplyfy the logic here and not call takeInquiry when the chat is resumed from on-hold. Rather after the chat is resumed from on-hold, we check if the previously serving agent is available. If they are, then we simply resume the chat and move it in-progress. Previosly we used to place the chat in the routing queue and then the routing queue used to assign it to the previously serving agent. But this step seemed a bit unnecessary so I've removed it. Note: We're still respecting agent's availablity with this new approach and if the agent isn't available then we place the chat in the routing queue

apps/meteor/ee/app/livechat-enterprise/server/hooks/resumeOnHold.ts
```diff
@@ -1,46 +1,73 @@
...
 
-		// Need to read the room every time, the room object is not updated
-		const room = await LivechatRooms.findOneById(rid, { projection: { t: 1, v: 1, onHold: 1 } });
-		if (!room) {
+	if (message.token && room.onHold) {
+		callbackLogger.debug(`[afterSaveMessage] Room ${rid} is on hold, resuming it now since visitor sent a message`);
+
+		try {
+			const { comment: resumeChatComment, resumedBy } = await resumeOnHoldCommentAndUser(updatedRoom);
+			await OmnichannelEEService.resumeRoomOnHold(updatedRoom, resumeChatComment, resumedBy);
```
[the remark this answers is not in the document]
murtaza98: I don't think that would be a good idea and we're not doing this right now. Here on line 57, we're only resolving the comment and commentor. We're not sending that message. Rather this info is being passed to the resumeOnHold service which would use this info to send the system message once the chat gets resumed.

Sidenote: We're passing this value as a param since chat can be resumed in 2 ways:
1) First by a visitor when they send a message
2) Or by an agent when they click on "Resume" button on the UI

apps/meteor/ee/app/livechat-enterprise/server/lib/AutoTransferChatScheduler.ts
```diff
@@ -76,35 +77,40 @@ class AutoTransferChatSchedulerClass {
...
 		const { roomId } = data;
 
-		if (await this.transferRoom(roomId)) {
+		try {
+			await this.transferRoom(roomId);
 			await LivechatRooms.setAutoTransferredAtById(roomId);
+		} catch (error) {
+			logger.error(`Error while executing job ${SCHEDULER_NAME} for room ${roomId}:`, error);
+		} finally {
+			await this.unscheduleRoom(roomId);
```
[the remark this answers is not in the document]
murtaza98: Makes sense. Let me move it there

apps/meteor/ee/app/livechat-enterprise/server/lib/AutoTransferChatScheduler.ts
```diff
@@ -76,35 +77,40 @@ class AutoTransferChatSchedulerClass {
...
+			logger.error(`No agent found to transfer room ${room._id} which hasn't been answered in ${timeoutDuration} seconds`);
+			return;
 		}
 
-		return false;
+		await forwardRoomToAgent(room, {
+			userId: agent.agentId,
+			transferredBy: await this.getSchedulerUser(),
+			transferredTo: agent,
+			scope: 'autoTransferUnansweredChatsToAgent',
```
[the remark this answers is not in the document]
murtaza98: That would be for the system messsages which we have defined here -> https://github.com/RocketChat/Rocket.Chat/blob/OC-529/packages/livechat/src/components/helpers.js#L196

apps/meteor/ee/app/livechat-enterprise/server/lib/logger.ts
murtaza98: Lol! I'd defined it before so as to use it within this new service. However ended up defining a new logger over there and not using it. Glad you caught it. Thanks!

aleksandernsilva (dismissed):

KevLehman (dismissed):

aleksandernsilva (approved):

apps/meteor/ee/app/livechat-enterprise/server/api/rooms.ts
```diff
@@ -1,56 +1,80 @@
...
 	'livechat/room.onHold',
-	{ authRequired: true, permissionsRequired: ['on-hold-livechat-room'] },
+	{ authRequired: true, permissionsRequired: ['on-hold-livechat-room'], validateParams: isLivechatRoomOnHoldProps },
 	{
 		async post() {
 			const { roomId } = this.bodyParams;
-			if (!roomId || roomId.trim() === '') {
-				return API.v1.failure('Invalid room Id');
+			if (roomId.trim() === '') {
+				throw new Error('error-invalid-room');
```
ggazzo, line 18: using the validator is possible to receive an empty string?

apps/meteor/client/views/room/components/body/composer/ComposerOmnichannel/hooks/useResumeChatOnHoldMutation.ts
```diff
@@ -1,18 +1,24 @@
...
 	const queryClient = useQueryClient();
 
 	return useMutation(
-		async (rid) => {
-			await resumeChatOnHold(rid, { clientAction: true });
+		async (roomId) => {
+			try {
+				await resumeChatOnHold({ roomId });
+			} catch (error) {
+				dispatchToastMessage({ type: 'error', message: error });
```
ggazzo, line 20: do not handle errors inside the mutation, there is a prop to manage that

apps/meteor/tests/data/livechat/rooms.ts
```diff
@@ -319,3 +319,11 @@ export const startANewLivechatRoomAndTakeIt = async ({
 
 	return { room, visitor };
 };
+
+export const placeRoomOnHold = async (roomId: string): Promise<void> => {
```
ggazzo, line 323: what do you think we don't do this anymore? I see several of these functions being created, one for each call, this is almost what an sdk should do.

it seems to me that we are meta programming the tests, sometimes more logic is introduced into these methods and it becomes difficult to understand what a simple post call should be doing

apps/meteor/ee/app/livechat-enterprise/server/hooks/resumeOnHold.ts
```diff
@@ -1,46 +1,68 @@
...
-		}
+	return message;
+};
 
-		return message;
-	},
-	callbacks.priority.HIGH,
-	'livechat-resume-on-hold',
-);
+callbacks.add('afterSaveMessage', handleAfterSaveMessage, callbacks.priority.HIGH, 'livechat-resume-on-hold');
```
ggazzo, line 68: Was this role re-created by some conflict resolution?

ggazzo (changes requested):
this is not a simple fix, correct? several modifications and even a new endpoint was created. remember to keep the pull request as short and direct as possible, otherwise some reasons and reasons may get lost in the middle of so much change

apps/meteor/ee/app/livechat-enterprise/server/api/rooms.ts
murtaza98: Yups, you're right. This is not needed

apps/meteor/ee/app/livechat-enterprise/server/hooks/resumeOnHold.ts
murtaza98: Sorry I didn't understood what you meant here. Could you please elaborate?

apps/meteor/tests/data/livechat/rooms.ts
murtaza98: I'm not sure I got this as well. Are you suggesting to create a SDK for testing or perhaps use the existing SDK somehow over here?
The main reason y I added this method because I had to use the same API call on 2 different tests. So instead of rewriting the whole logic for requests, I isolated it to a method call and called the method in these 2 tests. This makes the code more readable IMO, plus creates a helper which we can re-use in the future if we want to in other tests.
image

apps/meteor/client/views/room/components/body/composer/ComposerOmnichannel/hooks/useResumeChatOnHoldMutation.ts
murtaza98: Thanks for the tip! I wasn't aware of this

aleksandernsilva (dismissed):

KevLehman (approved):
Think we can change the title to "improve: xxxx" as this is a ton of changes for a fix :)

murtaza98:
> Ok, unit test fail is never a good sign grimacing

Not sure... I saw it fail on an another PR. Let's retry and see

aleksandernsilva (dismissed):

KevLehman (dismissed):

ggazzo (approved):

murtaza98:
Hello @KevLehman @aleksandernsilva @ggazzo Could you please re-approve this PR? Your previous approvals got lost while solving conflicts

aleksandernsilva (approved):

KevLehman (approved):

PR #28254, opened 2023-03-03
Regression: Fix intermitent apps engine startup

thassiov:

Makes the apps limitation functionality depend on methods from the Apps Engine's service instead of a dynamic import of the apps orchestrator, making the engine's load process deterministic.

## Steps to test or reproduce

It's an "trial and error" situation. When starting up the server, the apps engine might load or not.

### DIFF
--- a/apps/meteor/ee/app/license/server/lib/isUnderAppLimits.ts
+++ b/apps/meteor/ee/app/license/server/lib/isUnderAppLimits.ts
@@ -1,13 +1,10 @@
-import type { AppManager } from '@rocket.chat/apps-engine/server/AppManager';
+import type { IAppInfo } from '@rocket.chat/apps-engine/definition/metadata';
+import { Apps } from '@rocket.chat/core-services';
 
 import type { ILicense, LicenseAppSources } from '../../definition/ILicense';
 
-export async function isUnderAppLimits(
-	{ appManager }: { appManager: AppManager },
-	licenseAppsConfig: NonNullable<ILicense['apps']>,
-	source: LicenseAppSources,
-): Promise<boolean> {
-	const apps = appManager.get({ enabled: true }).filter((app) => app.getStorageItem().installationSource === source);
+export async function isUnderAppLimits(licenseAppsConfig: NonNullable<ILicense['apps']>, source: LicenseAppSources): Promise<boolean> {
+	const apps = Apps.getApp({ enabled: true }).filter((app: IAppInfo) => Apps.getStorageItemById(app.id).installationSource === source);
 	const configKey = `max${source.charAt(0).toUpperCase()}${source.slice(1)}Apps` as keyof typeof licenseAppsConfig;
 	const configLimit = licenseAppsConfig[configKey];
 
--- a/apps/meteor/ee/app/license/server/lib/isUnderAppLimits.ts
+++ b/apps/meteor/ee/app/license/server/lib/isUnderAppLimits.ts
@@ -3,16 +3,14 @@ import { Apps } from '@rocket.chat/core-services';
 import type { ILicense, LicenseAppSources } from '../../definition/ILicense';
 
 export async function isUnderAppLimits(licenseAppsConfig: NonNullable<ILicense['apps']>, source: LicenseAppSources): Promise<boolean> {
-	const apps = await Apps.getApp({ enabled: true });
+	const apps = await Apps.getApps({ enabled: true });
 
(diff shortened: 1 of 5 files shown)


apps/meteor/ee/app/license/server/lib/isUnderAppLimits.ts
```diff
@@ -1,13 +1,19 @@
...
+	const apps = await Apps.getApp({ enabled: true });
+
+	if (!apps || !Array.isArray(apps)) {
+		return true;
+	}
+
+	const promisedStorageItems = apps.map(async (app) => Apps.getAppStorageItemById(app.id));
+	const activeAppsFromSameSource = await Promise.all(promisedStorageItems).then((items) =>
+		items.filter((item) => item && item.installationSource === source),
+	);
```
sampaiodiego, line 15: it feels very confusing if you mix `await` and `then`, specially on the same command.

```suggestion
	const storageItems = await Promise.all(apps.map(async (app) => Apps.getAppStorageItemById(app.id)));
	const activeAppsFromSameSource = storageItems.filter((item) => item && item.installationSource === source);
```

sampaiodiego (changes requested):
I have these suggestions code wise.. still testing the fix meanwhile 😉

thassiov:
> I have these suggestions code wise.. still testing the fix meanwhile wink

Adjustments done. Looks better indeed

apps/meteor/ee/app/license/server/lib/isUnderAppLimits.ts
```diff
@@ -1,13 +1,17 @@
...
-	const apps = appManager.get({ enabled: true }).filter((app) => app.getStorageItem().installationSource === source);
+export async function isUnderAppLimits(licenseAppsConfig: NonNullable<ILicense['apps']>, source: LicenseAppSources): Promise<boolean> {
+	const apps = await Apps.getApps({ enabled: true });
+
+	if (!apps || !Array.isArray(apps)) {
+		return true;
+	}
+
+	const storageItems = await Promise.all(apps.map((app) => Apps.getAppStorageItemById(app.id)));
+	const activeAppsFromSameSource = storageItems.filter((item) => item && item.installationSource === source);
```
KevLehman, line 13: item?.installationSource works?

thassiov: It does indeed. Ty!

packages/core-services/src/index.ts
```diff
@@ -115,6 +115,7 @@ export {
 
 // TODO think in a way to not have to pass the service name to proxify here as well
 export const Authorization = proxifyWithWait<IAuthorization>('authorization');
+export const Apps = proxifyWithWait<IAppsEngineService>('apps-engine');
```
ggazzo, line 118: worries me a little that the best way to avoid cyclical imports is to inject the object after

apps/meteor/ee/app/license/server/license.ts
```diff
@@ -237,11 +220,11 @@ class LicenseClass {
...
 
 	async canEnableApp(source: LicenseAppSources): Promise<boolean> {
-		if (!this.Apps?.isInitialized()) {
+		if (!Apps.isInitialized()) {
 			return false;
 		}
 
-		return isUnderAppLimits({ appManager: this.Apps.getManager() as AppManager }, this.appsConfig, source);
+		return isUnderAppLimits(this.appsConfig, source);
 	}
```
ggazzo, line 228: this is not supposed to be a license service concern

ggazzo (changes requested):
I would take a completely different approach, really avoiding Apps using License and License using Apps that's too bad

apps/meteor/ee/app/license/server/license.ts
d-gubert: In the same vein, you would say that the `canAddNewUser` method is not supposed to be a license service concern as well, correct?

sampaiodiego (approved):
While I agree the concerns might not be split the best way as possible, this is not supposed to fix that.. This should fix the regression making apps unusable.

We can address other issues and improvements later.

apps/meteor/ee/app/license/server/license.ts
ggazzo: sure. the best explanation I can give you: I was not involved on `canAddNewUser` review  🤷

PR #28266, opened 2023-03-03
Regression: Disabled grandfathered apps must have the option to be enabled

d-gubert:

Grandfathered apps will not follow the new enabling flow and it should always be possible to enable them

## Further comments

MKP-273

### DIFF
--- a/apps/meteor/client/views/marketplace/AppMenu.js
+++ b/apps/meteor/client/views/marketplace/AppMenu.js
@@ -344,7 +344,7 @@ function AppMenu({ app, isAppDetailsPage, ...props }) {
 								{t('Enable')}
 							</>
 						),
-						disabled: !appCountQuery?.data?.hasUnlimitedApps && appCountQuery?.data?.enabled >= appCountQuery?.data?.limit,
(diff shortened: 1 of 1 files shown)


apps/meteor/client/views/marketplace/AppMenu.js
```diff
@@ -344,7 +344,7 @@ function AppMenu({ app, isAppDetailsPage, ...props }) {
 								{t('Enable')}
 							</>
 						),
-						disabled: !appCountQuery?.data?.hasUnlimitedApps && appCountQuery?.data?.enabled >= appCountQuery?.data?.limit,
+						disabled: !app.migrated && !appCountQuery?.data?.hasUnlimitedApps && appCountQuery?.data?.enabled >= appCountQuery?.data?.limit,
```
sampaiodiego, line 347: question, is this just a client validation? isn't there anything preventing them from being enabled on server side as well?

d-gubert: There is, yes

tiagoevanp (approved):

sampaiodiego (approved):

Issue #27206, opened 2022-11-07
Message_QuoteChainLimit has off-by-one error


nmagedman:
### Description:
Commit 5afe5ceb72888f0a0d51544d89fe1b4567ff5e85 ("Chore: Migrate oembed to ts (#25622)", released in RC v5.0.0) introduced a regression in the recursive processing of `Message_QuoteChainLimit`.  The change creates an off-by-one error, cutting off the quote chain at the wrong point.

So, for example, if Message_QuoteChainLimit is at its default value of 2 and I quote a message that itself quotes a chain of other messages, RC v4.x would show:

- my new message, quoting...
  - an older message, quoting...
    - a still older message

However, under RC v5.x, without changing any configuration, the same action would show:

- my new message, quoting...
  - an older message, quoting...
    - a still older message, quoting...
      - a very old message

Additionally, Message_QuoteChainLimit has a bug in that setting it to 1 (or 0) isn't honored.  The implementation always truncates at no less than 2.

As an aside, the documentation for this setting does not clearly define what constitutes a Quote Chain.
I wasn't certain whether "an older message" (in the above example) counted as part of the quote chain or not; that is, is the _first_ quoted message not considered _"chained"_ (excluding it from the count)?  Or is the _entire_ chain counted as a group (including it in the count)?  I turned to the documentation for clarification, but https://docs.rocket.chat/guides/administration/admin-panel/settings/message says merely:
 > **Maximum Number of Chained Quotes:** Lets you set the maximum number of Chained quotes.

### Steps to reproduce:

1. Go to /admin/settings/Message and set **Maximum Number of Chained Quotes** to  2
2. Enter a chat room and post a message.
3. Click on the "Curly Quotes" Quote icon for the newly created message.
4. Type in a message and post it.
5. Repeat steps 3-4 several times, each time clicking [Quote] on the most recently created message.

### Expected behavior:

The fourth message should not include the original message in the quote chain.  Instead, message number 2 should be the oldest post in the chain.

### Actual behavior:

The fourth message includes all 3 prior messages in the quote chain, including the original message, even though we have exceeded the limit of 2.

<img width="254" alt="Screen Shot 2022-11-07 at 23 19 19" src="https://user-images.githubusercontent.com/1042298/200417673-29e754ec-e2c8-4807-94ea-5758e423c366.png">

### Server Setup Information:

- Version of Rocket.Chat Server:  5.1.2
- Operating System: Linux
- Deployment Method: tar
[...]

PR #28281, opened 2023-03-05
fix: Quotes chain off by one error in quote chain limit settings

jayesh-jain252:

Fixed off by one error in Quote Chain Limit. Now if a user sets Message_QuoteChainLimit (default is 2) the number of quotes chained matches the setting.
Changes are made in apps/meteor/app/oembed/server/jumpToMessage.ts file

With Message_QuoteChainLimit=2
Before:
quote_limit2_before

After:
quote_limit2_after

## Issue(s)

#27206

## Further comments

Remaining Changes:
- [x] Message_QuoteChainLimit 1 and 0 results in quote chain of default value 2

TC-256


hugocostadev:
Hi @jayesh-jain252 thanks for your contribution, I'll review it soon

jayesh-jain252:
If a user sets the value of "Maximum Number of Chained Quotes" setting (at  /admin/settings/Message) 1 or 0, the setting does not apply and quotes chain length remains 2 (default value).

As per my view:
If a user sets "Maximum Number of Chained Quotes" = 0 then there should be no message in quote chain or user should be prompted "not allowed to quote message"
If a user sets "Maximum Number of Chained Quotes" = 1 then there should be only one message in quote chain 
Mock (should work like this):
quote_val1_1  quote_val1_2

But in Both cases (0 and 1) Its get a default chain of 2 messages
quote_val2_2

jayesh-jain252:
Do let me know if my understanding is wrong here

hugocostadev:
You are partially right @jayesh-jain252, in case of 0 as the value of the setting we should consider the same as 1...

I made a commit to complement the task

jayesh-jain252:
I tried logging the data received from data.ts file(/apps/meteor/client/lib/chats/data.ts)
1. When there is no quote data returned is just plain text

No quote

2. When we quote a message data returned is in form of `[ ] (URL) /n message`

one_quote

3. Chaining Quotes for second time. There is no URL present in Quote Preview

no_url

4. If we chain quotes for more than 2 times. There is URL in Quote Preview

url_present

I am guessing that this value of msg is given to QuoteAttachment as text prop
(at apps/meteor/client/views/room/components/body/composer/messageBox/MessageBoxReplies.tsx)
But I am not able to understand why URL is shown after chaining for more than 2 times

@hugocostadev would you please guide me on how should I approach this problem.

guijun13 (approved):

hugocostadev:
> composer/messageBox/MessageBoxReplies.tsx) But I am not able to understand why URL is shown after chaining for more than 2 times

Hi @jayesh-jain252 you can create a separate issue/pull request for that issue, I need to investigate here to check what is happening...

jayesh-jain252:
> 
> Hi @jayesh-jain252 you can create a separate issue/pull request for that issue, I need to investigate here to check what is happening...

Sure, I will do that..

jayesh-jain252:
> > composer/messageBox/MessageBoxReplies.tsx) But I am not able to understand why URL is shown after chaining for more than 2 times
> 
> Hi @jayesh-jain252 you can create a separate issue/pull request for that issue, I need to investigate here to check what is happening...

Hi @hugocostadev, I opened a new PR for the link appearing in quotes preview #28443, so I am removing the second checklist from Remaining Changes.

guijun13 (approved):

PR #28312, opened 2023-03-07
Chore: Change names of requested apps sorting options

rique223:

Changed the name of the requested app-related sorting options to cause less confusion regarding their expected behavior and solved some UX problems with the initial state of the sorting filter when on different marketplace pages. Also fixed some typos.
Demo image:
image


dougfabris (changes requested):
why not *JUST* check if `isRequested` on the initial `setSortFilterStructure`
no need at all for this `useEffect`

rique223:
> why not **JUST** check if `isRequested` on the initial `setSortFilterStructure`?
> no need at all for this `useEffect`

Because what you are describing, if I understood correctly, would only work on the first access of this component and would not change to the "unread requested first" sorting when accessing the requested apps page.

dougfabris (approved):

PR #28317, opened 2023-03-07
Chore: remove unused prop in register workspace modal

csuadev:
Remove prop that wasn't needed at the end in the registering workspace modal.


jcbrtl (approved):
This is just a cleanup but I'm kinda paranoid at this stage, so we double checked everything anyway (thanks @janainaCoelhoRocketchat ) :+1:

yash-rajpal (approved):

Issue #28223, opened 2023-03-01
No Feedback for invalid password or user Id


Aashray446:
### Steps to reproduce:

1. Go to Create an account on the main Login page
2. Try creating a profile with invalid usernames like ak@123 or with an invalid password

### Expected behavior:
The user should get feedback if the password policy doesn't match or for any other invalid input

### Actual behavior:

1. There is no feedback but there is an error message in the console
2. If we correct our input and re-submit the form, there are chances that we get "Email Already registered".

### No Feedback screenshot
![issue-5](https://user-images.githubusercontent.com/69245931/222179590-5652a297-1b76-42d7-9aa8-e4aed1464d45.gif)
### Email already registered even it's fresh attempt after invalid input attempt
![issue-6](https://user-images.githubusercontent.com/69245931/222179546-030ffda5-1a8b-428b-8954-28cf05098fbd.gif)

### Server Setup Information:

- Version of Rocket.Chat Server: 6.0.0-rc.3
- Operating System: Linux
- Deployment Method: 
- Number of Running Instances: 
- DB Replicaset Oplog: Enabled
- NodeJS Version: 14.21.2
- MongoDB Version: 5.0.5

### Client Setup Information

- Desktop App or Browser Version:
- Operating System:

Aashray446:
@dudanogueira I am currently working on this issue. My approach is to add missing error-catching conditional statements , same as here : 
![image](https://user-images.githubusercontent.com/69245931/222385669-c3860637-1c7a-4bdb-bfb2-2b30d41fe272.png)

PR #28329, opened 2023-03-08
[FIX] Invalid Username prompt shown to user for invalid input 

chinma-yyy:
### Before
Screenshot from 2023-03-08 21-46-03

### After
Screenshot from 2023-03-08 21-44-33

## Issue(s)

#28223 One task of this issue is fixed where the invalid username prompt is given to user

On create an account page enter any username that doesn't match the username policy but the prompt is not shown and also the user is created irrespective of the username validation which leads to **Email-already-exists** error due to user creation in database

## Steps to test or reproduce

On create an account page enter any invalid username 

## Further comments

So I have added a direct check before registerUser method call just after the API call to check for username validation which returns API failure on invalid username . Earlier the user was being registered and then the username was being set by calling setUsername method which resulted in unacceptable results.

### DIFF
--- a/apps/meteor/app/api/server/v1/users.ts
+++ b/apps/meteor/app/api/server/v1/users.ts
@@ -44,6 +44,7 @@ import { resetTOTP } from '../../../2fa/server/functions/resetTOTP';
 import { isValidQuery } from '../lib/isValidQuery';
 import { getURL } from '../../../utils/server';
 import { getUploadFormData } from '../lib/getUploadFormData';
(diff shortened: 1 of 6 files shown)


chinma-yyy:
Hey @dudanogueira can you have a look ?✌️

chinma-yyy:
Hey @hugocostadev can you review this . This is a very important issue.

chinma-yyy:
Hey @dudanogueira  @hugocostadev  any comments?

chinma-yyy:
hey @yash-rajpal the pr is ready with changes!!

PR #28330, opened 2023-03-08
[NEW] Federation public room search

MarcosSpessatto:


lmauromb (dismissed):

lmauromb (dismissed):

lmauromb (approved):

gabriellsh (approved):

apps/meteor/tests/e2e/federation/page-objects/fragments/home-sidenav.ts
```diff
@@ -61,9 +61,8 @@ export class FederationSidenav {
 		await this.page.locator('role=navigation >> role=button[name=Search]').click();
 		await this.page.locator('role=search >> role=searchbox').focus();
 		await this.page.locator('role=search >> role=searchbox').type(name);
-		await this.page.locator(`role=search >> role=listbox >> role=link >> text="${name}"`).first().waitFor();
-		await this.page.waitForTimeout(2000);
-		await this.page.locator(`role=search >> role=listbox >> role=link >> text="${name}"`).first().click();
+		await this.page.waitForTimeout(3000);
+		await this.page.locator(`role=search >> role=listbox >> role=link`).first().click();
```
ggazzo, line 65: is `await this.page.locator(`role=search >> role=listbox >> role=link >> text="${name}"`).first().waitFor();` not enough? I can see why waiting after the element is displayed brings any value to the test
I'm against timeout in tests, usually introduces flaky tests

packages/core-typings/src/IRoom.ts
```diff
@@ -124,6 +124,8 @@ export const isDiscussion = (room: Partial<IRoom>): room is IRoom => !!room.prid
 export const isPrivateDiscussion = (room: Partial<IRoom>): room is IRoom => isDiscussion(room) && room.t === 'p';
 export const isPublicDiscussion = (room: Partial<IRoom>): room is IRoom => isDiscussion(room) && room.t === 'c';
 
+export const isPublicRoom = (room: Partial<IRoom>): room is IRoom => room.t === 'c';
```
ggazzo, line 127: I know that you are just following the pattern from the functions above, but none of them are really good, receiving a partial and assuring this is a full complete type just because we tested one prop is not a good thing

ggazzo (approved):
Just see the comments nothing really to worry but if you have a better solution to be added fell free to do so.

ggazzo (approved):

PR #28349, opened 2023-03-10
chore: upgrade meteor to version 2.11.0

ggazzo:


sampaiodiego (changes requested):
still missing at least:

- update NodeJS version
- update TS version

Issue #28193, opened 2023-02-27
Emoji don't work after text in continous


Pheewww:
### Description:

Emoji don't work after text without space.

### Steps to reproduce:

1. Go to any text channel
2. send a message with an emoji that is without space.

### Expected behavior:

The emoji is sent with the text message.

### Actual behavior:

The emoji text for system rendering is sent along with your message.

[emoji_after_text.webm](https://user-images.githubusercontent.com/67512605/221626952-7e8fc088-07b7-47dc-9298-f1c3578a9851.webm)

### Client Setup Information

- Desktop App or Browser Version: firefox
- Operating System: win 10

Pheewww:
I want to know how the Unicode Support is created and how Rendering and Displaying of unicode characters is done. i can certainly work in it also.

Bankar-Swapnil:
Hello is this issue still available? Can I work on this

Mahmoudgalalz:
I'm working on it, and I will send the PR sooner,

shevchukma:
Already discussed [#27505](https://github.com/RocketChat/Rocket.Chat/issues/27505)

PR #28216, opened 2023-03-01
fix: Emoji don't work after text in contiunous

Bankar-Swapnil:
For solving this issue,
changes done in apps/meteor/client/views/room/components/body/composer/messageBox/MessageBox.tsx file and checked it on my local environment and it works fine. Please review this.

## Proposed changes (including videos or screenshots)

https://user-images.githubusercontent.com/98152507/222060666-433e2f2c-2d6e-415d-87ef-d6b55f5ff72a.mp4

## Issue(s)
Closes #28193 

## Steps to test or reproduce
1. Go to any text channel.
2. send a message with an emoji that is without space.
3. now emoji is sent with the text message.

PR #28360, opened 2023-03-10
chore: Emoji not rendering after text #28193

AdityaSingh-02:

when we were using emoji followed by text the emoji used to get converted into text form and not into emoji
so i have just added an space and that fixes the problem 

https://user-images.githubusercontent.com/94185167/224341785-6b98ee9d-6f48-46b1-ab05-b2442ff76e53.mp4

## Issue(s)

Closes -  Emoji don't work after text in continuos #28193
## Steps to test or reproduce

Open any channel and type some text followed by an emoji without adding any space


package.json
```diff
@@ -61,5 +61,8 @@
 	"resolutions": {
 		"minimist": "1.2.6",
 		"adm-zip": "0.5.9"
+	},
+	"dependencies": {
+		"textarea-caret": "^3.1.0"
```
ggazzo, line 66: this dependency in this package is unnecessary

ggazzo (changes requested):

package.json
AdityaSingh-02: I updated this branch and these files came out of nowhere, i have now removed them you can check please

yash-rajpal:
Closing in favor of #28216 

Thanks for your contribution.

Issue #28375, opened 2023-03-12
Labels of theme switch not clickable


paulchen:
### Description:

Rocket.Chat 6.0.0 finally brings the awesome dark mode to the general public. It can be configured in the menu that is shown when clicking the user avatar in the top left corner:

![image](https://user-images.githubusercontent.com/404840/224563911-d88f0f5f-b7f9-4a88-8059-b3aea6c28417.png)

However, the three different options can only be selected by clicking on the radio button. Clicking on the label does nothing.

I find this to be unintuitive: On the one hand, I can click on all the other menu items as well. On the other hand, being able to click on the label of a radio button is pretty common in applications and on website.

### Steps to reproduce:

1. Click on the user avatar in the top left corner to open the user menu.
2. Click on the label of any of the radio buttons in the 'Theme' section.

### Expected behavior:

The radio button next to the label is selected.

### Actual behavior:

Nothing happens.

### Server Setup Information:

- Version of Rocket.Chat Server: 6.0.0
- Deployment Method: Docker
- Number of Running Instances: 1
- MongoDB Version: 6.0.4

### Client Setup Information

- Desktop App or Browser Version:
  - Version 111.0.5563.64 (Offizieller Build) Arch Linux (64-Bit)
  - Mozilla Firefox for Arch Linux 110.0.1 (64-bit)
  - Electron client 3.8.16
- Operating System: Arch Linux

However, I think this behaviour is operating system and browser agnostic.

Issue #27847, opened 2023-01-25
When clicking on display option their checked status does not changes


tushartalesara:
### Description:
When clicking on name of any display option which is unchecked does not changes to checked, but when we hover over them cursor changes to pointer.

### Steps to reproduce:

1. Click on display option in header of sidebar
2. Click/hover on name of any option available

### Expected behavior:
When we click on name of any unchecked option it should get checked.

### Actual behavior:
When we hover cursor changes to pointer but on clicking nothing happens

[simplescreenrecorder-2023-01-26_00.13.11.webm](https://user-images.githubusercontent.com/114513938/214656695-c961f3e7-1adb-41bd-82aa-c57c0e987c89.webm)

### Server Setup Information:

- Version of Rocket.Chat Server: 5.5
- Operating System: Ubuntu 22.04LTS

### Client Setup Information

- Desktop App or Browser Version: Browser Version
- Operating System: Ubuntu 22.04LTS
- NodeJS Versiojn: 14.19.3

tushartalesara:
Should I submit a PR regarding this issue?

Vip2799:
hello, I would like to fix this issue. please assign this issue to me

AdityaSingh-02:
hey @Vip2799 i was already working on it and it is completed now  i have opened PR already

PR #28397, opened 2023-03-13
chore: Theme changes on clicking Label

AdityaSingh-02:

As we know Rocket.chat v6.0.0 came with Different themes , where we could simply switch between dark and light mode, but the issue was that When we were wishing to change theme we could only do that by clicking the radio button, but i have made few changes in file where we can also change theme by clicking the label

https://user-images.githubusercontent.com/94185167/224779377-822bb01c-83d0-4714-b54b-ef4239c96fbf.mp4

## Issue(s)

closes issue Labels of theme switch not clickable https://github.com/RocketChat/Rocket.Chat/issues/28375
## Steps to test or reproduce

Try to change themes


AdityaSingh-02:
yeah sure i would like to solve this at component level, What should i do right now close this PR and submit new one or make changes in this particular one only?

AdityaSingh-02:
hey @ggazzo i had already created an similar PR for issue https://github.com/RocketChat/Rocket.Chat/issues/27847 
my PR #27881 , and i believe only by adding onClick property solves the problem, as back then also i only made few change in click property

ggazzo (approved):

PR #28401, opened 2023-03-13
fix: Generic Table Search

hugocostadev:

When the user changes the page and tries to search for a room that is not currently visible on the page it’s not returning any results. This is also reproducible in the Users table 

**Current Behavior:**
It returns the list filtering by the name
task_table

**Previous Behavior:**
Not found - It was not finding the room because it’s filtering only in the current page
search-pagination-rooms

## Steps to test or reproduce

1. Go to Admin → Rooms list
2. Go to the last page possible
3. Search for a room that is shown on the first page

## Further comments

TC-524

I also added the `refetchOnWindowFocus: false,` to useQuery methods to avoid re-fetching data on user focus on the page, this avoid a lot requests


apps/meteor/client/components/GenericTable/GenericTable.tsx
```diff
@@ -55,14 +55,23 @@ const GenericTable = forwardRef(function GenericTable<
...
-	}, [params, current, itemsPerPage, setParams]);
+		setParams({
+			...params,
+			text: params.text || '',
+			current: prevSearchTerm === params.text ? current : 0,
+			itemsPerPage,
+		});
+
+		setCurrent(prevSearchTerm === params.text ? current : 0);
+		setPrevSearchTerm(params.text || '');
```
ggazzo, line 73: do you know that because of that `prevSearchTerm` you are causing an extra render every time? the effect runs 2x
try to create a different effect only if text changes

ggazzo (changes requested):

apps/meteor/client/components/GenericTable/GenericTable.tsx
```diff
@@ -55,14 +55,23 @@ const GenericTable = forwardRef(function GenericTable<
...
 
 	useEffect(() => {
-		setParams({ ...params, text: params.text || '', current, itemsPerPage });
-	}, [params, current, itemsPerPage, setParams]);
+		setParams({
+			...params,
+			text: params.text || '',
+			current: prevSearchTerm === params.text ? current : 0,
+			itemsPerPage,
+		});
```
tassoevan, line 70:
```suggestion
		setParams((prevParams) => ({
			...params,
			text: params.text || '',
			current: prevParams.text === params.text ? current : 0,
			itemsPerPage,
		}));
```

tassoevan (dismissed):

apps/meteor/client/components/GenericTable/GenericTable.tsx
hugocostadev: Thanks a lot Tasso, I followed your suggestion

apps/meteor/client/components/GenericTable/GenericTable.tsx
hugocostadev: You are right, I followed another approach, thanks!

apps/meteor/client/views/admin/rooms/RoomsTable.tsx
```diff
@@ -219,33 +189,67 @@ const RoomsTable = ({ reload }: { reload: MutableRefObject<() => void> }): React
...
+					<StatesIcon name='magnifier' />
+					<StatesTitle>{t('No_results_found')}</StatesTitle>
+				</States>
+			)}
+			{isError && (
+				<States>
+					<StatesIcon name='warning' variation='danger' />
+					<StatesTitle>{t('Something_went_wrong')}</StatesTitle>
+					<StatesActions>
+						<StatesAction onClick={() => refetch()}>{t('Reload_page')}</StatesAction>
```
ggazzo, line 248: be careful when making major changes unrelated to the correctness or purpose of the PR

In addition, this component is deprecated, we have a 2.0 version so it doesn't seem to make much sense to maintain

The "generic table" was not as generic as we thought, that's why we created V2, as you can see V1 can use v2 (since this one is considerably more generic)

hugocostadev: Got it! Thanks @ggazzo

guijun13 (dismissed):

guijun13 (approved):

PR #28405, opened 2023-03-13
refactor: Feature parity for raw models

KevLehman:
## Further comments

This PR migrates all model functions from meteor models under server/models to raw models.
EE models will be done separately, as they require other things before. 

ARCH-172


apps/meteor/server/models/raw/LivechatRooms.js
```diff
@@ -1,8 +1,10 @@
 import { escapeRegExp } from '@rocket.chat/string-helpers';
+import { Settings } from '@rocket.chat/models';
 
 import { BaseRaw } from './BaseRaw';
 import { getValue } from '../../../app/settings/server/raw';
 import { readSecondaryPreferred } from '../../database/readSecondaryPreferred';
+import { trim } from '../../../lib/utils/stringUtils';
```
sampaiodiego, line 7: this is causing problem since it depends on `dompurify` to provide a `stripTags` function, which you don't even use.

I'd suggest to use native `trim` ;)

sampaiodiego (changes requested):

apps/meteor/app/api/server/v1/im.ts
```diff
@@ -66,13 +66,13 @@ API.v1.addRoute(
...
 	{
-		post() {
+		async post() {
 			const users =
 				'username' in this.bodyParams
 					? [this.bodyParams.username]
 					: this.bodyParams.usernames.split(',').map((username: string) => username.trim());
 
-			const room = createDirectMessage(users, this.userId, this.bodyParams.excludeSelf);
+			const room = await createDirectMessage(users, this.userId, this.bodyParams.excludeSelf);
```
ggazzo, line 75: why are adding await here if this message still uses fibers (and .js) ?

KevLehman: Huh, I changed createDirectMessage to use new models, but seems merge develop overwrote the file 🤔

PR #28413, opened 2023-03-14
fix: Show all users on rooms members list

yash-rajpal:

image

## Further comments

TC-429


apps/meteor/server/lib/findUsersOfRoom.ts
```diff
@@ -33,8 +33,8 @@ export function findUsersOfRoom({
 			federated: 1,
 		},
 		sort: {
-			statusConnection: -1,
 			...(sort || { [settings.get('UI_Use_Real_Name') ? 'name' : 'username']: 1 }),
+			statusConnection: -1,
```
LucianoPierdona, line 37:
```suggestion
			statusConnection: -1,
			...(sort || { [settings.get('UI_Use_Real_Name') ? 'name' : 'username']: 1 }),
```

LucianoPierdona (changes requested):

apps/meteor/server/lib/findUsersOfRoom.ts
yash-rajpal: Done!

LucianoPierdona (approved):

PR #28416, opened 2023-03-14
fix: `emoji` property in messages not being respected

### LINKED ISSUE #26374: Avatar emoji in Incomming webhooks no longer work since 5.0.0
### Description:

The emoji or icon_emoji parameter in incoming WebHooks no longer work, neither does the Emoji option in the GUI setup of the WebHook. The message just shows the default profile picture of the "post as" user.

### Steps to reproduce:

1. Create Incoming WebHook that has the "Emoji (optional)" field set(e.g. to :ghost:)
2. Test the WebHook with a curl(or any other tool), preferably with the emoji option set

### Expected behavior:

Message appears with emoji as avatar icon.

### Actual behavior:

Message appears with default "post as" user avatar.

### Server Setup Information:

- Version of Rocket.Chat Server: Rocket 5.0.0 
- Operating System: Debian Buster
- Deployment Method: tar
- DB Replicaset Oplog: 
- NodeJS Version: 14.19.3
- MongoDB Version: 4.4.15

### Client Setup Information

- Desktop App or Browser Version: Tested with Firefox, chrome and desktop app
- Operating System: Win 10

gabriellsh:

There's a property on the message object called `emoji` which allow the sender (bot, app, API, integration, etc) to replace the user avatar with an emoji. Since we rewrote the messages to React, this was not being supported.
image

## Issue(s)

Closes #26374
## Further comments

TC-105


yash-rajpal (dismissed):

apps/meteor/client/components/message/variants/RoomMessage.tsx
```diff
@@ -69,8 +69,9 @@ const RoomMessage = ({ message, sequential, all, mention, unread, context, ignor
 		>
 			<MessageLeftContainer>
 				{!sequential && message.u.username && !selecting && (
-					<UserAvatar
-						url={message.avatar}
+					<MessageAvatar
+						emoji={message.emoji}
```
ggazzo, line 73: why not keeping this explicit?

```
{emoji && <EmojiAvatar />}
{!emoji && <UserAvatar />}
```
?

gabriellsh: Because I would have to replicate this same check in many places (threads, threadlist, discussion list, thread preview, etc). Doing this I keep this logic contained in a single place, which IMO makes it easier to maintain in case we change this.

I'm happy to change this if you think it would be better.

yash-rajpal (dismissed):

yash-rajpal (approved):

tassoevan (approved):

PR #28417, opened 2023-03-14
refactor: Convert chat endpoints to TS

MarcosSpessatto:

This PR converts all the `chat.` endpoint to Typescript.
Jira: ARCH-163


apps/meteor/app/api/server/v1/chat.ts
```diff
@@ -301,9 +311,9 @@ API.v1.addRoute(
 
 API.v1.addRoute(
 	'chat.update',
-	{ authRequired: true },
+	{ authRequired: true, validateParams: isChatUpdateProps },
 	{
-		post() {
+		async post() {
 			check(
```
KevLehman, line 317: If you have the validations from validateParams you can remove check

apps/meteor/app/api/server/v1/chat.ts
```diff
@@ -1,26 +1,49 @@
...
+import { executeSendMessage } from '../../../lib/server/methods/sendMessage';
 
 API.v1.addRoute(
 	'chat.delete',
-	{ authRequired: true },
+	{ authRequired: true, validateParams: isChatDeleteProps },
 	{
-		post() {
+		async post() {
 			check(
```
KevLehman, line 47: If you use validateParams you can remove check

KevLehman (dismissed):

ggazzo (dismissed):

apps/meteor/tests/end-to-end/api/05-chat.js
```diff
@@ -27,7 +27,7 @@ describe('[Chat]', function () {
 				.expect(400)
 				.expect((res) => {
 					expect(res.body).to.have.property('success', false);
-					expect(res.body).to.have.property('error', '[invalid-channel]');
+					expect(res.body).to.have.property('errorType', 'invalid-params');
```
sampaiodiego, line 30: we should not change error types/messages

sampaiodiego (changes requested):

apps/meteor/tests/end-to-end/api/05-chat.js
MarcosSpessatto: I agree with not changing the properties, but what about the messages? Wouldn't it be better to keep the properties as it was, but also maintain the validations for the endpoints?

MarcosSpessatto: I've rolled everything back 🤙

apps/meteor/app/api/server/v1/chat.ts
```diff
@@ -74,18 +71,15 @@ API.v1.addRoute(
 				throw new Meteor.Error('error-roomId-param-invalid', 'The "lastUpdate" query parameter must be a valid date.');
 			}
 
-			let result;
-			Meteor.runAsUser(this.userId, () => {
-				result = Meteor.call('messages/get', roomId, { lastUpdate: new Date(lastUpdate) });
-			});
+			const result = await Meteor.call('messages/get', roomId, { lastUpdate: new Date(lastUpdate) });
```
sampaiodiego, line 74: for next ones it would be good if you already change to `Meteor.callAsync` , this way we don't need to change it again 😛 

`await Meteor.call()` doesn't have any real impact.. but since we're going to change **all** `Meteor.call()` to `Meteor.callAsync` later on, that's ok 😉

sampaiodiego (approved):

Issue #28378, opened 2023-03-13
Routing issue in Private Apps section of Marketplace.


ayush3160:
### Description:

The Private Apps section of the Marketplace seems to have a routing bug where there is unexpected behavior when navigating through different tabs of an app. Upon opening an app, it redirects to (/info/${id_of_app}/details), but when trying to access the logs of the app, the route changes to /info/${id_of_app}/details/logs. Upon returning to the details tab, the route becomes /info/${id_of_app}/details/details. This is not the expected behavior of the routes and indicates a bug in the routing mechanism.
### Steps to reproduce:

1.  Open the Private apps section of marketplace. http://localhost:3000/marketplace/private/list
2. Click on any installed app , it will redirect to details page of the app.
3. Now open the logs of app and notice the route (/details/logs).
4. Come back to details and again notice the route (/details/details).

### Expected behavior:

The expected behaviour of routing should be similar to the routing in marketplace apps. Which is like upon opening an app , it redirects to (/info/${id_of_app}/${version}/details) and for the logs the route should be (/info/${id_of_app}/${version}/logs).

https://user-images.githubusercontent.com/89914602/224616363-ff4a7630-9b28-4220-8169-0f77eb70c629.mp4

### Actual behavior:

https://user-images.githubusercontent.com/89914602/224614658-c72d3cec-8345-45c6-9898-57f4a6141046.mp4

### Server Setup Information:

- Version of Rocket.Chat Server: 6.0.0-develop
- Operating System: Windows 
- Number of Running Instances: 1 
- NodeJS Version: 14

### Client Setup Information

- Desktop App or Browser Version: Chrome
- Operating System: Windows

IllTamer:
Thanks for your feedback @ayush3160 , I will work on this in two days

ayush3160:
Thanks for showing interest @IllTamer , But I am working on this issue. If I'll face any blocker's, I will definietly reach out to you.

dudanogueira:
Nice collaboration here, guys, Congrats!

let us know if of any blockers. I'll raise awareness to our app squad.

Thanks!!

yash-rajpal:
Fixed by #28418

PR #28418, opened 2023-03-14
fix: Routing fixed for private apps in marketplace section

ayush3160:

This pull request fixes the irregular behavior of routing in the Private Apps section of the Marketplace. The issue was caused by the passing of three parameters (id of app, marketplaceVersion, and tab) to the route when opening an app's details. In the case of private apps, the marketplaceVersion was returning undefined, which caused the version to take on the value of the tab when redirecting. To resolve this issue, I have updated the code to pass the version of the app when the marketplaceVersion returns undefined. This fix ensures that the routes behave as expected in the Private Apps section of the Marketplace.

https://user-images.githubusercontent.com/89914602/225092278-48854c4e-9135-4f6a-ae3e-6f99250dfa45.mp4

## Issue(s)

#28378 
## Steps to test or reproduce

1. Open the Private apps section of marketplace. http://localhost:3000/marketplace/private/list
2. Click on any installed app , it will redirect to details page of the app.
3. Now open the logs of app and notice the route (/${version_of_app}/logs).


yash-rajpal (approved):
Thanks for your contribution.

tiagoevanp (approved):

PR #28419, opened 2023-03-14
fix: Capitalized letters are bypassing the alphabetical ordering on Admin lists

LucianoPierdona:

The alphabetical ordering on these lists is listing capitalized words “first”, and only then listing non-capitalized ones.

## Steps to test or reproduce

Create a few rooms/users that should follow alphabetical ordering in a clear way for ease of testing.

1. E.g.: Carry, Chelsea, Cicero, Colleen, and Columbus.
2. Choose one of the rooms/users and change its name/username to a non-capitalized word.
3. For this example, I’ll change “Cicero” to “cicero”
4. Go to the User listing on Admin Setting 

### Expected Behavior

The Users created in the Steps above should appear in this order: 

- Carry; Chelsea; cicero; Colleen; Columbus;

The alphabetical order should be kept despite the capitalized letters.

### Current Behavior

The capitalized words are listed first, The same names as the above section would list:

- Carry; Chelsea; Colleen; Columbus; cicero;

## Further comments

[TC-525]

[TC-525]: https://rocketchat.atlassian.net/browse/TC-525?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


guijun13 (dismissed):

guijun13 (approved):

apps/meteor/app/api/server/lib/rooms.ts
```diff
@@ -47,7 +47,13 @@ export async function findAdminRooms({
 
 	const { cursor, totalCount } = result;
 
-	const [rooms, total] = await Promise.all([cursor.sort(sort || { default: -1, name: 1 }).toArray(), totalCount]);
+	const [rooms, total] = await Promise.all([
+		cursor
```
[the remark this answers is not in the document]
totoi690: I guess the `{ locale: 'en' }` just implements the non case sensitive sort @KevLehman :
https://www.mongodb.com/docs/manual/reference/collation-locales-defaults/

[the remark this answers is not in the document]
matheusbsilva137: Hey @heitortanoue , since 'caseFirst' is the only property that matters to us here, maybe we could specify it instead of the locale. Check my review

apps/meteor/app/api/server/lib/rooms.ts
```diff
@@ -39,7 +39,13 @@ export async function findAdminRooms({
...
+			.collation({ locale: 'en' })
```
matheusbsilva137, line 44:
```suggestion
			.collation({ caseFirst: 'off' })
```

matheusbsilva137 (changes requested):

apps/meteor/app/api/server/lib/rooms.ts
matheusbsilva137: Actually, locale is a required field in collation :grimacing: so dismiss this suggestion, let's move on with the other approach

matheusbsilva137 (dismissed):

matheusbsilva137 (commented):
Approved it since this is working fine, but I still have some performance concerns
Could using `collation` on the `cursor` object affect performance in this case? I think this would only be prejudicial if applied to an aggregation or so, but not sure
cc @sampaiodiego

apps/meteor/tests/end-to-end/api/01-users.js
```diff
@@ -747,6 +747,43 @@ describe('[Users]', function () {
 				.end(done);
 		});
 
+		it('should sort for user statuses and check if user is correctly sorted by name', async () => {
```
heitortanoue, line 750: Change the test description, it is not sorting by user status.

matheusbsilva137: tkss, fixed

PR #28260, opened 2023-03-03
chore: convert Rooms Model Raw to TypeScript

totoi690:
Convert the file `Rooms.ts`, located on `/Rocket.Chat/apps/meteor/server/models/raw/Rooms.ts` to TypeScript

TC-495

PR #28426, opened 2023-03-15
fix: admin room checkbox and refactor

totoi690:
## Description:
Discussion filter is being prioritized on the filter - even if the user selects more the one it will only return discussion if the checkbox is selected

## Preconditions:
Have created dm’s, channels, and teams before starting the testing

## Steps to reproduce:
- Go to Admin -> Rooms
- Select all the checkboxes

## Expected Behavior:
It should filter by all of the filters selected 

## Actual Behavior:
If the user selects discussion, it only filters by discussion - ignoring all the other checkboxes

### DIFF
--- a/apps/meteor/server/models/raw/Rooms.js
+++ b/apps/meteor/server/models/raw/Rooms.js
@@ -62,85 +62,45 @@ export class RoomsRaw extends BaseRaw {
 		return statistic;
 	}
 
-	findByNameOrFnameContainingAndTypes(name, types, discussion = false, teams = false, showOnlyTeams = false, options = {}) {
+	findByNameOrFnameContainingAndTypes(name, types, discussion = false, teams = false, options = {}) {
 		const nameRegex = new RegExp(escapeRegExp(name).trim(), 'i');
 
-		const onlyTeamsQuery = showOnlyTeams ? { teamMain: { $exists: true } } : {};
-
-		const teamCondition = teams
-			? {}
-			: {
-					teamMain: {
-						$exists: false,
-					},
-			  };
-
 		const query = {
-			t: {
-				$in: types,
-			},
-			prid: { $exists: discussion },
-			$or: [
-				{ name: nameRegex, federated: { $ne: true } },
-				{ fname: nameRegex },
+			$and: [
+				name
+					? {
+							$or: [
+								{ name: nameRegex, federated: { $ne: true } },
+								{ fname: nameRegex },
+								{
+									t: 'd',
+									usernames: nameRegex,
+								},
+							],
+					  }
+					: {},
 				{
-					t: 'd',
-					usernames: nameRegex,
+					$or: [
+						{
+							t: {
+								$in: types,
+							},
+						},
+						...(discussion ? [{ prid: { $exists: true } }] : []),
+						...(teams
+							? [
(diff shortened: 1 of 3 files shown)


packages/model-typings/src/models/IRoomsModel.ts
```diff
@@ -12,18 +12,7 @@ export interface IRoomsModel extends IBaseModel<IRoom> {
...
-		discussion?: boolean,
-		teams?: boolean,
-		showOnlyTeams?: boolean,
-		options?: any,
-	): any;
-
-	findByTypes(types: any, discussion?: boolean, teams?: boolean, onlyTeams?: boolean, options?: any): any;
-
-	findByNameOrFnameContaining(name: any, discussion?: boolean, teams?: boolean, onlyTeams?: boolean, options?: any): any;
+	findByNameOrFnameContainingAndTypes(name: any, types: any, discussion?: boolean, teams?: boolean, options?: any): any;
```
gabriellsh, line 15: Is it possible to correct the typing since you've modified this line?

apps/meteor/server/models/raw/Rooms.js
```diff
@@ -62,87 +62,50 @@ export class RoomsRaw extends BaseRaw {
...
+					  }
+					: {},
 			],
-			...teamCondition,
-			...onlyTeamsCondition,
+			...(!discussion ? { prid: { $exists: false } } : {}),
+			...(!teams ? { teamMain: { $exists: false } } : {}),
 		};
 
+		console.log(JSON.stringify(query));
```
gabriellsh, line 108: Log :)

packages/model-typings/src/models/IRoomsModel.ts
totoi690: Heyy, I rewrote all this file to typescript on [this PR](https://github.com/RocketChat/Rocket.Chat/pull/28260)

apps/meteor/app/api/server/v1/rooms.ts
```diff
@@ -387,7 +387,7 @@ API.v1.addRoute(
 				await findAdminRooms({
 					uid: this.userId,
 					filter: filter || '',
-					types: types || [],
+					types: types && !Array.isArray(types) ? [types] : undefined,
```
LucianoPierdona, line 390: Do you know if there's any case where `Array.isArray(types)` is truthy? And if there is, shouldn't we support that?

totoi690: There is an error here that should be fixed in another PR: when you send only one type (like ['p']), it comes along as a string, not inside an array.

LucianoPierdona (dismissed):

matheusbsilva137 (dismissed):

matheusbsilva137 (dismissed):

apps/meteor/server/models/raw/Rooms.js
matheusbsilva137, line 1: We should remove this file, it doesn't exist in `develop` anymore

matheusbsilva137 (changes requested):

matheusbsilva137 (approved):

matheusbsilva137 (approved):

PR #28429, opened 2023-03-15
fix: Remove read receipts on messages deletion

### LINKED ISSUE #27926: Not removed read receipts for non-existent messages
### Description:

We have Retention Policy enabled. Messages are automatically deleted, but their read receipts remain in the rocketchat_read_receipts collection. 

### Steps to reproduce:

1. Enable Retention Policy 
2. Wait until old messages are deleted
3. Check old entries in the collection rocketchat_read_receipts.

### Expected behavior:

When the Retention Policy is enabled, when messages are deleted, the read reports for them are also deleted.
The messages have already been deleted. It makes no sense to keep records of their reading.

### Actual behavior:

In our test environment, the Retention Policy is set to TWO DAYS, with read receipts stored for years.
For example, here is the oldest report for November 30, 2021.
` db.rocketchat_read_receipts.find().sort({ts:1}).limit(1);
{ "_id" : "HimeyueXs4cK8NWs8", "roomId" : "4uQ5KRSPzYnRhscvKM9pgeKRjfGpTQZ3gn", "userId" : "M9pgeKRjfGpTQZ3gn", "messageId" : "Q8h9hyCSb94wWaMWW", "ts" : ISODate("2021-11-30T06:44:12.257Z"), "_updatedAt" : ISODate("2021-11-30T06:44:12.258Z") }`

### Server Setup Information:

- Version of Rocket.Chat Server: 5.4.1
- Operating System: Oracle Linux Server release 8.6
- Deployment Method: docker
- DB Replicaset Oplog: yes
- NodeJS Version: v14.19.3
- MongoDB Version: 5.0.14 

### Client Setup Information

- Desktop App or Browser Version: Chrome 108
- Operating System: Win10

matheusbsilva137:

 - Delete read read receipts on messages deletion (for individual messaages, "prune messages" option or deleting rooms).

## Issue(s)

Fix #27926

## Steps to test or reproduce

- When erasing a single message, all of its corresponding read receipts shouild also de removed;
- When erasing a room, all of its messages and read receipts should also be removed;
- When using the "Prune Messages" options in a room, read receipts should be removed along with their corresponding messages (read receipts from messages that were not chosen to be deleted should remain).

## Further comments

The exceptions provided to the "Prune Messages" feature will only be applied to read receipts from new thread messages, discussions or messages pinned after updating to this PR (for old thread messages, discussions or messages pinned before the update, read receipts will be removed regardless of the exceptions chosen). That's an intended behavior, chosen to avoid huge migrations or slowliness in room deletions.

TC-457


apps/meteor/ee/server/models/raw/ReadReceipts.ts
```diff
@@ -16,4 +17,83 @@ export class ReadReceiptsRaw extends BaseRaw<ReadReceipt> implements IReadReceip
...
+			query.drid = { $exists: false };
+		}
+
+		if (ignoreThreads) {
+			query.tmid = { $exists: false };
+			query.tcount = { $exists: false };
+		}
+
+		if (users.length) {
+			const uids = await Users.findByUsernames(users, { projection: { _id: 1 } })
```
sampaiodiego, line 79: pls don't call a model from another model

apps/meteor/server/models/raw/Messages.ts
```diff
@@ -608,19 +614,21 @@ export class MessagesRaw extends BaseRaw<IMessage> implements IMessagesModel {
...
-					'otr',
-					otrSystemMessages.USER_JOINED_OTR,
-					otrSystemMessages.USER_REQUESTED_OTR_KEY_REFRESH,
-					otrSystemMessages.USER_KEY_REFRESHED_SUCCESSFULLY,
-				],
+				$in: OTRMessageTypes,
 			},
 			ts: { $lte: ts },
 		};
+		await ReadReceipts.removeByRoomIdAndTypesUntilDate(roomId, OTRMessageTypes, ts);
```
sampaiodiego, line 631: pls don't call a model from another model

apps/meteor/server/models/raw/Messages.ts
```diff
@@ -1092,6 +1100,7 @@ export class MessagesRaw extends BaseRaw<IMessage> implements IMessagesModel {
 			},
 		};
 
+		await ReadReceipts.setPinnedByMessageId(_id, pinned);
```
sampaiodiego, line 1103: pls don't call a model from another model

apps/meteor/server/models/raw/Messages.ts
```diff
@@ -1247,7 +1256,8 @@ export class MessagesRaw extends BaseRaw<IMessage> implements IMessagesModel {
 
 	// REMOVE
 
-	removeByRoomIds(rids: string[]): Promise<DeleteResult> {
+	async removeByRoomIds(rids: string[]): Promise<DeleteResult> {
+		await ReadReceipts.removeByRoomIds(rids);
```
sampaiodiego, line 1260: pls don't call a model from another model

apps/meteor/server/models/raw/Messages.ts
```diff
@@ -1310,6 +1320,7 @@ export class MessagesRaw extends BaseRaw<IMessage> implements IMessagesModel {
 
 		if (!limit) {
 			const count = (await this.deleteMany(query)).deletedCount;
+			await ReadReceipts.removeByIdPinnedTimestampLimitAndUsers(rid, pinned, ignoreDiscussion, ts, users, ignoreThreads);
```
sampaiodiego, line 1323: pls don't call a model from another model

apps/meteor/server/models/raw/Messages.ts
```diff
@@ -1310,6 +1320,7 @@ export class MessagesRaw extends BaseRaw<IMessage> implements IMessagesModel {
...
 
 			// decrease message count
 			await Rooms.decreaseMessageCountById(rid, count);
```
sampaiodiego, line 1326: I know you'll see this and will think that I'm crazy since we're already using models inside of models.. but this is legacy code and is considered technical debt that should be removed.

sampaiodiego (changes requested):
I have added a few comments in places where you're calling a model from inside another model. this is an anti-pattern and should not be accepted.

from an architecture point of view models (or the data layer) are dumb, they don't know any business logic, so they don't how different models should interact with other.

so whenever you're calling another model you're most likely adding some business logic, for example: when a message is inserted we should bump a field on rooms model. so this is *not* a concern of Message model, which means this logic should be added in a higher layer of the stack.. in our case this layer is the "service layer".. so for this example we should have a Message service with an "create" action, and this business logic should leave there =)

I hope this helps you to understand the concepts and provide enough information to refactor this PR. if you still have any question feel free to reach me out 😉

apps/meteor/server/models/raw/Messages.ts
matheusbsilva137: Fixed

apps/meteor/server/models/raw/Messages.ts
matheusbsilva137: Fixed! For this one I had to change the value returned by this method so that we know exactly what messages have been deleted

apps/meteor/server/models/raw/Messages.ts
matheusbsilva137: I completely understand hahah I've just removed the `ReadReceipts` model import from inside the `Messages` model.

apps/meteor/app/lib/server/functions/deleteRoom.ts
```diff
@@ -1,11 +1,12 @@
-import { Messages, Rooms, Subscriptions } from '@rocket.chat/models';
+import { Messages, Rooms, Subscriptions, ReadReceipts } from '@rocket.chat/models';
 
 import { callbacks } from '../../../../lib/callbacks';
 import { FileUpload } from '../../../file-upload/server';
 
 export const deleteRoom = async function (rid: string): Promise<void> {
 	await FileUpload.removeFilesByRoomId(rid);
 	await Messages.removeByRoomId(rid);
+	await ReadReceipts.removeByRoomId(rid);
```
KevLehman, line 9: @sampaiodiego , wdyt of making this a `Promise.all` or `allSettled`?

apps/meteor/app/slackbridge/server/SlackAdapter.js
```diff
@@ -1007,12 +1007,11 @@ export default class SlackAdapter {
 			};
 
 			if (!isImporting) {
-				await Messages.setPinnedByIdAndUserId(
-					`slack-${slackMessage.attachments[0].channel_id}-${slackMessage.attachments[0].ts.replace(/\./g, '-')}`,
-					rocketMsgObj.u,
-					true,
-					new Date(parseInt(slackMessage.ts.split('.')[0]) * 1000),
-				);
+				const messageId = `slack-${slackMessage.attachments[0].channel_id}-${slackMessage.attachments[0].ts.replace(/\./g, '-')}`;
```
KevLehman, line 1010: You check for `attachments[0].text` but this both props could be still undefined 👀 

worth checking before using, or are we sure they'll be there?

apps/meteor/app/livechat/server/lib/Livechat.js
```diff
@@ -955,6 +957,7 @@ export const Livechat = {
 		for await (const room of cursor) {
 			await FileUpload.removeFilesByRoomId(room._id);
 			await Messages.removeByRoomId(room._id);
+			await ReadReceipts.removeByRoomId(room._id);
```
KevLehman, line 960: Can you change this to an `allSettled`?

KevLehman (changes requested):

apps/meteor/app/lib/server/functions/deleteRoom.ts
sampaiodiego: this whole PR was actually in my queue to take a deeper look, since I think this is a lot of changes whereas I think to trigger an action after a message was deleted should be only a matter of adding a hook (either a `this.onEvent` or `callback.add(...)`) somewhere.. 

maybe we don't have such hook, but we should.. imagine if for everything we want to do we need to find all these places again and add an additional call to do whatever we need. 🙈

sampaiodiego: looking at this file specifically I can see there is a callback `afterDeleteRoom` .. so instead of adding the `ReadReceipts.removeByRoomId` here, we should actually call it from a callback for `afterDeleteRoom`.

apps/meteor/app/livechat/server/lib/Livechat.js
sampaiodiego: should we rather call `deleteRoom`? 😬

matheusbsilva137: Not sure if we can call `deleteRoom`. This would erase all subscriptions, but looking at the next line, looks like that's not what we want :thinking: maybe @KevLehman can help us here :thinking: 
Changing this to an `allSettled` for now

apps/meteor/app/lib/server/functions/deleteRoom.ts
matheusbsilva137: Moved it to the callback, tkss

I though about moving some more stuff into callbacks, but I guess that's not always possible with the callbacks we have by now.
1. For `cleanRoomHistory` we gotta know exactly what are the messages erased
2. For one specific message being deleted, I thought about calling this model from the `afterDeleteMessage` callback, but I'd have to replicate a big part of the code from the deleteMessage function (since I guess we'd only like to delete read receipts when `Message_KeepHistory` and `Message_ShowDeletedStatus` are disabled). Not sure what you think would be the best option in this case

apps/meteor/app/slackbridge/server/SlackAdapter.js
matheusbsilva137: I've just added the check for both props. We're sure they'll be there, but I guess it's worth checking

matheusbsilva137:
CI is not passing here :man_shrugging: 
I replicated the changes from this PR on #29720

PR #28437, opened 2023-03-16
chore: ddp streamer typings

ggazzo:
first step to type ddp streams responsible for the real data communication


packages/ui-contexts/src/hooks/useStream.ts
```diff
@@ -1,14 +1,54 @@
+import type { ServerStreamerNames, StreamerEvents } from '@rocket.chat/ui-contexts/src/ServerContext/streams';
 import { useContext, useMemo } from 'react';
 
 import { ServerContext } from '../ServerContext';
 
-export const useStream = (
-	streamName: string,
+// T extends (...args: any) => any> = T extends (...args: infer P) => any ? P : never;
```
tassoevan, line 6:
```suggestion
```

packages/ui-contexts/src/hooks/useStream.ts
```diff
@@ -1,14 +1,54 @@
...
+
+/*
+ * @param streamName The name of the stream to subscribe to
+ * @returns A function that can be used to subscribe to the stream
+ * the main difference between this and useStream is that this function
+ * will only subscribe to the `stream + key` only once, but you can still add multiple callbacks
+ * to the same path
+ */
+
+export function useSingleStream<StreamName extends ServerStreamerNames>(
```
tassoevan, line 45: Why not go with this being the new behavior of `useStream`? I'm pretty sure it was the desired one.

tassoevan (changes requested):

tassoevan (approved):

PR #28445, opened 2023-03-16
fix: marketplace requested  glitch

ggazzo:


dougfabris (dismissed):
As we do have permission checking if the user can manage apps, it's better for the user experience to show the option to access the Requested Apps Page through the menu and it's not necessary to show the loader for the option. Because of it, we should just show the loader the data we're waiting for, in that case, the data that renders the `Badge`. In that sense, it removes the glitch and improves the overall experience of opening the dropdown menu. Approving it!

dougfabris (approved):

Issue #25172, opened 2022-04-13
Sidebar collapses on clicking the kebab menu, for currently non-opened channel


Dnouv:
### Description:

On small size screen devices clicking on the kebab menu icon for channels other than the one which is opened collapses the sidebar, and the options float on the chat screen.

### Steps to reproduce:

1. Resize the browser window to be less than 766 px width.
2. Now, the sidebar is collapsed by default.
3. Open any channel; let's call it the "present channel".
4. Open the sidebar by clicking on the hamburger icon in the top left corner.
5. Now click the kebab icon (the one with three dots) to open up the "present channel" options "Hide, Mark Unread,..."
6. Notice that the sidebar remains open.
7. Now, click the kebab menu icon for any other channel; the sidebar is collapsed.

### Expected behavior:

The sidebar remains open.

### Actual behavior:

<img alt="gif showing the issue behavior" src="https://user-images.githubusercontent.com/61188295/163172605-42bb6968-d0f1-426d-b17f-6e232dac9daf.gif" width="400"/>
The sidebar is collapsed, and the options float over the chat window.

### Server Setup Information:

- Version of Rocket.Chat Server: 4.6
- Operating System: Windows
- Deployment Method: Open Server

### Relevant logs:

Nothing logged on console.

PR #28446, opened 2023-03-16
fix: Fixed sidebar collapses on clicking kebab menu

jayesh-jain252:

Sidebar was collapsing on small screen when clicked on kebab menu of channels other than the current channel

After:

Rocket Chat Sidebar After

## Issue(s)

#25172


jayesh-jain252:
Hi, @dougfabris could you review this. Let me know if there is something more I could add.

jayesh-jain252:
@dougfabris I removed the onClick function as it was triggering this behaviour. Let me know if this will work or I need to add something more.

PR #28450, opened 2023-03-16
fix: Changing the app's error verification

AllanPazRibeiro:
This PR is a consequence of this apps engine pr: https://github.com/RocketChat/Rocket.Chat.Apps-engine/pull/598

Due to the limitations of the VM2 sandbox, errors generated by apps were not being properly verified.


apps/meteor/app/file-upload/server/lib/FileUpload.js
```diff
@@ -115,7 +115,7 @@ export const FileUpload = {
 		try {
 			Promise.await(Apps.triggerEvent(AppEvents.IPreFileUpload, { file, content }));
 		} catch (error) {
-			if (error instanceof AppsEngineException) {
+			if (error.name === AppsEngineException.name) {
```
dapi, line 118: Hi, @AllanPazRibeiro !

Why did you choice to use the string comparison instead of type checking? Looks weird

d-gubert: Hey @dapi, this is because the actual code from the app runs inside an isolated context which requires every object it needs enclosed in a sandbox. This means that the error thrown by the app **will never be an instance of the `AppsEngineException` we have access to in our main context**. This was causing problems as Rocket.Chat wasn't able to determine what was an `AppsEngineException` anymore, and as such we decided to add a "type identifier" as a primitive value which we can assert

PR #28456, opened 2023-03-16
refactor: Models feature parity

KevLehman:
## Further comments

This PR migrates all model functions from meteor models under server/models to raw models.
EE models will be done separately, as they require other things before.

ARCH-172


apps/meteor/app/importer/server/classes/ImportDataConverter.ts
```diff
@@ -371,6 +371,8 @@ export class ImportDataConverter {
 				this.saveError(_id, e instanceof Error ? e : new Error(String(e)));
 			}
 		});
+
+		await Promise.all(promises);
```
murtaza98, line 375: Wondering if we should use AllSettled here instead. So if any one of the users has some record that isn't valid and is throwing errors then only that particular user won't be imported. The rest of the users will still get imported this way.

apps/meteor/ee/server/models/raw/LivechatRooms.ts
```diff
@@ -32,6 +32,7 @@ declare module '@rocket.chat/model-typings' {
 	}
 }
 
+// @ts-expect-error - Model is in JS, and types are getting weird :)
```
murtaza98, line 35: 😢

apps/meteor/server/models/raw/LivechatRooms.js
```diff
@@ -12,6 +13,40 @@ export class LivechatRoomsRaw extends BaseRaw {
 		super(db, 'room', trash);
 	}
 
+	// move indexes from constructor to here using IndexDescription as type
```
murtaza98, line 16:
```suggestion
```

apps/meteor/server/models/raw/LivechatDepartmentAgents.ts
```diff
@@ -113,4 +150,208 @@ export class LivechatDepartmentAgentsRaw extends BaseRaw<ILivechatDepartmentAgen
...
+		return this.updateOne(
+			{
+				agentId: agent.agentId,
+				departmentId: agent.departmentId,
+			},
+			{
+				$set: {
+					username: agent.username,
+					departmentEnabled: agent.departmentEnabled,
+					count: parseInt(`${agent.count}`),
```
murtaza98, line 179:
```suggestion
					count: parseInt(`${agent.count}`),
					order: parseInt(`${agent.order}`),
```
Aren't they already numbers now?

apps/meteor/server/models/raw/LivechatRooms.js
```diff
@@ -12,6 +13,40 @@ export class LivechatRoomsRaw extends BaseRaw {
...
+	modelIndexes() {
```
murtaza98, line 17:
```suggestion
	protected modelIndexes() {
```

apps/meteor/server/models/raw/LivechatDepartment.ts
```diff
@@ -151,4 +188,115 @@ export class LivechatDepartmentRaw extends BaseRaw<ILivechatDepartmentRecord> im
...
+
+	findEnabledWithAgents(projection: FindOptions<ILivechatDepartment>['projection'] = {}): FindCursor<ILivechatDepartment> {
+		const query = {
+			numAgents: { $gt: 0 },
+			enabled: true,
+		};
+		return this.find(query, projection && { projection });
+	}
+
+	findEnabledWithAgentsAndBusinessUnit(
```
murtaza98, line 249: I see an override for this function over here within base model which adds some custom login of Business unit. So wanted to confirm if that wouldn't get affected by this migration

KevLehman: Not yet, as this only migrates the function. There's another PR for updating usages

sampaiodiego (dismissed):

sampaiodiego (approved):

murtaza98 (approved):

PR #28458, opened 2023-03-16
refactor: Remove `ImportData` meteor model

KevLehman:
ARCH-918


apps/meteor/app/importer-slack/server/importer.js
```diff
@@ -235,7 +237,7 @@ export class SlackImporter extends Base {
 			// If we have no slack message yet, then we can insert them instead of upserting
 			this._useUpsert = !Messages.findOne({ _id: /slack\-.*/ });
 
-			zip.forEach((entry) => {
+			const operations = zip.map(async (entry) => {
```
pierre-lehnen-rc, line 240: I'll run some stress tests on this to ensure this change doesn't affect the operation. The zip loop is a very sensitive part of the import process.

KevLehman: Go there. I was thinking on changing this to either this async map or a for await of, let me know your results 🤗

apps/meteor/server/lib/ldap/Connection.ts
```diff
@@ -349,9 +349,9 @@ export class LDAPConnection {
 
 				const entries: T[] = [];
 
-				res.on('searchEntry', (entry) => {
+				res.on('searchEntry', async (entry) => {
```
pierre-lehnen-rc, line 352: ldapjs doesn't support async callbacks. Use `Promise.await` here and let's rework the ldap stuff on a separate PR.

apps/meteor/server/lib/ldap/Connection.ts
```diff
@@ -481,9 +481,9 @@ export class LDAPConnection {
 
 			const entries: T[] = [];
 
-			res.on('searchEntry', (entry) => {
+			res.on('searchEntry', async (entry) => {
```
pierre-lehnen-rc, line 484: same as above.

apps/meteor/server/lib/ldap/Connection.ts
```diff
@@ -547,9 +547,9 @@ export class LDAPConnection {
 			let entries: T[] = [];
 			const internalPageSize = pageSize * 2;
 
-			res.on('searchEntry', (entry) => {
+			res.on('searchEntry', async (entry) => {
```
pierre-lehnen-rc, line 550: same

pierre-lehnen-rc (changes requested):

pierre-lehnen-rc (commented):
Actually, can you leave the whole Connection.ts file untouched (adding Promise.awaits elsewhere)?

Instead of reworking the ldap to work asynchronously, I'll split the Virtual Importer from the Base code so that the BaseImporter will be async and Virtual continues to be sync.

pierre-lehnen-rc (approved):

PR #28472, opened 2023-03-17
refactor: `LivechatDepartmentAgents` removal

KevLehman:
[ARCH-172] [ARCH-926]

## Further comments

[ARCH-172]: https://rocketchat.atlassian.net/browse/ARCH-172?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[ARCH-926]: https://rocketchat.atlassian.net/browse/ARCH-926?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/meteor/app/lib/server/methods/setUsername.ts
```diff
@@ -10,13 +10,6 @@ import { checkUsernameAvailability } from '../functions';
 import { RateLimiter } from '../lib';
 import { saveUserIdentity } from '../functions/saveUserIdentity';
 
-declare module '@rocket.chat/ui-contexts' {
```
sampaiodiego, line 13: this was recent added so I guess we should add it back

apps/meteor/ee/app/livechat-enterprise/server/business-hour/Multiple.ts
```diff
@@ -231,6 +231,7 @@ export class MultipleBusinessHoursBehavior extends AbstractBusinessHourBehavior
 			if ((await LivechatDepartmentAgents.findByAgentId(agentId).count()) === 0) {
 				agentIdsWithoutDepartment.push(agentId);
 			}
+			// TODO: We're doing a full fledged aggregation with lookups and getting the whole array just for getting the length? :(
```
sampaiodiego, line 234: ouch 🙈

apps/meteor/server/models/raw/LivechatDepartmentAgents.ts
```diff
@@ -139,8 +139,8 @@ export class LivechatDepartmentAgentsRaw extends BaseRaw<ILivechatDepartmentAgen
 		return this.find({ departmentId: { $in: departmentIds } }, options);
 	}
 
-	findAgentsByAgentIdAndBusinessHourId(_agentId: string, _businessHourId: string): [] {
-		return [];
+	findAgentsByAgentIdAndBusinessHourId(_agentId: string, _businessHourId: string): Promise<ILivechatDepartmentAgents[]> {
+		return [] as unknown as Promise<ILivechatDepartmentAgents[]>;
```
sampaiodiego, line 143: an empty array is not enough?
```suggestion
	async findAgentsByAgentIdAndBusinessHourId(_agentId: string, _businessHourId: string): Promise<ILivechatDepartmentAgents[]> {
		return [];
```

packages/ui-contexts/src/ServerContext/methods.ts
```diff
@@ -20,6 +20,7 @@ export interface ServerMethods {
 	'updateIncomingIntegration': (...args: any[]) => any;
 	'updateOutgoingIntegration': (...args: any[]) => any;
 	'checkRegistrationSecretURL'(hash: string): boolean;
+	setUsername(username: string, param?: { joinDefaultChannelsSilenced?: boolean }): Promise<string>;
```
sampaiodiego, line 23: oh, now that I get to the end of the page I see you "moved it back" to here.

from what I got the change to a augmentation is the expected as the method typing is not supposed to be spread outside the Meteor project.

sampaiodiego (dismissed):

murtaza98 (dismissed):

apps/meteor/app/lib/server/methods/setUsername.ts
KevLehman: hm, would be cool to know why. I moved the declaration to the actual type on UI-context package.

packages/ui-contexts/src/ServerContext/methods.ts
KevLehman: 🤔 All right, can restore the agumentation. But, why are we doing the augmentation way? Like, we have around 12 methods here, what's the diff between the methods already here and the ones in the augmentation if all of them go to the same type? 👀

sampaiodiego (approved):

PR #28475, opened 2023-03-17
chore: ignore .eslintcache

KevLehman:


ggazzo (changes requested):
instead of only ignoring the file, why not prevent them for being created ?

KevLehman:
I'd say because locally it does help to speed up the lint on subsequent runs. When committed is when it doesn't make sense as the file has the full paths for each file, which will change per machine

PR #28483, opened 2023-03-18
refactor: remove Promise.await

ggazzo:


apps/meteor/app/authorization/server/functions/canSendMessage.ts
```diff
@@ -59,13 +59,16 @@ async function canSendMessageAsync(
 	return room;
 }
 
+/* deprecated */
```
sampaiodiego, line 62: do you know this comment means nothing right?

sampaiodiego (dismissed):

apps/meteor/ee/app/livechat-enterprise/server/lib/VisitorInactivityMonitor.ts
```diff
@@ -43,7 +43,7 @@ export class VisitorInactivityMonitor {
 			name: this._name,
 			schedule: (parser) => parser.cron(everyMinute),
 			job: () => {
-				Promise.await(this.handleAbandonedRooms());
+				return this.handleAbandonedRooms();
```
sampaiodiego, line 46: have you checked if `SyncedCron` supports `Promises`? I mean, it tracks how much time a cronjob has run, so it kinda waits for the function to execute.. 

what if we let this `Promise.await` until we take a look at the `SyncedCron` package?

ggazzo, line 46:
```suggestion
				return Promise.await(this.handleAbandonedRooms());
```

ggazzo: you are right I thought this one was from agenda, the meteor package doenst handle async functions

sampaiodiego (approved):

PR #28490, opened 2023-03-20
chore: Disable omnichannel endpoints and widget serving code with a new setting

cauefcr:

this disables omnichannel entirely from the backend, removing the widget serving code, and the omnichannel specific endpoints, this breaks custom integration when the setting is disabled, as well as custom apps.

## Issue(s)

OC-362
## Steps to test or reproduce

go to the omnichannel settings page, and check widget is disabled option, if disabled, no endpoint for omnichannel should work, and after a restart not even the widget, re-enabling enables the endpoints and the widget right away, without the need for a restart.


MartinSchoeler:
Moving this to draft, since the task context and objective is not that certain, let's take a step back and try to understand it better. Also this would be a BREAK since `this breaks custom integration when the setting is disabled, as well as custom apps.`

PR #28503, opened 2023-03-20
refactor(models): Use Messages Raw model (1/N)

sampaiodiego:
## Issue(s)

ARCH-929

### DIFF
--- a/apps/meteor/app/mentions/server/methods/getUserMentionsByChannel.js
+++ b/apps/meteor/app/mentions/server/methods/getUserMentionsByChannel.js
@@ -1,7 +1,8 @@
 import { Meteor } from 'meteor/meteor';
 import { check } from 'meteor/check';
+import { Messages } from '@rocket.chat/models';
 
-import { Rooms, Users, Messages } from '../../../models/server';
+import { Rooms, Users } from '../../../models/server';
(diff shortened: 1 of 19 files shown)


apps/meteor/app/mentions/server/methods/getUserMentionsByChannel.js
```diff
@@ -24,6 +25,6 @@ Meteor.methods({
 			});
 		}
 
-		return Messages.findVisibleByMentionAndRoomId(user.username, roomId, options).fetch();
+		return Messages.findVisibleByMentionAndRoomId(user.username, roomId, options).toArray();
```
KevLehman, line 28: In raw model, this uses findpaginated

KevLehman (changes requested):

apps/meteor/app/mentions/server/methods/getUserMentionsByChannel.js
sampaiodiego: well, that explains the failed test 😬

sampaiodiego: should be fixed now.. created both paginated/non-paginated variantes.

ggazzo (approved):

PR #28505, opened 2023-03-20
refactor(models): Use Messages Raw model (2/N)

sampaiodiego:
## Issue(s)

ARCH-929


apps/meteor/client/methods/unpinMessage.ts
```diff
@@ -26,7 +26,7 @@ Meteor.methods<ServerMethods>({
 			return 0;
 		}
 		dispatchToastMessage({ type: 'success', message: t('Message_has_been_unpinned') });
-		return ChatMessage.update(
+		ChatMessage.update(
```
KevLehman, line 29: needed?

apps/meteor/app/lib/server/functions/deleteMessage.ts
```diff
@@ -63,7 +64,8 @@ export async function deleteMessage(message: IMessage, user: IUser): Promise<voi
 	Rooms.decreaseMessageCountById(message.rid, 1);
 
 	if (showDeletedStatus) {
-		Messages.setAsDeletedByIdAndUser(message._id, user);
+		// TODO is there a better way to tell TS "IUser[username]" is not undefined?
```
KevLehman, line 67: NonNullable maybe, but's the same syntax lol

KevLehman: There's also the ! (non null assertion) but it's banned in our rules

apps/meteor/client/methods/unpinMessage.ts
sampaiodiego: yes.. this is client stub version that should reflect same as server version

ggazzo (approved):

PR #28508, opened 2023-03-20
chore: apps using raw models

thassiov:

Makes apps engine use raw models only, including its own models (apps, apps persistence and apps logs) and external ones, like subscriptions, room, users, etc.


tapiarafael:
Closing PR since the changes already have been done and it's already on the develop branch.

PR #28509, opened 2023-03-20
refactor: Remove `apps` & `app logs` models

KevLehman:
ARCH-916 ARCH-915

## Further comments

Needs: https://github.com/RocketChat/Rocket.Chat.Apps-engine/pull/603


thassiov (approved):
lgtm

apps/meteor/app/apps/server/bridges/api.ts
```diff
@@ -103,7 +103,8 @@ export class AppApisBridge extends ApiBridge {
 	}
 
 	private _appApiExecutor(endpoint: IApiEndpoint, appId: string): RequestHandler {
-		return (req: IRequestWithPrivateHash, res: Response): void => {
+		// eslint-disable-next-line @typescript-eslint/no-misused-promises
+		return async (req: IRequestWithPrivateHash, res: Response): Promise<void> => {
```
sampaiodiego, line 107: looks like there is already a `then/catch` for the `this.orch` operation, so it is not needed to convert this to `async` and disable eslint

apps/meteor/server/services/apps-engine/service.ts
```diff
@@ -31,10 +31,14 @@ export class AppsEngineService extends ServiceClassInternal implements IAppsEngi
...
 			if (!app) {
-				return;
+				/* empty */
 			}
 
-			await Apps.getManager()?.removeLocal(appId);
+			// Models removal made this more obvious: this event is getting duplicated (or triplicated)
+			// Commenting it while apps team provides a fix
+			// Apparently, since some things are now "promises", theres bigger change of this event being called while the app is being removed
+			// Triggering the whole remove process again, which fails (cause some data is not there anymore)
```
sampaiodiego, line 40: isn't it better to check the required data and just ignore if that is not there anymore? I mean, if this code is not executing at all it means the app will not be removed by any means right? which IMO is worse and not expected.

sampaiodiego (changes requested):
my previous comments are change requests 😬

apps/meteor/server/services/apps-engine/service.ts
KevLehman: Good question. When the event is received, the app appears to be in a "normal" state, with no indication of it being disabled. Apps engine internals still have it around and "available", but since the process has already started, there's _some_ data missing (slashCommands in this case).

So, would be cool to have a function on apps engine of the type `isAppValid` that checks all of this info, so in here we can use it.

What I don't understand is: why the event is being triggered when the app is still being deleted (on AppsManager.remove) and why .remove doesn't validate that the app has all the required data before attempting to delete 🤔

rodrigok: @KevLehman will you revert this?

KevLehman: All right, let's do something. Let's left this PR as "waiting", i'll revert the event change, and when yours is merged, I'll try this again to see if it goes well. Good? 👀

PR #28514, opened 2023-03-21
fix: Apps.engine resource consumption

tapiarafael:

Added some validations to avoid doing unnecessary updates and also avoid loop that causes high server usage

## Steps to test or reproduce
1. Install an app that has configuration options;
2. Change any settings;
3. Save the changes;

### DIFF
--- a/apps/meteor/server/services/apps-engine/service.ts
+++ b/apps/meteor/server/services/apps-engine/service.ts
@@ -25,6 +25,13 @@ export class AppsEngineService extends ServiceClassInternal implements IAppsEngi
 		});
 
 		this.onEvent('apps.added', async (appId: string): Promise<void> => {
+			// if the app already exists in this instance, don't load it again
+			const app = Apps.getManager()?.getOneById(appId);
+
+			if (app) {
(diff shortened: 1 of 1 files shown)


apps/meteor/server/services/apps-engine/service.ts
```diff
@@ -64,7 +71,16 @@ export class AppsEngineService extends ServiceClassInternal implements IAppsEngi
 			}
 		});
 
-		this.onEvent('apps.settingUpdated', async (appId: string, setting: ISetting): Promise<void> => {
+		this.onEvent('apps.settingUpdated', async (appId: string, setting: ISetting & { id: string }): Promise<void> => {
+			const app = Apps.getManager()?.getOneById(appId);
+			const oldSetting = app?.getStorageItem().settings[setting.id].value;
+
+			// avoid updating the setting if the value is the same,
+			// creating an infinite loop
```
d-gubert, line 79:
```suggestion
			// which caused an infinite loop
```

d-gubert (approved):

PR #28518, opened 2023-03-21
[NEW] App bridge for Moderation Features

Dnouv:

Expose a new bridge to allow for handling moderation features:
1. Reporting a message
2. Coming soon...


apps/meteor/app/apps/server/bridges/moderation.ts
```diff
@@ -0,0 +1,27 @@
...
+			throw new Error('Invalid message id');
+		}
+
+		if (!description) {
+			throw new Error('Invalid description');
+		}
+
+		await reportMessage(messageId, description, userId || 'rocket.cat');
+
+		// Meteor.call('reportMessage', { messageId, description });
```
tapiarafael, line 25: If this Meteor.call would not be used, it's better to remove it, isn't it?

tapiarafael (changes requested):

tapiarafael (approved):

PR #28531, opened 2023-03-21
refactor: move `Subscriptions` 1x

KevLehman:
ARCH-931


apps/meteor/server/models/raw/Subscriptions.ts
```diff
@@ -160,7 +160,7 @@ export class SubscriptionsRaw extends BaseRaw<ISubscription> implements ISubscri
 			'roles': roleId,
 		};
 
-		return this.findOne(query, { projection: { roles: 1 } });
+		return !!(await this.findOne(query, { projection: { roles: 1 } }));
```
sampaiodiego, line 163:
```suggestion
		return !!(await this.findOne(query, { projection: { _id: 1 } }));
```

apps/meteor/server/models/raw/Subscriptions.ts
```diff
@@ -706,13 +706,15 @@ export class SubscriptionsRaw extends BaseRaw<ISubscription> implements ISubscri
...
 
-	updateHideUnreadStatusById(_id: string, hideUnreadStatus: true): Promise<UpdateResult> {
+	updateHideUnreadStatusById(_id: string, hideUnreadStatus: boolean): Promise<UpdateResult> {
 		const query = {
 			_id,
 		};
 
 		const update: UpdateFilter<ISubscription> = {
 			$set: {
+				// @ts-expect-error - Check types of subscription
```
sampaiodiego, line 716: 👀

sampaiodiego (changes requested):

apps/meteor/server/models/raw/Subscriptions.ts
KevLehman: Yep. Basically:
ISubscription.hideUnreadStatus has "true" as type
But, when an endpoint calls updateHideUnreadStatusById, it passes a boolean 🤔 (x === '1')

So, not sure which of the types should prevail. IMHO, ISubscription should have boolean as type instead of true.

ggazzo: why not 

```
const update: UpdateFilter<ISubscription> = hideUnreadStatus
			? {
					$set: {
						hideUnreadStatus,
					},
			  }
			: {
					$unset: {
						hideUnreadStatus: 1,
					},
			  };
```

ggazzo (changes requested):

apps/meteor/server/models/raw/Subscriptions.ts
KevLehman: It changes what the function was doing originally, but works for the type. Since it doesn't change the logic, let me do it

PR #28536, opened 2023-03-21
fix: return file description when quoting

guijun13:
## Steps to test or reproduce

Steps to reproduce:

- Open a chat
- Send a file with a description
- Send a video message with a description
- Send an audio message with a description
- Quote the previous message
 
Expected Behavior:
It should include the description in the quote preview and in the quote

## Further comments

TC-484


yash-rajpal (changes requested):
Inside a QuoteAttachment, the other attachments are rendered using DefaultAttachment component, currently the DefaultAttachment doesn't render the `description` field, maybe we can render this description in Default attachment, that would might be better.

hugocostadev:
Closing in favor of: https://github.com/RocketChat/Rocket.Chat/pull/28579

PR #28584, opened 2023-03-23
fix: Hide System Messages for room requires refresh

gabriellsh:

Adittionaly fixed a bug in the Edit Channel contextual bar where the `hide system messages` option wasn't updating on removing the last item.

## Steps to test or reproduce

- Hide all needed system messages in room settings

- Try to perform actions like "add user" or "remove user"

- System messages are not hidden but become hidden after refreshing their own webpage, at the same time it is still not hidden for those users who have not yet refreshed their own webpage
## Further comments

TC-455


gabriellsh:
There is an issue when saving `hide system messages` when editing the channel through the room info contextual bar. When saving with the field empty, it doesn't save. This can be easily circumvented by unchecking the boolean flag on top of the select.

guijun13 (approved):

tassoevan (approved):

PR #28589, opened 2023-03-23
feat: Add setting to synchronize LDAP info on OAuth logins

matheusbsilva137:

 - Add a "Update User Data on Login with OAuth services" setting under LDAP Data Sync settings to enable users to have their data synced with LDAP servers on login with OAuth services.

## Steps to test or reproduce

Go to **Workspace > Settings > LDAP > Data Sync** and enable **"Update User Data on Login"** and **"Update User Data on Login with OAuth services"**. Then try to login using an OAuth service. The logged in user should be synced with its LDAP server corresponding registry (the username should match) on login

## Further comments

TC-433


totoi690 (dismissed):

apps/meteor/app/custom-oauth/server/custom_oauth_server.js
```diff
@@ -437,11 +439,15 @@ Accounts.updateOrCreateUserFromExternalService = function (...args /* serviceNam
 	const [serviceName, serviceData] = args;
 
 	const user = updateOrCreateUserFromExternalService.apply(this, args);
+	const fullUser = Users.findOneById(user.userId);
+	if (settings.get('LDAP_Update_Data_On_OAuth_Login')) {
+		LDAP.loginRequest(fullUser.username);
```
pierre-lehnen-rc, line 444: loginRequest is async

apps/meteor/server/services/ldap/service.ts
```diff
@@ -7,8 +7,8 @@ import { LDAPManager } from '../../lib/ldap/Manager';
 export class LDAPService extends ServiceClassInternal implements ILDAPService {
 	protected name = 'ldap';
 
-	async loginRequest(username: string, password: string): Promise<LDAPLoginResult> {
-		return LDAPManager.login(username, password);
+	async loginRequest(username: string, password?: string): Promise<LDAPLoginResult> {
+		return password ? LDAPManager.login(username, password) : LDAPManager.loginAuthenticatedUser(username);
```
pierre-lehnen-rc, line 11: instead of adding this logic check here, simply make `loginAuthenticatedUser` public and call it directly.

pierre-lehnen-rc (changes requested):

pierre-lehnen-rc (commented):
This new feature was not necessary to fix the mentioned issue, but it's a useful new feature regardless. 
Either way please make sure that the LDAP sync works together with OAuth accounts even if this setting is turned off.

.changeset/purple-tigers-taste.md
```diff
@@ -0,0 +1,6 @@
+---
+"@rocket.chat/meteor": patch
+"@rocket.chat/core-services": patch
```
sampaiodiego, line 3: `patch` should be used for fixes.. since this is a new feature, it should be `minor`..

sampaiodiego (approved):
the only issue is with changesets but it won't affect anything, so I'm approving

.changeset/purple-tigers-taste.md
matheusbsilva137: Fixed!

pierre-lehnen-rc (approved):

PR #28596, opened 2023-03-24
refactor: rooms model remove findOneByIdOrName to async

ggazzo:


apps/meteor/app/reactions/server/setReaction.ts
```diff
@@ -48,6 +47,10 @@ async function setReaction(room: IRoom, user: IUser, message: IMessage, reaction
 		});
 	}
 
+	// if (!('reactions' in message)) {
+	// 	return;
+	// }
+
```
KevLehman, line 53:
```suggestion
```

KevLehman (changes requested):

apps/meteor/app/reactions/server/setReaction.ts
ggazzo: this function is so stupid 😢

PR #28605, opened 2023-03-24
regression: Thread Preview breaking room

hugocostadev:

This regression was caused by this PR: https://github.com/RocketChat/Rocket.Chat/pull/28454 

Webhooks/integrations sometimes send an empty array to the messages object, because of that was necessary to double check if the array in not empty.


yash-rajpal (approved):

gabriellsh:
I think we can merge this, but just one comment:

If we knowingly accept for empty arrays to come in attachments, we should fix the `attachments` property in `IMessage` to reflect this. If we shouldn't, we must then not allow this to happen in the backend.

Let's treat this as a hotfix and investigate a little further

sampaiodiego:
@gabriellsh I understand your point, in fact we have discussed similar situations a few times, but in this case an empty array of an object is "**valid**"  and correctly typed.. as you can see https://www.typescriptlang.org/pt/play?ts=4.7.4#code/C4TwDgpgBAwg9gWzAGwgDwPICMBWUC8UA3gFBTlQB2AhghAFxQDOwATgJaUDmA3GRQDdqydgBNGWOHFTVKfAL4kSoSFACyskPCSo0BYv3JDkAVwhNG2lOmw4A2gF0FSgMZxKLZsDisG6zVa6+qQUUMZmFlCOJPJ8JEA

this `{ attachments: MessageAttachment[] }` doesn't mean `attachments` is not empty.. it means it is an array and every array value will be the type `MessageAttachment`

hugocostadev:
> @gabriellsh I understand your point, in fact we have discussed similar situations a few times, but in this case an empty array of an object is "**valid**" and correctly typed.. as you can see https://www.typescriptlang.org/pt/play?ts=4.7.4#code/C4TwDgpgBAwg9gWzAGwgDwPICMBWUC8UA3gFBTlQB2AhghAFxQDOwATgJaUDmA3GRQDdqydgBNGWOHFTVKfAL4kSoSFACyskPCSo0BYv3JDkAVwhNG2lOmw4A2gF0FSgMZxKLZsDisG6zVa6+qQUUMZmFlCOJPJ8JEA
> 
> this `{ attachments: MessageAttachment[] }` doesn't mean `attachments` is not empty.. it means it is an array and every array value will be the type `MessageAttachment`

But is it necessary to return an empty array instead of not returning an `attachment` object?

sampaiodiego:
I think it is from that file indeed..

and it is not *required* to save/return an empty array.. but my point was that an empty array is still a **valid typed value** ..

gabriellsh:
Oh yes, of course. I wonder why typescript wouldn't error since `message.attachments[0]` could be undefined... `isQuoteAttachment` only accepts `MessageAttachmentsBase` as parameter...

PR #27024, opened 2022-10-06
feat: apps to service

thassiov:
## Proposed changes (including videos or screenshots)

Make the apps-engine work as a service.

## Further comments

AECO-55
Depends on https://github.com/RocketChat/Rocket.Chat.Apps-engine/pull/587

PR #28612, opened 2023-03-24
refactor: Remove promise.await from apps

MarcosSpessatto:

Jira:
ARCH-382
ARCH-383
ARCH-384
ARCH-385


sampaiodiego (approved):
@RocketChat/apps can you please help confirm this is not creating a breaking change?

d-gubert:
> @RocketChat/apps can you please help confirm this is not creating a breaking change?

https://github.com/RocketChat/Rocket.Chat/pull/27024 was doing something similar, there's a couple things to note:

* https://github.com/RocketChat/Rocket.Chat/pull/27024/files#diff-20896fd9739a3566192ff645f3e4ef9e38bad33201de1a51f9eed80820a7facdR26 - this method needs to still run in a Fiber - changing that would break compatibility
* https://github.com/RocketChat/Rocket.Chat/pull/27024/files#diff-cec952ea73b6c415413ceede969dca56342823768fe967d09590d5637f10aff2 - same case for this one

ggazzo:
what does that mean? that this pr is useless or that we should keep it?
I don't think the PR you mentioned is in a healthy state enough to delivered (too many changes, too many conflicts).

If you can guarantee that this pr is good, providing approval or requesting changes, we can continue with the plan.

ggazzo:
Since we are following this idea of doing changes small as we can and keep everything working I cant see any problem on keeping the pointed changes in mind and apply them later
> Or do you think that we should simply keep it in the radar for the future?

PR #28613, opened 2023-03-24
fix: `UI_Use_Real_Name` not being respected in quotes and RoomForeword

gabriellsh:

Updated pertinent places and made it so user `name` is always sent with message (empty if non-existant).

Before:
image
image

After:
image
image

## Issue(s)

fix https://github.com/RocketChat/Rocket.Chat/issues/23051

## Further comments

TC-563


gabriellsh:
Unfortunately, for attachments (such as quotes), this setting only takes effect when sending the attachment. After that, if you change the setting, it'll show the value that was sent at that time (either name or username). This is a software design limitation related to how attachments work currently.

hugocostadev (commented):
Waiting conflicts to be solved

hugocostadev (approved):

d-gubert (approved):
All good on Apps-Engine side!

PR #28617, opened 2023-03-24
fix: room export by email

albuquerquefabio:

We improve the logic on the "/room.export" endpoint.

## Issue(s)

- /room.export 
  - When one tries to export to email, that is using the param `{"type": "email"}`, we get the error `"Cannot read property 'map' of undefined"`, even when "toUsers" and "toEmails" is provided.
  - Subject can not be empty and also must be validated before and after the submission.
 
## Further comments

This is PR https://github.com/RocketChat/Rocket.Chat/pull/24157 which I extracted this content. Why? The solution proposed here is much more simple than the original PR.


sampaiodiego (changes requested):
missing tests

PR #28647, opened 2023-03-27
refactor(Meteor.user): Remove all occurrences

rodrigok:

ARCH-180


MarcosSpessatto (dismissed):

apps/meteor/app/2fa/client/TOTPOAuth.js
```diff
@@ -105,7 +105,7 @@ Meteor.loginWithLinkedin = function (options, cb) {
 	overrideLoginMethod(loginWithLinkedin, [options], cb, loginWithLinkedinAndTOTP);
 };
 
-Accounts.onPageLoadLogin((loginAttempt) => {
+Accounts.onPageLoadLogin(async (loginAttempt) => {
```
KevLehman, line 108: Should this be a promise.await for now? Accounts.onPageLoadLogin doesn't seems to accept an async callback, so 2fa may not work 👀

ggazzo: you are right, but since process2faReturn is already async function I dont believe running this sync is the required

KevLehman (approved):

PR #28651, opened 2023-03-28
refactor: Remove promise.await 7/N

ggazzo:


apps/meteor/app/lib/server/functions/setStatusText.ts
```diff
@@ -36,14 +36,17 @@ async function _setStatusTextPromise(userId: string, statusText: string): Promis
...
-	return Promise.await(_setStatusTextPromise(userId, statusText));
-}
-
-export const setStatusText = RateLimiter.limitFunction(_setStatusText, 5, 60000, {
-	async 0() {
-		// Administrators have permission to change others status, so don't limit those
-		const userId = Meteor.userId();
-		return !userId || !(await hasPermissionAsync(userId, 'edit-other-user-info'));
+export const setStatusText = RateLimiter.limitFunction(
+	function _setStatusText(userId: any, statusText: string): boolean {
```
rodrigok, line 40: Did you forget to change this to async?

ggazzo: not exactly  the Promise.await removal is pretty obvious so somewhere/somehow someone is going to remove, I was only  "preparing the field" to do so. I didnt check every place where the function is being used, because I was focusing in different 'task/file' I can do this on other steps

Issue #23866, opened 2021-12-04
Header / Footer in plaintext emails


psy:
### Description:

While the header and footer for emails can be customized via settings and is added to the html part of mails it is missing in the plaintext part. If header and footer are only used to style emails, this is mostly irrelevant. But as they may contain relevant information (e.g. a link to an imprint), they should also be added to the plaintext part of mails.

### Steps to reproduce:

1. Look into source of an email sent by rocketchat
2. Find header / footer in html part
3. Search for it in plaintext part.

### Expected behavior:

Header and footer should also be added to the plaintext part.

### Actual behavior:

Plaintext part misses header and footer.

### Server Setup Information:

- Version of Rocket.Chat Server: 3.8.13
- Operating System: Debian

### Additional context

Example mail:
```
Subject: SMTP Test Email
Date: Sat, 04 Dec 2021 14:09:41 +0000
MIME-Version: 1.0

----_NmP-fe5a5cf915038dbf-Part_1
Content-Type: text/plain
Content-Transfer-Encoding: 7bit

You have successfully sent an email
----_NmP-fe5a5cf915038dbf-Part_1
Content-Type: text/html; charset=utf-8
Content-Transfer-Encoding: quoted-printable

<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Transitional//EN" "http://www.=
w3.org/TR/xhtml1/DTD/xhtml1-transitional.dtd"><html xmlns=3D"http://www.w3.=
org/1999/xhtml" style=3D"font-family: -apple-system,BlinkMacSystemFont,=
'Segoe UI',Roboto,Oxygen,Ubuntu,Cantarell,'Helvetica Neue','Apple Color =
Emoji','Segoe UI Emoji','Segoe UI Symbol','Meiryo UI',Arial,=
sans-serif;"><head><!-- If you delete this tag, the sky will fall on your =
head --><meta name=3D"viewport" content=3D"width=3Ddevice-width"><meta =
http-equiv=3D"Content-Type" content=3D"text/html; charset=3DUTF-8"><title>R=
ocket.Chat Cloud</title></head><body bgcolor=3D"#F7F8FA" =
style=3D"font-family: -apple-system,BlinkMacSystemFont,'Segoe UI',Roboto,=
Oxygen,Ubuntu,Cantarell,'Helvetica Neue','Apple Color Emoji','Segoe UI =
[...]

PR #28656, opened 2023-03-28
Fix header/footer for plainText

anmode:

  Follow up from this: https://github.com/RocketChat/Rocket.Chat/issues/23866#issuecomment-990377299

## Issue(s)
 #23866 

## Further comments
Here's one way to modify the wrap function to achieve this:
```
export const wrap = (html: string, data: { [key: string]: unknown } = {}): string => {
  if (settings.get('email_plain_text_only')) {
    return replace(html, data);
  }

  if (!body || !contentHeader || !contentFooter) {
    throw new Error('`body`, `contentHeader`, or `contentFooter` is not set yet');
  }

  const plaintext = replace(`{{body}}`, data);
  const wrappedPlaintext = `${contentHeader}\n${plaintext}\n${contentFooter}`;
  const wrappedHtml = replaceEscaped(body.replace('{{body}}', html), data);

  return { html: inlinecss(wrappedHtml), plaintext: wrappedPlaintext };
};
```
The modified function first checks whether the email is plaintext-only or HTML. If it's plaintext-only, the function only replaces variables in the plaintext version of the email and returns the result.

If the email includes HTML, the function checks whether` body`, `contentHeader`, and `contentFooter` have been set. These variables are set by the settings.watchMultiple function, which watches for changes to the email header and footer settings and updates the body variable accordingly.

If body, `contentHeader`, and `contentFooter` are all set, the function generates the plaintext version of the email by replacing variables in {{body}} with the provided data object. Then, it wraps the plaintext content with the header and footer and returns the result.

The function also generates the HTML version of the email by replacing `{{body}}` in the body variable with the provided HTML content, then escaping variables in the result. Finally, it returns an object with both the plaintext and HTML versions of the email.

I really like the idea discussed in https://github.com/RocketChat/Rocket.Chat/pull/23928#issue-1077846367

### DIFF
--- a/apps/meteor/app/mailer/server/api.ts
+++ b/apps/meteor/app/mailer/server/api.ts
@@ -70,11 +70,7 @@ export const replaceEscaped = (str: string, data: { [key: string]: unknown } = {
 };
 
 export const wrap = (html: string, data: { [key: string]: unknown } = {}): string => {
-	if (settings.get('email_plain_text_only')) {
-		return replace(html, data);
-	}
-
-	if (!body) {
+    if (!body) {
 		throw new Error('`body` is not set yet');
 	}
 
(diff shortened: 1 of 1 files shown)


harshitcompcode:
Hi there ,allow me to work on this issue

PR #28657, opened 2023-03-28
fix: reportmessage import error and archivereportsprop

Dnouv:

Fixes the validation params of `moderation.markChecked` and import error due to migration of Apps Engine code to EE.


debdutdeb:
Can we import ee code into community, have you talked about this with Doug? cc @d-gubert

Dnouv:
I saw the same import in the other file over here -> https://github.com/RocketChat/Rocket.Chat/blob/f10aaa6723b0e44ef331afa98da88647ca50e143/apps/meteor/app/lib/server/functions/addUserToRoom.ts#L7

So I think we could import it, or maybe not?

Dnouv:
Change already committed on the feature branch.

PR #28658, opened 2023-03-28
chore: improve tests for livechat/config api

murtaza98:
## Issue(s)

https://rocketchat.atlassian.net/browse/OC-848
## Further comments

this should help us detect errors like  https://github.com/RocketChat/Rocket.Chat/pull/28602


KevLehman (approved):

apps/meteor/tests/end-to-end/api/livechat/11-livechat.ts
```diff
@@ -102,6 +105,66 @@ describe('LIVECHAT - Utils', function () {
 
 			await deleteCustomField(customFieldName);
 		});
+		(IS_EE ? it : it.skip)('should return online as true if there is at least one agent online', async () => {
```
sampaiodiego, line 108: am I the only one who doesn't like this? why not using like this?

```suggestion
		it('should return online as true if there is at least one agent online', async () => {
			if (!IS_EE) {
				this.skip();
			}
```

murtaza98: https://www.reddit.com/r/ProgrammerHumor/comments/b6z1fc/ternaries_are_very_sophisticated/

murtaza98: PS: I usually go with this approach because it seems more compact, but if you want me to change it, let me know :)

sampaiodiego (approved):

PR #28666, opened 2023-03-28
refactor: AutoComplete Sanitization

dougfabris:


ggazzo (dismissed):

ggazzo (dismissed):

packages/core-typings/src/ISubscription.ts
```diff
@@ -14,6 +14,7 @@ export interface ISubscription extends IRocketChatRecord {
 
 	name: string;
 
+	avatarETag?: string;
```
tassoevan, line 17: Actually, I think this comes from `IRoom`.

tassoevan (changes requested):

tassoevan (approved):

PR #28669, opened 2023-03-28
refactor: Remove Users from fibers 3N

MarcosSpessatto:

### DIFF
--- a/apps/meteor/app/apps/server/bridges/messages.ts
+++ b/apps/meteor/app/apps/server/bridges/messages.ts
@@ -5,8 +5,9 @@ import type { IUser } from '@rocket.chat/apps-engine/definition/users';
 import type { IRoom } from '@rocket.chat/apps-engine/definition/rooms';
 import type { ISubscription } from '@rocket.chat/core-typings';
 import { api } from '@rocket.chat/core-services';
+import { Users } from '@rocket.chat/models';
 
-import { Messages, Users, Subscriptions } from '../../../models/server';
+import { Messages, Subscriptions } from '../../../models/server';
 import { updateMessage } from '../../../lib/server/functions/updateMessage';
 import { executeSendMessage } from '../../../lib/server/methods/sendMessage';
 import notifications from '../../../notifications/server/lib/Notifications';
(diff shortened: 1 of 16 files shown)


apps/meteor/app/apps/server/bridges/rooms.ts
```diff
@@ -190,7 +191,7 @@ export class AppRoomBridge extends RoomBridge {
 	private async getUsersByRoomIdAndSubscriptionRole(roomId: string, role: string): Promise<IUser[]> {
 		const subs = await Subscriptions.findByRoomIdAndRoles(roomId, [role], { projection: { uid: '$u._id', _id: 0 } });
 		// Was this a bug?
-		const users = await Users.findByIds(subs.map((user) => user.u._id));
+		const users = await Users.findByIds(subs.map((user: { uid: string }) => user.uid));
```
MarcosSpessatto, line 194: @KevLehman I intentionally reverted your change here, I believe it was indeed correct if you look at the projection at the above line. @RocketChat/apps can you please confirm that?

apps/meteor/app/apps/server/bridges/messages.ts
```diff
@@ -78,14 +83,12 @@ export class AppMessageBridge extends MessageBridge {
...
 
-		Users.findByIds(users, { fields: { _id: 1 } })
-			.fetch()
-			.forEach(
-				({ _id }: { _id: string }) =>
-					void api.broadcast('notify.ephemeralMessage', _id, room.id, {
-						...msg,
-					}),
-			);
+		(await Users.findByIds(users, { fields: { _id: 1 } }).toArray()).forEach(
```
ggazzo, line 86: I think here we can use the foreach from the cursor, no need to use toArray
```suggestion
		await Users.findByIds(users, { fields: { _id: 1 } }).forEach(
```

ggazzo (dismissed):

apps/meteor/app/bot-helpers/server/index.ts
```diff
@@ -111,63 +110,75 @@ class BotHelpers {
...
-			return false;
-		}
-		return this._allUsers.fetch().map((user) => user.username);
+		return (async () => {
+			if (!this.userFields.hasOwnProperty('username')) {
+				this.requestError();
+				return false;
+			}
+			return (await this._allUsers.toArray()).map((user) => user.username);
+		})();
```
ggazzo, line 131: I dont see the point of doing this, could you explain to me?

PR #28678, opened 2023-03-29
refactor(models): Use Messages Raw model (8/N)

sampaiodiego:


apps/meteor/app/livechat/server/lib/LivechatTyped.ts
```diff
@@ -325,9 +332,16 @@ class LivechatClass {
...
-		});
+		await Messages.createWithTypeRoomIdMessageUserAndUnread(
+			'livechat_transcript_history',
+			room._id,
+			'',
+			user as IRegisterUser,
+			settings.get('Message_Read_Receipt_Enabled'),
+			{
+				requestData: { type, visitor, user },
+			} as any, // TODO should a message really have "requestData"? if so, it should be on IMessage
```
KevLehman, line 343: Should be. AFAIK, it's on IOmnichannelSystemMessage, but not on the base type

PR #28684, opened 2023-03-29
refactor: Rewrite Users raw model to TS

gabriellsh:


apps/meteor/server/models/raw/Users.ts
```diff
@@ -0,0 +1,1231 @@
...
+		};
+
+		const update = {
+			$set: {
+				statusLivechat: status,
+				...extraFields,
+			},
+		};
+
+		return this.update(query, update);
```
MarcosSpessatto, line 438: I believe it would be better to use `.updateOne`  for this and all the subsequent applicable cases. Or `updateMany` if that's the case instead of  `update` (it's deprecated)

MarcosSpessatto (changes requested):
Also, please take a look at the work @KevLehman did on the Users model already, just to avoid too lose his work here. https://github.com/RocketChat/Rocket.Chat/pull/28456/files#diff-6b3dd01a9425ad473d8713c6d4f6de2127be8f2b15fd8a0ee4a3c0f5ccc87026

PR #28687, opened 2023-03-29
refactor: Remove Users from fibers 5

MarcosSpessatto:


apps/meteor/app/lib/server/functions/addUserToRoom.ts
```diff
@@ -85,16 +85,16 @@ export const addUserToRoom = async function (
 				},
 			};
 			if (room.teamMain) {
-				await Message.saveSystemMessage('added-user-to-team', rid, userToBeAdded.username, userToBeAdded, extraData);
+				await Message.saveSystemMessage('added-user-to-team', rid, userToBeAdded.username || '', userToBeAdded, extraData);
```
ggazzo, line 88: throw an error if username is empty. saving a message with an invalid user is not a valid planned case

ggazzo (changes requested):

apps/meteor/app/integrations/server/api/api.js
```diff
@@ -51,7 +52,7 @@ function buildSandbox(store = {}) {
 		},
 	};
 	Object.keys(Models)
-		.filter((k) => !k.startsWith('_'))
+		.filter((k) => !forbiddenModelMethods.includes(k))
```
KevLehman, line 55: Won't this allow accessing __proto__ stuff and so on?

apps/meteor/app/integrations/server/methods/outgoing/updateOutgoingIntegration.ts
```diff
@@ -96,7 +95,7 @@ Meteor.methods<ServerMethods>({
 					triggerWordAnywhere: integration.triggerWordAnywhere,
 					runOnEdits: integration.runOnEdits,
 					_updatedAt: new Date(),
-					_updatedBy: Users.findOne(this.userId, { fields: { username: 1 } }),
+					_updatedBy: await Users.findOne(this.userId, { projection: { username: 1 } }),
```
KevLehman, line 98: I thougt findOne received an object 🤔 it's customized on the model?

apps/meteor/app/lib/server/functions/createRoom.ts
```diff
@@ -49,16 +48,16 @@ export const createRoom = async <T extends RoomType>(
...
 
 	if (!ownerUsername || !owner) {
 		throw new Meteor.Error('error-invalid-user', 'Invalid user', {
 			function: 'RocketChat.createRoom',
 		});
 	}
 
-	if (!members.includes(owner)) {
-		members.push(owner.username);
+	if (owner.username && !(members as string[]).includes(owner.username)) {
```
KevLehman, line 59: I'd suggest to use a type validator instead of coercion

KevLehman (changes requested):

apps/meteor/app/integrations/server/methods/outgoing/updateOutgoingIntegration.ts
MarcosSpessatto: No, you are right, there's a lot of confusion with those methods because the previous version (sync) was allowing queries like this one, receiving just the id on findOne, update, etc. Will change, tks.

apps/meteor/app/integrations/server/api/api.js
MarcosSpessatto: Not really, this is the way the previous code was doing. I know, I agree that we should not expose all the models inside the integration script, but the old version was doing this. @rodrigok @ggazzo @sampaiodiego any thoughts on this?

PR #28697, opened 2023-03-30
refactor: Bump TypeScript version

tassoevan:


apps/meteor/app/api/server/v1/assets.ts
```diff
@@ -35,7 +35,7 @@ API.v1.addRoute(
 				await Meteor.callAsync('refreshClients');
 			}
 
-			return API.v1.success();
+			return API.v1.success<void>();
```
KevLehman, line 38: Wouldn't be possible to add T = void to the success type? 👀

PR #28702, opened 2023-03-30
refactor: Remove Users from fibers 7

MarcosSpessatto:


apps/meteor/app/slackbridge/server/SlackAdapter.js
```diff
@@ -1123,13 +1127,13 @@ export default class SlackAdapter {
...
-	uploadFileFromSlack(details, slackFileURL, rocketUser, rocketChannel, timeStamp, isImporting) {
+	async uploadFileFromSlack(details, slackFileURL, rocketUser, rocketChannel, timeStamp, isImporting) {
 		const requestModule = /https/i.test(slackFileURL) ? https : http;
 		const parsedUrl = url.parse(slackFileURL, true);
 		parsedUrl.headers = { Authorization: `Bearer ${this.apiToken}` };
-		requestModule.get(
+		await requestModule.get(
 			parsedUrl,
-			Meteor.bindEnvironment((stream) => {
+			await Meteor.bindEnvironment(async (stream) => {
```
sampaiodiego, line 1136: not sure this has the expected behavior..

`requestModule.get` 's second param is a callback.. looks like you're adding `await` because of you removed the `Promise.await` from `sendMessage`.. but the `await` you added will not `await sendMessage`, it will just `await Meteor.bindEnvironment`.. is that the case?

I guess you don't need this `await` and can do `void sendMessage`, since it is being called on `fileStore.insert`'s callback

sampaiodiego (changes requested):

apps/meteor/app/slackbridge/server/SlackAdapter.js
MarcosSpessatto: That's correct, thanks for the heads up, will remove that.

apps/meteor/server/methods/setAvatarFromService.ts
```diff
@@ -12,7 +12,7 @@ import { hasPermissionAsync } from '../../app/authorization/server/functions/has
 declare module '@rocket.chat/ui-contexts' {
 	// eslint-disable-next-line @typescript-eslint/naming-convention
 	interface ServerMethods {
-		setAvatarFromService(dataURI: Buffer | Blob, contentType?: string, service?: string, userId?: string): void;
+		setAvatarFromService(dataURI: Buffer | Blob, contentType?: string, service?: string, userId?: string): Promise<void>;
```
ggazzo, line 15:
```suggestion
		setAvatarFromService(dataURI: Buffer | Blob, contentType?: string, service?: string, userId?: string): void;
```

PR #28710, opened 2023-03-30
refactor: Move callbacks to ts 1X

KevLehman:


apps/meteor/app/livechat/server/hooks/sendToCRM.ts
```diff
@@ -38,55 +76,48 @@ const getAdditionalFieldsByType = (type, room) => {
...
+			messages: [],
+		};
 
-	postData.messages = [];
-
-	let messages;
+	let messages: IOmnichannelSystemMessage[] | null = null;
 	if (typeof includeMessages === 'boolean' && includeMessages) {
-		messages = await Messages.findVisibleByRoomId(room._id, { sort: { ts: 1 } }).toArray();
+		messages = (await Messages.findVisibleByRoomId(room._id, { sort: { ts: 1 } }).toArray()) as IOmnichannelSystemMessage[];
```
sampaiodiego, line 98: do you really need to cast is as `IOmnichannelSystemMessage` ? looks like you're properly testing it later, using like `isOmnichannelNavigationMessage`..

if really required, can you please improve the model implementation to support a generic so you could do like this:
```suggestion
		messages = await Messages.findVisibleByRoomId<IOmnichannelSystemMessage>(room._id, { sort: { ts: 1 } }).toArray();
```

apps/meteor/app/livechat/server/hooks/sendToCRM.ts
```diff
@@ -38,55 +76,48 @@ const getAdditionalFieldsByType = (type, room) => {
...
+	includeMessages: boolean | IOmnichannelSystemMessage[] = true,
+): Promise<OmnichannelRoomWithExtraFields> {
 	if (!settings.get('Livechat_webhookUrl')) {
 		return room;
 	}
 
-	const postData = await Livechat.getLivechatRoomGuestInfo(room);
-
-	postData.type = type;
+	const postData: Awaited<ReturnType<typeof Livechat.getLivechatRoomGuestInfo>> & { type: string; messages: IOmnichannelSystemMessage[] } =
```
sampaiodiego, line 89: why such TS witchcraft? 🙈 I'm sorry I'm dumb

sampaiodiego (changes requested):

apps/meteor/app/livechat/server/hooks/sendToCRM.ts
KevLehman: :kek:

It's saying: "Give me the same return type of `livechat.getlivechatroomGuestInfo` after being awaited, and add { type, messages } props to it (as they don't come from the function call)"

sampaiodiego: once we have it typed this will be removed then, right?
image

KevLehman: image

PR #28715, opened 2023-03-30
refactor: `Rooms` model 2X

KevLehman:


apps/meteor/app/channel-settings/server/methods/saveRoomSettings.ts
```diff
@@ -424,7 +423,7 @@ async function saveRoomSettings(
 		});
 	}
 
-	const room = Rooms.findOneById(rid) as IRoom | undefined;
+	const room = (await Rooms.findOneById(rid)) as IRoom | undefined;
```
sampaiodiego, line 426:
```suggestion
	const room = await Rooms.findOneById(rid);
```

apps/meteor/app/discussion/server/hooks/propagateDiscussionMetadata.ts
```diff
@@ -62,7 +61,7 @@ callbacks.add(
 callbacks.add(
 	'afterDeleteRoom',
 	async (rid) => {
-		RoomsSync.find({ prid: rid }, { fields: { _id: 1 } }).forEach(({ _id }: Pick<IRoom, '_id'>) => deleteRoom(_id));
+		await Rooms.find({ prid: rid }, { projection: { _id: 1 } }).forEach(({ _id }: Pick<IRoom, '_id'>) => deleteRoom(_id));
```
sampaiodiego, line 64: using `await ... forEach()` will not await for all async operations on the `forEach` callbacks..

since this is a heavy operation, I'd change it to a `for await` in this case.. wdyt?

sampaiodiego (changes requested):

apps/meteor/app/discussion/server/hooks/propagateDiscussionMetadata.ts
KevLehman: Ah, that's because deleteRoom is not async yet 👀 was planning to remove it with the Promise.awaits. Can change this to be for/await/of tho

sampaiodiego (approved):

PR #28720, opened 2023-03-31
chore: convert Livechat.js to ts (1/N)

murtaza98:
## Issue(s)

https://rocketchat.atlassian.net/browse/OC-856


apps/meteor/app/livechat/server/api/v1/visitor.ts
```diff
@@ -29,14 +30,13 @@ API.v1.addRoute('livechat/visitor', {
 		});
 
 		const { token, customFields } = this.bodyParams.visitor;
-		const guest: ILivechatVisitorDTO = { ...this.bodyParams.visitor };
+		const guest: Parameters<typeof LivechatTyped.registerGuest>[0] = {
```
KevLehman, line 33: Is this needed?

apps/meteor/server/models/raw/Users.js
```diff
@@ -1611,6 +1611,7 @@ export class UsersRaw extends BaseRaw {
 
 		const options = {
 			projection: {
+				_id: 1,
```
KevLehman, line 1614: Isn't _id always there?

KevLehman (dismissed):

apps/meteor/app/livechat/server/api/v1/visitor.ts
murtaza98: Apparently its not. Not sure Y added it before :thinking:

apps/meteor/server/models/raw/Users.js
murtaza98: Yups, lets remove it

apps/meteor/app/livechat/server/api/v1/visitor.ts
```diff
@@ -29,14 +30,13 @@ API.v1.addRoute('livechat/visitor', {
 		});
 
 		const { token, customFields } = this.bodyParams.visitor;
-		const guest: ILivechatVisitorDTO = { ...this.bodyParams.visitor };
+		const guest = {
+			...this.bodyParams.visitor,
+			phone: this.bodyParams.visitor.phone ? { number: this.bodyParams.visitor.phone as string } : undefined,
```
KevLehman, line 35: Why?

If `phone` accept undefined, why is the ternary needed?

KevLehman: worst case, something like: `typeof xxx.phone === 'string' ? xxx.phone : undefined` to avoid the unsafe cast

KevLehman (changes requested):

KevLehman (approved):

ggazzo (dismissed):

murtaza98:
Hi @KevLehman @ggazzo Could you please re-approve this PR? Lost your previous approval while resolving conflicts :grimacing: Thanks :v:

KevLehman (dismissed):

sampaiodiego (approved):

apps/meteor/app/livechat/server/lib/QueueManager.js
```diff
@@ -73,7 +73,13 @@ export const QueueManager = {
 		await queueInquiry(room, inquiry, agent);
 		logger.debug(`Inquiry ${inquiry._id} queued`);
 
-		return LivechatRooms.findOneById(rid);
+		const newRoom = await LivechatRooms.findOneById(rid);
+		if (!newRoom) {
+			logger.debug(`Room with id ${rid} not found`);
```
sampaiodiego, line 78: should be a `logger.error` since it is being `throw` ?

apps/meteor/app/livechat/server/lib/QueueManager.js
```diff
@@ -73,7 +73,13 @@ export const QueueManager = {
...
+			throw new Error('room-not-found');
```
sampaiodiego, line 79: what happens to this error thrown? is it caught somewhere?

apps/meteor/app/livechat/server/api/v1/room.ts
```diff
@@ -59,14 +59,14 @@ API.v1.addRoute('livechat/room', {
...
 			}
 
-			let agent;
+			let agent: SelectedAgent | undefined;
 			const agentObj = agentId && (await findAgent(agentId));
 			if (agentObj) {
 				if (isAgentWithInfo(agentObj)) {
 					const { username = undefined } = agentObj;
-					agent = { agentId, username };
+					agent = { agentId, username } as SelectedAgent;
```
sampaiodiego, line 67: this type cast looks wrong.. `username` cannot be `undefined` in `SelectedAgent`..

apps/meteor/app/livechat/server/api/v1/room.ts
```diff
@@ -59,14 +59,14 @@ API.v1.addRoute('livechat/room', {
...
 				} else {
-					agent = { agentId };
+					agent = { agentId } as SelectedAgent;
```
sampaiodiego, line 69: this looks wrong as well, since `username` is mandatory on `SelectedAgent`.. maybe `SelectedAgent` needs to be modified to allow both use cases on this file, and then the type cast can be removed.

sampaiodiego (changes requested):
replaying comments from the other PR

apps/meteor/app/livechat/server/lib/QueueManager.js
murtaza98: Indeed it should be an error

apps/meteor/app/livechat/server/lib/QueueManager.js
murtaza98: Well we're not explicitly catching this error, rather it's being handled as a generic error. Which means the calling service/lib will be responsible for handling this error. So if it's an API call, then this error would be returned within the error message of the api response. On the other hand, if this is an App's engine call, then this error would be thrown to app and they'd be responsible for handling it appropriately based on their use-case.
Additionally I don't really see this error being thrown since this room was just created by the system just a few lines before this code.

apps/meteor/app/livechat/server/api/v1/room.ts
murtaza98: Username seems to be an redundant prop in this type, so it makes sense to make it optional

apps/meteor/app/livechat/server/api/v1/room.ts
murtaza98: No more typecasting is needed here

KevLehman (approved):

sampaiodiego (approved):

PR #28738, opened 2023-03-31
fix: marketplace filters are changing size

felipe-rod123:

Created a aditional prop `fixFiltersSize` on the file `apps/meteor/client/views/marketplace/AppsPage/AppsFilters.tsx`, to adapt the filters size.

Previous version:

New version:

## Further comments

AECO-51


rique223 (approved):

apps/meteor/client/views/marketplace/AppsPage/AppsFilters.tsx
```diff
@@ -59,6 +59,8 @@ const AppsFilters = ({
 		private: t('Search_Private_apps'),
 	};
 
+	const fixFiltersSize = window.innerWidth > 1060 ? { maxWidth: 'x200', minWidth: 'x200' } : null;
```
tassoevan, line 62: This not going to work well if the viewport resizes (e.g. changing window size, opening side panels like DevTools). Prefer to use `useMediaQuery` hook.

tassoevan (changes requested):

apps/meteor/client/views/marketplace/AppsPage/AppsFilters.tsx
felipe-rod123: Thanks for the feedback! :)

tiagoevanp (approved):

apps/meteor/client/views/marketplace/AppsPage/AppsFilters.tsx
```diff
@@ -59,6 +59,8 @@ const AppsFilters = ({
 		private: t('Search_Private_apps'),
 	};
```
ggazzo, line 61: I think  useBreakpoint is better so we dont need to use "magical" numbers

ggazzo (changes requested):

Issue #28744, opened 2023-04-01
OAuthApplication Edit Page fails to load the OAuthApp setting.


abhinavkrin:
### Description:
When trying to open an OAuthApp setting, the page shows a loading status and eventually fails to load the OAuthApp setting page.

### Steps to reproduce:

1. Go to workspace -> OAuthApps
2. Click on the desired OAuthApp to open
3. Now the next page will fail to load the settings page for the chosen OAuthApp.

### Expected behavior:
The OAuthApp should be loaded and the UI to configure the OAuthApp should be displayed.

### Actual behavior:
It shows the loading skeleton and then after some time an error is shown that 
```
must have required property 'clientId'
 must have required property 'appId'
 must NOT have additional properties
 must match exactly one schema in oneOf [invalid-params]`
```
### Server Setup Information:

- Version of Rocket.Chat Server: 6.2.0-develop
- Operating System: Debian 11
- Deployment Method: yarn dsv
- Number of Running Instances: 1
- DB Replicaset Oplog: 
- NodeJS Version: v14.21.3
- MongoDB Version: 6.0.3 / wiredTiger (oplog Enabled)

### Client Setup Information

- Desktop App or Browser Version: Chrome Version 111.0.5563.64 (Official Build) (64-bit)
- Operating System: Debian 11

### Relevant logs:
Browser logs:-
![Screenshot from 2023-04-01 14-20-21](https://user-images.githubusercontent.com/15830206/229276071-03f61479-effc-4f65-88c6-b69984d84496.png)

### Screenshots
![ajv-validation-error](https://user-images.githubusercontent.com/15830206/229276087-999a7922-99e0-4f7f-9f59-79680dfe4a40.png)

tassoevan:
Closed by #26810.

PR #26810, opened 2022-09-05
chore: create EditOauthApp endpoints

felipe-rod123:
## Proposed changes (including videos or screenshots)
Created 'oauth-apps.updateOAuthApp' and other endpoints for the `apps/meteor/client/views/admin/oauthApps/EditOauthApp.tsx` file, and added Ajv validations.

## Steps to test or reproduce
 1. Enter on Administration settings
 2. Open the OAuth section
 3. Create new OAuth Provider
 4. Try to edit the create OAuth Provider

## Further comments
TC-369

PR #28745, opened 2023-04-01
fix: oauthapps settings page crash

abhinavkrin:
The oauth app setting page was crashing and could not load the setting page. It showed an error 
```
must have required property 'clientId'
 must have required property 'appId'
 must NOT have additional properties
 must match exactly one schema in oneOf [invalid-params]
```
The issue was cause by the ajv schema validation for `oauth-apps.get` end point due to a typo
schema
In line 35 it should be _id instead of appId.

## Issue(s)

#28744 
ajv-validation-error

## Steps to test or reproduce

1. Go to workspace -> OAuthApps
2. Click on the desired OAuthApp to open
3. Now the next page will fail to load the settings page for the chosen OAuthApp.


tassoevan (approved):

tassoevan:
Addressed and expanded by #26810. Regardless, thank you very much!

PR #28767, opened 2023-04-03
refactor: remove Users from fibers 12

MarcosSpessatto:


apps/meteor/client/lib/utils/timeAgo.ts
```diff
@@ -9,8 +9,8 @@ import { getUserPreference } from '../../../app/utils/lib/getUserPreference';
 
 const dayFormat = ['h:mm A', 'H:mm'];
 
-export const timeAgo = (date: MomentInput): string => {
-	const clockMode = Tracker.nonreactive(() => getUserPreference(Meteor.userId(), 'clockMode', false) as number | boolean);
+export const timeAgo = async (date: MomentInput): Promise<string> => {
+	const clockMode = await Tracker.nonreactive(async () => getUserPreference(Meteor.userId(), 'clockMode', false) as number | boolean);
```
MarcosSpessatto, line 13: @ggazzo I need some help here, I'm not sure if this makes sense and how to handle the `timeAgo` where it's being called.

sampaiodiego: this will not work since `getUserPreference` is now importing from `@rocket.chat/models`, which doesn't exist on client..

sampaiodiego: I think you should split `getUserPreference` into two functions: one for client using specific imports of client files, and another one to be used on server which uses `@rocket.chat/models` ..

sampaiodiego (changes requested):

apps/meteor/client/lib/utils/timeAgo.ts
MarcosSpessatto: yes, I saw that, my bad. I didn't realize that we already have a client-side function. I already addressed that. 🤙

ggazzo (approved):

PR #28783, opened 2023-04-03
feat: Add custom OAuth setting to allow merging users to others from distinct services

matheusbsilva137:

 - Add "Merge users from distinct services" to Custom OAuth services (new and existing ones) in order to allow merging OAuth users to its matching counterparts from distinct services;
 - Fix OAuth users being merged to users from distinct servers (services' `id`s were incorrectly extracted).

## Further comments

[TC-433]

[TC-433]: https://rocketchat.atlassian.net/browse/TC-433?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


pierre-lehnen-rc (dismissed):

apps/meteor/server/startup/migrations/v298.ts
```diff
@@ -0,0 +1,25 @@
+import { settings, settingsRegistry } from '../../../app/settings/server';
+import { addMigration } from '../../lib/migrations';
+
+addMigration({
+	version: 293,
```
sampaiodiego, line 5: version number is wrong, not matching file name

apps/meteor/server/startup/migrations/v298.ts
```diff
@@ -0,0 +1,25 @@
...
+	async up() {
+		const customOauthServices = settings.getByRegexp(/Accounts_OAuth_Custom-[^-]+$/im);
```
sampaiodiego, line 7: please a model and do a query instead.. it is not safe to use the settings cache during migrations

.changeset/dirty-singers-promise.md
```diff
@@ -0,0 +1,6 @@
+---
+"@rocket.chat/meteor": patch
+"@rocket.chat/rest-typings": patch
```
sampaiodiego, line 3: `patch` are meant for fixes.. since this is a new feature, it should be a `minor`

sampaiodiego (changes requested):

apps/meteor/server/startup/migrations/v299.ts
```diff
@@ -0,0 +1,27 @@
...
+import { addMigration } from '../../lib/migrations';
+
+addMigration({
+	version: 299,
+	async up() {
+		const customOauthServices = await Settings.find({ _id: /Accounts_OAuth_Custom-[^-]+$/im }).toArray();
+		const serviceNames = customOauthServices.map(({ _id }) => _id.replace('Accounts_OAuth_Custom-', ''));
+
+		serviceNames.forEach(async (serviceName) => {
+			await settingsRegistry.add(`Accounts_OAuth_Custom-${serviceName}-merge_users_distinct_services`, '', {
```
sampaiodiego, line 13: the default value is `false` right? I didn't check every other field
```suggestion
			await settingsRegistry.add(`Accounts_OAuth_Custom-${serviceName}-merge_users_distinct_services`, false, {
```

apps/meteor/server/startup/migrations/v299.ts
```diff
@@ -0,0 +1,27 @@
+import { Settings } from '@rocket.chat/models';
+
+import { settingsRegistry } from '../../../app/settings/server';
+import { addMigration } from '../../lib/migrations';
+
+addMigration({
+	version: 299,
+	async up() {
+		const customOauthServices = await Settings.find({ _id: /Accounts_OAuth_Custom-[^-]+$/im }).toArray();
```
sampaiodiego, line 9: the previous one was working, but since there are other change requests, I thought it was good to add this one as well
```suggestion
		const customOauthServices = await Settings.find({ _id: /Accounts_OAuth_Custom-[^-]+$/ }, { projection: { _id: 1 } }).toArray();
```

apps/meteor/server/startup/migrations/v299.ts
```diff
@@ -0,0 +1,27 @@
...
+		const serviceNames = customOauthServices.map(({ _id }) => _id.replace('Accounts_OAuth_Custom-', ''));
+
+		serviceNames.forEach(async (serviceName) => {
```
sampaiodiego, line 12: `Array.forEach` is not `async`.. you can either use `await Promise.all( ... )` with `serviceNames.map` or use `for await`:
```suggestion
		for await (const serviceName of serviceNames) {
```

sampaiodiego (changes requested):

apps/meteor/server/startup/migrations/v299.ts
matheusbsilva137: Oops, fixed! I rechecked the other ones and they're ok. Tkss

tapiarafael (approved):

sampaiodiego (approved):

Issue #28130, opened 2023-02-23
[6.0.0] crash with "TypeError: Cannot read property 'charAt' of undefined"


himpierre:
### Description:

Hello Devs.

After updating to 6.0.0-rc.1 from 5.4.3 (docker) the App crashes with:

```
=== UnHandledPromiseRejection ===
TypeError: Cannot read property 'charAt' of undefined
    at ee/app/license/server/lib/isUnderAppLimits.ts:11:33
    at /app/bundle/programs/server/npm/node_modules/meteor/promise/node_modules/meteor-promise/fiber_pool.js:43:40
 => awaited here:
    at Function.Promise.await (/app/bundle/programs/server/npm/node_modules/meteor/promise/node_modules/meteor-promise/promise_server.js:56:12)
    at ee/server/apps/orchestrator.js:169:48
    at /app/bundle/programs/server/npm/node_modules/meteor/promise/node_modules/meteor-promise/fiber_pool.js:43:40
 => awaited here:
    at Function.Promise.await (/app/bundle/programs/server/npm/node_modules/meteor/promise/node_modules/meteor-promise/promise_server.js:56:12)
    at ee/server/apps/orchestrator.js:165:3
    at /app/bundle/programs/server/npm/node_modules/meteor/promise/node_modules/meteor-promise/fiber_pool.js:43:40
---------------------------------
Errors like this can cause oplog processing errors.
Setting EXIT_UNHANDLEDPROMISEREJECTION will cause the process to exit allowing your service to automatically restart the process
Future node.js versions will automatically exit the process
```

Had to roll back to 5.4.3.

### Steps to reproduce:

1. Uprade to 6.0.0-rc.1 from 5.4.3
2. Docker Container chrashes

### Expected behavior:

Docker shoud run with 6.0.0-rc.1.

### Actual behavior:

Crash.

### Server Setup Information:

- Version of Rocket.Chat Server: 6.0.0-rc.1
- Operating System: Linux
- Deployment Method: docker
- Number of Running Instances: 1
- DB Replicaset Oplog: yes
- NodeJS Version: 
- MongoDB Version: 4.4

### Client Setup Information

- Desktop App or Browser Version: irrelevant
- Operating System: irrelevant

Any hints?

cheers, t.

dudanogueira:
Hi, Thanks for your issue report!

This is on your staging environment Or is this on a fresh new install?

himpierre:
[...]

PR #28806, opened 2023-04-05
fix: Add function to get installation source

d-gubert:

There have been reports of problems regarding the missing information of `installationSource` in apps. The only scenario where this information could be missing is if migrations failed to run properly when upgrading to 6.0. We haven't been able to identify the root cause yet, so we decided to provide a workaround in the meanwhile

## Issue(s)

fix https://github.com/RocketChat/Rocket.Chat/issues/28130


tapiarafael (approved):

AllanPazRibeiro (dismissed):
LGTM

AlenDavid (dismissed):

tapiarafael (approved):

PR #28822, opened 2023-04-05
refactor: remove some Meteor.bindEnvironment

MarcosSpessatto:

ARCH-948


apps/meteor/app/livechat/server/lib/Livechat.js
```diff
@@ -777,28 +777,26 @@ export const Livechat = {
...
 			attempts - 1 && Livechat.webhookLogger.warn('Will try again in 10 seconds ...');
-			setTimeout(
-				Meteor.bindEnvironment(function () {
-					Livechat.sendRequest(postData, callback, attempts - 1);
-				}),
-				10000,
-			);
+			setTimeout(function () {
+				Livechat.sendRequest(postData, callback, attempts - 1);
+			}, 10000);
```
sampaiodiego, line 799:
```suggestion
			setTimeout(async function () {
				await Livechat.sendRequest(postData, callback, attempts - 1);
			}, 10000);
```

apps/meteor/app/version-check/server/index.ts
```diff
@@ -12,15 +12,15 @@ if (SyncedCron.nextScheduledAtDate(jobName)) {
 	SyncedCron.remove(jobName);
 }
 
-const addVersionCheckJob = Meteor.bindEnvironment(() => {
+const addVersionCheckJob = () => {
```
sampaiodiego, line 15: I was expecting synced cron to require bindEnvironment 😳

sampaiodiego (changes requested):

apps/meteor/app/version-check/server/index.ts
MarcosSpessatto: That's correct, I'll add it back.

apps/meteor/app/custom-sounds/server/startup/custom-sounds.js
```diff
@@ -46,6 +46,9 @@ Meteor.startup(function () {
 
 		const file = await RocketChatFileCustomSoundsInstance.getFileWithReadStream(fileId);
 		if (!file) {
+			res.writeHead(404);
```
MarcosSpessatto, line 49: @sampaiodiego I've changed this to a 404 error because it was doing nothing here, leading to a timeout error. I don't think this should be considered a breaking change.

sampaiodiego: that's great.. we desperately need tests these endpoints =)

sampaiodiego (approved):

PR #28825, opened 2023-04-06
Remove saml wrapAsync

rodrigok:

### DIFF
--- a/apps/meteor/app/meteor-accounts-saml/server/lib/SAML.ts
+++ b/apps/meteor/app/meteor-accounts-saml/server/lib/SAML.ts
@@ -374,22 +374,28 @@ export class SAML {
 		res.end();
 	}
 
-	private static processAuthorizeAction(res: ServerResponse, service: IServiceProviderOptions, samlObject: ISAMLAction): void {
+	private static async processAuthorizeAction(
+		res: ServerResponse,
+		service: IServiceProviderOptions,
+		samlObject: ISAMLAction,
+	): Promise<void> {
 		service.id = samlObject.credentialToken;
 
 		const serviceProvider = new SAMLServiceProvider(service);
-		serviceProvider.getAuthorizeUrl((err, url) => {
-			if (err) {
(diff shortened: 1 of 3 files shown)


apps/meteor/server/lib/http/call.ts
```diff
@@ -177,12 +184,21 @@ function _call(httpMethod: string, url: string, options: HttpCallOptions, callba
...
+	// If the options argument was omitted, adjust the arguments:
 	if (!callback && typeof optionsOrCallback === 'function') {
 		return _call(httpMethod, url, {}, optionsOrCallback as callbackFn);
 	}
 
 	return _call(httpMethod, url, optionsOrCallback as HttpCallOptions, callback as callbackFn);
 }
 
-export const httpCall = Meteor.wrapAsync(httpCallAsync);
+export const httpCall = async (httpMethod: string, url: string, options: HttpCallOptions) => {
```
sampaiodiego, line 195: I wonder if we should let this as is but start migrating code using this function to use `fetch` instead.

rodrigok: I did the migration to fetch already where it was possible. Had to change this to use Promise.await for the integrations part and let us decide the solution for the integrations later

rodrigok: Thinking better now, we may need to wait Kevin's work replacing the HTTPs by fetch, I didn't realized the other methods as `HTTP.post` were calling the `HTTP.call` as well :(

apps/meteor/server/lib/http/call.ts
```diff
@@ -177,12 +184,21 @@ function _call(httpMethod: string, url: string, options: HttpCallOptions, callba
...
+	return new Promise((resolve, reject) => {
+		httpCallAsync.bind(HTTP)(httpMethod, url, options, (error, result) => {
```
KevLehman, line 197: Q: why do we need to bind HTTP?

PR #28835, opened 2023-04-06
refactor: Convert small files to typescript

KevLehman:


apps/meteor/app/ui-message/client/index.ts
```diff
@@ -1 +1 @@
-import './popup/messagePopupSlashCommandPreview.ts';
+import './popup/messagePopupSlashCommandPreview';
```
KevLehman, line 1:
```suggestion
import './popup/messagePopupSlashCommandPreview.ts';
```

apps/meteor/app/lib/server/functions/saveCustomFieldsWithoutValidation.ts
```diff
@@ -37,7 +39,7 @@ export const saveCustomFieldsWithoutValidation = async function (userId, formDat
 				update.$set[modifyRecordField.field] = customFields[fieldName];
 			}
 
-			await Users.updateOne(userId, update);
+			await Users.updateOne({ _id: userId }, update);
```
sampaiodiego, line 42: I guess this was not working? 🤔  it this function being used anywhere? should we add some tests?

apps/meteor/app/lib/server/functions/parseUrlsInMessage.ts
```diff
@@ -1,21 +1,23 @@
...
 	}
 
 	message.html = message.msg;
 	message = Markdown.code(message);
 
-	const urls = message.html.match(getMessageUrlRegex()) || [];
+	const urls = message.html?.match(getMessageUrlRegex()) || [];
 	if (urls) {
-		message.urls = [...new Set(urls)].map((url) => ({ url }));
+		message.urls = [...new Set(urls)].map((url) => ({ url, meta: {} }));
```
sampaiodiego, line 16: `meta: {}`?

apps/meteor/app/lib/server/functions/saveCustomFieldsWithoutValidation.ts
KevLehman: Yep, was not working 😬 


Let's see if we have tests for it

apps/meteor/app/lib/server/functions/parseUrlsInMessage.ts
KevLehman: It's for "metadata", required in the urls type, but not being passed here 👀 

I can make it optional, aa seems to not cause any problems not being there lol

apps/meteor/app/lib/server/functions/saveCustomFieldsWithoutValidation.ts
sampaiodiego: oh, we handle that 💪 

https://github.com/RocketChat/Rocket.Chat/blob/bc22cfa4c08fbdca4963040177b18fb190775704/apps/meteor/server/models/raw/BaseRaw.ts#L160

sampaiodiego: oops, that's on `findOne`, not `updateOne` 🤦

sampaiodiego: how is TS not complaining about it? can `Filter` be a string? 🤔

KevLehman: Good question. Maybe one type on mongo side is too broad or smth, cause this is something that happened multiple times as of now (string where Filter is expected) 😬

apps/meteor/app/lib/server/functions/insertMessage.ts
```diff
@@ -1,9 +1,15 @@
 import { Messages, Rooms } from '@rocket.chat/models';
+import type { IMessage, IRoom, IUser } from '@rocket.chat/core-typings';
 
 import { validateMessage, prepareMessageObject } from './sendMessage';
 import { parseUrlsInMessage } from './parseUrlsInMessage';
 
-export const insertMessage = async function (user, message, rid, upsert = false) {
+export const insertMessage = async function (
+	user: IUser,
```
tassoevan, line 8: `question` Must it be a full `IUser`? Making it like `Pick<IUser, '_id' | 'username'>` would avoid the assertion at https://github.com/RocketChat/Rocket.Chat/pull/28835/files#diff-4c5584bef947bc711ab14cc203bb3694c22c5a939ca4d4eed9340c9c45750b93R625

KevLehman: 🤔 let me try

tassoevan (approved):

PR #28870, opened 2023-04-11
fix: Inline code formatted as email

hugocostadev:

The Back-stick char  ` was being considered to be part of the local-part of an email address. 

**Previous:**
image

**Current (NOW):**
image

fixed in: https://github.com/RocketChat/fuselage/pull/1021


hugocostadev:
fixed by: https://github.com/RocketChat/Rocket.Chat/pull/28904

PR #28872, opened 2023-04-11
refactor: Rate Limit & DDP Rate Limit to internal packages

KevLehman:
ARCH-975


apps/meteor/app/lib/server/startup/rateLimiter.js
```diff
@@ -47,9 +47,9 @@ DDPRateLimiter._increment = function (input) {
 
 // Need to override the meteor's code duo to a problem with the callback reply
 // being shared among all matchs
-RateLimiter.prototype.check = function (input) {
+RateLimiter.prototype.check = async function (input, sess) {
```
sampaiodiego, line 50: I guess we should move this patch to the main code, since we have the whole code now.

sampaiodiego (changes requested):

apps/meteor/app/lib/server/startup/rateLimiter.js
KevLehman: The patch is actually ported to inside the package, but kept this one in there just in case 🤔 will try to remove

PR #28893, opened 2023-04-12
fix(client): Small UI bugs

tassoevan:

It fixes an insidious bug affecting the focus of the message composer whenever the user start typing. It also reverts some refactored code that affected desktop app users (e.g. cut/copy/paste not working).


jeanfbrito (approved):
:heart:

PR #28909, opened 2023-04-13
refactor: Remove Meteor.runAsUser part 1

MarcosSpessatto:


apps/meteor/app/lib/server/functions/setUsername.ts
```diff
@@ -3,15 +3,70 @@ import { Accounts } from 'meteor/accounts-base';
...
 import { SystemLogger } from '../../../../server/lib/logger/system';
+import { callbacks } from '../../../../lib/callbacks';
+import { joinDefaultChannels } from './joinDefaultChannels';
+
+export const setUsernameWithValidation = async (userId: string, username: string, joinDefaultChannelsSilenced?: boolean): Promise<void> => {
+	if (!username) {
+		throw new Meteor.Error('error-invalid-username', 'Invalid username', { method: 'setUsername' });
+	}
+
+	const user = await Users.findOneById(userId);
```
KevLehman, line 24: Since you're not using the full user, a projection would be good :)

MarcosSpessatto: My main concern here is the callback.run we have below, it's relying on the whole user 😞

ggazzo (approved):

PR #28927, opened 2023-04-15
fix: Change audit main titles

felipe-rod123:

Change  audit main titles from `Message Auditing` to `Audit messages`.

Change from `Message auditing log` to `Audit logs`.

## Further comments

WM-13


rique223 (dismissed):
Please keep the old keys here and change only the translation value on the dictionary. This way we avoid breaking other language translations. E.G.: instead of creating an Audit_logs, lets do "Message_auditing_log": "Audit logs"

hugocostadev (approved):

PR #28951, opened 2023-04-17
refactor: Remove bindenv from settings & centralize settings creation

KevLehman:


KevLehman:
Drafting for now as meteor still doesn't support top level await on 3.0

apps/meteor/app/lib/server/startup/oAuthServicesUpdate.js
```diff
@@ -194,10 +194,10 @@ function customOAuthServicesInit() {
...
-				addOAuthService(name, values);
+				await addOAuthService(name, values);
 			}
 		}
-	});
+	}
 }
 
-customOAuthServicesInit();
+await customOAuthServicesInit();
```
MarcosSpessatto, line 203: Would not be better to create a function as you did in the other ones? 🤔

apps/meteor/server/settings/accounts.ts
```diff
@@ -0,0 +1,949 @@
...
+			await this.add('Accounts_Password_History_Amount', 5, {
+				type: 'int',
+				enableQuery,
+				i18nLabel: 'Password_History_Amount',
+				i18nDescription: 'Password_History_Amount_Description',
+			});
+		});
+	});
+
+await settingsRegistry.addGroup('OAuth', async function () {
```
MarcosSpessatto, line 794: Any reason not to have this inside a function as well?

MarcosSpessatto (dismissed):
I've found these other 2 cases [orchestrator.ts](https://github.com/RocketChat/Rocket.Chat/blob/refactor/bind-env-settings/apps/meteor/ee/server/apps/orchestrator.js#L243-L293) and [assets.ts](https://github.com/RocketChat/Rocket.Chat/blob/refactor/bind-env-settings/apps/meteor/app/assets/server/assets.ts#L359-L364). Shouldn't we move those 2 to functions as well?

apps/meteor/app/lib/server/startup/oAuthServicesUpdate.js
KevLehman: Ah, this is already a function 😬 but didn't move to the other places as this has some extra logic in there (getting settings, env, creating oauth classes etc) so left it for a future refactor: separate setting creation logic from oauth logic and break down the file

apps/meteor/server/settings/accounts.ts
KevLehman: Good catch! This should be on Oauth file 😬

KevLehman:
About assets and orchestrator:

- orchestrator is EE so left intact (since we have the UI and BE to decide whether or not to use these, I don't think it affects but yeah, ee)
- assets has some "customization" to the settings, so left intact

rodrigok (approved):

Issue #22145, opened 2021-05-26
member list is randomly sorted


CvH:
### Description:

The users at the memberlist are randomly sorted and it is difficult to find someone without actually using the search, `/directory/users` works properly in that regard.
The "sort by" setting is also ignored for it.

### Steps to reproduce:

1. open the memberlist at the general channel

### Expected behavior:

list the members in alphabetically

### Actual behavior:

![image](https://user-images.githubusercontent.com/1355173/119620725-9b9fff00-be05-11eb-91e1-154ce6a430f3.png)

### Server Setup Information:

- Version of Rocket.Chat Server: 3.14.0
- Operating System: Ubuntu
- Deployment Method: snap
- Number of Running Instances: 1
- DB Replicaset Oplog: -
- NodeJS Version: v12.22.1
- MongoDB Version: 3.6.14

### Client Setup Information

- Desktop App or Browser Version: Chrome latest
- Operating System: Win/Linux

emikolajczak:
Hi, I confirm. It's hard to say what it is sorted by. Certainly not by Full name or Username

ankar84:
Confirm that issue. Users complaining about that.

johncrisp:
First thing is please upgrade immediately to 3.14.4

See the release notes for why.

Has anyone tested this on either 3.14.4 or 3.15?

CvH:
@johncrisp currently running 3.15, still the same issue appears (updated issue description with new version)

lebkayn:
on 3.16, sorting by name seems to be working. but sorting by activity is totally random

faustfaustov:
> on 3.16, sorting by name seems to be working. but sorting by activity is totally random

on 3.17.1, still sorting by activity is random

CvH:
still a problem at the latest official version currently available 3.18.1

ankar84:
3.18.2 same issue

CvH:
4.3.1 same issue

ankar84:
@dudanogueira @debdutdeb  please take a look at that issue.
It's not so critical in functionality aspect, it's a UI\UX bug, but very annoying to users.

AmiKonst:
So. Where is the solution?

CvH:
5.4 same issue

varun-s22:
Hey there.. is this issue still open?
can i take it?

RG7279805:
@debdutdeb if no one is assigned yet can i work on this issue?

[...]

PR #28954, opened 2023-04-18
chore #22145 issue : Sorting the Member List according to names of the users

M-coder-design:


debdutdeb (changes requested):
should be handled by backend, not manually in the frontend. the endpoint takes a `sort` parameter, as I can see, please use that.

PR #28958, opened 2023-04-18
chore: Improve microservice deserialization of Buffer values

d-gubert:

Currently, if a service sends a Buffer in a method call or expects a buffer as the return value of a call, the serializer returns a `Uint8Array` instead, as that's how EJSON handles these cases.

This PR changes any `Uint8Array` sent as params of service calls or as a result value back into a Buffer, so the services don't need to change their expected values.

## Further comments

There are services in the upcoming Apps-Engine microservices migration that handle Buffers, this PR is required for those services to operate correctly


apps/meteor/ee/server/startup/broker.ts
```diff
@@ -67,7 +67,36 @@ class EJSONSerializer extends Base {
...
+		 * message to an action, i.e. the returning value of a service API call.
+		 * We also want these to be parsed back into a Buffer
+		 */
+		return this.rebufferize(EJSON.parse(buf.toString()));
+	}
+
+	/**
+	 * Recrusively convert all EJSON.Binary objects to Buffers
+	 */
+	rebufferize(obj: unknown): any {
```
ggazzo, line 86:
```suggestion
	private rebufferize(obj: unknown): any {
```

tassoevan (changes requested):
`Serialized` is used by client (web/browser) code too, therefore `Buffer` cannot be used here. Keeping it as `Uint8Array` fits a little bit more, however I'm not 100% sure if we're assuming `Serialized` is JSON or EJSON decoded after encoded.

ggazzo (changes requested):
also looking at your case, you claim that serialized keeps the buffer structure, but Date does not. The curious thing is that EJSON.parse keeps the dates, so something leads me to believe that you want to to use serialized in the wrong place

apps/meteor/ee/server/startup/broker.ts
```diff
@@ -67,7 +67,36 @@ class EJSONSerializer extends Base {
...
+
+		// We exclude strings because they are also iterable but are a primitive type
+		if (obj && typeof obj !== 'string') {
+			// Arrays and objects are iterable and can be handled the same way
+			Object.entries(obj).forEach(([key, value]) => {
+				(obj as Record<string, unknown>)[key] = this.rebufferize(value);
+			});
+		}
+
+		return obj;
```
ggazzo, line 99: this can also cause performance issues it is preferable perform such cases on the "application/function" level

PR #28930, opened 2023-04-15
Added css variable for highlight colour

ravijyoti3:
## Proposed changes (including videos or screenshots)

The word added to highlight in user preferences, gets highlighted successfully. Also the extra spacing from the left has been removed.
image

## Issue(s)

fixes #28845 

## Steps to test or reproduce

1.  Define a custom highlight eg: testing.
2. Compose a message containing the word "testing".
3. Submit the message.

PR #28963, opened 2023-04-18
fix: Fixed custom highlight issue

### LINKED ISSUE #28845: Custom highlights invisible
### Description:

Custom highlights are invisible.

### Steps to reproduce:

1. Define a custom highlight, e.g. `test2`.
2. Compose a message consisting of the words `test1`, `test2` and `test3`, each separated by a line break.
3. Submit this message.

### Expected behavior:

It is clearly visible that the message contains the message `test2`.

### Actual behavior:

I can't clearly see that the message contains  `test2`.

image

### Server Setup Information:

- Version of Rocket.Chat Server: 6.1.1
- Deployment Method: Docker
- MongoDB Version: 6.0.5

### Client Setup Information

Issue is client-agnostic.

### Additional context

As can be seen in the above screenshot, the text `test2` is slightly moved to the right. It turns out that HTML elements for highlighting that word are actually present as this screenshot from Chromium's DevTools shows:

image

The element with the class `highlight-text` even has the CSS property `background-color` set:

image

However, the variable `--selection-background` is not defined anywhere.

vishalhampiholi:

Fixed the highlighting issue 

Before the fix 

before change

After the fix 

After change

Also fixed the small spacing issue .

## Issue(s)

Closes #28845 

## Steps to test or reproduce

1. Define a custom highlight, e.g. test2.
2.Compose a message consisting of the words test1, test2 and test3, each separated by a line break.
3.Submit this message.


hugocostadev:
Please, update your branch and check it again, seems that a lot of unnecessary stuff were added

dougfabris:
Closing this one in favor of: #28930
Thanks for your contribution @vishalhampiholi

PR #28965, opened 2023-04-18
fix: Fixing the user status change by app

AllanPazRibeiro:
Fixing a bug in the app bridge where the user prop was not being considered on the user reactive change regarding the user status. 

## Issue(s)
The app bridge was not correctly handling changes to the user status when the user prop was updated. This led to inconsistencies in the app's UI, where the user status would not be properly reflected in certain cases.

To fix this issue, I updated the app bridge to properly consider changes to the user prop when handling user status changes. This ensures that the app's UI accurately reflects the user's current status at all times.

## Steps to test or reproduce
Basically, you only need to create a simple app that changes the user status and status text by any trigger. Now the user status will change instantly.

### DIFF
--- a/apps/meteor/app/apps/server/bridges/users.ts
+++ b/apps/meteor/app/apps/server/bridges/users.ts
@@ -106,11 +106,7 @@ export class AppUserBridge extends UserBridge {
 		return true;
 	}
 
-	protected async update(
-		user: IUser & { id: string },
-		fields: Partial<IUser> & { statusDefault: string },
-		appId: string,
-	): Promise<boolean> {
+	protected async update(user: IUser & { id: string }, fields: Partial<IUser>, appId: string): Promise<boolean> {
 		this.orch.debugLog(`The App ${appId} is updating a user`);
 
 		if (!user) {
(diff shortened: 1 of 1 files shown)


tiagoevanp (approved):

apps/meteor/app/apps/server/bridges/users.ts
```diff
@@ -121,13 +121,6 @@ export class AppUserBridge extends UserBridge {
 			return true;
 		}
 
-		const { status } = fields;
-		delete fields.status;
```
d-gubert, line 125: Is it necessary to remove the `status` field? It is valid in Rocket.Chat's `IUser` type def https://github.com/RocketChat/Rocket.Chat/blob/390523f3103134a767286420a15a5d02696cafc5/packages/core-typings/src/IUser.ts#L119-L140 - maybe it should be kept?

d-gubert (changes requested):
Can you please remove the `& { statusDefault: string }` part from the `fields`  param in the `update` function as well? It's very confusing and inaccurate, because the Apps-Engine will never send this property from the apps side

apps/meteor/app/apps/server/bridges/users.ts
AllanPazRibeiro: the first one makes sense to me, I will be removing the `{ statusDefault: string }` but I don't get the second comment, I am removing all the lines since it is not relevant to remove the field status since it will be considered in the User update

AllanPazRibeiro: Previously I used the UserPresence.changeDefaultStatus, then @sampaiodiego change it to use the prop statusDefault, but I don't think this is the case now, we need to change the status not the defaultStatus

sampaiodiego (approved):

d-gubert (approved):

PR #28967, opened 2023-04-19
fix: removed the avatar appearing besides the messages 

ShrutiSharma77:


ShrutiSharma77:
removed the avatar showing besides messages in omnichannel section.

PR #28974, opened 2023-04-19
regression: `Not authorized` error when loading a room

gabriellsh:

Listen for changes in the `user` object from `useUser`
This also fixes some logic regarding anonymous reading. (Also a regression)

## Steps to test or reproduce

Login with a user, open a room and refresh without cache. It happens only sometimes, but most times this error appears.
## Further comments

This happened because the `user` object comes from `Meteor.user()` and we provide it in a context above the Room. Unfortunately, this user depends on meteor reactivity, and `username` is left out on the first load, meaning the checks inside `useOpenRoom` fail, even though milliseconds after it runs, username is present.

[SUP-75]

[SUP-75]: https://rocketchat.atlassian.net/browse/SUP-75?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/meteor/app/lib/server/functions/loadMessageHistory.ts
+++ b/apps/meteor/app/lib/server/functions/loadMessageHistory.ts
@@ -14,7 +14,8 @@ export async function loadMessageHistory({
 	showThreadMessages = true,
 	offset = 0,
 }: {
-	userId: string;
+	// userId is undefined if user is reading anonymously
+	userId?: string;
 	rid: string;
 	end: Date | undefined;
 	limit?: number;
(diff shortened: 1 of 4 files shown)


sampaiodiego (changes requested):
looks like you created the branch based on `develop`, so it brought some undesired things to merge.. can you please clean that up?

gabriellsh:
Closing because I cherry-picked my commits to https://github.com/RocketChat/Rocket.Chat/pull/29006

Issue #23977, opened 2021-12-17
Private message unread count 0 for normal user 


nishlashakya:
### Steps to reproduce:

1. call `/api/v1/subscriptions.get` 
2. response object has unread count 0 even if the channel has unread message

### Expected behavior:

get the number of unread count from the channels

### Actual behavior:

![image](https://user-images.githubusercontent.com/13099175/146484987-73141625-d893-4edf-b7b5-c6067d3b0054.png)

### Server Setup Information:

- Version of Rocket.Chat Server: 
rocket chat iframe integration

yksolanki9:
I wish to take up this issue. Please assign it to me

nishlashakya:
thanks @yksolanki9

AbhishekJamhoriya:
Hi, @nishlashakya  is this issue still open if yes, can you please assign me this issue.

nishlashakya:
I am not able to assign the issue. Can you please @dudanogueira ?

nishlashakya:
@dudanogueira hi! when can we expect this fix?

try-catch-stack:
Hey @nishlashakya 
May I know what options are selected in the Unread count settings present in Administration > General .You can only change it if you have the admin rights of the server.

![image](https://user-images.githubusercontent.com/77742477/149526824-4dde7936-c8b9-429c-8cea-da7c70edaf2b.png)

Because by default , it is set to user mentions only and hence the api will return unread count of only the messages in which you are mentioned. Changing it to All messages will return unread count of all messages.

nishlashakya:
hello @try-catch-stack I have tried by updating `Unread count of DIrect Message` to `All messages`. I only get the correct unread count for **admin users** and for **normal users** the unread count is still 0 even it has many unread messages.

Manan17:
Hi is this issue still open if yes, can you please assign me this issue.

J-Ankit2020:
@nishlashakya  can I work on this issue I might be able to fix this

arunvigneshn:
Rocket chat version : 4.7.0

Hi @nishlashakya 

This bug is the default configuration from the admin so only message count not getting updated for rooms, On changing the unread count setting from "User and Group mentions only" setting to "All" it works fine. And it works for both admin and normal users. it was also mentioned in the doc.

To change the settings
Administration -> General -> unread count drop down.
Change the "User and Group mentions only" to "All"

[...]

PR #28988, opened 2023-04-20
fix: Private message unread count 0

Simer13:
Response object has unread count 0 even if the channel has unread message.

#23977 Fixed the error 

The proposed changes are in the typescript file of the following codes- 
apps/meteor/app/api/server/v1/subscriptions.ts

The part where the count was not getting updated has been changed further.


Simer13:
My bad i mistakenly commited incomplete file instead of the some file i changed later on. I will once again open a pr with the right change.

Simer13:
Give me a span of 3/4 days as soon as i complete my exams i will look through this again.

debdutdeb:
you can add more commits here, no need for another pr :)

PR #28989, opened 2023-04-20
chore: add more omnichannel tests

murtaza98:

### DIFF
--- a/apps/meteor/tests/end-to-end/api/livechat/00-rooms.ts
+++ b/apps/meteor/tests/end-to-end/api/livechat/00-rooms.ts
@@ -19,6 +19,7 @@ import {
 	sendMessage,
 	startANewLivechatRoomAndTakeIt,
 	closeOmnichanelRoom,
+	createManager,
 } from '../../../data/livechat/rooms';
 import { addPermissions, updateEEPermission, updatePermission, updateSetting } from '../../../data/permissions.helper';
 import { createUser, login } from '../../../data/users.helper.js';
@@ -340,6 +341,17 @@ describe('LIVECHAT - rooms', function () {
 			const room = await createLivechatRoom(visitor.token);
 
 			await request.get(api('livechat/room.join')).set(credentials).query({ roomId: room._id }).send().expect(200);
+		});
+		it('should allow managers to join a room which is already being served by an agent', async () => {
+			const {
+				room: { _id: roomId },
+			} = await startANewLivechatRoomAndTakeIt();
+
+			const manager: IUser = await createUser();
+			const managerCredentials = await login(manager.username, password);
+			await createManager(manager.username);
+
+			await request.get(api('livechat/room.join')).set(managerCredentials).query({ roomId }).send().expect(200);
 		});
 	});
 
--- a/apps/meteor/tests/data/livechat/rooms.ts
+++ b/apps/meteor/tests/data/livechat/rooms.ts
@@ -153,13 +153,13 @@ export const createAgent = (overrideUsername?: string): Promise<ILivechatAgent> 
 			});
 	});
 
-export const createManager = (): Promise<ILivechatAgent> =>
+export const createManager = (overrideUsername?: string): Promise<ILivechatAgent> =>
 	new Promise((resolve, reject) => {
 		request
 			.post(api('livechat/users/manager'))
 			.set(credentials)
 			.send({
-				username: adminUsername,
+				username: overrideUsername || adminUsername,
 			})
 			.end((err: Error, res: DummyResponse<ILivechatAgent>) => {
 				if (err) {


KevLehman (dismissed):

apps/meteor/tests/end-to-end/api/livechat/00-rooms.ts
```diff
@@ -464,6 +476,23 @@ describe('LIVECHAT - rooms', function () {
 
 			await request.get(api('livechat/room.join')).set(credentials).query({ roomId: room._id }).send().expect(200);
 		});
+		it('should allow managers to join a room which is already being served by an agent', async () => {
+			await updateSetting('Livechat_Routing_Method', 'Manual_Selection');
+			// delay for 1 second to make sure the routing queue gets stopped
+			await sleep(1000);
```
ggazzo, line 482: Avoid sleeping inside tests as much as possible, unless this is a functionality requirement. This slows down tests and can cause flaky tests if more time is needed.

apps/meteor/tests/data/permissions.helper.ts
```diff
@@ -63,3 +64,30 @@ export const removePermissions = async (perms: string[]) => {
...
 	await updateManyPermissions(perms);
 };
+
+type Permission = typeof permissions[number]['_id']
+
+export const removePermissionFromAllRoles = async (permission: Permission) => {
+    await updatePermission(permission, []);
+};
+
+export const addPermissionToDefaultRoles = async (permission: Permission) => {
```
ggazzo, line 74: I don't know how the name of this function makes sense. after a little effort I understood that this function RESTORES the default values (but with a secret even if it is not EE), does it make sense to have this condition inside the test?

maybe there should be an endpoint that does that, don't you think?

ggazzo (changes requested):

apps/meteor/app/authorization/server/functions/upsertPermissions.ts
```diff
@@ -5,230 +5,12 @@ import { Permissions, Settings } from '@rocket.chat/models';
...
-		{ _id: 'clear-oembed-cache', roles: ['admin'] },
-		{ _id: 'videoconf-ring-users', roles: ['admin', 'owner', 'moderator', 'user'] },
-		{ _id: 'view-moderation-console', roles: ['admin'] },
-		{ _id: 'manage-moderation-actions', roles: ['admin'] },
-		{ _id: 'bypass-time-limit-edit-and-delete', roles: ['bot', 'app'] },
-	];
-
 	for await (const permission of permissions) {
-		await Permissions.create(permission._id, permission.roles);
+		const roles = permission.roles.map((role) => role);
```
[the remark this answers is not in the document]
KevLehman: Apparently because `permission.roles` was defined `as const` so the type was a `readonly` intersection... removing

apps/meteor/tests/end-to-end/api/livechat/00-rooms.ts
KevLehman: The sleep when changing routing methods is for the queue to start processing or to stop doing it. Not ideal, but we have to deal with it while we have the arcane queue we have for omni.

apps/meteor/tests/data/permissions.helper.ts
```diff
@@ -63,3 +64,30 @@ export const removePermissions = async (perms: string[]) => {
...
+
+export const restorePermissionToRoles = async (permission: Permission) => {
+    const defaultPermission = permissions.find((p) => p._id === permission);
+    if (!defaultPermission) {
+        throw new Error(`No default roles found for permission ${permission}`);
+    }
+
+    const mutableDefaultRoles: string[] = defaultPermission.roles.map((r) => r);
+
+    if (!IS_EE) {
```
KevLehman, line 82: Do we actually need this logic here?

apps/meteor/tests/end-to-end/api/livechat/09-visitors.ts
```diff
@@ -783,4 +788,125 @@ describe('LIVECHAT - visitors', function () {
 			expect(res.body.visitors[0]).to.have.property('visitorEmails');
 		});
 	});
+	describe('omnichannel/contact', () => {
+		let contact: ILivechatVisitor;
+		it('should fail if user doesnt have view-l-room permission', async () => {
+			await updatePermission('view-l-room', []);
```
KevLehman, line 794: Cannot use your helpers here? 👀

KevLehman (changes requested):

apps/meteor/tests/data/permissions.helper.ts
murtaza98: The api is throwing an error if it doesn't find a role and this is causing issue with these enterprise roles like livechat-monitor. That's y I added this filter here to filter those roles on CE.

apps/meteor/tests/end-to-end/api/livechat/09-visitors.ts
murtaza98: Indeed we should. Done, pls check :heavy_check_mark:

KevLehman (approved):

PR #28996, opened 2023-04-20
fix: registration form shows when should be disabled

hugocostadev:

The 'Create an account shows up on login even if the registration form is 'Disable' or 'Secret URL' through the setting `Workspace > Accounts > Registration > Registration Form`

e.g: 
image

**Previous:** 
image

**Now:**
image

## Steps to test or reproduce

1. Open Rocket.Chat client 
2. Go to `Workspace > Accounts > Registration > Registration Form`
3. Set the value to 'Disable' or 'Secret URL' 
4. On Login Page the link to the registration form still shows up.


yash-rajpal (approved):

hugocostadev:
Put on hold for now since there are other important scenarios that need to be handled

Issue #28586, opened 2023-03-23
Login Terms layout option does not have any effect


Gummikavalier:
### Description:

Custom layout `Login Terms` option does not have any effect to the login screen in RC 6.0.

We've used this to inform our users of extra details before they log in.
Not many people bother to read the actual terms behind the particular link reserved for it and we got law advice that the most crucial terms should be visible in the main login view directly.

### Steps to reproduce:

1. Go to `Admin -> Settings -> Layout -> Home page content`
2. Alter the field `Login Terms`.
3. Check the login screen for changes.

### Expected behavior:

Edited terms should come to effect under the login elements.

### Actual behavior:

The contents of the login terms remain always the same:
![Screenshot from 2023-03-23 20-45-59](https://user-images.githubusercontent.com/6958942/227317191-74540ea6-b3a4-4953-a839-00371e371122.png)

### Server Setup Information:

- Version of Rocket.Chat Server: 6.0.1
- Operating System: RHEL8
- Deployment Method: tar
- Number of Running Instances: 10
- DB Replicaset Oplog: Yes
- NodeJS Version: 14.21.2
- MongoDB Version: 5.0

### Client Setup Information

- Desktop App or Browser Version: Any

jeremyd100:
This issue is also present in version 5.4.4, running within Docker Compose, running against 5.0 Mongo DB.

People that are running on a context path for example /rocketchat, will notice this problem. People running on the default path probably have not noticed this.

The links within the default text will break things.

Ahamad-Tanzeel-Khan:
I think I can work on this issue, can anyone please assign me this project.

hugocostadev:
> I think I can work on this issue, can anyone please assign me this project.

No need to ask for permission to fix the issue, feel free to work on it! 
Just check if someone it's already raised a PR for that (should be linked in this issue).
Best of luck 🚀 👏 

---

**Questions? Help needed? Feature Requests?** 

- Join our [Open Server](https://open.rocket.chat/) in the #support channel and feel free to raise a question
- Join our [Community Forum](https://forums.rocket.chat/) and search/create a post there
- Feature Request: Open an issue in [Feature Request](https://github.com/RocketChat/feature-requests) repository

Dass404:
This article may help to solve this issue- [click here](https://www.mysourcebook.com/2023/04/login-terms-layout-option-does-not-have-any-effect-in-rocketchat-rocket.chat.html) to read.

[...]

PR #28999, opened 2023-04-20
fix: Login Terms custom content

hugocostadev:

The custom layout **Login Terms** option does not have any effect on the login screen.

**Previous:** 

**Now:**
image

## Issue(s)

fix https://github.com/RocketChat/Rocket.Chat/issues/28586

### DIFF
--- a/packages/web-ui-registration/src/components/LoginTerms.tsx
+++ b/packages/web-ui-registration/src/components/LoginTerms.tsx
@@ -1,15 +1,13 @@
-import { Trans } from 'react-i18next';
 import type { ReactElement } from 'react';
-import { Link, HorizontalWizardLayoutCaption } from '@rocket.chat/layout';
+import { HorizontalWizardLayoutCaption } from '@rocket.chat/layout';
+import { useSetting } from '@rocket.chat/ui-contexts';
+import { Box } from '@rocket.chat/fuselage';
 
 export const LoginTerms = (): ReactElement => {
+	const loginTerms = useSetting('Layout_Login_Terms') as string;
 	return (
 		<HorizontalWizardLayoutCaption>
-			<Trans
-				defaults='By proceeding you are agreeing to our <0>{{terms}}</0>, <1>{{privacyPolicy}}</1> and <2>{{legalNotice}}</2>.'
-				values={{ terms: 'Terms of Service', privacyPolicy: 'Privacy Policy', legalNotice: 'Legal Notice' }}
-				components={[<Link href='/terms-of-service' />, <Link href='/privacy-policy' />, <Link href='/legal-notice' />]}
-			/>
+			<Box withRichContent dangerouslySetInnerHTML={{ __html: loginTerms }} />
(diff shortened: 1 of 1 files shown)


yash-rajpal (approved):

ggazzo (dismissed):
are we no longer planning to use translation when text is the default?

hugocostadev:
> are we no longer planning to use translation when text is the default?

Thanks, man! I fixed it and took the chance to change the input to code and add a translation for this content for EN and PT-BR

packages/web-ui-registration/src/components/LoginTerms.tsx
```diff
@@ -1,14 +1,20 @@
...
+
 	return (
 		<HorizontalWizardLayoutCaption>
-			<Trans
-				defaults='By proceeding you are agreeing to our <0>{{terms}}</0>, <1>{{privacyPolicy}}</1> and <2>{{legalNotice}}</2>.'
-				values={{ terms: 'Terms of Service', privacyPolicy: 'Privacy Policy', legalNotice: 'Legal Notice' }}
-				components={[<Link href='/terms-of-service' />, <Link href='/privacy-policy' />, <Link href='/legal-notice' />]}
+			<Box
+				withRichContent
+				dangerouslySetInnerHTML={{ __html: loginTerms !== loginTermsDefault ? loginTerms : t('Layout_Login_Terms_Content') }}
```
debdutdeb, line 17: can we not just set some value as default and handle that instead of this comparison? This strict string comparison is a bit weird considering the end result could be the same but with different formatting, rendering this bit of code useless in its intent.

the end effect would also be weird imo. if the value is **exactly** as the default considered here, if I have a server with portuguese as the server lang, the rendered content will be in portuguese, even though the setting has it in english? one space after <a and before href, and suddenly translation is gone. This would be a nighmare to troubleshoot too for those unaware of this piece of code. 

what if we leave the default empty, and if empty, run it by the default content catering to different locales. If somebody really wants to disable it, they can put an empty div there. Any other content would be exactly rendered as it is in the setting.

debdutdeb (approved):
backend - ok.

Please see the comment.

packages/web-ui-registration/src/components/LoginTerms.tsx
debdutdeb: in this case, I think we can let older ones stay the way they are. And just have this behaviour for the new one. So removing the default value without migration.

so workspaces with the setting set, will show what's in the setting. if they want locale support with the default content, remove the value.

We can add it to the description to make it clear?

IMO it's also worth checking the pre-react-rewrite behaviour.

debdutdeb (approved):
Code wise lgtm. 

@ggazzo Idk if we are only adding migrations only to major versions or not, if so Ig we won't be able to ship it soon. Maybe in that case we should remove the migration, and have this behaviour only for new installs as I suggested initially. Otherwise, this is fine.

PR #29012, opened 2023-04-21
refactor: add I18Next on server-side and deliver files through it

MarcosSpessatto:

This PR adds the i18next to the server side, I added it behind the already existing translation service, so it'll be easy if we decide to migrate it in the future.

The end goal with this change is to replace TAPi18n with i18next completely, We continue reading the files from the i18n folder(symlink) for now, but I'll put all the i18n files into the private folder as soon as I finish all the migration.

I'll address in the next PRs:
- Replace all the server calls to tap and use the translator service instead #29022).
- Add support for i18next on the client side (we already have a hook to do that, but there are still some files relying on TAPi18n to translate), and load everything through "raw" endpoints and remove the TAPi18n way of load translations. (#29023)
- Replace all the client (and missing on server side) calls to TAPi18n and use the previously created i18next helper OR use the hook (if possible). #29034
- Put all the i18n files into the private folder, and remove TAPi18n completely from the project. #29036

Jira: ARCH-1071 ARCH-1014


KevLehman:
Just curious, why does this need to be a service instead of a package? 👀 

btw, we do have a "translation" service that is a proxy for tapi18n, so maybe we can reuse it in this PR if it keeps being a service

MarcosSpessatto:
Yup, it's using the existing service, there's no specific reason to use a service instead of a package, I just wanted to reuse the service. But it might be a better idea to have a package instead.

PR #29020, opened 2023-04-24
fix: System throws an error while visitor tries to register in public queue.

murtaza98:


apps/meteor/app/livechat/server/lib/Livechat.js
```diff
@@ -154,7 +154,7 @@ export const Livechat = {
 			}
 
 			const onlineAgents = await LivechatDepartmentAgents.getOnlineForDepartment(dept._id);
-			if (onlineAgents && onlineAgents.length) {
+			if (onlineAgents && (await onlineAgents.count())) {
```
KevLehman, line 157: Depending on the priorities inside the squad, would be good to start moving this big file to TS, at least, doing a feature pair TS version of it and then migrating step by step :)

KevLehman (approved):

apps/meteor/app/livechat/server/lib/Livechat.js
murtaza98: Totally agree and already started working on it here -> https://github.com/RocketChat/Rocket.Chat/pull/28720

PS: This function was also converted to typescript in above PR. If possible, please review it whenever you're free so we can start merging it ;)

sampaiodiego (dismissed):
is this a regression for 6.2.0 or is it broken on `develop` only?

if a regression, pls change the base branch to `release-6.2.0`, thx

murtaza98:
> is this a regression for 6.2.0 or is it broken on `develop` only?
> 
> if a regression, pls change the base branch to `release-6.2.0`, thx

After further analysis we found this isn't a regression, rather a bug in previous version. I'll change the PR title

aleksandernsilva (approved):

KevLehman (approved):

ggazzo (approved):

sampaiodiego (approved):

PR #29038, opened 2023-04-25
remove http from server

KevLehman:

### DIFF
--- a/apps/meteor/app/autotranslate/server/deeplTranslate.ts
+++ b/apps/meteor/app/autotranslate/server/deeplTranslate.ts
@@ -3,7 +3,6 @@
  */
 
 import { TAPi18n } from 'meteor/rocketchat:tap-i18n';
-import { HTTP } from 'meteor/http';
 import _ from 'underscore';
 import type {
 	IMessage,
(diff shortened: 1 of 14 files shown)


apps/meteor/app/autotranslate/server/deeplTranslate.ts
```diff
@@ -13,6 +12,7 @@ import type {
 	ITranslationResult,
 	ISupportedLanguage,
 } from '@rocket.chat/core-typings';
+import { serverFetch as fetch } from '@rocket.chat/server-fetch';
```
ggazzo, line 15: is not the ideia of exporting `serverFetch` avoid use the same `fetch` name? why not keep `serverFetch` everywhere?

ggazzo (changes requested):

apps/meteor/app/autotranslate/server/deeplTranslate.ts
KevLehman: Think it's more a preference thing. Part of my goal with the exported name is that ppl using it remember is not the standard fetch.

Renaming it to fetch also makes "easier" to move away from it when we use node's built in fetch in the future.

PR #29054, opened 2023-04-25
fix: Fixed canned responses popup adding ! prefix

aleksandernsilva:
## Issue(s)

[SUP-179](https://rocketchat.atlassian.net/browse/SUP-179)

## Further comments

[SUP-179]: https://rocketchat.atlassian.net/browse/SUP-179?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


MartinSchoeler (dismissed):

apps/meteor/app/ui-message/client/popup/components/composerBoxPopup/ComposerBoxPopupCannedResponse.tsx
```diff
@@ -13,7 +13,9 @@ const ComposerPopupCannedResponse = ({ shortcut, text }: ComposerBoxPopupCannedR
 			<OptionColumn>
 				<strong>{shortcut}</strong>
 			</OptionColumn>
-			<OptionContent>{text}</OptionContent>
+			<Box withTruncatedText flexGrow={1} width={0}>
```
aleksandernsilva, line 16: flexGrow 1 + width 0 is to prevent the text from affecting the width of the flex container and still retain the ellipsis behavior.

MartinSchoeler (dismissed):

apps/meteor/app/ui-message/client/popup/components/composerBoxPopup/ComposerBoxPopupCannedResponse.tsx
dougfabris: Guys, we shouldn't use custom componentization inside our build-in components, `<OptionContent />` should have the ellipsis behavior by default for long texts, if it's being used correctly, if not we should consider checking whats wrong and fixing it on the Fuselage side! Please let's consider this for all similar cases.

dougfabris (changes requested):

apps/meteor/app/ui-message/client/popup/components/composerBoxPopup/ComposerBoxPopupCannedResponse.tsx
aleksandernsilva: Changed my approach. Thanks for the heads up.

PR #27653, opened 2022-12-27
[FIX] Remove invalid settings at startup

KevLehman:
## Further comments

Fixes #26516

PR #29056, opened 2023-04-25
fix: Custom OAuth settings are not visible

matheusbsilva137:

 - Ensure that custom OAuth settings have their `public` field defined so that they're displayed in the UI.

## Steps to test or reproduce

Create a custom OAuth in an older RC version (I'd suggest 4.x or prior) and check if its settings have the `public` field in the DB (or erase it manually). Upgrade to 6.x and try to check the custom OAuth settings in the UI.
**Current behavior:** the settings are not displayed;
**New behavior:** the settings have been added the `public` field in the DB (with the `false` value) and, hence, are visible in the UI.

## Further comments

Introduced in #27653
SUP-208


totoi690 (approved):

debdutdeb:
@matheusbsilva137 conflicts 😬

apps/meteor/app/lib/server/functions/addOAuthService.ts
```diff
@@ -9,27 +9,31 @@ export async function addOAuthService(name: string, values: { [k: string]: strin
 	name = capitalize(name);
 	await settingsRegistry.add(`Accounts_OAuth_Custom-${name}`, values.enabled || false, {
 		type: 'boolean',
+		public: false,
```
sampaiodiego, line 12: there is no need to say `public: false` since all settings have this by default, see:
https://github.com/RocketChat/Rocket.Chat/blob/d34ce1569bfa01b99477751b07c5726a9d511e73/apps/meteor/app/settings/server/functions/getSettingDefaults.ts#L35

apps/meteor/server/startup/migrations/v298.ts
```diff
@@ -0,0 +1,11 @@
+import { Settings } from '@rocket.chat/models';
+
+import { addMigration } from '../../lib/migrations';
+
+addMigration({
+	version: 295,
```
sampaiodiego, line 6: migration version is not matching the file name

sampaiodiego (changes requested):
I'm not sure I understood the reasoning behind this changes.. the linked PR #27653 doesn't seem to be the reason.

based on the PR description, some specific version was creating Custom OAuth settings without the `secret` field?

apps/meteor/server/startup/migrations/v298.ts
matheusbsilva137: Oops, fixed :nerd_face:

matheusbsilva137:
> based on the PR description, some specific version was creating Custom OAuth settings without the `secret` field?

Yep @sampaiodiego . We received a customer report about this, their Custom OAuth settings were not being displayed. I guess they created it in a quite old version where this field was not inserted (the linked PR assumes that settings without the `public` field are invalid)
Inserting the `public` field fixed the issue for them. So I'm doing this for all user to assure this doesn't happen again.

pierre-lehnen-rc (approved):

PR #29062, opened 2023-04-26
fix: system does not allow agents to accept chats in progress equal to the no. of simultaneous chats.

murtaza98:
## Further comments

Note: This PR is derived from [this PR](https://github.com/RocketChat/Rocket.Chat/pull/28252), but it just contains the fix and not the refactoring we've made over there. The intent of making this PR is to include it within patches for previous versions.


KevLehman (commented):
Given the nature of the fix, would you mind adding some api tests? 🙏🏽

murtaza98:
> Given the nature of the fix, would you mind adding some api tests? 🙏🏽

I agree and I've already added the tests for on-hold feature over here -> https://github.com/RocketChat/Rocket.Chat/pull/28252

Note: This PR is already derived from that PR I mentioned above however it just contains the fix and not the refactoring. The intent of making this PR is to include it within patches for previous versions. 

To avoid confusion, let me move this PR to draft for now and we can move it back to review once we have more info about which versions to patch

PR #29090, opened 2023-04-28
fix: misleading of 'total' in team members & members fixed

### LINKED ISSUE #29041: Number of group members wrongly replaced with number of online group members
Hallo,

thanks for developing this cool functional chat system!

On a channel, when expanding the members list, the default behavior is to show only the users who are online.

[FEATURE REQUEST] It would be preferable to show all users as the default, or at least allow choosing the initial view via a setting.

[BUG] The caption appearing over the list reports the same number for online users and total users (in this screenshot it reads: "Show: 3 Total: 3"). This is really misleading! The second number should correspond to the real total number of group members.

image

anikdhabal:

In team members or members **total** is defined to the total number of members when type is **All** and total number of online when type is **Online**. That could be misleading to the end user and could get confused by the **Total** field.

Before:

Now:

## Issue(s)

fix #29041


anikdhabal:
@hugocostadev can you review it sir?

hugocostadev:
HI @anikdhabal , thanks for your contribution

Design team feedback:  _I'd rather change the label to `Showing 2 of 2` or `Showing 19 of 833`_

can you do this change? Thanks

hugocostadev (dismissed):

hugocostadev (approved):

apps/meteor/client/views/room/contextualBar/RoomMembers/RoomMembers.tsx
```diff
@@ -118,11 +118,7 @@ const RoomMembers = ({
 				{!loading && members.length > 0 && (
 					<Box pi='x18' pb='x12'>
 						<Box is='span' color='hint' fontScale='p2'>
-							{t('Showing')}: {members.length}
-						</Box>
-
-						<Box is='span' color='hint' fontScale='p2' mis='x8'>
-							{t('Total')}: {total}
+							{t('Showing')} {members.length} {t('of')} {total}
```
gabriellsh, line 121: Please, create a single translation key for the whole text.

Ex: `{t('Showing__current__of__total', { current: members.length, total })}`

gabriellsh: This is needed because in certain languages the order of verbs/nouns/connectors can change, and the only way to provide a correct translation would be to make a string for the whole text.

gabriellsh (changes requested):

apps/meteor/client/views/room/contextualBar/RoomMembers/RoomMembers.tsx
anikdhabal: Fixed

anikdhabal:
@gabriellsh @hugocostadev I've created a single translation key.

hugocostadev (approved):

hugocostadev (approved):

dougfabris (dismissed):

hugocostadev (approved):

PR #29097, opened 2023-04-28
refactor: create i18n package

ggazzo:


KevLehman (dismissed):

packages/i18n/src/locales
```diff
@@ -0,0 +1 @@
+../../../apps/meteor/packages/rocketchat-i18n/i18n
```
MarcosSpessatto, line 1: Any specific reason to use a symbolic link here? Since we are replacing TAPPi18n completely, wouldn't be better to remove the rocketchat-i18n from the meteor packages folder?

MarcosSpessatto (dismissed):

sampaiodiego (changes requested):
looks like the language selection in "My Account" is not working as before:

before:
image

now:
image

the same for the Language setting on Admin:
image


they are also missing the "Default" value.


some weird behavior, after I change my preferred language on My Account, it doesn't translate everything without a refresh.. but it changes the language for *some* strings.

packages/i18n/src/locales
ggazzo: the lingo hub must to be configured and to be honest I'm not fluent with that tool so I'm keeping this for next steps

packages/i18n/src/index.mjs
```diff
@@ -0,0 +1,120 @@
...
+	`const languages: string[];
+export default languages;`,
+);
+
+fs.writeFileSync(`./dist/index.mjs`, esm);
+fs.writeFileSync(`./dist/index.js`, cjs);
+fs.writeFileSync(`./dist/index.d.ts`, tds);
+fs.writeFileSync(`./dist/stats.txt`, interpolationReplacementsStats());
+
+// files.forEach((file) => {
```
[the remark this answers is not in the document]
ggazzo: just kept there just in case we need to run it again 🤷

ggazzo:
> some weird behavior, after I change my preferred language on My Account, it doesn't translate everything without a refresh.. but it changes the language for _some_ strings.

it should be solved by now ;)

ggazzo:
WHERE IS MY APPROVAL @sampaiodiego ❤️ ?

PR #29101, opened 2023-04-30
fix: Add phone number validation 

### LINKED ISSUE #20528: Missing phone number validation
### Description:
Phone Number validation should be present and only phone numbers should be allowed in phone field.

### Steps to reproduce:

1. Contact center, and make a new contact
2. No phone number validation is present.

### Expected behavior:
Phone number should be validated

### Actual behavior:
Phone number is not validated

### Server Setup Information:

- Version of Rocket.Chat Server: 3.11.0-develop
- Operating System: Ubuntu
- Deployment Method: 
- DB Replicaset Oplog: 
- NodeJS Version: 12.18.4
- MongoDB Version: 4.2.8

### Client Setup Information

- Desktop App or Browser Version: Chrome version 87
- Operating System: Ubuntu

### Additional context

I can enter anything in phone number field.
image

anikdhabal:

Before:-
Screenshot 2023-04-30 011711

Now:-
Screenshot 2023-04-30 064147

## Issue(s)

fix #20528 
## Steps to test or reproduce

1.Open Contact Center
2.Click on New Contact

### DIFF
--- a/apps/meteor/client/views/omnichannel/directory/contacts/contextualBar/ContactNewEdit.tsx
+++ b/apps/meteor/client/views/omnichannel/directory/contacts/contextualBar/ContactNewEdit.tsx
@@ -7,6 +7,7 @@ import { useForm } from 'react-hook-form';
 
 import { hasAtLeastOnePermission } from '../../../../../../app/authorization/client';
 import { validateEmail } from '../../../../../../lib/emailValidator';
+import { validatePhone } from '../../../../../../lib/phoneValidator';
 import { withDebouncing } from '../../../../../../lib/utils/highOrderFunctions';
 import { CustomFieldsForm } from '../../../../../components/CustomFieldsFormV2';
 import VerticalBar from '../../../../../components/VerticalBar';
@@ -119,8 +120,11 @@ const ContactNewEdit = ({ id, data, close }: ContactNewEditProps): ReactElement 
 	};
 
 	const isPhoneValid = async (phone: string): Promise<boolean | string> => {
-		if (!phone || initialValue.phone === phone) {
+		if (phone === initialValue.phone) {
 			return true;
+		}
+		if(!validatePhone(phone)){
+			return t('error-invalid-phone-number');
 		}
 
 		const { contact } = await getContactBy({ phone });
--- a/apps/meteor/lib/phoneValidator.ts
+++ b/apps/meteor/lib/phoneValidator.ts
@@ -1,0 +1,6 @@
+export const validatePhone = (phone: string): boolean => {
+    const rfcPhoneRegex =/^[+]?[\s./0-9]*[(]?[0-9]{1,4}[)]?[-\s./0-9]*$/g;
+       
+          return rfcPhoneRegex.test(phone);
+};
+
--- a/apps/meteor/packages/rocketchat-i18n/i18n/en.i18n.json
+++ b/apps/meteor/packages/rocketchat-i18n/i18n/en.i18n.json
@@ -1951,6 +1951,7 @@
   "error-invalid-param": "Invalid param",
   "error-invalid-params": "Invalid params",
   "error-invalid-permission": "Invalid permission",
+  "error-invalid-phone-number": "Invalid phone number",
   "error-invalid-port-number": "Invalid port number",
   "error-invalid-priority": "Invalid priority",
   "error-invalid-redirectUri": "Invalid redirectUri",


apps/meteor/lib/phoneValidator.ts
```diff
@@ -0,0 +1,6 @@
+export const validatePhone = (phone: string): boolean => {
+    const rfcPhoneRegex =/^[+]?[\s./0-9]*[(]?[0-9]{1,4}[)]?[-\s./0-9]*$/g;
```
debdutdeb, line 2: where did you get this regex?

debdutdeb, line 2:
```suggestion
    const regex = /^[+]?[\s./0-9]*[(]?[0-9]{1,4}[)]?[-\s./0-9]*$/g;
```

apps/meteor/lib/phoneValidator.ts
```diff
@@ -0,0 +1,6 @@
...
+       
+          return rfcPhoneRegex.test(phone);
```
debdutdeb, line 4:
```suggestion
          return regex.test(phone);
```

apps/meteor/client/views/omnichannel/directory/contacts/contextualBar/ContactNewEdit.tsx
```diff
@@ -119,9 +120,12 @@ const ContactNewEdit = ({ id, data, close }: ContactNewEditProps): ReactElement
 	};
 
 	const isPhoneValid = async (phone: string): Promise<boolean | string> => {
-		if (!phone || initialValue.phone === phone) {
+		if (phone === initialValue.phone) {
```
debdutdeb, line 123: this changes the logic, I'm not confident if this is expected elsewhere or not.

previously an empty input would've returned true, despite of if it matches initialValue or not. e.g. `isPhoneValid('')` would've been true even if `initialValue.phone != ''` which isn't the case anymore here.

cc @RocketChat/omnichannel

debdutdeb (changes requested):

apps/meteor/lib/phoneValidator.ts
anikdhabal: @debdutdeb from quera. It supports almost all international number.

apps/meteor/client/views/omnichannel/directory/contacts/contextualBar/ContactNewEdit.tsx
anikdhabal: ok

apps/meteor/client/views/omnichannel/directory/contacts/contextualBar/ContactNewEdit.tsx
```diff
@@ -119,9 +120,12 @@ const ContactNewEdit = ({ id, data, close }: ContactNewEditProps): ReactElement
...
 			return true;
 		}
+		if(!validatePhone(phone)){
```
murtaza98, line 126: This is only getting validated on the FE, what about on the BE side?

apps/meteor/lib/phoneValidator.ts
```diff
@@ -0,0 +1,6 @@
+export const validatePhone = (phone: string): boolean => {
+    const regex = /^[+]?[\s./0-9]*[(]?[0-9]{1,4}[)]?[-\s./0-9]*$/g;
```
murtaza98, line 2: Instead of hardcoding this, I'd ask you to make this configurable by adding a setting for it on the admin panel. This way if some user is trying to enter a number which is valid for them, then they can modify the regex to suite their needs

murtaza98:
Hi @anikdhabal A gentle reminder about the review suggestions. Thanks for your contribution :rocket:

anikdhabal:
> Hi @anikdhabal A gentle reminder about the review suggestions. Thanks for your contribution 🚀

Yes sir, I will fix it by today. Thank you for your suggestion.

KevLehman:
Hey hey! Thanks for your contribution!

There is a reason we cannot implement strict phone number validation right now on this field 😬 and it's because some integrations rely on this field being "freeform" to save some data. For example, whatsapp and telegram apps (and twilio integration) create phone numbers in the form of `integration:+50505050505` which would conflict with the validations on that field (when you edit one contact with this kind of data, the validation will trigger and won't let you save)
image

So, we can't merge this one _yet_, till we move that data elsewhere (which is in our plans)

PR #29115, opened 2023-05-02
fix: Incomplete emails are being accepted at user registration

felipe-rod123:

Created a RegExp validator to check the emails in `packages/web-ui-registration/src/RegisterForm.tsx` and display an error.

## Further comments

WM-30


hugocostadev (changes requested):
This forms make use of useForm hook from `react-hook-form`
Use the hook features like `rule` instead of creating your own validation. 

BTW, check other forms that are validating the email field, maybe there is a util function or an easier way to check this

packages/web-ui-registration/src/RegisterForm.tsx
```diff
@@ -33,6 +34,9 @@ export const RegisterForm = ({ setLoginRoute }: { setLoginRoute: DispatchLoginRo
 
 	const formLabelId = useUniqueId();
 	const registerUser = useRegisterMethod();
+	const [errorOnRegister, setErrorOnRegister] = useState<string | undefined>(undefined);
+	const [invalidEmailInput, setInvalidEmailInput] = useState<boolean>(false);
+	const debouncedInvalidEmailInput = useDebouncedValue(invalidEmailInput, 5000);
```
hugocostadev, line 39: 5 seconds?  It's too much I think, decrease to 2000 at max

hugocostadev (changes requested):
Why not use useForm from react-hook-form ?

hugocostadev (changes requested):
Why not use useForm from react-hook-form ?

hugocostadev (dismissed):

hugocostadev (approved):

PR #29119, opened 2023-05-02
refactor: More files to ts - 2

KevLehman:


apps/meteor/app/2fa/server/MethodInvocationOverride.ts
```diff
@@ -2,14 +2,25 @@ import { DDPCommon } from 'meteor/ddp-common';
...
+		userId?: string;
+	}) {
 		const result = super(options);
 		const currentInvocation = DDP._CurrentInvocation.get();
 
 		if (currentInvocation) {
 			this.twoFactorChecked = currentInvocation.twoFactorChecked;
 		}
 
+		// @ts-expect-error - `super` return is typed as `void` even when types says it returns an instance of `MethodInvocation`
```
ggazzo, line 23: is this something really necessary ? I think constructor always returns a instance to its right have a return void

KevLehman: It should be `instance of X` for sure, but seems like an ongoing issue on TS

https://github.com/microsoft/TypeScript/issues/37847

PR #29129, opened 2023-05-02
chore: hide releases tab for private apps on marketplace

csuadev:
Hide the releases tab for private apps to fix error when trying to get info from marketplace releases.

image

## Issue(s)
[https://rocketchat.atlassian.net/browse/AECO-53](https://rocketchat.atlassian.net/browse/AECO-53)


apps/meteor/client/views/marketplace/AppDetailsPage/AppDetailsPageTabs.tsx
```diff
@@ -51,7 +51,7 @@ const AppDetailsPageTabs = ({
 					{t('Security')}
 				</Tabs.Item>
 			)}
-			{marketplace !== false && (
+			{marketplace !== false && context !== 'private' && (
```
tiagoevanp, line 54: I think this marketplace test is not necessary, this prop always came as undefined... 🤔 

I suggest to check the necessity and try to find what is happening. 

If the test is from a boolean, why the type is unknown?

tiagoevanp (changes requested):

tiagoevanp (approved):

PR #29140, opened 2023-05-03
fix: Handler for "apps.added" event not working properly

d-gubert:

THIS IS A TEST. We're trying an approach to address reports of app management inconsistencies in the product.

This PR has been created so that a docker image is pushed to our registry for further testing.

## Further comments

AECO-100
AECO-150


d-gubert:
Not applicable anymore

PR #29142, opened 2023-05-03
feat: Implement new Administration kebab reroute algorithm

rique223:
## Jira task: [WM-27](https://rocketchat.atlassian.net/browse/WM-27?atlOrigin=eyJpIjoiODQxMDQzMDhjNTAyNDcxOWE1ZThhNWZmZTA2ZTg5MDciLCJwIjoiaiJ9)
Implemented a new custom hook that finds the first page an user has permission to see before redirecting them in order to avoid infinite loops and other UX related problems. Currently it is being used only for the 'Workspace' option of the administration kebab menu but it was designed to be used with any page that follows the same modular permission logic, E.G.: Omnichannel and possibly Marketplace in a probable future.

## Further comments

[WM-27]: https://rocketchat.atlassian.net/browse/WM-27?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


hugocostadev (changes requested):
Hi @rique223 , to make it more simple and avoid unhandled scenarios I suggest: 
- using the `getAdminSidebarItems` method to set the default route for admin

hugocostadev (approved):

PR #29151, opened 2023-05-04
regression: Fix UserCard height to avoid position glitch

yash-rajpal:

The user card has different dimensions when it is loading and when it has loaded the contents. Due to this change in dimensions, it sometimes changes the position of the userCard. So, making the user card as fixed size so this doesn't happen.

## Further comments

[TC-652]

[TC-652]: https://rocketchat.atlassian.net/browse/TC-652?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


dougfabris:
Did you test it with big bio content?

yash-rajpal:
> Did you test it with big bio content?

Yes, big bio is truncated.

image

gabriellsh (approved):

PR #29152, opened 2023-05-04
chore(ignore): just something for sing.li

debdutdeb:


apps/meteor/server/models/raw/Messages.ts
```diff
@@ -83,7 +83,7 @@ export class MessagesRaw extends BaseRaw<IMessage> implements IMessagesModel {
 		const query: Filter<IMessage> = {
 			'_hidden': { $ne: true },
 			'mentions.username': username,
-			rid,
+			rid: 'GENERAL',
```
Dnouv, line 86: Hmm, looks like not standard, but for Sing, I will require a change, please?

Dnouv (approved):
LGTM!

PR #29153, opened 2023-05-04
fix: Whatsapp quotes not working due to livechat url change

aleksandernsilva:


aleksandernsilva:
Hey @ggazzo !
This PR is trying to fix an issue that me an Tasso have been dealing for a while. You can see some of our previous attempts here: https://github.com/RocketChat/Rocket.Chat/pull/28957.

The original issue was caused by the Toolbox's `useRoute` not being updated when `openRoomInfo()` was called. The cleanest solution so far is the one implemented in the PR above (#28957), but it caused problems in other parts of the app that depends on the url without the tab.

For those reasons I'm considering returning the original `setTimeout`, which seems to be the only solution working at the moment. I know this workaround is far from ideal. Do you have any suggestions on ways to improve this behavior?

ggazzo (approved):

aleksandernsilva:
It's already been QA'd, the labels just weren't updated 😉

Issue #29087, opened 2023-04-28
Unable to enable read-only mode when creating a channel


shevchukma:
### Description:

When creating a channel as a regular user, it is not possible to enable the read-only option. However, this option is active for the Administrator. The user is forced to create a regular channel, and then, in the settings, switch it to read-only mode.
![image](https://user-images.githubusercontent.com/104189570/235135563-642bcc49-a8a8-4202-8222-5d81bb4f5e6c.png)

### Steps to reproduce:

1. Click on the button for creating channels from a user with the "User" role.
![image](https://user-images.githubusercontent.com/104189570/235138497-2e544f49-5bb4-45f3-8435-44dc91f70921.png)
2. Try to enable read-only mode and create a channel. 

### Expected behavior:

It is possible to enable read-only mode when creating a channel as a regular user.

### Actual behavior:

Read-only mode can only be enabled by an administrator when creating a channel.

### Server Setup Information:

- Version of Rocket.Chat Server: 6.1.4 (on open.rocket.chat the issue is also reproduced)
- Operating System: CentOS 7.9
- Deployment Method: docker
- Number of Running Instances: 3
- DB Replicaset Oplog: yes
- NodeJS Version: v14.21.2
- MongoDB Version: 5.0.15

### Client Setup Information

- Desktop App or Browser Version: Microsoft Edge 111.0.1661.43 
- Operating System: Windows 10

### Additional context

I know that a room can be set to read-only after it's created, but this confuses users. Especially considering that this switch is there, but only administrators can switch it.

Ninad3721:
Hello @shevchukma  ,I would like to solve this issue. This is my first contribution to this project , can you tell me how can we switch profile from user to administrator and vice-versa.

PR #29168, opened 2023-05-08
fix: unable to enable read-only mode  by the normal user while creating a new channel

Ninad3721:

image

This PR will remove the fix the issues of a normal user not able to enable read only mode while creating the channel.

## Issue(s)

Fixes https://github.com/RocketChat/Rocket.Chat/issues/29087
## Steps to test or reproduce

1.Login in as a normal user (not admin)
2.Click on create channel (top left corner of home page)

The read only mode is not disabled anymore


Nabhag8848:
Hey @Ninad3721 you should avoid pushing package-lock.json + any .gitpod file which is not intended unless its related to gitpod. Make an appropriate changes while pushing to same branch.

Ninad3721:
Ok @Nabhag8848 I will do the necessary changes.

dougfabris:
@Ninad3721 Just removing the permission checking is not the proper fix, because of that, I'm going to close this one!
Thank you so much for your contribution!

PR #29179, opened 2023-05-08
refactor: Convert livechat rooms model to ts

KevLehman:


murtaza98 (approved):

apps/meteor/server/models/raw/BaseRaw.ts
```diff
@@ -177,7 +177,7 @@ export abstract class BaseRaw<
 		return this.col.find(query, optionsDef);
 	}
 
-	findPaginated<P extends Document = T>(query: Filter<T>, options?: FindOptions<P extends T ? T : P>): FindPaginated<FindCursor<WithId<P>>>;
+	findPaginated<P extends Document = T>(query: Filter<T>, options?: FindOptions<P extends T ? T : P>): FindPaginated<FindCursor<WithId<T>>>;
```
ggazzo, line 180: is there any reason why you changed this?

ggazzo (changes requested):

apps/meteor/server/models/raw/BaseRaw.ts
KevLehman: The problem is this when using P:
```
IOmnichannelRoom is not assignable to EnhancedOmit<P, '_id'>
```

At some point, P != IOmnichannelRoom and types get crazy. Changing it to use T works, but it only works because T is not coming from the func but from the model. 

Do you know any way to fix this? Otherwise, I'll put an any as return type.

murtaza98 (dismissed):

.changeset/spicy-books-act.md
sampaiodiego, line 1: maybe remove the changeset? 😬

murtaza98: Done ;)

sampaiodiego (approved):

PR #29184, opened 2023-05-09
regression: Delete Message Modal not closing on error in deleting a message

Dnouv:

This PR fixes the delete message modal on Moderation Console not closing, if an error occurs while deleting the message


debdutdeb (changes requested):
It's either msgid or userid, this would not indicate the correct outcome if there is no msgid with the passed one to begin with.

put that in an else block and that should be ok (as ajv will make sure at least one is passed).

debdutdeb:
> This PR fixes the error occurring when trying to delete a single reported message.

Also, just to clarify, the patch works on the dismiss endpoint, not message delete one?

apps/meteor/app/api/server/v1/moderation.ts
```diff
@@ -159,14 +159,14 @@ API.v1.addRoute(
 			const { userId, msgId, reason, action: actionParam } = this.bodyParams;
 
 			if (userId) {
-				const user = await Users.findOneById(userId, { projection: { _id: 1 } });
+				const user = await ModerationReports.findOne({ 'message.u._id': userId }, { projection: { _id: 1 } });
 				if (!user) {
 					return API.v1.failure('user-not-found');
 				}
```
debdutdeb, line 165:
```suggestion
				const report = await ModerationReports.findOne({ 'message.u._id': userId, _hidden: { $ne: true } }, { projection: { _id: 1 } });
				if (!report) {
					return API.v1.failure('no-reports-found');
				}
```

apps/meteor/app/api/server/v1/moderation.ts
```diff
@@ -159,14 +159,14 @@ API.v1.addRoute(
...
 			}
 
 			if (msgId) {
-				const message = await Messages.findOneById(msgId, { projection: { _id: 1 } });
+				const message = await ModerationReports.findOne({ 'message._id': msgId }, { projection: { _id: 1 } });
 				if (!message) {
 					return API.v1.failure('error-message-not-found');
 				}
```
debdutdeb, line 172:
```suggestion
				const report = await ModerationReports.findOne({ 'message._id': msgId, _hidden: { $ne: true } }, { projection: { _id: 1 } });
				if (!report) {
					return API.v1.failure('no-reports-found');
				}
```

debdutdeb (changes requested):
might require a prettier format.

hugocostadev (dismissed):

debdutdeb (dismissed):

apps/meteor/client/views/admin/moderation/hooks/useDeleteMessage.tsx
```diff
@@ -4,20 +4,36 @@ import React from 'react';
...
 
 	const handleDeleteMessages = useMutation({
-		mutationFn: deleteMessage,
+		mutationFn: async ({ msgId, roomId, asUser }: { msgId: string; roomId: string; asUser: boolean }) => {
+			try {
+				await getMessage({ msgId });
+			} catch (error) {
+				return error;
+			}
+			return deleteMessage({ msgId, roomId, asUser });
```
ggazzo, line 27: is this really the way?
I'm afraid sometimes the user will not be able to get the message through `chat.getMessage` or even delete the message using 'chat.detele' for permission reasons but still able to moderate, maybe the we should use the same logic from `/v1/moderation.user.deleteReportedMessages` but only for one message

the hook `useDeleteMessagesAction` looks ideal btw

ggazzo (changes requested):

apps/meteor/client/views/admin/moderation/hooks/useDeleteMessage.tsx
debdutdeb: We basically want to handle a specific not-found situation for chat.delete. For any other case, block the dismiss action. This does make the assumption that user x can read all message. 

> the hook useDeleteMessagesAction looks ideal btw

that I think is too much? delete all message for a user vs a single random message. 

At this point i'm thinking modifying the dismissReport endpoint to pass an optional like `deleteMessages` to first delete (or ignore if doesn't exist) then mark hidden

Dnouv: Yes, that had be good. But, I would prefer to not modify the API endpoints now, considering the release deadline 😬.

debdutdeb: Don't think you need to worry about that too much.

hugocostadev (approved):

apps/meteor/client/views/admin/moderation/hooks/useDeleteMessage.tsx
Dnouv: It makes sense, and there might be cases where the moderator might not have access to messages, but as it is now, the Moderation Console is not yet available for scoped roles, so it might be very few cases where it does happen. 

And the main concern for opening this fix was that if there was any such error while deleting a single message, the error was shown as a toast, but the confirmation modal was not getting closed. And despite the message not being deleted, the reports against that message were cleared out, which is not something good, as per my understanding.

And after this fix, the delete confirmation model will get closed in case of any error on trying to delete the message. But the reports won't be dismissed; in the meanwhile, the user who was trying to perform this moderation action can figure out why there was an error, resolve it, come back, and delete it again. As a last resort, if it is some strange unknown issue, they can force dismiss the reported message using the 'Dismiss Reports' action. So, as far as I could see, it is all covered. I am open to suggestions, in case, I overlooked any cases.

Thank you!

ggazzo (approved):

PR #29186, opened 2023-05-09
test: Make PDF generation sync on test mode

KevLehman:

### DIFF
--- a/ee/packages/omnichannel-services/src/OmnichannelTranscript.ts
+++ b/ee/packages/omnichannel-services/src/OmnichannelTranscript.ts
@@ -136,12 +136,19 @@ export class OmnichannelTranscript extends ServiceClass implements IOmnichannelT
 
 		await LivechatRooms.setTranscriptRequestedPdfById(details.rid);
 
+		// Make the whole process sync when running on test mode
+		// This will prevent the usage of timeouts on the tests of this functionality :)
+		if (process.env.TEST_MODE) {
+			await this.workOnPdf({ details: { ...details, from: this.name }, template: 'omnichannel-transcript' });
+			return;
+		}
+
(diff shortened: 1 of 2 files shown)


murtaza98 (approved):

ee/packages/omnichannel-services/src/OmnichannelTranscript.ts
```diff
@@ -323,45 +328,58 @@ export class OmnichannelTranscript extends ServiceClass implements IOmnichannelT
...
+			await this.doRender({ data, details });
 		} catch (error) {
 			await this.pdfFailed({ details, e: error as Error });
 		} finally {
 			this.currentJobNumber--;
 		}
 	}
 
-	async doRender({ template, data, details }: { template: Templates; data: WorkerData; details: WorkDetailsWithSource }): Promise<void> {
+	async promisifyStream(stream: NodeJS.ReadableStream, buffer: Uint8Array[]) {
```
sampaiodiego, line 339: maybe you can use this one instead? https://github.com/RocketChat/Rocket.Chat/blob/eb40ccef9f9672e0ade130d429040efa2a0c20c4/apps/meteor/app/file-upload/server/lib/streamToBuffer.ts#L3

maybe move it to a lib? wdyt?

sampaiodiego (changes requested):

ee/packages/omnichannel-services/src/OmnichannelTranscript.ts
KevLehman: All right, moved the function to `tools` package, but won't refactor the other places on this PR 👀 

but now we have a central func for doing so.

sampaiodiego (approved):

PR #29203, opened 2023-05-10
feat: access-marketplace permission

yash-rajpal:

access-marketplace permission for marketplace section

image

## Further comments

TC-626


gabriellsh (changes requested):
If I access the marketplace without permission, I can see the sidebar. In this case we should show the page no found.

image

gabriellsh (dismissed):

tiagoevanp (commented):
I have a question, why we are fixing permissions for admin and users only? What kind of role shouldn't be able to visit marketplace view since we are trying to expose the apps? The code looks fine to me, but I want to have better explanation about the reason for this PR to accept it... 🤔

tiagoevanp (dismissed):

gabriellsh:
> I have a question, why we are fixing permissions for admin and users only? What kind of role shouldn't be able to visit marketplace view since we are trying to expose the apps? The code looks fine to me, but I want to have better explanation about the reason for this PR to accept it... 🤔

We added this new permission as default for regular users, it doesn't change behaviour, but now if an admin wants to remove this permission, they can do it.

gabriellsh (approved):

matheusbsilva137 (approved):
Back-end LGTM

PR #29207, opened 2023-05-10
refactor: replace `useForm` in favor of `RHF` on `TwoFactorTOTP`

juliajforesti:


apps/meteor/client/views/account/security/TwoFactorTOTP.tsx
```diff
@@ -80,20 +77,23 @@ const TwoFactorTOTP = (props: ComponentProps<typeof Box>): ReactElement => {
...
-			const result = await verifyCodeFn(authCode);
+	const handleVerifyCode = useCallback(
+		async ({ authCode }) => {
+			try {
+				const result = await verifyCodeFn(authCode);
 
-			if (!result) {
-				return dispatchToastMessage({ type: 'error', message: t('Invalid_two_factor_code') });
-			}
+				if (!result) {
```
dougfabris, line 85: IMO we can remove this check, it's unnecessary.

dougfabris (changes requested):

apps/meteor/client/views/account/security/TwoFactorTOTP.tsx
juliajforesti: not really, `2fa:validateTempToken` method can return `undefined`

dougfabris (approved):

PR #29219, opened 2023-05-12
[FIX] app status inconsistencies when running multiple instances in a cluster

d-gubert:

App status inconsistencies between multiple instances in a cluster boil down to the fact that the Apps-Engine is currently responsible for orchestrating when these events are triggered and is overly verbose in doing so.

Upon analysis, the framework itself _should not_ have the concept of "other instances" - this is a deployment detail of the host system, and as such should be controlled by the host. The correct solution for this problem is to review this notification system, potentially removing it from the framework and leaving the responsibility solely for Rocket.Chat.

However, this is hindering the current app management experience for workspaces, so this PR cuts the control of some notifications that come from the framework (the more problematic ones) and moves the control over to RC in a short and practical way.

This is done by turning the methods of the most problematic events in the `AppActivationBridge` into no-ops, and instead triggering the `AppServerNotifier` directly in the api endpoints that are applicable.

It is _not_ the most correct solution to the problem, but due to time constraints and urgency this will be applied first so we can move with the correct solution in a future point.

## Steps to test or reproduce

1 - Setup a clustered deployment (6.x), either micro-services or high availability;
2 - Install an app on instance A
3 - App will not be available in other instances

### DIFF
--- a/apps/meteor/app/apps/server/bridges/activation.ts
+++ b/apps/meteor/app/apps/server/bridges/activation.ts
@@ -11,12 +11,19 @@ export class AppActivationBridge extends ActivationBridge {
 		super();
 	}
 
-	protected async appAdded(app: ProxiedApp): Promise<void> {
-		await this.orch.getNotifier().appAdded(app.getID());
+	protected async appAdded(_app: ProxiedApp): Promise<void> {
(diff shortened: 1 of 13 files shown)


AllanPazRibeiro (approved):
LGTM

PR #29221, opened 2023-05-12
chore: Remove Meteor.Streamer client usage

ggazzo:
[ARCH-1112]
the idea is to remove the old meteor-streamer package, and use the same interface that the new sdk uses.

We are not using the new sdk yet, we are using a wrapper implementing the same methods but based on meteor.

that way we can refactor all the code without worrying about  new bugs being introduced, then we can run our standalone application

## Further comments

[ARCH-1112]: https://rocketchat.atlassian.net/browse/ARCH-1112?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/meteor/client/views/meet/CallPage.tsx
```diff
@@ -71,36 +71,34 @@ const CallPage: FC<CallPageProps> = ({
...
-				visitorId,
-			);
-			Notifications.onRoom(roomId, 'webrtc', (type: any, data: any) => {
+			Notifications.onVisitor(WEB_RTC_EVENTS.WEB_RTC, visitorId, (type: any, data: any) => {
+				if (data.room == null) {
+					return;
+				}
+				webrtcInstance.onUserStream(type, data);
+			});
+			// TODO: I have no idea what this is doing, how this works, why it's here or how to type it
```
sampaiodiego, line 80: once this is merged nobody will know who "I" is.. maybe you should sign this comment?

btw, I don't see much relevance of having this comment in the code as well

apps/meteor/ee/app/canned-responses/client/streamer.js
sampaiodiego, line 1: delete file?

ggazzo: my bad

apps/meteor/client/views/meet/CallPage.tsx
ggazzo: just wanted to make clear why I inserted any instead o typing the stream. I

apps/meteor/ee/app/canned-responses/client/streamer.js
ggazzo: my bad thanks

sampaiodiego (approved):

PR #29227, opened 2023-05-12
regression: Reply in DM not working with hide_FlexTab

yash-rajpal:
## Further comments

TC-699

### DIFF
--- a/apps/meteor/client/components/message/toolbox/MessageActionMenu.tsx
+++ b/apps/meteor/client/components/message/toolbox/MessageActionMenu.tsx
@@ -26,6 +26,7 @@ const MessageActionMenu = ({ options, ...props }: MessageActionMenuProps): React
 		.map(({ color, ...option }) => ({
 			...option,
 			...(color === 'alert' && { variant: 'danger' as const }),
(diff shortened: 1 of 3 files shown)


apps/meteor/client/components/message/toolbox/MessageActionMenu.tsx
```diff
@@ -26,6 +26,7 @@ const MessageActionMenu = ({ options, ...props }: MessageActionMenuProps): React
 		.map(({ color, ...option }) => ({
 			...option,
 			...(color === 'alert' && { variant: 'danger' as const }),
+			...(option.id === 'reply-directly' && { role: 'link' as const }),
```
gabriellsh, line 29: Please, pass the role here `apps/meteor/app/ui-utils/client/lib/messageActionDefault.ts` Line: 28

gabriellsh (changes requested):

apps/meteor/client/components/message/toolbox/MessageActionMenu.tsx
yash-rajpal: Done! Thanks!

PR #29253, opened 2023-05-16
refactor: Move units check outside of model for finds

KevLehman:
OC-950

## Further comments

- Removes Units check from inside `find` and `findpaginated` functions.
- Since updates are async, having the extra filter inside is not a problem.
- Removes `Promise.await` introduced to fix a regression on 6.2


apps/meteor/ee/server/models/raw/LivechatRooms.ts
```diff
@@ -159,6 +168,7 @@ export class LivechatRoomsRawEE extends LivechatRoomsRaw implements ILivechatRoo
 		);
 	}
 
+	// TODO - check if removable, looks unused
```
murtaza98, line 171: I think you're right here. This indeed doesn't seem to be used anywhere. Perhaps we can remove it?

apps/meteor/ee/app/livechat-enterprise/server/lib/query.helper.js
```diff
@@ -18,13 +18,18 @@ export const addQueryRestrictionsToRoomsModel = (originalQuery = {}) => {
 	return query;
 };
 
-export const addQueryRestrictionsToDepartmentsModel = async (originalQuery = {}) => {
-	const query = { ...originalQuery, type: { $ne: 'u' } };
+export const restrictQuery = async (originalQuery = {}) => {
```
murtaza98, line 21: Is this still needed, considering you've already moved this logic to `applyDepartmentRestrictions.ts`

murtaza98 (approved):

murtaza98 (dismissed):

d-gubert (approved):
All good on apps side

ggazzo (approved):

d-gubert (approved):
Still good on apps side

aleksandernsilva (approved):

PR #29261, opened 2023-05-16
refactor: Big enterprise file to ts

KevLehman:
OC-980

## Further comments

There was a disconnection between the model types and endpoint types, so this PR closes that gap for units, departments, tags and slas.


packages/rest-typings/src/v1/omnichannel.ts
```diff
@@ -3008,7 +3008,21 @@ export type OmnichannelEndpoints = {
...
+				description?: string | undefined;
+				showOnRegistration: boolean;
+				email: string;
+				showOnOfflineForm: boolean;
+				requestTagBeforeClosingChat?: boolean | undefined;
+				chatClosingTags?: string[] | undefined;
+				fallbackForwardDepartment?: string | undefined;
+				departmentsAllowedToForward?: string[] | undefined;
+			};
+			agents: any[];
```
ggazzo, line 3024: 👀

packages/rest-typings/src/v1/omnichannel.ts
```diff
@@ -3008,7 +3008,21 @@ export type OmnichannelEndpoints = {
...
+				name: string;
+				description?: string | undefined;
+				showOnRegistration: boolean;
+				email: string;
+				showOnOfflineForm: boolean;
+				requestTagBeforeClosingChat?: boolean | undefined;
+				chatClosingTags?: string[] | undefined;
+				fallbackForwardDepartment?: string | undefined;
+				departmentsAllowedToForward?: string[] | undefined;
+			};
```
ggazzo, line 3023: i think this is pretty useful in other parts of the code, maybe it should became a type

btw I'm missing the validator

ggazzo (dismissed):

packages/rest-typings/src/v1/omnichannel.ts
KevLehman: Good one, forgot to centralize it. 

On the validator, yeah 😬 the endpoint has some "custom" behavior of adding validators to the custom fields that I haven't looked at how to do with ajv without the "additionalProps": true, cause that allows everything 🤔 so, that's why there's no validator for now

murtaza98 (approved):

PR #29264, opened 2023-05-17
refactor: `AppsProvider`

tiagoevanp:
## Further comments

[AECO-206]

[AECO-206]: https://rocketchat.atlassian.net/browse/AECO-206?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ggazzo (changes requested):
image
I would suggest another folder structure

/providers
	Appsprovider.tsx
	/lib
		appsReducer.ts -> good candidate to have unit tests
		registerLicence -> I dont think this file is really necesssary, I think you just removed from the provider because the code was too big but since you dont want to use in a different place, maybe we should not allow it be imported

ggazzo (approved):

PR #29274, opened 2023-05-17
refactor: Convert Omnichannel helper ee to ts

KevLehman:

### DIFF
--- a/apps/meteor/app/livechat/server/lib/RoutingManager.js
+++ b/apps/meteor/app/livechat/server/lib/RoutingManager.js
@@ -196,7 +196,8 @@ export const RoutingManager = {
 
 		if (!agent) {
 			logger.debug(`Cannot take Inquiry ${inquiry._id}: Precondition failed for agent`);
-			return callbacks.run('livechat.onAgentAssignmentFailed', { inquiry, room, options });
+			await callbacks.run('livechat.onAgentAssignmentFailed', { inquiry, room, options });
+			return;
 		}
 
 		await LivechatInquiry.takeInquiry(_id);
--- a/apps/meteor/ee/app/livechat-enterprise/server/lib/Helper.ts
+++ b/apps/meteor/ee/app/livechat-enterprise/server/lib/Helper.ts
@@ -8,6 +8,8 @@ import {
 	Users,
 } from '@rocket.chat/models';
 import { api } from '@rocket.chat/core-services';
+import type { Document } from 'mongodb';
+import type { ILivechatInquiryRecord, IOmnichannelRoom, IOmnichannelServiceLevelAgreements } from '@rocket.chat/core-typings';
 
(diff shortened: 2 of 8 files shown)


apps/meteor/ee/app/livechat-enterprise/server/lib/Helper.ts
```diff
@@ -111,34 +138,34 @@ const dispatchWaitingQueueStatus = async (department) => {
 
 	const queueInfo = await getQueueInfo(department);
 	queue.forEach((inquiry) => {
-		dispatchInquiryPosition(inquiry, queueInfo);
+		void dispatchInquiryPosition(inquiry, queueInfo);
```
murtaza98, line 141: Considering this is an async function, WDYT about using Promise.all() to make this entire function sync?

apps/meteor/app/livechat/server/lib/RoutingManager.js
```diff
@@ -196,7 +196,8 @@ export const RoutingManager = {
 
 		if (!agent) {
 			logger.debug(`Cannot take Inquiry ${inquiry._id}: Precondition failed for agent`);
-			return callbacks.run('livechat.onAgentAssignmentFailed', { inquiry, room, options });
+			await callbacks.run('livechat.onAgentAssignmentFailed', { inquiry, room, options });
```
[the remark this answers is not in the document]
murtaza98: The value from the callback won't get passed to `forwardRoomToDepartment` in this case since you're not returning it again in this method. Perhaps you can change it to:

```
const value = await callbacks.run('livechat.onAgentAssignmentFailed', { inquiry, room, options });
return value;
```

apps/meteor/ee/app/livechat-enterprise/server/lib/Helper.ts
murtaza98: Makes sense :+1:

murtaza98 (approved):

.changeset/modern-coins-know.md
```diff
@@ -0,0 +1,5 @@
+---
+"@rocket.chat/meteor": patch
+---
+
+refactor: Convert Omnichannel helper ee to ts
```
sampaiodiego, line 5: refactor usually shouldn't go to a change log.. but if you want to have it there, it is fine =)

sampaiodiego (approved):

.changeset/modern-coins-know.md
KevLehman: All right 🤔 didn't know that. We have other refactors with changesets lol

sampaiodiego: changesets are customer facing (or end user) release notes, so that's why **usually** refactors don't deserve a changeset 😉

PR #29290, opened 2023-05-18
fix: Omnichannel queue not running for all queues

KevLehman:
## Further comments

Fix Public queue running even when no room is assigned to it. 
Even when this doesn't produce any noticeable effect, causes logs to always show that inquiries are running on "public" queue instead of actual department queue.


murtaza98 (dismissed):

KevLehman:
Skipping QA as there's no actual way of QAing 😬

murtaza98 (approved):

murtaza98 (approved):

ggazzo (approved):

PR #29307, opened 2023-05-20
fix: Member list is randomly sorted 

Keanby:


debdutdeb (changes requested):
Should return a sorted list from the be already, configurable by some sort param.

Please check many of our other endpoint code that accepts a sort parameter which then returns a sorted list based on the key to the client.

apps/meteor/client/views/room/contextualBar/RoomMembers/RoomMembersWithData.tsx
```diff
@@ -95,6 +95,20 @@ const RoomMembersWithData = ({ rid }: { rid: IRoom['_id'] }): ReactElement => {
 	if (state.tab === ROOM_MEMBERS_TABS.ADD) {
 		return <AddUsers rid={rid} onClickBack={handleBack} reload={reload} />;
 	}
```
KevLehman, line 97: Shouldn't this sort be made on server?

KevLehman (changes requested):
See comments

KevLehman:
Thanks! I'll close this one in favor of #29228 :)

PR #29313, opened 2023-05-22
fix: message deletion fails if has files attached on filesystem storage

### LINKED ISSUE #29266: Application Error When Deleting uploaded Files After upgrade to 6.2.0 (FileSystem Type)
### Description:

Application Error When Deleting uploaded Files After upgrade to 6.2.0

### Steps to reproduce:

1. Go to Direct Message
2. upload a file and try to delete on
3. Refresh your tab and you can see you file deleted succesfully but in file menu you can see the file name and you cant delete it

check this video:
https://mega.nz/file/KjJyERaA#jQyzvcl_MMYLqH3nMpB2RDesYX00JzQpRGsm88EQzKY

### Server Setup Information:

- Version of Rocket.Chat Server: 6.2.0
- Operating System: Rocky linux 9
- Deployment Method: Manually
- DB Replicaset Oplog: 
- NodeJS Version: 14.21.3
- MongoDB Version:6.0.5 / wiredTiger (oplog Enabled)

### LINKED ISSUE #29292: Unable to delete users
### Description:

This is a stub for the issue where deleting users fails using admin web console.

The issue started in RC 6.2.x.

Additional conditions to reproduce the issue are still unknown at the time of writing this.

### Steps to reproduce:

1. Go to Administration -> Users
2. Choose an existing user
3. Choose Delete and confirm the deletion in the popup

### Expected behavior:

The user account gets deleted.

### Actual behavior:

A red notification consisting of word `ENOENT` flashes at the top right corner after pressing `Delete` in the confirmation popup. User account does not get deleted.

### Server Setup Information:

- Version of Rocket.Chat Server: 6.2.1
- Operating System: RHEL8
- Deployment Method: tar
- DB Replicaset Oplog: Yes
- NodeJS Version: 12.21.3
- MongoDB Version: 5.0

### Client Setup Information

- Desktop App or Browser Version: The latest Chromium

### Additional context

This is likely to be related to the code that checks and assigns new owners to the channels owned by the user.

Deleting any newly created users does not trigger the issue, nor does it happen with every single old user either.

Tested with a clean browser session (private window).

### Relevant logs:

...

debdutdeb:
## Issue(s)

Closes #29292 
Closes #29266 

## Further comments

Introduced in https://github.com/RocketChat/Rocket.Chat/pull/28769

(Store).removeById calls implemented (Store).delete, which, in case of filesystem store calls (Store).removeById again ending up with stat() failing as file was already deleted.


.changeset/friendly-spies-retire.md
```diff
@@ -0,0 +1,5 @@
+---
+"@rocket.chat/meteor": minor
```
debdutdeb, line 2:
```suggestion
"@rocket.chat/meteor": patch
```

debdutdeb:
btw - I would avoid the 2nd call instead of wrapping in try catch, but that will require modifying removeById and I don't know enough about this part of the codebase to make that decision, might break it someplace else where it expects the code to run.

the distinction in my head for remove vs delete keywords is that delete works on actuak file removal, remove whereas removing a record from database. but removeById already handled a part of actual deletion, some temporary file thing, I dont know what that is. 

Avoiding the second call is the better way IMO, as when removing a room heavy in attachments, the second batch of stat calls can be expensive and unnecessary.

apps/meteor/server/ufs/ufs-local.ts
```diff
@@ -69,11 +69,14 @@ export class LocalStore extends Store {
...
-				await unlink(path);
-				await this.removeById(fileId);
+			try {
+				const statResult = await stat(path);
+				if (statResult?.isFile()) {
+					await unlink(path);
+					await this.removeById(fileId);
+				}
+			} catch (_e) {
+				// noop
```
KevLehman, line 79: Is it good enough to silently fail? Wouldn't be better to update the callers of `this.delete` to do something if an error happens? Cause this way, there's no current indication of the file being unexistent, not even a log 👀

sampaiodiego (commented):
should we cover this with tests?

debdutdeb:
@sampaiodiego since this is dependent on upload store, do you think it'd make more sense to cover tests for that instead? this affects user deletion, avatar change too, room deletion, anything that requires deleting a file from file system. we'd have to cover all of that other wise. Idk 🙈 need you help

sampaiodiego:
hmmm.. I was wondering that we already have tests for most (maybe all) of those use cases, but we test only with the default GridFS file storage.. so ideally we need to expand that to every file storage supported, which might be tricky.

I'll gonna merge this for now, but pls let's not lose track of these missing tests..

sampaiodiego (approved):

Issue #29312, opened 2023-05-22
The Phone number does not have validation check in omnichannel 


Ykumar1415:
Description:
When we create a new contact, the phone section does not have a limit for the length of phone numbers; it also accepts the alphabet. 

Steps to reproduce:
Go to the omnichannel contact center
Click on the new contact
Expected behavior:
It should check the format of the phone number whether it is valid or not.

Actual behavior:

https://github.com/RocketChat/Rocket.Chat/assets/97088265/cd5801cc-676b-4f88-9a3b-044f750a7e0e

Server Setup Information:
Rocket.Chat Server: 6.3.0-develop
Operating System: Ubuntu
Deployment Method: Gitpod
Number of Running Instances: 1
DB Replicaset Oplog: Enabled
NodeJS Version: 14.21.3 
MongoDB Version:6.0.3

Ykumar1415:
I am currently working on this issue.

anikdhabal:
This issue has been created before #20528

debdutdeb:
@Ykumar1415 please check first if there is any existing open issue before creating another one.

debdutdeb:
Duplicate of https://github.com/RocketChat/Rocket.Chat/issues/20528

debdutdeb:
closing in favour of #20528

Ykumar1415:
> @Ykumar1415 please check first if there is any existing open issue before creating another one. You can react with a 👍🏼 to any issue to show your support, or add a comment in that issue.

sorry sir, but due to not having any search box option this happened I searched for this issue in the good first issue but it was not there so I created a new one.

Ykumar1415:
Should i make a pull request again  referring to that issue number

anikdhabal:
One PR has been created before #29101.Hopefully I will fix it by today.

PR #29315, opened 2023-05-22
feat: omnichannel phone number validation

Ykumar1415:
Proposed changes (including video):

https://github.com/RocketChat/Rocket.Chat/assets/97088265/511a62cf-5c39-4698-bada-7cc9f4640e8d

Issue(s):
The Phone number does not have validation check in omnichannel #29312

Steps to test:
Go to the omnichannel contact center
Click on the new contact
It will check the format of the phone number whether it is valid or not.

Further comments:
1. When the Phone number is already present or not checks in a function then 
2. The Phone Number is valid or not is also will be checked. 
3. - the Phone Number can contain only numbers and country code . and length should be limited.


debdutdeb:
Just noting here there is already another pr from the community open for this - if we don't get any response for 1-2 weeks, we can move on with this pr then :) 

https://github.com/RocketChat/Rocket.Chat/pull/29101

cc @murtaza98

KevLehman:
Closing this as we'll go with the other PR when we're ready for this change

Thanks for your contribution:)

PR #29318, opened 2023-05-22
chore: replace meteor.call -> sdk.call

ggazzo:


apps/meteor/app/message-mark-as-unread/client/actionButton.ts
```diff
@@ -6,29 +6,31 @@ import { messageArgs } from '../../../client/lib/utils/messageArgs';
...
 		label: 'Mark_unread',
 		context: ['message', 'message-mobile', 'threads'],
-		action(_, props) {
+		async action(_, props) {
 			const { message = messageArgs(this).msg } = props;
-			return Meteor.call('unreadMessages', message, async function (error: unknown) {
-				if (error) {
+			return sdk
+				.call('unreadMessages', message)
+				.then(async () => {
```
sampaiodiego, line 21: since you change `action` to `async`, why are you using `then` instead of `await`?

sampaiodiego (changes requested):

sampaiodiego (approved):

KevLehman (approved):

PR #29323, opened 2023-05-22
fix: Add missing awaits to .count() calls

KevLehman:


murtaza98 (approved):

KevLehman:
Skipping QA as almost all the missing awaits are used in logs

PR #29356, opened 2023-05-24
fix: Register Username form showing before homepage

hugocostadev:

Meteor Minimongo Collection was causing delays in populating user data in the front end, causing layout updates with different rules (the username was not being sent to UI the first time sending the user to the username registration form).

To avoid the incomplete stream from Meteor, I used the REST endpoint with the existing hook to get user info to get the full user data and check whether it has a username or not.

## Steps to test or reproduce

1. Go to localhost:3000 
2. Open the Network tab and select the Throttling option to a slower network
3. Try to log in as a normal user
4. You should see the Username registration for a moment, if not, try again until happens

## Further comments

WM-46


rique223 (approved):
LGTM!

PR #29357, opened 2023-05-24
fix: Remove room from UI when another agent takes it

KevLehman:
## Issue(s)

[OC-680](https://rocketchat.atlassian.net/browse/OC-680)
OC-873
[OC-1025](https://rocketchat.atlassian.net/browse/OC-1025)

## Further comments

[OC-680]: https://rocketchat.atlassian.net/browse/OC-680?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/meteor/app/livechat/server/hooks/notifyInquiryTaken.ts
+++ b/apps/meteor/app/livechat/server/hooks/notifyInquiryTaken.ts
@@ -1,0 +1,59 @@
+import type { Filter } from 'mongodb';
+import { LivechatDepartmentAgents, LivechatRooms, Users } from '@rocket.chat/models';
+import { api } from '@rocket.chat/core-services';
+import { UserStatus } from '@rocket.chat/core-typings';
+import type { IUser } from '@rocket.chat/core-typings';
+
+import { callbacks } from '../../../../lib/callbacks';
+
+callbacks.add(
+	'livechat.afterTakeInquiry',
+	async (inquiry) => {
+		const { rid } = inquiry;
+		const room = await LivechatRooms.findOneById(rid);
+
+		if (!room) {
+			return inquiry;
+		}
+
+		const { servedBy, departmentId } = room;
+
+		if (!servedBy?._id) {
+			return inquiry;
+		}
+
+		const extraQuery: Filter<IUser> = {
+			_id: { $nin: [servedBy._id] },
+		};
+
+		if (departmentId) {
+			extraQuery._id = {
+				$in: (await LivechatDepartmentAgents.findByDepartmentIds([departmentId], { projection: { agentId: 1 } }).toArray()).map(
+					(d) => d.agentId,
+				),
+				$nin: [servedBy._id],
+			};
+		}
+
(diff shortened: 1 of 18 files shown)


aleksandernsilva (approved):

apps/meteor/server/models/raw/Users.js
```diff
@@ -2875,4 +2875,27 @@ export class UsersRaw extends BaseRaw {
...
+		}
+
+		return this.find(
+			{
+				roles: {
+					$all: ['livechat-agent'],
+					$nin: ['livechat-manager', 'livechat-monitor'],
+				},
+				statusLivechat: 'available',
+				status: 'online',
```
murtaza98, line 2895: Shouldn't we also respect `Livechat_enabled_when_agent_idle` setting over here?

KevLehman: Hmmmm, we can remove this filter as all we need is the "available" thing

murtaza98 (dismissed):

aleksandernsilva (approved):

murtaza98 (approved):

apps/meteor/app/livechat/server/hooks/notifyInquiryTaken.ts
```diff
@@ -0,0 +1,50 @@
...
+			void api.broadcast('omnichannel.events', user._id, {
+				event: 'InquiryTaken',
+				data: {
+					inquiryId: inquiry._id,
+					roomId: room._id,
+					agentId: servedBy._id,
+				},
+			});
+		}
+	},
```
ggazzo, line 47: I think this should/could be done inside the `listener.modules`

apps/meteor/server/modules/listeners/listeners.module.ts
```diff
@@ -344,6 +344,10 @@ export class ListenersModule {
 			notifications.notifyUserInThisInstance(userId, 'voip.events', data);
 		});
 
+		service.onEvent('omnichannel.events', (userId, data): void => {
+			notifications.notifyUserInThisInstance(userId, 'omnichannel.events', data);
+		});
```
ggazzo, line 349: I'm not sure informing every agent online (by user) about on inquiry being taken this is most performant approach.

If I'm not mistaken we already have the `roomData` that will be triggered and only who open the channel will receive

apps/meteor/app/livechat/server/hooks/notifyInquiryTaken.ts
KevLehman: And i was trying to not to add custom logic inside that file 😬

apps/meteor/server/modules/listeners/listeners.module.ts
sampaiodiego: agreed.. that would be much better to have some logic applied on on client side when data is received through the `room-data` stream

PR #29373, opened 2023-05-25
feat: add support for `MONOLITH_TRANSPORTER` env var

sampaiodiego:


rodrigok (dismissed):

apps/meteor/ee/tests/unit/apps/meteor/ee/server/local-services/instance/getTransporter.spec.ts
```diff
@@ -0,0 +1,25 @@
+import { expect } from 'chai';
```
ggazzo, line 1: I think the unit tests should live side by side with the files it calls `colocating`

ggazzo (changes requested):

.changeset/mean-guests-love.md
```diff
@@ -0,0 +1,11 @@
+---
+'@rocket.chat/meteor': minor
+---
+
+feat: *Enterprise* Add support for different transporters to connect multiple monolith instances.
+
+To use that, you can use the `TRANSPORTER` env var adding "monolith+" to the transporter value. To use NATS for example, your env var should be:
+
+```bash
+export TRANSPORTER="monolith+nats://localhost:4222"
```
geekgonecrazy, line 10: Is there a real downside to just making it use transporter just like normal?  This means if someone deploys this then wants to move to microservices they’ve got to now remove that instead of add microservices

debdutdeb: yeah I agree with Aaron

debdutdeb:
this I believe will break at addOfflineNode btw .. nats transporter doesn't have an addOfflineNode method. which makes me think this won't respect instance_ip at all.

debdutdeb:
i'll give it a try now

debdutdeb (approved):

ggazzo (approved):

PR #29375, opened 2023-05-25
chore: Change ddpsdk.stream to return subscription

ggazzo:


jeanfbrito (approved):

packages/ddp-client/__tests__/Connection.spec.ts
```diff
@@ -13,60 +14,44 @@ afterEach(() => {
 	WS.clean();
 });
 
+// 87,92,105,134,142,166
```
ggazzo, line 17:
```suggestion
```

jeanfbrito (dismissed):

PR #29381, opened 2023-05-27
Fix login terms issue#28586

Irtiza751:


debdutdeb:
Hi @Irtiza751 , thanks for the PR, but unfortunately this is not the right fix.

A PR is already open with the correct fix - https://github.com/RocketChat/Rocket.Chat/pull/28999

I'll close this one in favour of that one.

Please, if you have any questions regarding the PR, feel free to leave a question there.

PR #29393, opened 2023-05-29
feat: new moderation setting to enable trust roles (WM-74)

Dnouv:

This PR introduces trust/reputation roles. To make it more customizable, there is a boolean setting, `Trust Roles`, once enabled it creates two new locked roles with varying permissions to the workspace; this is to be used by the RC Apps to provide different levels of services/privileges to the users, based on their trust/reputation roles.
The Reputation App will manage the award/deduction of reputation points.

The last two columns are the new roles
image

## Further comments

JIRA: [WM-74](https://rocketchat.atlassian.net/browse/WM-74)

[WM-74]: https://rocketchat.atlassian.net/browse/WM-74?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/meteor/app/moderation/lib/permissions.ts
```diff
@@ -0,0 +1,23 @@
...
+export const novice: IPermission['_id'][] = [
+	'view-joined-room',
+	'preview-c-room',
+	'leave-c',
+	'join-without-join-code',
+	'view-joined-room',
+	'view-history',
+];
+
+export const explorer: IPermission['_id'][] = [
```
debdutdeb, line 12: can you share the role => permissions on open for review? Purpose of a role and given permissions.

debdutdeb (dismissed):
We need to make this a service - this is not how we should create these anymore. Let's catch up tomorrow.

the meteor stuff is also not required

apps/meteor/app/moderation/lib/permissions.ts
Dnouv: Sure, I will do that. Thank you!

hugocostadev (dismissed):

tapiarafael (dismissed):

apps/meteor/server/services/moderation/service.ts
```diff
@@ -0,0 +1,48 @@
...
+
+import { createOrUpdateProtectedRoleAsync } from '../../lib/roles/createOrUpdateProtectedRole';
+
+export class ModerationService extends ServiceClassInternal implements IModerationService {
+	protected name = 'moderation';
+
+	async resetUserRoles(roles: string[]): Promise<void> {
+		const usersInRole = await Users.findUsersInRoles(roles, undefined, { projection: { _id: 1 } }).toArray();
+
+		const usersToRemove = usersInRole.map((user) => user._id);
```
debdutdeb, line 13: you can call map on the cursor directly

apps/meteor/server/models/raw/Roles.ts
```diff
@@ -48,6 +48,42 @@ export class RolesRaw extends BaseRaw<IRole> implements IRolesModel {
...
+		if (!Array.isArray(roles)) {
+			roles = [roles];
+			console.warn('[WARN] Roles.addRolesByUserIds: roles should be an array');
+		}
+
+		const users = await Users.find({ _id: { $in: userIds } }, { projection: { _id: 1 } }).toArray();
+
+		const userIdsToAddRoles = users.map((user) => user._id);
+
+		for await (const roleId of roles) {
```
debdutdeb, line 66: roles here isn't an async iterartor

apps/meteor/server/models/raw/Roles.ts
```diff
@@ -100,6 +136,30 @@ export class RolesRaw extends BaseRaw<IRole> implements IRolesModel {
 		return true;
 	}
 
+	async removeUsersRoles(uids: IUser['_id'][], roles: IRole['_id'][], scope?: IRoom['_id']): Promise<boolean> {
+		if (process.env.NODE_ENV === 'development' && (scope === 'Users' || scope === 'Subscriptions')) {
+			throw new Error('Roles.removeUserRoles method received a role scope instead of a scope value.');
+		}
+
+		for await (const roleId of roles) {
```
debdutdeb, line 144: not async iterator or generator

apps/meteor/server/models/raw/Subscriptions.ts
```diff
@@ -222,6 +222,21 @@ export class SubscriptionsRaw extends BaseRaw<ISubscription> implements ISubscri
...
 	}
 
+	removeRolesByUserIds(uids: IUser['_id'][], roles: IRole['_id'][], rid: IRoom['_id']): Promise<Document | UpdateResult> {
+		const query = {
+			'u._id': { $in: uids },
+			rid,
+		};
+
+		const update = {
+			$pullAll: {
```
debdutdeb, line 232: $pull is enough, roles is a set by operation.

apps/meteor/server/models/raw/Subscriptions.ts
```diff
@@ -270,6 +285,31 @@ export class SubscriptionsRaw extends BaseRaw<ISubscription> implements ISubscri
...
+	async addRolesByUserIds(userIds: IUser['_id'][], roles: IRole['_id'][], roomId?: IRoom['_id']): Promise<Document | UpdateResult> {
+		if (!Array.isArray(userIds)) {
+			userIds = [userIds];
+			console.warn('[WARN] Subscriptions.addRolesByUserIds: userIds should be an array');
+		}
+
+		if (!Array.isArray(roles)) {
+			roles = [roles];
+			console.warn('[WARN] Subscriptions.addRolesByUserIds: roles should be an array');
+		}
```
debdutdeb, line 297: same comment as before

apps/meteor/server/services/moderation/service.ts
```diff
@@ -0,0 +1,48 @@
...
+
+		// remove the roles from the users
+		await Roles.removeUsersRoles(usersToRemove, roles);
+
+		// add the default 'user' role to the users
+		await Roles.addRolesByUserIds(usersToRemove, ['user']);
```
debdutdeb, line 19: not all workspace defaults to user role. this should be skipped. as when the trust roles are added, we're not unsetting the default role either. upto the admin to change the default role.

debdutdeb (changes requested):
I see a disconnection between part of the intent behind some of the code and the rest.

The service methods seem partially only intended for trust roles, then also not. 

for example, resetRoles method, by signature it's only intended for global roles, therefore no need to use the roles model. You can just use the users model.

On the other hand, by implementation, partially looks like you're trying to make the methods generic as in "reset any roles", but not handling room scoped roles well. 

I suggest not trying anything generic here and just focusing on the global trust roles. Thus, 
1. avoiding the roles model changes (not to mention the methods currently added should be in a service rather)
2. avoiding subscription model changes
3. moving most code in the moderation service
4. using just the users model

this will simplify the pr quite a lot. You'll need two methods, `resetTrustRoles` or `removeTrustRoles` and `addTrustRoles`.

apps/meteor/server/models/raw/Roles.ts
```diff
@@ -48,6 +48,42 @@ export class RolesRaw extends BaseRaw<IRole> implements IRolesModel {
...
+
+		const users = await Users.find({ _id: { $in: userIds } }, { projection: { _id: 1 } }).toArray();
+
+		const userIdsToAddRoles = users.map((user) => user._id);
+
+		for await (const roleId of roles) {
+			const role = await this.findOneById<Pick<IRole, '_id' | 'scope'>>(roleId, { projection: { scope: 1 } });
+
+			if (!role) {
+				console.warn(`[WARN] Roles.addRolesByUserIds: role: ${roleId} not found`);
```
debdutdeb, line 70: only on development

apps/meteor/server/models/raw/Users.js
```diff
@@ -1090,6 +1108,24 @@ export class UsersRaw extends BaseRaw {
...
+	 * @param {string[]} uids the list of user ids to remove roles from
+	 * @param {IRole['_id']} roles the list of role ids to remove
+	 */
+	removeRolesByUserIds(uids, roles) {
+		const query = {
+			_id: { $in: uids },
+		};
+
+		const update = {
+			$pullAll: {
```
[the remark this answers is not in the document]
Dnouv: I think `pullAll` would be good enough to avoid making the query more complex.

apps/meteor/server/models/raw/Roles.ts
```diff
@@ -48,6 +48,42 @@ export class RolesRaw extends BaseRaw<IRole> implements IRolesModel {
...
+		}
+
+		if (!Array.isArray(roles)) {
+			roles = [roles];
+			console.warn('[WARN] Roles.addRolesByUserIds: roles should be an array');
+		}
+
+		const users = await Users.find({ _id: { $in: userIds } }, { projection: { _id: 1 } }).toArray();
+
+		const userIdsToAddRoles = users.map((user) => user._id);
```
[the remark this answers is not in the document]
Dnouv: No, we are just creating `userIdsToAddRoles=['id1', 'id2']` from `users=[{_id: "id1"}, {_id: "id2"}]`.

apps/meteor/server/models/raw/Roles.ts
```diff
@@ -100,6 +136,30 @@ export class RolesRaw extends BaseRaw<IRole> implements IRolesModel {
...
+			switch (role.scope) {
+				case 'Subscriptions':
+					scope && (await Subscriptions.removeRolesByUserIds(uids, [roleId], scope));
+					break;
+				case 'Users':
+				default:
+					await Users.removeRolesByUserIds(uids, [roleId]);
+			}
+		}
+		return true;
```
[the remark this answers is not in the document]
Dnouv: To keep this new method consistent with the other methods, you see all other similar methods are returning a boolean, so that's why I did not change it.

apps/meteor/server/services/moderation/service.ts
```diff
@@ -0,0 +1,48 @@
...
+			});
+
+			// add the permissions to the role
+			const promises = permissions.map((permission) => Permissions.addRole(permission, roleName));
+			await Promise.all(promises);
+		} catch (error) {
+			console.error(`Error adding permissions to role ${roleName}: ${error}`);
+		}
+	}
+}
```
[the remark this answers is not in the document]
Dnouv: Can we use `Permissions.insert...` here inside the service?

apps/meteor/server/models/raw/Roles.ts
Dnouv: Yes, it is not, but the `await` keyword is necessary in this case to ensure that the loop waits for each asynchronous operation to complete before moving on to the next iteration.

apps/meteor/server/services/moderation/service.ts
Dnouv: The app will unset the default roles when installed; if it is not done, then the system won't work because usually, the permissions check is done considering the highest level role.
For example, if a user has two roles, `novice` and `users`, then there won't be any restriction, right? Because when checking for permission, the `users` will be given the preference. 
Please correct me if I misunderstood something. Thank you!

apps/meteor/server/models/raw/Roles.ts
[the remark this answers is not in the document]
debdutdeb: That's not what I'm saying. The user ids are passed to the method to begin with. Why do a find on the same ids again?

apps/meteor/server/models/raw/Roles.ts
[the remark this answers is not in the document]
debdutdeb: I'll have to check that then, but here it doesn't make any sense. No matter what it'll always return true if iirc, otherwise it'll jump out of the stack.

apps/meteor/server/services/moderation/service.ts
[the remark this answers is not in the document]
debdutdeb: Don't see a reason to not to.

apps/meteor/server/models/raw/Roles.ts
debdutdeb: But you are awaiting each promise inside the loop anyway. How does this benefit us, while it'll make the process slower because of repeated unwrapping.

apps/meteor/server/services/moderation/service.ts
Dnouv: Sorry, I did not understand this comment; what would this do?

apps/meteor/server/services/moderation/service.ts
debdutdeb: Ok. But user is not always the default role. You need to reset to the default role and not to "user" specifically.

apps/meteor/server/models/raw/Roles.ts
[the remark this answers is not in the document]
Dnouv: Just to make sure, the ids provided are valid, and there exist such userIds.

apps/meteor/server/models/raw/Roles.ts
Dnouv: It won't work without it, I already tried. :)

apps/meteor/server/services/moderation/service.ts
debdutdeb: When the setting is disabled, it'll then be safe to reset roles to present default ones.

Dnouv: Ok, I removed the part which adds the `users` role.

hugocostadev (dismissed):

debdutdeb (dismissed):

ggazzo (changes requested):
Looking at the changes, I was a little confused about some decisions, after I read the ADR and it seems to me that all this was done so some app, given some decision, can change the user's roles.


Here are my comments and then my suggestions.

I don't see the need to use listeners on the backend. like it or not this is yet another callback on the system and potentially modifying large amounts of user.

I don't understand the need to create a service. Services are cool, but each service adds an overhead and the whole system becomes aware of this new service, this is all because of 2 methods that won't be called on a regular basis.

I don't like to use frontend events in the global scope. this forces loading unnecessary stuff at unnecessary times.


That said, here are my suggestions on how I would implement

First step:
I would simply create the new permissions like any other, no settings, no listeners.
This saves a lot of code and avoids a lot of things happening.


Second: Through an API create a quick method for promoting/demoting multiple users. Instead of deactivating the feature (which currently doesn't exist, every smart part is an app), create an endpoint that allows you to modify a list of users at once. The same could be accessible by the apps engine.


Just that.

If there are no apps intercepting and modifying users, nothing but two new permissions happen.

apps/meteor/server/services/moderation/service.ts
```diff
@@ -0,0 +1,74 @@
...
+import { explorerPermissions as explorerExplicitPermissions, novicePermissions, trustRoles } from '../../../app/moderation/lib/permissions';
+import { Logger } from '../../lib/logger/Logger';
+
+export class ModerationService extends ServiceClassInternal implements IModerationService {
+	protected name = 'moderation';
+
+	private logger = new Logger('Moderation');
+
+	async removeTrustRoles(): Promise<void> {
+		const role = await Roles.findOneById(trustRoles[0], { projection: { _id: 1 } });
```
KevLehman, line 15: Instead of `trustRoles[0]` why not making trustRoles something like an enum or so? That way you can use trustRoles.XXX instead of a hard coded number

apps/meteor/server/services/moderation/service.ts
```diff
@@ -0,0 +1,74 @@
...
+		if (!role) {
+			return;
+		}
+		// remove all role mentions from user records that have it
+		// TODO: move to models
+		try {
+			await Users.updateMany({ roles: { $in: trustRoles } }, { $pullAll: { roles: trustRoles as unknown as string[] } });
+			await Permissions.updateMany({ roles: { $in: trustRoles } }, { $pullAll: { roles: trustRoles as unknown as string[] } });
```
KevLehman, line 23: These queries should be on model :)

apps/meteor/server/services/moderation/service.ts
```diff
@@ -0,0 +1,74 @@
...
+					},
+					{ $addToSet: { roles: { $each: trustRoles as unknown as string[] } } },
+				),
+				Permissions.updateMany(
+					{
+						_id: { $in: explorerExplicitPermissions },
+					},
+					{ $addToSet: { roles: 'explorer' } },
+				),
+			]);
```
KevLehman, line 69: Same with this queries

apps/meteor/server/services/moderation/service.ts
```diff
@@ -0,0 +1,74 @@
...
+			return;
+		}
+
+		try {
+			await Promise.all([
+				/*
+				 * explorer has all permissions novice does
+				 * and more, so we write to the same records only once
+				 */
+				Permissions.updateMany(
```
KevLehman, line 57: I know this is the case for this current scenario, but what happens if you later want to add 10 more "trust roles"? Will you do this update for each of them? 👀 

Why not using something like an object `{ [k: role]: permissions[] }` ?

KevLehman (changes requested):

apps/meteor/server/services/moderation/service.ts
debdutdeb: maybe .. yeah. at the end of the day it would check for any one roles, be it indexed or an enum.

i agree with Gazzo's recommendations though .. so I'll leave this one unresolved and wait for Devanshu to come back online 😬

apps/meteor/server/services/moderation/service.ts
debdutdeb: not sure what you mean. in more role case definitely wouldn't write all updates manually, but in this case there were just two, and this was just easier to read.

apps/meteor/server/services/moderation/service.ts
Dnouv: Thank you so much for the review. I agree, will move it!

hugocostadev (dismissed):

debdutdeb:
Btw - this is on hold afaik.

Please correct me if Im wrong @Dnouv

Dnouv:
Thank you, Hugo, for approving the changes. 

Yes, this is currently in Product and UX discussion.

PR #29401, opened 2023-05-30
fix: Importer crashes when sending the "active status" e-mail notification to users

matheusbsilva137:

 - Fix `getImportProgress` endpoint returning incorrect count info (`total: 0, completed: 0`);
 - Fix server crash when sending the "active status" e-mail notification to users on import.

## Further comments

TC-727


gabriellsh (approved):
Frontend LGTM!

apps/meteor/app/lib/server/functions/setUserActiveStatus.ts
```diff
@@ -115,6 +115,9 @@ export async function setUserActiveStatus(userId: string, active: boolean, confi
 	if (!active && !settings.get('Accounts_Send_Email_When_Deactivating')) {
 		return true;
 	}
+	if (!user.emails) {
```
sampaiodiego, line 118: should this be a short circuit for all other empty possibilities as well? like:

```suggestion
	if (!user.emails || !Array.isArray(user.emails) || user.emails.length === 0) {
```

if so, the `Array.isArray` can be removed from line 123

sampaiodiego (changes requested):

sampaiodiego (approved):

sampaiodiego (approved):

gabriellsh (approved):
Not sure why it's pending a FE review, but here you go

PR #29406, opened 2023-05-30
refactor: convert routing manager to ts

KevLehman:
## Further comments

OC-1020


packages/core-typings/src/omnichannel/routing.ts
```diff
@@ -12,7 +12,8 @@ export type RoutingMethodConfig = {
 	autoAssignAgent: boolean;
 };
 
-export type SelectedAgent = { agentId: string; username: string };
+export type SelectedAgent = { agentId?: string; username?: string };
```
ggazzo, line 15: is that possible? having an agent with no id or username?

ggazzo (changes requested):

packages/core-typings/src/omnichannel/routing.ts
KevLehman: It's "technically" possible, as some places return `{}` or `null` as the agent. But for the case it matters here, at least `agentId` is required and is present always, so we should be good by making it required 👀 

Good one

murtaza98 (dismissed):

murtaza98 (approved):

apps/meteor/app/livechat/server/lib/RoutingManager.ts
```diff
@@ -101,17 +153,17 @@ export const RoutingManager = {
...
 
 		await LivechatRooms.changeAgentByRoomId(rid, agent);
-		await Rooms.incUsersCountById(rid);
+		await Rooms.incUsersCountById(rid, 1);
 
 		const user = await Users.findOneById(agent.agentId);
 		const room = await LivechatRooms.findOneById(rid);
 
-		await Message.saveSystemMessage('command', rid, 'connected', user);
+		user && (await Message.saveSystemMessage('command', rid, 'connected', user));
```
sampaiodiego, line 161: this is the pattern we use, right?
```suggestion
		if (user) {
			await Message.saveSystemMessage('command', rid, 'connected', user));
		}
```

sampaiodiego (changes requested):

apps/meteor/app/livechat/server/api/lib/livechat.ts
```diff
@@ -126,7 +126,7 @@ export function getRoom({
 	return LivechatTyped.getRoom(guest, message, roomInfo, agent, extraParams);
 }
 
-export async function findAgent(agentId: string): Promise<void | { hiddenInfo: true } | ILivechatAgent> {
+export async function findAgent(agentId?: string): Promise<void | { hiddenInfo: true } | ILivechatAgent> {
```
ggazzo, line 129: is this really correct? why you are trying to find the agent if you dont have the id?

apps/meteor/app/livechat/server/lib/RoutingManager.ts
```diff
@@ -19,7 +29,45 @@ import { Apps, AppEvents } from '../../../../ee/server/apps';
 
 const logger = new Logger('RoutingManager');
 
-export const RoutingManager = {
+type Routing = {
```
ggazzo, line 32: since you are not exporting this type  why not just type the functions implemented on the RoutingManager ?

ggazzo (dismissed):

apps/meteor/app/livechat/server/lib/RoutingManager.ts
KevLehman: Not quite following, this type is typing the fucntions implemented on routingmanager 🤔 

But, RoutingManager is an object, so the types are "separated". Planned to convert it to a class when we move it to a service.

apps/meteor/app/livechat/server/methods/getNextAgent.ts
```diff
@@ -38,7 +38,7 @@ Meteor.methods<ServerMethods>({
 		}
 
 		const agent = await Livechat.getNextAgent(department);
-		if (!agent) {
+		if (!agent?.agentId) {
```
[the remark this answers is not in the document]
KevLehman: I'll keep old bhv just in case 😬

murtaza98 (approved):

ggazzo (approved):

PR #29416, opened 2023-06-01
fix: Imported messages are not displayed

matheusbsilva137:

 - Fix issue with imported messages not being displayed in the room.
 - Fix importer agent being added as a member of every imported room.

## Issue(s)

Introduced in #28687 and #28695

## Further comments

Imported messages were added along with a `u._id` field, which caused them not to be displayed (messages would have a `u._id` field  and ALSO the `u`object with the `_id` and `username` fields).

TC-732


apps/meteor/app/lib/server/functions/createRoom.ts
```diff
@@ -24,6 +24,7 @@ export const createRoom = async <T extends RoomType>(
 	name: T extends 'd' ? undefined : string,
 	ownerUsername: string | undefined,
 	members: T extends 'd' ? IUser[] : string[] = [],
+	excludeSelf?: boolean,
```
totoi690, line 27: Couldn't we here transform all the optional parameters into a object so we don't have to add this `excludeSelf` in all other files?

pierre-lehnen-rc: At least move the new param to the existing options object - also add a comment to the type definition to clarify what this param does, as "excludeSelf" by itself doesn't say much.

apps/meteor/app/lib/server/functions/insertMessage.ts
```diff
@@ -32,8 +32,6 @@ export const insertMessage = async function (
 		}
 		await Messages.insertOne({
```
pierre-lehnen-rc, line 33: this insertOne is supposed to only run when `existingMessage` is undefined - can you add an `else` block around it?
as it is right now it'll be duplicating messages if the import is executed multiple times.

pierre-lehnen-rc (changes requested):

apps/meteor/app/lib/server/functions/insertMessage.ts
matheusbsilva137: Thankss, fixed!

apps/meteor/app/lib/server/functions/createRoom.ts
matheusbsilva137: I'll work on it in a separate PR just so that we do not hold this one for the patch release

gabriellsh (approved):

pierre-lehnen-rc (approved):

PR #29436, opened 2023-06-02
chore: dup of #28720 for tests

murtaza98:


apps/meteor/app/livechat/server/lib/QueueManager.js
```diff
@@ -73,7 +73,13 @@ export const QueueManager = {
 		await queueInquiry(room, inquiry, agent);
 		logger.debug(`Inquiry ${inquiry._id} queued`);
 
-		return LivechatRooms.findOneById(rid);
+		const newRoom = await LivechatRooms.findOneById(rid);
+		if (!newRoom) {
+			logger.debug(`Room with id ${rid} not found`);
```
sampaiodiego, line 78: should be a `logger.error` since it is being `throw` ?

apps/meteor/app/livechat/server/api/v1/room.ts
```diff
@@ -59,14 +59,14 @@ API.v1.addRoute('livechat/room', {
...
 			}
 
-			let agent;
+			let agent: SelectedAgent | undefined;
 			const agentObj = agentId && (await findAgent(agentId));
 			if (agentObj) {
 				if (isAgentWithInfo(agentObj)) {
 					const { username = undefined } = agentObj;
-					agent = { agentId, username };
+					agent = { agentId, username } as SelectedAgent;
```
sampaiodiego, line 67: this type cast looks wrong.. `username` cannot be `undefined` in `SelectedAgent`..

apps/meteor/app/livechat/server/api/v1/room.ts
```diff
@@ -59,14 +59,14 @@ API.v1.addRoute('livechat/room', {
...
 				} else {
-					agent = { agentId };
+					agent = { agentId } as SelectedAgent;
```
sampaiodiego, line 69: this looks wrong as well, since `username` is mandatory on `SelectedAgent`.. maybe `SelectedAgent` needs to be modified to allow both use cases on this file, and then the type cast can be removed.

sampaiodiego (changes requested):

KevLehman:
Hey @sampaiodiego , this PR should be in draft, but pls move your comments to #28720 as this is supposed to be the same PR (not completely sure _why_ this extra PR was needed tho, as we only do these for community)

sampaiodiego:
> Hey @sampaiodiego , this PR should be in draft, but pls move your comments to #28720 as this is supposed to be the same PR (not completely sure _why_ this extra PR was needed tho, as we only do these for community)

I'm confused 😞

murtaza98:
Sorry about the confusion guys. I was struggling with trying to reproduce an issue which QA team were facing on the QA environment. That issue was not reproducible anywhere else locally expect for that QA env. Hence we thought that something much be going wrong on that QA env due to which we decided to spin up yet another QA env for testing this PR. Hence I created this new PR to spin up a brand new QA env. 
PS: Good news is that we found a solution to the problem we were facing before. It apparently was an issue with cronjobs not working on these QA env which was fixed here https://github.com/RocketChat/Rocket.Chat/pull/29449

PR #29455, opened 2023-06-05
fix: Show agent email not being passed to model calls

KevLehman:
OC-957


.changeset/brown-forks-fry.md
```diff
@@ -0,0 +1,5 @@
+---
+"@rocket.chat/meteor": patch
+---
+
+Fixed a problem where the setting `Livechat_show_agent_email` was not being used by the back when returning agent's info
```
debdutdeb, line 5: Do you think, since this is supposed to be for users, we should instead of using setting ids here, provide setting paths like `Omnichannel > Email > Show Agent Email` ?

KevLehman: :hmmm: not really sure. But I do agree we can remove the _ from the setting name lol

KevLehman, line 5:
```suggestion
Fixed a problem where the setting `Show Agent Email` from Omnichannel was not being used by the back when returning agent's info
```

apps/meteor/app/apps/server/bridges/livechat.ts
```diff
@@ -88,7 +89,7 @@ export class AppLivechatBridge extends LivechatBridge {
 
 		let agentRoom;
 		if (agent?.id) {
-			const user = await Users.getAgentInfo(agent.id);
+			const user = await Users.getAgentInfo(agent.id, settings.get('Livechat_show_agent_email'));
```
murtaza98, line 92: I think we can ignore the settings call here since we're not using the email's value

apps/meteor/server/models/raw/Users.js
```diff
@@ -1644,7 +1644,7 @@ export class UsersRaw extends BaseRaw {
 		};
 
 		if (showAgentEmail) {
-			options.fields.emails = 1;
+			options.projection.emails = 1;
```
murtaza98, line 1647: The if here seems a bit unnecessary here, wdyt? Perhaps we can move the logic up while the option's prop is being built... perhaps something like this:

```
const options = {
			projection: {
				name: 1,
				username: 1,
				phone: 1,
				customFields: 1,
				status: 1,
				livechat: 1,
				...(showAgentEmail && { emails: 1 }),
			},
		};
```

apps/meteor/app/apps/server/bridges/livechat.ts
KevLehman: hmmm, not really sure. The original model implementation run the setting for all cases, so I mimic it on here. We may need to revisit these places later to see if we still need this data there, but didn't want to miss any place and cause a random issue somewhere

murtaza98 (dismissed):

sampaiodiego (approved):

apps/meteor/app/apps/server/bridges/livechat.ts
```diff
@@ -89,7 +90,7 @@ export class AppLivechatBridge extends LivechatBridge {
 
 		let agentRoom: SelectedAgent | undefined;
 		if (agent?.id) {
-			const user = await Users.getAgentInfo(agent.id);
+			const user = await Users.getAgentInfo(agent.id, settings.get('Livechat_show_agent_email'));
```
d-gubert, line 93: Is it possible to use the `Settings` raw model here? We're working on removing these fiber usages on our Microservices PRs

d-gubert (changes requested):

apps/meteor/app/apps/server/bridges/livechat.ts
KevLehman: Hey! There shouldn't be a problem, since `settings.get` doesn't rely on fibers since some time ago (it actually never _relied_ on them, but there was a kind of middleware that used fibers, which was removed)

KevLehman: If you find any problems after this, call Diego :trollface:

d-gubert: Oh, ok, cool!

d-gubert (approved):
All good on apps side

pierre-lehnen-rc (dismissed):

murtaza98 (approved):

pierre-lehnen-rc (approved):

PR #29457, opened 2023-06-05
fix: handle list type values in app setting comparison (old vs new)

AllanPazRibeiro:
This pull request proposes changes to compare the equality of app settings when the settings are represented as arrays. The comparison is performed using JSON.stringify().

Changes Made:

Utilized JSON.stringify() to obtain string representations of the previous and new settings and compared the strings to determine the equality of the app settings:

`JSON.stringify(oldSetting) === JSON.stringify(setting.value)`

## Steps to test or reproduce

1. Generate an app with a room selector setting
2. Install it on the server
3. Change the room setting to a different one
4. Save it


apps/meteor/server/services/apps-engine/service.ts
```diff
@@ -90,6 +90,18 @@ export class AppsEngineService extends ServiceClassInternal implements IAppsEngi
 
 			// avoid updating the setting if the value is the same,
 			// which caused an infinite loop
+			// and sometimes the settings can be an array
+			// so we need to convert it to JSON stringified to compare it
+
+			if (Array.isArray(oldSetting) && Array.isArray(setting.value)) {
+				if (JSON.stringify(oldSetting) === JSON.stringify(setting.value)) {
```
debdutdeb, line 97: general query, is the insert function always sorted?

KevLehman, line 97: This could fail if the settings have different order, but not sure if that's really a problem.

KevLehman: Jesus man, I posted it first :kek:

debdutdeb: 😂 sorry man didn't see hahahaha

debdutdeb: consider mine resolved 😂

AllanPazRibeiro: The order does not matter in this case, we just need to avoid it if it is strictly equal.

KevLehman: No no, yeah. I was wondering because `['a','b','c']` stringified will be different from `['b','c','a']`.

Same array, different order, will say they're "different" when they're not

d-gubert: I don't think this would be considered an issue here, since the settings are generated as a form to the user to interact with and changes of order really do mean that the value has changed.

KevLehman: Noice, I'll resolve this then :)

d-gubert (approved):

KevLehman (approved):

debdutdeb (approved):

apps/meteor/server/services/apps-engine/service.ts
```diff
@@ -90,7 +90,10 @@ export class AppsEngineService extends ServiceClassInternal implements IAppsEngi
 
 			// avoid updating the setting if the value is the same,
 			// which caused an infinite loop
-			if (oldSetting === setting.value) {
+			// and sometimes the settings can be an array
+			// so we need to convert it to JSON stringified to compare it
+
+			if (JSON.stringify(oldSetting) === JSON.stringify(setting.value)) {
```
ggazzo, line 96: it that the best way to compare arrays?

PR #29461, opened 2023-06-05
feat: `Contextualbar` resizable

dougfabris:

Kapture 2023-06-23 at 18 47 30

## Issue(s)

Closes https://github.com/RocketChat/feature-requests/issues/159


Gummikavalier:
A suggestion: Currently we run the separate thread view as big as this with Custom CSS:

Screenshot from 2023-09-27 16-50-47

We hope that this separated view could be stretched from its default 50:50 form factor to something like in the picture.

At this custom size it is much more pleasing to read and focus on the discussion at hand.

dougfabris:
@Gummikavalier Thanks for the heads up, we'll consider it on delivering this feature!

ggazzo (approved):

MartinSchoeler (approved):

Gummikavalier:
This one is currently broken in RC 6.8.0; Hovering mouse simply does nothing. I cleared our custom CSS first for this test.

dougfabris:
@Gummikavalier it was released as a feature preview, in order to be able to use it you have to enable the feature under User Account > Feature preview and of course the feature preview should be active on the workspace also.

Gummikavalier:
@dougfabris Ahh, I eyeballed for it under admin settings but couldn't find it first. Found the `Feature preview` now.

Resizing works a treat now. Thanks! :heart_eyes:

Gummikavalier:
@dougfabris I'll be updating one production system this week to RC 6.8.0 so we may get some comments after that.

A forewarning though; We've spoiled our users for years with that a lot more extended custom `Expand` thread view mode I mentioned earlier here. So it is not likely many will switch from expand thread view to to regular thread view due to latter having only 50% max size. :thinking:

Most rejoicing would be heard from option to resize both thread views. And extra hugs from ability to tune the channel listing bar as well. :hugs: 

(Yes, appetite grows from these small tasters. :grin: )

Gummikavalier:
@dougfabris Actually even my personal preference is the Expanded view as it darkens the main channel and focuses the eyes on the thread. It stands out better and the general esthetics is less cluttered. Meaning that the user is more focused to the discussion.

Several exactly same looking parallel text columns is quite confusing (to me at least). Looking at Discord often makes me vince how bad it can be. :nauseated_face: 

With regular view if you were able to increase the size to higher than 50%, it'd probably also reformat the messages in the main channel view quite thin. Not sure, would probably need some prototyping to see how it'd work.

If only regular mode is the way to go in the future to simplify design, maybe some coloring or hue changes to either main channel or thread background could help with that.

dougfabris:
@Gummikavalier Yeah, maybe a focus mode option that makes the main channel darker instead of extended. Thank your for bring your opinion, I will add it to the discussion with the design team

Gummikavalier:
@dougfabris No problem and thank _you_. Regarding on comments on resized contextual bar; it is great that particularly message search, threads listing and members list views can be extended; it making them easier to read.

Gummikavalier:
@dougfabris FYI: Not reporting this as an official bug, as the feature is in preview. If you hover the mouse in expanded view over a line where the divider would be in collapsed view, it shows through as a blue line.

Screenshot from 2024-05-17 13-56-44

Probably not worth fixing if expanded view would be going away, though.

dougfabris:
@Gummikavalier This is a know issue indeed, thank you for reporting. But yeah, we have to check first about the usage of the extended view if we doesn't keep it, theres no why fix it

PR #29500, opened 2023-06-07
feat: Add Apps-engine Thread Bridge

ggazzo:


apps/meteor/app/apps/server/converters/threads.ts
```diff
@@ -0,0 +1,203 @@
...
+			audioSize: 'audio_size',
+			videoUrl: 'video_url',
+			videoType: 'video_type',
+			videoSize: 'video_size',
+			fields: 'fields',
+			actionButtonsAlignment: 'button_alignment',
+			actions: 'actions',
+			type: 'type',
+			description: 'description',
+			author: (attachment: ExcludeUndefined<IMessage['attachments']>[number]) => {
```
d-gubert, line 168: Is there a difference in using this type expression instead of the plain `MessageAttachment` type?

apps/meteor/app/apps/server/converters/threads.ts
```diff
@@ -0,0 +1,203 @@
...
+import { Messages } from '@rocket.chat/models';
+import type { IMessage as AppsEngineMessage } from '@rocket.chat/apps-engine/definition/messages';
+import type { IRoom } from '@rocket.chat/apps-engine/definition/rooms';
+
+import { transformMappedData } from '../../../../ee/lib/misc/transformMappedData';
+
+// eslint-disable-next-line @typescript-eslint/naming-convention
+interface Orchestrator {
+	rooms: () => {
+		convertById(id: string): Promise<unknown>;
```
d-gubert, line 11: Curious as to why you preferred "casting" the return type when using the method instead of typing the return here :thinking:

d-gubert (changes requested):
A few comments on some of the types used

apps/meteor/app/apps/server/converters/threads.ts
ggazzo: Exactly the same, there is no difference. but since we were manipulating IMesssage types, I thought it would make more sense to continue with IMessage usage

apps/meteor/app/apps/server/converters/threads.ts
```diff
@@ -0,0 +1,203 @@
...
+			},
+		};
+
+		return Promise.all(attachments.map(async (attachment) => transformMappedData(attachment, map)));
+	}
+}
+
+// exclude undefined values
+
+type ExcludeUndefined<T> = T extends undefined ? never : T;
```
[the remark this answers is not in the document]
ggazzo: I think you are right no significant difference

d-gubert (approved):

PR #29519, opened 2023-06-12
ci: New action to publish releases

sampaiodiego:

There are two new workflows using the new action:

- One is manual, used to trigger new releases for current version or a new patch release
- The other runs on every push to master and automatically cut a release and publishes packages to Npm registry


.github/workflows/publish-release.yml
```diff
@@ -0,0 +1,35 @@
...
+      - name: Install Dependencies
+        run: yarn
+
+      - name: Build
+        run: yarn build
+
+      - name: Cut release
+        uses: ./packages/release-action
+        with:
+          action: cut
```
sampaiodiego, line 32:
```suggestion
          action: publish
```

packages/release-action/src/bumpNextVersion.ts
sampaiodiego, line 1: need a way to trigger the next release candidate

ggazzo (approved):

PR #29521, opened 2023-06-12
fix: Ignore archived departments from business hours queries

KevLehman:
## Issue(s)

https://rocketchat.atlassian.net/browse/OC-1050


apps/meteor/server/models/raw/LivechatDepartment.ts
```diff
@@ -330,6 +336,14 @@ export class LivechatDepartmentRaw extends BaseRaw<ILivechatDepartment> implemen
 
 		return this.find(query, options);
 	}
+
+	findNotArchived(options: FindOptions<ILivechatDepartment> = {}): FindCursor<ILivechatDepartment> {
+		const query = {
+			$or: [{ archived: { $exists: false } }, { archived: false }],
```
sampaiodiego, line 342: can `archived` have all the three states? like `false`, `true` and `undefined`? 

well maybe my bad on previous review, but ideally it should have only two states.. either: 
- always `true` or `false`
- or `true/false` and `undefined`

anyways, this query could be simpler, like `{ archived: { $ne: true } }` .. and unfortunately this query doesn't uses the most of the index.. negative conditions must go through all records anyways (the index will have only the records **with** the field, but the query also matches records that are not there.

sampaiodiego (changes requested):

apps/meteor/ee/app/livechat-enterprise/server/business-hour/Helper.ts
```diff
@@ -6,9 +6,19 @@ import { LivechatBusinessHours, LivechatDepartment, LivechatDepartmentAgents, Us
...
 const getAllAgentIdsWithoutDepartment = async (): Promise<string[]> => {
+	// Fetch departments with agents excluding archived ones (disabled ones still can be tied to business hours)
+	// Then find the agents that are not in any of those departments
+
+	// TODO: define behaviors for disabled departmnents
+	const departmentIds = (await LivechatDepartment.findNotArchived({ projection: { _id: 1 } }).toArray()).map(({ _id }) => _id);
+
 	const agentIdsWithDepartment = (
-		await LivechatDepartmentAgents.find({ departmentEnabled: true }, { projection: { agentId: 1 } }).toArray()
+		await LivechatDepartmentAgents.find(
```
murtaza98, line 16: We can make this find distinct ;)

apps/meteor/ee/app/livechat-enterprise/server/business-hour/Helper.ts
```diff
@@ -6,9 +6,19 @@ import { LivechatBusinessHours, LivechatDepartment, LivechatDepartmentAgents, Us
...
+			{ departmentEnabled: true, departmentId: { $in: departmentIds } },
+			{ projection: { agentId: 1 } },
+		).toArray()
 	).map((dept: any) => dept.agentId);
```
murtaza98, line 20: Not related but I think we can remove `any` over here

apps/meteor/server/models/raw/LivechatDepartment.ts
KevLehman: Sadly yes, a `never archived` department won't have the property `archived` in it. Only after a dep is archived then the property becomes `true`|`false`.

Good part is departments never grow that big for indexes to be critical, so both $ne: true & $exists: false should have similar perf.

apps/meteor/ee/app/livechat-enterprise/server/business-hour/Helper.ts
KevLehman: image

murtaza98 (approved):

apps/meteor/server/models/raw/LivechatDepartment.ts
```diff
@@ -330,6 +336,14 @@ export class LivechatDepartmentRaw extends BaseRaw<ILivechatDepartment> implemen
 
 		return this.find(query, options);
 	}
+
+	findNotArchived(options: FindOptions<ILivechatDepartment> = {}): FindCursor<ILivechatDepartment> {
+		const query = {
+			$or: [{ archived: { $ne: false } }],
+		};
```
sampaiodiego, line 343:
```suggestion
		const query = { archived: { $ne: false } };
```

apps/meteor/server/models/raw/LivechatDepartment.ts
```diff
@@ -83,6 +83,12 @@ export class LivechatDepartmentRaw extends BaseRaw<ILivechatDepartment> implemen
 				},
 				sparse: true,
 			},
+			{
+				key: {
+					archived: 1,
+				},
+				sparse: true,
+			},
```
sampaiodiego, line 91: per my previous comment, this index won't be used by the query on `findNotArchived`, so need to have it.
```suggestion
```

sampaiodiego (changes requested):

apps/meteor/server/models/raw/LivechatDepartment.ts
KevLehman: Oh no, this index is not for _that_ query. Is for the findArchived we have. (That query is outside the model apparently)

Didn't want to do a PR solely for this

sampaiodiego: oh, ok then =)

sampaiodiego (approved):

murtaza98 (approved):

PR #29525, opened 2023-06-12
feat: use persistent queue for matrix events on federation

lmauromb:

- introduces a new package, `@rocket.chat/queue-wrapper`, which is used to interact with `mongo-message-queue` to have a DB queue.
- replaces the InMemoryQueue Wrapper with an Adapter (**Queue**) which uses the new `@rocket.chat/queue-wrapper`

## Issue(s)

Jira: https://rocketchat.atlassian.net/browse/FED-107 and https://rocketchat.atlassian.net/browse/FED-108

## Steps to test or reproduce

This is a behind the scenes change, federation behavior should be the same.


MarcosSpessatto (changes requested):
Everything looks good to me, but when we are running a CE environment, this is not working, I believe we are creating the `handleMatrixEvent` method on the EE service only.

apps/meteor/server/models/raw/PersistentQueue.ts
```diff
@@ -0,0 +1,19 @@
+import type { Db, IndexDescription } from 'mongodb';
+import type { IPersistentQueueModel } from '@rocket.chat/model-typings';
+
+import { BaseRaw } from './BaseRaw';
+
+export class PersistentQueueModel extends BaseRaw<any> implements IPersistentQueueModel {
```
MarcosSpessatto, line 6: I think we should provide the correct typings for what's being handled here, instead of `any`. As the example here: https://github.com/RocketChat/Rocket.Chat/blob/develop/apps/meteor/server/models/raw/MatrixBridgedRoom.ts#L7
and the type: https://github.com/RocketChat/Rocket.Chat/blob/develop/apps/meteor/server/models/raw/MatrixBridgedRoom.ts#L1

MarcosSpessatto (changes requested):

MarcosSpessatto (dismissed):

apps/meteor/server/models/raw/PersistentQueue.ts
```diff
@@ -0,0 +1,20 @@
+import type { Db, IndexDescription } from 'mongodb';
+import type { IPersistentQueueModel } from '@rocket.chat/model-typings';
+import type { IPersistentQueue } from '@rocket.chat/core-typings';
+
+import { BaseRaw } from './BaseRaw';
+
+export class PersistentQueueModel extends BaseRaw<IPersistentQueue> implements IPersistentQueueModel {
```
KevLehman, line 7: If this is to add indexes to the collection mongomessagequeue uses, i vaguely remember it was `_queue`

Additional, if it's the same collection and you're adding indexes here, you can remove the index creation step on the queue-worker microservice

packages/model-typings/src/models/IPersistentQueueModel.ts
```diff
@@ -0,0 +1,3 @@
+import type { IBaseModel } from './IBaseModel';
+
+export type IPersistentQueueModel = IBaseModel<any>;
```
KevLehman, line 3: 👀

packages/queue-wrapper/src/logger.ts
```diff
@@ -0,0 +1,3 @@
+import { pino } from 'pino';
+
+export const Logger = pino({});
```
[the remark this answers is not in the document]
lmauromb: what do you think about extracting our current Logger into a package and import it directly?

apps/meteor/server/models/raw/PersistentQueue.ts
lmauromb: you're right this is to setup the indexes! 

But we're not using the same collection since in one moment both this new package and the service used by Omnichannel will live at the same time. 

An additional PR could be made to update Omnichannel to this package and remove the other service.

packages/model-typings/src/models/IPersistentQueueModel.ts
lmauromb: my bad, I think we can use this, right?

```typescript
import type { IBaseModel } from './IBaseModel';
import type { IPersistentQueue } from '@rocket.chat/core-typings'; 

export type IPersistentQueueModel = IBaseModel<IPersistentQueue>;
```

PR #29531, opened 2023-06-13
chore: change public room deletion behaviour 

totoi690:
Change behavior so when a user is the last owner of a public room and the user is deleted, the room is also deleted.

## Steps to test or reproduce
- Create a public room
- Set a user as its owner (make it the last owner and last user of the room)
- Delete the user
The public room should be deleted

## Further comments

TC-749

### DIFF
--- a/apps/meteor/app/lib/server/functions/getRoomsWithSingleOwner.ts
+++ b/apps/meteor/app/lib/server/functions/getRoomsWithSingleOwner.ts
@@ -63,12 +63,12 @@ export async function getSubscribedRoomsForUserWithDetails(
 					roomData.shouldChangeOwner = true;
 				}
 
-				// If there's no subscriber available to be the new owner and it's not a public room, we can remove it.
-				if (!roomData.shouldChangeOwner && roomData.t !== 'c') {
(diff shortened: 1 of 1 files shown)


matheusbsilva137 (approved):

ggazzo:
at the end of the day the admin can 'steal' rooms from others by having sufficient permissions, so this should not be  a privacy issue.

Now imagine a company that deletes a user (or a few users or an area) and the room disappears and can never be recovered

jsdgdo:
From a UX standpoint I also don't think deleting the room is the best approach. I did some benchmarking and none of our competitors does so. My suggestion would be:
1. Not deleting the channel
2. Add an `empty` tag to it when listed in the directory
3. Change the color of the people count when it's 0
4. When the last member of a public channel is deleted: 
a. transfer ownership to the workspace creator (or upmost admin)
b. Notify them as such, telling them that the channel has no members, and for them to check out what's in there, if it's important, and if not to maybe delete it. 
c. Start a 30 day timer, to send that notification again. Remind them, and suggest to delete again. If they don't delete it after that then they can keep the clutter.

PR #29543, opened 2023-06-14
fix: Remove the association between BH and disabled/archived departments

KevLehman:
[OC-1056]
[OC-1059]
[OC-1050]

## Further comments

Tests added:
- https://rocketchat.atlassian.net/browse/OC-1078

[OC-1056]: https://rocketchat.atlassian.net/browse/OC-1056?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[OC-1059]: https://rocketchat.atlassian.net/browse/OC-1059?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[OC-1050]: https://rocketchat.atlassian.net/browse/OC-1050?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/meteor/ee/app/livechat-enterprise/server/business-hour/Multiple.ts
```diff
@@ -132,6 +132,27 @@ export class MultipleBusinessHoursBehavior extends AbstractBusinessHourBehavior
...
+		if (!department.businessHourId) {
+			return;
+		}
+		// Get business hour
+		const businessHour = await this.BusinessHourRepository.findOneById(department.businessHourId);
+		if (!businessHour) {
+			return;
+		}
+		// Check if i'm the only department on this business hour
+		const imTheOnlyOne = !(await LivechatDepartment.countByBusinessHourId(businessHour._id));
```
murtaza98, line 145: 2 points here:
1) Shouldn't we be checking enabled departments only over here?
2) I'm a bit confused if this works. Let's say there's only one dep connected to BH and it's the one which we're disabling right now, then "imTheOnlyOne" would be (!1) => false. On the other hand, if there are more deps connected to this BH, then "imTheOnlyOne" would be "(!2)" => false. Due to this "closeBusinessHour" never gets executed. Please let me know if I'm missing something here 😅

KevLehman: For 1: no, we want all. There's another job that will remove all the disabled departments whenever the server starts + business hours is enabled, so in here we can be safe we won't count disabled ones.

For 2: :kek: forgot

murtaza98 (dismissed):

murtaza98 (dismissed):

apps/meteor/ee/app/livechat-enterprise/server/business-hour/Multiple.ts
```diff
@@ -132,6 +132,38 @@ export class MultipleBusinessHoursBehavior extends AbstractBusinessHourBehavior
...
+	}
+
+	async onDepartmentArchived(department: Pick<ILivechatDepartment, '_id'>): Promise<void> {
+		const dbDepartment = await LivechatDepartment.findOneById(department._id, { projection: { businessHourId: 1, _id: 1 } });
+
+		if (!dbDepartment) {
+			return;
+		}
+
+		console.log(this.onDepartmentDisabled);
```
murtaza98, line 163:
```suggestion
```

murtaza98 (approved):

apps/meteor/server/models/raw/LivechatDepartment.ts
```diff
@@ -345,6 +350,60 @@ export class LivechatDepartmentRaw extends BaseRaw<ILivechatDepartment> implemen
...
+					},
+				},
+				{
+					$group: {
+						_id: '$businessHourId',
+						validDepartments: {
+							$push: {
+								$cond: {
+									if: {
+										$or: [
```
murtaza98, line 377: Couldn't we simplify this OR by just having the enabled prop set to true considering archived departments always have "enabled" set to false

murtaza98 (dismissed):

ggazzo (approved):
try to avoid multiple fixes in a single pr, it makes the review job tiring and difficult to understand. Even if it takes a little more work, try to do one fix at a time

murtaza98 (dismissed):

murtaza98 (dismissed):

murtaza98 (approved):

apps/meteor/app/livechat/server/lib/Livechat.js
```diff
@@ -755,7 +757,11 @@ export const Livechat = {
 			});
 		}
 
-		return LivechatDepartmentRaw.archiveDepartment(_id);
+		await LivechatDepartmentAgents.disableAgentsByDepartmentId(_id);
+		await LivechatDepartmentRaw.archiveDepartment(_id);
+
+		this.logger.debug('Running livechat.afterDepartmentArchived callback for department:', _id);
```
sampaiodiego, line 763:
```suggestion
		this.logger.debug({ msg: 'Running livechat.afterDepartmentArchived callback for department:', departmentId: _id });
```

apps/meteor/ee/app/livechat-enterprise/server/business-hour/Helper.ts
```diff
@@ -6,9 +6,17 @@ import { LivechatBusinessHours, LivechatDepartment, LivechatDepartmentAgents, Us
...
-	const agentIdsWithDepartment = (
-		await LivechatDepartmentAgents.find({ departmentEnabled: true }, { projection: { agentId: 1 } }).toArray()
-	).map((dept: any) => dept.agentId);
+	// Fetch departments with agents excluding archived ones (disabled ones still can be tied to business hours)
+	// Then find the agents that are not in any of those departments
+
+	// TODO: define behaviors for disabled departmnents
+	const departmentIds = (await LivechatDepartment.findNotArchived({ projection: { _id: 1 } }).toArray()).map(({ _id }) => _id);
+
+	const agentIdsWithDepartment = await LivechatDepartmentAgents.col.distinct('agentId', {
```
sampaiodiego, line 15: even though `col` is public, you should not use it from outside of the model.. `col` being public is a tech debt we need to pay.

please create a model method to perform this query.

sampaiodiego (changes requested):

murtaza98 (approved):

apps/meteor/app/livechat/server/business-hour/BusinessHourManager.ts
```diff
@@ -115,12 +159,27 @@ export class BusinessHourManager {
 			callbacks.priority.HIGH,
 			'business-hour-livechat-on-save-agent-department',
 		);
+		callbacks.add(
+			'livechat.afterDepartmentDisabled',
+			this.behavior.onDepartmentDisabled.bind(this),
+			callbacks.priority.HIGH,
+			'business-hour-livechat-on-department-disabled',
+		);
+		// TODO: Binding the correct "this" for the context of onDepartmentArchived. Should we do the same for other callbacks?
```
ggazzo, line 168: this comment is not really necessary. If you have a question, or believe something is wrong, please find out and act accordingly. this kind of thought scattered in the code does not add anything

btw: passing a method of an object as a callback and expecting this to remain the same is wrong and amateur, whoever does this should be aware
```suggestion
```

ggazzo (changes requested):

ggazzo (approved):

apps/meteor/app/livechat/server/business-hour/BusinessHourManager.ts
KevLehman: The comment is indeed outdated, so thanks for removing it.

As you see, the idea of the calls never implies they're "the same", rather, than for some reason we're passing `this` which in the context is an instance of `businessHourManager` instead of `this.behavior` since the calls are coming from that context. However, none of the calls actually reference `this` inside, so the comment was meant to handle it later and test accordingly, since for now, it's not causing any harm 🤗

murtaza98 (approved):

ggazzo (approved):

PR #29553, opened 2023-06-15
member list is randomly sorted #22145

namanchandak:


debdutdeb:
Hi, thanks for your contribution. Unfortunately, this does not fix the issue. A pr is already open (possibly even merged at this point) with the fix.

namanchandak:
Can you please tell me in which file in the codebase the members page is there?
I am  going to change it

> Hi, thanks for your contribution. Unfortunately, this does not fix the issue. A pr is already open (possibly even merged at this point) with the fix.

PR #29572, opened 2023-06-16
chore: improve omnichannel tests reliability

aleksandernsilva:

### DIFF
--- a/apps/meteor/tests/e2e/page-objects/fragments/home-sidenav.ts
+++ b/apps/meteor/tests/e2e/page-objects/fragments/home-sidenav.ts
@@ -64,10 +64,14 @@ export class HomeSidenav {
 		await this.page.locator(`//li[@class="rcx-option"]//div[contains(text(), "${status}")]`).click();
 	}
 
-	async openChat(name: string): Promise<void> {
-		await this.page.locator('role=navigation >> role=button[name=Search]').click();
-		await this.page.locator('role=search >> role=searchbox').type(name);
-		await this.page.locator(`role=search >> role=listbox >> role=link >> text="${name}"`).click();
+	async openChat(name: string, search = true): Promise<void> {
+		if (search) {
+			await this.page.locator('role=navigation >> role=button[name=Search]').click();
+			await this.page.locator('role=search >> role=searchbox').type(name);
+			await this.page.locator(`role=search >> role=listbox >> role=link >> text="${name}"`).click();
+		} else {
+			await this.getSidebarItemByName(name).click();
+		}
 	}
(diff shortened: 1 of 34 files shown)


apps/meteor/tests/e2e/page-objects/fragments/home-sidenav.ts
```diff
@@ -64,10 +64,14 @@ export class HomeSidenav {
...
-		await this.page.locator('role=search >> role=searchbox').type(name);
-		await this.page.locator(`role=search >> role=listbox >> role=link >> text="${name}"`).click();
+	async openChat(name: string, search = true): Promise<void> {
+		if (search) {
+			await this.page.locator('role=navigation >> role=button[name=Search]').click();
+			await this.page.locator('role=search >> role=searchbox').type(name);
+			await this.page.locator(`role=search >> role=listbox >> role=link >> text="${name}"`).click();
+		} else {
+			await this.getSidebarItemByName(name).click();
+		}
```
ggazzo, line 74: why not call `getSidebarItemByName`directly ?

ggazzo (changes requested):
just keep in mind  that you are actually changing the test before people were opening through search and now the test is being performed agains the regular sidebar


btw is this a problem yet? I thought it was not happening anymore those kind of errors

apps/meteor/tests/e2e/page-objects/fragments/home-sidenav.ts
aleksandernsilva: > just keep in mind that you are actually changing the test before people were opening through search and now the test is being performed agains the regular sidebar

That's the objective ;) I want **Omnichannel tests** to get the room directly from the sidebar. This should prevent a problem we are facing where it takes some time for the room to show up in the queue. I created this draft specifically to confirm that.

> why not call `getSidebarItemByName` directly ?

Actually might be a good idea.

MartinSchoeler (approved):

MartinSchoeler (dismissed):

apps/meteor/tests/e2e/page-objects/omnichannel-triggers.ts
```diff
@@ -13,7 +13,7 @@ export class OmnichannelTriggers {
 	}
 
 	get btnNew(): Locator {
-		return this.page.locator('text=New');
+		return this.page.locator('button >> text=New');
```
ggazzo, line 16:
```suggestion
		return this.page.locator('role=button[name="New"');
```

ggazzo (approved):

PR #29581, opened 2023-06-19
chore: Add eslint `react-jsx-key` rule on client

dougfabris:


gabriellsh (dismissed):
Why didn't we have this before????? Nice one

tassoevan (approved):

PR #29589, opened 2023-06-20
fix: Dont update a user's livechat status when its status is offline

KevLehman:
OC-1069

### DIFF
--- a/apps/meteor/server/models/raw/Users.js
+++ b/apps/meteor/server/models/raw/Users.js
@@ -968,6 +968,7 @@ export class UsersRaw extends BaseRaw {
 	setLivechatStatusActiveBasedOnBusinessHours(userId) {
 		const query = {
 			_id: userId,
+			status: 'online',
 			openBusinessHours: {
 				$exists: true,
 				$not: { $size: 0 },


apps/meteor/server/models/raw/Users.js
```diff
@@ -968,6 +968,7 @@ export class UsersRaw extends BaseRaw {
 	setLivechatStatusActiveBasedOnBusinessHours(userId) {
 		const query = {
 			_id: userId,
+			status: 'online',
```
KevLehman, line 971:
```suggestion
			status: { $ne: 'offline' },
```

apps/meteor/app/livechat/server/business-hour/index.ts
```diff
@@ -16,8 +16,11 @@ Meteor.startup(async () => {
 	businessHourManager.registerBusinessHourBehavior(new BusinessHourBehaviorClass());
 	businessHourManager.registerBusinessHourType(new DefaultBusinessHour());
 
-	Accounts.onLogin(
-		async ({ user }: { user: any }) =>
-			user?.roles?.includes('livechat-agent') && !user?.roles?.includes('bot') && businessHourManager.onLogin(user._id),
+	// Note: This shouldn't happen directly on login, but after the login process has already finished
+	// but meteor doesn't have "afterLogin" hooks
+	Accounts.onLogin(async ({ user }: { user: any }) =>
```
murtaza98, line 21:
```suggestion
	Accounts.onLogin(async ({ user }: { user: IUser }) =>
```

murtaza98: Minor typing change

murtaza98 (dismissed):

murtaza98:
@KevLehman Can you think of any API tests we can add here to verify this fix?

murtaza98 (dismissed):

apps/meteor/app/livechat/server/business-hour/index.ts
```diff
@@ -16,8 +17,11 @@ Meteor.startup(async () => {
...
 
-	Accounts.onLogin(
-		async ({ user }: { user: any }) =>
-			user?.roles?.includes('livechat-agent') && !user?.roles?.includes('bot') && businessHourManager.onLogin(user._id),
+	// Note: This shouldn't happen directly on login, but after the login process has already finished
+	// but meteor doesn't have "afterLogin" hooks
+	Accounts.onLogin(async ({ user }: { user: IUser }) =>
+		setTimeout(async () => {
+			void (user?.roles?.includes('livechat-agent') && !user?.roles?.includes('bot') && businessHourManager.onLogin(user._id));
+		}, 1000),
```
ggazzo, line 25: could you explain why? i dont think timeouts are the best approach to solve this kind of situation we know that 1s can be too much in some cases , but with in a slow env can be not enough we dont want to gamble the agent availability

ggazzo (dismissed):

murtaza98 (dismissed):

apps/meteor/app/livechat/server/business-hour/index.ts
KevLehman: Yep. My understanding was that the `status` was being set by meteor on login, but after a lookup, found that presence is the one in charge. 1s should be enough in most cases for waiting the `status` to be set, however, when presence runs as MS, this could take longer (though this taking longer than 1s should be worrysome)

Changed the impl to use an event after the presence service sets the status prop, so should handle those cases too

murtaza98 (approved):

ggazzo (approved):

Issue #29319, opened 2023-05-22
SlackBridge error receiving messages but sending is fine


dmotors:
### Description:

Server is running on Snap installation.

Ever since 5/17/23 no messages from Slack has been received. When I send a message from Rocketchat it goes to Slack fine. No settings has been altered in the SlackBridge settings.

### Steps to reproduce:

Send message from Slack.

### Expected behavior:

Messages sent from Slack should be sent to Rocketchat channel with invited bot user.

### Actual behavior:

No messages received.

### Server Setup Information:

- Version of Rocket.Chat Server: 6.2.2
- Operating System: Windows 11
- Deployment Method: snap
- Number of Running Instances: 1
- DB Replicaset Oplog: 
- NodeJS Version: v14.21.2
- MongoDB Version: 5.0.15

### Client Setup Information

- Desktop App or Browser Version: 3.9.4
- Operating System: Windows 11

### Additional context

Things that can trigger error logs:
- Adding/Removing user
- Sending message from Rocket
- Enabling/disabling SlackBridge

### Relevant logs:

```
[...]

PR #29592, opened 2023-06-20
fix: slackbridge was not receiving updates from Slack events anymore

MarcosSpessatto:

Jira: SUP-267
Closes https://github.com/RocketChat/Rocket.Chat/issues/29319


apps/meteor/app/slackbridge/server/RocketAdapter.js
```diff
@@ -228,11 +228,11 @@ export default class RocketAdapter {
 	}
 
 	async getChannel(slackMessage) {
-		return slackMessage.channel ? this.findChannel(slackMessage.channel) || this.addChannel(slackMessage.channel) : null;
+		return slackMessage.channel ? (await this.findChannel(slackMessage.channel)) || this.addChannel(slackMessage.channel) : null;
 	}
 
 	async getUser(slackUser) {
-		return slackUser ? this.findUser(slackUser) || this.addUser(slackUser) : null;
+		return slackUser ? (await this.findUser(slackUser)) || this.addUser(slackUser) : null;
```
lmauromb, line 235: notice that in other calls to `this.addUser` they put an `await` first, should we add one too?

MarcosSpessatto: No, in this case where we don't add the `await` is because we are returning the promise directly, otherwise, it'll be a redundant return await.

lmauromb (approved):

ggazzo (approved):

PR #29611, opened 2023-06-21
fix: cannot invite LDAP users via `/invite-all` slashcommand

tapiarafael:

It wasn't possible to invite LDAP users through `/invite-all-from` and `/invite-all-to` slashcommands since the users were without usernames in the Subscription model.

With this fix, when joining in the default room, all the necessary fields will be stored in the DB, avoiding problems when inviting users in the future.

## Further comments
[SUP-11](https://rocketchat.atlassian.net/browse/SUP-11)

[SUP-11]: https://rocketchat.atlassian.net/browse/SUP-11?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/meteor/app/importer/server/classes/ImportDataConverter.ts
```diff
@@ -303,6 +303,8 @@ export class ImportDataConverter {
 		}
 		await this.updateUser(user, userData);
 
+		Object.assign(user, userData);
```
pierre-lehnen-rc, line 306: Those two objects do not have the same type for all attributes. Instead of assigning the data directly here, you should load the user from the db again.

pierre-lehnen-rc (changes requested):

matheusbsilva137 (approved):

pierre-lehnen-rc:
The changes from this PR were included on #29935

PR #29652, opened 2023-06-27
regression: Livechat UiKit button not working

murtaza98:
## Issue(s)

https://rocketchat.atlassian.net/browse/OC-1099

Livechat was not passing visitor authentication headers while sending UIKit button events to server.


packages/livechat/src/lib/uiKit.js
```diff
@@ -104,16 +104,22 @@ export const triggerAction = async ({ appId, type, actionId, rid, mid, viewId, c
...
+					rid,
+					mid,
+					viewId,
+					container,
+					triggerId,
+					payload,
+				},
+				{
+					appId,
+					visitorToken: Livechat.credentials.token,
```
ggazzo, line 120: what is the advantage of having a method in the sdk that cannot even pass the automatic token to you?

ggazzo (dismissed):

packages/ddp-client/src/livechat/LivechatClientImpl.ts
```diff
@@ -324,6 +324,13 @@ export class LivechatClientImpl extends DDPSDK implements LivechatStream, Livech
 		});
 	}
 
+	async sendUiInteraction(
+		payload: OperationParams<'POST', '/apps/ui.interaction/:id'>,
+		{ appId, visitorToken }: { appId: string; visitorToken: string },
```
MartinSchoeler, line 329: Instead of sending the token via params, use `this.token`, you can even check to see if the token exists and send the appropriate error message

MartinSchoeler (changes requested):

packages/ddp-client/src/livechat/LivechatClientImpl.ts
murtaza98: This was helpful! Thanks :hugs:

KevLehman (approved):

KevLehman (approved):

aleksandernsilva (approved):

Issue #29637, opened 2023-06-23
File uploads even size limit exceed


olodar:
### Description:

For example, I have set limit for file to 15 MiB:
![Screenshot from 2023-06-23 20-28-17](https://github.com/RocketChat/Rocket.Chat/assets/32993068/22931494-8b7a-484f-a8e9-0472c281da89)
But users anyway can send files with higher size, and files fully uploads to server, after uploading file simply drops, no any notification about exceed file size or error.
<img width="713" alt="2023-06-23_20-28" src="https://github.com/RocketChat/Rocket.Chat/assets/32993068/ff2595f3-4681-4c48-bd71-8b6dd2bc5d85">

### Steps to reproduce:

1.  Set file upload limit
2.  Try to upload file excees limit you set

### Expected behavior:

There's should be some notification when I trying to send file with higher size, for example "File X excess file upload limit (15 MiB)", and it should be shown BEFORE uploading.

### Actual behavior:

Rocketchat accepts big files, uploads it fully then drops

### Server Setup Information:

- Version of Rocket.Chat Server: 6.2.8
- Operating System: Ubuntu 22.04
- Deployment Method: tar
- Number of Running Instances: 1
- DB Replicaset Oplog: ?
- NodeJS Version: 14.21.3
- MongoDB Version: 6.0.6

### Client Setup Information

- Desktop App or Browser Version: Chromium 111.0.5563.65
- Operating System: Ubuntu 22.04

sivasathyaseeelan:
Hi, I would like to work on this issue, can you please assign it to me?

olodar:
> Hi, I would like to work on this issue, can you please assign it to me?

Hi, how can I do that?

hugocostadev:
No need to ask for permission to fix the issue, feel free to work on it! 
Just check if someone it's already raised a PR for that (should be linked in this issue).
Best of luck 🚀 👏 

---

**Questions? Help needed? Feature Requests?** 

- Join our [Open Server](https://open.rocket.chat/) in the #support channel and feel free to raise a question
- Join our [Community Forum](https://forums.rocket.chat/) and search/create a post there
- Feature Request: Open an issue in [Feature Request](https://github.com/RocketChat/feature-requests) repository

Aditay2002:
@hugocostadev should i use some react library to provide the notification or just give an alert about the file limit exceeds?

Sayan4444:
Please close this issue as it has been solved

PR #29677, opened 2023-06-29
fix: File uploads even size limit exceed

AdityaSingh-02:

https://github.com/RocketChat/Rocket.Chat/assets/94185167/1c9f3898-27b6-4ca7-abb3-001698c95033

rc2

## Issue(s)

solves #29637  
## Further comments

SUP-413


olodar:
Good. But I think it will be better notify after you selected file, i.e. before File upload Window.

AdityaSingh-02:
> Good. But I think it will be better notify after you selected file, i.e. before File upload Window.

Hey The file gets uploaded when you click the send button and before that it's just a preview, so it works

AdityaSingh-02:
> Good. But I think it will be better notify after you selected file, i.e. before File upload Window.

hmm, I thought about that but I wasn't able to do that, The Error Toast wasn't popping up while doing that so I had to take this approach

yash-rajpal (approved):

PR #29683, opened 2023-06-30
fix: verifying if both app and environment is enterprise to enable

AllanPazRibeiro:
This PR adds verification of the "isEnterpriseOnly" flag for the app, ensuring that only enterprise apps can be enabled in an enterprise environment. The PR includes the following changes:

In the frontend code (`AppMenu.js`):

- Checks if the license is enterprise using the isEnterprise() method.
- Renders the "Enable App" button only if the app is marked as enterprise-only and the license is enterprise.

In the backend code (`:id/status` controller):

- Uses the isEnterprise() method to determine if the license is enterprise.
- Verifies if the app is marked as enterprise-only and the license is an enterprise by requesting it to the `/v1/apps/${appId}?appVersion=${version}` marketplace API endpoint.
- If both conditions are met, proceeds with enabling the app.
- If the app is enterprise-only but the environment is not valid (not enterprise or no longer enterprise), returns an error with the message "Invalid environment for enabling an enterprise app".

These changes ensure that only enterprise apps can be enabled in a valid enterprise environment, providing better control and preventing invalid configurations.

## Steps to test or reproduce

1. On an enterprise env, install some enterprise apps (Google Calendar/drive, Instagram....)
2. Change the env from enterprise to free version
3. Try to enable the app, it shouldn't be possible since the button will not be visible

With that in mind, you can test the backend part by performing a `curl`, for example:
```
curl -X POST \
   <your-env>/apps/:<app-id>/status
  -H "Content-Type: application/json" \
  -H "x-auth-token: <auth-token>" \
  -H "x-user-id: <user-id>" \
  --data-raw '{"status":"manually_enabled"}' \
  --compressed
```
If the env is not enterprise anymore you should face an error that might look like: "'Invalid environment for enabling enterprise app'"
## Further comments

AECO-275


apps/meteor/client/views/marketplace/AppMenu.js
```diff
@@ -334,20 +341,19 @@ function AppMenu({ app, isAppDetailsPage, ...props }) {
...
-							<>
-								<Icon name='check' size='x16' marginInlineEnd='x4' />
-								{t('Enable')}
-							</>
-						),
-						disabled: !app.migrated && !appCountQuery?.data?.hasUnlimitedApps && appCountQuery?.data?.enabled >= appCountQuery?.data?.limit,
-						action: handleEnable,
-					},
-				}),
+			...(shouldBeAbleToEnable && {
```
tiagoevanp, line 344: I think is better to use `is[something]` to name boolean variables... `isEnableable`

apps/meteor/client/views/marketplace/AppMenu.js
```diff
@@ -294,6 +298,9 @@ function AppMenu({ app, isAppDetailsPage, ...props }) {
 			}),
 		};
 
+		const isEnterpriseFeasible = app.isEnterpriseOnly && isEnterprise;
```
tiagoevanp, line 301: I think you could remove `Feasible`...

apps/meteor/client/views/marketplace/AppMenu.js
```diff
@@ -40,6 +41,9 @@ function AppMenu({ app, isAppDetailsPage, ...props }) {
 	const buildExternalUrl = useEndpoint('GET', '/apps');
 	const syncApp = useEndpoint('POST', `/apps/${app.id}/sync`);
 	const uninstallApp = useEndpoint('DELETE', `/apps/${app.id}`);
+	const { data } = useIsEnterprise();
+
+	const isEnterprise = data?.isEnterprise ?? false;
```
tiagoevanp, line 46: I Would Suggest to add in the end of the variable name `License`... to make it more clear

apps/meteor/client/views/marketplace/AppMenu.js
```diff
@@ -40,6 +41,9 @@ function AppMenu({ app, isAppDetailsPage, ...props }) {
 	const buildExternalUrl = useEndpoint('GET', '/apps');
 	const syncApp = useEndpoint('POST', `/apps/${app.id}/sync`);
 	const uninstallApp = useEndpoint('DELETE', `/apps/${app.id}`);
+	const { data } = useIsEnterprise();
+
+	const isEnterpriseLicense = data?.isEnterprise ?? false;
```
AlenDavid, line 46: is it possible for data.isEnterprise to always be a boolean? so we can avoid the increase in complexity by adding ?? false here.

apps/meteor/client/views/marketplace/AppMenu.js
```diff
@@ -334,20 +341,19 @@ function AppMenu({ app, isAppDetailsPage, ...props }) {
...
-							<>
-								<Icon name='check' size='x16' marginInlineEnd='x4' />
-								{t('Enable')}
-							</>
-						),
-						disabled: !app.migrated && !appCountQuery?.data?.hasUnlimitedApps && appCountQuery?.data?.enabled >= appCountQuery?.data?.limit,
-						action: handleEnable,
-					},
-				}),
+			...(isPossibleToEnableApp && {
```
AlenDavid, line 344: Would it be possible to refactor this object to be created outside this lines and avoid having a high cyclomatic complexity in here?

AlenDavid (commented):
Left a few comments!

apps/meteor/ee/server/apps/communication/rest.ts
```diff
@@ -1124,14 +1124,34 @@ export class AppsRestApi {
...
+					const appId = this.urlParams.id;
+					const { version } = this.bodyParams;
+
+					const headers = getDefaultHeaders();
+
+					const request = await fetch(`${baseUrl}/v1/apps/${appId}?appVersion=${version}`, {
+						headers,
+					});
+
+					const [data] = await request.json();
```
KevLehman, line 1137: Check if response is ok before consuming

apps/meteor/ee/server/apps/communication/rest.ts
```diff
@@ -1124,14 +1124,34 @@ export class AppsRestApi {
 						return API.v1.failure('Invalid status provided, it must be "status" field and a string.');
 					}
 
+					const baseUrl = orchestrator.getMarketplaceUrl();
+					const appId = this.urlParams.id;
+					const { version } = this.bodyParams;
+
+					const headers = getDefaultHeaders();
+
+					const request = await fetch(`${baseUrl}/v1/apps/${appId}?appVersion=${version}`, {
```
KevLehman, line 1133: Pass `appVersion` as query param instead of on the URL

apps/meteor/ee/server/apps/communication/rest.ts
```diff
@@ -1124,14 +1124,34 @@ export class AppsRestApi {
 						return API.v1.failure('Invalid status provided, it must be "status" field and a string.');
 					}
 
+					const baseUrl = orchestrator.getMarketplaceUrl();
+					const appId = this.urlParams.id;
```
KevLehman, line 1128: You can destructure and rename in place too

apps/meteor/ee/server/apps/communication/rest.ts
```diff
@@ -1124,14 +1124,34 @@ export class AppsRestApi {
...
+
 					const prl = manager.getOneById(this.urlParams.id);
 
 					if (!prl) {
 						return API.v1.notFound(`No App found by the id of: ${this.urlParams.id}`);
 					}
 
+					const storedApp = prl.getStorageItem();
+
+					if (![AppStatus.DISABLED, AppStatus.MANUALLY_DISABLED].includes(this.bodyParams.status)) {
```
KevLehman, line 1147: Add `status` to the destructure of `bodyParams`

KevLehman (changes requested):

apps/meteor/client/views/marketplace/AppMenu.js
AllanPazRibeiro: Since the data can be undefined, I am just guaranteeing that this variable is at least false

apps/meteor/client/views/marketplace/AppMenu.js
AllanPazRibeiro: Where do you think that could be a good place for this object?

apps/meteor/client/views/marketplace/AppMenu.js
d-gubert: I don´t think this is a necessary rule-of-thumb when naming booleans. `isSomething`, `canDoSomething`, `hasSomething` and I guess even `shouldSomething` are valid names that clearly send the message they are booleans IMO

d-gubert (changes requested):
I'm wondering if sending the version from client to server when using the `/:id/status` endpoint is the best approach. Why do we need to send this info? It should be available already on the server, because the app is already installed, right?

AllanPazRibeiro:
> I'm wondering if sending the version from client to server when using the `/:id/status` endpoint is the best approach. Why do we need to send this info? It should be available already on the server, because the app is already installed, right?

The app version is needed to perform the request to the marketplace API.

I don't like this approach as well, but I didn't find any better way, do you have a better approach in mind?

apps/meteor/ee/server/apps/communication/rest.ts
```diff
@@ -1124,14 +1124,42 @@ export class AppsRestApi {
...
+
+					mktAppsUrl.searchParams.set('appVersion', String(version));
+
+					const request = await fetch(mktAppsUrl.toString(), {
+						headers,
+					});
+
+					if (!request.ok) {
+						return API.v1.failure('Unable to retrieve app from marketplace');
+					}
```
d-gubert, line 1143: Can you confirm what `request.ok` means? The marketplace-api will respond with a `404` if the appId is not found. This could prevent private apps from having their status changed via the user

apps/meteor/ee/server/apps/communication/rest.ts
```diff
@@ -1124,14 +1124,42 @@ export class AppsRestApi {
...
+
+					const headers = getDefaultHeaders();
+
+					const mktAppsUrl = new URL(`${baseUrl}/v1/apps/${appId}`);
+
+					mktAppsUrl.searchParams.set('appVersion', String(version));
+
+					const request = await fetch(mktAppsUrl.toString(), {
+						headers,
+					});
```
d-gubert, line 1139: Technically, we have the `response` assigned to the variable here, not the `request` :stuck_out_tongue: But what I want to point out here is: you're making the request _first thing when the endpoint is hit_ - however we just want to prevent apps from _being enabled_ if they are enterprise only and the workspace doesn´t have a valid license anymore.

You could prevent unnecessary network calls if you just did this in the cases where it is required

d-gubert (changes requested):

apps/meteor/client/views/marketplace/AppMenu.js
tiagoevanp: There is another option by casting using boolean cast `!!data?.isEnterprise`

apps/meteor/ee/server/apps/communication/rest.ts
```diff
@@ -1124,14 +1128,48 @@ export class AppsRestApi {
...
+						return API.v1.notFound(`No App found by the id of: ${appId}`);
+					}
+
+					const storedApp = prl.getStorageItem();
+
+					const { version } = prl.getInfo();
+
+					const { isEnterpriseOnly: isStoredAppEnterpriseOnly } = storedApp.marketplaceInfo as IMarketplaceInfo;
+
+					let isAppEnterpriseOnly = isStoredAppEnterpriseOnly ?? false;
```
d-gubert, line 1148: In the case where this is true, do we still need to make the request to the `marketplace-api`? For me it seems that we don't but I may be missing something

apps/meteor/ee/server/apps/communication/rest.ts
```diff
@@ -1124,14 +1128,48 @@ export class AppsRestApi {
...
+						});
+
+						if (appInfoResponse?.ok) {
+							const [data] = await appInfoResponse.json();
+
+							isAppEnterpriseOnly = data?.isEnterpriseOnly;
+						}
+
+						if (![AppStatus.DISABLED, AppStatus.MANUALLY_DISABLED].includes(status)) {
+							if (isAppEnterpriseOnly) {
```
d-gubert, line 1165: There doesn't seem to be any benefits on splitting this if statement. Please prefer using the more concise `&` operator

apps/meteor/ee/server/apps/communication/rest.ts
```diff
@@ -1124,14 +1128,48 @@ export class AppsRestApi {
 						return API.v1.failure('Invalid status provided, it must be "status" field and a string.');
 					}
 
-					const prl = manager.getOneById(this.urlParams.id);
+					const baseUrl = orchestrator.getMarketplaceUrl();
```
d-gubert, line 1131: This variable, the `headers` one on line 1135, `version` on line 1144, `isStoredAppEnterpriseOnly` on line 1146 and `isAppEnterpriseOnly` on line 1148 are all used _exclusively_ inside the if statement that start on line 1150 - declaring them outside of that block makes code harder to read in the case they are irrelevant and they consume resources even when they're not used.

Please move them into the scope where they're used

apps/meteor/ee/server/apps/communication/rest.ts
```diff
@@ -1124,14 +1128,48 @@ export class AppsRestApi {
...
+							if (isAppEnterpriseOnly) {
+								return API.v1.failure('Invalid environment for enabling enterprise app');
+							}
+						}
 					}
 
-					if (AppStatusUtils.isEnabled(this.bodyParams.status)) {
-						if (!(await canEnableApp(prl.getStorageItem()))) {
+					if (AppStatusUtils.isEnabled(status)) {
+						if (!(await canEnableApp(storedApp))) {
```
d-gubert, line 1172: Take the opportunity to improve this if statement as well as mentioned above

apps/meteor/ee/server/apps/communication/rest.ts
```diff
@@ -413,6 +415,7 @@ export class AppsRestApi {
 					}
 
 					void orchestrator.getNotifier().appAdded(info.id);
+					console.log('🚀 ~ file: rest.ts:419 ~ AppsRestApi ~ post ~ info:', info);
```
d-gubert, line 418:
```suggestion
```

apps/meteor/ee/server/apps/communication/rest.ts
```diff
@@ -880,6 +883,7 @@ export class AppsRestApi {
 						const request = await fetch(`${baseUrl}/v1/workspaces/${workspaceIdSetting.value}/apps/${this.urlParams.id}`, { headers });
 						statusCode = request.status;
 						result = await request.json();
+						console.log('🚀 ~ file: rest.ts:883 ~ AppsRestApi ~ post ~ result:', result);
```
d-gubert, line 886:
```suggestion
```

d-gubert (changes requested):

d-gubert (changes requested):

apps/meteor/ee/server/apps/communication/rest.ts
```diff
@@ -1120,24 +1120,57 @@ export class AppsRestApi {
...
+							return API.v1.failure('Invalid environment for enabling enterprise app');
 						}
 					}
 
-					const result = await manager.changeStatus(prl.getID(), this.bodyParams.status);
+					if (AppStatusUtils.isEnabled(status) && !(await canEnableApp(storedApp))) {
+						return API.v1.failure('Enabled apps have been maxed out');
+					}
 
+					const result = await manager.changeStatus(prl.getID(), status);
```
KevLehman, line 1173: You can abstract this to a func, and make this easier to read:

```
if (isEnterprise)
	doTheCheckHere()
	return

doWhateverIsInsideTheIfRightNow()

File: apps/meteor/ee/server/apps/communication/rest.ts [hunk 19]
```diff
@@ -1120,24 +1120,57 @@ export class AppsRestApi {
...
+								headers,
+							});
+
+							if (!appInfoResponse.ok) {
+								const result = await appInfoResponse.json();
+								throw new Error(result?.error || 'Error fetching app information from the Marketplace.');
+							}
+
+							const [data] = await appInfoResponse.json();
+							isAppEnterpriseOnly = data?.isEnterpriseOnly;
```
KevLehman, line 1158: This could be its own function, something like "getMarketplaceInfoForApp(appId)"

KevLehman (changes requested):

apps/meteor/ee/server/apps/communication/rest.ts
```diff
@@ -1120,24 +1120,57 @@ export class AppsRestApi {
...
+					if (!isEnterprise()) {
+						const { version } = prl.getInfo();
+						const baseUrl = orchestrator.getMarketplaceUrl();
+						const headers = getDefaultHeaders();
+
+						const appInfosURL = new URL(`${baseUrl}/v1/apps/${appId}`);
+						appInfosURL.searchParams.set('appVersion', String(version));
+
+						try {
+							const appInfoResponse = await fetch(appInfosURL.toString(), {
```
d-gubert, line 1148: This request fails if the app is not found in the marketplace-api. This means calls to this endpoint for private apps will not work, as pointed out before https://github.com/RocketChat/Rocket.Chat/pull/29683#discussion_r1256327089

d-gubert (changes requested):

apps/meteor/client/views/marketplace/AppMenu.js
```diff
@@ -140,7 +144,7 @@ function AppMenu({ app, isAppDetailsPage, ...props }) {
 		const confirm = async () => {
 			closeModal();
 			try {
-				const { status } = await setAppStatus({ status: 'manually_disabled' });
+				const { status } = await setAppStatus({ status: 'manually_disabled', version: app.version });
```
d-gubert, line 147: The field `version` shouldn't be sent anymore to the endpoint - remove all usages

d-gubert (changes requested):

apps/meteor/ee/server/apps/marketplace/appEnableCheck.ts
```diff
@@ -0,0 +1,44 @@
...
+
+	const appInfosURL = new URL(`${baseUrl}/v1/apps/${appId}`);
+	appInfosURL.searchParams.set('appVersion', String(version));
+
+	try {
+		const { isEnterpriseOnly } = await getMarketplaceAppInfo({ baseUrl, headers, appId, version });
+
+		if (isEnterpriseOnly) {
+			isAppEnterpriseOnly = isEnterpriseOnly;
+		}
```
KevLehman, line 35: Isn't this if the same as just `isAppEnterpriseOnly = isEnterpriseOnly`? 👀

apps/meteor/ee/server/apps/marketplace/appEnableCheck.ts
```diff
@@ -0,0 +1,42 @@
...
+	marketplaceInfo,
+}: {
+	baseUrl: string;
+	headers: Record<string, any>;
+	appId: string;
+	version: string;
+	logger: Logger;
+	status: AppStatus;
+	marketplaceInfo?: IMarketplaceInfo;
+}) => {
```
AlenDavid, line 23: Can you put this in an interface? Its consuming quite a sum of lines

apps/meteor/ee/server/apps/communication/rest.ts
```diff
@@ -1120,24 +1122,46 @@ export class AppsRestApi {
...
+					const { id: appId } = this.urlParams;
+					const { status } = this.bodyParams;
+
+					if (!status || typeof status !== 'string') {
 						return API.v1.failure('Invalid status provided, it must be "status" field and a string.');
 					}
 
-					const prl = manager.getOneById(this.urlParams.id);
-
+					const prl = manager.getOneById(appId);
```
AlenDavid, line 1132: Can we use a better naming for this variable? prl is not giving the idea of what this is responsible for.

AlenDavid (commented):
Left some comments!

AlenDavid (dismissed):
Good work!

d-gubert (dismissed):

KevLehman (approved):

d-gubert (changes requested):
Please add a changeset to the PR

.changeset/small-rice-repair.md
```diff
@@ -0,0 +1,5 @@
+---
+'@rocket.chat/meteor': minor
+---
+
+Avoid an enterprise app to be enabled in a non-enterprise environment
```
d-gubert, line 5:
```suggestion
---
'@rocket.chat/meteor': patch
---

Fix validation in app status call that allowed Enterprise apps to be enabled in invalid environments
```

d-gubert (changes requested):

d-gubert (dismissed):

KevLehman (approved):

AlenDavid (approved):

tiagoevanp (approved):

tiagoevanp (approved):

d-gubert (approved):

AlenDavid (approved):

PR #29701, opened 2023-07-03
regression: unable to view reported messages of deleted users

Dnouv:
If a user deletes their own account or if the user account is deleted. Then, on the Moderation Console the Admin was not able to view the reported messages.
This PR introduces a fix for this issue.

If the "Delete all messages" is selected under the `Message Erasure` (`Setting>Messages`) setting, then if a user account is deleted, all the reported messages will be hidden, and they no longer show up on the Moderation Console.

If any other option is selected in the setting `Message Erasure`, then the reported messages will now show up in the Moderation Console, but there will be an additional Callout component, and some moderation actions against the user will be disabled since they are not possible to take on a deleted user.

> Why not delete all the reports against a message when the message is deleted?

Eren got trolled by a random user, Eren got sad, and he reported those messages, expecting to get some justice, but soon the troller deleted all those messages (now the reports will also be dismissed/hidden), and here Eren is left waiting for Moderator to take some action on this troller 😟 .

## Why introduce a new method

Before this PR is merged, there could be some edge cases where the user is already deleted along with all messages, and this will cause an error while deleting their messages (as these messages no longer exist in the `messages` collection but do show up in the `moderation_reports` collection because, at the time of reporting a message, we keep a snapshot of the message) from the Moderation Console.
The error occurs because before the `deleteReportedMessage` used to call the `Messages` model method `cloneAndSaveAsHistoryById`, which as written [here](https://github.com/RocketChat/Rocket.Chat/blob/c10137e015b2cfa42aa7976d88e512e2ed0ed34e/apps/meteor/server/models/raw/Messages.ts#L1011-L1012) returns an error if a message is not found. 

To tackle this issue, my first approach was inside the `deleteReportedMessages` function. I called the 
```ts
const existingMids = Messages.find({ _id: { $in: ids } }, {projection: {_id: 1}}).toArray() // where ids is an array of message-ids that are to be deleted

await Promise.all(existingMids.map((mid) => Messages.cloneAndSaveAsHistoryById(mid, user)))
```

however, the above approach loads all the messages into the memory, blocking the Event Loop. And after getting feedback from discussions with some more experienced members, we found a better way of doing this by introducing a new method, which takes in the `message` as an argument.

Advantages of the new method.
- Reduced the DB call by 50%; before, it used to be 2 trips per message to delete, now it's only 1.
- No need to load all messages into memory since the Mongo cursor loads the documents in batches.

this returns the messages

## Steps to test or reproduce

1. Create a temporary user (temp).
2. Log in as the temp user, go into any channel and send a message, and report this message.
3. Go to “My Account“ and click “Delete My Account“
4. Head over to the Moderation Console from an Admin account and try clicking on the row of the “temp” user.
5. The Context bar will remain blank and an error shows up `invalid-user`

## Further comments

[WM-120](https://rocketchat.atlassian.net/browse/WM-120)

[WM-120]: https://rocketchat.atlassian.net/browse/WM-120?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


tapiarafael (dismissed):

apps/meteor/app/api/server/v1/moderation.ts
```diff
@@ -73,9 +73,12 @@ API.v1.addRoute(
 
 			const { count = 50, offset = 0 } = await getPaginationItems(this.queryParams);
 
-			const user = await Users.findOneById(userId, { projection: { _id: 1 } });
-			if (!user) {
-				return API.v1.failure('error-invalid-user');
+			const existingReport = await ModerationReports.findOne(
+				{ 'message.u._id': userId, '_hidden': { $ne: true } },
+				{ projection: { _id: 1 } },
+			);
```
debdutdeb, line 79: Move this to the model.

apps/meteor/app/api/server/v1/moderation.ts
```diff
@@ -73,9 +73,12 @@ API.v1.addRoute(
...
+			if (!existingReport) {
+				return API.v1.failure('no-report-found-for-this-user');
```
debdutdeb, line 81:
```suggestion
				return API.v1.failure('no-report-found');
```

apps/meteor/app/api/server/v1/moderation.ts
```diff
@@ -126,9 +129,12 @@ API.v1.addRoute(
...
```
debdutdeb, line 137: same suggestions as previous

apps/meteor/app/lib/server/functions/deleteUser.ts
```diff
@@ -25,7 +26,9 @@ export async function deleteUser(userId: string, confirmRelinquish = false): Pro
 		projection: { username: 1, avatarOrigin: 1, roles: 1, federated: 1 },
 	});
 
-	if (!user) {
+	const actionUserId = Meteor.userId();
+
+	if (!user || !actionUserId) {
```
debdutdeb, line 31: pass the userid as an argument, avoid using meteor.

apps/meteor/server/lib/moderation/deleteReportedMessages.ts
```diff
@@ -11,15 +11,22 @@ export async function deleteReportedMessages(messages: IMessage[], user: IUser):
 	const showDeletedStatus = settings.get('Message_ShowDeletedStatus');
 	const files: string[] = [];
 	const messageIds: string[] = [];
-	for (const message of messages) {
+
+	const promises = messages.map(async (message) => {
+		const foundMessage = await Messages.findOneById(message._id, { projection: { _id: 1 } });
```
debdutdeb, line 16: I don't get this change. why do you have to check for every id?

debdutdeb (changes requested):
for `moderation.user.deleteReportedMessages` I'm wondering what happens if a user is deleted. Isn't all of their messages get deleted with them?

Dnouv:
Thank you for the review, @debdutdeb ! 

> for moderation.user.deleteReportedMessages I'm wondering what happens if a user is deleted. Isn't all of their messages get deleted with them?

Yes, you are partially correct; under the settings (`Settings>Messages>Message Erasure Type`), the WSpace Admin can configure what behavior they want, we provide three options:
1. **Keep Messages and User Name**: The message and files history of the user will be deleted from Direct Messages and will be kept in other rooms. 
2. **Delete All Messages**: All messages and files from the user will be deleted from the database and it will not be possible to locate the user anymore. 
3. **Remove link between user and messages**: This option will assign all messages and files of the user to Rocket.Cat bot and Direct Messages are going to be deleted.

And in this PR, I have added a few lines that will hide reports if option 2 is selected for the workspace and user account deletion is initiated.

And to manage the edge cases where the user was already deleted along with all of their messages before this PR was merged, I have modified the existing logic in `deleteReportedMessages` function. What it does is calls the `Messages.findOneById` and if the message is not found we don't take any action on that.

Thank you!

apps/meteor/app/lib/server/functions/deleteUser.ts
Dnouv: That would require modifying the existing `deleteUser` function, and so potentially I will have to modify the files which call it, would that be ok?

debdutdeb:
> And in this PR, I have added a few lines that will hide reports if option 2 is selected for the workspace and user account deletion is initiated.

That's a good catch. But imo should be done for all types. How a message is deleted shouldn't matter - if it's deleted it's not viewabl thus reports should be dismissed. 

> And to manage the edge cases where the user was already deleted along with all of their messages before this PR was merged, I have modified the existing logic in deleteReportedMessages function. What it does is calls the Messages.findOneById and if the message is not found we don't take any action on that.

This part I dislike. It's a lot of roundtrips. What happens if an id is invalid in the following operations? Aren't they ignored (deleteMany, updateMany)?

Dnouv:
> This part I dislike. It's a lot of roundtrips. What happens if an id is invalid in the following operations? Aren't they ignored (deleteMany, updateMany)?

I don't know why, but this was the first time I saw an error thrown inside a Model 

https://github.com/Dnouv/Rocket.Chat/blob/1424fa8bd64e684f36ffe59c23af574b25ea0e82/apps/meteor/server/models/raw/Messages.ts#L1011

Dnouv:
> But imo should be done for all types. How a message is deleted shouldn't matter - if it's deleted it's not viewabl thus reports should be dismissed.

No, actually, sorry but I personally think we should not do this, I thought from a Moderator's POV:

Eren got trolled by a random user, Eren got sad, and he reported those messages, expecting to get some justice, but soon the troller deletes all those messages, and here Eren is left waiting for Moderator to take some action on this troller 😟 .

apps/meteor/app/api/server/v1/moderation.ts
Dnouv: Thank you! I will create a new method to handle this.

apps/meteor/app/lib/server/functions/deleteUser.ts
Dnouv: Here we need the uid of the user who is calling this function. Because while hiding the reported messages, the moderator's id is stored along with the action.

If a user deletes their own account, we can use the passed `user._id`, but if someone else calls the `deleteUser`, the userId of the user whose account is going to be deleted is passed on, and we don't need that. 

I hope that makes things clear. Thank you!

tapiarafael (dismissed):

hugocostadev (dismissed):

debdutdeb (changes requested):

apps/meteor/server/lib/moderation/deleteReportedMessages.ts
```diff
@@ -20,9 +21,15 @@ export async function deleteReportedMessages(messages: IMessage[], user: IUser):
...
+
 	if (keepHistory) {
 		if (showDeletedStatus) {
-			await Promise.all(messageIds.map((id) => Messages.cloneAndSaveAsHistoryById(id, user as any)));
+			const existingMessages = await Messages.find({ _id: { $in: messageIds } }).toArray();
+
+			if (existingMessages.length === 0) {
+				return;
+			}
+			await Promise.all(existingMessages.map((msg) => Messages.cloneAndSaveAsHistoryById(msg._id, user as any)));
```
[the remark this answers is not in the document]
debdutdeb: adding a new method won't change existing behaviour.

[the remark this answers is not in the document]
debdutdeb: actually nvm no need for another method. my bad

hugocostadev (approved):

KevLehman:
Isn't this like too big for a regression? 👀

Dnouv:
Is it? There is a fix in this PR related to i18n (the WM-143), which was classified as a regression.

tapiarafael (approved):

PR #29705, opened 2023-07-03
fix: Omnichannel Visitor Abandonment feature not working for apps

murtaza98:

Visitor abandonment feature used to rely on `room.v.lastMessageTs` property to calculate the predicted time when the chat should be considered abandoned. Now when apps were updating the room properties like customFields, then the app converters were overriding lastMessageTs property ( [reference](https://github.com/RocketChat/Rocket.Chat/blob/74aa6770881eb620a2275b84c55465d7552e4597/apps/meteor/app/apps/server/converters/rooms.js#L41) ), thus making this abandonment feature not work for rooms created from apps. 

So to solve this issue, I may have updated the app's converters to not override lastMessageTs. However I did not go that route since I didn't find any significant reference to `room.v.lastMessageTs` - hence I felt there was no need to store and maintain this property. There's only one reference where this property was being used and I've updated it to query the database direct, and get the lastMessage timestamp from source.

## Issue(s)

https://rocketchat.atlassian.net/browse/SUP-66


apps/meteor/server/models/raw/Messages.ts
```diff
@@ -1649,4 +1649,13 @@ export class MessagesRaw extends BaseRaw<IMessage> implements IMessagesModel {
 		};
 		return this.updateOne(query, update);
 	}
+
+	findVisitorLastMessageByRoomId(rid: string): Promise<IMessage | null> {
+		const query = {
+			rid,
+			token: { $exists: true },
```
KevLehman, line 1656: :hmmm: not sure if we need a `rid + token` index for this

KevLehman (changes requested):

apps/meteor/server/models/raw/Messages.ts
murtaza98: You're right, I believe we'll need this index. However considering this is Messages collection which is always growing rapidly, I'd like to be cautious and not add this index because this is only needed for a single feature and majority of the application would not use this index. So let me try exploring other ways we can fix this issue

murtaza98:
Closing this in favour of https://github.com/RocketChat/Rocket.Chat/pull/29724 . The approach which I was taking here would require adding an sparingly used index on Messages collection which is likely the biggest collection we have on DB. This index would not be used by majority of the application, rather it would be used for for a single visitor abandonment feature. Hence I've tried to solve the underlying issue with a different approach on the new PR which involves preventing app's converters to override the lastMessageTs field

PR #29717, opened 2023-07-04
fix: VoIP call UI becoming non responsive

aleksandernsilva:


MartinSchoeler (dismissed):

apps/meteor/client/providers/CallProvider/CallProvider.tsx
dougfabris, line 1: Why not use `useCustomSound` in this React provider?

aleksandernsilva: Old implementation, didn't know the hook existed. I took the opportunity and changed a few things.

MartinSchoeler (approved):

PR #29719, opened 2023-07-04
chore: throttle exceptions counter increment

sampaiodiego:

Throttle down uncaught exceptions counter to every 10 seconds. This is to avoid too many operations into the settings collection.


rodrigok (approved):

packages/tools/src/counter.ts
```diff
@@ -0,0 +1,20 @@
...
+		},
+		wait,
+		{ leading: false },
+	);
+
+	return () => {
+		counter++;
+		throttledFn();
+	};
+}
```
ggazzo, line 20: I dont think this should be placed here, its not quite useful outside that scope

not sure about the `lodash` necessity as well

apps/meteor/app/error-handler/server/lib/RocketChat.ErrorHandler.ts
```diff
@@ -1,9 +1,14 @@
 import { Meteor } from 'meteor/meteor';
 import { Settings, Users, Rooms } from '@rocket.chat/models';
+import { throttledCounter } from '@rocket.chat/tools';
 
 import { settings } from '../../../settings/server';
 import { sendMessage } from '../../../lib/server';
 
+const incException = throttledCounter((counter) => {
```
KevLehman, line 8: wdyt of ignoring this setting on the `watch.settings` watcher?

packages/tools/src/counter.ts
ggazzo, line 20:
```suggestion
export const throttle = <R, A extends any[]>(
    fn: (...args: A) => R,
    delay: number
) => {
    let wait = false;
    let timeout: undefined | number;

    return Object.assign(
        (...args: A) => {
            if (wait) return undefined;
  
            const val = fn(...args);
  
            wait = true;
  
            timeout = window.setTimeout(() => {
                wait = false;
            }, delay);
  
            return val;
        }, {
        stop: () => {
            clearTimeout(timeout);
        },
    });
};
```

apps/meteor/app/error-handler/server/lib/RocketChat.ErrorHandler.ts
```diff
@@ -1,9 +1,14 @@
 import { Meteor } from 'meteor/meteor';
 import { Settings, Users, Rooms } from '@rocket.chat/models';
+import { throttledCounter } from '@rocket.chat/tools';
```
ggazzo, line 3:
```suggestion
import { throttle } from '@rocket.chat/tools';
```

apps/meteor/app/error-handler/server/lib/RocketChat.ErrorHandler.ts
```diff
@@ -1,9 +1,14 @@
...
 
 import { settings } from '../../../settings/server';
 import { sendMessage } from '../../../lib/server';
 
+const incException = throttledCounter((counter) => {
+	Settings.incrementValueById('Uncaught_Exceptions_Count', counter).catch(console.error);
+}, 10000);
+
```
ggazzo, line 11:
```suggestion
const incException = (() => {
    let counter = 0;
    const commit = throttle((increment: number) => {
        console.log(counter);
        counter = 0;
    }, 10000);
    return () => {
        counter++;
        commit(counter);
    };
})()
```

apps/meteor/app/error-handler/server/lib/RocketChat.ErrorHandler.ts
sampaiodiego: the main issue is that we'd still receive it from mongo.. so we'd need to remove it on oplog/change stream query, which is not good 😬

ggazzo (approved):

PR #29720, opened 2023-07-04
feat: Delete read receipts on messages deletion

### LINKED ISSUE #27926: Not removed read receipts for non-existent messages

matheusbsilva137:

 - Delete read read receipts on messages deletion (for individual messaages, "prune messages" option or deleting rooms).

## Issue(s)

Fix #27926

## Steps to test or reproduce

- When erasing a single message, all of its corresponding read receipts should also de removed;
- When erasing a room, all of its messages and read receipts should also be removed;
- When using the "Prune Messages" options in a room, read receipts should be removed along with their corresponding messages (read receipts from messages that were not chosen to be deleted should remain).

## Further comments

The exceptions provided to the "Prune Messages" feature will only be applied to read receipts from new thread messages, discussions or messages pinned after updating to this PR (for old thread messages, discussions or messages pinned before the update, read receipts will be removed regardless of the exceptions chosen). That's an intended behavior, chosen to avoid huge migrations or slowliness in room deletions.

TC-457


apps/meteor/app/lib/server/functions/deleteMessage.ts
```diff
@@ -43,6 +43,7 @@ export async function deleteMessage(message: IMessage, user: IUser): Promise<voi
 
 	if (deletedMsg?.tmid) {
 		await Messages.decreaseReplyCountById(deletedMsg.tmid, -1);
+		await ReadReceipts.incrementThreadMessagesCountById(deletedMsg.tmid, -1);
```
KevLehman, line 46: Kinda confusing the name is `increment` but is actually `decrementing` 👀

apps/meteor/app/slackbridge/server/SlackAdapter.js
```diff
@@ -1006,13 +1006,12 @@ export default class SlackAdapter {
...
-				await Messages.setPinnedByIdAndUserId(
-					`slack-${slackMessage.attachments[0].channel_id}-${slackMessage.attachments[0].ts.replace(/\./g, '-')}`,
-					rocketMsgObj.u,
-					true,
-					new Date(parseInt(slackMessage.ts.split('.')[0]) * 1000),
-				);
+			if (!isImporting && slackMessage.attachments[0].channel_id && slackMessage.attachments[0].ts) {
+				const messageId = `slack-${slackMessage.attachments[0].channel_id}-${slackMessage.attachments[0].ts.replace(/\./g, '-')}`;
+				await Messages.setPinnedByIdAndUserId(messageId, rocketMsgObj.u, true, new Date(parseInt(slackMessage.ts.split('.')[0]) * 1000));
+				if (settings.get('Message_Read_Receipt_Store_Users')) {
```
KevLehman, line 1012: A suggestion, maybe having the setting check inside the service and decide there if doing something is better than checking for the setting on each place we want to use it? 👀

KevLehman (changes requested):

apps/meteor/app/lib/server/functions/cleanRoomHistory.ts
```diff
@@ -95,6 +95,16 @@ export async function cleanRoomHistory({
 		fromUsers,
 		ignoreThreads,
 	);
+
+	if (!limit) {
+		const uids = await Users.findByUsernames(fromUsers, { projection: { _id: 1 } })
+			.map((user: IUser) => user._id)
```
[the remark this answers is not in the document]
matheusbsilva137: Fixed

apps/meteor/app/lib/server/functions/deleteMessage.ts
matheusbsilva137: Done, thankss!

apps/meteor/app/slackbridge/server/SlackAdapter.js
matheusbsilva137: Since in this case I'm calling the raw model directly you mean adding the setting check in the raw model code? :thinking: I haven't seen us doing that in other places

KevLehman: No no, thought this was a call to the ReadReceipts service 😬 

If it's the model, there's nothing much we can do, unless we wrap the calls in the service. Not sure if that would be in scope for this one tho

apps/meteor/app/livechat/server/lib/Livechat.js
```diff
@@ -638,13 +639,18 @@ export const Livechat = {
...
 		}
 
-		await Subscriptions.removeByVisitorToken(token);
-		await LivechatRooms.removeByVisitorToken(token);
-		await LivechatInquiry.removeByVisitorToken(token);
+		await Promise.allSettled([
+			Subscriptions.removeByVisitorToken(token),
+			LivechatRooms.removeByVisitorToken(token),
+			LivechatInquiry.removeByVisitorToken(token),
+		]);
```
debdutdeb, line 653: imo we should log if any of these fails at least - silently getting ignored feels odd. Or is there a reason for ignoring? If so can you add a comment please?

.changeset/tame-pens-occur.md
```diff
@@ -0,0 +1,6 @@
+---
+"@rocket.chat/meteor": patch
+"@rocket.chat/model-typings": patch
+---
+
+fix: Delete read receipts on messages deletion
```
debdutdeb, line 6:
```suggestion
Deletes read receipts on messages deletion
```

debdutdeb (changes requested):

.changeset/tame-pens-occur.md
debdutdeb: another option could be `Fixed read receipts not getting deleted after corresponding message is deleted` or something like that - but idon't think this has to be a "fix" changelog entry - could very well be a new thing for the change log. 

either way - would be better to not repeat the pr title and make it more changelog friendly (think users reading it)

apps/meteor/app/livechat/server/lib/Livechat.js
matheusbsilva137: Hmm agree, I think it's ok to add it here. Done

apps/meteor/app/slackbridge/server/SlackAdapter.js
matheusbsilva137: Since the `MessageReads` service is only used for read receipts in threads as of now, I guess we can keep this out of scope for now. But it'd certainly be a great improvement once we include all of the read-receipts features in the `MessageReads` service (or any other)

apps/meteor/app/discussion/server/hooks/propagateDiscussionMetadata.ts
```diff
@@ -100,3 +100,13 @@ callbacks.add(
 	callbacks.priority.LOW,
 	'CleanDiscussionMessage',
 );
+
+callbacks.add(
+	'afterDeleteRoom',
+	async (rid) => {
+		await ReadReceipts.removeByRoomId(rid);
```
sampaiodiego, line 107: this file is not the best fit for this action, since this is not related to discussions.

apps/meteor/app/livechat/server/lib/Livechat.js
```diff
@@ -638,13 +639,32 @@ export const Livechat = {
...
 		for await (const room of cursor) {
-			await FileUpload.removeFilesByRoomId(room._id);
-			await Messages.removeByRoomId(room._id);
+			const result = await Promise.allSettled([
+				FileUpload.removeFilesByRoomId(room._id),
+				Messages.removeByRoomId(room._id),
+				ReadReceipts.removeByRoomId(room._id),
+			]);
+
+			for (const r of result) {
```
sampaiodiego, line 648: instead of looping through all promises, you could just use `Promise.all` and you'll get the error if any of them fails

apps/meteor/app/threads/server/functions.ts
```diff
@@ -21,6 +21,7 @@ export async function reply({ tmid }: { tmid?: string }, message: IMessage, pare
 	];
 
 	await Messages.updateRepliesByThreadId(tmid, addToReplies, ts);
+	await ReadReceipts.incrementThreadMessagesCountById(tmid);
```
sampaiodiego, line 24: you don't need to keep track of `tcount` for read receipts.. you just need to flag them as threads

sampaiodiego (changes requested):

apps/meteor/app/lib/server/functions/deleteMessage.ts
```diff
@@ -43,6 +43,7 @@ export async function deleteMessage(message: IMessage, user: IUser): Promise<voi
 
 	if (deletedMsg?.tmid) {
 		await Messages.decreaseReplyCountById(deletedMsg.tmid, -1);
+		await ReadReceipts.decrementThreadMessagesCountById(deletedMsg.tmid);
```
KevLehman, line 46: I'm curious, why we store this info in the readreceipts model?

apps/meteor/ee/server/models/raw/ReadReceipts.ts
```diff
@@ -16,4 +17,92 @@ export class ReadReceiptsRaw extends BaseRaw<ReadReceipt> implements IReadReceip
 	findByMessageId(messageId: string): FindCursor<ReadReceipt> {
 		return this.find({ messageId });
 	}
+
+	removeByUserId(userId: string): Promise<DeleteResult> {
```
KevLehman, line 21: Do we need an index on userId?

apps/meteor/ee/server/models/raw/ReadReceipts.ts
```diff
@@ -16,4 +17,92 @@ export class ReadReceiptsRaw extends BaseRaw<ReadReceipt> implements IReadReceip
...
+
+		if (users.length) {
+			query.userId = { $in: users };
+		}
+
+		return this.deleteMany(query);
+	}
+
+	setPinnedByMessageId(messageId: string, pinned?: boolean): Promise<Document | UpdateResult> {
+		if (pinned == null) {
```
KevLehman, line 91: In which cases can you pass null in as `pinned`?

Why cannot use a default value of `true` for pinned param?

KevLehman (changes requested):

apps/meteor/app/threads/server/functions.ts
matheusbsilva137: I first thought of keeping the counter in order to erase data from the new `message_reads` collection when the counter hits zero... (not in this PR, but we could do that at some point :thinking: )
but I'm not sure if we'd like to erase data from that collection, since once a message turns into a thread, we won't ever turn it into a regular message once again

matheusbsilva137: I'll change this to a flag. If we decide to delete data from the `message_reads` collection at some point too, we can use a flag to control that

apps/meteor/app/lib/server/functions/deleteMessage.ts
matheusbsilva137: I first thought of keeping the counter in order to erase data from the new `message_reads` collection when the counter hits zero... (not in this PR, but we could do that at some point thinking )
but I'm not sure if we'd like to erase data from that collection, since once a message turns into a thread, we won't ever turn it into a regular message once again. Anyways, I updated this to a flag now

apps/meteor/ee/server/models/raw/ReadReceipts.ts
matheusbsilva137: Done, I added indexes to `userId` and `roomId` too
cc @sampaiodiego

apps/meteor/ee/server/models/raw/ReadReceipts.ts
matheusbsilva137: Done!

apps/meteor/app/lib/server/functions/cleanRoomHistory.ts
```diff
@@ -86,7 +86,7 @@ export async function cleanRoomHistory({
 		}
 	}
 
-	const count = await Messages.removeByIdPinnedTimestampLimitAndUsers(
+	const { count, selectedMessageIds } = await Messages.removeByIdPinnedTimestampLimitAndUsers(
```
MarcosSpessatto, line 89: I believe this could lead to confusion, since `selectedMessageIds` will be undefined if limit is undefined or 0. We are indeed checking it below, but I think it should be more explicit, I'd say it would be better to separate the model method into 2, one for deleting and another one for querying. What is your opinion?

MarcosSpessatto (changes requested):

apps/meteor/app/slackbridge/server/SlackAdapter.js
```diff
@@ -1223,12 +1222,11 @@ export default class SlackAdapter {
 						],
 					};
 
-					await Messages.setPinnedByIdAndUserId(
-						`slack-${pin.channel}-${pin.message.ts.replace(/\./g, '-')}`,
-						msgObj.u,
-						true,
-						new Date(parseInt(pin.message.ts.split('.')[0]) * 1000),
-					);
+					const messageId = `slack-${pin.channel}-${pin.message.ts.replace(/\./g, '-')}`;
```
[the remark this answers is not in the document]
matheusbsilva137: Sounds good for me too. Done!

apps/meteor/server/models/raw/Messages.ts
```diff
@@ -1355,8 +1355,34 @@ export class MessagesRaw extends BaseRaw<IMessage> implements IMessagesModel {
...
+					_id: 1,
+				},
+				limit,
+			}).toArray()
+		).map(({ _id }) => _id);
+	}
+
+	async removeByIdPinnedTimestampLimitAndUsers(
+		rid: string,
+		pinned: boolean,
```
MarcosSpessatto, line 1370: I believe we should change this parameter to `ignore` or `exclude` pinned since it looks like the behavior intended here. Or the below filter `query.pinned = { $ne: true };` could be wrong, I'm not sure which is the expected behavior though.

matheusbsilva137: great suggestion, agreed. I changed it to `ignorePinned` (in the `find` also)

MarcosSpessatto (approved):

KevLehman (approved):

apps/meteor/server/models/raw/Messages.ts
```diff
@@ -1387,23 +1413,26 @@ export class MessagesRaw extends BaseRaw<IMessage> implements IMessagesModel {
...
+			query.pinned = { $ne: true };
+		}
+
+		if (ignoreDiscussion) {
+			query.drid = { $exists: false };
+		}
+
+		if (ignoreThreads) {
+			query.tmid = { $exists: false };
+			query.tcount = { $exists: false };
```
sampaiodiego, line 1432: `tcount` doesn't exist anymore

sampaiodiego (changes requested):

apps/meteor/ee/server/models/raw/ReadReceipts.ts
```diff
@@ -10,10 +11,92 @@ export class ReadReceiptsRaw extends BaseRaw<ReadReceipt> implements IReadReceip
...
+			query.pinned = { $ne: true };
+		}
+
+		if (ignoreDiscussion) {
+			query.drid = { $exists: false };
+		}
+
+		if (ignoreThreads) {
+			query.tmid = { $exists: false };
+			query.thread = { $ne: true };
```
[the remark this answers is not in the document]
matheusbsilva137: I replaced this `thread` flag by the `tmid` in threads' main messages too.

sampaiodiego (dismissed):

MarcosSpessatto (approved):

sampaiodiego (approved):

MarcosSpessatto (approved):

PR #29722, opened 2023-07-04
regression: Livechat not unsubscribing after the chat is closed

aleksandernsilva:
## Issue(s)

[OC-1117](https://rocketchat.atlassian.net/browse/OC-1117)

## Further comments

[OC-1117]: https://rocketchat.atlassian.net/browse/OC-1117?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/ddp-client/src/livechat/LivechatClientImpl.ts
```diff
@@ -359,6 +359,11 @@ export class LivechatClientImpl extends DDPSDK implements LivechatStream, Livech
 		return this.rest.put(`/v1/livechat/message/${id}`, params);
 	}
 
+	unsubscribeAll(): Promise<unknown> {
+		const subscriptions = Array.from(this.client.subscriptions.keys());
+		return Promise.all(subscriptions.map((subscription) => this.client.unsubscribe(subscription)));
```
KevLehman, line 364: Can we just 
```suggestion
		return Promise.all(subscriptions.map(this.client.unsubscribe));
```
or would it mess up with the types?

aleksandernsilva: @KevLehman It wouldn't, I even considered it, but I'm not sure if `unsubscribe` uses `this` in its internal implementation. Didn't think the risk was worth 😬

KevLehman (approved):

ggazzo (approved):

MartinSchoeler (approved):

PR #29724, opened 2023-07-05
chore: Improve system messages for omni-visitor abandonment feature

murtaza98:

~Visitor abandonment feature used to rely on `room.v.lastMessageTs` property to calculate the predicted time when the chat should be considered abandoned. Now when apps were updating the room properties like customFields, then the app converters were overriding lastMessageTs property ( [reference](https://github.com/RocketChat/Rocket.Chat/blob/74aa6770881eb620a2275b84c55465d7552e4597/apps/meteor/app/apps/server/converters/rooms.js#L41) ), thus making this abandonment feature not work for rooms created from apps. 
So to solve this, I've moved the `lastMessageTs` prop out of `room.v`, and instead I'm storing it as `room.contactLastMessageTs`. This will prevent the app's engine to override this prop, thus fixing the main issue due to which app's room were not being abandoned automatically.~

Update: This PR has been updated to now only include changes to system messages related to visitor abandonment feature. The rest of the changes wrt visitor abandonment feature not working has been moved to a separate PR here -> https://github.com/RocketChat/Rocket.Chat/pull/29756


KevLehman (changes requested):
What about compatibility with old rooms that have `v.lastMessage`? Since we're changing the property where we take info, we do need to make sure we keep compatibility with them. 

Or we can create a single shot function that moves `v.lastMessage` to `contactLastMessage`

KevLehman (dismissed):

KevLehman (dismissed):

apps/meteor/server/models/raw/LivechatRooms.ts
```diff
@@ -2383,13 +2383,13 @@ export class LivechatRoomsRaw extends BaseRaw<IOmnichannelRoom> implements ILive
...
 
-	setVisitorLastMessageTimestampByRoomId(roomId: string, lastMessageTs: Date) {
+	setVisitorLastMessageTimestampByRoomId(roomId: string, contactLastMessageAt: Date) {
 		const query = {
 			_id: roomId,
 		};
 		const update = {
 			$set: {
-				'v.lastMessageTs': lastMessageTs,
+				contactLastMessageAt,
```
ggazzo, line 2392: what is the point of adding this field? is this really necessary to fix the problem? or its like a improvement/refactor?
I could not find the exact line that actually fixes the problem

ggazzo (changes requested):

sampaiodiego (dismissed):
if you were able to pin point the problem root (https://github.com/RocketChat/Rocket.Chat/blob/74aa6770881eb620a2275b84c55465d7552e4597/apps/meteor/app/apps/server/converters/rooms.js#L41), why not fixing the code only there? I mean, all these changes are just an alternative solution, but it doesn't fix the root cause of the problem, right?

ggazzo (dismissed):
Great explanation @murtaza98, but I would like to enforce: on patches keep your changes small as possible, if you are fixing something focus on the fix, if you are refactoring something try to refactor without change any behavior, if you are creating something new, avoid fixes and code changes that dont are directly related with the new feature


about your change in specific I think its only about change the bridge as you mentioned. that would be a great fix (2..3 lines of code) 

if you are planning to replace some field or optimize any message let this to another pull request

murtaza98:
Thanks for the feedback everyone :raised_hands: As per your suggestions, I've created a separate PR with changes only to app's bridges which should fix the original issue -> https://github.com/RocketChat/Rocket.Chat/pull/29756
Wrt this PR, I'd still like to keep this it has some improvement to the system messages, so I've renamed it accordingly.

apps/meteor/ee/app/livechat-enterprise/server/lib/VisitorInactivityMonitor.ts
```diff
@@ -170,4 +155,41 @@ export class VisitorInactivityMonitor {
 
 		this._initializeMessageCache();
 	}
+
+	private async getDefaultAbandonedCustomMessage(abandonmentAction: 'close' | 'on-hold', visitorId: string) {
+		this.logger.debug(`Getting default abandoned custom message for ${abandonmentAction}`);
```
KevLehman, line 160:
```suggestion
```

KevLehman (changes requested):

KevLehman (dismissed):

apps/meteor/ee/app/livechat-enterprise/server/lib/VisitorInactivityMonitor.ts
```diff
@@ -170,4 +155,39 @@ export class VisitorInactivityMonitor {
 
 		this._initializeMessageCache();
 	}
+
+	private async getDefaultAbandonedCustomMessage(abandonmentAction: 'close' | 'on-hold', visitorId: string) {
+		const visitor = await LivechatVisitors.findOneById(visitorId, {
```
KevLehman, line 160:
```suggestion
		const visitor = await LivechatVisitors.findOneById<Pick<ILivechatVisitor, 'name' | 'username'>>(visitorId, {
```

KevLehman (changes requested):

KevLehman (approved):

PR #29736, opened 2023-07-05
refactor: Create `config` package

KevLehman:

- Creates `config` package
- Package is in charge of controlling env variables
- Package automatically validates env variables and properly type them


KevLehman:
image

ggazzo:
maybe I missed some conversation and the point here, but if the package does anything besides renaming the vars, I would stay with the `process.env`, at least we know from where this value comes

also this can prevent some bundlers plugins (like terser) to remove dead code in some cases

PR #29756, opened 2023-07-07
fix: Prevent app's bridges to override lastMsg prop

murtaza98:

### DIFF
--- a/apps/meteor/app/apps/server/converters/rooms.js
+++ b/apps/meteor/app/apps/server/converters/rooms.js
@@ -38,11 +38,15 @@ export class AppRoomsConverter {
 		let v;
 		if (room.visitor) {
 			const visitor = await LivechatVisitors.findOneById(room.visitor.id);
+			// For the patch, we're temporarily storing the last message timestamp in the unmapped properties
+			// We'll refactor this later in next versions
+			const lastMessageTs = room?._unmappedProperties_?.lastMessageTs;
 			v = {
 				_id: visitor._id,
 				username: visitor.username,
 				token: visitor.token,
 				status: visitor.status || 'online',
+				...(lastMessageTs && { lastMessageTs }),
 			};
 		}
 
@@ -109,6 +113,12 @@ export class AppRoomsConverter {
 	async convertRoom(room) {
 		if (!room) {
 			return undefined;
+		}
+
+		// Add room.v.lastMessageTs to room.lastMessageTs if it exists so that the converter doesn't override it
+		// This is a temporary fix until we can refactor this properly
+		if (room.v?.lastMessageTs) {
+			room.lastMessageTs = room.v.lastMessageTs;
 		}
 
 		const map = {


KevLehman (approved):

KevLehman (approved):

apps/meteor/app/apps/server/converters/rooms.js
```diff
@@ -38,11 +38,14 @@ export class AppRoomsConverter {
...
+
+			const lastMessageTs = room?.visitor?.lastMessageTs;
 			v = {
 				_id: visitor._id,
 				username: visitor.username,
 				token: visitor.token,
 				status: visitor.status || 'online',
+				...(lastMessageTs && { lastMessageTs }),
 			};
 		}
```
ggazzo, line 50: why not?
```suggestion
			v = {
				...room?.visitor,
				_id: visitor._id,
				username: visitor.username,
				token: visitor.token,
				status: visitor.status || 'online',
			};
		}
```

d-gubert: I guess this could be a problem if there are properties that have been renamed when sent over to the Apps-Engine. For instance, this would make both `id` and `_id` properties available in the converted `v` variable

d-gubert: What I think would work is spreading the `visitor` instead of `rooms?.visitor`

apps/meteor/app/apps/server/converters/rooms.js
```diff
@@ -172,9 +175,14 @@ export class AppRoomsConverter {
...
 				}
 
+				const { lastMessageTs } = v;
+
 				delete room.v;
 
-				return this.orch.getConverters().get('visitors').convertById(v._id);
+				return {
+					...(await this.orch.getConverters().get('visitors').convertById(v._id)),
+					...(lastMessageTs && { lastMessageTs }),
```
d-gubert, line 184: I would really recommend a comment above this line clarifying why we're adding this field here. On a first glance, it looks like the filed should just be added to the `VisitorsConverter.convertById` method. Leaving it like this will generate confusion in the future

d-gubert:
> The app's bridges were overriding this last msg prop

Just clarifying that it's not that the bridge was overriding the prop, it simply was not aware of it as it is a newer prop in comparison to when the bridge/converter was written. Ideally, for the future, we should consider how apps will interact with new data being added to models

apps/meteor/app/apps/server/converters/rooms.js
ggazzo: ok, maybe my question was not clear enough, I suggested "spread" the visitor not because I was not aware that both types are different, but was more like "are we not missing any other field? or in the feature if we add another field the bridge will remove it"?  and actually looks like we are, the phone field, but thankfully I believe this field doesnt even exist so we are fine

murtaza98: I get your point @ggazzo but I can't think of a way to apply your suggestions here. You see right now, app's and core are fundamentally storing 2 different objects which at first glance seems identical but they aren't. 
- The `room.v` prop on core is supposed to store information related to visitor + **ITS CURRENT CHANNEL**. So suppose you have a contact whom we're contacting using SMS via 2 phone no's, let's call X and Y. Then if the contact sends a message using X phone number, then `room.v.phoneNo` would be `X` and correspondingly we'll store the timestamp of the last message from this visitor from X phone no on `room.v.lastMessageTs`. 
- Now on the app's side, the bridge is mapping `room.v` to `room.visitor` and this visitor object only contains visitor info and **DOESN'T CONTAIN ANY INFO ABOUT THE CHANNEL**. 

This PR aims to add the additional channel info to "room.visitor" level on the apps. Right now we have 2 such channel props, one is lastMessageTs and other is phone (thanks for pointing this out btw, I'll update this PR with it shortly ;) ).
Now coming back to your suggestion, as pointed out by Kevin, room.visitor on app's has a bunch of more visitor related props which we're not storing on room.v level on core. Due to this we're manually mapping these values here. I'll further add comments to show this however if you can think of a better way to manage these typings, please feel free to share.

apps/meteor/app/apps/server/converters/rooms.js
murtaza98: Good idea! Added a comment here :heavy_check_mark:

sampaiodiego (approved):

sampaiodiego (approved):

apps/meteor/app/livechat/server/hooks/saveLastVisitorMessageTs.ts
```diff
@@ -2,15 +2,24 @@ import { isOmnichannelRoom } from '@rocket.chat/core-typings';
...
 	async function (message, room) {
 		if (!(isOmnichannelRoom(room) && room.v.token)) {
 			return message;
 		}
+		if (message.t) {
+			return message;
+		}
 		if (message.token) {
 			await LivechatRooms.setVisitorLastMessageTimestampByRoomId(room._id, message.ts);
+			callbackLogger.debug({
```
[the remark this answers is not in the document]
KevLehman: I removed this one as it's indeed not needed.

Would be cool to have a kind of workbench to see how debug logs affect perf in the real world, cause we are adding more and more

PR #29763, opened 2023-07-07
regression: option to start video conferences disappearing for regular users

pierre-lehnen-rc:


sampaiodiego (changes requested):
pls change the regression to a single command

sampaiodiego (approved):

PR #29781, opened 2023-07-11
fix: Pruning messages in a room results in an incorrect message counter

matheusbsilva137:

 - Pruning message feature (fired by the `rooms.cleanHistory` endpoint) now disconsiders deleted messages and its system messages (created when the "Show Deleted Status" or "Keep Per Message Editing History" settings are enabled) when decreasing room messages' counter.

## Steps to test or reproduce

1. Go to **Workspace > Settings > Message** and enable both "Show Deleted Status" and "Keep Per Message Editing History" settings.
2. Create a room, send some messages in it and erase one of them;
3. Prune this room's messages;
4. Access **Workspace > Rooms** and check the messages counter from the created room.
**Old behavior:** the counter is `-2` (incorrect since we're counting the hidden deleted message and its system message, which should not be subtracted from the counter when pruning messages -- since they are never added to the counter).
**Current behavior:** the counter is `0`.

## Further comments

A migration to adjust the message counter of rooms (eg the ones with negative counters) may not be feasible. So we'll stick with solving the root cause for now.

### DIFF
--- a/apps/meteor/server/models/raw/Messages.ts
+++ b/apps/meteor/server/models/raw/Messages.ts
@@ -25,6 +25,7 @@ import type {
 	Document,
 	UpdateFilter,
 } from 'mongodb';
+import { isDeletedMessage } from '@rocket.chat/core-typings';
 import { escapeRegExp } from '@rocket.chat/string-helpers';
 import { Rooms } from '@rocket.chat/models';
 
@@ -1362,22 +1363,28 @@ export class MessagesRaw extends BaseRaw<IMessage> implements IMessagesModel {
 			return count;
 		}
 
-		const messagesToDelete = (
-			await this.find(query, {
-				projection: {
-					_id: 1,
-				},
-				limit,
-			}).toArray()
-		).map(({ _id }) => _id);
+		const messagesToDelete = await this.find(query, {
+			projection: {
+				_id: 1,
+				_hidden: 1,
+				t: 1,
+				editedAt: 1,
+				editedBy: 1,
+			},
+			limit,
+		}).toArray();
 
-		const count = (
-			await this.deleteMany({
-				_id: {
-					$in: messagesToDelete,
-				},
-			})
-		).deletedCount;
+		const notCountedMessages = messagesToDelete.filter((message) => message._hidden === true || isDeletedMessage(message)).length;
+		const messagesIdsToDelete = messagesToDelete.map(({ _id }) => _id);
+
+		const count =
+			(
+				await this.deleteMany({
+					_id: {
+						$in: messagesIdsToDelete,
+					},
+				})
+			).deletedCount - notCountedMessages;
 
 		if (count) {
 			// decrease message count


tapiarafael (approved):

apps/meteor/server/models/raw/Messages.ts
```diff
@@ -1362,22 +1363,28 @@ export class MessagesRaw extends BaseRaw<IMessage> implements IMessagesModel {
...
+		}).toArray();
 
-		const count = (
-			await this.deleteMany({
-				_id: {
-					$in: messagesToDelete,
-				},
-			})
-		).deletedCount;
+		const notCountedMessages = messagesToDelete.filter((message) => message._hidden === true || isDeletedMessage(message)).length;
```
MarcosSpessatto, line 1377: Wouldn't it be better to filter this at the DB level? Or even use only the `deleteMany` method, without querying the message first (not sure if it's possible 🤔 )

apps/meteor/server/models/raw/Messages.ts
```diff
@@ -1362,22 +1363,28 @@ export class MessagesRaw extends BaseRaw<IMessage> implements IMessagesModel {
...
-		).map(({ _id }) => _id);
+		const messagesToDelete = await this.find(query, {
+			projection: {
+				_id: 1,
+				_hidden: 1,
+				t: 1,
+				editedAt: 1,
+				editedBy: 1,
+			},
+			limit,
```
MarcosSpessatto, line 1374: I believe we are forgetting to apply this change when there is no limit defined, right?

MarcosSpessatto (changes requested):

MarcosSpessatto:
@matheusbsilva137 can you please also add a test case for this if possible?

apps/meteor/server/models/raw/Messages.ts
matheusbsilva137: Done! I think using only `deleteMany` isn't possible due to the `limit` param :thinking:

apps/meteor/server/models/raw/Messages.ts
matheusbsilva137: Oopsss, thanks. great catch :grimacing: 
Fixed!

MarcosSpessatto (changes requested):

apps/meteor/server/models/raw/Messages.ts
```diff
@@ -1351,8 +1351,23 @@ export class MessagesRaw extends BaseRaw<IMessage> implements IMessagesModel {
...
+						_id: 1,
+					},
+					limit,
+				},
+			).toArray()
+		).length;
+
 		if (!limit) {
-			const count = (await this.deleteMany(query)).deletedCount;
+			const count = (await this.deleteMany(query)).deletedCount - notCountedMessages;
```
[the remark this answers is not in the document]
matheusbsilva137: Since `limit` is `undefined` in this case, aren't the results not `limit`ed? :thinking: 

We can't just set it to `0` since there may be a `ts` filter in the query too. So we're not necessarily deleting _everything_ (all messages) within this method, it could also be all messages in a specific time period (or all messages except from threads, discussions etc.).

MarcosSpessatto (approved):

PR #29797, opened 2023-07-12
chore: move some file functions to upload service

MarcosSpessatto:

FED-114


apps/meteor/server/services/upload/service.ts
```diff
@@ -22,12 +22,31 @@ export class UploadService extends ServiceClassInternal implements IUploadServic
...
+		roomId: string,
+		user: IUser,
+	): Promise<{
+		files: {
+			_id: string;
+			name: string | undefined;
+			type: string | undefined;
+		}[];
+		attachments: MessageAttachment[];
+	}> {
```
KevLehman, line 49: Not sure what's better, typing this again, or type it as `ReturnType<parseFileIntoMessageAttachments>` now that's properly typed

apps/meteor/server/services/federation/infrastructure/rocket-chat/adapters/File.ts
```diff
@@ -23,7 +23,7 @@ export class RocketChatFileAdapter {
 			try {
 				const { files, attachments } = await parseFileIntoMessageAttachments(uploadedFile, internalRoomId, internalUser);
 
-				resolve({ files, attachments });
+				resolve({ files: files as FileProp[], attachments });
```
KevLehman, line 26: If you're casting files as FileProp, why not making `parseFileINtoMessageAttachment` return a `FileProp[]`?

KevLehman (changes requested):

apps/meteor/server/services/federation/infrastructure/rocket-chat/adapters/File.ts
MarcosSpessatto: makes sense 🤙

packages/core-typings/src/IMessage/MessageAttachment/Files/FileProp.ts
```diff
@@ -1,7 +1,7 @@
 export type FileProp = {
 	_id: string;
-	name: string;
-	type: string;
-	format: string;
-	size: number;
+	name?: string;
+	type?: string;
+	format?: string;
+	size?: number;
```
KevLehman, line 6: Hmmm, a type with only ID as required, maybe it's not the proper type for your use case?

KevLehman (approved):

sampaiodiego (dismissed):

KevLehman (approved):

elfajriharsuda (approved):

sampaiodiego (approved):

PR #29800, opened 2023-07-12
chore: move some functions to a (new) user service

MarcosSpessatto:

FED-114


ggazzo (approved):

ggazzo (changes requested):
instead of creating a new user service one this should not be related to accounts?

pierre-lehnen-rc (changes requested):

pierre-lehnen-rc (dismissed):

PR #29801, opened 2023-07-12
chore: Change NetworkBroker Serializer to CBOR-X

d-gubert:

The motivation for this is that, for Apps-Engine running on Microservices, we'll need access to the `Upload` service to grab a file's content. It just so happens that our current serializer (EJSON) transforms Buffers to `UInt8Array`s before transporting it via network (microservice deployment), but doesn´t automatically transform it back into `Buffer`. 

Additionally, if the workspace is NOT running in microservice mode, the service will be able to deal with Buffers natively (as expected). This behavior screws our service typing, because instead of having the method typed like this `getFileBuffer(...args): Promise<Buffer>` we´d need to change it into `getFileBuffer(...args): Promise<Buffer | UInt8Array>` - which sucks as there is nothing in the method's implementation that refers to `UInt8Array`, making the code less intuitive to read.

An alternative would be to force the return of the method to _always be `UInt8Array`_, but.... that feels awkward.

I've picked `cbor-x` from [moleculer's own list of supported serializers](https://moleculer.services/docs/0.14/networking.html#Serialization) mostly because of it's speed and flexibility characteristics. See https://github.com/moleculerjs/moleculer/pull/905 and https://github.com/kriszyp/cbor-x


apps/meteor/ee/server/startup/broker.ts
```diff
@@ -10,7 +10,7 @@ const {
 	TRANSPORTER = '',
 	CACHE = 'Memory',
 	// SERIALIZER = 'MsgPack',
-	SERIALIZER = 'EJSON',
+	SERIALIZER = 'CBOR',
```
KevLehman, line 13: Do this serializer works when the data is going through TCP transporter? Probably yes, but would like to confirm

d-gubert: Yes, it does. Moleculer even separates these concepts as they can be agnostic of each other's details. But thanks for asking, I checked it anyway and it does work :wink:

PR #29804, opened 2023-07-12
fix: Spotlight search does not find rooms with special characters

### LINKED ISSUE #22833: Bug in search with cyrillic alphabet
### Description:

Hi guys!

Well, there was an issue on mobile about Russian alphabet search, which we are already solving. However, we noticed that the spotlight does not return any rooms when searching with Cyrillic alphabet, only users. So I created a public channel on `open.rocket.chat` with the name: `ЩДФЩДФ` to test the request

#### Endpoint
`POST` https://open.rocket.chat/api/v1/method.call/spotlight

#### Data Sent
```json
{
  "message": "{\"method\":\"spotlight\",\"params\":[\"ЩДФЩДФ\",[],{\"users\":true,\"rooms\":true}]}"
}
```                    

#### Data Received
```json
{
  "message": "{\"msg\":\"result\",\"result\":{\"users\":[],\"rooms\":[]}}",
  "success": true
}
```

### Steps to reproduce:

1.  Click on `Search`
2.  Write `ЩДФЩДФ`, or with any case type

### Expected behavior:

- Expect that the room is returned from API
### Actual behavior:

- Doesn't return the room from API

### Server Setup Information:

`open.rocket.chat`

matheusbsilva137:

 - Search also for `fname` (and not only `name`) on spotlight rooms search.

## Issue(s)

Fixes #22833

## Further comments

SUP-253


apps/meteor/server/models/raw/Rooms.ts
```diff
@@ -1327,9 +1330,12 @@ export class RoomsRaw extends BaseRaw<IRoom> implements IRoomsModel {
 				},
 				includeFederatedRooms
 					? {
-							$or: [{ $and: [{ $or: [{ federated: { $exists: false } }, { federated: false }], name }] }, { federated: true, fname: name }],
+							$or: [
```
KevLehman, line 1333: If you're adding nameCondition to both conditions, why not moving it to the main query?

matheusbsilva137: Not sure if I can do that due to [line 1335](https://github.com/RocketChat/Rocket.Chat/pull/29804/files#diff-4812f4201fc467c5333be9eba5448d71fa67a6ad0e0be89e9645dddcae5066ccR1335)
When `federated` is true, we should only search by `fname` :thinking: I think that's intended behavior

reinaldonetof:
Maybe fixes this: https://github.com/RocketChat/Rocket.Chat/issues/22833 ?

tapiarafael (approved):

KevLehman (dismissed):

tassoevan (dismissed):

KevLehman (dismissed):

tassoevan (dismissed):

tassoevan (approved):

heitortanoue (approved):

PR #29813, opened 2023-07-13
regression: app's not able to upload binary files

murtaza98:


packages/server-fetch/src/parsers.ts
```diff
@@ -32,18 +32,18 @@ const jsonParser = (options: ExtendedFetchOptions) => {
...
 };
 
 const getParser = (contentTypeHeader?: string): ((options: ExtendedFetchOptions) => FetchOptions) => {
 	switch (contentTypeHeader) {
 		case 'application/json':
 			return jsonParser;
-		case 'application/x-www-form-urlencoded':
-			return urlencodedParser;
 		default:
-			return jsonParser;
```
KevLehman, line 46: I'm trying to think on how changing the default parser could affect other calls. 

For now, maybe adding a new case for the type of data you're sending works? 👀

KevLehman (approved):

PR #29818, opened 2023-07-13
feat: Report User

gabriellsh:

 - Add the `moderation.reportUser` endpoint, which enables users to report others.

## Issue(s)

[TC-779](https://rocketchat.atlassian.net/browse/TC-779)

## Steps to test or reproduce

The endpoint can be called as follows (`description` is required):
```
curl -H "X-Auth-Token: PP8mRUeNKIH2wdh6BhQVIBsXxuvq5Xhqg4kYzrakfN9" \
   -H "X-User-Id: JmR3T4TFfppJCfmq9" \
   -H "Content-type:application/json" \
   http://localhost:3000/api/v1/moderation.reportUser \
   -d '{ "userId": "edjvbnb7pGC3MZNSP", "description": "test" }'
```
Which should return a successful response:
```
{
  "acknowledged": true,
  "insertedId": "64c7cbce7d87bb28d7e5ab77",
  "success": true
}
```

## Further comments

[TC-779]: https://rocketchat.atlassian.net/browse/TC-779?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/model-typings/src/models/IModerationReportsModel.ts
```diff
@@ -17,7 +17,13 @@ export interface IModerationReportsModel extends IBaseModel<IModerationReport> {
...
 	): ReturnType<IBaseModel<IModerationReport>['insertOne']>;
 
-	findReportsGroupedByUser(
+	createWithDescriptionAndUser(
+		reportedUser: Exclude<IModerationReport['reportedUser'], undefined>,
+		description: string,
+		reportedBy: IModerationReport['reportedBy'],
+	): ReturnType<IBaseModel<IModerationReport>['insertOne']>;
+
+	findMessageReportsGroupedByUser(
```
matheusbsilva137, line 26: We should also ensure that this will only return message reports (with `message: { $exists: true }` or something similar)

packages/model-typings/src/models/IModerationReportsModel.ts
```diff
@@ -27,7 +33,7 @@ export interface IModerationReportsModel extends IBaseModel<IModerationReport> {
 	countReportsInRange(latest: Date, oldest: Date, selector: string): Promise<number>;
```
matheusbsilva137, line 33: This should also be updated to `countMessageReportsInRange` (+ the code should be updated too to ensure we only count message reports)

matheusbsilva137 (changes requested):

packages/model-typings/src/models/IModerationReportsModel.ts
matheusbsilva137: Fixed. I added this check to the `getSearchQueryForSelector` method.

packages/rest-typings/src/v1/moderation/ModerationReportUserPOST.ts
```diff
@@ -0,0 +1,25 @@
...
+	properties: {
+		userId: {
+			type: 'string',
+		},
+		description: {
+			type: 'string',
+			nullable: true,
+		},
+	},
+	required: ['userId', 'description'],
```
heitortanoue, line 21: If the `description` is optional (nullable), wouldn't be better to remove it as required?

matheusbsilva137: Sorry, `description` should actually be required :grimacing:

apps/meteor/server/models/raw/ModerationReports.ts
```diff
@@ -213,7 +226,7 @@ export class ModerationReportsRaw extends BaseRaw<IModerationReport> implements
 		return this.updateMany(query, update);
 	}
 
-	async hideReportsByUserId(userId: string, moderatorId: string, reason: string, action: string): Promise<UpdateResult | Document> {
+	async hideMessageReportsByUserId(userId: string, moderatorId: string, reason: string, action: string): Promise<UpdateResult | Document> {
```
Dnouv, line 229: I would recommend we keep as it was before, and maybe extend it.

Why: From the main console, if Admin selects the action "Dismiss Reports" it does not matter if user reports or message reports we dismiss all of them at once. And we could further (if required) introduce an optional param `userReport?: true` which if passed, we process with hiding the user reports, else we only hide the message reports.

apps/meteor/app/api/server/v1/moderation.ts
```diff
@@ -243,3 +249,30 @@ API.v1.addRoute(
...
+	},
+	{
+		async post() {
+			const { userId, description } = this.bodyParams;
+
+			const {
+				user: { _id, name, username, createdAt },
+			} = this;
+
+			const reportedUser = await Users.findOneById(userId, { projection: { _id: 1, name: 1, username: 1 } });
```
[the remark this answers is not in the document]
gabriellsh: This could be done, but there's a couple of regards.
1. Will we update these fields for every report when the user is updated? If not, it should be correctly described as a field that was only recorded at the time of the report. If we are going to update then we should implement the reactivity.

2. Are we going to use `createdAt` for something? As far as I know there's no use for it currently. This one doesn't need to be updated though, and should be easier to include.

apps/meteor/server/models/raw/ModerationReports.ts
gabriellsh: Actually I just noticed this change wasn't done by me, but the explanation still stands.

packages/core-typings/src/IModerationReport.ts
```diff
@@ -11,15 +16,25 @@ export interface IModerationInfo {
...
 	_hidden?: boolean;
 }
 
+export type MessageReport = Omit<IModerationReport, 'reportedUser'> & {
+	room: Exclude<IModerationReport['room'], undefined>;
+	message: Exclude<IModerationReport['message'], undefined>;
+};
+
+export type UserReport = Omit<IModerationReport, 'message' | 'room'> & {
+	reportedUser: Exclude<IModerationReport['reportedUser'], undefined>;
```
matheusbsilva137, line 35: I was talking to @Dnouv and we're wondering if user reports should also include a room :thinking: 
Not sure if it's the right path to have it as a required param or if we should even have it at all. Maybe using as an optional param could work
I think rooms are unrelated enough from user reports to make us not store them, but I'd like to have architecture's opinion on this
cc @sampaiodiego

apps/meteor/app/api/server/v1/moderation.ts
```diff
@@ -35,7 +37,11 @@ API.v1.addRoute(
 
 			const escapedSelector = escapeRegExp(selector);
 
-			const reports = await ModerationReports.findReportsGroupedByUser(latest, oldest, escapedSelector, { offset, count, sort }).toArray();
+			const reports = await ModerationReports.findMessageReportsGroupedByUser(latest, oldest, escapedSelector, {
+				offset,
+				count,
+				sort,
```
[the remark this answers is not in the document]
Dnouv: Thank you for the heads up, @gabriellsh !

I understand, so we can't modify the data format returned from an endpoint unless it is a notified breaking change. That lefts three options for us:

1. As you said, modify the designs & have tabs on the main console.
2. Make two API calls, one for getting the message reports, and a second for getting the user reports, combine the results on FE, and show them on the moderation console.
3. Use a new endpoint that would return the combined data without the need for making two calls as required in the second.

I am not sure, about the product decision on changing the designs, but to me, the second solution sounds sort of hacky. What do you think would be the best approach for us?

Thanks!

apps/meteor/app/api/server/v1/moderation.ts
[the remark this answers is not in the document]
Dnouv: Thanks for bringing up the above points; I understand that's the thing to consider.

After thinking about it for a while and considering the previous comment, _we might have to consider modifying the design_. 

**Approach 1:**
I think this is more suitable for user reports:
```ts
{
_id: string,
reportedUser: IUser<roles[], _id, name, uname, createdAt, email>,
reporter: IUser<_id, name, uname>[],
description: Array<string>,
moderationInfo: ...
}
``` 

Why the above recommendation:

The user reports are amazingly different from user-generated content reporting, especially we remove the `messages` variable; that was the main reason I did not proceed with the above schema for the message reports. So every time someone reports a user, we upsert the data. This covers the concern of constantly updating the data.

The above schema could have potential drawbacks, like making the queries based on `reporter` more complex. But at the same time, it simplifies the query that we use on the Main Console.
The main concern would be the increasing document size, as more reports against the user are submitted. However, I assume that won't be the case, considering we have a hard limit of 16MB per document. 

And please feel free to correct me if I miss assumed something or missed something. Thank you!


**Approach 2:**

It is good to have historical data for Moderation purposes. For example, if a user never had a `moderator` or any critical role, it helps the moderator understand the context better. And if the moderator sees that this user has changed over emails multiple times, that raises a red flag.

The `createdAt` is as well helpful information for Moderator; if a user created their account two days back and got multiple reports, it increases the probability that this account was created for malicious purposes.

_However, none of the above guarantees any fact; after all, the final action to take should be handled by the Moderators reasoning_

Thank you!

packages/core-typings/src/IModerationReport.ts
debdutdeb: if it is not being used today or really no plans from product either, i don't think it's worth adding an unrelated patch to this pr.

apps/meteor/app/api/server/v1/moderation.ts
```diff
@@ -5,6 +5,8 @@ import {
 	isArchiveReportProps,
 	isReportInfoParams,
 	isReportMessageHistoryParams,
+	isModerationReportUserPost,
+	// isUserReportsGet,
```
MarcosSpessatto, line 9:
```suggestion
```

apps/meteor/server/models/raw/ModerationReports.ts
```diff
@@ -21,21 +29,37 @@ export class ModerationReportsRaw extends BaseRaw<IModerationReport> implements
...
-		description: string,
+		description: IModerationReport['description'],
 		room: IModerationReport['room'],
 		reportedBy: IModerationReport['reportedBy'],
 	): ReturnType<BaseRaw<IModerationReport>['insertOne']> {
 		const record: Pick<IModerationReport, 'message' | 'description' | 'ts' | 'reportedBy' | 'room'> = {
 			message,
 			description,
 			reportedBy,
+			// reportedUser: message.u,
```
MarcosSpessatto, line 40:
```suggestion
```

apps/meteor/tests/end-to-end/api/27-moderation.ts
```diff
@@ -598,4 +599,42 @@ describe('[Moderation]', function () {
...
+				.post(api('moderation.reportUser'))
+				.set(credentials)
+				.send({
+					userId: userToBeReported?._id,
+				})
+				.expect('Content-Type', 'application/json')
+				.expect(400)
+				.expect((res: Response) => {
+					expect(res.body).to.have.property('success', false);
+					expect(res.body).to.have.property('error');
```
MarcosSpessatto, line 636: Is it possible to assert the error message?

apps/meteor/tests/end-to-end/api/27-moderation.ts
```diff
@@ -598,4 +599,42 @@ describe('[Moderation]', function () {
 				});
 		});
 	});
+
+	describe('[/moderation.reportUser]', () => {
+		let userToBeReported: IUser;
+
+		before(async () => {
+			userToBeReported = await createUser();
```
MarcosSpessatto, line 607: Is it possible to remove the user after the test cases are executed?

MarcosSpessatto (changes requested):

packages/core-typings/src/IModerationReport.ts
```diff
@@ -11,15 +16,25 @@ export interface IModerationInfo {
 }
 
 export interface IModerationReport extends IRocketChatRecord {
-	message: IMessage;
+	message?: IMessage;
+	room?: Pick<IRoom, '_id' | 'name' | 'fname' | 't' | 'federated' | 'prid'>;
+	reportedUser?: Pick<IUser, '_id' | 'username' | 'name'>;
```
matheusbsilva137, line 21:
```suggestion
	reportedUser?: Pick<IUser, '_id' | 'username' | 'name' | 'emails' | 'createdAt'>;
```

matheusbsilva137 (changes requested):

apps/meteor/app/api/server/v1/moderation.ts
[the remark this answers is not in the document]
matheusbsilva137: Hey @Dnouv !
As I mentioned, we discussed these two proposals with the Architecture team and came to the conclusion that reports should behave like a snapshot of the moment of the report. Hence, we should store each of them in a separate record and do not upsert the reported user data (then, if someone is reported due to their username, we wouldn't lose track of it if the reported user updates it).
Still, inserting more data into the report sounds like a good idea. We now also store `emails` and the `createdAt` field.

apps/meteor/app/api/server/v1/moderation.ts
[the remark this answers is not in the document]
matheusbsilva137: Since the moderation console design for user reports is still a wip, for now I guess we can keep the behavior as it is and make sure it's clear that current methods only deal with message reports -- and not introduce a breaking change. But this definitely will be changed depending on some design choices (I'll try to make some suggestions on it later).

matheusbsilva137 (approved):

heitortanoue (dismissed):

packages/core-typings/src/IModerationReport.ts
sampaiodiego: agreed

debdutdeb (dismissed):
please change the response structure 🙈 just success: true or with the filtered document, just not the raw driver/mongo response

apps/meteor/app/api/server/v1/moderation.ts
```diff
@@ -243,3 +248,30 @@ API.v1.addRoute(
...
+
+			const reportedUser = await Users.findOneById(userId, { projection: { _id: 1, name: 1, username: 1, emails: 1, createdAt: 1 } });
+
+			if (!reportedUser) {
+				return API.v1.failure('Invalid user id provided.');
+			}
+
+			const response = await ModerationReports.createWithDescriptionAndUser(reportedUser, description, { _id, name, username, createdAt });
+
+			return API.v1.success(response);
```
debdutdeb, line 274:
```suggestion
			await ModerationReports.createWithDescriptionAndUser(reportedUser, description, { _id, name, username, createdAt });

			return API.v1.success();
```
:)

ggazzo (approved):

tapiarafael (approved):

PR #29827, opened 2023-07-14
chore: Improve bio and nickname exceeded max length errors

matheusbsilva137:

 - Throw more meaningful errors when setting bio or nickname that exceeds max length (introduce `error-bio-size-exceeded` and `error-nickname-size-exceeded)`.

## Further comments

TC-723


KevLehman (commented):
Can i haz some tests? 😬

apps/meteor/app/lib/server/functions/saveUser.js
```diff
@@ -237,11 +237,16 @@ export async function validateUserEditing(userId, userData) {
 
 const handleBio = (updateUser, bio) => {
 	if (bio && bio.trim()) {
-		if (typeof bio !== 'string' || bio.length > MAX_BIO_LENGTH) {
+		if (typeof bio !== 'string') {
```
KevLehman, line 240: It's kinda weird that we do a `trim` and then we check for the value to be an string tbh

matheusbsilva137: Agree. Removed this check from `bio` and `username` in this file

KevLehman: For safety i think you can keep them, but will be better placed before trying to call `.trim()` :)

tapiarafael (approved):

tapiarafael (dismissed):

tapiarafael (approved):

PR #29832, opened 2023-07-15
fix: not able do see reports about deleted users

ggazzo:
## Further comments

Predecessor PR: https://github.com/RocketChat/Rocket.Chat/pull/29701 
[WM-120](https://rocketchat.atlassian.net/browse/WM-120)

[WM-120]: https://rocketchat.atlassian.net/browse/WM-120?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Dnouv:
Made some changes for following:

1. User actions on the Moderation Console table e.g. "Deactivate user" was not disabled.
C. Added back the hook `useIsUserDeleted` which is being called inside the `ModerationConsoleAction`

2. The GUI crashes if reports are being checked (inside a context bar) of a user, who's never been reported.
C. Some additional optional chaining in the hook, which is getting the username, or name

3. Type errors on the `cloneAndSaveAsHistoryByRecord`
C. added a `// @ts-expect-error`

Thank you!

Dnouv:
I had to force push because I did a `develop` merge and reverted the same commit. But the commit history was still showing up, so removed all of it using a rebase.
Thank you!

apps/meteor/app/api/server/v1/moderation.ts
```diff
@@ -73,10 +73,7 @@ API.v1.addRoute(
 
 			const { count = 50, offset = 0 } = await getPaginationItems(this.queryParams);
 
-			const user = await Users.findOneById(userId, { projection: { _id: 1 } });
-			if (!user) {
-				return API.v1.failure('error-invalid-user');
-			}
+			const user = await Users.findOneById(userId, { projection: { _id: 1, username: 1, name: 1 } });
```
KevLehman, line 76:
```suggestion
			const user = await Users.findOneById<Pick<IUser, 'username' | 'name' | '_id'>>(userId, { projection: { _id: 1, username: 1, name: 1 } });
```

apps/meteor/server/lib/moderation/deleteReportedMessages.ts
```diff
@@ -22,7 +22,14 @@ export async function deleteReportedMessages(messages: IMessage[], user: IUser):
 	}
 	if (keepHistory) {
 		if (showDeletedStatus) {
-			await Promise.all(messageIds.map((id) => Messages.cloneAndSaveAsHistoryById(id, user as any)));
+			const cursor = Messages.find({ _id: { $in: messageIds } });
+
+			for await (const doc of cursor) {
+				await Messages.cloneAndSaveAsHistoryByRecord(
+					doc,
+					user as Required<Pick<IUser, '_id' | 'username' | 'name'>> & { username: NonNullable<IUser['username']> },
```
KevLehman, line 30:
```suggestion
					user as Required<Pick<IUser, '_id' | 'name'>> & { username: NonNullable<IUser['username']> },
```

apps/meteor/app/api/server/v1/moderation.ts
Dnouv: Thank you!

apps/meteor/server/lib/moderation/deleteReportedMessages.ts
Dnouv: Thanks a lot for the review!

ggazzo:
thanks for the additions  @Dnouv,  but I deleted the hook for a good reason (actually 3)

- during the load state, the callout is going to be shown, blinking/disappearing after the user data arrives which is a bad ux
- one fetch per row potentially will  trigger the rate limiter if you have many reports
- if the user doesnt exist or if you get a rate limiter error react query will retry at least 3x, in a huge list you are going to fetch a bunch of times
I added this info to the aggregation we are going to be forced to check if the user exists anyway, so I believe it doesn’t add extra load to the operation

Dnouv:
I agree; we can remove them. Thank you so much for the changes! I missed this case. Indeed I noticed the query resulted in an error for the deleted users, and it was being called thrice in a row. 

Can the delay between two consecutive retries be modified (just asking out of curiosity)?

KevLehman (approved):

pierre-lehnen-rc (approved):

ggazzo:
> I agree; we can remove them. Thank you so much for the changes! I missed this case. Indeed I noticed the query resulted in an error for the deleted users, and it was being called thrice in a row.
> 
> Can the delay between two consecutive retries be modified (just asking out of curiosity)?
yes, react query allow us to change the behaviour

KevLehman:
@Dnouv , maybe you can update the api tests for this case? 🙏🏽  (on a separate PR so we don't have this staying open for longer)

Dnouv:
> yes, react query allow us to change the behaviour

Interesting, thank you so much!

> @Dnouv , maybe you can update the api tests for this case? 🙏🏽 (on a separate PR so we don't have this staying open for longer)

Sure, I would be happy to do that. Just confirming, do we need this test PR today?

ggazzo:
@Dnouv 
> Sure, I would be happy to do that. Just confirming, do we need this new test PR today?

no need

Dnouv (approved):

PR #29837, opened 2023-07-17
chore: move getUrl to meteor service

MarcosSpessatto:


KevLehman (dismissed):

KevLehman:
My only concern is if we start to treat the meteor service as a "utils" service 🤔 at some point, we may need to put boundaries on what this service is/was for

ggazzo (approved):

PR #29838, opened 2023-07-17
fix: SAML full name update is not mirrored in DMs

matheusbsilva137:

 - Update DM Subscriptions on SAML `fullName` changes.

## Further comments

TC-740


apps/meteor/app/meteor-accounts-saml/server/lib/SAML.ts
```diff
@@ -220,8 +221,8 @@ export class SAML {
 			},
 		);
 
-		if (username && username !== user.username) {
-			await saveUserIdentity({ _id: user._id, username });
+		if (username && fullName && (username !== user.username || fullName !== user.name)) {
```
pierre-lehnen-rc, line 224: If there's no username or no fullName, the other will also not be saved.

matheusbsilva137: oops, thanks for noticing this :grimacing: please check #30177

PR #29862, opened 2023-07-18
regression: Align user status on discussions header

yash-rajpal:
## Further comments

TC-826


apps/meteor/client/views/room/Header/ParentRoom.tsx
```diff
@@ -18,7 +19,9 @@ const ParentRoom = ({ room }: ParentRoomProps): ReactElement => {
 	return (
 		<HeaderTag onClick={handleClick}>
 			<HeaderTagIcon icon={icon} />
-			{roomCoordinator.getRoomName(room.t, room)}
+			<Box display='inline-block' mis='x6'>
```
gabriellsh, line 22: I think this could break the styles a little if we have an icon different than the user status. See `HeaderTagIcon`, it already has margins when the provided icon is not a react element. I believe it's better to add these margins inside `HeaderTagIcon` component. Just make sure it's not being used with another purpose elsewhere.

gabriellsh (changes requested):

apps/meteor/client/views/room/Header/ParentRoom.tsx
yash-rajpal: Yess! You are right. The header tag icon is not being used at other places with icon element so added the styles to the component itself.
Thanks!

gabriellsh (approved):
lgtm

PR #29868, opened 2023-07-19
fix: Avoid blank spaces being sent as comment when comment is required

KevLehman:
OC-455


murtaza98 (approved):

murtaza98 (approved):

murtaza98 (approved):

MarcosSpessatto (dismissed):

aleksandernsilva (approved):

murtaza98 (approved):

MarcosSpessatto (changes requested):
@murtaza98 Sorry, I think I don't express myself correctly, I was talking about creating a test case for the problem we are solving here.

It's good to have the test case you created, let's keep that as well, but I believe we should create a test case sending the comment as an empty string with one or more blank(s) space(s) in it.

Something like:
```
it('should not close a room without comment', async () => {
			await restorePermissionToRoles('close-others-livechat-room');
			const visitor = await createVisitor();
			const { _id } = await createLivechatRoom(visitor.token);
			const response = await request.post(api('livechat/room.closeByUser')).set(credentials).send({ rid: _id, comment: '  ' }).expect(400);

			expect(response.body).to.have.property('success', false);
			expect(response.body).to.have.property('error', 'error-comment-is-required');
		});
```

murtaza98 (approved):

MarcosSpessatto (approved):

ggazzo (approved):

PR #29869, opened 2023-07-19
fix: Permissions Table > Roles sdk stream

hugocostadev:

The permission table was not receiving updates on Role changes

The roles stream was not running on the `Meteor.startup`, probably the stream was being invalidate before finishing, to avoid this behavior I added a `Tracker.afterFlush` to schedules a function to be called during the next flush, or later in the current flush if one is in progress.

## Steps to test or reproduce

1. Head over to the Workspace>Permissions
2. Delete/Create/Edit any custom role.
3. The permissions table header data won’t get updated unless a refresh is done.

## Further comments

WM-69


rique223 (approved):

apps/meteor/client/startup/roles.ts
```diff
@@ -32,8 +32,11 @@ Meteor.startup(() => {
...
 			return;
 		}
-		sdk.stream('roles', ['roles'], (role) => {
-			events[role.type]?.(role);
+
+		Tracker.afterFlush(() => {
+			sdk.stream('roles', ['roles'], (role) => {
+				events[role.type]?.(role);
+			});
 		});
```
ggazzo, line 40: could you please check if this solves the issue?
```suggestion
		if (Meteor.status().status !== 'connected') {
			return;
		}
		sdk.stream('roles', ['roles'], (role) => {
			events[role.type]?.(role);
		});
```

hugocostadev: I checked and don't work :/

tassoevan (approved):

PR #29870, opened 2023-07-19
feat: Add `membersByHighestRole` endpoints

matheusbsilva137:

 - Add `channels.membersByHighestRole` and `groups.membersByHighestRole` endpoints to enable users to retrieve members of a room ordered by their highest (most important) role.

## Steps to test or reproduce

The endpoint's usage is the same as the current `.members`:
```
curl -H "X-Auth-Token: 9HqLlyZOugoStsXCUfD_0YdwnNnunAJF8V47U3QHXSq" \
     -H "X-User-Id: aobEdbYhXfu5hkeqG" \
     http://localhost:3000/api/v1/groups.membersByHighestRole?roomId=ByehQjC44FwMeiLbX
```
And results are as follows:
```
{
  "members": [
    {
      "_id": "JmR3T4TFfppJCfmq9",
      "status": "offline",
      "_updatedAt": "2023-07-20T20:43:05.895Z",
      "name": "Matheus Barbosa Silva",
      "statusConnection": "offline",
      "username": "matheus.barbosa",
      "avatarETag": "fdJn5SqCwg4ZwbywD",
      "highestRole": {
        "role": "owner",
        "level": 0
      }
    },
    {
      "_id": "edjvbnb7pGC3MZNSP",
      "status": "offline",
      "_updatedAt": "2023-07-17T21:42:31.845Z",
      "username": "john",
      "name": "John",
      "statusConnection": "offline",
      "highestRole": {
        "role": "member",
        "level": 2
      }
    }
  ],
  "count": 2,
  "offset": 0,
  "total": 2,
  "success": true
}
```
 - The `members` array contains a list of users ordered from highest to lowest role.
 - A numeric representation of the highest role level is given in the `roleLevel` field (inside the `highestRole` object), which ranges from 0 to 2. 0 is used for owners, 1 is used for moderators and 2 is used for any other room scoped roles.
 - The fields returned for each user are similar to those from the `.members` endpoints, except from the new `highestRole` object (which contains the `level` and `role` fields) and `statusConnection`. The `roles` field returned within each user contains only room scoped roles (and not global roles, as in the `.members` endpoints).
**Note:** The same applies to the `channels.membersByHighestRole` endpoint

## Further comments

The following rank between room scoped roles is used: **Owner > Moderator > Other room scoped roles**.
TC-837


debdutdeb:
following your description, this is then basically a filter right? I'm not understanding the connection between the description (filtering by `role` field) and the name of the endpoint `membersByHighestRole` - what does it mean by `by highest role` and how does the endpoint respect that definition? 

and if it's just a filter why can it not be added to existing members endpoints?

matheusbsilva137:
> following your description, this is then basically a filter right? 
> 
> and if it's just a filter why can it not be added to existing members endpoints?

Hey @debdutdeb !

Yeah, it is some sort of filter, but since this filter wouldn't be applied to a field that is returned in the endpoint, I felt like this should be something else (new). Just like we have the `channels.list` and `channels.list.joined` endpoints -- the latter can be seen as a "filter", but since it works different internally, this is a whole new endpoint.

Although I've taken the decision to split this in a new endpoint for now, this still needs to be discussed with the Archicteture team, so it may change :thinking: But I'm sure that adding a new param to the `members` endpoints is a feasible approach too.

> I'm not understanding the connection between the description (filtering by `role` field) and the name of the endpoint `membersByHighestRole` - what does it mean by `by highest role` and how does the endpoint respect that definition?

We're not just filtering by `role` in this endpoint. Instead, we've defined a _hierarchy_ between room scoped roles and we're using it to define which is the highest (most important) role of a user in a room.
A user can have multiple roles in a room (eg they can be moderator and owner at the same time), but when filtering by highest role, we'll only return users that do not have any role above the one specified (if moderator is given, we won't return owners; or if leader is given, we won't return moderators or owners).

There is an improvement planned to members lists in rooms: instead of displaying a single list with all the users, we're going to split it in multiple lists (one for owners, other for moderators, other for leaders, and other list for any other role or no role at all) that's where this new endpoint fits.

apps/meteor/server/models/raw/Users.js
```diff
@@ -258,6 +258,112 @@ export class UsersRaw extends BaseRaw {
...
 
+	findPaginatedActiveUsersByRoomIdWithHighestRole(
+		searchTerm,
+		rid,
+		options,
+		searchFields,
+		extraQuery = [],
+		{ startsWith = false, endsWith = false } = {},
+	) {
+		if (options == null) {
```
KevLehman, line 269: Why not making this the default value and update the callers to not pass null? 👀

apps/meteor/server/models/raw/Users.js
```diff
@@ -258,6 +258,112 @@ export class UsersRaw extends BaseRaw {
...
+			options = {};
+		}
+
+		const termRegex = new RegExp((startsWith ? '^' : '') + escapeRegExp(searchTerm) + (endsWith ? '$' : ''), 'i');
```
KevLehman, line 273: Make what's inside a variable , so the name helps to understand better

apps/meteor/server/models/raw/Users.js
```diff
@@ -258,6 +258,112 @@ export class UsersRaw extends BaseRaw {
...
+
+		const orStmt = (searchFields || []).reduce(function (acc, el) {
```
KevLehman, line 275: Can't this be done with a map?

apps/meteor/server/models/raw/Users.js
```diff
@@ -258,6 +258,112 @@ export class UsersRaw extends BaseRaw {
...
+		}, []);
+
+		const query = {
+			$and: [
+				{
+					active: true,
+					__rooms: rid,
+					username: { $exists: true },
+					// if the search term is empty, don't need to have the $or statement (because it would be an empty regex)
+					...(searchTerm && orStmt.length > 0 && { $or: orStmt }),
```
KevLehman, line 287:
```suggestion
					...(searchTerm && orStmt.length && { $or: orStmt }),
```

More of a preference of playing with the boolean conversion of these

apps/meteor/server/models/raw/Users.js
```diff
@@ -258,6 +258,112 @@ export class UsersRaw extends BaseRaw {
...
+			{
+				$project: {
+					...options.projection,
+					roles: '$sub.roles',
+				},
+			},
+			{
+				$addFields: {
+					highestRole: {
+						$cond: [{ $in: ['owner', '$roles'] }, 'owner', { $cond: [{ $in: ['moderator', '$roles'] }, 'moderator', 'member'] }],
```
KevLehman, line 342: A suggestion: instead of doing the same condition twice to check highest and level, why not making this one return an object like { role: owner, level: 1 }

KevLehman (changes requested):

apps/meteor/server/models/raw/Users.js
matheusbsilva137: Done! btw this is the same we're doing on `findPaginatedByActiveUsersExcept`, so I changed it there too

apps/meteor/server/models/raw/Users.js
matheusbsilva137: Done! this is also the same we're doing on `findPaginatedByActiveUsersExcept`, so I changed it there too

apps/meteor/server/models/raw/Users.js
matheusbsilva137: Ohh I agree it'd look better. Update, tkss :)

apps/meteor/server/models/raw/Users.js
matheusbsilva137: Ohh, I guess it can. I've just updated both occurrences

apps/meteor/app/api/server/v1/channels.ts
```diff
@@ -1044,6 +1045,57 @@ API.v1.addRoute(
...
+			const { offset: skip, count: limit } = await getPaginationItems(this.queryParams);
+			const { sort = {} } = await this.parseJsonQuery();
+
+			check(
+				this.queryParams,
+				Match.ObjectIncluding({
+					status: Match.Maybe([String]),
+					filter: Match.Maybe(String),
+				}),
+			);
```
KevLehman, line 1071: no ajv? 👀

apps/meteor/app/api/server/v1/groups.ts
```diff
@@ -738,6 +739,59 @@ API.v1.addRoute(
...
+			const { sort = {} } = await this.parseJsonQuery();
+
+			check(
+				this.queryParams,
+				Match.ObjectIncluding({
+					role: Match.Maybe(String),
+					status: Match.Maybe([String]),
+					filter: Match.Maybe(String),
+				}),
+			);
```
KevLehman, line 766: same

apps/meteor/server/models/raw/Users.js
```diff
@@ -270,18 +262,115 @@ export class UsersRaw extends BaseRaw {
...
 
+	findPaginatedActiveUsersByRoomIdWithHighestRole(
+		searchTerm,
+		rid,
+		searchFields,
+		options = {},
+		extraQuery = [],
+		{ startsWith = false, endsWith = false } = {},
+	) {
+		const regexString = (startsWith ? '^' : '') + escapeRegExp(searchTerm) + (endsWith ? '$' : '');
```
KevLehman, line 273: The regexp match could be problematic perf-wise, generally is better to always check for the match for the start of the string so the index could cover it, but at least on the usage on this PR, the option is not passed so it will try to match without index. 

Depending on the amount of `search fields` and the number of users, this will be slow.

KevLehman: idk if that kind of search is a requirement, so I'm just sharing the thought on here.

apps/meteor/server/models/raw/Users.js
```diff
@@ -270,18 +262,115 @@ export class UsersRaw extends BaseRaw {
...
+		const limit =
+			options.limit !== 0
+				? [
+						{
+							$limit: options.limit,
+						},
+				  ]
+				: [];
+
+		return this.col.aggregate([
```
KevLehman, line 309: A suggestion about this aggregation: i don't see any index that could cover the whole match, it will go either on `active` idx or `__rooms` idx, depending on which one is more selective. 

Can we try to run an explain on an env with a ton of users? Like open, just to see how it performs? It touches 2 of the biggest collections so this could take a ton of time to compute 😬

apps/meteor/server/models/raw/Users.js
```diff
@@ -270,18 +262,115 @@ export class UsersRaw extends BaseRaw {
...
+								...options.sort,
+							},
+						},
+						...skip,
+						...limit,
+					],
+					totalCount: [{ $group: { _id: null, total: { $sum: 1 } } }],
+				},
+			},
+		]);
```
KevLehman, line 359: This query could potentially operate on a ton of results, it may need `allowDiskUse` option to be passed

KevLehman (changes requested):
image

apps/meteor/app/api/server/v1/channels.ts
matheusbsilva137: This one used `ajv` already, so there was no use for this `check` statement. Removed it

apps/meteor/app/api/server/v1/groups.ts
matheusbsilva137: Done!

MarcosSpessatto (changes requested):
Hey @matheusbsilva137, you mentioned this `we'll only return users that do not have any role above the one specified (if moderator is given, we won't return owners; or if leader is given, we won't return moderators or owners).`.  Sorry, but I could not find this implementation in the code, am I missing something?

apps/meteor/app/api/server/v1/groups.ts
```diff
@@ -738,6 +739,49 @@ API.v1.addRoute(
...
+				limit,
+				filter,
+				...(sort?.username && { sort: { username: sort.username } }),
+			});
+
+			const result = await cursor.toArray();
+			const {
+				members,
+				totalCount: [{ total } = { total: 0 }],
+			} = result[0];
```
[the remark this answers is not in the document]
matheusbsilva137: Hmmm I think it'd be good to have it. I've just added this check and a new error (for channels and groups).

apps/meteor/server/models/raw/Users.js
matheusbsilva137: This new endpoint will replace the current `.members` endpoints in the UI (we'll use it to display the members lists in rooms). So it should have the same capabilities as `.members`, which means this is a requirement for now :/

apps/meteor/server/models/raw/Users.js
matheusbsilva137: Hmmm that'd be nice. I'll talk to the architecture team and check if that's possible
Anyways, my guess here is that it should perform fine :thinking: it's essentially the same query we use in `findPaginatedByActiveUsersExcept` (the `__rooms` condition is not explicitly stated there, but it is defined in the `extraQuery` param), which is used in the `.members` endpoints currently (within the `findUsersOfRoom` method)
I can't spot any difference between them other than this new one is running in an aggregation
cc @sampaiodiego

apps/meteor/tests/end-to-end/api/03-groups.js
```diff
@@ -863,6 +863,220 @@ describe('[Groups]', function () {
 		});
 	});
 
+	describe('/groups.membersByHighestRole', () => {
+		let testGroup;
+		let testUser;
+		before('create a group', async () => {
```
MarcosSpessatto, line 869: Same as above.

MarcosSpessatto (changes requested):

apps/meteor/tests/end-to-end/api/03-groups.js
```diff
@@ -863,6 +863,220 @@ describe('[Groups]', function () {
...
+					expect(member.roles[0]).to.be.equal('moderator');
+
+					const { highestRole } = member;
+					expect(highestRole).to.have.property('role', 'moderator');
+					expect(highestRole).to.have.property('level', 1);
+				})
+				.end(done);
+		});
+
+		it('should return the correct highest role when searching for a moderator user', (done) => {
```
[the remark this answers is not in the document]
matheusbsilva137: oops, thanks for pointing this :grimacing:

packages/model-typings/src/models/IUsersModel.ts
```diff
@@ -39,11 +39,19 @@ export interface IUsersModel extends IBaseModel<IUser> {
...
 		searchTerm: any,
 		exceptions: any,
-		options: any,
 		searchFields: any,
+		options?: any,
 		extraQuery?: any,
 		params?: { startsWith?: boolean; endsWith?: boolean },
 	): FindPaginated<FindCursor<T>>;
+	findPaginatedActiveUsersByRoomIdWithHighestRole<T = IUser>(
+		searchTerm: any,
```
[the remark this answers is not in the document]
matheusbsilva137: Done! Since the params for this method are similar to the ones in `findPaginatedByActiveUsersExcept`, I was also trying to type it, but that was harder than I thought :zipper_mouth_face: at least the new one is typed now

MarcosSpessatto (approved):

apps/meteor/tests/end-to-end/api/03-groups.js
```diff
@@ -863,6 +863,220 @@ describe('[Groups]', function () {
...
+					expect(res.body).to.have.property('total');
+					expect(res.body).to.have.property('offset');
+
+					const member = res.body.members[0];
+					expect(member).to.have.property('roles');
+					expect(member).to.have.property('_id');
+					expect(member).to.have.property('username');
+					expect(member).to.have.property('name');
+					expect(member).to.have.property('status');
+					expect(member).to.have.property('highestRole');
```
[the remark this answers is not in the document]
matheusbsilva137: I added this check, but it was causing some troubles, so I had to remove it. I was trying to assure it exists, but this is not a default field for all users. The ones created in the test suite do not have it.

MarcosSpessatto (dismissed):

MarcosSpessatto (dismissed):

.changeset/thick-swans-drop.md
```diff
@@ -0,0 +1,8 @@
+---
+"@rocket.chat/meteor": patch
+"@rocket.chat/core-typings": patch
+"@rocket.chat/model-typings": patch
+"@rocket.chat/rest-typings": patch
+---
+
+Added `membersByHighestRole` endpoints, which enables users to retrieve room members sorted by their room-scoped roles
```
ggazzo, line 8:
```suggestion
---
"@rocket.chat/meteor": minor
"@rocket.chat/core-typings": minor
"@rocket.chat/model-typings": minor
"@rocket.chat/rest-typings": minor
---

Added `membersByHighestRole` endpoints, which enables users to retrieve room members sorted by their room-scoped roles
```

ggazzo (changes requested):

MarcosSpessatto (dismissed):

apps/meteor/server/models/raw/Users.js
matheusbsilva137: Please consider this while reviewing, @ggazzo @sampaiodiego

ggazzo (dismissed):

PR #29871, opened 2023-07-19
fix: Agent count not updating after removing agents

KevLehman:
OC-1043


apps/meteor/app/livechat/server/hooks/afterAgentRemoved.ts
```diff
@@ -0,0 +1,14 @@
+import { LivechatDepartment, Users, LivechatDepartmentAgents, LivechatVisitors } from '@rocket.chat/models';
+
+import { callbacks } from '../../../../lib/callbacks';
+
+callbacks.add('livechat.afterAgentRemoved', async ({ agent }) => {
+	const departmentIds = (await LivechatDepartmentAgents.findByAgentId(agent._id).toArray()).map((department) => department.departmentId);
+	console.log('afterAgentRemoved', agent, departmentIds);
```
aleksandernsilva, line 7: console.log 😿

aleksandernsilva (changes requested):

apps/meteor/app/livechat/server/hooks/afterAgentRemoved.ts
KevLehman: :(

aleksandernsilva (dismissed):

apps/meteor/app/livechat/server/hooks/afterAgentRemoved.ts
```diff
@@ -0,0 +1,13 @@
+import { LivechatDepartment, Users, LivechatDepartmentAgents, LivechatVisitors } from '@rocket.chat/models';
+
+import { callbacks } from '../../../../lib/callbacks';
+
+callbacks.add('livechat.afterAgentRemoved', async ({ agent }) => {
+	const departmentIds = (await LivechatDepartmentAgents.findByAgentId(agent._id).toArray()).map((department) => department.departmentId);
+	await Promise.all([
+		Users.removeAgent(agent._id),
+		LivechatDepartmentAgents.removeByAgentId(agent._id),
+		agent.username && LivechatVisitors.removeContactManagerByUsername(agent.username),
```
murtaza98, line 10: I think this might be a problem since I don't see username prop on the projection of this query

apps/meteor/app/livechat/server/lib/Livechat.js
```diff
@@ -574,11 +574,7 @@ export const Livechat = {
 		const { _id } = user;
 
 		if (await removeUserFromRolesAsync(_id, ['livechat-agent'])) {
-			await Promise.all([
-				Users.removeAgent(_id),
-				LivechatDepartmentAgents.removeByAgentId(_id),
-				LivechatVisitors.removeContactManagerByUsername(username),
-			]);
+			await callbacks.run('livechat.afterAgentRemoved', { agent: user });
```
murtaza98, line 577: The agent param expects `IUser` however here it is getting only Pick<IUser, '_id'>

apps/meteor/server/models/raw/LivechatDepartment.ts
```diff
@@ -312,6 +312,10 @@ export class LivechatDepartmentRaw extends BaseRaw<ILivechatDepartment> implemen
 		return this.updateOne({ _id }, { $set: { numAgents } });
 	}
 
+	decreaseNumberOfAgentsByIds(_ids: string[]): Promise<Document | UpdateResult> {
+		return this.updateOne({ _id: { $in: _ids } }, { $inc: { numAgents: -1 } });
```
murtaza98, line 316: Wouldn't this by updateMany?

murtaza98 (dismissed):

MartinSchoeler (approved):

packages/model-typings/src/models/IUsersModel.ts
```diff
@@ -379,4 +379,9 @@ export interface IUsersModel extends IBaseModel<IUser> {
 	countRemote(options?: FindOptions<IUser>): Promise<number>;
 	findOneByImportId(importId: string, options?: FindOptions<IUser>): Promise<IUser | null>;
 	removeAgent(_id: string): Promise<UpdateResult>;
+	findAgentsWithDepartments<T = ILivechatAgent>(
+		roles: string,
```
MarcosSpessatto, line 383: Isn't this an array of strings also? 🤔

apps/meteor/app/livechat/server/lib/Livechat.js
```diff
@@ -560,10 +564,15 @@ export const Livechat = {
 		return false;
 	},
 
+	async afterRemoveAgent(user) {
+		await callbacks.run('livechat.afterAgentRemoved', { agent: user });
```
MarcosSpessatto, line 568: Any specific reason to put this in a callback? Since it's being called only from here, it'd be clear to just put that code here instead of creating one callback more.

MarcosSpessatto (dismissed):

packages/model-typings/src/models/IUsersModel.ts
murtaza98: No it isn't. If you check the typing here it is a string -> https://github.com/RocketChat/Rocket.Chat/blob/b198164368ac6d56101869c63ac073b3b8ee8b3a/apps/meteor/app/livechat/server/api/lib/users.ts#L15

But I'm glad you raised this because the variable name definitely needs to be updated here to reflect it is singular and not plural

murtaza98 (approved):

MarcosSpessatto (approved):

ggazzo (approved):

PR #29873, opened 2023-07-19
fix: Trim URL on Livechat init

MartinSchoeler:
OC-1176
Fixes an issue with spaces on the url when running the `init()` livechat function


KevLehman (commented):
Changeset :)

KevLehman (approved):

aleksandernsilva (approved):

tassoevan (approved):
Maybe we should also validate the URL for complete sanitization.

PR #29883, opened 2023-07-20
chore: isolate callbacks and typings

ggazzo:
nowadays callbacks group ALL types needed for use, forcing various parts of the application to import types that will not be used

the intention is to allow each callback to have its object and its separate types.

### DIFF
--- a/apps/meteor/app/dolphin/server/lib.ts
+++ b/apps/meteor/app/dolphin/server/lib.ts
@@ -5,6 +5,7 @@ import type { IUser } from '@rocket.chat/core-typings';
 import { settings } from '../../settings/server';
 import { CustomOAuth } from '../../custom-oauth/server/custom_oauth_server';
 import { callbacks } from '../../../lib/callbacks';
+import { beforeCreateUserCallback } from '../../../lib/callbacks/beforeCreateUserCallback';
 
 const config = {
 	serverURL: '',
@@ -22,6 +23,7 @@ const config = {
 const Dolphin = new CustomOAuth('dolphin', config);
 
 function DolphinOnCreateUser(options: any, user?: IUser) {
+	// TODO: callbacks Fix this
 	if (user?.services?.dolphin?.NickName) {
 		user.username = user.services.dolphin.NickName;
 	}
@@ -48,5 +50,5 @@ Meteor.startup(async () => {
 		await ServiceConfiguration.configurations.upsertAsync({ service: 'dolphin' }, { $set: data });
 	}
 
-	callbacks.add('beforeCreateUser', DolphinOnCreateUser, callbacks.priority.HIGH, 'dolphin');
+	beforeCreateUserCallback.add(DolphinOnCreateUser, callbacks.priority.HIGH, 'dolphin');
 });
--- a/apps/meteor/app/integrations/server/triggers.ts
+++ b/apps/meteor/app/integrations/server/triggers.ts
@@ -1,4 +1,5 @@
 import { callbacks } from '../../../lib/callbacks';
+import { afterLeaveRoomCallback } from '../../../lib/callbacks/afterLeaveRoomCallback';
 import { triggerHandler } from './lib/triggerHandler';
 
 const callbackHandler = function _callbackHandler(eventType: string) {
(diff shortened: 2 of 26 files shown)


tassoevan (approved):

apps/meteor/app/dolphin/server/lib.ts
```diff
@@ -22,6 +23,7 @@ const config = {
 const Dolphin = new CustomOAuth('dolphin', config);
 
 function DolphinOnCreateUser(options: any, user?: IUser) {
+	// TODO: callbacks Fix this
```
AlenDavid, line 26: Is this TODO being tracked in a task? Can you create a technical debt ticket to address it if you're not going to do it in this PR?

apps/meteor/app/integrations/server/triggers.ts
```diff
@@ -12,6 +13,6 @@ callbacks.add('afterCreateChannel', callbackHandler('roomCreated'), callbacks.pr
 callbacks.add('afterCreatePrivateGroup', callbackHandler('roomCreated'), callbacks.priority.LOW, 'integrations-roomCreated');
 callbacks.add('afterCreateUser', callbackHandler('userCreated'), callbacks.priority.LOW, 'integrations-userCreated');
 callbacks.add('afterJoinRoom', callbackHandler('roomJoined'), callbacks.priority.LOW, 'integrations-roomJoined');
-callbacks.add('afterLeaveRoom', callbackHandler('roomLeft'), callbacks.priority.LOW, 'integrations-roomLeft');
+afterLeaveRoomCallback.add(callbackHandler('roomLeft'), callbacks.priority.LOW, 'integrations-roomLeft');
 callbacks.add('afterRoomArchived', callbackHandler('roomArchived'), callbacks.priority.LOW, 'integrations-roomArchived');
 callbacks.add('afterFileUpload', callbackHandler('fileUploaded'), callbacks.priority.LOW, 'integrations-fileUploaded');
```
AlenDavid, line 18: This introduces a different style to handle this callbacks. Are we sure we want this changes to propagate? Should we update the other callback methods to reflect this new change?

apps/meteor/app/lib/server/functions/createRoom.ts
```diff
@@ -34,7 +35,15 @@ export const createRoom = async <T extends RoomType>(
 	}
 > => {
 	const { teamId, ...extraData } = roomExtraData || ({} as IRoom);
-	await callbacks.run('beforeCreateRoom', { type, name, owner: ownerUsername, members, readOnly, extraData, options });
+	await beforeCreateRoomCallback.run({
+		type,
+		// name,
```
AlenDavid, line 40: Please avoid having commented code.

apps/meteor/tests/unit/server/federation/infrastructure/rocket-chat/hooks/hooks.spec.ts
```diff
@@ -24,117 +26,155 @@ const { FederationHooks } = proxyquire
 				priority: { HIGH: 'high' },
 				remove,
 				add: (_name: string, callback: (...args: any[]) => void, _priority: string, _id: string) => {
+					console.log('ADD fuck');
```
AlenDavid, line 29: Do we need this console.log?

AlenDavid (commented):
Leaving some comments. Good work here!

apps/meteor/tests/unit/server/federation/infrastructure/rocket-chat/hooks/hooks.spec.ts
ggazzo: nop, my bad I was forced to debug because this injection creates two instances, I though my rebase had removed

apps/meteor/app/integrations/server/triggers.ts
ggazzo: I'm glad you noticed, that's exactly the idea, we want and we're going to use the new way, as we have 1000 callbacks and I don't have a way to do that in a pr.  so its going to be something gradual, btw the use of callbacks is deprecated

apps/meteor/app/dolphin/server/lib.ts
ggazzo: if i had created a task i would not have put a todo in the code. as you can see i prefixed it with 'callbacks' this will be handled in the next tasks related to callbacks

tassoevan (approved):

MartinSchoeler (approved):

MarcosSpessatto (approved):

pierre-lehnen-rc (approved):
:ninja: :eyes:

PR #29884, opened 2023-07-21
chore: delete users after API test is complete

murtaza98:


KevLehman (approved):
Nice, this should give us enough time to work on the refactor :)

PR #29902, opened 2023-07-24
fix: Deduplicate dynamic emails before creating the visitors

KevLehman:
OC-554

## Further comments

Current behavior is:
- Assume we have setup an email inbox a@b.com
- Someone from <hello@rocket.chat> sends an email, with name <Mario>
- Later, someone with name <Wario> sends an email from the same <hello@rocket.chat> (this is possible with most emails, such as notifications, newsletters, programattic emails, etc)
- Our system will incorrectly assume both emails are from _the exact same person_, and will use the data from the first email to create the visitor, causing all subsequent emails from <hello@rocket.chat> to be shown as if "Mario" wrote them, ignoring "Wario".

New behavior takes this into consideration by providing the email we receive, and generating a more "emailish" name for visitors from this channel of the form "Name <email>".

We'll revisit this whole behavior when this feature gets moved outside of Core, to handle other cases ignored right now, which will provide support for the "unique visitor" ideal we want to have :)


murtaza98 (approved):

MarcosSpessatto (dismissed):

ggazzo (dismissed):

murtaza98:
There were some issues with the approach taken on this PR as the fix was working for email channels but was breaking other channels. We'd need to re-visit this... until then keeping this as draft

murtaza98 (approved):

PR #29914, opened 2023-07-25
fix: Performance issue when using api to create users

rodrigok:


apps/meteor/server/models/raw/Users.js
```diff
@@ -66,6 +66,18 @@ export class UsersRaw extends BaseRaw {
 			{ key: { language: 1 }, sparse: true },
 			{ key: { 'active': 1, 'services.email2fa.enabled': 1 }, sparse: true }, // used by statistics
 			{ key: { 'active': 1, 'services.totp.enabled': 1 }, sparse: true }, // used by statistics
+			// Used for canse insensitive queries
```
KevLehman, line 69:
```suggestion
			// Used for case insensitive queries
```

ggazzo (approved):

PR #29916, opened 2023-07-25
refactor: Unify all the code for user creation into a new service.

pierre-lehnen-rc:

Moves the user creation code from Meteor's `Accounts.createUserAsync` and `Accounts.insertUserDoc` to a new `User` service on rocket.chat, with only minimal changes to behavior.

Changes:
 - `Accounts.createUserAsync` would query the user collection to check if the specified username and email were not already in use - but this check was only valid in regular user registration because in many of the places we call this function (such as OAuth), the username and email were filled in through callbacks after this check is done. So I dropped this check completely as in one way or another we already rely on the indexes for those fields.
 - In some places we passed an attribute named `globalRoles` to the user param when creating users with `Accounts.insertUserDoc`. I moved this attribute to the options param instead, so that the user object is compatible with `IUser`.
 - `Accounts.onCreateUser` and `Accounts.validateNewUser` will no longer be called. We had a few references to those in the code base and I moved all of them into the User service as well, **except** for the OAuth `validateNewUser` function that I moved into the `beforeCreateUser` callback instead.
 - Any call to `Accounts.insertUserDoc` will be redirected to this new service automatically. OAuth login is still handled by Meteor code so it relies on this. I made no such fallbacks for `Accounts.validateNewUser`.
 - Includes a change made on PR #29925 to save the user's roles before running the afterCreateUser callback (If that PR is already merged then this is not an additional change)

## Further comments

This adds a reference to a new meteor package, `oauth-encryption`. This package was previously used internally on `Accounts.insertUserDoc` to handle a special case in OAuth logins, where some data may have been encrypted and stored before the user's ID was generated. The `insertUserDoc` function decrypts that data and re-encrypts using the ID once it is available.


apps/meteor/server/services/user/service.ts
```diff
@@ -0,0 +1,333 @@
...
+			globalRoles.push(...options.globalRoles);
+		}
+
+		delete options.globalRoles;
+
+		if (doc.services && !doc.services.password) {
+			const defaultAuthServiceRoles = parseCSV(settings.get('Accounts_Registration_AuthenticationServices_Default_Roles') || '');
+
+			if (defaultAuthServiceRoles.length > 0) {
+				globalRoles.push(...defaultAuthServiceRoles);
```
KevLehman, line 154: Think concat should be faster in this case

apps/meteor/server/services/user/service.ts
```diff
@@ -0,0 +1,333 @@
...
+							address: service.email,
+							verified,
+						},
+					];
+				}
+
+				Object.keys(service).forEach((key) => {
+					const value = service[key];
+					// If there's any value that was encrypted without an userId, decrypt it and re-encrypt with the id.
+					if (OAuthEncryption?.isSealed(value)) {
```
KevLehman, line 197: Isn't there an alternative to this one that's not a meteor package? 👀

apps/meteor/server/services/user/service.ts
```diff
@@ -0,0 +1,333 @@
...
+		}
+
+		if (!hasAdmin && roles.includes('admin') && settings.get('Show_Setup_Wizard') === 'pending') {
+			await Settings.updateValueById('Show_Setup_Wizard', 'in_progress');
+		}
+
+		if (newUser.username) {
+			await this.addUserToDefaultChannels(newUser, options);
+			await this.callAfterCreateUser(newUser);
+			await this.setDefaultAvatar(newUser);
```
KevLehman, line 241: Can this use a Promise.all?

apps/meteor/server/services/user/service.ts
```diff
@@ -0,0 +1,333 @@
...
+		if (settings.get('Accounts_SetDefaultAvatar') === true) {
+			const avatarSuggestions = await getAvatarSuggestionForUser(user);
+
+			for await (const service of Object.keys(avatarSuggestions)) {
+				const avatarData = avatarSuggestions[service];
+				if (service !== 'gravatar') {
+					return setAvatarFromServiceWithValidation(user._id, avatarData.blob, '', service);
+				}
+			}
+		}
```
KevLehman, line 275:
```suggestion
		if (settings.get('Accounts_SetDefaultAvatar') !== true) {
			return;
		}
		const avatarSuggestions = await getAvatarSuggestionForUser(user);
		for await (const service of Object.keys(avatarSuggestions)) {
			const avatarData = avatarSuggestions[service];
			if (service !== 'gravatar') {
				return setAvatarFromServiceWithValidation(user._id, avatarData.blob, '', service);
			}
		}
```

apps/meteor/server/services/user/service.ts
```diff
@@ -0,0 +1,333 @@
...
+		const _id = await this.insertOne(user);
+
+		const hasAdmin = await Users.findOneByRolesAndType('admin', 'user', { projection: { _id: 1 } });
+		if (!roles.includes('admin') && !hasAdmin) {
+			roles.push('admin');
+		}
+
+		await addUserRolesAsync(_id, roles);
+
+		const newUser = await Users.findOne({
```
KevLehman, line 226: Maybe `findOneById`

pierre-lehnen-rc:
As this first PR focuses on unifying the code while changing as little as possible, I've marked as resolved the suggestions related to code that was only moved from one file to another. I'm working on a second PR with the actual improvements over this code and I'll consider those suggestions there.

apps/meteor/server/services/user/service.ts
pierre-lehnen-rc: I didn't want to affect the order in which they are run - yet.

apps/meteor/server/services/user/service.ts
```diff
@@ -0,0 +1,333 @@
...
+			active: doc.active ?? !settings.get('Accounts_ManuallyApproveNewUsers'),
+			...(!!name && { name }),
+			roles: [],
+		};
+
+		if (user.services) {
+			const verified = settings.get<boolean>('Accounts_Verify_Email_For_External_Accounts');
+
+			for (const service of Object.values(user.services) as Record<string, any>[]) {
+				if (!user.name && typeof service === 'object' && ('name' in service || 'username' in service)) {
```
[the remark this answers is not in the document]
pierre-lehnen-rc: Ideally, yes. In this case the type this would assert to is not defined yet. I'll try to include its definition on the next PR.

PR #29925, opened 2023-07-26
regression: Server isn't adding agent required properties to users created/updated through admin UI

murtaza98:
(cherry picked from commit 932b81a0a97fbf145bd4a97a76a3d81607f26f97)

## Issue(s)

https://rocketchat.atlassian.net/browse/OC-1194

### DIFF
--- a/apps/meteor/app/lib/server/functions/saveUser.js
+++ b/apps/meteor/app/lib/server/functions/saveUser.js
@@ -329,6 +329,12 @@ const saveNewUser = async function (userData, sendPassword) {
 		}
 	}
 
+	const userUpdated = await Users.findOneById(_id);
+	await callbacks.run('afterSaveUser', {
+		user: userUpdated,
+		oldUser: null,
+	});
+
 	return _id;
 };
 
--- a/apps/meteor/app/lib/server/functions/saveUser.js
+++ b/apps/meteor/app/lib/server/functions/saveUser.js
@@ -329,12 +329,6 @@ const saveNewUser = async function (userData, sendPassword) {
 		}
 	}
 
-	const userUpdated = await Users.findOneById(_id);
-	await callbacks.run('afterSaveUser', {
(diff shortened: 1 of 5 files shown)


KevLehman (approved):

pierre-lehnen-rc:
Isn't that by design? There's an `afterCreateUser` callback as well and I see some places adding the same function to both.

apps/meteor/app/lib/server/functions/saveUser.js
```diff
@@ -329,6 +329,12 @@ const saveNewUser = async function (userData, sendPassword) {
 		}
 	}
 
+	const userUpdated = await Users.findOneById(_id);
+	await callbacks.run('afterSaveUser', {
+		user: userUpdated,
+		oldUser: null,
+	});
+
```
ggazzo, line 337: this is not supposed to be an after create user?

apps/meteor/app/lib/server/functions/saveUser.js
```diff
@@ -329,6 +329,12 @@ const saveNewUser = async function (userData, sendPassword) {
 		}
 	}
 
+	const userUpdated = await Users.findOneById(_id);
+	await callbacks.run('afterSaveUser', {
+		user: userUpdated,
+		oldUser: null,
+	});
```
ggazzo, line 336:
```suggestion
	await callbacks.run('afterSaveUser', {
		user: await Users.findOneById(_id),
		oldUser: null,
	});
```

ggazzo (dismissed):
if, and I'm not saying this is the right way, but if it were the right solution, it wouldn't be a regression it wasn't just introduced, it's old behavior, it deserves a changeset and a proper commit. also i would point this pr to release and not to your own pr

pierre-lehnen-rc (approved):

KevLehman (approved):

PR #29933, opened 2023-07-26
fix: Block user creation if theres no assigned role

rique223:


ggazzo (changes requested):
keep refactors/improvements/fix in different pull requests

PR #29935, opened 2023-07-27
feat: new Endpoints for Bulk User Creation

pierre-lehnen-rc:


apps/meteor/app/importer/server/classes/ImportDataConverter.ts
```diff
@@ -281,27 +288,69 @@ export class ImportDataConverter {
...
+				joinDefaultChannels: false,
+				skipEmailValidation: true,
+				skipAdminCheck: true,
+				skipAdminEmail: true,
+				skipOnCreateUserCallback: this._options.skipUserCallbacks,
+				skipBeforeCreateUserCallback: this._options.skipUserCallbacks,
+				skipAfterCreateUserCallback: this._options.skipUserCallbacks,
+				skipDefaultAvatar: true,
+				skipAppsEngineEvent: !!process.env.IMPORTER_SKIP_APPS_EVENT,
+				...(userData.roles?.length ? { globalRoles: userData.roles } : {}),
```
pierre-lehnen-rc, line 320: globalRoles must be sent to the second argument instead of the first.

pierre-lehnen-rc (commented):
Note for myself.

apps/meteor/app/api/server/v1/users.ts
```diff
@@ -393,10 +393,16 @@ API.v1.addRoute(
...
 			const { fields } = await this.parseJsonQuery();
 
-			const user = await getFullUserDataByIdOrUsername(this.userId, {
-				filterId: (this.queryParams as any).userId,
-				filterUsername: (this.queryParams as any).username,
-			});
+			const searchTerms: [string, 'id' | 'username' | 'importId'] | false =
+				('userId' in this.queryParams && !!this.queryParams.userId && [this.queryParams.userId, 'id']) ||
+				('username' in this.queryParams && !!this.queryParams.username && [this.queryParams.username, 'username']) ||
+				('importId' in this.queryParams && !!this.queryParams.importId && [this.queryParams.importId, 'importId']);
```
KevLehman, line 399: Would it be the same to use ajv like this https://stackoverflow.com/a/48778838 ?

We validate they're not empty when present, and on here, you can just
`searcterms = this.queryparams.userid || username || importid`

apps/meteor/app/authentication/server/startup/index.js
```diff
@@ -157,8 +157,34 @@ const getLinkedInName = ({ firstName, lastName }) => {
 	return `${firstName} ${lastName}`;
 };
 
+const validateEmailDomain = (user) => {
+	if (user.type === 'visitor') {
+		return true;
+	}
+
+	let domainWhiteList = settings.get('Accounts_AllowedDomainsList');
+	if (_.isEmpty(domainWhiteList?.trim())) {
```
KevLehman, line 166: DDP Streamer has an isEmpty function, **ee/apps/ddp-streamer/src/lib/utils.ts**

Not sure if moving that to the `utils` package and using it would be better here.

apps/meteor/app/authentication/server/startup/index.js
```diff
@@ -157,8 +157,34 @@ const getLinkedInName = ({ firstName, lastName }) => {
...
+		return true;
+	}
+
+	domainWhiteList = domainWhiteList.split(',').map((domain) => domain.trim());
+
+	if (user.emails && user.emails.length > 0) {
+		const email = user.emails[0].address;
+		const inWhiteList = domainWhiteList.some((domain) => email.match(`@${escapeRegExp(domain)}$`));
+
+		if (inWhiteList === false) {
```
KevLehman, line 176: `.some` should only return boolean so
```suggestion
		if (!inWhiteList) {
```

apps/meteor/app/api/server/v1/users.ts
pierre-lehnen-rc: In this case other than the value I also need to know which attribute the value came from, so that wouldn't work. Ajv is already validating the param structure.

apps/meteor/app/authentication/server/startup/index.js
pierre-lehnen-rc: This piece is not new code, I just moved it from a callback to a function. I intend to remove it completely in a subsequent PR but for this release I'm not changing the implementation.

ggazzo (dismissed):

KevLehman (dismissed):
None of the comments need to be addressed by this PR, but maybe on the improvements one :)

apps/meteor/ee/server/startup/seatsCap.ts
```diff
@@ -30,6 +30,17 @@ callbacks.add(
 	'check-max-user-seats',
 );
 
+callbacks.add(
+	'beforeUserImport',
+	async ({ usersCount }: { usersCount: number }) => {
```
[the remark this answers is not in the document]
pierre-lehnen-rc: You just caught a bug ;)

KevLehman (approved):

ggazzo (approved):

PR #29507, opened 2023-06-09
chore: Use FlowRoute through a facade

tassoevan:
## Issue(s)

[ARCH-1127](https://rocketchat.atlassian.net/browse/ARCH-1127)

## Further comments

[ARCH-1127]: https://rocketchat.atlassian.net/browse/ARCH-1127?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #29945, opened 2023-07-27
regression: Login with SAML not working

matheusbsilva137:

 - Create a new `/saml` route for logins with SAML, which redirects to the home page on succesful logins.

## Issue(s)

Introduced in #29507

## Further comments

TC-857


pierre-lehnen-rc (dismissed):

pierre-lehnen-rc (approved):

tassoevan (approved):
Oh, man. It took more time to realize what was going on than I want to admit. Around `<HomePage />` there is `<MainLayout />`, and it does "tricks" when there is no user logged.

PR #29953, opened 2023-07-28
fix: Update channels.ts

zero-commit:
added some extra lines to sort members object on the basis of username. line no 1036 to line no 1044


debdutdeb:
we are already sorting by username, check mongo query's options parameter.

i'm closing this one, but please feel free to ask any questions

PR #29955, opened 2023-07-28
refactor: UiKit surfaces and related stuff

tiagoevanp:


apps/meteor/client/components/message/uikit/UiKitMessage.tsx
```diff
@@ -28,14 +28,14 @@ const patchMessageParser = () => {
...
-type UiKitSurfaceProps = {
+type UiKitMessageProps = {
 	mid: IMessage['_id'];
 	blocks: MessageSurfaceLayout;
 	rid: IRoom['_id'];
 	appId?: string | boolean; // TODO: this is a hack while the context value is not properly typed
 };
 
-const UiKitSurface = ({ mid: _mid, blocks, rid, appId }: UiKitSurfaceProps): ReactElement => {
+const UiKitMessage = ({ mid: _mid, blocks, rid, appId }: UiKitMessageProps): ReactElement => {
```
tassoevan, line 38: I think this name gives the user the wrong idea of what it's supposed to be. I myself would expect `UiKitMessage` to be a complete message component with rendered content from UIKit, when in reality is just a fraction of content that is embedded into a complete message component.

tassoevan (changes requested):

tassoevan (approved):

PR #29963, opened 2023-07-29
fix: Performance issue on username availability check

rodrigok:


pierre-lehnen-rc:
This will also drastrically improve room creation performance, as we use `Users.findOneByUsernameIgnoringCase` to load the room owner's data.

ggazzo (approved):

PR #29969, opened 2023-07-31
fix: users without field roles in db

tapiarafael:

Fix the problem of users not having the field `roles` in database when the default role setting is not set.

## Steps to test or reproduce
1. Go to workspace > configuration > accounts > register;
 Set the Registration Form as `Secret URL`;
 Set the Default Roles for Users field as blank;
2. Using the registration url, create a user (`https://[workspace_url]/register/[secret_hash]`)
3. With an admin user, go to [workspace > users] and try to check the created user info;

A grey screen with an error will be shown since the user doesn't have the `roles` field. It also possible to check via DB in the `users` collection.

## Further comments

[SUP-207](https://rocketchat.atlassian.net/browse/SUP-207)

[SUP-207]: https://rocketchat.atlassian.net/browse/SUP-207?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/meteor/server/methods/registerUser.ts
```diff
@@ -129,6 +129,12 @@ Meteor.methods<ServerMethods>({
 			// throw new Meteor.Error 'error-email-send-failed', 'Error trying to send email: ' + error.message, { method: 'registerUser', message: error.message }
 		}
 
+		const defaultRoles = settings.get('Accounts_Registration_Users_Default_Roles');
+
+		if (!defaultRoles) {
+			await Users.updateOne({ _id: userId }, { $set: { roles: [] } });
+		}
+
```
pierre-lehnen-rc, line 137:
```suggestion
```

Do not handle this here, but on the same place where the roles are loaded today:

https://github.com/RocketChat/Rocket.Chat/blob/dce4a829fa63fcb55ede1bd49bbe0b78cb927af9/apps/meteor/app/authentication/server/startup/index.js#L253

And instead of using an empty array, add `user` to the role list if the list is empty.

apps/meteor/server/startup/migrations/v303.ts
```diff
@@ -0,0 +1,20 @@
...
+	async up() {
+		const defaultRoles = settings.get('Accounts_Registration_Users_Default_Roles');
+
+		await Users.updateMany(
+			{
+				roles: { $exists: false },
+			},
+			{ $set: { roles: defaultRoles ? parseCSV(String(defaultRoles)) : [] } },
+		);
+	},
```
pierre-lehnen-rc, line 19: Instead of loading the setting value directly, use `getNewUserRoles` to handle the parsing for you.
If the list is empty, set them to `['user']` instead of an empty array.

apps/meteor/server/startup/migrations/index.ts
```diff
@@ -35,4 +35,6 @@ import './v298';
 import './v299';
 import './v300';
 import './v301';
+import './v302';
+import './v303';
```
pierre-lehnen-rc, line 39: You're importing two new files but only creating one new migration.

apps/meteor/server/methods/registerUser.ts
```diff
@@ -129,6 +129,12 @@ Meteor.methods<ServerMethods>({
 			// throw new Meteor.Error 'error-email-send-failed', 'Error trying to send email: ' + error.message, { method: 'registerUser', message: error.message }
 		}
 
+		const defaultRoles = settings.get('Accounts_Registration_Users_Default_Roles');
+
+		if (!defaultRoles) {
+			await Users.updateOne({ _id: userId }, { $set: { roles: [] } });
```
matheusbsilva137, line 135: I'd suggest adding a new method to the `Users` model instead of calling this directly :thinking:

pierre-lehnen-rc (changes requested):

apps/meteor/server/methods/registerUser.ts
debdutdeb: > And instead of using an empty array, add user to the role list if the list is empty.
but that goes against the intention, what iF X wants to not set any default role?

debdutdeb: if we want at least one role, and that being "user", it should reflect on the ui as well.

tapiarafael: Okay, I'll move the logic to the correct place.

> And instead of using an empty array, add user to the role list if the list is empty.

Aren't we forcing some role if the default role is empty? This is the correct behavior to expect from the application?

apps/meteor/server/startup/migrations/index.ts
tapiarafael: Because the `v302` was not imported before. Should I ignore it, then?

apps/meteor/server/methods/registerUser.ts
tapiarafael: We already have a method to insert a role to a user. We can use `addUserRoleAsync` or even `Roles.addUserRoles`; however, none of them allow us to insert an empty array of roles or a role with an invalid id. Since this is an isolated scenario and we, in fact, don't want to be this to become a common thing, I think that it's ok for us to update directly.
But please, share your thoughts if you disagree.

apps/meteor/server/lib/roles/addUserRoles.ts
```diff
@@ -14,6 +14,11 @@ export const addUserRolesAsync = async (userId: IUser['_id'], roleIds: IRole['_i
 		throw new MeteorError('error-invalid-user', 'Invalid user');
 	}
 
+	if (!roleIds?.length) {
```
pierre-lehnen-rc, line 17: The purpose of this function is only to add new roles to users (or to subscriptions), it should never remove any role that may already be assigned.

apps/meteor/server/startup/migrations/index.ts
pierre-lehnen-rc: As it's unrelated to your PR you can ignore it. I'll figure out why it's there and fix it on a separate PR.

pierre-lehnen-rc (changes requested):

apps/meteor/server/methods/registerUser.ts
pierre-lehnen-rc: We have several places that force the 'user' role if nothing else is set, but there's actually no reason for that. Leaving it empty is fine.

pierre-lehnen-rc (approved):

pierre-lehnen-rc (changes requested):
Actually, here on `insertUserDoc` you can set the empty array to the `user` object directly and it will be included on the first insertion, so the extra update operation is not needed.

pierre-lehnen-rc (dismissed):

matheusbsilva137 (dismissed):

apps/meteor/server/startup/migrations/v303.ts
```diff
@@ -0,0 +1,19 @@
...
+	name: 'Add default roles to users without roles field',
+	async up() {
+		const defaultRoles = getNewUserRoles();
+
+		await Users.updateMany(
+			{
+				roles: { $exists: false },
+			},
+			{ $set: { roles: defaultRoles } },
+		);
```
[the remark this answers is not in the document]
debdutdeb: IMO setting to just empty array is better here, goes with existing state and expectation.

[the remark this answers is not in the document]
tapiarafael: Yeah, that makes sense. I was thinking about a similar problem that we had with roles. But the best way to deal with this scenario is just to set an empty array of roles. The problem was only possible because the Default Role setting was set to empty.

debdutdeb (approved):

debdutdeb (approved):

pierre-lehnen-rc (approved):

ggazzo (approved):

PR #29984, opened 2023-08-01
fix: hide dm with deleted user

heitortanoue:

DM with deleted user (or deleted DM room) can now be hidden if the subscription still exists.

## Steps to test or reproduce

1. Open a DM with someone so that it appears at the side bar
2. Delete this DM room directly inside the database (I used MongoDB Compass for this step)
3. Try to open the DM room now, it will throw an error (room not found). If you try to hide it, it will show an unauthorized warning.

## Further comments

SUP-245

### DIFF
--- a/apps/meteor/app/api/server/v1/im.ts
+++ b/apps/meteor/app/api/server/v1/im.ts
@@ -123,18 +123,35 @@ API.v1.addRoute(
 			if (!roomId) {
 				throw new Meteor.Error('error-room-param-not-provided', 'Body param "roomId" is required');
 			}
-			const canAccess = await canAccessRoomIdAsync(roomId, this.userId);
-			if (!canAccess) {
-				return API.v1.unauthorized();
-			}
 
-			const { room, subscription } = await findDirectMessageRoom({ roomId }, this.userId);
+            let rid: IRoom["_id"] = roomId;
+            let subscription: ISubscription | null = null;
 
-			if (!subscription?.open) {
-				return API.v1.failure(`The direct message room, is already closed to the sender`);
-			}
+            const roomExists = !!(await Rooms.findOneById(roomId));
+            if (!roomExists) {
+                // if the room doesnt exists, should check if the user has a subscription to it, and if has, close it
+                subscription = await Subscriptions.findOneByRoomIdAndUserId(roomId, this.userId);
+            } else {
+                const canAccess = await canAccessRoomIdAsync(roomId, this.userId);
+                if (!canAccess) {
+                    return API.v1.unauthorized();
+                }
 
-			await hideRoomMethod(this.userId, room._id);
+                const { room, subscription: subs } = await findDirectMessageRoom({ roomId }, this.userId);
+
+                rid = room._id;
+                subscription = subs;
+            }
+
+            if (!subscription) {
+                return API.v1.failure(`The user is not subscribed to the room`);
+            }
+
+            if (!subscription.open) {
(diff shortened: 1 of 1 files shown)


apps/meteor/app/api/server/v1/im.ts
```diff
@@ -123,18 +123,35 @@ API.v1.addRoute(
 			if (!roomId) {
 				throw new Meteor.Error('error-room-param-not-provided', 'Body param "roomId" is required');
 			}
-			const canAccess = await canAccessRoomIdAsync(roomId, this.userId);
-			if (!canAccess) {
-				return API.v1.unauthorized();
+
+			let rid: IRoom['_id'] = roomId;
+			let subscription: ISubscription | null = null;
```
matheusbsilva137, line 128:
```suggestion
			let subscription;
```
We only need to store the subscription here, and there's no need to specify the type since it can be infered from the returns later.
Generally, we should avoid using `let`. But I think that's a good use case for it.

apps/meteor/app/api/server/v1/im.ts
```diff
@@ -123,18 +123,35 @@ API.v1.addRoute(
...
+
+			const roomExists = !!(await Rooms.findOneById(roomId));
+			if (!roomExists) {
+				// if the room doesnt exists, should check if the user has a subscription to it, and if has, close it
```
matheusbsilva137, line 132:
```suggestion
				// even if the room doesn't exist, we should allow the user to close the subscription anyways
```
being picky here, but trying to make our intentions more explicit

apps/meteor/app/api/server/v1/im.ts
```diff
@@ -123,18 +123,35 @@ API.v1.addRoute(
...
+				subscription = await Subscriptions.findOneByRoomIdAndUserId(roomId, this.userId);
+			} else {
+				const canAccess = await canAccessRoomIdAsync(roomId, this.userId);
+				if (!canAccess) {
+					return API.v1.unauthorized();
+				}
+
+				const { room, subscription: subs } = await findDirectMessageRoom({ roomId }, this.userId);
+
+				rid = room._id;
```
matheusbsilva137, line 142:
```suggestion
				const { subscription: subs } = await findDirectMessageRoom({ roomId }, this.userId);
```
We don't really need to get the room object since we're only using its `_id` property (which we already have, `roomId`). We do know that the `roomId` provided is correct since it's been validated internally by the `findDirectMessageRoom` function.

apps/meteor/app/api/server/v1/im.ts
```diff
@@ -123,18 +123,35 @@ API.v1.addRoute(
...
+				return API.v1.failure(`The user is not subscribed to the room`);
+			}
 
-			if (!subscription?.open) {
+			if (!subscription.open) {
 				return API.v1.failure(`The direct message room, is already closed to the sender`);
 			}
 
-			await hideRoomMethod(this.userId, room._id);
+			await hideRoomMethod(this.userId, rid);
```
matheusbsilva137, line 154:
```suggestion
			await hideRoomMethod(this.userId, roomId);
```
Saving one variable here, let's use the one we already have.

matheusbsilva137 (changes requested):

matheusbsilva137 (approved):

MarcosSpessatto:
Hey @heitortanoue, is this happening only when we delete the users directly on the database, or is it happening even if we delete by the "regular" methods (e.g. using the UI, etc)?

PR #29985, opened 2023-08-01
refactor: PasswordVerifier

ggazzo:


packages/ui-contexts/src/hooks/useVerifyPassword.ts
```diff
@@ -22,38 +20,44 @@ export const useVerifyPassword = (password?: string) => {
...
+
+export const useVerifyPasswordByPolices = (policies: PasswordPolicies) => {
+	return useCallback(
+		(password: string) => {
+			return policies.reduce((passwordVerifications, [name, rules]) => {
+				if (name === 'get-password-policy-forbidRepeatingCharacters') return passwordVerifications;
+
+				if (rules) {
+					passwordVerifications[name] = {
+						isValid: password.length < 0 || passwordVerificationsTemplate[name](password, Object.values(rules)[0]),
```
rique223, line 41: Is the verification `password.length < 0` correct? Strings cant have negative lengths and I believe the `watch('password')` function never returns a negative number so this check would be meaningless.

packages/ui-contexts/src/hooks/useVerifyPassword.ts
```diff
@@ -22,38 +20,44 @@ export const useVerifyPassword = (password?: string) => {
...
+						limit: Object.values(rules)[0],
+					};
+					return passwordVerifications;
+				}
+
+				passwordVerifications[name] = {
+					isValid: password.length < 0 || passwordVerificationsTemplate[name](password),
```
rique223, line 48: Same here

packages/ui-client/src/components/PasswordVerifier/PasswordVerifier.tsx
```diff
@@ -14,7 +14,7 @@ export const PasswordVerifier = ({ password }: PasswordVerifierProps) => {
 
 	const passwordVerifications = useVerifyPassword(password);
 
-	if (!passwordVerifications) {
+	if (!passwordVerifications.length) {
```
rique223, line 17: passwordVerifications is not an array

rique223 (changes requested):

packages/ui-contexts/src/hooks/useVerifyPassword.ts
```diff
@@ -22,38 +20,45 @@ export const useVerifyPassword = (password?: string) => {
...
+	return useCallback(
+		(password: string): PasswordVerifications => {
+			return policies
+				.map(([name, rules]) => {
+					if (name === 'get-password-policy-forbidRepeatingCharacters') return;
+
+					if (rules) {
+						return {
+							name,
+							isValid: password.length === 0 || passwordVerificationsTemplate[name](password, Object.values(rules)[0]),
```
rique223, line 43: This verification should be password.length !== 0

packages/ui-contexts/src/hooks/useVerifyPassword.ts
```diff
@@ -22,38 +20,45 @@ export const useVerifyPassword = (password?: string) => {
...
+							limit: Object.values(rules)[0],
+						};
+					}
+
+					return {
+						name,
+						isValid: password.length === 0 || passwordVerificationsTemplate[name](password),
```
rique223, line 50: Same here

rique223 (changes requested):

rique223 (approved):

PR #29993, opened 2023-08-02
feat: Accessibility Page

juliajforesti:


packages/rest-typings/src/v1/users/UsersSetPreferenceParamsPOST.ts
```diff
@@ -4,6 +4,9 @@ import Ajv from 'ajv';
 const ajv = new Ajv({
 	coerceTypes: true,
 });
+export type FontSize = '100%' | '14px' | '18px' | '20px' | '24px' | '28px' | '32px' | '36px' | '40px' | '44px' | '48px';
+
+// export type FontSize = '14' | '18' | '20' | '24' | '28' | '32' | '36' | '40' | '44' | '48';
```
MarcosSpessatto, line 9:
```suggestion
```

packages/rest-typings/src/v1/users/UsersSetPreferenceParamsPOST.ts
```diff
@@ -42,6 +45,7 @@ export type UsersSetPreferencesParamsPOST = {
 		dontAskAgainList?: Array<{ action: string; label: string }>;
 		featuresPreview?: { name: string; value: boolean }[];
 		themeAppearence?: ThemePreference;
+		fontSize?: FontSize;
```
MarcosSpessatto, line 48: Can we add this `fontSize` property here: https://github.com/RocketChat/Rocket.Chat/blob/develop/apps/meteor/server/methods/saveUserPreferences.ts#L7 and here: https://github.com/RocketChat/Rocket.Chat/blob/develop/apps/meteor/server/methods/saveUserPreferences.ts#L51 as well?

MarcosSpessatto (changes requested):

PR #30010, opened 2023-08-03
refactor: integrate backend and frontend password policies

felipe-rod123:
## Further comments

WM-133


apps/meteor/app/lib/server/functions/saveUser.js
```diff
@@ -382,12 +383,18 @@ export const saveUser = async function (userId, userData) {
...
-	) {
+	const validPassword = () => {
+		try {
+			passwordPolicy.validate(userData.password);
+		} catch (err) {
+			if (err instanceof PasswordPolicyError) {
+				throw new Meteor.Error(err.error, err.message, err.reasons);
+			}
+			throw err;
+		}
```
debdutdeb, line 394: should, instead of converting to a MeteorError, making frontend not expecting a meteorerror, not be the way? idk maybe that's a lot of work idk, just curious.

apps/meteor/app/lib/server/lib/passwordPolicy.ts
```diff
@@ -0,0 +1,44 @@
+import PasswordPolicy from '@rocket.chat/account-utils';
+
+import { settings } from '../../../settings/server';
+
+export const passwordPolicy = new PasswordPolicy.PasswordPolicy({
```
MarcosSpessatto, line 5: Can we import the class only? e.g. (`import { PasswordPolicy } from '@rocket.chat/account-utils'`);

apps/meteor/server/methods/registerUser.ts
```diff
@@ -82,8 +82,6 @@ Meteor.methods<ServerMethods>({
 			}
 		}
 
-		passwordPolicy.validate(formData.pass);
```
MarcosSpessatto, line 85: Shouldn't we keep the validation here? 🤔

apps/meteor/server/methods/saveUserProfile.ts
```diff
@@ -19,6 +20,7 @@ import { compareUserPasswordHistory } from '../lib/compareUserPasswordHistory';
 const MAX_BIO_LENGTH = 260;
 const MAX_NICKNAME_LENGTH = 120;
 
+// eslint-disable-next-line complexity
```
MarcosSpessatto, line 23: I'd suggest not disabling the warning here, it'd be better to have it explicit, so we can refactor it as soon as possible.

apps/meteor/server/methods/saveUserProfile.ts
```diff
@@ -128,7 +130,14 @@ async function saveUserProfile(
 				});
 			}
 
-			passwordPolicy.validate(settings.newPassword);
+			try {
```
MarcosSpessatto, line 133: Shouldn't it be better to move this logic to a function to some place instead of replicating it everywhere?

packages/account-utils/src/PasswordPolicyClass.ts
```diff
@@ -34,25 +74,108 @@ class PasswordPolicy {
...
-	set forbidRepeatingCharactersCount(value) {
-		this._forbidRepeatingCharactersCount = value;
-		this.regex.forbiddingRepeatingCharacters = new RegExp(`(.)\\1{${this.forbidRepeatingCharactersCount},}`);
+	get passwordForbidRepeatingCharactersCount() {
+		return this.forbidRepeatingCharactersCount;
 	}
 
-	get forbidRepeatingCharactersCount() {
-		return this._forbidRepeatingCharactersCount;
+	set passwordForbidRepeatingCharactersCount(value: number) {
```
MarcosSpessatto, line 81: I believe it would be better to remove all the `set`s and let the people use only through the constructor.

packages/account-utils/src/PasswordPolicyClass.ts
```diff
@@ -1,8 +1,36 @@
-import { Random } from '@rocket.chat/random';
 import generator from 'generate-password';
-import { Meteor } from 'meteor/meteor';
+
+import { PasswordPolicyError } from './PasswordPolicyError';
 
 class PasswordPolicy {
+	regex: {
```
MarcosSpessatto, line 6: I think it should be better to encapsulate all the properties as `private`. Wdyt?

apps/meteor/app/lib/server/functions/saveUser.js
```diff
@@ -1,3 +1,4 @@
+import { PasswordPolicyError } from '@rocket.chat/account-utils';
```
MarcosSpessatto, line 1: I think it is worth moving this to the `utils` package instead of creating a new one, any opinion on this?

MarcosSpessatto (changes requested):

apps/meteor/app/lib/server/functions/saveUser.js
felipe-rod123: Hello there! I think the frontend uses that to create the `ToastBar` warnings, but I'm not sure hahah
since this was the previous behavior, perhaps we can keep it like that, and change in another PR. What do you think?

apps/meteor/app/lib/server/lib/passwordPolicy.ts
felipe-rod123: Hi! Good idea!

apps/meteor/server/methods/saveUserProfile.ts
felipe-rod123: Ok, I'll remove this line!

packages/account-utils/src/PasswordPolicyClass.ts
felipe-rod123: Moved it to the constructor!

packages/account-utils/src/PasswordPolicyClass.ts
felipe-rod123: Hmm, I think those props can't be private, since we are updating them with the `settings.watch()` method on the backend... and the frontend uses the `useSetting()` hook to keep it synced with the backend instance.

packages/account-utils/src/PasswordPolicyClass.ts
```diff
@@ -27,32 +67,111 @@ class PasswordPolicy {
...
+			validationReturn.push({
+				name: 'get-password-policy-mustContainAtLeastOneUppercase',
+				isValid: this.regex.mustContainAtLeastOneUppercase.test(password as string),
+			});
+		}
+
+		if (this.mustContainAtLeastOneNumber) {
+			validationReturn.push({
+				name: 'get-password-policy-mustContainAtLeastOneNumber',
+				isValid: this.regex.mustContainAtLeastOneNumber.test(password as string),
```
MarcosSpessatto, line 155:
```suggestion
				isValid: this.regex.mustContainAtLeastOneNumber.test(password),
```

packages/account-utils/src/PasswordPolicyClass.ts
```diff
@@ -27,32 +67,111 @@ class PasswordPolicy {
...
+			});
+		}
+
+		if (this.mustContainAtLeastOneSpecialCharacter) {
+			validationReturn.push({
+				name: 'get-password-policy-mustContainAtLeastOneSpecialCharacter',
+				isValid: this.regex.mustContainAtLeastOneSpecialCharacter.test(password as string),
```
MarcosSpessatto, line 162:
```suggestion
				isValid: this.regex.mustContainAtLeastOneSpecialCharacter.test(password),
```

packages/account-utils/src/PasswordPolicyClass.ts
```diff
@@ -27,32 +67,111 @@ class PasswordPolicy {
...
+			});
+		}
+
+		return validationReturn;
+	}
+
+	validate(password: string | unknown) {
```
MarcosSpessatto, line 169: Any reason for this not to be a string only?

MarcosSpessatto (changes requested):

Issue #29855, opened 2023-07-18
Admin Account Creation


WhiteTiburon:
### Description:

I am attempting to create an Initial user via the compose file following **https://docs.rocket.chat/setup-and-configure/accessing-your-workspace/creating-the-first-administrator** and I get a 

=== UnHandledPromiseRejection ===
errorClass [Error]: Match error: Expected string, got object
    at check (packages/check/match.js:36:17)
    at packages/accounts-password/password_server.js:324:3

 It leads back to the password not being sent as a string. However, I made sure to make the password a string and still get the same error.

### Steps to reproduce:

1. add variables to compose file
2. run compose file via docker-compose up -d

### Expected behavior:

Creates user and I am able to log in.

### Actual behavior:
I get this error in the logs:

Inserting admin user:
Name: Administrator
Email: me@me.com
Username: Jon
=== UnHandledPromiseRejection ===
errorClass [Error]: Match error: Expected string, got object
    at check (packages/check/match.js:36:17)
    at packages/accounts-password/password_server.js:324:3
    at /app/bundle/programs/server/npm/node_modules/meteor/promise/node_modules/meteor-promise/fiber_pool.js:43:40
 => awaited here:
    at Function.Promise.await (/app/bundle/programs/server/npm/node_modules/meteor/promise/node_modules/meteor-promise/promise_server.js:56:12)
    at server/startup/initialData.js:122:4
    at /app/bundle/programs/server/npm/node_modules/meteor/promise/node_modules/meteor-promise/fiber_pool.js:43:40 {
  path: '',
  sanitizedError: errorClass [Error]: Match failed [400]
      at errorClass.<anonymous> (packages/check/match.js:91:27)
      at new errorClass (packages/meteor.js:692:17)
      at check (packages/check/match.js:36:17)
      at packages/accounts-password/password_server.js:324:3
      at /app/bundle/programs/server/npm/node_modules/meteor/promise/node_modules/meteor-promise/fiber_pool.js:43:40 {
    isClientSafe: true,
    error: 400,
    reason: 'Match failed',
    details: undefined,
    errorType: 'Meteor.Error'
  },
  errorType: 'Match.Error'
}

Here is what I added to the compose file:

ADMIN_USERNAME: test
ADMIN_PASS: "password"
ADMIN_EMAIL: me@me.com

[...]

PR #30015, opened 2023-08-04
fix: user creation using ADMIN_* variables

debdutdeb:
## Further comments

Closes : #29855


apps/meteor/server/startup/initialData.js
```diff
@@ -12,6 +12,7 @@ import { settings } from '../../app/settings/server';
 import { validateEmail } from '../../lib/emailValidator';
 import { addUserRolesAsync } from '../lib/roles/addUserRoles';
 
+
```
MarcosSpessatto, line 15: Can we remove this?

MarcosSpessatto (changes requested):

apps/meteor/server/startup/initialData.js
debdutdeb:
```suggestion
```

pierre-lehnen-rc (approved):

MarcosSpessatto (approved):

PR #30038, opened 2023-08-08
chore: codeBlock theme colors

juliajforesti:

- apply tokens on codeBlock component

- dark-theme
image

- light-theme 
image


ee/packages/ui-theming/src/codeBlockStyles.ts
```diff
@@ -0,0 +1,74 @@
+export const codeBlock = `pre code.hljs {
+	display: block;
+	overflow-x: auto;
+	padding: 1em;
+}
+code.hljs {
+	padding: 3px 5px;
+}
+.hljs {
+	background: #1d1f21 !important !important;
```
tassoevan, line 10:
```suggestion
	background: #1d1f21 !important;
```

ee/packages/ui-theming/src/codeBlockStyles.ts
```diff
@@ -0,0 +1,74 @@
...
+.hljs-template-variable,
+.hljs-variable {
+	color: #8abeb7 !important;
+}
+.hljs-built_in,
+.hljs-quote,
+.hljs-section,
+.hljs-selector-class,
+.hljs-type {
+	color: #de935f !important;
```
tassoevan, line 67: `question` Why so many `!important` declarations?

tassoevan (changes requested):

ee/packages/ui-theming/src/codeBlockStyles.ts
juliajforesti: `hljs` injects a css at end of the `head`.
it would be correctly overwritten if the new style was imported on `packages/gazzodown/src/code/CodeBlock.tsx`
but since we need to conditionally toggle between these 2 styles (light theme uses the default, dark theme uses this one), the way out was to "manually" inject it along with our dark palette.

our palette is injected **before** on the dom, that's why `!important` is needed :(

ggazzo (approved):

Issue #15378, opened 2019-09-16
Un-muting user not idempotent (resulting in unwanted state)


winterstefan:
#### TL;DR
**Muting & then un-muting one user** has the side-effect, that he/she will be then **able to send messages**, even if the **entire room** gets set to **read-only**.

The root cause is that whilst un-muting the user, the name is removed from `room.muted` **and added** to `room.unmuted`. That's not idempotent.

### Description:

The PR https://github.com/RocketChat/Rocket.Chat/pull/11311 fixed some issues with read-only & muted of a room. Unfortunately, this fix introduced a side-effect, that may not be intended.

_Let's assume the following use case_

Action | Effect | read-only | muted | unmuted
------------ | ------------- | ------------- | ------------- | -------------
Chat `x` has participant `random-dude` (it's newly created) | Everybody can send messages | `false` | `[]` | `[]`
`random-dude` sends spam & therefore gets muted | Everybody but the `random-dude` can send messages | `false` | `['random-dude']` | `[]`
After stopping spamming, `random-dude` gets re-activated | Everybody can send messages  | `false` | `[]` | `['random-dude']` ❗️
Now, the entire chat needs to be muted for some time -> moderator sets it to read-only |  Nobody except the `random-dude` ❗️ can send messages | `true` | `[]` | `['random-dude']`

In my humble opinion, the marked (❗️) states are quiet odd. That means, **once a user gets re-activated**, the user is able to send messages **even if the entire room gets muted.**
Even if the user gets kicked, its username remains in the `room.unmuted` array.

This is very error-prone, because the moderator doesn't see which users still are able to send messages.

### Steps to reproduce:

- Follow the mentioned use-case from above
- For example, you may ...
  - Log into the admin UI as administator
  - Create a blank new room with a participant
  - Open the e.g. Chrome developer tools & watch the WebSocket frames being send
  - Use the administrator to follow the use-case 

### Expected behavior:

Muting and un-muting one user is idempotent. So a blank new user shall have the same muted-state as a 'muted and then unmuted user' (no record in neither `muted` and `unmuted`).

### Actual behavior:

A 'muted and then unmuted user' still can send messages in a muted room.

### Server Setup Information:

- Version of Rocket.Chat Server: 1.3.2
- Operating System: Linux
- Deployment Method: docker
- Number of Running Instances: 3 
- DB Replicaset Oplog: true
- NodeJS Version: 8.11.4
- MongoDB Version: 4.0.12

[...]

PR #30050, opened 2023-08-09
fix: Unmuting users on average rooms allows them to speak if room is later set as read-only

heitortanoue:

Users can no longer send messages in a read only room if they are not unmuted by an admin. 
Unmuting an user doesn't add it to the `unmuted` array anymore, just removing it from the `muted` array

## Issue(s)
#15378 

## Steps to test or reproduce

Action | Effect | read-only | muted | unmuted
------------ | ------------- | ------------- | ------------- | -------------
Chat `x` has participant `random-dude` (it's newly created) | Everybody can send messages | `false` | `[]` | `[]`
`random-dude` sends spam & therefore gets muted | Everybody but the `random-dude` can send messages | `false` | `['random-dude']` | `[]`
After stopping spamming, `random-dude` gets re-activated | Everybody can send messages  | `false` | `[]` | `['random-dude']` ❗️
Now, the entire chat needs to be muted for some time -> moderator sets it to read-only |  Nobody except the `random-dude` ❗️ can send messages | `true` | `[]` | `['random-dude']`

For testing the unmuting on the read only rooms I recommend being in an admin account and using the slash command `/unmute <username>`.

## Further comments

TC-815


pierre-lehnen-rc (commented):
Can you apply the same change to the `muteUserInRoom` method?

Otherwise if a user loses the permission to speak on a readonly room, they'll remain muted if the room stops being readonly.

gabriellsh (dismissed):
LGTM

apps/meteor/server/models/raw/Rooms.ts
```diff
@@ -1707,7 +1707,31 @@ export class RoomsRaw extends BaseRaw<IRoom> implements IRoomsModel {
 		return this.updateOne(query, update);
 	}
 
-	unmuteUsernameByRoomId(_id: IRoom['_id'], username: IUser['username']): Promise<UpdateResult> {
+	muteReadOnlyMutedUsernameByRoomId(_id: IRoom['_id'], username: IUser['username']): Promise<UpdateResult> {
```
MarcosSpessatto, line 1710: Would it not be better to be `muteReadOnlyUsernameByRoomId`?

apps/meteor/server/models/raw/Rooms.ts
```diff
@@ -1707,7 +1707,31 @@ export class RoomsRaw extends BaseRaw<IRoom> implements IRoomsModel {
...
+		const query: Filter<IRoom> = { _id };
```
MarcosSpessatto, line 1711: Doesn't it make sense to query also by readonly here just to make sure and to make the method name make more sense? `const query: Filter<IRoom> = { _id, ro: true };`

apps/meteor/server/models/raw/Rooms.ts
```diff
@@ -1707,7 +1707,31 @@ export class RoomsRaw extends BaseRaw<IRoom> implements IRoomsModel {
...
+		const update: UpdateFilter<IRoom> = {
+			$pull: {
+				muted: username,
+			},
+		};
+
+		return this.updateOne(query, update);
+	}
+
+	unmuteReadOnlyUsernameByRoomId(_id: string, username: string): Promise<UpdateResult> {
```
MarcosSpessatto, line 1734: I think we could also add the `{ to: true }` flag here as well, just to reflect the method name.

MarcosSpessatto (changes requested):

matheusbsilva137 (approved):

MarcosSpessatto (approved):
It'd be nice if we could add a test to cover this scenario, We can add that in a subsequent PR.

KevLehman (approved):

PR #30055, opened 2023-08-09
fix: remove code that was setting mail_url to 'undefined'

tapiarafael:

Remove code that was setting the env `MAIL_URL` to `'undefined'`. 
This causes an error since the lib `meteor/email` validates if the host is set before sending the email. By setting it as `'undefined'` instead of `undefined`, the lib tried to send the email to the host 'undefined' resulting in the error: 
`Error trying to send email: Invalid URL: undefined [error-email-send-failed]`.

## Steps to test or reproduce
1. Go to workspace > users > edit user;
2. Change the email to anything valid;
3. Save the changes;

## Further comments

[WM-151](https://rocketchat.atlassian.net/browse/WM-151)

[WM-151]: https://rocketchat.atlassian.net/browse/WM-151?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


pierre-lehnen-rc (changes requested):
Good catch!
But in order to ensure the MAIL_URL gets cleared when the setting is removed, use this instead of the old assignment.

```js
delete process.env.MAIL_URL;
```

pierre-lehnen-rc (approved):

sampaiodiego (approved):

sampaiodiego (changes requested):
can you please add a changeset for the fix? thx

sampaiodiego (dismissed):

sampaiodiego (approved):

PR #30063, opened 2023-08-10
fix: recurring ENOENT errors when processing export requests

debdutdeb:


apps/meteor/server/methods/requestDataDownload.ts
debdutdeb, line 1: creating these here is pointless if from when export is requested to when it gets processed, the container restarts and a persistent path is not set in setting and in deployment.

pierre-lehnen-rc (approved):

PR #30069, opened 2023-08-10
fix(meteor): Respect `UI_Use_Real_Name` setting for reactions

yash-rajpal:
## Further comments

[TC-867]

[TC-867]: https://rocketchat.atlassian.net/browse/TC-867?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Shivamdwivedi1403 (approved):

tassoevan (dismissed):

packages/rest-typings/src/v1/users/UsersGetNamesParamsGET.ts
```diff
@@ -0,0 +1,40 @@
+import Ajv from 'ajv';
+
+const ajv = new Ajv({
+	coerceTypes: true,
+});
+
+export type UsersGetNamesParamsGET = { userIds: string[] } | { usernames: string[] };
+
+const UsersGetNamesParamsGETSchema = {
+	anyOf: [
```
matheusbsilva137, line 10:
```suggestion
	oneOf: [
```
`anyOf` allows matching with multiple schemas. In this case, there is no interesection (due to `additionalProperties` set to `false`), so we'd end up always using only one of them. Still, I think we should use `oneOf` to state clearer what is the intention here.

apps/meteor/app/api/server/v1/users.ts
```diff
@@ -1255,6 +1256,25 @@ API.v1.addRoute(
 	},
 );
 
+API.v1.addRoute(
+	'users.getNames',
+	{ authRequired: true, validateParams: isUsersGetNamesParamsGETProps },
+	{
+		async get() {
+			const users = await Users.findByUsernames(this.queryParams.usernames, {
```
matheusbsilva137, line 1264: We're not covering the case of `usernames` not being sent as a param here. It could be `undefined` when `userIds` is provided instead.

packages/rest-typings/src/v1/users.ts
```diff
@@ -367,6 +368,16 @@ export type UsersEndpoints = {
...
 		POST: (params: { password: string; confirmRelinquish?: boolean }) => void;
 	};
+
+	'/v1/users.getNames': {
+		GET: (params: UsersGetNamesParamsGET) => {
+			users: {
+				userId: string;
+				username: string;
+				name: string;
+			}[];
```
matheusbsilva137, line 378:
```suggestion
			users: Pick<IUser, '_id', 'username', 'name'>[];
```

matheusbsilva137 (changes requested):

packages/rest-typings/src/v1/users.ts
debdutdeb: @matheusbsilva137 the fields are different tho 😬 also, i think, in IUser username is nullable, so is name, which again isn't what the endpoint is returning (the nullability isn't true, it is nullable for the sake of reusing the types in many places, does not signify username or name won't be in the resulting record (name prob for rocket.cat idk))

apps/meteor/client/components/message/content/reactions/Reaction.tsx
```diff
@@ -41,29 +43,53 @@ const Reaction = ({ hasReacted, counter, name, names, ...props }: ReactionProps)
...
+	const { refetch } = useQuery(
+		['users.getNames', names],
+		async () => {
+			if (names.length === 0) {
+				return undefined;
+			}
+
+			const users: string[] = showRealName
+				? (await getNames({ usernames: names }))?.users?.map((user) => user.name) || []
+				: names.map((name) => `@${name}`);
```
debdutdeb, line 65: about the `|| []` here, if there is a situation where the chain may break and result in undefined or something (as the typing says here), wdyt about falling back to showing usernames and not nothing?

apps/meteor/app/api/server/v1/users.ts
debdutdeb: @matheusbsilva137 isn't that handled by ajv )isUsersGetNamesParamsGETProps) ?

yash-rajpal: Nice catch, I completely overlooked that part 🙈

apps/meteor/client/components/message/content/reactions/Reaction.tsx
yash-rajpal: Makes sense, it shouldn't happen but usernames is better than nothing.

gabriellsh (approved):

tassoevan (dismissed):

apps/meteor/app/api/server/v1/users.ts
```diff
@@ -1255,6 +1256,38 @@ API.v1.addRoute(
...
+					projection: {
+						name: 1,
+						username: 1,
+					},
+				}).toArray();
+			}
+
+			return API.v1.success({
+				users,
+			});
```
matheusbsilva137, line 1286:
```suggestion
			const options = {
				projection: {
					name: 1,
					username: 1,
				},
			};

			if ('userIds' in this.queryParams) {
				return API.v1.success({
					users: await Users.findByIds(this.queryParams.userIds, options).toArray(),
				});
			}

			return API.v1.success({
				users: await Users.findByUsernames(this.queryParams.usernames, options).toArray(),
			});
```
Looks fine, but we usually avoid using let

matheusbsilva137 (changes requested):

apps/meteor/app/api/server/v1/users.ts
```diff
@@ -1255,6 +1256,38 @@ API.v1.addRoute(
 	},
 );
 
+API.v1.addRoute(
+	'users.getNames',
```
sampaiodiego, line 1260: I'm not totally against this, but I was wondering if we shouldn't be able to use `/api/v1/users.list` for this purpose.. 

or maybe, even better, we could have an endpoint within the "messages" namespace to get the reactions list from it.. this way we don't need to provide a list of users to get some info, we'd provide just the message id.. wdyt?

yash-rajpal: `users.list` is a paginated API, also doesn't return for a set of particular users like we need in this case.

yash-rajpal: I like the idea of having such implementation under `messages` namespace and using the `messageId` to get normalized reactions list, but it would have been more expensive to fetch the message reactions using messageId and then querying for user's details, what do you think about this, as this API will be called a lot of times?

matheusbsilva137 (approved):

apps/meteor/app/api/server/v1/users.ts
matheusbsilva137: Yep, also looks like a good idea to me to add a new `messages` endpoint for this at some point
Since we must consider heavy usage for this now, it might be better to keep this new endpoint as it is and avoid this extra db trip :thinking:

sampaiodiego: having this endpoint being called multiple times with a single (and constant) parameter is a ideal scenario for caching.. in the other hand, if we'd like to add cache to `users.getNames` endpoint it would not be as efficient since it would receive a lot of different combination of params..

btw, the payload query for `users.getNames` doesn't scale well.. it is kinda rare having too many reactions in a single message, but there are no limits still.. so you may need to provide a list of hundreds of user IDs as the payload.. not ideal too..

other reasons I'm also not a fan of the `users.getNames` is that we're creating another very specific endpoint.. that's the point when I mentioned the `users.list` endpoint, it _should_ be able to provide this data, but unfortunately it doesn't =/ in this same principle, maybe we could just improve/expand the existing `chat.getMessage` endpoint to return this data, maybe..

yash-rajpal: As suggested, now using `chat.getMessage` API to get normalized reactions.

PR #30084, opened 2023-08-15
fix: Unread mark not showing

gabriellsh:

Unread Indicator behavior -

- Entering a unread room should show a red indicator above the first unread message
- If the user sends a message, we hide the indicator.
- If the first unread message is visible, we mark the room as read, but keep the indicator until interaction.
- If user gets out of the current window (alt+tab or click other window) and the room had been marked as read already, we update the indicator to be above the new messages.

## Issue(s)

[TC-758](https://rocketchat.atlassian.net/browse/TC-758)
## Further comments

[TC-758]: https://rocketchat.atlassian.net/browse/TC-758?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


tassoevan (approved):

tassoevan (dismissed):

tassoevan (changes requested):

apps/meteor/client/lib/chats/readStateManager.ts
```diff
@@ -0,0 +1,174 @@
...
+	public debouncedMarkAsRead = withDebouncing({ wait: 1000 })(() => {
+		try {
+			return this.markAsRead();
+		} catch (e) {
+			console.error(e);
+		}
+	});
+
+	// this will always mark as read.
+	public markAsRead() {
```
tassoevan, line 165: `question` Why not async?

apps/meteor/client/lib/chats/readStateManager.ts
```diff
@@ -0,0 +1,174 @@
...
+		}
+
+		return this.markAsRead();
+	}
+
+	public debouncedMarkAsRead = withDebouncing({ wait: 1000 })(() => {
+		try {
+			return this.markAsRead();
+		} catch (e) {
+			console.error(e);
```
tassoevan, line 160: `question` Why catch this error just to log it?

apps/meteor/client/views/room/body/hooks/useReadMessageWindowEvents.ts
```diff
@@ -1,42 +1,12 @@
...
-
-		return () => {
-			window.removeEventListener('blur', handleWindowBlur);
-			window.removeEventListener('focus', handleWindowFocus);
-			window.removeEventListener('touchend', handleWindowTouchEnd);
-			window.removeEventListener('keyup', handleWindowKeyUp);
-		};
-	}, [rid]);
+		const unsub = chat?.readStateManager.handleWindowEvents();
+		return unsub;
```
tassoevan, line 10:
```suggestion
		return chat?.readStateManager.handleWindowEvents();
```

tassoevan (changes requested):

apps/meteor/client/lib/chats/readStateManager.ts
gabriellsh: The `subscriptions.read` api call is very strict about sending too many requests. I added a debounce to it, but it can still trigger the delay. If `subscriptions.read` endpoint is blocked and I send a message, it blocks the message from being sent until the endpoint is active again (takes about 30s or longer). This way, if it happens to trigger the rate limiter, I can still send messages without issues.

Also, this is very similar to the current behaviour, since a bunch of too-many-requests errors pop in the console, but I can send messages normally.

tassoevan (approved):

PR #30088, opened 2023-08-15
feat: apply group filter to LDAP sync process

tapiarafael:
Use the group filter to filter the LDAP integration and sync only the users that match the criteria.

Today we sync all the users from the Base DN and use the group filter only to restrict the ones that can access the RC. The goal is to only sync the users in the group filter setting to avoid importing unwanted users and consuming unnecessary seats.

## Steps to test or reproduce
1. Create a `groupOfNames` in the LDAP;
2. Use the `cn` attribute to define the group name;
3. Set in the `member` attribute the user that you want to have access to the RC (create multiple `member` attributes if you need multiple users);
4. In Rocket.chat, navigate to workspace > settings > LDAP > User Search;
5. Enable the Group Filter and fill the fields;
6. Sync the data and check the users page;
7. Try to login with a user that is in the RC database, but not in the filtered group;

## Further comments

[SUP-305](https://rocketchat.atlassian.net/browse/SUP-305)

[SUP-305]: https://rocketchat.atlassian.net/browse/SUP-305?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/meteor/ee/server/lib/ldap/Manager.ts
```diff
@@ -43,7 +46,15 @@ export class LDAPEEManager extends LDAPManager {
...
 			}
 
+			const membersOfGroupFilter = await ldap.searchMembersOfGroupFilter();
+
 			await converter.convertUsers({
+				beforeImportFn: (async (data: IImportUser): Promise<boolean> => {
+					if (!ldap.options.groupFilterEnabled) {
+						return true;
+					}
+					return membersOfGroupFilter.includes(data.importIds[0]);
```
pierre-lehnen-rc, line 56: You need to use the value of the `LDAP_Group_Filter_Group_Member_Format` setting here.

apps/meteor/ee/server/lib/ldap/Manager.ts
```diff
@@ -43,9 +46,22 @@ export class LDAPEEManager extends LDAPManager {
...
-				afterImportFn: (async (data: IImportUser, _type: string, isNewRecord: boolean): Promise<void> =>
-					this.advancedSync(ldap, data, converter, isNewRecord)) as ImporterAfterImportCallback,
+				beforeImportFn: (async ({ options }: IImportRecord): Promise<boolean> => {
+					if (!ldap.options.groupFilterEnabled || !ldap.options.groupFilterGroupMemberFormat) {
+						return true;
+					}
+
+					const memberFormat = ldap.options.groupFilterGroupMemberFormat
+						?.replace(/#{username}/g, options?.username as string)
+						.replace(/#{userdn}/g, options?.dn as string);
```
pierre-lehnen-rc, line 59: The `username` may be undefined here. `dn` will never be with the current implementation, but it's worth validating it to be safe.

pierre-lehnen-rc (changes requested):

apps/meteor/ee/server/lib/ldap/Manager.ts
```diff
@@ -564,7 +580,7 @@ export class LDAPEEManager extends LDAPManager {
 
 			if (ldapUser) {
 				const userData = this.mapUserData(ldapUser, user.username);
-				converter.addUserSync(userData);
+				converter.addUserSync(userData, { dn: ldapUser.dn, username: this.getLdapUsername(data) });
```
pierre-lehnen-rc, line 583:
```suggestion
				converter.addUserSync(userData, { dn: ldapUser.dn, username: this.getLdapUsername(ldapUser) });
```

my bad.

apps/meteor/server/lib/ldap/Connection.ts
```diff
@@ -391,6 +391,49 @@ export class LDAPConnection {
...
+		const result = await this.searchRaw(this.options.baseDN, searchOptions);
+
+		if (!Array.isArray(result) || result.length === 0) {
+			searchLogger.debug({ msg: 'No groups found', result });
+			return [];
+		}
+
+		const membersBuffer: Buffer[] = result[0].raw[this.options.groupFilterGroupMemberAttribute] as Buffer[];
+
+		return membersBuffer.map((member) => member.toString());
```
pierre-lehnen-rc, line 434:
```suggestion
		return this.extractLdapAttribute(result[0].raw[this.options.groupFilterGroupMemberAttribute]);
```

apps/meteor/server/lib/ldap/Connection.ts
```diff
@@ -391,6 +392,49 @@ export class LDAPConnection {
...
+
+		const result = await this.searchRaw(this.options.baseDN, searchOptions);
+
+		if (!Array.isArray(result) || result.length === 0) {
+			searchLogger.debug({ msg: 'No groups found', result });
+			return [];
+		}
+
+		const members = this.extractLdapAttribute(result[0].raw[this.options.groupFilterGroupMemberAttribute]) as string[];
+
```
pierre-lehnen-rc, line 434:
```suggestion
		const members = this.extractLdapAttribute(result[0].raw[this.options.groupFilterGroupMemberAttribute]) as string | string[];

```

pierre-lehnen-rc (approved):

tassoevan (approved):

PR #30092, opened 2023-08-15
chore: move logger to a separate package

pierre-lehnen-rc:


apps/meteor/server/lib/logger/startup.ts
```diff
@@ -1,7 +1,6 @@
+import { setQueueLimit, logLevel, type LogLevelSetting } from '@rocket.chat/logger';
+
 import { settings } from '../../../app/settings/server';
-import type { LogLevelSetting } from './logLevel';
-import { logLevel } from './logLevel';
-import { setQueueLimit } from './logQueue';
```
sampaiodiego, line 4: I wonder if the "log queue" feature should be kept on meteor app.. it is probably not relevant to other packages/apps using the logger

pierre-lehnen-rc: I moved it there to ensure it runs before the main pino instance is instantiated - so Pino can detect that `process.stdout.write` has been modified.

packages/logger/package.json
```diff
@@ -0,0 +1,28 @@
+{
+	"name": "@rocket.chat/logger",
+	"version": "0.0.1",
+	"private": true,
+	"devDependencies": {
+		"@types/jest": "^27.4.1",
+		"eslint": "~8.45.0",
+		"jest": "~29.5.0",
+		"ts-jest": "~29.0.5",
+		"typescript": "~5.0.2"
```
sampaiodiego, line 10: we're using TS 5.1 elsewhere.. can you pls update to 5.1 and also check if the other versions are also matching?

packages/logger/package.json
```diff
@@ -0,0 +1,28 @@
...
+		"dev": "tsc -p tsconfig.json --watch --preserveWatchOutput"
+	},
+	"main": "./dist/index.js",
+	"typings": "./dist/index.d.ts",
+	"files": [
+		"/dist"
+	],
+	"dependencies": {
+		"@rocket.chat/emitter": "next",
+		"pino": "^8.14.1"
```
sampaiodiego, line 26: maybe a good opportunity to update to pino 8.15?

sampaiodiego (changes requested):

sampaiodiego (dismissed):

sampaiodiego (approved):

PR #30108, opened 2023-08-17
chore: convert push notifications code to typescript

pierre-lehnen-rc:


apps/meteor/app/push/server/apn.ts
```diff
@@ -1,22 +1,46 @@
...
 
-let apnConnection;
-
-export const sendAPN = ({ userToken, notification, _removeToken }) => {
-	if (typeof notification.apn === 'object') {
-		notification = Object.assign({}, notification, notification.apn);
+let apnConnection: apn.Provider | undefined;
+
+// the definition included in the module does not include the set-only attributes
+class ApnNotification extends apn.Notification {
```
sampaiodiego, line 11: instead of declaring a new class that needs to be used with a type cast later on, I'd actually recommend augmenting the "official" one:
```ts
declare module 'apn' {
	// eslint-disable-next-line @typescript-eslint/naming-convention
	interface Notification {
		setContentAvailable: (value: boolean | 1 | 0) => void;
		category: string | undefined;
		body: string;
		title: string;
	}
}
```

apps/meteor/app/push/server/apn.ts
```diff
@@ -1,22 +1,46 @@
...
+	_removeToken: (token: IAppsTokens['token']) => void;
+}) => {
+	if (!apnConnection) {
+		throw new Error('Apn Connection not initialized.');
 	}
 
 	const priority = notification.priority || notification.priority === 0 ? notification.priority : 10;
 
-	const note = new apn.Notification();
+	const note = new apn.Notification() as ApnNotification;
```
sampaiodiego, line 39:
```suggestion
	const note = new apn.Notification();
```

apps/meteor/app/push/server/apn.ts
```diff
@@ -1,22 +1,46 @@
...
 
 	note.expiry = Math.floor(Date.now() / 1000) + 3600; // Expires 1 hour from now.
-	note.badge = notification.badge;
-	note.sound = notification.sound;
+	note.badge = notification.badge as number; // the module accepts undefined, but is typed wrong
+	note.sound = notification.sound as string; // the module accepts undefined, but is typed wrong
```
sampaiodiego, line 43: even if the type is incorrect, we can make it work:
```suggestion
	if (notification.badge) {
		note.badge = notification.badge;
	}
	if (notification.sound) {
		note.sound = notification.sound;
	}
```

sampaiodiego (changes requested):

apps/meteor/app/push/server/apn.ts
pierre-lehnen-rc: I had done this at first, but ultimately decided to go with the typecast instead, to ensure that undefined values will continue to be set to the attribute.

apps/meteor/app/push/server/apn.ts
pierre-lehnen-rc: Benefits of taking so long to open a PR, I guess. 
When I first wrote this code this augmentation wouldn't work because those attributes only have setters. Typescript 5.1 added support for different types on setters/getters and now this works.

sampaiodiego (approved):

Issue #30049, opened 2023-08-09
[6.3.0] SAML login breaks in mobile app


svenseeberg:
### Description:

When logging in with the mobile app via a SAML IdP, the app logs in but stays in the SSO browser widget. The app itself does not register a successful login. Instead, the Rocket.Chat website is shown in the browser widget.

### Steps to reproduce:

1. Configure SAML IdP
2. Log in via native app
3. App stays in SSO widget, even after successful login

### Expected behavior:

The app should register the successful login and return to its native chat options

### Actual behavior:

![image](https://github.com/RocketChat/Rocket.Chat/assets/15197000/bcb8ee28-9e6d-4e32-8a57-a2f5c3e87d20)

### Server Setup Information:

- Version of Rocket.Chat Server: 6.3.0
- Operating System: Debian 11
- Deployment Method: tar
- Number of Running Instances: 1
- DB Replicaset Oplog: 1
- NodeJS Version: v14.21.3
- MongoDB Version: 5.0.19

### Client Setup Information

- Desktop App or Browser Version: Android 4.39.0.43604
- Operating System: Android 13

### Additional context

N/A

### Relevant logs:

The app sends a request to `/api/v1/login` with a JSON body:
```json
{
"saml": true,
"credentialToken": "MySecret"
}
```
The server answers with HTTP 401 Unauthorized and the following message:
```json
{
"status": "error",
"message": "No matching login attempt found",
"error": 12345678
}
```
The message obviously comes from https://github.com/RocketChat/Rocket.Chat/blob/856c23555a43853165ba9d9daeed508ccf4c6892/apps/meteor/app/meteor-accounts-saml/server/loginHandler.ts#L23

arthur-trt:
Facing the same problem here with Android and iOS

anicoa:
In CredentialTokens.create the token is written into collection rocketchat_credential_tokens into field inResponseToId:

```
{ "_id": "id","userInfo":{"profile":{"inResponseToId":"JBvsCVkOeEPRH1v8N", 
```

But CredentialTokens.findOneNotExpiredById is looking for the token in _id:

```
"filter":{"_id":"JBvsCVkOeEPRH1v8N","expireAt":{"$gt":{"$date":"2023-08-10T14:02:30.620Z"}
```

arthur-trt:
Issue still present in 6.3.1

HunKjj:
Yes 6.3.1 same issue

[...]

PR #30121, opened 2023-08-19
fix: broken SAML login on react native apps

anicoa:
## Proposed changes 

changes `findOneNotExpiredById` in `./apps/meteor/server/models/raw/CredentialTokens.ts` to look for the token in `userInfo.profile.inResponseToId` instead using the `_id`

## Issue(s)

Solves https://github.com/RocketChat/Rocket.Chat/issues/30049

## Steps to test or reproduce

- Open react native app
- authenticate with saml


pierre-lehnen-rc (changes requested):
The query itself is not wrong, the error is somewhere else.
I'll investigate.

pierre-lehnen-rc:
The issue is actually on the url we redirect the user to after a succesful login. It changed on 6.3 but this change was not reflected on our mobile app. We'll fix it there.

PR #30123, opened 2023-08-19
feat: Hide chat console during verification process in Omnichannel rooms

bhardwajdisha:
During the omnichannel verification process, the chat console will be concealed, and in its place, a message footer will be presented. This footer will include a "Cancel Verification" button.

https://github.com/RocketChat/Rocket.Chat/assets/67470541/e44efbe7-2c70-424c-b580-47f499a05d00

## Steps to test or reproduce
1. Go to omnichannel, join an open room
2. Click on verify user button from quick action buttons.

### DIFF
--- a/apps/meteor/app/livechat/server/api/v1/room.ts
+++ b/apps/meteor/app/livechat/server/api/v1/room.ts
@@ -458,21 +458,14 @@ API.v1.addRoute(
 	{ authRequired: true, permissionsRequired: ['initiate-livechat-verification-process'] },
 	{
 		async put() {
-			// const { roomId } = this.queryParams;
 			const { roomId } = this.bodyParams;
 			if (!roomId) {
 				throw new Error('invalid-room');
 			}
-			// const room = await LivechatRooms.findOneById(roomId, { _id: 1 });
-			// if (!room) {
-			// 	throw new Error('invalid-room');
-			// }
-			console.log('hi', roomId);
 			const roomAfterVerificationStateUpdate = await LivechatRooms.updateVerificationStatusById(roomId, RoomVerificationState.unVerified);
-			// if (!roomAfterVerificationStateUpdate) {
-			// 	return API.v1.failure();
-			// }
-			console.log('hi', roomAfterVerificationStateUpdate);
+			if (!roomAfterVerificationStateUpdate) {
+				return API.v1.failure();
+			}
 			return API.v1.success();
 		},
 	},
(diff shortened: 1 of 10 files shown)


apps/meteor/client/views/room/composer/ComposerOmnichannel/ComposerOmnichannelVerification.tsx
```diff
@@ -0,0 +1,28 @@
...
+		try {
+			await cancelVerification({ roomId: room._id });
+		} catch (error) {
+			dispatchToastMessage({ type: 'error', message: error });
+		}
+	};
+	return (
+		<MessageFooterCallout>
+			<MessageFooterCalloutContent>{t('This_conversation_is_currently_undergoing_a_verification_process')}</MessageFooterCalloutContent>
+			<MessageFooterCalloutAction onClick={handleCancelVerification}>{t('Cancel')}</MessageFooterCalloutAction>
```
murtaza98, line 25: Is it possible to make this a red considering it falls under danger action?

apps/meteor/client/views/room/composer/ComposerOmnichannel/ComposerOmnichannelVerification.tsx
```diff
@@ -0,0 +1,28 @@
+import { MessageFooterCallout, MessageFooterCalloutAction, MessageFooterCalloutContent } from '@rocket.chat/ui-composer';
+import { useEndpoint, useToastMessageDispatch, useTranslation } from '@rocket.chat/ui-contexts';
+// useToastMessageDispatch,
```
murtaza98, line 3:
```suggestion
```

apps/meteor/app/livechat/server/api/v1/room.ts
```diff
@@ -453,3 +454,35 @@ API.v1.addRoute(
 		},
 	},
 );
+API.v1.addRoute(
```
murtaza98, line 457: Instead of spreading all the endpoints created for this verification feature across multiple files, how about we group all of them together and put them in a single file?

packages/ui-composer/src/MessageFooterCallout/MessageFooterCalloutAction.tsx
```diff
@@ -6,7 +6,8 @@ const MessageFooterCalloutAction = forwardRef<HTMLButtonElement, ComponentProps<
 	props,
 	ref,
 ): ReactElement {
-	return <Button mi={4} ref={ref} primary small flexShrink={0} {...props} />;
+	const modifiedProps = props.danger ? { ...props, primary: false } : props;
+	return <Button mi={4} ref={ref} primary small flexShrink={0} {...modifiedProps} />;
```
murtaza98, line 10: Could you check if we really need to do this override within this generic component? My understanding is that since `props.danger` is added after primary, it'd take precedence over primary and the button would be displayed as red.

murtaza98: Just tested this locally with the old code and just "danger" prop set and found it working as expected

bhardwajdisha: It didn’t worked out for me so that’s why I tried this one.

murtaza98: That's odd! Let's debug this on our call today

murtaza98 (approved):

Issue #30111, opened 2023-08-17
Regression: Since V6.3.1 http.post in apps, timeout seems to be shorter than 15 sec. 


FrankDaze:
### Description:

We run our own ChatGPT app with a special terms of use agreement process. In V 6.2.6 everything was fine until we updated to V 6.3.1 

Now the behavior is: I ask for "which color has the moon" (quick answer) works.
I ask "please create a trailwind table with 34 columns and the first 2 are sticky" and after appox. 13 seconds I receive an undefined from the http.post as answer. 

All short questions work. Questions with longer answers run into timeout. I tried as well to add timeout:  60000 to the http.post options but no change.

### Steps to reproduce:

I think you can reproduce the behavior by creating an app this a http.post where response takes longer than 15 seconds. 

### Expected behavior:

I get an answer for whatever I ask

### Actual behavior:

There is no error message to screenshot. http.post returns only "undefined "

### Server Setup Information:

Operating System: OS des Docker-Images (Debian GNU/Linux 11 (bullseye))
Deployment Method: Helm via official Chart and ArgoCD (Seperate MongoDB)
Number of Running Instances: rocketchat - 1 / mongodb-replicaset: 3 as stateful set /mongodb-arbiter: 1 as stateful set)
MongoDB-Version: 6.0.8-debian-11-r12 (bitnami/helm)
NodeJS-Version: Version of Docker image (v14.21.3)

### Client Setup Information

- Desktop App or Browser Version: App and Browser same behavior
- Operating System: Mac, Windows, Linux

casalsgh:
We'll investigate with internal teams and get back

FrankDaze:
I think I should add the coding part of my app to be sure, I made no mistake:

`
const ok = await http.post('https://api.openai.com/v1/chat/completions', {
                headers: {
                    'Content-Type': 'application/json',
                    'Accept': 'application/json',
                    'Authorization': `Bearer ${apiKey}`
                },
                content: JSON.stringify(content)
         });

        if(typeof(ok)==="undefined")
        {

            this.sendMessage(sender, room, read, modify, { text: "Got no answer from ChatGPT" }, "ChatGPT");
            return;
        }
`

florisvangeel:
in what file did you add this?
having issues with all 6.3.x perhaps it helps :)

FrankDaze:
Hi Floris, 

this is in the main file of my app.

PR #30129, opened 2023-08-21
fix: HTTP timeout too short on Apps-Engine bridge

d-gubert:

Previously there was no timeout configured for HTTP requests made by apps via the Apps-Engine bridge, however the underlying library that handled this was changed and introduced a timeout of 20s - breaking compatibility to some extent. 

Here we set the timeout to 3m as it seems to be an appropriate balance.

## Issue(s)

https://github.com/RocketChat/Rocket.Chat/issues/30111

### DIFF
--- a/apps/meteor/app/apps/server/bridges/http.ts
+++ b/apps/meteor/app/apps/server/bridges/http.ts
@@ -7,6 +7,11 @@ import type { AppServerOrchestrator } from '../../../../ee/server/apps/orchestra
 
 const isGetOrHead = (method: string): boolean => ['GET', 'HEAD'].includes(method.toUpperCase());
 
+// Previously, there was no timeout for HTTP requests.
+// We're setting the default timeout now to 3 minutes as it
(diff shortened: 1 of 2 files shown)


apps/meteor/app/apps/server/bridges/http.ts
```diff
@@ -7,6 +7,11 @@ import type { AppServerOrchestrator } from '../../../../ee/server/apps/orchestra
 
 const isGetOrHead = (method: string): boolean => ['GET', 'HEAD'].includes(method.toUpperCase());
 
+// Previously, there was no timeout for HTTP requests.
+// We're setting the default timeout now to 3 minutes as it
+// seems to be a good balance
+const DEFAULT_TIMEOUT = 3 * 60 * 1000;
```
d-gubert, line 13: Not completely sure about the time here - if anyone wants to chime in, please do :)

Dnouv (approved):
LGTM!

PR #30169, opened 2023-08-24
chore: Add missing index on Permissions collection

KevLehman:


sampaiodiego (commented):
what about moving the queries to the model as well? 😬

sampaiodiego (approved):

PR #30173, opened 2023-08-24
chore: don't allow to report own messages and user actions menu dividers

yash-rajpal:
## Further comments

TC-929


apps/meteor/app/ui-utils/client/lib/messageActionDefault.ts
```diff
@@ -241,9 +241,9 @@ Meteor.startup(async () => {
 				},
 			});
 		},
-		condition({ subscription, room }) {
+		condition({ subscription, room, message }) {
```
gabriellsh, line 244: I think you can get the user from this object, instead of using `Meteor.userId()`

gabriellsh (changes requested):

apps/meteor/app/ui-utils/client/lib/messageActionDefault.ts
yash-rajpal: Done!

gabriellsh (approved):

PR #30175, opened 2023-08-25
chore: move team mentions to ce

heitortanoue:
Move the Team Mentions feature to the RC Community Edition

## Steps to test or reproduce

Without the EE license, try to use the team mention feature by sending a message with `@team`, all team members should be notified.

## Further comments

TC-901


yash-rajpal (dismissed):
LGTM

apps/meteor/app/mentions/server/getMentionedTeamMembers.ts
```diff
@@ -0,0 +1,27 @@
...
+callbacks.add('beforeGetMentions', async (mentionIds: string[], extra?: IExtraDataForNotification) => {
+	const { otherMentions } = extra ?? {};
+
+	const teamIds = otherMentions?.filter(({ type }) => type === 'team').map(({ _id }) => _id);
+
+	if (!teamIds?.length) {
+		return mentionIds;
+	}
+
+	const members: ITeamMember[] = await Team.getMembersByTeamIds(teamIds, { projection: { userId: 1 } });
```
KevLehman, line 21: Don't you get the type from getMembersByTeamIds?

apps/meteor/app/mentions/server/getMentionedTeamMembers.ts
```diff
@@ -0,0 +1,27 @@
...
+	mentionIds.push(
+		...new Set(members.map(({ userId }: { userId: string }) => userId).filter((userId: string) => !mentionIds.includes(userId))),
```
KevLehman, line 23: `members` already has a type, so it should be inferred here 👀

KevLehman (changes requested):

matheusbsilva137 (changes requested):
I guess we should also remove `'teams-mention'` from `BundleFeature` and from the enterprise bundle (in `bundles.ts`), shouldn't we? :thinking:

matheusbsilva137 (approved):

PR #30192, opened 2023-08-28
fix: User timezone not being respected on Current Chat's filter

murtaza98:
## Issue(s)

https://rocketchat.atlassian.net/browse/OC-822


murtaza98:
Closing this in favour of the new approach we're taking here. The new approach involves updating the client to convert dates to UTC format and then pass it to backend; opposite of what we were doing where which involved backend converting the dates to UTC

PR #30194, opened 2023-08-28
fix: `default` field is not returned from the `setDefault` endpoint when setting to false

matheusbsilva137:

 - Do not remove the `default` field from rooms when it's set to `false` (since this would cause a breaking change in the `.setDefault` endpoints).

## Issue(s)

Introduced in #28260

## Steps to test or reproduce

Use the `channels.setDefault` endpoint to make a room default in the workspace. Then use it again to _unset_ it as default.
**Current behavior:** the `default` field is not returned in the response when setting it to `false`.
**Expected behavior:** the `default` field should be returned (with `false` as its value)

## Further comments

TC-930

It'd be a nice change not to store the default field with `false` value, though (since we never check for a `false` value in this field). Let's keep this a to-do for the next major release.

### DIFF
--- a/apps/meteor/server/models/raw/Rooms.ts
+++ b/apps/meteor/server/models/raw/Rooms.ts
@@ -1348,7 +1348,7 @@ export class RoomsRaw extends BaseRaw<IRoom> implements IRoomsModel {
 			t: {
 				$in: types,
 			},
-			...(defaultValue ? { default: true } : { default: { $exists: false } }),
+			...(defaultValue ? { default: true } : { default: { $ne: true } }),
 		};
 
 		return this.find(query, options);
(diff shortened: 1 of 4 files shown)


apps/meteor/server/models/raw/Rooms.ts
```diff
@@ -1348,7 +1348,7 @@ export class RoomsRaw extends BaseRaw<IRoom> implements IRoomsModel {
 			t: {
 				$in: types,
 			},
-			...(defaultValue ? { default: true } : { default: { $exists: false } }),
```
KevLehman, line 1351: If `default` is always there, maybe the type should not be optional? 👀

matheusbsilva137: Actually, it should only be there once the room is set as default. After that, setting it to false should not erase the field so as to not break the `setDefault` endpoints' returns.
Just creating a non-default room won't create this field too, it's not always there.

MarcosSpessatto (approved):

PR #30201, opened 2023-08-28
regression: sidebar actions pressed state reactivity

yash-rajpal:
## Steps to test or reproduce

 - Go to home by pressing the Home in sidebar actions
 - Now go to some room 
 - Home sidebar action state still seems active

## Further comments

TC-936


dougfabris (dismissed):
LGTM

dougfabris (approved):

PR #30202, opened 2023-08-28
fix: new `overrideDestinationChannelEnabled` param in integrations' endpoints is required

matheusbsilva137:

 - Make `overrideDestinationChannelEnabled` param optional in `integrations.create` and `integration.update` endpoints (in order to avoid breaking changes).
 - Remove `overrideDestinationChannelEnabled` as a valid param when updating outgoing integrations (this only makes sense for ingoing integrations).

## Issue(s)

Introduced in #26875

## Steps to test or reproduce

Check the integrations.create and integrations.update endpoints. The new `overrideDestinationChannelEnabled` param should not be required anymore, and its value should be set to false when not provided in the creation.

## Further comments

TC-935


gabriellsh (dismissed):

packages/rest-typings/src/v1/integrations/IntegrationsCreateProps.ts
```diff
@@ -70,7 +70,7 @@ const integrationsCreateSchema = {
 				},
 				overrideDestinationChannelEnabled: {
 					type: 'boolean',
-					nullable: false,
+					nullable: true,
```
debdutdeb, line 73: if ajv is being used, then why Match too?

matheusbsilva137: ~i was afraid of removing it and creating a new issue~ done!

MarcosSpessatto (changes requested):
@matheusbsilva137 I believe we still need the Match verification (even if we use the AJV validator). The AJV validator validates only when the request comes from the rest API, but we could also receive a request from the Meteor method 😞. I think it is worth keeping the Match validation, at least until we get rid of the meteor methods.

ggazzo (approved):

PR #30222, opened 2023-08-30
feat: Save visitor's activity on agent's interaction

KevLehman:
NBJ-301


apps/meteor/app/livechat/server/hooks/markRoomResponded.ts
```diff
@@ -23,6 +24,12 @@ callbacks.add(
 			await LivechatRooms.setAgentLastMessageTs(room._id);
 		}
 
+		// Return MMMM-YY from moment
```
MarcosSpessatto, line 27: Isn't this YYYY-MM? 👀

apps/meteor/server/models/raw/LivechatVisitors.ts
```diff
@@ -365,6 +366,32 @@ export class LivechatVisitorsRaw extends BaseRaw<ILivechatVisitor> implements IL
...
+	markVisitorActiveForPeriod(visitorId: string, period: string): Promise<UpdateResult> {
+		const query = {
+			_id: visitorId,
+		};
+
+		const update = {
+			$push: {
+				activity: {
+					$each: [period],
+					$slice: -12,
```
MarcosSpessatto, line 388: Any specific reason for us to keep only the most 12 recent ones? 🤔

apps/meteor/server/models/raw/LivechatVisitors.ts
```diff
@@ -32,6 +32,7 @@ export class LivechatVisitorsRaw extends BaseRaw<ILivechatVisitor> implements IL
 			{ key: { username: 1 } },
 			{ key: { 'contactMananger.username': 1 }, sparse: true },
 			{ key: { 'livechatData.$**': 1 } },
+			{ key: { activity: 1 } },
```
MarcosSpessatto, line 35: I wonder if it would be better to use a sparse index instead. 🤔

MarcosSpessatto (changes requested):

apps/meteor/server/models/raw/LivechatVisitors.ts
KevLehman: No actual reason, just thought 1Y of data would be enough for most cases. We can increase it, what i don't want is keeping it unbounded.

apps/meteor/server/models/raw/LivechatVisitors.ts
KevLehman: Huh, this should had a partialFilterexp 🤔

apps/meteor/app/livechat/server/hooks/markRoomResponded.ts
```diff
@@ -26,6 +27,19 @@ callbacks.add(
 			return message;
 		}
 
+		// Return YYYY-MM from moment
+		const monthYear = moment().format('YYYY-MM');
+		const isVisitorActive = await LivechatVisitors.isVisitorActiveOnPeriod(room.v._id, monthYear);
+		if (!isVisitorActive) {
+			await LivechatVisitors.markVisitorActiveForPeriod(room.v._id, monthYear);
```
MarcosSpessatto, line 34: Is it better if we use the same logic as you are using with the Room? You are using `$addToSet` there... or at least just calling the update operation without checking it first (adding the query filter to the update operation, something like `markVisitorActiveForPeriodIfNotActive`), do you think this makes sense? 🤔

MarcosSpessatto (changes requested):

apps/meteor/app/livechat/server/hooks/markRoomResponded.ts
KevLehman: I thought of having the same logic, but at the same time, it's really difficult for a room to span across multiple months... unless there's a weird workspace over there that never closes them. But it's a really low probability of happening, so the $addToSet should work there.

For visitors, they're reused infinitely, so having the check for "no more than X" it's better suited there.

The problem: on mongo 4.x, this type of update (preventing duplications) while limiting the size of the array is not possible with addToSet, so to avoid the array growing infinitely from duplicate values & limiting the size, we do the query this way

MarcosSpessatto (dismissed):

aleksandernsilva (approved):

ggazzo (approved):

PR #30235, opened 2023-08-31
chore: Fix envvar name in patch notes

d-gubert:

This might look silly but since the typo is the name of the envvar I thought it was worth to fix it. This is, of course, only possible because the patch note hasn't been released in any candidate or version yet.


KevLehman (approved):

sampaiodiego:
from what I could see both are being used 🤔

PR #30241, opened 2023-08-31
chore: Move bad words filter callback to service

sampaiodiego:


apps/meteor/server/services/messages/service.ts
```diff
@@ -2,17 +2,42 @@ import type { IMessageService } from '@rocket.chat/core-services';
...
+	async created() {
+		await this.configureBadWords();
+	}
+
+	private async configureBadWords() {
+		if (!this.isBadWordsFilterEnabled()) {
+			return;
+		}
+
+		settings.watchMultiple(
```
MarcosSpessatto, line 29: I believe if the setting to enable bad words is `false` when the service was created, this will not execute and will not listen for changes on the setting after the service was created. 😬

MarcosSpessatto (changes requested):

apps/meteor/server/services/messages/service.ts
sampaiodiego: good call.. let me fix that

MarcosSpessatto (approved):

PR #30245, opened 2023-08-31
feat: Disable and annonimize visitors instead of removing and stats

KevLehman:
NBJ-317
NBJ-318
NBJ-319


MarcosSpessatto (changes requested):
@KevLehman what about the other `LivechatVisitors` methods?

I saw you created some methods to retrieve only the enabled visitors. I assume if you didn't put the enabled filter there as well it's because they don't need it. But I'm concerned because there are a lot of places all over the code using those methods.

MarcosSpessatto (approved):

PR #30269, opened 2023-09-04
feat: push notification statistics

heitortanoue:

Added statistic `push`, which is a number (int) with three bits representing booleans, following the structure:
```
1 1 1
| | |
| | +- push enabled = 0b1 = 1
| +--- push gateway enabled = 0b10 = 2
+----- push gateway changed = 0b100 = 4
```

## Further comments

TC-875


matheusbsilva137 (dismissed):

PedroRorato (approved):

apps/meteor/app/statistics/server/lib/statistics.ts
```diff
@@ -506,6 +506,10 @@ export const statistics = {
 		statistics.totalCustomRoles = await RolesRaw.findCustomRoles({ readPreference }).count();
 		statistics.totalWebRTCCalls = settings.get('WebRTC_Calls_Count');
 		statistics.uncaughtExceptionsCount = settings.get('Uncaught_Exceptions_Count');
+		statistics.pushEnabled = settings.get('Push_enable');
+
+		const defaultGateway = (await Settings.findOneById('Push_gateway'))?.packageValue;
```
sampaiodiego, line 511: if you're looking for `packageValue` field only, add a projection:
```suggestion
		const defaultGateway = (await Settings.findOneById('Push_gateway', { projection: { packageValue: 1 } }))?.packageValue;
```

sampaiodiego (changes requested):
regarding the reasoning behind the need to have this information added to the statistics, I do think it makes more sense to track `Push_enable_gateway` than if the gateway was changed or not.. because usually people will either use our apps with our gateway or change do not use our gateway with their own mobile apps.

and since we've been adding too much information to the statistics recently, I'd like to encourage some thinking on how to improve it.. for example, if we're tracking many different push related settings, maybe we can use a binary value? like:

we want to track 3 different settings, each one of them receives receives a bit:

```
1 1 1
| | |
| | +- push enabled = 0b1 = 
| +--- push gateway enabled = 0b10 = 2
+----- push gateway changed = 0b100 = 4
```

this way we could have a single value sent to the statistics using bitwise operators, like this:

```js
const pushEnabled = settings.get('Push_enable') ? 1 : 0;
const pushGatewayEnabled = settings.get('Push_enable_gateway') ? 2 : 0;
const gatewayChanged = hasGatewayChanged() ? 4 : 0;

statistics.push = pushEnabled | pushGatewayEnabled | gatewayChanged;
```

just an idea I think we should consider =)

apps/meteor/client/views/admin/info/UsageCard.stories.tsx
```diff
@@ -243,6 +243,8 @@ export default {
 			totalCustomRoles: 0,
 			totalWebRTCCalls: 0,
 			uncaughtExceptionsCount: 0,
+			pushEnabled: false,
```
matheusbsilva137, line 246:
```suggestion
```
I think you forgot to remove this line :thinking: please re-add it if I'm wrong

.changeset/nice-chairs-add.md
```diff
@@ -0,0 +1,8 @@
+---
+"@rocket.chat/meteor": minor
+"@rocket.chat/core-typings": minor
+---
+
+Added `pushEnabled` and `pushGatewayChanged` statistics.
+ - `pushEnabled`: mirrors the value of the `Push_enable` setting;
+ - `pushGatewayChanged`: a boolean value to indicate whether the `Push_gateway` setting is set as default or not;
```
matheusbsilva137, line 8: Please update this description

matheusbsilva137 (changes requested):

matheusbsilva137 (approved):

sampaiodiego (approved):

yash-rajpal (approved):

PR #30277, opened 2023-09-04
fix: engagement dashboard not working

tapiarafael:
The aggregate function was not being call because of the name in the cron job is being the same of the finish sessions function, resulting in not having information to display on the Engagement Dashboard.

## Steps to test or reproduce
1. Use the platform to generate session records;
2. Wait for the cron job to run _(02:00 am)_;
3. Check if any data was generated in `admin > engagement`;

## Further comments

[WM-158](https://rocketchat.atlassian.net/browse/WM-158)

[WM-158]: https://rocketchat.atlassian.net/browse/WM-158?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/meteor/app/statistics/server/lib/SAUMonitor.ts
```diff
@@ -321,7 +321,7 @@ export class SAUMonitorClass {
 		logger.info('[aggregate] - Aggregating data.');
 
 		const date = new Date();
-		date.setDate(date.getDate() - 0); // yesterday
+		date.setDate(date.getDate() - 1); // yesterday
```
matheusbsilva137, line 324:
```suggestion
```
The idea in this line is just to make clear that this date is set like this so that we can aggregate the data from "yesterday" (this date is a superior constraint, as you can check in `dailySessionsOfYesterday`). So I think the approaches we can choose here are either to remove this line or keep it as it is so that the engagement dashboard continues to work
Honestly, I think it's confusing as it is hahah so maybe we should remove it

matheusbsilva137 (changes requested):

apps/meteor/app/statistics/server/lib/SAUMonitor.ts
tapiarafael: Okay, got it. I was thinking that was wrong since `getDate() - 0` isn't yesterday 😆

pierre-lehnen-rc (approved):

matheusbsilva137 (dismissed):

MarcosSpessatto (approved):

PR #30282, opened 2023-09-05
fix: Presence broadcast isn't enabled after license validation

matheusbsilva137:

### DIFF
--- a/ee/packages/presence/src/Presence.ts
+++ b/ee/packages/presence/src/Presence.ts
@@ -39,9 +39,13 @@ export class Presence extends ServiceClass implements IPresence {
 			}
 		});
 
-		this.onEvent('license.module', ({ module, valid }) => {
+		this.onEvent('license.module', async ({ module, valid }) => {
 			if (module === 'scalability') {
 				this.hasLicense = valid;
+				const presenceBroadcastDisabled = await Settings.findOneById('Troubleshoot_Disable_Presence_Broadcast');
+				if (presenceBroadcastDisabled) {
+					await this.toggleBroadcast(!presenceBroadcastDisabled.value);
+				}
 			}
 		});
 	}


ee/packages/presence/src/Presence.ts
```diff
@@ -39,9 +39,13 @@ export class Presence extends ServiceClass implements IPresence {
 			}
 		});
 
-		this.onEvent('license.module', ({ module, valid }) => {
+		this.onEvent('license.module', async ({ module, valid }) => {
```
MarcosSpessatto, line 42: I believe this partially fixes the problem. Testing it locally, it looks like this event is not being received (even having a valid license). I'm unsure if this is the intended behavior or a problem with the `onModule` function.

Also, if I'm not wrong,  `started()` method solves the problem. https://github.com/RocketChat/Rocket.Chat/blob/develop/ee/packages/presence/src/Presence.ts#L61

The only problem that could happen is, while `started()` is executing AND before it fully finishes, we receive an event on the `watch.instanceStatus` turning the setting to true again. What I'm trying to say is, the execution must be at [this](https://github.com/RocketChat/Rocket.Chat/blob/develop/ee/packages/presence/src/Presence.ts#L239) point when the [`started()`](https://github.com/RocketChat/Rocket.Chat/blob/develop/ee/packages/presence/src/Presence.ts#L63) finishes.

I think @sampaiodiego opinion is really important here.

MarcosSpessatto (changes requested):

ee/packages/presence/src/Presence.ts
sampaiodiego: I don't think this is the best solution either..

if the problem is really the method `validateAvailability` being called **before** the start up procedure, we should actually make sure that doesn't happen by adding some kind of control like `started = true`.. or maybe making like `hasLicense = boolean | null` , so if it is still `null` we ignore all events.

ee/packages/presence/src/Presence.ts
```diff
@@ -232,14 +232,19 @@ export class Presence extends ServiceClass implements IPresence {
...
+		if (this.hasLicense !== false) {
 			return;
 		}
 
 		if (this.getTotalConnections() > MAX_CONNECTIONS) {
 			this.broadcastEnabled = false;
 
 			await Settings.updateValueById('Presence_broadcast_disabled', true);
+		} else {
+			const presenceBroadcastDisabled = await Settings.findOneById('Troubleshoot_Disable_Presence_Broadcast');
```
debdutdeb, line 244: this looks expensive, doesn't it? potentially two trips for every new connection?

matheusbsilva137: Yeah, it did :grimacing: 
I've just had a better idea, I'll check the `broadcastEnabled` property to make sure that presence broadcast is always enabled when there's a valid license (so we just do it once, not at every new connection).

ee/packages/presence/src/Presence.ts
matheusbsilva137: Thought of another approach: [here](https://github.com/RocketChat/Rocket.Chat/pull/30282/files#diff-4e5b24fc44a40a717ed7e5f19fcbbb5c978f13eac88725313af59ff3c7def3e9R236)
It's basically calling `toggleBroadcast` directly on new connections if the license is active and presence broadcast is disabled

ee/packages/presence/src/Presence.ts
```diff
@@ -39,9 +39,17 @@ export class Presence extends ServiceClass implements IPresence {
...
 		});
 
-		this.onEvent('license.module', ({ module, valid }) => {
+		this.onEvent('license.module', async ({ module, valid }) => {
 			if (module === 'scalability') {
 				this.hasLicense = valid;
+
+				if (!this.broadcastEnabled && valid) {
+					// broadcast should always be enabled if license is active (unless the troubleshoot setting is on)
+					const presenceBroadcastDisabled = await Settings.findOneById('Troubleshoot_Disable_Presence_Broadcast');
```
sampaiodiego, line 48: checking this setting is not wrong, but it is not a concern of this package..

we need to create an alternate mechanism to properly respect this setting, as it is already being "ignored" in other places.

sampaiodiego (changes requested):

sampaiodiego (approved):

benedikt-wegmann:
@matheusbsilva137 @sampaiodiego Please backport this change to the 6.3 and 6.4 branches as well, because - frankly - this is a pain and happens occasionally without any (intended) change on a running installation, like restarting a service.

(ping @ankar84 )

benedikt-wegmann:
Ok, seems like it is included in the 6.3.7 changelog as well. :+1:

PR #29529, opened 2023-06-13
fix: Newly added agent not following business hours

murtaza98:
## Issue(s)

https://rocketchat.atlassian.net/browse/OC-1047
## Further comments

List of changes being done over here:
- Fix the issue where newly added agents were not following the business hour rule.
- Cleanup opened BH prop from agents when they are removed as agents
- Listen to add and remove agent events from the user's admin panel. Similarly, listen to activate and deactivate user events and cleanup/add agent-related fields
- Avoid double checking whether agent's BH is active when the agent manually tries to turn on their Livechat status

Suggested API tests for future:
- [x] When a new agent is added, verify if they follow the default BH rule.
- [x] When a user is removed as an agent, verify if BH-related props were cleaned from their user object.

PR #30284, opened 2023-09-05
fix: agent role being removed upon user deactivation

murtaza98:
## Further comments

This reverts the change we did here in terms of removing agent role upon user deactivation -> https://github.com/RocketChat/Rocket.Chat/pull/29529


KevLehman (dismissed):

apps/meteor/app/livechat/server/business-hour/Single.ts
```diff
@@ -26,6 +27,39 @@ export class SingleBusinessHourBehavior extends AbstractBusinessHourBehavior imp
...
+			return;
+		}
+
+		await Users.addBusinessHourByAgentIds([agentId], defaultBusinessHour._id);
+
+		businessHourLogger.debug({
+			msg: 'Business hours found. Moving agent to AVAILABLE status',
+			agentId,
+			newStatus: ILivechatAgentStatus.AVAILABLE,
+		});
```
ggazzo, line 60: are we going to debug every single step we code? is this really necessary ?

murtaza98: I did found them quite useful while debugging locally but I guess I can re-check if we can remove some of them

murtaza98: Removed a few logs. The once I kept around are at the end of an execution path, so although at first it might seem like a function has a bit too much logs, but keep in mind that at runtime only once of those paths will be executed so only one log will get emitted

KevLehman: Side note: Yeah, we're doing some efforts on removing some of the logs, but on this specific feature, having a good amount of them is really useful to map problems.

KevLehman (approved):

KevLehman (approved):

ggazzo (changes requested):
please remove all the useless debug not only the ones that I pointed

murtaza98:
> please remove all the useless debug not only the ones that I pointed

@ggazzo Did you get a chance to look at the discussion here? https://github.com/RocketChat/Rocket.Chat/pull/30284#discussion_r1317253996

If you still think some logs here are excessive, then we'd appreciate if you point us to these logs specifically, and share us your point of view on why you think it would be un-necessary :pray:

KevLehman (approved):
I think we can improve a bit on log writing 🤔 the actual text of it sounds a bit too verbose.

Anyways, don't think that should prevent the PR from being reviewed, specially when there's no agreement or guide in how to put "good logs" and everyone has a diff opinion:)

ggazzo:
I certainly saw the thread, sure you've been reading my review requests? because it seems like I was ignored twice.

> > please remove all the useless debug not only the ones that I pointed
> 
> @ggazzo Did you get a chance to look at the discussion here? [#30284 (comment)](https://github.com/RocketChat/Rocket.Chat/pull/30284#discussion_r1317253996)
> 
that's exactly what I did by asking you to remove the logs again. If I may ask, you seem so concerned about debugging, are there any tests ensuring that what you did fixed the problem?
> If you still think some logs here are excessive, then we'd appreciate if you point us to these logs specifically, and share us your point of view on why you think it would be un-necessary 🙏



Just because we "didnt create a standard", that doesn't mean we need to opt for the worst choice and that we are allowed to do anything justifiably, you have done this repeatedly through the features and I don't see it helping us to prevent bugs. if you don't feel confident in the conditions made there are tests to mitigate these issues.

> I think we can improve a bit on log writing 🤔 the actual text of it sounds a bit too verbose.
> 
> Anyways, don't think that should prevent the PR from being reviewed, specially when there's no agreement or guide in how to put "good logs" and everyone has a diff opinion:)

ggazzo:
I'm not going to approve because I dont

ggazzo:
@murtaza98 would you mind to cherry-pick this to branch release you indicated ? I'm afraid the changes are not 100% compatible, thanks

KevLehman:
@ggazzo , don't think we have permissions to cherry pick onto that branch. Best I can do is to get yet another PR targetted to the release branch.

murtaza98:
In terms of logs, I do understand that I'm adding a bit too much specifically for Business Hour (BH) feature; and it's mainly because of the instability surrounding this feature. The support tickets surrounding this feature usually are intermittent in nature and that's where I've usually found these debug logs to be quite helpful in the past. 

But considering we've already put in a considerable effort in the past month towards the stability of this feature, I guess we could now stop being extra cautious and could've removed all the debug logs we added here. I'll keep a note of this and if don't see any crazy support tickets surrounding this feature in the coming months, then I'll make time to remove all such logs.

Your suggestion about adding tests is certainly a good idea @ggazzo  and I'm planning to add those in a separate PR

ggazzo:
> @ggazzo , don't think we have permissions to cherry pick onto that branch. Best I can do is to get yet another PR targetted to the release branch.
> 
> IF not needed, lmk so i can close it #30323

could you share with us the error you faced during the cherry-pick? this is supposed to be the normal way to do patches, if you dont have permissions something is wrong

murtaza98:
@ggazzo I faced this error while pushing a commit to the release branch. Perhaps its the same Kevin was facing

image

PR #30286, opened 2023-09-05
test: tests for muting and unmuting

heitortanoue:

Added tests for `muteUserInRoom` and `unmuteUserInRoom`, in standard and in read-only rooms.

## Further comments

TC-942


apps/meteor/tests/end-to-end/api/24-methods.js
```diff
@@ -2397,4 +2397,274 @@ describe('Meteor.methods', function () {
...
+				.expect(200)
+				.expect((res) => {
+					expect(res.body).to.have.property('success', true);
+					expect(res.body).to.have.nested.property('channel._id');
+					expect(res.body).to.have.nested.property('channel.name', channelName);
+					expect(res.body).to.have.nested.property('channel.t', 'c');
+					expect(res.body).to.have.nested.property('channel.msgs', 0);
+					rid = res.body.channel._id;
+				})
+				.end(done);
```
matheusbsilva137, line 2425: You can use the `createRoom` helper instead

apps/meteor/tests/end-to-end/api/24-methods.js
```diff
@@ -2397,4 +2397,274 @@ describe('Meteor.methods', function () {
...
+			const username = `user.test.${Date.now()}`;
+			const email = `${username}@rocket.chat`;
+			request
+				.post(api('users.create'))
+				.set(credentials)
+				.send({ email, name: username, username, password: username })
+				.end((err, res) => {
+					testUser = res.body.user;
+					done();
+				});
```
matheusbsilva137, line 2438: You can use the `createUser` helper instead

apps/meteor/tests/end-to-end/api/24-methods.js
```diff
@@ -2397,4 +2397,274 @@ describe('Meteor.methods', function () {
...
+			request
+				.post(api('channels.invite'))
+				.set(credentials)
+				.send({
+					roomId: rid,
+					userId: testUser._id,
+				})
+				.expect('Content-Type', 'application/json')
+				.expect(200)
+				.end(done);
```
matheusbsilva137, line 2451: Provide the `userId` in the `members` array from the `createRoom` helper so that this step isn't needed

apps/meteor/tests/end-to-end/api/24-methods.js
```diff
@@ -2397,4 +2397,274 @@ describe('Meteor.methods', function () {
...
+					user: testUser.username,
+					password: testUser.username,
+				})
+				.expect('Content-Type', 'application/json')
+				.expect(200)
+				.expect((res) => {
+					testUserCredentials['X-Auth-Token'] = res.body.data.authToken;
+					testUserCredentials['X-User-Id'] = res.body.data.userId;
+				})
+				.end(done);
```
matheusbsilva137, line 2467: You can use the `login` helper instead

apps/meteor/tests/end-to-end/api/24-methods.js
```diff
@@ -2397,4 +2397,274 @@ describe('Meteor.methods', function () {
...
+							expect(res.body).to.have.a.property('message').that.is.a('string');
+							const data = JSON.parse(res.body.message);
+							expect(data).to.have.a.property('msg', 'result');
+							expect(data).to.have.a.property('id', 'id');
+							expect(data).not.to.have.a.property('error');
+						})
+						.end(done);
+				});
+
+				it('muted user should not be able to send message', (done) => {
```
matheusbsilva137, line 2497: Please prefer using async functions instead of `done` in the new lines you added. You can change all `it` and `before` to be async (`it('something', async () => )` so you just need to `await` instead of calling `end(done`))

matheusbsilva137 (changes requested):
Can we also add a test for the case of changing a standard room to read-only? In order to check if previously unmuted users are now muted

matheusbsilva137 (approved):

PR #30290, opened 2023-09-05
chore: Improve cache of static files

rodrigok:


.changeset/wicked-humans-hang.md
```diff
@@ -0,0 +1,5 @@
+---
+"@rocket.chat/meteor": minor
```
sampaiodiego, line 2:
```suggestion
"@rocket.chat/meteor": patch
```

it needs to be flagged as `patch` if we want to target a patch release.. otherwise it will bump a minor version

apps/meteor/app/cors/server/cors.ts
```diff
@@ -77,6 +79,17 @@ WebApp.rawConnectHandlers.use((_req: http.IncomingMessage, res: http.ServerRespo
...
+	// eslint-disable-next-line @typescript-eslint/no-namespace
+	namespace WebApp {
+		function categorizeRequest(
+			req: http.IncomingMessage,
+		): { arch: string; path: string; url: UrlWithParsedQuery } & Record<string, unknown>;
+	}
+}
+
+// These routes already handle cache control on their own
+const cacheControlledRoutes: Array<string> = ['/assets', '/custom-sounds', '/emoji-custom', '/avatar', '/file-upload'];
```
sampaiodiego, line 92: if we have the regexps already on this array, we don't need to do a `new RegExp` on every request:

```suggestion
const cacheControlledRoutes: Array<string> = ['/assets', '/custom-sounds', '/emoji-custom', '/avatar', '/file-upload'].map((route) => new RegExp(`^${route}`, 'i'));
```

apps/meteor/app/cors/server/cors.ts
```diff
@@ -86,6 +99,37 @@ WebAppInternals.staticFilesMiddleware = function (
 	next: NextFunction,
 ) {
 	res.setHeader('Access-Control-Allow-Origin', '*');
+	const { arch, path, url } = WebApp.categorizeRequest(req);
+
+	if (
+		Meteor.isProduction &&
+		!cacheControlledRoutes.some((route) => {
+			return new RegExp(`^${route}`, 'gi').test(path);
+		})
```
sampaiodiego, line 108: complementing the suggestion above:
```suggestion
		!cacheControlledRoutes.some((regexp) => regexp.test(path))
```

sampaiodiego (changes requested):

apps/meteor/app/cors/server/cors.ts
gabriellsh: I was using the global option before, and I know that running the same regex with the global option enabled can have problems. I just realized I don't actually need the 'g' flag 🙈 . Thanks for the tip.

apps/meteor/app/ui-master/server/inject.ts
```diff
@@ -76,27 +76,33 @@ export const injectIntoHead = (key: string, value: Injection): void => {
 };
 
 export const addScript = (key: string, content: string): void => {
+	const currentHash = crypto.createHash('sha1').update(content).digest('hex');
+	key += `_${currentHash}`;
```
sampaiodiego, line 80: the `key` provided to `injectIntoHead ` cannot have the hash, it must stay the same, otherwise multiple assets will be created everytime the content changes..

apps/meteor/app/ui-master/server/inject.ts
```diff
@@ -76,27 +76,33 @@ export const injectIntoHead = (key: string, value: Injection): void => {
...
 		type: 'JS',
-		tag: `<script id="${key}" type="text/javascript" src="${`${getURL(key)}.js?${currentHash}`}"></script>`,
+		tag: `<script id="${key}" type="text/javascript" src="${`${getURL(key)}.js`}"></script>`,
 		content,
 	});
 };
 
 export const addStyle = (key: string, content: string): void => {
+	const currentHash = crypto.createHash('sha1').update(content).digest('hex');
+	key += `_${currentHash}`;
```
sampaiodiego, line 96: same as above, the `key` must not change

sampaiodiego (changes requested):

apps/meteor/app/ui-master/server/inject.ts
rodrigok: And they will be added to the HTML? Not having the hash in the key can create issues with caching wrong files when doing rolling upgrades

sampaiodiego (approved):

PR #30296, opened 2023-09-06
feat: User Panel > Tab PENDING

felipe-rod123:

### DIFF
--- a/apps/meteor/client/views/admin/users/AdminUsersPage.tsx
+++ b/apps/meteor/client/views/admin/users/AdminUsersPage.tsx
@@ -25,7 +25,7 @@ const UsersPage = (): ReactElement => {
 	const canCreateUser = usePermission('create-user');
 	const canBulkCreateUser = usePermission('bulk-register-user');
 
-	const [tab, setTab] = useState<'all' | 'invited' | 'new' | 'active' | 'deactivated'>('all');
+	const [tab, setTab] = useState<'all' | 'invited' | 'pending' | 'active' | 'deactivated'>('all');
 
 	useEffect(() => {
 		if (!context || !seatsCap) {
@@ -51,7 +51,7 @@ const UsersPage = (): ReactElement => {
 					) : (
 						<ButtonGroup>
 							{canCreateUser && (
-								<Button icon='user-plus' onClick={() => router.navigate('/admin/users/new')}>
+								<Button icon='user-plus' onClick={() => router.navigate('/admin/users/pending')}>
 									{t('New')}
 								</Button>
 							)}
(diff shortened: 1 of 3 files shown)


apps/meteor/client/views/admin/users/hooks/useFilterActiveUsers.ts
```diff
@@ -1,9 +1,6 @@
 import type { IUser } from '@rocket.chat/core-typings';
 
-export const useFilterActiveUsers = (
-	users: Pick<IUser, '_id' | 'username' | 'name' | 'status' | 'roles' | 'emails' | 'active' | 'avatarETag'>[] | undefined,
```
rique223, line 4: This typing change is not correct, rest-typings for users.list have been set up to only return this subset of params of the users object, changing it to `Partial<IUser>` will make all fields in the User Object optional and cause errors like the one in the line 180 of `UsersTable.ts` in which the `_id` of the users returned to `filteredUsers` can be undefined whilst the `_id` of the users that `UsersTableRow` expect can not. This happens because IUser sets `_id` as mandatory and this should be respected.

apps/meteor/client/views/admin/users/UsersTable/UsersTable.tsx
```diff
@@ -133,17 +151,34 @@ const UsersTable = ({ reload, tab, onReload }: UsersTableProps): ReactElement |
...
 						<GenericTableHeader>{headers}</GenericTableHeader>
 						<GenericTableBody>
+							{/* TODO: fix user type */}
 							{filteredUsers.map((user) => (
 								<UsersTableRow
 									key={user._id}
 									onClick={handleClickOrKeyDown}
 									mediaQuery={mediaQuery}
-									user={user}
+									user={user as any}
```
rique223, line 180: This problem here was caused due to the change of the type of the users param of the three filter hooks, TS is infering the type of user to be Partial<IUser>[] when in fact it is and shouldnt be this.

rique223 (changes requested):

apps/meteor/client/views/admin/users/AdminUsersPage.tsx
```diff
@@ -68,18 +70,18 @@ const UsersPage = (): ReactElement => {
 						<Tabs.Item selected={!tab || tab === 'all'} onClick={() => setTab('all')}>
 							{t('All')}
 						</Tabs.Item>
-						<Tabs.Item selected={tab === 'invited'} onClick={() => setTab('invited')}>
-							{t('Invited')}
-						</Tabs.Item>
-						<Tabs.Item selected={tab === 'new'} onClick={() => setTab('new')}>
-							{t('New_users')}
+						<Tabs.Item selected={tab === 'pending'} onClick={() => setTab('pending')}>
+							{pendingActionsCount === 0 ? t('Pending') : `${t('Pending')} (${pendingActionsCount})`}
```
rique223, line 74: pendingActionsCount is defined and used but since the set is never ran the number will never change. Why is it incomplete?

rique223 (changes requested):

PR #30313, opened 2023-09-06
chore: remove istanbul from prod env

ggazzo:

### DIFF
--- a/.github/actions/build-docker/action.yml
+++ b/.github/actions/build-docker/action.yml
@@ -28,6 +28,7 @@ runs:
         path: /tmp/build
 
     - name: Unpack build
+      shell: bash
       run: |
         cd /tmp/build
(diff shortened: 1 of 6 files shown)


.github/actions/build-docker/action.yml
```diff
@@ -0,0 +1,69 @@
...
+
+    - run: yarn build
+      shell: bash
+
+    - name: Build Docker images
+      shell: bash
+      run: |
+        args=(rocketchat)
+
+        if [[ '${{ matrix.platform }}' = 'alpine' ]]; then
```
sampaiodiego, line 54: there is no `matrix` here.. this value need to come from an input

sampaiodiego (changes requested):

sampaiodiego (approved):

PR #30315, opened 2023-09-06
regression: Rooms page filters faulty behavior

rique223:
## Jira task: [WM-173](https://rocketchat.atlassian.net/browse/WM-173?atlOrigin=eyJpIjoiNDM3NjJjZGNjOWIxNDY1NTlhYzk2MTliZDdkMmVlMzYiLCJwIjoiaiJ9)

This pull request addresses two issues:

1. The empty state for the rooms table was not displaying correctly due to an error in the empty verification process, which was considering the wrong rooms list.

2. The rooms filters were only filtering the current page of the rooms list instead of filtering the entire list. This issue arose because the rooms.adminRooms endpoint, the source of the rooms list, returns a paginated list, and the front-end only had access to the current page of this pagination. This regression highlighted the need to rework the back-end of the rooms page to accommodate visibility filter parameters. Given the current context, this PR refactors the rooms filters to function with the existing back-end, ensuring the stability of the page until it becomes possible to implement the alternative approach.

Demo gif of the paliative behavior:
paliative_rooms_filters
P.S.: The Channels option title was changed to "Public channels" after the recording of this gif;

## Further comments

[WM-173]: https://rocketchat.atlassian.net/browse/WM-173?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


hugocostadev (approved):

apps/meteor/client/views/admin/rooms/RoomsTable.tsx
```diff
@@ -248,24 +233,24 @@ const RoomsTable = ({ reload }: { reload: MutableRefObject<() => void> }): React
...
-						itemsPerPage={itemsPerPage}
-						count={data?.total || 0}
-						onSetItemsPerPage={setItemsPerPage}
-						onSetCurrent={setCurrent}
-						{...paginationProps}
-					/>
-				</>
-			)}
-			{isSuccess && data && data.rooms.length === 0 && <GenericNoResults />}
+
```
dougfabris, line 236: why? There's no need to show the table headers without results

apps/meteor/client/views/admin/rooms/RoomsTableFilters.tsx
```diff
@@ -27,46 +22,52 @@ const roomTypeFilterStructure = [
...
+// 	// 	checked: false,
+// 	// },
+// ] as OptionProp[];
+
 const RoomsTableFilters = ({ setFilters }: { setFilters: Dispatch<SetStateAction<any>> }): ReactElement => {
 	const t = useTranslation();
 	const [text, setText] = useState('');
 	const [roomTypeOptions, setRoomTypeOptions] = useState<OptionProp[]>(roomTypeFilterStructure);
-	const [roomVisibilityOptions, setRoomVisibilityOptions] = useState<OptionProp[]>(roomVisibilityFilterStructure);
+	// const [roomVisibilityOptions, setRoomVisibilityOptions] = useState<OptionProp[]>(roomVisibilityFilterStructure);
```
dougfabris, line 64: why leave it commented?

dougfabris (changes requested):

apps/meteor/client/views/admin/rooms/RoomsTableFilters.tsx
hugocostadev: We are going to implement when the backend does that

dougfabris: I would recommend not leaving comments on the code and reimplement when backend is done

apps/meteor/client/views/admin/rooms/RoomsTable.tsx
rique223: It already doesn't show it:
image

I also found it weird that it did this but since it did I didn't botter adding the duplicated verification. Do you think its good to do it?

dougfabris (approved):

PR #30321, opened 2023-09-07
fix: update type of message bridge's create method

Dnouv:

- A minor enhancement of types for the message `create` bridge
- This absolves us from either calling getById of room bridge, essentially touching the db to satisfy typescript of all the required fields (room id verification is already done in the bridge/converter, so this call would just add overhead without any benefits) or force cast and risk future changes breaking current code

## Issue(s)

https://github.com/RocketChat/Rocket.Chat.Apps-engine/pull/660#discussion_r1315682415
## Further comments

[AECO-360](https://rocketchat.atlassian.net/browse/AECO-360)

[AECO-360]: https://rocketchat.atlassian.net/browse/AECO-360?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


debdutdeb:
Don't think this needs a changeset from core.

debdutdeb (approved):
lgtm.

maybe remove the changeset, but apps team would be the best people to comment on that .

AllanPazRibeiro (approved):

PR #30332, opened 2023-09-08
regression: `AccountProfileForm` not reseting

dougfabris:

### DIFF
--- a/apps/meteor/client/views/account/profile/AccountProfileForm.tsx
+++ b/apps/meteor/client/views/account/profile/AccountProfileForm.tsx
@@ -22,7 +22,6 @@ import UserAvatarEditor from '../../../components/avatar/UserAvatarEditor';
 import { useUpdateAvatar } from '../../../hooks/useUpdateAvatar';
 import { USER_STATUS_TEXT_MAX_LENGTH, BIO_TEXT_MAX_LENGTH } from '../../../lib/constants';
 import type { AccountProfileFormValues } from './getProfileInitialValues';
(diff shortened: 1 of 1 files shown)


apps/meteor/client/views/account/profile/AccountProfileForm.tsx
```diff
@@ -97,10 +96,10 @@ const AccountProfileForm = (props: AllHTMLAttributes<HTMLFormElement>): ReactEle
...
 			await updateOwnBasicInfo(
 				{
-					...(allowRealNameChange ? { realname: name } : {}),
-					...(allowEmailChange && user ? getUserEmailAddress(user) !== email && { email } : {}),
-					...(canChangeUsername ? { username } : {}),
-					...(allowUserStatusMessageChange ? { statusText } : {}),
+					realname: name,
+					...(user ? getUserEmailAddress(user) !== email && { email } : {}),
+					username,
+					statusText,
```
juliajforesti, line 102: I'm not able to add a comment to the line in question (112) since it was not changed on this pr
but now that `reset` is working properly, it doesn't make sense to reset with initial values, after updating:
```
reset(getProfileInitialValues(user));
```
if you change any field and save, it will fetch the previous info - as it does on first form load.
only if you reload, it brings the updated info. 

I would just suggest removing this reset on save - it only makes sense now on `cancel` action.

juliajforesti (changes requested):

juliajforesti (approved):

PR #30354, opened 2023-09-11
fix: Consider only sessions from last few days for SAU/MAU

sampaiodiego:

It changes the filter for the cronjob that runs daily to calculate the workspace MAU to fetch sessions only from 3 days ago.

This change is will only have an impact in case the cronjob has been skipped for quite a few days and will prevent large workspace from having the database hammered when it runs again.

### DIFF
--- a/apps/meteor/server/models/raw/Sessions.ts
+++ b/apps/meteor/server/models/raw/Sessions.ts
@@ -169,7 +169,7 @@ const getProjectionByFullDate = (): { day: string; month: string; year: string }
 export const aggregates = {
 	dailySessions(
 		collection: Collection<ISession>,
-		{ from, to }: { from: Date; to: Date },
+		{ start, end }: DestructuredRange,
 	): AggregationCursor<
 		Pick<ISession, 'mostImportantRole' | 'userId' | 'day' | 'year' | 'month' | 'type'> & {
 			time: number;
@@ -185,9 +185,7 @@ export const aggregates = {
 					lastActivityAt: { $exists: true },
 					device: { $exists: true },
 					type: 'session',
-
-					// we filter by createdAt to be able to perform a range query, since the "day", "month" and "year" fields were not meant for that
-					createdAt: { $gte: from, $lt: to },
+					...matchBasedOnDate(start, end),
 				},
 			},
 			{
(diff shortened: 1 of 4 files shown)


apps/meteor/server/models/raw/Sessions.ts
```diff
@@ -178,115 +178,103 @@ export const aggregates = {
...
-					$match: {
-						time: { $gt: 0 },
-					},
+		const pipeline = [
+			{
+				$match: {
+					userId: { $exists: true },
+					lastActivityAt: { $exists: true },
+					device: { $exists: true },
+					type: 'session',
```
KevLehman, line 187: Do you think an index on all the fields would cover the query better? 👀

apps/meteor/server/models/raw/Sessions.ts
```diff
@@ -178,115 +178,103 @@ export const aggregates = {
...
+			},
+		];
+
+		return collection.aggregate<
+			Pick<ISession, 'mostImportantRole' | 'userId' | 'day' | 'year' | 'month' | 'type'> & {
+				time: number;
+				sessions: number;
+				devices: ISession['device'][];
+				_computedAt: string;
+			}
```
KevLehman, line 276: This should be its own type, since you're reusing it above

apps/meteor/server/models/raw/Sessions.ts
```diff
@@ -1616,4 +1604,25 @@ export class SessionsRaw extends BaseRaw<ISession> implements ISessionsModel {
...
+	}
+
+	async updateAllSessionsByDateToComputed(from: Date, to: Date): Promise<UpdateResult | Document> {
+		return this.updateMany(
+			{
+				type: 'session',
+
+				// we filter by createdAt to be able to perform a range query, since the "day", "month" and "year" fields were not meant for that
+				createdAt: { $gte: from, $lt: to },
+			},
```
KevLehman, line 1619: Do we need an index on type+createdAt? Not sure how the collection data is composed

MarcosSpessatto (approved):

apps/meteor/server/models/raw/Sessions.ts
sampaiodiego: 👍 we can do that later

PR #30373, opened 2023-09-12
feat: New records page analytics tab

rique223:
## Jira task: [WM-172](https://rocketchat.atlassian.net/browse/WM-172?atlOrigin=eyJpIjoiNjI3NDQ3ZTAwNjc5NDQ0ZWIzZjBkMTAwNTM1NTI5NTgiLCJwIjoiaiJ9)

Changed the name/route of the administration Logs page to Records, implemented a tab-based layout in the page, and added a new "Analytics Reports" tab that retrieves data from the /v1/statistics endpoint, presenting it beautified and indented with tabs.

Demo gif:
records_analytics

## Further comments

[WM-172]: https://rocketchat.atlassian.net/browse/WM-172?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/meteor/client/views/admin/viewLogs/AnalyticsReports.tsx
```diff
@@ -0,0 +1,38 @@
...
+						By sending your statistics, you'll help us identify how many instances of
+						<Link href='https://rocket.chat/' target='_blank' rel='noopener noreferrer'>
+							Rocket.Chat
+						</Link>
+						are deployed, as well as how good the system is behaving, so we can further improve it. Don't worry, as no user information is
+						sent and all the information we receive is kept confidential.
+					</Trans>
+				</Box>
+			</Box>
+			<Box display='flex' flexDirection='column' padding={8} flexGrow={1} color='default' bg='neutral' borderRadius='x4' overflow='scroll'>
```
hugocostadev, line 31: avoid using 'x4' , use only the number 
```suggestion
			<Box display='flex' flexDirection='column' padding={8} flexGrow={1} color='default' bg='neutral' borderRadius={4} overflow='scroll'>
```

apps/meteor/client/views/admin/viewLogs/AnalyticsReports.tsx
```diff
@@ -0,0 +1,38 @@
...
+				<pre>{JSON.stringify(analytics, null, '\t')}</pre>
```
hugocostadev, line 32: That's OK, but I guess that you can use the highlight.js to beautify the payload if you want, the dependency is already there

apps/meteor/client/views/admin/viewLogs/hooks/useAnalyticsObject.ts
```diff
@@ -0,0 +1,8 @@
+import { useEndpoint } from '@rocket.chat/ui-contexts';
+import { useQuery } from '@tanstack/react-query';
+
+export const useAnalyticsObject = () => {
+	const getAnalytics = useEndpoint('GET', '/v1/statistics');
+
+	return useQuery(['analytics'], () => getAnalytics({}));
+};
```
hugocostadev, line 8: don't return the entire useQuery return, you can return only the necessary objects, and also you can get rid of `data` and send only the payload

apps/meteor/client/views/admin/viewLogs/ViewLogsPage.tsx
```diff
@@ -1,18 +1,33 @@
...
+import React, { useState } from 'react';
 
 import Page from '../../../components/Page';
+import AnalyticsReports from './AnalyticsReports';
 import ServerLogs from './ServerLogs';
+import { useAnalyticsObject } from './hooks/useAnalyticsObject';
 
 const ViewLogsPage = (): ReactElement => {
 	const t = useTranslation();
+	const { data, isSuccess } = useAnalyticsObject();
```
hugocostadev, line 14: You can fetch the data inside the `AnalyticsReport`

apps/meteor/packages/rocketchat-i18n/i18n/en.i18n.json
```diff
@@ -423,6 +424,7 @@
   "Analytics_features_users_Description": "Tracks custom events related to actions related to users (password reset times, profile picture change, etc).",
   "Analytics_Google": "Google Analytics",
   "Analytics_Google_id": "Tracking ID",
+  "analytics.page.bySending": "By sending your statistics, you'll help us identify how many instances of <1>Rocket.Chat</1> are deployed, as well as how good the system is behaving, so we can further improve it. Don't worry, as no user information is sent and all the information we receive is kept confidential.",
```
hugocostadev, line 427: Try to follow the same key structure for naming your keys...

hugocostadev (changes requested):

apps/meteor/client/views/admin/viewLogs/AnalyticsReports.tsx
rique223: The spec for this doesn't include highlighting so I figured it wasn't necessary. Also, I'd like to avoid depending on an external library to do such a simple thing. This goes without mentioning the overhead of using highlight.js.

apps/meteor/client/views/admin/viewLogs/ViewLogsPage.tsx
rique223: If I did this it would re-run the query, and consequently the rest request, on every tab change, creating a lot of unnecessary re-requests. On this note I belive it would be good to refactor the ServerLogs component to avoid unnecessary requests as well, but since it is quite old and probably will be deprecated in the future I didn't try to.

apps/meteor/client/views/admin/viewLogs/hooks/useAnalyticsObject.ts
rique223: If I do this react query will stop interacting correctly with itself and cause TS errors and, consequently, unnecessary overhead, also will make future usage of this hook harder than it needs to be.

rique223: If I do this react query will stop interacting correctly with itself and cause TS errors and, consequently, unnecessary overhead, also will make future usage of this hook harder than it needs to be.

rique223: For an example:
image
image

apps/meteor/client/views/admin/viewLogs/ViewLogsPage.tsx
hugocostadev: Why would be re-render? The useQuery has a cache mechanism that prevent that, it's just a good practice to use this hook, this was mentioned on the front-end meeting

hugocostadev: Also, we need to handle error scenario and refetch if necessary

hugocostadev (changes requested):

hugocostadev (approved):

tassoevan (approved):

PR #30381, opened 2023-09-13
chore: Add the possibility for removing the watcher for the message collection

MarcosSpessatto:

https://rocketchat.atlassian.net/browse/ARCH-1208


packages/core-services/src/events/listeners.ts
```diff
@@ -0,0 +1,12 @@
...
+import type { IServiceClass } from '../types/ServiceClass';
+
+export const dbWatchersDisabled = ['yes', 'true'].includes(String(process.env.DISABLE_DB_WATCHERS).toLowerCase());
+
+export const listenToMessageSentEvent = (service: IServiceClass, action: (message: IMessage) => Promise<void>): void => {
+	if (dbWatchersDisabled) {
+		return service.onEvent('message.sent', (message: IMessage) => action(message));
+	}
+	return service.onEvent('watch.messages', ({ message }) => action(message));
+};
```
MarcosSpessatto, line 12: @ggazzo @sampaiodiego I decided to create another event to handle the "manual" process of sending events instead of using the existing `watch.messages`.

PR #30383, opened 2023-09-13
fix: Custom translations do not work

matheusbsilva137:
## Issue(s)

Introduced in #29097

## Further comments

TC-953


yash-rajpal (approved):

tassoevan (changes requested):
Is ignoring the prefix (instead of assuming it's `core`) a good idea? Also, what about server-side translations?

yash-rajpal:
> Is ignoring the prefix (instead of assuming it's `core`) a good idea? Also, what about server-side translations?

The final `results` doesn't contain the namespace prefix, we remove the namespace prefix before adding it to the final translations object. In this case it wasn't working for custom-translations not having namespace prefix so this fix handles that case.

tassoevan (approved):

PR #30408, opened 2023-09-15
chore: Adapt some messaging features to work without the DB watcher

MarcosSpessatto:

This PR adapts the following features to work without the DB watcher:
- Send message;
- Update message;
- Delete message;
- Save system messages (only the ones sent through the MessageService)
- Discussions;

ARCH-1209
ARCH-1210
ARCH-1211
ARCH-1212
ARCH-1213
ARCH-1215
ARCH-1216
ARCH-1241
ARCH-1234
ARCH-1230


apps/meteor/app/lib/server/functions/deleteMessage.ts
```diff
@@ -86,11 +86,19 @@ export async function deleteMessage(message: IMessage, user: IUser): Promise<voi
...
 
-	await callbacks.run('afterDeleteMessage', deletedMsg, room);
-
 	// decrease message count
 	await Rooms.decreaseMessageCountById(message.rid, 1);
 
+	await callbacks.run('afterDeleteMessage', deletedMsg, room);
+
+	const mustNotifyNewMessageSent = keepHistory || showDeletedStatus;
+	if (mustNotifyNewMessageSent) {
```
sampaiodiego, line 95:
```suggestion
	if (keepHistory || showDeletedStatus) {
```

apps/meteor/app/lib/server/lib/notifyUsersOnMessage.js
```diff
@@ -191,4 +191,9 @@ export async function notifyUsersOnMessage(message, room) {
 	return message;
 }
 
-callbacks.add('afterSaveMessage', (message, room) => notifyUsersOnMessage(message, room), callbacks.priority.LOW, 'notifyUsersOnMessage');
+callbacks.add(
+	'afterSaveMessage',
+	(message, room) => notifyUsersOnMessage(message, room),
+	callbacks.priority.MEDIUM,
```
sampaiodiego, line 197: any reason to change this to MEDIUM? just curious

sampaiodiego (changes requested):

apps/meteor/app/lib/server/lib/notifyUsersOnMessage.js
MarcosSpessatto: This comes from the discussion feature. I changed that because I need that to run before the discussion callback, otherwise, we would end up without the correct number of messages within that room. https://github.com/RocketChat/Rocket.Chat/blob/28419b8278704677ee03a054ddfb66bff2b8f818/apps/meteor/app/lib/server/lib/notifyUsersOnMessage.js#L187

sampaiodiego (approved):

PR #30414, opened 2023-09-16
Live preview message composer

imf-ali:


imf-ali:
Code line added
image

 Output:
image

tassoevan (changes requested):
Rocket.Chat uses it's own parser (`@rocket.chat/message-parser`) instead of Markdown because we do not follow the CommonMark specification. Also, this parser outputs a structure that is not HTML, therefore it's not the format I'd personally pass to Quill to render. Maybe doing a translation between the message parser AST output and Quill's Deltas is the way to go.

imf-ali:
https://github.com/RocketChat/Rocket.Chat/pull/31731
Attaching this PR for reference, as came up with new implementation suggested by @tassoevan and @hugocostadev

PR #30430, opened 2023-09-18
chore:  remove connectToCloud option

ggazzo:
[NBJ-366]

## Further comments

[NBJ-366]: https://rocketchat.atlassian.net/browse/NBJ-366?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


tiagoevanp (dismissed):

apps/meteor/app/cloud/server/methods.ts
```diff
@@ -151,7 +150,9 @@ Meteor.methods<ServerMethods>({
 			});
 		}
 
-		return disconnectWorkspace();
+		await disconnectWorkspace();
```
sampaiodiego, line 153: instead of calling a function that has a meaningless error I'd suggest to return an error/message that actually tells to the client/end user what is going on.

apps/meteor/app/cloud/server/functions/disconnectWorkspace.ts
```diff
@@ -1,17 +1,4 @@
...
-	if (!connectToCloud) {
-		return true;
-	}
-
-	await Settings.updateValueById('Register_Server', false);
-
-	await syncWorkspace(true);
-
-	return true;
+	throw new Error('Deprecated');
```
sampaiodiego, line 3: any special reason to keep this function around? it is an internal function, IMO it is better to remove it and properly assess what it was supposed to do.

sampaiodiego (changes requested):

apps/meteor/app/cloud/server/methods.ts
sampaiodiego: I guess we need to also deprecate `cloud:disconnectWorkspace` method, right?

sampaiodiego (approved):

PR #30464, opened 2023-09-21
feat: Bot warning when mentioning users not in a channel

gabriellsh:
## Issue(s)

[TC-948](https://rocketchat.atlassian.net/browse/TC-948)
## Further comments

[TC-948]: https://rocketchat.atlassian.net/browse/TC-948?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


guijun13 (dismissed):

apps/meteor/app/lib/server/startup/mentionUserNotInChannel.ts
```diff
@@ -0,0 +1,143 @@
...
+
+		// get all users that are mentioned but not in the channel
+		const mentionsUsersNotInChannel = mentions.filter(({ _id }) => !subs.some((sub) => sub.u._id === _id));
+
+		if (!mentionsUsersNotInChannel.length) {
+			return message;
+		}
+
+		const canAddUsersToThisRoom = await hasPermissionAsync(message.u._id, 'add-user-to-joined-room', message.rid);
+		const canAddToAnyRoom = await (room.t === 'c'
```
MarcosSpessatto, line 89: We were adding the users by default only on public channels, now it looks like we are doing the same for private channels (if the user has permission) as well, are we ok with this behavior change?

apps/meteor/app/lib/server/startup/mentionUserNotInChannel.ts
```diff
@@ -0,0 +1,143 @@
...
+		if (isEditedMessage(message) || (message.ts && Math.abs(moment(message.ts).diff(moment())) > 60000) || !message.mentions) {
+			return message;
+		}
+
+		const room = await Rooms.findOneById(message.rid);
+		if (!room || isDirectMessageRoom(room) || isRoomFederated(room) || room.t === 'l') {
+			return message;
+		}
+
+		const mentions = message.mentions.filter(({ _id }) => _id !== 'all' && _id !== 'here');
```
MarcosSpessatto, line 69: I think we could invert this with the Rooms.find, this way if there are no mentions we could save a query. WDYT?

apps/meteor/server/modules/core-apps/mention.module.ts
```diff
@@ -0,0 +1,115 @@
...
+			if (!roomPath) {
+				throw new Error('Mention bot - Failed to retrieve path to room');
+			}
+
+			const link = new URL(Meteor.absoluteUrl(roomPath));
+			link.searchParams.set('msg', message._id);
+			const text = `[ ](${link.toString()})`;
+
+			// forwards message to all DMs
+			await processWebhookMessage(
```
MarcosSpessatto, line 93: Any specific reason to use this function instead `sendMessage()`?

apps/meteor/server/modules/core-apps/mention.module.ts
```diff
@@ -0,0 +1,115 @@
...
+			}
+
+			const roomPath = roomCoordinator.getRouteLink(sub.t, { rid: sub.rid, name: sub.name });
+			if (!roomPath) {
+				throw new Error('Mention bot - Failed to retrieve path to room');
+			}
+
+			const link = new URL(Meteor.absoluteUrl(roomPath));
+			link.searchParams.set('msg', message._id);
+			const text = `[ ](${link.toString()})`;
```
MarcosSpessatto, line 90: When mentioning someone in private rooms and selecting the share-message (via DM) option, the user who receives the notification about the mention, when clicking on this link is redirected to a `room not found` page (expected behavior, since they are not part of the private room), I just want to make sure this is the expected UX we want. 😬

apps/meteor/server/modules/core-apps/mention.module.ts
```diff
@@ -0,0 +1,115 @@
...
+		if (!message) {
+			throw new Error('Mention bot - Failed to retrieve message information');
+		}
+
+		if (actionId === 'dismiss') {
+			void api.broadcast('notify.ephemeralMessage', payload.user._id, payload.room, {
+				msg: i18n.t(
+					'You_mentioned___mentions__but_theyre_not_in_this_room',
+					{
+						mentions: `@${usernames.join(', @')}`,
```
MarcosSpessatto, line 48: We could extract this to a variable since it's being used in many places.

apps/meteor/server/modules/core-apps/mention.module.ts
```diff
@@ -0,0 +1,115 @@
...
+		const {
+			actionId,
+			payload: { value: stringifiedMentions, blockId: referenceMessageId },
+		} = payload;
+
+		const mentions = retrieveMentionsFromPayload(stringifiedMentions);
+
+		const usernames = mentions.map(({ username }) => username);
+
+		const message = await Messages.findOneById(referenceMessageId);
```
MarcosSpessatto, line 37: Couldn't we retrieve only the fields we need here?

apps/meteor/server/modules/core-apps/mention.module.ts
```diff
@@ -0,0 +1,115 @@
...
+				),
+				tmid: message.tmid,
+				_id: payload.message,
+				mentions,
+			});
+			return;
+		}
+
+		if (actionId === 'share-message') {
+			const sub = await Subscriptions.findOneByRoomIdAndUserId(payload.room, payload.user._id);
```
MarcosSpessatto, line 77: Couldn't we retrieve only the fields we need here?

MarcosSpessatto (changes requested):

apps/meteor/app/lib/server/startup/mentionUserNotInChannel.ts
gabriellsh: Actually the behaviour changed, since we're no longer adding autmatically. This permission check happens to decide wether we show the "Add them" button on the bot message, so it was actually planned.

apps/meteor/server/modules/core-apps/mention.module.ts
gabriellsh: This function accepts many `rids`, sendMessage only sends one message at a time.

apps/meteor/server/modules/core-apps/mention.module.ts
gabriellsh: So, this was also pointed out from design recently. We introduced a message together with the quote so that it is clearer that you're not part of the room.

Being redirected to a `room not found` is now being dealt as separate bug, since sharing a message through the context menu or copying a message link and sending it in other channels produce the same behaviour.

MarcosSpessatto (approved):

yash-rajpal (approved):

yash-rajpal (approved):

PR #30484, opened 2023-09-25
regression: livechat uiKit actions not working

MartinSchoeler:
OC-1315

### DIFF
--- a/packages/livechat/src/lib/random.ts
+++ b/packages/livechat/src/lib/random.ts
@@ -4,6 +4,6 @@ export const chooseElement = Random.choice;
 
 export const createRandomString = Random._randomString;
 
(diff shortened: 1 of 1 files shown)


packages/livechat/src/lib/random.ts
```diff
@@ -4,6 +4,6 @@ export const chooseElement = Random.choice;
 
 export const createRandomString = Random._randomString;
 
-export const createRandomId = Random.id;
+export const createRandomId = Random.id.bind(Random);
```
ggazzo, line 7: why not ?
```suggestion
export const createRandomId = () => Random.id();
```

aleksandernsilva (approved):

PR #30488, opened 2023-09-25
fix: Max rooms per guest is limited on CE

matheusbsilva137:

 - Do not check the `max-rooms-per-guest` limit when adding users to rooms on CE.

## Issue(s)

Introduced in #26400

## Further comments

SUP-342


yash-rajpal (approved):

DATA-HORIZON:
It should also be mentioned on the project website, that guest users are available in the community edition.
Right now it looks to me, that only Enterprise users have this feature:
https://www.rocket.chat/pricing

image

matheusbsilva137:
Fixed in #30287

PR #30499, opened 2023-09-26
chore: Improve `groups.create` endpoint for large amounts of members

sampaiodiego:

Some key changes:

- changed the signature of `createRoom` function, to receive an `IUser` instead of a `username`. This was to prevent multiple trips to DB being done to fetch data from the same user
- change how subscriptions are being created, now it creates an array of all subscriptions that need to be created and create them all at once using `insertMany`. Before they were being created synchronously one by one
- update members count in the room object in a single command instead of single increments for each user
- update user's `__rooms` property with a single `updateMany` command instead of multiple `updateOne` for each user/member provided

You can see below a load test comparing the old and the new code.

This test was performed by calling the endpoint providing an array with 100 members during 2 minutes by 5 simultaneous virtual users:
image

This one was also provided an array with 100 members but during 5 minutes by 5 simultaneous virtual users. another difference is that the user calling the API was already a member of 70k+ rooms:
image

we can see in both cases an improvement of around 50% either in number of requests or response time.

I still plan to add some tests.. I noticed we don't have tests covering even common use cases.


ggazzo (approved):

apps/meteor/app/lib/server/functions/createRoom.ts
```diff
@@ -21,10 +20,90 @@ const isValidName = (name: unknown): name is string => {
...
+	const memberIds = [];
+
+	const membersCursor = Users.findUsersByUsernames(members, {
+		projection: { 'username': 1, 'settings.preferences': 1, 'federated': 1, 'roles': 1 },
+	});
+
+	for await (const member of membersCursor) {
+		try {
+			await callbacks.run('federation.beforeAddUserToARoom', { user: member, inviter: owner }, room);
+			await callbacks.run('beforeAddedToRoom', { user: member, inviter: owner });
```
pierre-lehnen-rc, line 63: The callback definition expects a full IUser object, but this `member` here is fetched with a projection. Maybe we can change the callback definition to specific what it actually uses?

sampaiodiego: I improved the callback types 😉

pierre-lehnen-rc (approved):

PR #30500, opened 2023-09-26
fix: use setImmediate to handle username update

tapiarafael:

When updating a username with many messages and mentions, the request can result in a TimeOut error since the request can take a long time to be finished.
To prevent this, we're executing the username update inside a `setImmediate` to avoid blocking the process, doing the username update in the background.

## Further comments

[SUP-248](https://rocketchat.atlassian.net/browse/SUP-248)
[ADR-60](https://github.com/RocketChat/adrs/pull/50)

[SUP-248]: https://rocketchat.atlassian.net/browse/SUP-248?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/meteor/tests/end-to-end/api/01-users.js
```diff
@@ -1522,31 +1522,6 @@ describe('[Users]', function () {
...
-						userId: targetUser._id,
-						data: {
-							username: name,
-						},
-					})
-					.expect('Content-Type', 'application/json')
-					.expect(400)
-					.expect((res) => {
-						expect(res.body).to.have.property('success', false);
-						expect(res.body).to.have.property('error', 'Could not save user identity [error-could-not-save-identity]');
```
sampaiodiego, line 1541: I wonder if this validation (and others) should still be done "sync" .. 🤔  because the heavy operations we want to avoid using `setImmediate` are all the many MongoDB updates it does that take some time..

so like if we "accept" the update request, it should be done.. it would be weird if we "accept" the update, return a response to the client, but right after that we figure out the name is invalid..

tapiarafael: I moved the `setImmediate` inside the `saveUserIdentity` and created a flag to decide if the update should run in the background or not to avoid unexpected behavior from other files that use this method.

sampaiodiego (changes requested):
there are more operations on line 71 that should also go to the new function `updateUsernameReferences` - https://github.com/RocketChat/Rocket.Chat/pull/30500/files?diff=split&w=0#diff-3665ef25f844f29569d95ff6d3acd1574e5e3d37707020c980073f08069306f9R71

sampaiodiego (approved):

PR #30508, opened 2023-09-27
fix: seatsCap wrong limit calculation

tapiarafael:

It is impossible to use the total number of license seats. The user is prevented from creating a new user when the seat counter is on the last available seat.
Setting the default value of the counter in the validation to zero instead of one, the user will be able to use the total license seat capacity.

## Steps to test or reproduce
In a workspace with an enterprise license:
1. Create users until you have only one seat left;
2. Try to create a new user to reach the maximum seat capacity;
3. An `error-license-user-limit-reached` error will be shown;

## Further comments

[SUP-364](https://rocketchat.atlassian.net/browse/SUP-364)

[SUP-364]: https://rocketchat.atlassian.net/browse/SUP-364?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/meteor/ee/app/license/server/license.ts
```diff
@@ -223,7 +223,7 @@ class LicenseClass {
 		EnterpriseLicenses.emit('invalidate');
 	}
 
-	async canAddNewUser(userCount = 1): Promise<boolean> {
+	async canAddNewUser(userCount = 0): Promise<boolean> {
```
pierre-lehnen-rc, line 226: This function checks if adding a new user would cause the limit to be crossed, so 1 **is** the right value here. If this is causing issues, then something must be using this function in the wrong way.

pierre-lehnen-rc (changes requested):

apps/meteor/ee/app/license/server/license.ts
pierre-lehnen-rc: The error is inside the function; It should be doing `maxActiveUsers >=`, not `maxActiveUsers >`

tapiarafael:
Fixed by https://github.com/RocketChat/Rocket.Chat/pull/30287

PR #30510, opened 2023-09-27
fix: Multiple BH setup page not loading upon server restart

murtaza98:
## Steps to test or reproduce

Note: this issue is only reproducible when running the server in dev environment using `yarn dsv` command

1.  Set BH to multiple mode
2. Create at-least one custom BH
3. Restart the server from Admin -> Settings -> General -> Restart
4. Upon restart, go to Business hour config page (i.e. Omnichannel ->  Business hour)

Expected Behaviour:
See a list of all custom Business hours created

Actual Behaviour:
You only get to see a single business hour which is the default one


murtaza98:
Closing since this fix is being applied here https://github.com/RocketChat/Rocket.Chat/pull/30565

PR #30522, opened 2023-09-28
fix: Remove model-level query restrictions for monitors

KevLehman:
Release 6.3.8

### DIFF
--- a/apps/meteor/app/livechat/server/lib/Helper.js
+++ b/apps/meteor/app/livechat/server/lib/Helper.js
@@ -483,7 +483,13 @@ export const forwardRoomToDepartment = async (room, guest, transferData) => {
 			throw new Error('error-no-agents-online-in-department');
 		}
 		// if a chat has a fallback department, attempt to redirect chat to there [EE]
-		return !!callbacks.run('livechat:onTransferFailure', { room, guest, transferData });
+		const transferSuccess = !!(await callbacks.run('livechat:onTransferFailure', room, { guest, transferData }));
+
+		// On CE theres no callback so it will return the room
+		if (typeof transferSuccess !== 'boolean' || !transferSuccess) {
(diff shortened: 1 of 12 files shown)


aleksandernsilva (dismissed):

ggazzo (changes requested):
It seems to me that there is more than exactly the changeset describes.
if possible keep the changes punctual as possible.

- It seems that part of it is already fixed in another pull request, wouldn't a cherry-pick for 6.3.9 do the trick?
- hasMonitor role test looks like something "new" in the ui not understand how this is actually a fix
- hard to see exactly where the query restrictions changes really are
- since part of this is a fix being fixed for the second time, would be an automated test an option?
- by any chance of this code affected  is already being tested? looks like an important old mechanism/feature

KevLehman:
3 things:
- Part of the changes were in another PR, to avoid conflicts and simplify life, I replayed those here. Moreover, not all changes from that PR should be migrated. Those changes are also related to how monitor role worked around transcript feature, so was better to have them at a single place for QAing
- The "model-level restrictions" are the overriden update/updateOne/remove/etc methods removed from models. We no longer want to have those. Since that's the "most important" change, that's the one that i included on the changeset
- The monitor change on UI it's one that can be moved to another PR, honestly, I just didn't want to create a PR for a single line since all of them have to go QA+1week review so, packed everything here. I agree we can have a separate PR for it if you think that makes sense.

On the tests side, there's no test for this _specific_ functionality sadly. We never tested the variety of cases the monitor/unit restrictions apply on different endpoints. We want to do that when we add them back on 6.x, as you see, removing them didn't cause any test to fail which is concerning, but expected lol

ggazzo (approved):

aleksandernsilva (approved):

PR #30523, opened 2023-09-28
test: refactor API E2E tests for groups endpoints

sampaiodiego:

I needed to add a few new tests and realised the whole file was being executed in many weird ways. So I refactor it to use `describe` correctly for each endpoint. I also added a few clean up routines but it is still not covering all tests.

Some tests were depending on the setting `E2E_Enable` to be set as `true`, but I don't know how they were "working" since it was not being set anymore. This issue started to happen when I used `describe` for all tests. I wonder if some failures were hidden previsouly.


apps/meteor/tests/end-to-end/api/03-groups.js
```diff
@@ -132,47 +141,63 @@ describe('[Groups]', function () {
 		});
 
 		it('should create the encrypted room by default', async () => {
+			await updateSetting('E2E_Enable', true);
 			await updateSetting('E2E_Enabled_Default_PrivateRooms', true);
```
ggazzo, line 145: is this not a good place to use before(...) ?

apps/meteor/tests/end-to-end/api/03-groups.js
```diff
@@ -132,47 +141,63 @@ describe('[Groups]', function () {
...
+				.expect((res) => {
+					expect(res.body).to.have.property('success', true);
+					expect(res.body).to.have.nested.property('group.name', `default-encrypted-${apiPrivateChannelName}`);
+					expect(res.body).to.have.nested.property('group.t', 'p');
+					expect(res.body).to.have.nested.property('group.msgs', 0);
+					expect(res.body).to.have.nested.property('group.encrypted', true);
+				});
+
+			await updateSetting('E2E_Enable', false);
+			await updateSetting('E2E_Enabled_Default_PrivateRooms', false);
```
ggazzo, line 164: same here but `after` if the test fails it never cleans the flag

ggazzo (approved):

PR #30524, opened 2023-09-29
fix: Ensure user always has an array of roles

wojtek555:

There are lots of places in code where application relies on user's roles. Currently application doesn't guarantee that property `roles` exists in `user` object and that this property is an array, even if it exists. Why? Because application saves user record to database with non-existing `roles` property at first step. Then, in next steps, application updates this record in order to create proper `roles`. These updates can fail either due to admin's error (improper settings regarding default user's roles, for instance) or due to hardware or network errors just after creating user's record.

Therefore most of interactions with `user.roles` nowadays should be secured against non-existing `roles` property and its possibly improper type. Unfortunately it's not done properly in many existing places in code. Instead of continuing that error-prone securing in many places, we should try to ensure that property `roles` always exists in `user`. The best place for doing this is [the method Accounts.onCreateUser()](https://docs.meteor.com/api/accounts-multi.html#AccountsServer-onCreateUser).

### DIFF
--- a/apps/meteor/app/authentication/server/startup/index.js
+++ b/apps/meteor/app/authentication/server/startup/index.js
@@ -187,6 +187,10 @@ const onCreateUserAsync = async function (options, user = {}) {
 		await beforeCreateUserCallback.run(options, user);
 	}
 
+	if (!(user.roles && Array.isArray(user.roles))) {
+		user.roles = [];
+	}
+
 	user.status = 'offline';
 	user.active = user.active !== undefined ? user.active : !settings.get('Accounts_ManuallyApproveNewUsers');
 


wojtek555:
I've just noticed that my fix is similar to the fix done in https://github.com/RocketChat/Rocket.Chat/pull/29969/. However my fix is better, because it works either when user created via `Accounts.insertUserDocAsync()` or via the other method `Accounts.createUserAsync()` (API uses this latter, for instance). BTW I'm not sure whether we need to maintain both those methods for user's creation.

reetp:
I am going to refer this one for comments.

No idea why it is ignored.

PR #30540, opened 2023-10-01
fix: pass message update delta to utilize cache better and resolve less stuff on the server

debdutdeb:


apps/meteor/app/reactions/server/setReaction.ts
```diff
@@ -51,40 +38,62 @@ async function setReaction(room: IRoom, user: IUser, message: IMessage, reaction
...
 
-	if (userAlreadyReacted === shouldReact) {
+	if (shouldReact === userAlreadyReacted) {
 		return;
 	}
 
-	let isReacted;
+	let didReact;
 
 	if (userAlreadyReacted) {
```
matheusbsilva137, line 57:
```suggestion
	if (userAlreadyReacted && message.reactions) {
```
This should be ok to validate + allows us to remove the lint + typecheck errors and see the tests running

apps/meteor/app/reactions/server/setReaction.ts
```diff
@@ -51,40 +38,62 @@ async function setReaction(room: IRoom, user: IUser, message: IMessage, reaction
...
+					await Rooms.setReactionsInLastMessage(room._id, message.reactions);
+				}
 			}
-			await Messages.unsetReactions(message._id);
 		} else {
-			await Messages.setReactions(message._id, message.reactions);
-			if (isTheLastMessage(room, message)) {
-				await Rooms.setReactionsInLastMessage(room._id, message.reactions);
-			}
+			message.reaction[reaction].usernames.splice(userIndexInReactionList, 1);
```
matheusbsilva137, line 89:
```suggestion
			message.reactions[reaction].usernames.splice(userIndexInReactionList, 1);
```

apps/meteor/app/reactions/server/setReaction.ts
```diff
@@ -51,40 +38,62 @@ async function setReaction(room: IRoom, user: IUser, message: IMessage, reaction
...
-		removeUserReaction(message, reaction, user.username as string);
-		if (_.isEmpty(message.reactions)) {
-			delete message.reactions;
-			if (isTheLastMessage(room, message)) {
-				await Rooms.unsetReactionsInLastMessage(room._id);
+
+		if (reactedUsernamesForEmoji?.length === 1) {
+			// we can just pull the whole 'emoji' field
+			// @ts-ignore
+			if (message.reactions && Object.keys(message.reactions).length === 0) {
```
matheusbsilva137, line 67:
```suggestion
			if (message.reactions && Object.keys(message.reactions).length === 1) {
```
Isn't that the case for when the user removed the last reaction of a message (and no reaction is left after that)? Since this happens before the last reaction is removed, this should be `1`, or am I missing something?

matheusbsilva137 (changes requested):

apps/meteor/app/reactions/server/setReaction.ts
debdutdeb: you're right .. i had the key delete above this statement before, but then moved

apps/meteor/app/reactions/server/setReaction.ts
```diff
@@ -51,40 +38,62 @@ async function setReaction(room: IRoom, user: IUser, message: IMessage, reaction
...
+				// remove the whole reactions
+				delete message.reactions;
+				await Messages.updateOne({ _id: message._id }, { $unset: { reactions: 1 } });
+
+				if (isTheLastMessage(room, message)) {
+					await Rooms.unsetReactionsInLastMessage(room._id);
+				}
+			} else {
+				delete message.reactions[reaction];
+				await Messages.updateOne(
```
debdutdeb, line 77:
```suggestion
			if (message.reactions && Object.keys(message.reactions).length === 1) {
				// remove the whole reactions
				delete message.reactions;
				await Messages.updateOne({ _id: message._id }, { $unset: { reactions: 1 } });

				if (isTheLastMessage(room, message)) {
					await Rooms.unsetReactionsInLastMessage(room._id);
				}
			} else {
				// @ts-ignore
				delete message.reactions[reaction];
				await Messages.updateOne(
```

PR #30542, opened 2023-10-01
fix:  query paramters can be passed 

### LINKED ISSUE #25300: Fields Query Parameters not working on livechat rooms list endpoint
### Description:

Fields Query Parameters not working on livechat rooms list endpoint

### Steps to reproduce:

1. Send request with query or fields parameter on endpoint '/api/v1/livechat/rooms'
2. Returned all rooms without apply query or fields

### Expected behavior:

Rooms can be filtered by fields

### Actual behavior:
image

### Server Setup Information:

- Version of Rocket.Chat Server:  3.14.0
- Operating System:  Windows 11
- Number of Running Instances:  1

lokesh-wagh:

there are two files that are changed
apps/meteor/app/livechat/imports/server/rest/rooms.ts
and
apps/meteor/app/livechat/server/api/lib/rooms.ts

i have got the query paramter
then converted it into a parsedQ  object to be passed into the findRooms function
inside the defination of findRooms function i have recieved the object as customQuery 
they are then destructured and passed to the findRoomsWithCriteria function
 

## Issue(s)

#25300 

## Steps to test or reproduce

1. call the endpoint api/v1/livechat/rooms
2. pass a query object as parameter

## Further comments

closes #25300

### DIFF
--- a/apps/meteor/app/livechat/imports/server/rest/rooms.ts
+++ b/apps/meteor/app/livechat/imports/server/rest/rooms.ts
@@ -6,6 +6,9 @@ import { getPaginationItems } from '../../../../api/server/helpers/getPagination
 import { hasPermissionAsync } from '../../../../authorization/server/functions/hasPermission';
 import { findRooms } from '../../../server/api/lib/rooms';
 
+import { IOmnichannelRoom } from '@rocket.chat/core-typings';
+import type {Filter} from 'mongodb'
+
 const validateDateParams = (property: string, date?: string) => {
 	let parsedDate: { start?: string; end?: string } | undefined = undefined;
 	if (date) {
@@ -32,6 +35,9 @@ API.v1.addRoute(
 			const { sort, fields } = await this.parseJsonQuery();
 			const { agents, departmentId, open, tags, roomName, onhold } = this.queryParams;
 			const { createdAt, customFields, closedAt } = this.queryParams;
+			
+			const {query}=this.queryParams;
+			
 
 			const createdAtParam = validateDateParams('createdAt', createdAt);
 			const closedAtParam = validateDateParams('closedAt', closedAt);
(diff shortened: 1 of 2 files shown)


lokesh-wagh:
sorry for the PR title error

KevLehman:
Hey, thanks for contributing :)

Wanna share a bit more of context on what's your usecase so we can review this better? 
Opening custom queries can be dangerous and so we try to do them only when necessary

lokesh-wagh:
In the documentation of the api it say's that query parameters are supported
but there was no way to pass a query parameter so I added one

link to the documentation  : https://developer.rocket.chat/reference/api/rest-api/endpoints/omnichannel/livechat-endpoints/livechat-rooms/livechat-rooms-list

a normal usecase would be to query the livechat rooms based on how the field's 

example: if you want livechat room's with certain name's and a certain amount of message's

image

just like the documentation the query is applied and the found document's are returned

in the issue #25300 it also mention's that the api doesn't function according to the documentation
this PR fixes that

KevLehman:
I don't think we ever supported a `query` parameter like that 👀 (or at least, i cannot remember). What the endpoint supports is passing individual query parameters (URL query parameters, not mongo queries) to the endpoint in order to filter. For example:
`livechat/rooms?open=true` will return the rooms where `room.open === true`. If you want to filter by `fname`, we have the `roomName` query param that you can send in the URL.

We don't support filtering by `msgs` property as we haven't found a proper usecase for it. Maybe yours is the first one? 👀

lokesh-wagh:
Maybe I have misunderstood but in the documentation it say's that this endpoint support's query and field parameters are accepted

link to the endpoint : https://developer.rocket.chat/reference/api/rest-api/endpoints/omnichannel/livechat-endpoints/livechat-rooms/livechat-rooms-list
image
link to query and field's : https://developer.rocket.chat/reference/api/rest-api#query-and-fields
image

Can you please explain to me what these parameter's are so that I can avoid such mistake in future?

KevLehman:
Hey! It says there in the documentation: "some endpoints in Rocket.Chat API". Not every endpoint supports those fields. 

In the case of Omnichannel, I don't remember any endpoint that supports receiving and parsing an arbitrary query object.

General rule is: if an endpoint documentation doesn't explicitly tell that `query` is supported, it's almost certain is not supported.

For fields/sort, those are supported by the majority of endpoints that return data. There should be also a mention of those on the endpoint documentation so you're sure how to use them

If that solves your question, maybe we can close this PR? We generally try to avoid endpoints with user-defined queries for security :)

lokesh-wagh:
Also can you tell me what  issue #25300 actually want's so that i can do that and my work doesn't goes to waste

I am actually a begineer and really want to contribute to this project

PR #30546, opened 2023-10-02
chore: add the new endpoints to sync with cloud

ggazzo:

### DIFF
--- a/apps/meteor/app/cloud/server/functions/getWorkspaceLicense.ts
+++ b/apps/meteor/app/cloud/server/functions/getWorkspaceLicense.ts
@@ -51,6 +51,10 @@ const fetchCloudWorkspaceLicensePayload = async ({ token }: { token: string }): 
 
 export async function getWorkspaceLicense(): Promise<{ updated: boolean; license: string }> {
 	const currentLicense = await Settings.findOne('Cloud_Workspace_License');
+	// it should never happen, since even if the license is not found, it will return an empty settings
+	if (!currentLicense?._updatedAt) {
+		throw new CloudWorkspaceLicenseError('Failed to retrieve current license');
+	}
 
 	const fromCurrentLicense = async () => {
 		const license = currentLicense?.value as string | undefined;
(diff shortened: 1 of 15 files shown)


apps/meteor/app/cloud/server/functions/getWorkspaceLicense.ts
```diff
@@ -51,11 +51,15 @@ const fetchCloudWorkspaceLicensePayload = async ({ token }: { token: string }):
...
+	// it should never happen, since even if the license is not found, it will return an empty settings
+	if (!currentLicense?._updatedAt) {
+		throw new CloudWorkspaceLicenseError('Failed to retrieve current license');
+	}
 
 	const fromCurrentLicense = async () => {
 		const license = currentLicense?.value as string | undefined;
 		if (license) {
-			callbacks.run('workspaceLicenseChanged', license);
+			await callbacks.run('workspaceLicenseChanged', license);
```
tassoevan, line 62: Are you sure `await` is necessary, since `workspaceLicenseChanged` is listed under `EventLikeCallbackSignatures`?

tassoevan (changes requested):

apps/meteor/app/cloud/server/functions/supportedVersionsToken/supportedVersionsToken.ts
```diff
@@ -10,6 +10,12 @@ import { settings } from '../../../../settings/server';
 import { generateWorkspaceBearerHttpHeader } from '../getWorkspaceAccessToken';
 import { supportedVersionsChooseLatest } from './supportedVersionsChooseLatest';
 
+declare module '@rocket.chat/license' {
+	interface ILicenseV3 {
```
tassoevan, line 14: `question` Why this augmentation is needed?

ggazzo: so far we dont have decided if the supported version is going to be added to the license, also I'm still considering where the types should live: cloud/license/supportedVersions

apps/meteor/app/cloud/server/functions/getWorkspaceLicense.ts
ggazzo: hm not totally sure, but all callbacks.run returns a promise, if not we should use runAsync

tassoevan (approved):

PR #30550, opened 2023-10-02
fix: Remove monitors query restrictions on update

KevLehman:


apps/meteor/ee/app/livechat-enterprise/server/methods/getUnitsFromUserRoles.ts
```diff
@@ -17,7 +17,30 @@ async function getUnitsFromUserRoles(user: string | null): Promise<string[] | un
 	return LivechatUnit.findByMonitorId(user);
 }
 
+async function getDepartmentsFromUserRoles(user: string | null): Promise<string[] | undefined> {
```
sampaiodiego, line 20: maybe this is a change for another time, but looks like it doesn't make much sense the signature from the functions of this file to allow `user` to be `null`.. since they do nothing when that is the case..

I understand changing this for a patch may introduce a lot of undesired changes, but can you pls take a look into this in the near future?

KevLehman: Hey. This is only `string | null` since that's the type Meteor returns from `Meteor.userId`. I can move the validation to the meteor method and let this be just "string" if that's the problem.

Validation in both places will have the same behavior tho, both cases it will just do nothing, disregard of where the `if` is.

sampaiodiego: IMO this is much better.. I mean, for the scope of this function `user` **is** actually required.. having the signature right makes that clear..

I just saw this one is new, so yeah, having the test in the Meteor method is better.

but for the other ones like `getUnitsFromUser` , I see it's been used in many places so changing it may cause other unrelated changes..

KevLehman: No problems there. Tried and it worked well here

apps/meteor/ee/app/livechat-enterprise/server/hooks/applyDepartmentRestrictions.ts
```diff
@@ -25,7 +27,7 @@ callbacks.add(
 		}
 
 		cbLogger.debug('Applying department query restrictions');
-		return addQueryRestrictionsToDepartmentsModel(originalQuery);
+		return addQueryRestrictionsToDepartmentsModel(originalQuery, userId);
```
sampaiodiego, line 30: should a `userId` be provided on the line below as well?

https://github.com/RocketChat/Rocket.Chat/blob/32c6b0b308336f6db68aad35c16c5c2224865807/apps/meteor/app/livechat/server/lib/Livechat.js#L288

I mean, even though `userId` is optional in this callback, it doesn't do anything if it is not provided..

sampaiodiego (changes requested):

sampaiodiego (approved):

aleksandernsilva (approved):

PR #30554, opened 2023-10-03
feat: Introduce User Report Section on Moderation Console

Dnouv:

Introduce a new window (tab, to be precise) on the Moderation Console to view and manage reported users. 

**Added:**

- A new tab for Reported Users is now available on the Moderation Console page.
- A new endpoint has been added to retrieve user reports by user ID.
- A new endpoint has been added to dismiss user reports by user ID.
- A new endpoint has been added to fetch all the user reports grouped by unique users.
- A new component has been added to show user report details on the Contextual Bar.
- A new common filter component `ModFilter` can be used in both the User & Message Reports.
- A few `i18n` have been added.

**Changed:**

- Replaced `useRoute` with `useRouter` on the Moderation Console page.
- Added reported user name and username columns to the Moderation Console table.
- Shortened long file names, e.g., "ModConsoleReportDetails" instead of "ModerationConsoleReportDetails".
- Improved the no results page view with more informative messages for each tab.
- Changed the Date filters to a dropdown with pre-defined options, rather than a date selector, for better i18n.
- Added an `isUserReport` prop to the `useDismissUserAction` to enable dismissing user reports as well as message reports.
- Updated the `useResetAvatarAction` with more calls to invalidate react-queries.
- Renamed the `ReportMessageHistoryParams` to `GetUserReportsParams`, for more clarity.

## APIs
 
<details>
<summary>API Endpoint: moderation.dismissUserReports</summary>

### Description
This endpoint allows a moderator to dismiss user reports. The moderator could provide a reason and an action for the dismissal.

### Parameters
- **userId**: The ID of the user whose reports are to be dismissed.
- **reason**: The reason for dismissing the reports.
- **action**: The action taken upon dismissing the reports.

### Response
- **success**: A boolean indicating if the operation was successful.
- **message**: A message indicating the result of the operation.
</details>

<details>
<summary>API Endpoint: moderation.user.reportsByUserId</summary>

### Description
This endpoint retrieves all the moderation reports for a specific user.

### Query Parameters
- **userId**: The ID of the user whose reports are to be retrieved.
- **selector**: (Optional) A string to filter the reports.
- **count**: (Optional) The number of reports to retrieve. Default is 50.
- **offset**: (Optional) The number of reports to skip before starting to retrieve. Default is 0.

### Response
- **success**: A boolean indicating if the operation was successful.
- **user**: The user object.
- **reports**: An array of report objects.
- **count**: The number of reports returned.
- **total**: The total number of reports.
- **offset**: The offset value.
</details>

<details>
<summary>API Endpoint: moderation.userReports</summary>

### Description
This endpoint retrieves all the user reports.

### Query Parameters
- **latest**: (Optional) The latest date (in ISO format) from which to retrieve reports. Default is the current date.
- **oldest**: (Optional) The oldest date (in ISO format) from which to retrieve reports. Default is the date of the oldest report.
- **selector**: (Optional) A string to filter the reports.
- **count**: (Optional) The number of reports to retrieve. Default is 20.
- **offset**: (Optional) The number of reports to skip before starting to retrieve. Default is 0.

### Response
- **success**: A boolean indicating if the operation was successful.
- **reports**: An array of report objects.
- **count**: The number of reports returned.
- **total**: The total number of reports.
- **offset**: The offset value.
</details>

## Steps to test or reproduce

This video shows the main features of the product:

https://github.com/RocketChat/Rocket.Chat/assets/61188295/b2ee6963-4ce5-40c5-93cb-0bceaa408533

## Further comments

[AECO-336](https://rocketchat.atlassian.net/browse/AECO-336)

To the reviewer:

> This PR might look a bit large at first glance, and you might think it could be split into smaller ones. However, from my experience with similar changes in the past, I’ve found that keeping everything in one PR actually makes the review and testing process easier. 
That being said, I understand this might not be the most convenient approach for everyone. So, my apologies if this causes any inconvenience during your review. Thanks for your understanding and patience as you go through this PR. 🙏

[AECO-336]: https://rocketchat.atlassian.net/browse/AECO-336?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/meteor/app/api/server/v1/moderation.ts
```diff
@@ -113,6 +161,61 @@ API.v1.addRoute(
...
+				projection: {
+					_id: 1,
+					username: 1,
+					name: 1,
+					avatarETag: 1,
+					active: 1,
+					roles: 1,
+					emails: 1,
+					createdAt: 1,
+				},
```
debdutdeb, line 189: what defined these fields need to be exposed? also, we have view-other-user-info, IMO we should respect that permission here too.

apps/meteor/app/api/server/v1/moderation.ts
```diff
@@ -113,6 +161,61 @@ API.v1.addRoute(
...
+
+			if (user) {
+				user.emails = Array.from(emailSet.values());
+			}
+
+			return API.v1.success({
+				user,
+				reports,
+				count: reports.length,
+				total: totalCount,
```
debdutdeb, line 212: totalCount is a promise iirc.

debdutdeb (changes requested):

apps/meteor/app/api/server/v1/moderation.ts
Dnouv: Thank you for the comment, Deb!

> what defined these fields need to be exposed? 

Nothing precisely; it is just the info we must show on the Context Bar.

> we have view-other-user-info, IMO we should respect that permission here too.

Sure, we could do that, but should we? Because viewing the Moderation Console already requires permission, if an admin doesn't want the reported user details to be viewed by specific role users, they won't give permission to view the Moderation Console. 
Additionally, I think the details on MC, apart from the User info, are already quite sensitive. Thoughts?

apps/meteor/app/api/server/v1/moderation.ts
Dnouv: My bad; I will correct this while working on improving the total count. I plan on making it, equal to the total number of distinct users. Currently, it is shown as the total number of reports. But, it's still pending and might change.

AllanPazRibeiro (changes requested):

apps/meteor/app/api/server/v1/moderation.ts
```diff
@@ -63,11 +63,59 @@ API.v1.addRoute(
...
 
+API.v1.addRoute(
+	'moderation.userReports',
+	{
+		authRequired: true,
+		validateParams: isReportHistoryProps,
+		permissionsRequired: ['view-moderation-console'],
+	},
+	{
+		async get() {
```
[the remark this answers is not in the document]
Dnouv: Thank you for the review, Allan. I added separate helpers for two API calls, which had some extended logic, and there are error handlers.

AllanPazRibeiro (approved):
LGTM

apps/meteor/server/models/raw/ModerationReports.ts
```diff
@@ -24,6 +25,8 @@ export class ModerationReportsRaw extends BaseRaw<IModerationReport> implements
 			{ key: { 'message.rid': 1, 'ts': 1 } },
 			{ key: { userId: 1, ts: 1 } },
 			{ key: { 'message._id': 1, 'ts': 1 } },
+			{ key: { 'ts': 1, '_hidden': 1, 'message.u._id': 1 } },
+			{ key: { '_hidden': 1, 'reportedUser._id': 1, 'ts': 1 } },
```
debdutdeb, line 29: pls look into optimizing the indexes a bit, there are a lot for a rather small feature. for example, index \#3 and the last one could be compounded i believe. maybe more.

apps/meteor/server/models/raw/ModerationReports.ts
```diff
@@ -132,12 +135,97 @@ export class ModerationReportsRaw extends BaseRaw<IModerationReport> implements
...
+	}
+
 	countMessageReportsInRange(latest: Date, oldest: Date, selector: string): Promise<number> {
-		return this.col.countDocuments({
-			_hidden: { $ne: true },
-			ts: { $lt: latest, $gt: oldest },
-			...this.getSearchQueryForSelector(selector),
-		});
+		return this.col
+			.distinct('message.u._id', {
```
debdutdeb, line 200: this means "unique reports for a message sender" - so 2 different messages sent by the same user would count as one. should this not be all reports?

apps/meteor/server/models/raw/ModerationReports.ts
```diff
@@ -132,12 +135,97 @@ export class ModerationReportsRaw extends BaseRaw<IModerationReport> implements
...
+					$gt: oldest,
+				},
+				...this.getSearchQueryForSelector(selector),
+			})
+			.then((ids) => ids.length);
+	}
+
+	countUserReportsInRange(latest: Date, oldest: Date, selector: string): Promise<number> {
+		return this.col
+			.distinct('reportedUser._id', {
```
debdutdeb, line 215: this also i don't understand. counting user reports means all user reports .. not unique ones by reporter. if I report two users that will count to one?

debdutdeb (changes requested):
couple of comments left.

1 more point i'd like to hear your thoughts - for component names, why not use full intent, `ModerationFilter` instead of `ModFilter` - why shorten it? Can make it hard to comprehend for new readers 

didn't read all of the code, it was quite large. tried to put as much feedback as i could now

apps/meteor/client/views/admin/moderation/helpers/UserProfile.tsx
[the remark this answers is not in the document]
Dnouv: Thank you for the review. Yes, they are similar now and can be merged, I will do this. 

I kept it separate because, in the next iteration, there will be an additional active status. And we will have to separate it. So I decided to keep it different.

Dnouv (commented):
Thank you for the review.

> for component names, why not use full intent, ModerationFilter instead of ModFilter - why shorten it? Can make it hard to comprehend for new readers

I am also trying to get all the previous file names to `Mod` (maybe in the next PR). Because the names are getting quite long, and as for the confusion for new readers, all of this lives under the `moderation` folder. So, it shouldn't be an issue.

apps/meteor/app/api/server/v1/moderation.ts
```diff
@@ -63,11 +65,30 @@ API.v1.addRoute(
...
+			try {
+				const result = await getUserReports(this.queryParams, this.parseJsonQuery.bind(this));
```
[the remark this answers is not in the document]
debdutdeb: it's context transfer that is *costlier* than just passing what's required in form of primitives.

But that's not my worry. It's just .. weird :) It's unconventional in our codebase and with no real benefit. You's not really reducing the lines of code, as the function as to be called somewhere anyway, not that lines of code should matter IMO 

if you think making it pass queries as params doesn't give the function much value, then 100 percent would recommend removing the function and doing it as it is inside the handler.

apps/meteor/app/api/server/v1/moderation.ts
```diff
@@ -196,6 +236,33 @@ API.v1.addRoute(
...
+			const { userId, reason, action: actionParam } = this.bodyParams;
+
+			const sanitizedReason: string = reason ?? 'No reason provided';
+			const action: string = actionParam ?? 'None';
+
+			const { userId: moderatorId } = this;
+
+			if (!userId) {
+				return API.v1.failure('error-user-id-param-not-provided');
+			}
```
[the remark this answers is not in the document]
debdutdeb: userId could be checked right after the destructure, start-end together. 

If you're written go yet it's similar, we always return at the earliest, because of its design. Same thing here, as a nitpick

apps/meteor/client/views/admin/moderation/ModConsoleReportDetails.tsx
```diff
@@ -0,0 +1,45 @@
...
+};
+
+const tabs = ['Messages', 'Users'];
+
+const ModConsoleReportDetails = ({ userId, default: defaultTab, onRedirect }: ModConsoleReportDetailsProps): JSX.Element => {
+	const t = useTranslation();
+	const [tab, setTab] = useState<string>(defaultTab);
+	const moderationRoute = useRouter();
+
+	const handleTabClick = useMemo(() => (tab: string) => (): void => setTab(tab), [setTab]);
```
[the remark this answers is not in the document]
debdutdeb: I'd love to read it ! I don't play with react, been absorbing information like this :p please, if you find it, send it over to me.

apps/meteor/client/views/admin/moderation/ModerationConsolePage.tsx
```diff
@@ -1,37 +1,52 @@
...
+	const handleRedirect = useCallback(
+		async (mid: string) => {
+			try {
+				const permalink = await getPermaLink(mid);
+				window.open(permalink, '_self');
+			} catch (error) {
+				dispatchToastMessage({ type: 'error', message: error });
+			}
+		},
+		[dispatchToastMessage],
```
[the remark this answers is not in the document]
debdutdeb: I see. Thanks for explaining 🤗

apps/meteor/server/lib/moderation/getUserReportsByUid.ts
```diff
@@ -0,0 +1,55 @@
...
+
+		const [reports, total] = await Promise.all([cursor.toArray(), totalCount]);
+
+		const emailSet = new Map<IUserEmail['address'], IUserEmail>();
+
+		reports.flatMap((report) => report.reportedUser?.emails ?? []).forEach((email) => emailSet.set(email.address, email));
+
+		if (user) {
+			user.emails = Array.from(emailSet.values());
+		}
```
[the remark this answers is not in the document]
debdutdeb: Ah good point. Right because we keep snapshots of the records the emails can potentially be different. Nice.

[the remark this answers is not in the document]
debdutdeb: But I'm also thinking then this could be a single loop, could it not?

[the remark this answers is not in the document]
debdutdeb: because emails will always be length 1, we can treat it as a string and iterate over just once without the need for flatmap

apps/meteor/server/models/raw/ModerationReports.ts
debdutdeb: Please rename the model function then. The method there is not saying this :)

apps/meteor/server/models/raw/ModerationReports.ts
debdutdeb: Ok, same thing as before. Please update the method name. Because now it isn't really counting user reports - user reports mean ALL individual reports for any user. Rather something like `totalUsersReportedCountInRange` (mind the `ed`).

apps/meteor/app/api/server/v1/moderation.ts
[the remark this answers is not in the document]
Dnouv: Thank you. Looked up more on this and found the performance impact would be negligible in context transfer. 

I will leave this as a function; IIRC, in a few previous encounters, I got feedback about keeping this in separate functions for better code organization. But it was also mentioned that this is a unique solution, so I held on until I reached a threshold of the same suggestion and finally added it here.

Thank you for being so understanding. 🙏

apps/meteor/server/lib/moderation/getUserReportsByUid.ts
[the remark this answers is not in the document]
debdutdeb: empty emails can be for bots or app users i think .. not for normal users. the list type is meteor/accounts thing, the schema "supports" multiple emails, we don't. afaik.

apps/meteor/server/lib/moderation/getUserReportsByUid.ts
```diff
@@ -0,0 +1,55 @@
...
+		return {
+			user,
+			reports,
+			count: reports.length,
+			total,
+			offset,
+		};
+	} catch (error) {
+		throw new Error('Error while fetching users reports');
+	}
```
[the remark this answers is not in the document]
Dnouv: Thank you, done!

apps/meteor/client/views/admin/moderation/ModConsoleReportDetails.tsx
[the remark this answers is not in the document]
Dnouv: Thank you, done!

apps/meteor/server/lib/moderation/getUserReports.ts
```diff
@@ -0,0 +1,48 @@
...
+
+		return {
+			reports,
+			count: reports.length,
+			offset,
+			total,
+		};
+	} catch (error) {
+		throw new Error('Error while fetching user reports');
+	}
```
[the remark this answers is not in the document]
Dnouv: Thank you, done!

apps/meteor/server/models/raw/ModerationReports.ts
Dnouv: Thank you, done!

apps/meteor/server/models/raw/ModerationReports.ts
Dnouv: Thank you, done!

apps/meteor/server/lib/moderation/getUserReportsByUid.ts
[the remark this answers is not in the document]
Dnouv: I just tried out some benchmarks (assuming data with either 1 or 0 emails),

While processing 1 million users:
```
flatMap: 2.708s
new: 2.008s
```

For our use-case, 100 users:
```
flatMap: 0.114ms
new: 0.07ms
```

That's amazing; we saved ~0.044 ms. Made, the changes, marking them as done.

apps/meteor/client/views/admin/moderation/helpers/UserColumn.tsx
```diff
@@ -0,0 +1,45 @@
...
+	name?: string;
+	username?: string;
+	isDesktopOrLarger?: boolean;
+	isProfile?: boolean;
+};
+
+const UserColumn = ({ name, username, isDesktopOrLarger, isProfile }: UserColumnProps) => {
+	let size: 'x40' | 'x20' | 'x48' = 'x40';
+	if (isProfile) size = 'x48';
+	else if (isDesktopOrLarger) size = 'x20';
```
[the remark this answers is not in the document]
Dnouv: Thank you so much for the review.
 
Haha, I did it initially this way, but you know the lint said I can't use nested ternary `no-nested-ternary` 😂, so had to do it the other way.

apps/meteor/server/lib/moderation/getUserReportsByUid.ts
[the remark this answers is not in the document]
debdutdeb: not worried about performance here, given we know it's gonna be constant length of 1. it's just constructs that adds so even at large scale (like ours, not millions large) shouldn't be that bad with flatmap or something. my point here was for readability. if you can make it clear that the length is 1 expected, that makes it clear what the expectation is from the list. While right now (ike i was also confused at first) it's like we are expecting more than one emails to be associated with a user account.

apps/meteor/client/views/admin/moderation/MessageContextFooter.tsx
dougfabris, line 1: There's no need to `aria-label` and `title` for Buttons with labels

apps/meteor/client/views/admin/moderation/MessageContextFooter.tsx
```diff
@@ -16,7 +16,7 @@ const MessageContextFooter: FC<{ userId: string; deleted: boolean }> = ({ userId
 	const deleteMessagesAction = useDeleteMessagesAction(userId);
 
 	return (
-		<ButtonGroup width='full' stretch>
+		<ButtonGroup flexGrow={0} width='full' stretch>
```
dougfabris, line 19: Doing that you will remove the spacing logic in `ButtonGroup`

apps/meteor/client/views/admin/moderation/ModConsoleReportDetails.tsx
```diff
@@ -0,0 +1,45 @@
...
+
+type ModConsoleReportDetailsProps = {
+	userId: IUser['_id'];
+	default: string;
+	onRedirect: (mid: string) => void;
+};
+
+const tabs = ['Messages', 'Users'];
+
+const ModConsoleReportDetails = ({ userId, default: defaultTab, onRedirect }: ModConsoleReportDetailsProps): JSX.Element => {
```
dougfabris, line 18:
```suggestion
const ModConsoleReportDetails = ({ userId, default: defaultTab, onRedirect }: ModConsoleReportDetailsProps) => {
```

apps/meteor/client/views/admin/moderation/ModConsoleReportDetails.tsx
```diff
@@ -0,0 +1,45 @@
...
+
+	return (
+		<Contextualbar>
+			<ContextualbarHeader>
+				<ContextualbarTitle>{t('Reports')}</ContextualbarTitle>
+				<ContextualbarClose onClick={() => moderationRoute.navigate('/admin/moderation', { replace: true })} />
+			</ContextualbarHeader>
+			<Tabs>
+				{tabs.map((tabName) => (
+					<TabsItem key={tabName || ''} selected={tab === tabName} onClick={handleTabClick(tabName)}>
```
dougfabris, line 33:
```suggestion
					<TabsItem key={tabName || ''} selected={tab === tabName} onClick={() => setTab(tabName)}>
```

apps/meteor/client/views/admin/moderation/ModerationConsolePage.tsx
```diff
@@ -1,37 +1,52 @@
...
 
 	return (
 		<Page flexDirection='row'>
 			<Page>
 				<Page.Header title={t('Moderation')} />
 				<Page.Content>
-					<ModerationConsoleTable />
+					<Tabs>
+						{tabs.map((tabName) => (
+							<TabsItem key={tabName || ''} selected={tab === tabName} onClick={handleTabClick(tabName)}>
```
dougfabris, line 41:
```suggestion
							<TabsItem key={tabName || ''} selected={tab === tabName} onClick={() => setTab(tabName)}>
```

apps/meteor/client/views/admin/moderation/ModerationConsolePage.tsx
```diff
@@ -1,37 +1,52 @@
...
-	const handleRedirect = async (mid: string) => {
-		try {
-			const permalink = await getPermaLink(mid);
-			// open the permalink in same tab
-			window.open(permalink, '_self');
-		} catch (error) {
-			dispatchToastMessage({ type: 'error', message: error });
-		}
-	};
+	const handleTabClick = (tab: string) => () => setTab(tab);
```
dougfabris, line 32:
```suggestion

```

apps/meteor/client/views/admin/moderation/ModConsoleReportDetails.tsx
```diff
@@ -0,0 +1,45 @@
...
+};
+
+const tabs = ['Messages', 'Users'];
+
+const ModConsoleReportDetails = ({ userId, default: defaultTab, onRedirect }: ModConsoleReportDetailsProps): JSX.Element => {
+	const t = useTranslation();
+	const [tab, setTab] = useState<string>(defaultTab);
+	const moderationRoute = useRouter();
+
+	const handleTabClick = (tab: string) => (): void => setTab(tab);
```
dougfabris, line 23:
```suggestion

```

apps/meteor/client/views/admin/moderation/UserReports/ModConsoleUserActions.tsx
```diff
@@ -0,0 +1,41 @@
+import { useTranslation } from '@rocket.chat/ui-contexts';
+import React from 'react';
+
+import GenericMenu from '../../../../components/GenericMenu/GenericMenu';
+import useDeactivateUserAction from '../hooks/useDeactivateUserAction';
+import useDismissUserAction from '../hooks/useDismissUserAction';
+import useResetAvatarAction from '../hooks/useResetAvatarAction';
+import type { ModConsoleUserRowProps } from './ModConsoleUserTableRow';
+
+const ModConsoleUserActions = ({ report, onClick }: Omit<ModConsoleUserRowProps, 'isDesktopOrLarger'>): JSX.Element => {
```
dougfabris, line 10:
```suggestion
const ModConsoleUserActions = ({ report, onClick }: Omit<ModConsoleUserRowProps, 'isDesktopOrLarger'>) => {
```

apps/meteor/client/views/admin/moderation/UserReports/ModConsoleUsersTable.tsx
```diff
@@ -0,0 +1,161 @@
...
+		),
+		500,
+	);
+
+	const getReports = useEndpoint('GET', '/v1/moderation.userReports');
+
+	const dispatchToastMessage = useToastMessageDispatch();
+
+	const { data, isLoading, isSuccess, isError, refetch } = useQuery(['moderation.userReports', query], async () => getReports(query), {
+		onError: (error) => {
```
dougfabris, line 58: Avoid dispatching toast error on request, since you're already rendering a state error

apps/meteor/client/views/admin/moderation/UserReports/ModConsoleUsersTable.tsx
```diff
@@ -0,0 +1,161 @@
...
+			dispatchToastMessage({ type: 'error', message: error });
+		},
+	});
+
+	const handleRefetch = useMutableCallback(() => {
```
dougfabris, line 63: There's no need to this function, just call refetch when needed

apps/meteor/server/lib/moderation/getUserReportsByUid.ts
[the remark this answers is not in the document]
Dnouv: Ah, I misunderstood, but I think it is good to save on performance, although it is little sometimes, reiterating your point, what if the system is already at its capacity, stacking these improvements, really comes as critical.

apps/meteor/client/views/admin/moderation/UserReports/ModConsoleUsersTable.tsx
```diff
@@ -0,0 +1,161 @@
...
+		router.navigate({
+			name: 'moderation-console',
+			params: {
+				context: 'info',
+				id,
+			},
+		});
+	});
+
+	const headers = useMemo(
```
dougfabris, line 77: return an `ReactElement` instead

apps/meteor/client/views/admin/moderation/UserReports/UserReportInfo.tsx
```diff
@@ -0,0 +1,158 @@
...
+import React, { useMemo } from 'react';
+
+import GenericNoResults from '../../../../components/GenericNoResults';
+import UserCard from '../../../../components/UserCard';
+import { useFormatDate } from '../../../../hooks/useFormatDate';
+import ReportReason from '../helpers/ReportReason';
+import UserColumn from '../helpers/UserColumn';
+import UserContextFooter from './UserContextFooter';
+
+const isUser = (obj: any): obj is IUser => {
```
dougfabris, line 28: ???

apps/meteor/client/views/admin/moderation/UserReports/UserReportInfo.tsx
```diff
@@ -0,0 +1,158 @@
...
+	if (obj?.createdAt && typeof obj.createdAt === 'string') {
+		obj.createdAt = new Date(obj.createdAt);
+	}
+	return obj?.createdAt instanceof Date;
+};
+
+const UserReportInfo = ({ userId }: { userId: string }): JSX.Element => {
```
dougfabris, line 35:
```suggestion
const UserReportInfo = ({ userId }: { userId: string }) => {
```

apps/meteor/client/views/admin/moderation/UserReports/UserReportInfo.tsx
```diff
@@ -0,0 +1,158 @@
...
+	const {
+		data: report,
+		refetch: reloadUsersReports,
+		isLoading: isLoadingUsersReports,
+		isSuccess: isSuccessUsersReports,
+		isError,
+		dataUpdatedAt,
+	} = useQuery(
+		['moderation.usersReport', { userId }],
+		async () => {
```
dougfabris, line 50: just call getUsersReports, no need to instantiate a const and return it

apps/meteor/client/views/admin/moderation/UserReports/UserReportInfo.tsx
```diff
@@ -0,0 +1,158 @@
...
+			const reports = await getUserReports({ userId });
+			return reports;
+		},
+		{
+			onError: (error) => {
```
dougfabris, line 55: No need to this since you're rendering the error on the UI

apps/meteor/client/views/admin/moderation/helpers/ModerationFilter.tsx
```diff
@@ -0,0 +1,30 @@
...
+	setDateRange: (dateRange: { start: string; end: string }) => void;
+};
+
+const ModerationFilter = ({ setText, setDateRange }: ModerationFilterProps) => {
+	const t = useTranslation();
+
+	const handleChange = useCallback(({ text }): void => setText(text), [setText]);
+
+	return (
+		<Grid>
```
dougfabris, line 19: Avoid using `Grid` its you can pass `DataRangePicker` as children of FilterBytText the have them inline and customize the size if wanted

apps/meteor/client/views/admin/moderation/ModerationConsolePage.tsx
```diff
@@ -1,37 +1,52 @@
...
 
 	return (
 		<Page flexDirection='row'>
 			<Page>
 				<Page.Header title={t('Moderation')} />
 				<Page.Content>
-					<ModerationConsoleTable />
+					<Tabs>
```
dougfabris, line 39: This `Tabs` must live outside `Page.Content` to get this result: 

Screenshot 2023-10-19 at 18 38 38

dougfabris (changes requested):
I couldn't see the value of having two tabs, for messages and for users and then having a context bar that allows the user to check messages and user again. Honestly is confusing... IMO we should revisit this decision, it doesn't seem clear. 
Also, I realized a bunch of non-pattern UI things like loading through strings, spacings, and more. I would ask you to take more time revisiting the UI to guarantee these small issues
Finally, the first render of the users tab is blinking, please check the gif below: 

Kapture 2023-10-19 at 18 43 49

apps/meteor/client/views/admin/moderation/MessageContextFooter.tsx
Dnouv: Thank you, done!

apps/meteor/client/views/admin/moderation/ModConsoleReportDetails.tsx
Dnouv: Thank you, done!
Could you please help me understand why this is not something expected? I would appreciate your help

apps/meteor/client/views/admin/moderation/ModConsoleReportDetails.tsx
Dnouv: Thank you, done!

apps/meteor/client/views/admin/moderation/ModerationConsolePage.tsx
Dnouv: Thank you, done!

apps/meteor/client/views/admin/moderation/UserReports/UserReportInfo.tsx
Dnouv: Thank you, done!

apps/meteor/client/views/admin/moderation/UserReports/UserReportInfo.tsx
Dnouv: Thank you, done!

apps/meteor/client/views/admin/moderation/UserReports/UserReportInfo.tsx
Dnouv: Thank you, done!

apps/meteor/client/views/admin/moderation/helpers/ModerationFilter.tsx
Dnouv: I didn't know about this, but this is interesting. Thank you, done!

Curiosity: Do we avoid Grid whenever possible, or are there some specific cases, where we should use it.

apps/meteor/client/views/admin/moderation/ModerationConsolePage.tsx
Dnouv: Thank you, done!

apps/meteor/client/views/admin/moderation/ModerationConsolePage.tsx
Dnouv: Thank you, done!

apps/meteor/client/views/admin/moderation/ModConsoleReportDetails.tsx
Dnouv: Thank you, done!

apps/meteor/client/views/admin/moderation/UserReports/ModConsoleUserActions.tsx
Dnouv: Thank you, done!

apps/meteor/client/views/admin/moderation/UserReports/ModConsoleUsersTable.tsx
Dnouv: Thank you, done!

apps/meteor/client/views/admin/moderation/UserReports/ModConsoleUsersTable.tsx
Dnouv: Thank you, done!

apps/meteor/client/views/admin/moderation/UserReports/ModConsoleUsersTable.tsx
Dnouv: Thank you, done!

apps/meteor/client/views/admin/moderation/UserReports/UserReportInfo.tsx
Dnouv: Actually, the data we got from the react query had the `createdAt:string` but the actual types are `createdAt:Date`. To resolve this, I had done an unknown casting initially, `as unknown as IUser`, but changed it back to using a type guard (`isUser`) after the suggestion above, which I feel like better for avoiding unintentional break.

apps/meteor/client/views/admin/moderation/helpers/ModerationFilter.tsx
dougfabris: Just use `Grid` to render complex structures aiming to have a specific result on mobile

apps/meteor/client/views/admin/moderation/MessageContextFooter.tsx
dougfabris: There's no need to `title` for `Buttons` with labels

dougfabris (changes requested):

dougfabris (changes requested):
Some visual issues still remain:
_I would suggest some spacing between the contextual bar title and the tabs as well_

<div style='display:"flex"'>
Screenshot 2023-10-30 at 11 03 04
Screenshot 2023-10-30 at 11 03 19

</div>

Shouldn't at least the first message be shown by default, the purpose of the tab is to show the report why hide it?

<div style='display:"flex"'>
Screenshot 2023-10-30 at 11 03 30

</div>

Dnouv:
Thank you so much for the review, @dougfabris 

> I would suggest some spacing between the contextual bar title and the tabs as well

Sorry, but I did not clearly understand, the suggestion here. 

> Shouldn't at least the first message be shown by default, the purpose of the tab is to show the report why hide it?

Initially, we had the UI where we had the flow:
(Click on the "See Messages") -> _a random message_ (hover) -> In the message toolbar actions, there was a "View Reports" -> (Click on it and appears a second context bar, with report reasons) 

It was pretty difficult to notice how to view report reasons, so, we moved to this new UI, where we have the report reasons, and the messages on a single context bar. 
And, yes, initially, we had, all the reports visible without a collapsible button. But, there are situations, where, there are 10 messages, each having 10 reports, that would have made the context bar, very long to scroll to the last message. So, a collapsible button. 

And yes, the purpose was to view the reported messages. Because, many times, the Moderator only views the reported message and takes action, since these reported messages are usually self-explanatory.

I have tried to summarize the discussion, we had, while designing this UI, please feel free to let me know, if there are questions.

Thank you!

dougfabris:
> Sorry, but I did not clearly understand, the suggestion here.

We should have a spacing like this, not so tight with the title

Screenshot 2023-10-30 at 13 09 20


> And, yes, initially, we had, all the reports visible without a collapsible button. But, there are situations, where, there are 10 messages, each having 10 reports, that would have made the context bar, very long to scroll to the last message. So, a collapsible button.

I see, so that's ok then, but don't you think would be cool to have the report visible for the latest message?

apps/meteor/client/views/admin/moderation/ModConsoleReportDetails.tsx
dougfabris: It's not necessary to tell you're going to return a ReactElement, it's identified automatically

apps/meteor/client/views/admin/moderation/UserReports/UserReportInfo.tsx
dougfabris: You can use `Serialized` type to resolve it. It should be something like: `Serialized<IUser>`
`Serialized` is found on @rocket.chat/core-typings

apps/meteor/client/views/admin/moderation/UserReports/UserReportInfo.tsx
dougfabris: Remove the object from the keys as well, should be `['moderation.usersReport', userId]`

Dnouv:
Here are the updated screen captures; now we have better spacing and alignments. Thank you!

image

apps/meteor/client/views/admin/moderation/ModConsoleReportDetails.tsx
Dnouv: I see, interesting, thank you!

apps/meteor/client/views/admin/moderation/UserReports/UserReportInfo.tsx
Dnouv: Thank you; this is really helpful.

apps/meteor/client/views/admin/moderation/UserReports/UserReportInfo.tsx
Dnouv: Thank you, done!

apps/meteor/client/views/admin/moderation/helpers/ModerationFilter.tsx
Dnouv: Thank you, I understand, I will keep a note of this.

dougfabris (dismissed):
LGTM!

apps/meteor/client/views/admin/moderation/UserReports/ModConsoleUsersTable.tsx
```diff
@@ -0,0 +1,148 @@
...
+		() => ({
+			selector: debouncedText,
+			sort: JSON.stringify({ [sortBy]: sortDirection === 'asc' ? 1 : -1 }),
+			count: itemsPerPage,
+			offset: current,
+			latest: end ? `${new Date(end).toISOString().slice(0, 10)}T23:59:59.999Z` : undefined,
+			oldest: start ? `${new Date(start).toISOString().slice(0, 10)}T00:00:00.000Z` : undefined,
+		}),
+		[current, debouncedText, end, itemsPerPage, sortBy, sortDirection, start],
+	);
```
ggazzo, line 50: there is no need to memoize this object , usequery always serialize the object resulting in a hash

apps/meteor/client/views/admin/moderation/hooks/useResetAvatarAction.tsx
```diff
@@ -25,8 +25,11 @@ const useResetAvatarAction = (userId: string): GenericMenuItemProps => {
 
 	const onResetAvatar = async () => {
 		setModal();
-		handleResetAvatar.mutateAsync({ userId });
+		await handleResetAvatar.mutateAsync({ userId });
 		queryClient.invalidateQueries({ queryKey: ['moderation.reports'] });
+		queryClient.invalidateQueries({ queryKey: ['moderation.usersReport', { userId }] });
```
ggazzo, line 30: what about changing yours querykey to `['moderation', ...]` 
you can invalidate all of them just using `['moderation']`
```suggestion
		queryClient.invalidateQueries({ queryKey: ['moderation', 'usersReport', { userId }] });
```

apps/meteor/tests/data/moderation.helper.ts
```diff
@@ -0,0 +1,16 @@
...
+	});
+	return res.body;
+};
+
+export const getUsersReports = async (userId: string) => {
+	const res = await request.get(api('moderation.user.reportsByUserId')).set(credentials).query({
+		userId,
+	});
+	return res.body;
+};
```
ggazzo, line 16: try to avoid such functions it doest really help a lot something like that could be a bit. more usefull
```suggestion

const whatever = (url: string) => async (userId: string, reason: string) => {
	const res = await request.post(api(url)).set(credentials).send({
		userId,
		reason,
	});
	return res.body;
};

export const reportUser =whatever('moderation.reportUser')

export const getUsersReports = whatever('moderation.user.reportsByUserId'))
```


but still I would not omit the url from the tests

apps/meteor/client/views/admin/moderation/helpers/UserColumn.tsx
```diff
@@ -0,0 +1,45 @@
...
+import UserAvatar from '../../../../components/avatar/UserAvatar';
+
+type UserColumnProps = {
+	name?: string;
+	username?: string;
+	isDesktopOrLarger?: boolean;
+	isProfile?: boolean;
+};
+
+const UserColumn = ({ name, username, isDesktopOrLarger, isProfile }: UserColumnProps) => {
```
ggazzo, line 13: try to avoid props starting with `is` usually is should be used in functions

also usually using boolean props for customize components is a code smell , probably you are trying to save sometime here instead of using 2..3 different compoentns

ggazzo: https://www.informit.com/articles/article.aspx?p=1392524

“Boolean arguments loudly declare that the function does more than one thing. They are confusing and should be eliminated.”
— Clean Code, Robert C. Martin (Uncle Bob)

apps/meteor/client/views/admin/moderation/UserReports/UserReportInfo.tsx
```diff
@@ -0,0 +1,135 @@
...
+				<Callout mbs={8} type='warning' icon='warning'>
+					{t('Moderation_User_deleted_warning')}
+				</Callout>
+			</Box>
+		);
+	};
+
+	const renderUserReports = (reports: Serialized<Omit<UserReport, 'moderationInfo'>[]>) => {
+		return reports.map((report, ind) => (
+			<Box key={report._id}>
```
ggazzo, line 96: what is the point of using `Box` here?

apps/meteor/client/views/admin/moderation/UserReports/UserReportInfo.tsx
```diff
@@ -0,0 +1,135 @@
...
+				<ReportReason ind={ind + 1} uinfo={report.reportedBy?.username} msg={report.description} ts={new Date(report.ts)} />
```
ggazzo, line 97: what is `ind`?

apps/meteor/client/views/admin/moderation/UserReports/UserReportInfo.tsx
```diff
@@ -0,0 +1,135 @@
...
+	const userEmails = useMemo(() => {
+		if (!report?.user?.emails) {
+			return [];
+		}
+		return report.user.emails.map((email) => email.address);
+	}, [report]);
+
+	const renderUserDetails = (user: Serialized<IUser>) => {
+		return (
+			<Box>
```
ggazzo, line 60: do not use Box as fragment

ggazzo (changes requested):
please check my comments

apps/meteor/client/views/admin/moderation/UserReports/ModConsoleUsersTable.tsx
Dnouv: Nice catch. Thank you! I have corrected it.

apps/meteor/client/views/admin/moderation/hooks/useResetAvatarAction.tsx
Dnouv: That's a good idea. Thank you! Renamed the keys.

apps/meteor/client/views/admin/moderation/helpers/UserColumn.tsx
Dnouv: Thank you! This was a good read. 

I have changed the props to accept the `size` so I could use the component, instead of creating multiple of them.

apps/meteor/client/views/admin/moderation/UserReports/UserReportInfo.tsx
```diff
@@ -0,0 +1,135 @@
...
+
+	const renderDeletedUserWarning = () => {
+		return (
+			<Box>
+				<Callout mbs={8} type='warning' icon='warning'>
+					{t('Moderation_User_deleted_warning')}
+				</Callout>
+			</Box>
+		);
+	};
```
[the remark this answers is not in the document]
Dnouv: Sorry, but do we need subcomponents? Because I need to access some of the parents' state. And since this is not going to be used elsewhere, and the logic was not complex either, so I thought keeping it as a subfunction within the parent component seems like a good approach.

I might be missing something, please let me know if I did.

apps/meteor/client/views/admin/moderation/UserReports/UserReportInfo.tsx
Dnouv: Nice catch, not needed. I have removed it. Thank you!

Dnouv:
Converted to draft. I am working on it, after doing some tests, I will try to get this RFR (ready for review) by today. 
Thank you!

apps/meteor/tests/data/moderation.helper.ts
Dnouv: Thank you, I will definitely keep a note of it.

apps/meteor/client/views/admin/moderation/UserReports/UserReportInfo.tsx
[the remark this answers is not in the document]
Dnouv: Thank you! I have removed the sub-functions.

tassoevan (approved):

apps/meteor/server/models/raw/ModerationReports.ts
```diff
@@ -18,12 +19,11 @@ export class ModerationReportsRaw extends BaseRaw<IModerationReport> implements
 
 	modelIndexes(): IndexDescription[] | undefined {
 		return [
-			{ key: { 'ts': 1, 'reports.ts': 1 } },
-			{ key: { 'message.u._id': 1, 'ts': 1 } },
-			{ key: { 'reportedUser._id': 1, 'ts': 1 } },
-			{ key: { 'message.rid': 1, 'ts': 1 } },
+			{ key: { '_hidden': 1, 'ts': 1, 'reports.ts': 1 } },
```
sampaiodiego, line 22: if I read the code correctly, `reports.ts` is a field created in a pipeline of an aggregation, right? in this case an index will not be used..

apps/meteor/server/models/raw/ModerationReports.ts
```diff
@@ -132,6 +132,83 @@ export class ModerationReportsRaw extends BaseRaw<IModerationReport> implements
...
+			},
+			ts: {
+				$lt: latest,
+				$gt: oldest,
+			},
+			...(isMessageReports ? this.getSearchQueryForSelector(selector) : this.getSearchQueryForSelectorUsers(selector)),
+		};
+
+		const field = isMessageReports ? 'message.u._id' : 'reportedUser._id';
+		const ids = await this.col.distinct(field, query);
```
sampaiodiego, line 208: you don't need to fetch **all data** to know how many different values are.. you can can use an aggregation for that, something like:
```js
{
  $group: {
    _id: '$message.u._id', // group all values into a single _id key
  }
},
{
  $group: {
    _id: 'total', // group all previously values into a single value
    total: { $sum: 1 }, // sum the total values from previous pipeline
  }
}
```

apps/meteor/server/models/raw/ModerationReports.ts
```diff
@@ -18,12 +19,11 @@ export class ModerationReportsRaw extends BaseRaw<IModerationReport> implements
...
 	modelIndexes(): IndexDescription[] | undefined {
 		return [
-			{ key: { 'ts': 1, 'reports.ts': 1 } },
-			{ key: { 'message.u._id': 1, 'ts': 1 } },
-			{ key: { 'reportedUser._id': 1, 'ts': 1 } },
-			{ key: { 'message.rid': 1, 'ts': 1 } },
+			{ key: { '_hidden': 1, 'ts': 1, 'reports.ts': 1 } },
 			{ key: { userId: 1, ts: 1 } },
 			{ key: { 'message._id': 1, 'ts': 1 } },
+			{ key: { '_hidden': 1, 'message.u._id': 1, 'ts': 1 } },
```
sampaiodiego, line 25: I'm not totally sure the queries `{ _hidden: { $ne: true } }` will actually use this indexes.. can you please confirm by by doing an `.explain('executionStats')` on mongoshell?

it is also clear  that all indexes share the same `ts` field, I wonder if we can optimize those indexes by having less compound indexes being used by multiple queries.

sampaiodiego (changes requested):
general comment regarding changing indexes:

when you change an index, the old index is not removed, it will still be in the database even though our code doesn't "make sure" it is there on every start.

so if you need to "change" an index, just comment out the old ones in the model but with a comment saying they're deprecated and should be remove in the next major version via a migration.. and add the "changed" indexes as new ones.. 

I have added other inline comments as well.

apps/meteor/server/models/raw/ModerationReports.ts
Dnouv: Thank you, good catch. Yes, since it is created in the `$group` stage, it is not used.

apps/meteor/server/models/raw/ModerationReports.ts
Dnouv: Thank you, yes, we could do it, but I got to learn in the past, we had some issues with aggregation, so we should try to avoid the `aggregation` as much as possible. Please let me know if I misunderstood something.

apps/meteor/server/models/raw/ModerationReports.ts
Dnouv: Thank you, yes, it uses these indexes. I could not find a better solution for optimizing the common `ts` field because of respecting the ESR Rule. 
Both common fields `_hidden` and `ts` use range operators.

apps/meteor/server/models/raw/ModerationReports.ts
Dnouv: Had a discussion on this, and made the suggested changes. Thank you!

.changeset/rare-eels-fetch.md
sampaiodiego, line 1: pls add `@rocket.chat/meteor` to the changeset.. it is missing from the list, since there was indeed many many changes to it (like new endpoints and such).. but it is also the most important one, since it is used to generate the final release change log.

apps/meteor/server/models/raw/ModerationReports.ts
```diff
@@ -18,12 +19,17 @@ export class ModerationReportsRaw extends BaseRaw<IModerationReport> implements
 
 	modelIndexes(): IndexDescription[] | undefined {
 		return [
-			{ key: { 'ts': 1, 'reports.ts': 1 } },
-			{ key: { 'message.u._id': 1, 'ts': 1 } },
-			{ key: { 'reportedUser._id': 1, 'ts': 1 } },
-			{ key: { 'message.rid': 1, 'ts': 1 } },
-			{ key: { userId: 1, ts: 1 } },
-			{ key: { 'message._id': 1, 'ts': 1 } },
+			// deprecated
```
sampaiodiego, line 22:
```suggestion
			// TODO deprecated. remove within a migration in v7.0
```

apps/meteor/app/api/server/v1/moderation.ts
```diff
@@ -63,11 +64,65 @@ API.v1.addRoute(
...
+					reports,
+					count: reports.length,
+					offset,
+					total,
+				};
+				return API.v1.success(result);
+			} catch (error) {
+				SystemLogger.error(error);
+				return API.v1.failure('Error while fetching user reports');
+			}
```
sampaiodiego, line 116: this try/catch isn't really needed, since the API code has one already for each endpoint.

apps/meteor/app/api/server/v1/moderation.ts
```diff
@@ -113,6 +168,69 @@ API.v1.addRoute(
...
+					user,
+					reports,
+					count: reports.length,
+					total,
+					offset,
+				});
+			} catch (error) {
+				SystemLogger.error(error);
+				return API.v1.failure('Error while fetching user reports');
+			}
```
sampaiodiego, line 229: same as the previous one: this try/catch isn't really needed, since the API code has one already for each endpoint.

apps/meteor/tests/end-to-end/api/27-moderation.ts
```diff
@@ -73,6 +74,66 @@ describe('[Moderation]', function () {
 		});
 	});
 
+	describe('[/moderation.userReports]', () => {
```
sampaiodiego, line 77: this endpoint is missing some tests for fails, like missing permission or auth, invalid params, etc.

sampaiodiego (changes requested):
tests can be improved later during the release candidate cycle.. the other change requests should be very simple to get done before the cut. 😉

apps/meteor/app/api/server/v1/moderation.ts
```diff
@@ -113,6 +168,69 @@ API.v1.addRoute(
 	},
 );
 
+API.v1.addRoute(
+	'moderation.user.reportsByUserId',
```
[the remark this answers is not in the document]
Dnouv: Nice catch! Did not notice this. Thank you, changed!

.changeset/rare-eels-fetch.md
Dnouv: Thank you, added!

apps/meteor/app/api/server/v1/moderation.ts
Dnouv: Thank you, removed!

apps/meteor/app/api/server/v1/moderation.ts
Dnouv: Thank you, removed!

apps/meteor/tests/end-to-end/api/27-moderation.ts
Dnouv: Thank you. I will improve the tests in the next PR.

sampaiodiego (approved):

PR #30555, opened 2023-10-03
regression: fix initializing startup order

ggazzo:

### DIFF
--- a/packages/core-services/src/lib/Api.ts
+++ b/packages/core-services/src/lib/Api.ts
@@ -47,7 +47,7 @@ export class Api implements IApiService {
 
 	async broadcast<T extends keyof EventSignatures>(event: T, ...args: Parameters<EventSignatures[T]>): Promise<void> {
 		if (!this.broker) {
-			throw new Error(`No broker set to start: ${event}`);
+			throw new Error(`No broker set to broadcast: ${event}`);
 		}
 
 		return this.broker.broadcast(event, ...args);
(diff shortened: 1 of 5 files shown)


packages/core-services/src/lib/Api.ts
```diff
@@ -46,7 +46,11 @@ export class Api implements IApiService {
 	}
 
 	async broadcast<T extends keyof EventSignatures>(event: T, ...args: Parameters<EventSignatures[T]>): Promise<void> {
-		return this.broker?.broadcast(event, ...args);
+		if (!this.broker) {
+			throw new Error(`No broker set to start.${event}`);
```
ggazzo, line 50:
```suggestion
			throw new Error(`No broker set to start: ${event}`);
```

packages/core-services/src/lib/Api.ts
```diff
@@ -46,7 +46,11 @@ export class Api implements IApiService {
 	}
 
 	async broadcast<T extends keyof EventSignatures>(event: T, ...args: Parameters<EventSignatures[T]>): Promise<void> {
-		return this.broker?.broadcast(event, ...args);
+		if (!this.broker) {
+			throw new Error(`No broker set to start: ${event}`);
```
ggazzo, line 50:
```suggestion
			throw new Error(`No broker set to broadcast: ${event}`);
```

sampaiodiego (approved):

PR #30568, opened 2023-10-04
fix: Fixed editing the custom emoji for name field without file upload. #26980

AdityaRaut13:

1. I added to the `getUploadFormData` function, another parameter to check if the route requires the file or not, and made the changes in the code accordingly.

2. I also changed the cache control from the emoji-custom route to none. The problem with keeping a cache-control is that if a user changes to the previous name for a particular emoji, it takes the earlier result without validating it from the server, which essentially causes another bug on the client's side.

https://github.com/RocketChat/Rocket.Chat/assets/102903554/a1b15b58-60e0-4a10-881c-f134d1ddad86

https://github.com/RocketChat/Rocket.Chat/issues/26980

## Issue(s)

1. `getUploadFormData` required a file for further processing along the busboy stream parser.
4. There were cache-control on the custom-emoji being send to the user. 

## Steps to test or reproduce

1. Add a custom emoji in case you don't already have one.
2. Select a custom emoji for editing.
4. Edit its name or aliases.
5. Click the "Save" button.

### DIFF
--- a/apps/meteor/app/api/server/lib/getUploadFormData.ts
+++ b/apps/meteor/app/api/server/lib/getUploadFormData.ts
@@ -5,15 +5,18 @@ import type { ValidateFunction } from 'ajv';
 import busboy from 'busboy';
 import type { Request } from 'express';
 
-type UploadResult<K> = {
+type UploadedFile = {
 	file: Readable & { truncated: boolean };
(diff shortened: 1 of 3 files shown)


AdityaRaut13:
@dougfabris  can you please review this pr ?

PR #30570, opened 2023-10-04
chore: Register workspace from marketplace App Installation

csuadev:
When an admin tries to install an app and the workspace is not registered to the cloud, it shows the modal to register the workspace instead of an error and after the register finishes, it shows the app installation flow. 

## Issue(s)
[https://rocketchat.atlassian.net/browse/AECO-365](https://rocketchat.atlassian.net/browse/AECO-365)

## Steps to test or reproduce
- Have an unregistered workspace.
- Go to marketplace and try to install an app.


ggazzo (changes requested):
anyone able to install an app should be able to register the workspace? I dont think so. there is any condition to handle that case?

PR #30571, opened 2023-10-04
chore: Setup wizard translation update

tiagoevanp:
## Further comments

Related PR: https://github.com/RocketChat/fuselage/pull/1185

NBJ-417


ggazzo (changes requested):
can I ask you why you just inserted a translation key, but with no use?

tiagoevanp:
> can I ask you why you just inserted a translation key, but with no use?

This is a onboarding-ui translation... let me add co-related PR on fuselage

PR #30573, opened 2023-10-04
feat: Count daily and monthly peaks of concurrent connections

matheusbsilva137:

 - Add the `dailyPeakConnections` statistic for monitoring the daily peak of concurrent connections in a workspace;
 - Add `maxMonthlyPeakConnections` statistic for monitoring the last 30 days peak of concurrent connections in a workspace;

## Steps to test or reproduce

 * A new set of statistics for the current date can be generated using the [/statistics](https://developer.rocket.chat/reference/api/rest-api/endpoints/statistics/stats-endpoints/get-statistics) endpoint. The current amount of users connected to a workspace can be checked with the existing `presence.getConnections` endpoint (the value generated for the `dailyPeakConnections` statistic is the highest value that could be returned by this endpoint since the last time the statistics were saved via cronjob).

## Further comments

NBJ-416


apps/meteor/app/api/server/v1/presence.ts
```diff
@@ -14,6 +14,18 @@ API.v1.addRoute(
 	},
 );
 
+API.v1.addRoute(
+	'presence.getMonthlyPeakConnections',
```
ggazzo, line 18: instead of creating the endpoint why not just get the IStats as we are doing with other info?

matheusbsilva137: hmmm I thought an endpoint would fit better in this case since we don't want just the raw value that is in the statistics (the daily peak). Instead, we want the maximum daily peak from the last 30 days (+ consider the current peak from the ongoing day so that this info makes more sense) 🤔 
Do you think it'd be enough to grab this from the statistics?

janainaCoelhoRocketchat:
image
tested

apps/meteor/app/api/server/v1/presence.ts
sampaiodiego: I changed this to add the monthly data into the stats as well

sampaiodiego (approved):

sampaiodiego:
I've made a few changes to the PR.. removed the previously created new endpoint in favor of adding the values to the statistics.. 👍

hugocostadev (approved):

rodrigok:
We need to state clearly that this only works for single instance deployments since it does not sum the count of connections among instances.

sampaiodiego:
> We need to state clearly that this only works for single instance deployments since it does not sum the count of connections among instances.

@rodrigok this is not true.. it does count connections from all instances.. each instance store the concurrent connections to their record into the `instances` collection.. the Presence service then summarize them up..

PR #30577, opened 2023-10-05
refactor: AutoCompleteTagsMultiple to TS

MartinSchoeler:


apps/meteor/ee/client/hooks/useTagsList.ts
```diff
@@ -12,17 +11,19 @@ type TagsListOptions = {
 	viewAll?: boolean;
 };
 
+type TagsItemsList = { _id: string; label: string; value: string; _updatedAt: Date };
```
aleksandernsilva, line 14: This name in my opinion does not correspond to the type. The type seem to be of a singular item, so something like `TagsItem` _(or similar)_ might be more appropriate.

apps/meteor/ee/client/omnichannel/tags/AutoCompleteTagsMultiple.tsx
```diff
@@ -25,7 +38,7 @@ const AutoCompleteTagMultiple = (props) => {
 	const { phase: tagsPhase, items: tagsItems, itemCount: tagsTotal } = useRecordList(tagsList);
 
 	const tagsOptions = useMemo(() => {
-		const pending = value.filter(({ value }) => !tagsItems.find((tag) => tag.value === value));
+		const pending = value ? value.filter(({ value }) => !tagsItems.find((tag) => tag.value === value)) : [];
```
aleksandernsilva, line 41: I would just add a default value of `[]` to the props

apps/meteor/ee/client/omnichannel/tags/AutoCompleteTagsMultiple.tsx
```diff
@@ -41,9 +54,11 @@ const AutoCompleteTagMultiple = (props) => {
 			flexShrink={0}
 			flexGrow={0}
 			placeholder={t('Select_an_option')}
-			endReached={tagsPhase === AsyncStatePhase.LOADING ? () => {} : (start) => loadMoreTags(start, Math.min(50, tagsTotal))}
+			endReached={
+				tagsPhase === AsyncStatePhase.LOADING ? () => undefined : (start) => start && loadMoreTags(start, Math.min(50, tagsTotal))
```
aleksandernsilva, line 58: We have a lot of places doing this same pagination without the `start &&` check. Why is it necessary here?

aleksandernsilva (changes requested):

aleksandernsilva (approved):

PR #30606, opened 2023-10-09
chore: Remove text decoration from room tag

juliajforesti:

Screen Shot 2023-10-09 at 11 46 57 &#8594; Screen Shot 2023-10-09 at 11 47 40


dougfabris (approved):
As a workaround looks approvable to me!

PR #30610, opened 2023-10-09
fix: Read receipts are not created on the first time a user reads a room

matheusbsilva137:

 - When a user accesses a room for the first time, mark as read all the messages sent after they joined the room.

## Further comments

TC-986


sampaiodiego (dismissed):

heitortanoue (approved):

matheusbsilva137 (commented):
QA noticed that when creating a room already with a member, the first messages would still not be marked as read when reading the room with the invited user (that's because we **update** their subscription by the time the first message is sent, which makes `_updatedAt` timestamp bigger than the message's `ts`). That's now fixed by using the subscription's `ts` instead
Thanks @jessicaschelly

heitortanoue (approved):

PR #30616, opened 2023-10-10
fix: Change plan name from Enterprise to Premium on marketplace filtering

csuadev:
Due to the marketplace don't showing apps when Enterprise category is selected, change category name on filtered apps since a change on the database from all apps from Enterprise to Premium.


hugocostadev (approved):

casalsgh:
AECO-409 is the one related to this

PR #30625, opened 2023-10-11
fix: Message mentions displayed in the UI are outdated after changing username

matheusbsilva137:

 - Unset markdown (`md`) field when updating messages that mention the username changed (so that the client tries to parse it again with the updated content).

## Further comments

SUP-378


yash-rajpal (changes requested):
Shouldn't we parse the `newMessage` and update the `md` property?

matheusbsilva137:
Handled in #31860

PR #30631, opened 2023-10-12
refactor: Livechat functions to Typescript

KevLehman:
## Further comments

Also included: smol refactors to reduce db calls & missing types


apps/meteor/app/livechat/server/api/v1/room.ts
```diff
@@ -363,7 +368,7 @@ API.v1.addRoute(
 				throw new Error('invalid-visitor');
 			}
 
-			const room = await LivechatRooms.findOneById(rid, { _id: 1, v: 1 }); // TODO: check _id
+			const room = await LivechatRooms.findOneById(rid, { projection: { ...roomAccessAttributes, _id: 1, t: 1 } }); // TODO: check _id
```
KevLehman, line 371:
```suggestion
			const room = await LivechatRooms.findOneById(rid, { projection: { ...roomAccessAttributes, _id: 1, t: 1, v: 1 } }); // TODO: check _id
```

apps/meteor/app/livechat/server/api/v1/room.ts
```diff
@@ -352,7 +352,12 @@ API.v1.addRoute(
 
 API.v1.addRoute(
 	'livechat/room.visitor',
-	{ authRequired: true, permissionsRequired: ['view-l-room'], validateParams: isPUTLivechatRoomVisitorParams, deprecationVersion: '7.0.0' },
+	{
+		authRequired: true,
+		permissionsRequired: ['change-livechat-room-visitor'],
```
sampaiodiego, line 357: is this permission change intentional? if so, is this a bug fix? 😬

murtaza98, line 357:
```suggestion
		permissionsRequired: ['change-livechat-room-visitor', 'view-l-room'],
```

murtaza98: Considering previously we needed both these permissions to change room's visitor

KevLehman: view-l-room was there, but was actually not of any use. Inside the only function call, it checked for `change-livechat-room-visitor` which means l-room alone was useless.

So i removed it :)

murtaza98: Yeah makes sense :+1: Just thought about raising this since this is a refactoring PR which is changing some behaviour of the product

murtaza98 (approved):

apps/meteor/app/livechat/server/api/v1/room.ts
KevLehman: Thank you! Yeah, the thing is, the behavior keeps unchanged. We just check the permission before

sampaiodiego (approved):

debdutdeb (approved):

PR #30637, opened 2023-10-13
test: read receipts

heitortanoue:

Create a new file dedicated to the read-receipts feature tests and improve its coverage (including some edge-cases introduced in the last major release).

## Further comments

TC-502


apps/meteor/tests/end-to-end/api/24-methods.js
```diff
@@ -123,6 +126,276 @@ describe('Meteor.methods', function () {
 		});
 	});
 
+	describe('[@getReadReceipts]', () => {
+		it('should fail if not logged in', async () => {
```
pierre-lehnen-rc, line 130: This request will fail even if the user is logged in, since no params are being passed.
You should send an otherwise valid request and also check the error type to ensure it's the user state that is causing the request to fail.

apps/meteor/tests/end-to-end/api/24-methods.js
```diff
@@ -123,6 +126,276 @@ describe('Meteor.methods', function () {
...
+							id: 'id',
+							msg: 'method',
+						}),
+					})
+					.expect('Content-Type', 'application/json')
+					.expect(200)
+					.expect((res) => {
+						expect(res.body).to.have.a.property('success', true);
+						expect(res.body).to.have.a.property('message').that.is.a('string');
+						console.log('res.body', res.body);
```
pierre-lehnen-rc, line 215:
```suggestion
```

apps/meteor/tests/end-to-end/api/24-methods.js
```diff
@@ -123,6 +126,276 @@ describe('Meteor.methods', function () {
...
+						expect(res.body).to.have.a.property('message').that.is.a('string');
+
+						const data = JSON.parse(res.body.message);
+						expect(data).to.have.a.property('result').that.is.an('array');
+						expect(data.result.length).to.equal(1);
+						expect(data.result[0]).to.have.property('userId', credentials['X-User-Id']);
+					});
+			});
+
+			it("should read all main room's messages with the invited user", async () => {
```
pierre-lehnen-rc, line 250: This isn't a test, but a requirement for the next few tests; 
Move them all to a new `describe` section that runs independently from this one, with it's own `before` sub-section.

apps/meteor/tests/end-to-end/api/24-methods.js
```diff
@@ -123,6 +126,276 @@ describe('Meteor.methods', function () {
...
+						expect(res.body).to.have.a.property('message').that.is.a('string');
+
+						const data = JSON.parse(res.body.message);
+						expect(data).to.have.a.property('result').that.is.an('array');
+						expect(data.result.length).to.equal(1);
+						expect(data.result[0]).to.have.property('userId', credentials['X-User-Id']);
+					});
+			});
+
+			it('should read thread messages with the invited user', async () => {
```
pierre-lehnen-rc, line 322: Same as above, this one is not a test, but data manipulation.

apps/meteor/tests/end-to-end/api/24-methods.js
```diff
@@ -123,6 +126,276 @@ describe('Meteor.methods', function () {
...
+						const data = JSON.parse(res.body.message);
+						expect(data).to.have.a.property('result').that.is.an('array');
+						expect(data.result.length).to.equal(2);
+
+						const receiptsUserIds = [data.result[0].userId, data.result[1].userId];
+						expect(receiptsUserIds).to.have.members([credentials['X-User-Id'], user._id]);
+					});
+			});
+
+			it('should mark the thread as read', async () => {
```
pierre-lehnen-rc, line 377: This one is also data manipulation, that is not even being validated by anything.

pierre-lehnen-rc (changes requested):
You'll need multiple `describe` sections for all situations you're testing, each one with their own messages in the correct state.

1 - simple message and thread that nobody has read yet
2 - simple message and thread where the room message was read by another user but the thread message was not
3 - another test where everything was read by both users
4 - another test that checks the result of `getReadReceipts` after the call to `readThreads`

pierre-lehnen-rc (changes requested):
Avoid calling `expect` inside the `before` section of tests; if the call fails the test will fail anyway.

pierre-lehnen-rc (approved):

PR #30645, opened 2023-10-16
chore: Handling plural translations

yash-rajpal:
## Further comments

TC-993


tassoevan (changes requested):
Why not to use the own i18next pluralization handling? https://www.i18next.com/translation-function/plurals

tassoevan (approved):

PR #30655, opened 2023-10-17
chore: Add VSCode extensions recommendations

rodrigok:

### DIFF
--- a/.vscode/extensions.json
+++ b/.vscode/extensions.json
@@ -1,0 +1,9 @@
+{
+	"recommendations": [
+		"streetsidesoftware.code-spell-checker",
+		"streetsidesoftware.code-spell-checker-portuguese-brazilian",
+		"editorconfig.editorconfig",
+		"dbaeumer.vscode-eslint",
+		"eamodio.gitlens"
+	]
+}


.vscode/extensions.json
```diff
@@ -0,0 +1,9 @@
+{
+	"recommendations": [
+		"streetsidesoftware.code-spell-checker",
+		"streetsidesoftware.code-spell-checker-portuguese-brazilian",
+		"editorconfig.editorconfig",
+		"dbaeumer.vscode-eslint",
+		"eamodio.gitlens"
```
KevLehman, line 7: Should we add `esbenp.prettier-vscode` too?

hugocostadev: I think so...

.vscode/extensions.json
```diff
@@ -0,0 +1,9 @@
+{
+	"recommendations": [
+		"streetsidesoftware.code-spell-checker",
+		"streetsidesoftware.code-spell-checker-portuguese-brazilian",
+		"editorconfig.editorconfig",
+		"dbaeumer.vscode-eslint",
```
hugocostadev, line 6: Playwright VS code extension is a very useful tool to debug and write tests

```suggestion
		"dbaeumer.vscode-eslint",
		"ms-playwright.playwright",
```

KevLehman: Think playwright is more situational. For example, BE devs won't use it :trollface:

.vscode/extensions.json
tassoevan: It might conflict with the ESLint one.

.vscode/extensions.json
debdutdeb: Agreed with Kevin. 

But also doesn't hurt. Playwright is used a lot in the project so i think makes sense to include. Anyone not wanting can ignore.

.vscode/extensions.json
```diff
@@ -0,0 +1,9 @@
+{
+	"recommendations": [
+		"streetsidesoftware.code-spell-checker",
+		"streetsidesoftware.code-spell-checker-portuguese-brazilian",
```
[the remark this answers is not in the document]
rodrigok: It helps on translation files. We probably should have for every official supported language.

PR #30652, opened 2023-10-16
fix: processWebhookMessage & chat.postMessage types

gabriellsh:

PR #30656, opened 2023-10-17
chore: Call api only once when sharing a message

gabriellsh:
## Further comments

Depends of #30652


dougfabris (approved):
great catch my friend ❤️

PR #30664, opened 2023-10-17
refactor: Move functions out of `livechat.js`

KevLehman:


apps/meteor/app/livechat/server/lib/LivechatTyped.ts
```diff
@@ -1123,6 +1162,172 @@ class LivechatClass {
...
+		}
+
+		const cursor = LivechatRooms.findByVisitorToken(token);
+		for await (const room of cursor) {
+			await Promise.all([
+				FileUpload.removeFilesByRoomId(room._id),
+				Messages.removeByRoomId(room._id),
+				ReadReceipts.removeByRoomId(room._id),
+			]);
+		}
```
murtaza98, line 1215: Won't this be synchronous? Like it'd process one room after an other, instead of processing all of them without any sequence

apps/meteor/server/features/EmailInbox/EmailInbox_Incoming.ts
```diff
@@ -148,7 +147,7 @@ export async function onEmailReceived(email: ParsedMail, inbox: string, departme
 	const rid = room?._id ?? Random.id();
 	const msgId = Random.id();
 
-	Livechat.sendMessage({
+	LivechatTyped.sendMessage({
```
murtaza98, line 150:
```suggestion
	await LivechatTyped.sendMessage({
```

murtaza98: Missing await or was this intentially async?

apps/meteor/app/livechat/server/lib/LivechatTyped.ts
murtaza98: You're right... maybe we can combine all these queries into single one...

Like say so a single call to `Messages.removeByRoomIds(roomIds: string[])`

But obviously, that's out of scope for this PR

apps/meteor/server/features/EmailInbox/EmailInbox_Incoming.ts
murtaza98: Oops...I missed that. But y are we still using .then() lol :cry:

murtaza98 (approved):

d-gubert (approved):
All good on apps side

PR #30685, opened 2023-10-19
fix: i18n translations using sprintf post processor 

yash-rajpal:


ggazzo (approved):

ggazzo:
there is no need to QA this one I tested myself, actually I was about to make the SAME fix

PR #30695, opened 2023-10-19
test: License v3 - add test cases for empty limitations

ggazzo:


ee/packages/license/src/license.spec.ts
```diff
@@ -22,6 +22,19 @@ it('should not prevent if the counter is under the limit', async () => {
 	await expect(licenseManager.shouldPreventAction('activeUsers')).resolves.toBe(false);
 });
 
+it('should not prevent actions there is no limit set in the license', async () => {
```
sampaiodiego, line 25:
```suggestion
it('should not prevent actions if there is no limit set in the license', async () => {
```

sampaiodiego (changes requested):

sampaiodiego (approved):

PR #30714, opened 2023-10-22
fix: AppsProvider preventing Private Apps list from being shown on air-gapped environments

d-gubert:

Make both Marketplace and installed apps requests independent, so the installed apps can be shown even if the Marketplace apps request fails.

Also minor polishes to error message display and empty states of apps screens

### DIFF
--- a/apps/meteor/client/views/marketplace/AppsPage/AppsPageContent.tsx
+++ b/apps/meteor/client/views/marketplace/AppsPage/AppsPageContent.tsx
@@ -32,6 +32,7 @@ const AppsPageContent = (): ReactElement => {
 	const context = useRouteParameter('context');
 
 	const isMarketplace = context === 'explore';
+	const isPremium = context === 'premium';
 	const isRequested = context === 'requested';
 
 	const [freePaidFilterStructure, setFreePaidFilterStructure] = useState({
@@ -133,7 +134,8 @@ const AppsPageContent = (): ReactElement => {
 
 	const noInstalledApps = appsResult.phase === AsyncStatePhase.RESOLVED && !isMarketplace && appsResult.value.totalAppsLength === 0;
 
-	const noMarketplaceOrInstalledAppMatches = appsResult.phase === AsyncStatePhase.RESOLVED && isMarketplace && appsResult.value.count === 0;
+	const noMarketplaceOrInstalledAppMatches =
+		appsResult.phase === AsyncStatePhase.RESOLVED && (isMarketplace || isPremium) && appsResult.value.count === 0;
 
 	const noInstalledAppMatches =
 		appsResult.phase === AsyncStatePhase.RESOLVED &&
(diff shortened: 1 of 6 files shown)


.changeset/serious-cats-fetch.md
```diff
@@ -0,0 +1,5 @@
+---
+'@rocket.chat/meteor': patch
+---
+
+Fixed a problem that would prevent private apps to be shown on air-gapped environments
```
d-gubert, line 5:
```suggestion
Fixed a problem that would prevent private apps from being shown on air-gapped environments
```

apps/meteor/client/views/marketplace/AppsPage/AppsPageContent.tsx
```diff
@@ -186,6 +188,30 @@ const AppsPageContent = (): ReactElement => {
 		toggleInitialSortOption(isRequested);
 	}, [isMarketplace, isRequested, sortFilterOnSelected, t, toggleInitialSortOption]);
 
+	const getEmptyState = () => {
```
tiagoevanp, line 191: I think is better to encapsulate this function in a `React.useMemo` here...

apps/meteor/client/views/marketplace/helpers/handleAPIError.ts
```diff
@@ -8,10 +8,10 @@ const shouldHandleErrorAsWarning = (message: string): boolean => {
 };
 
 export const handleAPIError = (errorObject: unknown): void => {
-	const { message = '', error = '' } = errorObject as { message?: string; error?: string };
+	const { error = '', message = error } = errorObject as { message?: string; error?: string };
```
tiagoevanp, line 11: [visible_confusion] Why not use only error as message then?

tiagoevanp (changes requested):

apps/meteor/client/views/marketplace/helpers/handleAPIError.ts
d-gubert: [visible_confusion_as_well] The `errorObject` might contain `message`, or `error`, or `both`. In case it has both, prefer `error`

apps/meteor/client/views/marketplace/AppsPage/AppsPageContent.tsx
d-gubert: Not sure I agree, whenever the component is rerendered this function should be rerendered as well. From my point of view the operations in the function are lightweight enough (all are synchronous, simple if statements based on existing data) so that it does not justify the use of `useMemo`

apps/meteor/client/views/marketplace/helpers/handleAPIError.ts
tiagoevanp: Oh, I see... it's optional

apps/meteor/client/views/marketplace/AppsPage/AppsPageContent.tsx
tiagoevanp: Every single top change that doesn't affect the state of the return of `getEmptyState` could re-trigger another re-computation of values unnecessary... Are you sure we'll not have any scenario like this?

d-gubert: I am quite confident that the performance of encapsulating and retrieving the value of this function in a `useMemo` and simply executing the function again on every rerender will be very similar, with negligible difference. Also, memoizing the value will use more memory as React will need to store the value with the dependencies as keys for it.

If it was the case that this function would iterate over a large set of data, or require data from elsewhere asynchronously, I would say `useMemo` could help. In this specific case, I don't see benefits on using it

tiagoevanp (approved):

PR #30739, opened 2023-10-25
fix: invite link with Registration mode as Secret URL

hugocostadev:

In previous versions of Rocket Chat (before 5.4.0) it was possible to invite new external users to channels with an "Invite link" when public registration is set to secret_URL. But now it is not working because person will see the messages "New user registration is currently disabled"

Solution:
Created a new Login state `invite-register` and change the conditions to show the correct form

## Steps to test or reproduce

1. change Registration Form setting to “Secret URL”
2. Go to any channel public or private and generate invite link 
3. external user will not be able to use registration form with this link

## Further comments

There is still a bug with registration flow that it's not redirecting the user to login, but this is being handled by another PR ( 
https://github.com/RocketChat/Rocket.Chat/pull/30766 )

SUP-339


pierre-lehnen-rc (changes requested):
This would defeat the purpose of locking registration behind an invite link, as the invite hash can be omitted if the secret hash is known.

When we first implemented invite links, we made the registration form accept either the invite hash and the secret hash independently, that's probably still the best approach for this.

hugocostadev:
> This would defeat the purpose of locking registration behind an invite link, as the invite hash can be omitted if the secret hash is known.
> 
> 
> 
> When we first implemented invite links, we made the registration form accept either the invite hash and the secret hash independently, that's probably still the best approach for this.

Really? But only the invite hash was being verified...

And I didn't receive the information that the secret URL should be used only with invitations

hugocostadev:
> > This would defeat the purpose of locking registration behind an invite link, as the invite hash can be omitted if the secret hash is known.
> > When we first implemented invite links, we made the registration form accept either the invite hash and the secret hash independently, that's probably still the best approach for this.
> 
> Really? But only the invite hash was being verified...
> 
> And I didn't receive the information that the secret URL should be used only with invitations
> 
> Do you know where can I find this definition?
> 
> Checking the docs I saw:
> 
> > Secret URL: The form is only accessible using a specific URL.
> 
> https://docs.rocket.chat/use-rocket.chat/workspace-administration/settings/account-settings#registration

Forget it... I miss understood you in the first time. 
You are right, there is no need to do what I did, the problem was the condition to show the register form

pierre-lehnen-rc (approved):

PR #30745, opened 2023-10-25
feat: New permission for testing push notifications

yash-rajpal:
## Further comments

TC-998


apps/meteor/server/startup/migrations/v304.ts
```diff
@@ -0,0 +1,9 @@
+import { upsertPermissions } from '../../../app/authorization/server/functions/upsertPermissions';
```
pierre-lehnen-rc, line 1: You don't need to create migrations for permissions anymore. `upsertPermissions` is called automatically when the server starts after an update.

pierre-lehnen-rc (changes requested):

apps/meteor/server/startup/migrations/v304.ts
yash-rajpal: Done!

pierre-lehnen-rc (approved):

PR #30762, opened 2023-10-26
regression: Modal region for setup wizard

yash-rajpal:


apps/meteor/client/views/setupWizard/steps/RegisterServerStep.tsx
```diff
@@ -48,6 +49,7 @@ const RegisterServerStep = (): ReactElement => {
 			onSuccess: () => {
 				queryClient.invalidateQueries(['licenses']);
 				completeSetupWizard();
+				homeRoute.push();
```
ggazzo, line 52: why not add this redirect inside the `completeSetupWizard` ?

tiagoevanp: The redirect is already implemented in another place. Something is not triggering the re-render effect...

yash-rajpal: Upon looking more into it, the re-direct is already implemented in a hook called `useRouteLock` the issue is, this hook his not redirecting because the value of setting `Show_Setup_Wizard` is not changing upon completion of setup wizard due to TOTP and totp modal doesn't appear on setup-wizard. Looking more into this, if we can show totp modal here or skip the totp check 🤔

yash-rajpal: After talking to Tasso, we decided to implement a new meteor method which won't be guarded by totp, this method can be used to set some of the setup wizard settings.

apps/meteor/server/methods/setSetupWizardParameters.ts
```diff
@@ -0,0 +1,81 @@
...
+		);
+
+		if (settingsNotAllowed.length) {
+			throw new Meteor.Error('error-action-not-allowed', 'Editing settings is not allowed', {
+				method: 'setSetupWizardParameters',
+				settingIds: settingsNotAllowed,
+			});
+		}
+
+		console.log('changing settings - ', params);
```
tiagoevanp, line 75: this log is not necessary... right?

yash-rajpal: Definitely not, just removed it.

tiagoevanp (dismissed):

tiagoevanp (approved):

PR #30774, opened 2023-10-26
fix: Message mentions displayed in the UI are outdated after changing username

heitortanoue:

Reparse the markdown (`md`) field when updating messages that mention the username changed.

## Steps to test or reproduce

- Send a message mentioning a user.
- Change the user's username
- Check if the mentioned message is still working
(test also the `DISABLE_MESSAGE_PARSER` env config to see if it is correctly disabling the parsing)

## Further comments

SUP-378


apps/meteor/app/lib/server/functions/saveUserIdentity.ts
```diff
@@ -125,7 +126,8 @@ async function updateUsernameReferences({
 		const cursor = Messages.findByMention(previousUsername);
 		for await (const msg of cursor) {
 			const updatedMsg = msg.msg.replace(new RegExp(`@${previousUsername}`, 'ig'), `@${username}`);
-			await Messages.updateUsernameAndMessageOfMentionByIdAndOldUsername(msg._id, previousUsername, username, updatedMsg);
+			const updatedMd = messageTextToAstMarkdown(updatedMsg);
+			await Messages.updateUsernameAndMessageOfMentionByIdAndOldUsername(msg._id, previousUsername, username, updatedMsg, updatedMd);
```
yash-rajpal, line 130: Maybe change name of the update function to also include info about `md` field as a better practice 🤔

yash-rajpal (changes requested):

yash-rajpal (approved):

MarcosSpessatto (approved):
@heitortanoue Can we add some tests to cover this case and prevent that from happening in the future? We can create it in a separate PR if you prefer.

apps/meteor/app/lib/server/functions/saveUserIdentity.ts
```diff
@@ -125,7 +126,8 @@ async function updateUsernameReferences({
 		const cursor = Messages.findByMention(previousUsername);
 		for await (const msg of cursor) {
 			const updatedMsg = msg.msg.replace(new RegExp(`@${previousUsername}`, 'ig'), `@${username}`);
-			await Messages.updateUsernameAndMessageOfMentionByIdAndOldUsername(msg._id, previousUsername, username, updatedMsg);
+			const updatedMd = messageTextToAstMarkdown(updatedMsg);
+			await Messages.updateUsernameAndMessageAndMdOfMentionByIdAndOldUsername(msg._id, previousUsername, username, updatedMsg, updatedMd);
```
matheusbsilva137, line 130: I think we should check the `DISABLE_MESSAGE_PARSER` env var here. Otherwise, we'd parse messages even though this has been disabled. Maybe we could just unset the `md` field when this env var is enabled (and keep what you did for when it's not).

matheusbsilva137 (changes requested):

matheusbsilva137 (dismissed):
Code for this approach LGTM. We just gotta check if Architecture thinks we can move on with this one or if they have any other suggestion

apps/meteor/server/services/messages/service.ts
```diff
@@ -129,6 +129,36 @@ export class MessageService extends ServiceClassInternal implements IMessageServ
...
 
+	async updateUserReferences({
+		username,
+		previousUsername,
+		userId,
+	}: {
+		username: string;
+		previousUsername: string;
+		userId: string;
+	}): Promise<void> {
```
matheusbsilva137, line 140:
```suggestion
	async updateUserReferences(username: string, previousUsername: string, userId: string): Promise<void> {
```

Since the fields are all required, I think we can have normal params instead of an object here.

apps/meteor/app/lib/server/functions/saveUserIdentity.ts
```diff
@@ -119,14 +120,11 @@ async function updateUsernameReferences({
...
-		const cursor = Messages.findByMention(previousUsername);
-		for await (const msg of cursor) {
-			const updatedMsg = msg.msg.replace(new RegExp(`@${previousUsername}`, 'ig'), `@${username}`);
-			await Messages.updateUsernameAndMessageOfMentionByIdAndOldUsername(msg._id, previousUsername, username, updatedMsg);
-		}
+		await Message.updateUserReferences({
+			username,
+			previousUsername,
+			userId: user._id,
+		});
```
matheusbsilva137, line 127:
```suggestion
		await Message.updateUserReferences(username, previousUsername, user._id);
```
Changing this in order to follow the suggestion below.

matheusbsilva137 (changes requested):

matheusbsilva137 (approved):
Code LGTM. We just need to check if this approach looks fine to Architecture

PR #30780, opened 2023-10-27
fix: Remove auto reload

geekgonecrazy:
I know likely should be replaced with something.. but proposing this change to start the conversation.

Auto reload is awful.  We often end up in loops because the package just doesn't work very well.


debdutdeb:
Just a note, I think this will also disable hot reload on development. There is/should be a way to not include this in final bundles.

ggazzo:
closed by https://github.com/RocketChat/Rocket.Chat/pull/35466

PR #30799, opened 2023-10-29
fix: Message count is not changing after removing channel selection

felipe-rod123:

* Created a new hook to filter messages with the corresponding userId or roomId from the channel.
* Created a new `messages` prop on the IImportFileData.ts: `messages: Array<Pick<IImportMessage, '_id' | 'rid' | 'u'>>;`

## Issue(s)

WM-123
## Steps to test or reproduce

*Steps to reproduce:*

1. Open RC
2. Click on “Workspace” → Import
3. Click on “Import new file”
4. Select the import optione.g: Slack
5. Upload the file
6. In the table, remove some channels from the selection

*Expected Behavior:*
The number of messages should change after removing some channels (with messages) from the selection

*Actual Behavior:*
The number is fixed

image


apps/meteor/client/views/admin/import/PrepareImportPage.js
```diff
@@ -40,13 +41,31 @@ function PrepareImportPage() {
...
+		const filtered = users.filter(({ do_import }) => do_import);
+		setUsersCount(filtered.length);
+		return filtered;
+	}, [users]);
+
+	const selectedChannels = useMemo(() => {
+		const filtered = channels.filter(({ do_import }) => do_import);
+		setChannelsCount(filtered.length);
+		return filtered;
+	}, [channels]);
```
ggazzo, line 64: makes no sense creating a memo to set a state inside, if you are memoizing the array you can use array.length with no fear

ggazzo (changes requested):

apps/meteor/client/views/admin/import/PrepareImportPage.js
felipe-rod123: Good point! Thanks for the feedback!!

tapiarafael (approved):

hugocostadev (changes requested):
The reactivity in the front end is broken, after removing Channels/Users selection the message count it still not updating and navigating with the tabs the channel/ users selection resets with the initial value.

PR #30804, opened 2023-10-30
regression: do not cache the main document

rodrigok:

### DIFF
--- a/apps/meteor/app/cors/server/cors.ts
+++ b/apps/meteor/app/cors/server/cors.ts
@@ -103,7 +103,7 @@ WebAppInternals.staticFilesMiddleware = function (
 	res.setHeader('Access-Control-Allow-Origin', '*');
 	const { arch, path, url } = WebApp.categorizeRequest(req);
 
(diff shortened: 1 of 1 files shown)


apps/meteor/app/cors/server/cors.ts
```diff
@@ -103,7 +103,7 @@ WebAppInternals.staticFilesMiddleware = function (
 	res.setHeader('Access-Control-Allow-Origin', '*');
 	const { arch, path, url } = WebApp.categorizeRequest(req);
 
-	if (Meteor.isProduction && !cacheControlledRoutes.some((regexp) => regexp.test(path))) {
+	if (Meteor.isProduction && path.match(/\.(js|css|map)$/i) && !cacheControlledRoutes.some((regexp) => regexp.test(path))) {
```
pierre-lehnen-rc, line 106: Won't this also remove cache from fonts, images and any random file we may be serving?

rodrigok: @pierre-lehnen-rc good catch. It will, as it was before. We can fine tune this in another PR

pierre-lehnen-rc (approved):

PR #30807, opened 2023-10-30
refactor: Move functions out of `livechat.js`

KevLehman:
## Further comments

There's just one PR more apart from this one

Some types were changed to match the actual parameters that were being sent. Those types were created when almost everything was on JS so we expected them to have some smol issues


MarcosSpessatto (dismissed):

MarcosSpessatto (approved):

apps/meteor/app/livechat/server/lib/LivechatTyped.ts
```diff
@@ -1542,6 +1567,224 @@ class LivechatClass {
...
+
+		if (guestData?.name?.trim().length) {
+			const { _id: rid } = roomData;
+			const { name } = guestData;
+			return (
+				(await Rooms.setFnameById(rid, name)) &&
+				(await LivechatInquiry.setNameByRoomId(rid, name)) &&
+				// This one needs to be the last since the agent may not have the subscription
+				// when the conversation is in the queue, then the result will be 0(zero)
+				Subscriptions.updateDisplayNameByRoomId(rid, name)
```
sampaiodiego, line 1784: this also returns a `Promise`, no need to `await`? what is supposed to be returned here? `true` if all actions succeed?

sampaiodiego (changes requested):

apps/meteor/app/livechat/server/lib/LivechatTyped.ts
KevLehman: IIRC, we don't care about the return value for this call, as long as it doesn't fail. 👀

sampaiodiego: so does it makes sense to call 3 actions one by one and return a boolean? like

```ts
await Rooms.setFnameById(rid, name);
await LivechatInquiry.setNameByRoomId(rid, name);

// This one needs to be the last since the agent may not have the subscription
// when the conversation is in the queue, then the result will be 0(zero)
await Subscriptions.updateDisplayNameByRoomId(rid, name);

return true;
```
they could even all be in a `Promise.all` right?

KevLehman: Think we can move the first 2 to a promise.all, and leave the last one out, as the idea of it was to have it always as the last step :eyes:

That would work.

KevLehman: Actually, the promise.all worked equally well so, changed

sampaiodiego: great! I saw the comment but didn't actually got it why, since the commands were not really related, unless there was another side effect somewhere else.. well, good that you tested it =)

sampaiodiego (approved):

apps/meteor/app/livechat/server/lib/LivechatTyped.ts
KevLehman: Think this has been like this for a while, since fibers. Apparently, it was like a "sequential" chain that should return true if everything worked, and the last action was left "at last" since it may return 0, so it wont work with the `&&` that were using there. Then, this was moved to the "no fibers" world in the same fashion lol

murtaza98 (approved):

PR #30819, opened 2023-10-31
chore: add troubleshoot option to force cache flush

rodrigok:


pierre-lehnen-rc (approved):

debdutdeb (dismissed):

sampaiodiego (approved):

Gummikavalier:
Yes! Burn those cache entries in the login page with fire! :fire: :+1: 

(They give us a lot of headache, since all loginpage loop errors happen in the web client end and we cannot ask customers to start debugging it for us locally on their own time.)

PR #30843, opened 2023-11-03
fix: channel members sorting

thebluespecs:


thebluespecs:
@debdutdeb Can you have a look? Thanks

debdutdeb (changes requested):
Sort in driver function **can not** be broken, or at least we should hope not otherwise a whole lot of other things would be.

thebluespecs:
@debdutdeb broken not in general. But in this context. 
Like I mentioned switching from ascending to descending didn't do anything. 

It could be because of a number of reasons.
I could deep dive into that. 

But for now fixing it in the application layer looked fine with little overhead.

debdutdeb:
Looking at the code, it could happen because of sort is not passed to the method at all. Which would happen if sort param is not passed as part of the endpoint query (and specifically set to username). Furthermore there is no default sort (extension of last point) and no branching of default case based on real name usage setting. 

We wouldn't want to incur a technical debt here IMO.

thebluespecs:
@debdutdeb Can you have a look?

thebluespecs:
Screenshot 2023-12-05 at 8 06 13 PM

Is there a reason that the `status` is away but the `statusConnection` still stays online?

thebluespecs:
I also noticed that some users might not even have the `statusConnection` like the rocketchat user.

thebluespecs:
@ggazzo Can you have a look?

PR #30855, opened 2023-11-03
chore: Restore Managers SearchBar and add tests

MartinSchoeler:
OC-1368
OC-1363


apps/meteor/client/views/omnichannel/managers/ManagersTable.tsx
```diff
@@ -88,7 +100,7 @@ const ManagersTable = () => {
 			)}
 			{isSuccess && data.users.length > 0 && (
 				<>
-					<GenericTable>
+					<GenericTable aria-busy={text !== debouncedText} aria-live='assertive'>
```
dougfabris, line 103: why `aria-live`?

MartinSchoeler: I usually use `aria-live` to indicate that this area is `live` and prone to changes, shouldn't we be using it like that

aleksandernsilva (approved):

PR #30863, opened 2023-11-06
fix: remove pagination from slash commands

AlenDavid:

Removing not necessary pagination from apps.
Related to AECO-361 ->more than 50 slash commands, the commands after the 50 won't appear.  When installing the next app with slash commands, the current action is fetching from a paginated endpoint and is not asking for items after the first page.

## Issue(s)
https://rocketchat.atlassian.net/browse/CONN-111


casalsgh:
Related to internal task AECO-361

ggazzo (changes requested):
this is actually a break change, the pagination should be properly handled on the client side

PR #30868, opened 2023-11-06
feat: Add `push.info` endpoint to retrieve push gateway info

matheusbsilva137:

 - Add `push.info` (GET) endpoint to enable all authenticated users in the workspace to retrieve push gateway info.

## Steps to test or reproduce

**_Example request (using cURL):_**
```
curl 'http://localhost:3000/api/v1/push.info' \
  -H 'X-Auth-Token: G0Hi2k7AQnzmZenFQjDhShla4XfY2JefugNgjOhX8gA' \
  -H 'X-User-Id: kXQ3GFsuKBqtS6z9M'
```

**_Example response:_**
```
{
  "pushGatewayEnabled": true,
  "defaultPushGateway": false,
  "success": true
}
```
**Note:** `defaultPushGateway` should be `true` if the workspace is using the default push gateway (`https://gateway.rocket.chat`) or `false` otherwise.

## Further comments

TC-1014


yash-rajpal (dismissed):

MarcosSpessatto (approved):

sampaiodiego:
may I ask why they need that info?

reinaldonetof:
> may I ask why they need that info?

Hey, @sampaiodiego, the mobile team is currently working on implementing a new feature: Troubleshoot Notification. One of the key aspects of this task is determining whether the push gateway is enabled and if it's a custom push gateway. Importantly, all types of users should have access to this setting, not just admins or specific permission holders.

Screen Shot 2023-11-08 at 10 12 38

If you have any alternative approaches to make this setting 'public,' that would be fantastic. Currently, the mobile team doesn't need to know the specific details of the push gateway; they only need to ascertain if it's enabled and whether it's a custom push gateway or not. Additionally, we can access this parameter through settings or the API. Given that this is a new feature, we can obtain this value the same way the backend will handle it.

sampaiodiego:
thx @reinaldonetof .. that's very good info.. so for this feature I think we'll need some additional endpoints, either to provide extra information or to "try" things.. so maybe the best solution would be creating this endpoint (or multiple endpoints) first, so in this endpoint it would make this checks like and could return a payload like:

```
return {
  usingPushGateway: true,
  pushGatewayModified: false,
  pushGatewayPing: 123, // I'm being creative here =P
}
```

what do you guys think?

reinaldonetof:
> so for this feature I think we'll need some additional endpoints, either to provide extra information or to "try" things (like send a test push notification)..

About the `send a test push notification` @yash-rajpal did this: https://rocketchat.atlassian.net/browse/TC-998 // https://github.com/RocketChat/Rocket.Chat/pull/30745
I think this is ok, considering that we can enable/disable the button depending the permissions of the user

> maybe the best solution then would be creating this endpoint (or multiple endpoints) first, so in this endpoint it would make this checks like and could return a payload like:

Awesome, imho I was thinking in something like this. It's up to you.

packages/rest-typings/src/v1/push.ts
```diff
@@ -65,4 +65,10 @@ export type PushEndpoints = {
 			};
 		};
 	};
+	'/v1/push.info': {
+		GET: () => {
+			pushGateWayEnabled: boolean;
```
KevLehman, line 70:
```suggestion
			pushGatewayEnabled: boolean;
```

KevLehman (dismissed):

packages/rest-typings/src/v1/push.ts
matheusbsilva137: oopsss, thanks :hugs:

sampaiodiego (approved):

PR #30883, opened 2023-11-07
fix: Avoid sending empty objects as errors from meteor.methods result

KevLehman:

### DIFF
--- a/apps/meteor/app/api/server/v1/misc.ts
+++ b/apps/meteor/app/api/server/v1/misc.ts
@@ -492,6 +492,18 @@ declare module '@rocket.chat/rest-typings' {
 	}
 }
 
+const parseErrorObject = (err: any) => {
+	if (err instanceof Error) {
+		return {
+			message: err.message,
+			stack: process.env.TEST_MODE === 'true' ? err.stack : undefined,
+		};
+	}
+
+	// Meteor errors can be EJSONed without problems
+	return err;
+};
+
 const mountResult = ({
 	id,
 	error,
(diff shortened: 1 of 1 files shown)


apps/meteor/app/api/server/v1/misc.ts
```diff
@@ -492,6 +492,18 @@ declare module '@rocket.chat/rest-typings' {
 	}
 }
 
+const parseErrorObject = (err: any) => {
+	if (err instanceof Error) {
```
sampaiodiego, line 496: should we use `isMeteorError` instead?

https://github.com/RocketChat/Rocket.Chat/blob/de50690c839f4ad75baf9f7e7a29e45a1119fcbe/packages/core-services/src/MeteorError.ts#L22

KevLehman: huh didn't know that existed. Let's try

apps/meteor/app/api/server/v1/misc.ts
```diff
@@ -492,6 +493,18 @@ declare module '@rocket.chat/rest-typings' {
 	}
 }
 
+const parseErrorObject = (err: any) => {
+	if (!isMeteorError(err)) {
+		return {
+			message: err.message,
+			stack: process.env.TEST_MODE === 'true' ? err.stack : undefined,
```
sampaiodiego, line 500:
```suggestion
			...(process.env.TEST_MODE === 'true' ? { stack:  err.stack } : {}),
```

sampaiodiego (changes requested):

PR #30886, opened 2023-11-07
chore: Improve agent tests

MartinSchoeler:
OC-1396
OC-1397
OC-1398
OC-1399

### DIFF
--- a/apps/meteor/tests/e2e/omnichannel/omnichannel-agents.spec.ts
+++ b/apps/meteor/tests/e2e/omnichannel/omnichannel-agents.spec.ts
@@ -49,6 +49,7 @@ test.describe.serial('omnichannel-agents', () => {
 
 			await poOmnichannelAgents.inputSearch.fill('user1');
 			await expect(poOmnichannelAgents.firstRowInTable).toBeVisible();
+			await expect(poOmnichannelAgents.firstRowInTable).toContainText('user1');
 		});
 
 		await test.step('expect remove "user1" as agent', async () => {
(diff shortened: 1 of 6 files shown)


apps/meteor/tests/e2e/omnichannel/omnichannel-agents.spec.ts
```diff
@@ -1,44 +1,104 @@
...
+			await poOmnichannelAgents.firstRowInTable.click();
+			await poOmnichannelAgents.btnRemove.click();
+		});
 	});
 
-	test('expect remove "user1" as agent', async () => {
-		await poOmnichannelAgents.inputSearch.fill('user1');
-		await poOmnichannelAgents.btnDeletefirstRowInTable.click();
-		await poOmnichannelAgents.btnModalRemove.click();
+	test('OC - [EE] Manage Agents - Edit ', async ({ page }) => {
```
aleksandernsilva, line 88: This test seem to be incomplete 🤔 
What are we testing here? I don't see a counterpart in the CE test.

apps/meteor/tests/e2e/omnichannel/omnichannel-agents.spec.ts
```diff
@@ -1,44 +1,104 @@
...
+		await api.post('/settings/Omnichannel_enable_department_removal', { value: false }).then((res) => expect(res.status()).toBe(200));
+	});
+
+	test('OC - Manage Agents - Add, search and remove using table', async ({ page }) => {
+		await test.step('expect "user1" be first ', async () => {
+			await poOmnichannelAgents.inputUsername.type('user');
+			await expect(page.locator('role=option[name="user1"]')).toContainText('user1');
+
+			await poOmnichannelAgents.inputUsername.type('');
+		});
```
aleksandernsilva, line 43: This step seems redundant considering the next step does almost exactly the same actions.

apps/meteor/tests/e2e/omnichannel/omnichannel-agents.spec.ts
```diff
@@ -1,44 +1,104 @@
...
+
+		await test.step('expect add "user1" as agent', async () => {
+			await poOmnichannelAgents.inputUsername.type('user1');
+			await page.locator('role=option[name="user1"]').click();
+			await poOmnichannelAgents.btnAdd.click();
+
+			await poOmnichannelAgents.inputSearch.fill('user1');
+			await expect(poOmnichannelAgents.firstRowInTable).toBeVisible();
```
aleksandernsilva, line 51: Could you improve this assertion? It covers the search feature poorly imo. 
It works fine if the workspace is clean (which is likely the case on CI env), but we are just looking for *any* first row, not the correct first row. We could end up with a false positive.

aleksandernsilva (changes requested):

PR #30910, opened 2023-11-09
fix: change the push sound sent when the push is from video conference

GleidsonDaniel:

Corrects the push sound when the push is from videoconf, currently the system default sound, however videoconf has a specific sound for pushes.

## Further comments

Jira: [VC-155]

[VC-155]: https://rocketchat.atlassian.net/browse/VC-153


MarcosSpessatto (approved):

GleidsonDaniel:
> Please add the change set

I don't think there should be a changeset, it's a fix of a PR that was merged two days ago and they left this detail in the PR, the original PR that should have the changeset.

pierre-lehnen-rc (approved):
For reference: The push gateway will change `sound` to `soundname` automatically on android notifications.

The code that sends android notifications directly from rocket.chat also reads the `sound` attribute to fill the `soundname`.

pierre-lehnen-rc (changes requested):
missing changeset

pierre-lehnen-rc (approved):

PR #30914, opened 2023-11-09
chore: remove redundant indexes

sampaiodiego:

the indexes removed are actually redundant and are prefixes of other compound indexes.

Unused indexes consume storage space and increase write latency.


debdutdeb:
Don't we need migrations to actually remove those indexes from mongo?

sampaiodiego:
> Don't we need migrations to actually remove those indexes from mongo?

yes, we'd need a migration to remove them.. we're avoiding migrations during minor releases though, so we'd need to "schedule" a migration for the next major..

ggazzo (approved):

sampaiodiego:
already created this as a follow up https://github.com/RocketChat/Rocket.Chat/pull/30918

PR #30948, opened 2023-11-14
regression: adjust license

ggazzo:

### DIFF
--- a/apps/meteor/app/cloud/server/functions/syncWorkspace/syncCloudData.ts
+++ b/apps/meteor/app/cloud/server/functions/syncWorkspace/syncCloudData.ts
@@ -1,4 +1,5 @@
 import type { Cloud, Serialized } from '@rocket.chat/core-typings';
+import { DuplicatedLicenseError } from '@rocket.chat/license';
 import { serverFetch as fetch } from '@rocket.chat/server-fetch';
 import { v, compile } from 'suretype';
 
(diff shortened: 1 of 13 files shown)


apps/meteor/app/cloud/server/functions/syncWorkspace/syncCloudData.ts
```diff
@@ -74,14 +75,41 @@ export async function syncCloudData() {
...
+		/**
+		 * If some of CloudWorkspaceAccessError and CloudWorkspaceRegistrationError happens, makes no sense to run the legacySyncWorkspace
+		 * because it will fail too.
+		 * The DuplicatedLicenseError license error is also ignored because it is not a problem. the Cloud is allowed to send the same license twice.
+		 */
+		switch (true) {
+			case err instanceof CloudWorkspaceAccessError:
+			case err instanceof CloudWorkspaceRegistrationError:
+			case err instanceof DuplicatedLicenseError:
+				SystemLogger.info({
```
KevLehman, line 94:
```suggestion
				SystemLogger.error({
```

KevLehman (approved):

apps/meteor/app/cloud/server/functions/syncWorkspace/syncCloudData.ts
ggazzo: Considering the comment above, those errors are suppressed since everything is under control, so I decide to handle them as info (if the admin wants to follow all the steps), but since we handled its not an exception anymore

nmagedman:
It appears to me that when you do a change like this:

```
commit 4354cab319bba11288c90dccaea58b13492ad396
Author: Guilherme Gazzo <guilhermegazzo@gmail.com>
Date:   Tue Nov 14 02:32:34 2023 -0300

    ...
    * fix: Wrong index for Analytics
    ...

diff --git a/apps/meteor/server/models/raw/Analytics.ts b/apps/meteor/server/models/raw/Analytics.ts
index 254ff32965..4e95cadebb 100644
--- a/apps/meteor/server/models/raw/Analytics.ts
+++ b/apps/meteor/server/models/raw/Analytics.ts
@@ -14,7 +14,7 @@ export class AnalyticsRaw extends BaseRaw<IAnalytic> implements IAnalyticsModel
 	}

-		return [{ key: { date: 1 } }, { key: { 'room._id': 1, 'date': 1 }, unique: true }];
+		return [{ key: { date: 1 } }, { key: { 'room._id': 1, 'date': 1 }, unique: true, partialFilterExpression: { type: 'rooms' } }];
 	}

 	saveMessageSent({ room, date }: { room: IRoom; date: IAnalytic['date'] }): Promise<Document | UpdateResult> {
```

it can lead to the following fatal error on startup:

```
systemd[1]: Starting Rocket.Chat service...
rocket.chat[2187053]: Some indexes for collection 'rocketchat_analytics' could not be created:
rocket.chat[2187053]:         An existing index has the same name as the requested index. When index names are not specified, they are auto generated and can cause conflicts. Please refer to our documentation. Requested index: { v: 2, unique: true, key: { room._id: 1, date: 1 }, name: "room._id_1_date_1", partialFilterExpression: { type: "rooms" } }, existing index: { v: 2, unique: true, key: { room._id: 1, date: 1 }, name: "room._id_1_date_1" }
```

That is, you can't just simply change the options for an index.  You have to somehow handle the transition from the _old_ options to the _new_ options.

PR #30958, opened 2023-11-14
docs: fix local host port

JoshMcCullough:

- RC starts on port 3000, not 4000.


JoshMcCullough:
Sorry, I was looking at the wrong section.

PR #30960, opened 2023-11-14
fix: Livechat Reactivity issues with createOrUpdateGuest

MartinSchoeler:
Fixes issue with updating guest data using the Livechat API, the widget would not update after a call

OC-1413


rodrigok:
@casalsgh 6.4.6 was released without this PR, I'll remove it from the project

aleksandernsilva (approved):

PR #30969, opened 2023-11-14
fix: user and channels dropdowns not closing when clicking outside

guijun13:


ggazzo:
if this is a regression it should have no `changeset` otherwise this should be labeled as `fix`

PR #30996, opened 2023-11-16
chore: adapt read receipts feature to work without oplog on msgs

MarcosSpessatto:

ARCH-1235
ARCH-1236


apps/meteor/ee/server/lib/message-read-receipt/ReadReceipt.js
```diff
@@ -61,7 +65,10 @@ export const ReadReceipt = {
 		// mark message as read if the sender is the only one in the room
 		const isUserAlone = (await Subscriptions.countByRoomIdAndNotUserId(roomId, userId)) === 0;
 		if (isUserAlone) {
-			await Messages.setAsReadById(message._id);
+			const result = await Messages.setAsReadById(message._id);
+			if (result.modifiedCount > 0) {
+				void api.broadcast('notify.messagesRead', roomId);
```
sampaiodiego, line 70: this being a single message I guess we could use `message.sent`

packages/core-services/src/events/Events.ts
```diff
@@ -93,6 +93,13 @@ export type EventSignatures = {
 	'notify.deleteCustomSound'(data: { soundData: ICustomSound }): void;
 	'notify.updateCustomSound'(data: { soundData: ICustomSound }): void;
 	'notify.calendar'(uid: string, data: ICalendarNotification): void;
+	'notify.messagesRead'(
+		rid: string,
+		data: {
+			until?: Date;
+			tmid?: string;
+		},
```
sampaiodiego, line 101: this could be a plain object, just like the `messagesRead` Streamer, making both the same would be simpler IMO, like:
```
{
  rid: string;
  until?: Date;
  tmid?: string;
}
```

sampaiodiego (changes requested):

sampaiodiego (approved):

PR #31003, opened 2023-11-17
chore: lower-case meteor package name

JoshMcCullough:

Simple -- just lower-case the word "METEOR" `apps/meteor/.meteor/release` so we can easily use it for automation (e.g. just install the version exact package found in this file). Why have it upper-case if that is not valid NPM syntax?


debdutdeb:
image
I think this is a meteor requirement, i.e. it expects in capitals.

PR #31005, opened 2023-11-17
fix: fixing the href message link parsing

AllanPazRibeiro:

  This PR addresses a critical issue regarding broken links within messages sent by the 'rocket.cat' bot. The function `parseLinkUrl` has been introduced to handle the parsing of URLs containing HTML-encoded characters within the messages.
  The `parseLinkUrl` function uses the DOMParser to decode the HTML-encoded URL, followed by the URL constructor to parse it accurately. This helps in ensuring that links sent by the 'rocket.cat' bot are correctly formatted and functional.
  The primary goal is to fix the broken links and ensure consistent formatting in the messages sent by the bot.

## Issue(s)

[Jira Link](https://rocketchat.atlassian.net/browse/AECO-422)

## Steps to test or reproduce

1. Using a secondary account, open the marketplace app list
2. Request the installation of an app
3. Now using an admin account, you will be receiving a message from rocket.cat with the link to the requested app
4. This link should lead you to the requests tab on the app details page

### DIFF
--- a/.changeset/rare-cobras-poke.md
+++ b/.changeset/rare-cobras-poke.md
@@ -1,0 +1,5 @@
+---
+'@rocket.chat/gazzodown': major
+---
+
+Fixing the message link url parser
--- a/.changeset/rare-cobras-poke.md
+++ b/.changeset/rare-cobras-poke.md
@@ -1,5 +1,0 @@
----
-'@rocket.chat/gazzodown': major
----
-
-Fixing the message link url parser
--- a/.changeset/great-queens-joke.md
+++ b/.changeset/great-queens-joke.md
@@ -1,0 +1,5 @@
+---
+'@rocket.chat/meteor': patch
+---
+
+Fix rocket.cat's app request message escaping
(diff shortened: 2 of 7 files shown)


tassoevan (changes requested):
It feels like an overkill, since we shouldn't be feeding pre-escaped URLs. Also, the situation when this happens is, as far as I understand, limited to one automated message from rocket.cat. It might be better to fix the message itself.

.changeset/rare-cobras-poke.md
```diff
@@ -0,0 +1,5 @@
+---
+'@rocket.chat/gazzodown': major
```
tassoevan, line 2: I don't think that fixing a behavior of a component in one package is enough to release a new major version of it.

tassoevan (changes requested):

AllanPazRibeiro:
> It feels like an overkill, since we shouldn't be feeding pre-escaped URLs. Also, the situation when this happens is, as far as I understand, limited to one automated message from rocket.cat. It might be better to fix the message itself.

The message itself looks just fine, but somewhere between the message being sent and the message being parsed, the link is encoded with HTML-encoded characters, it may sound overkill, but I don't think it will break anything as long as the url is valid

But I am very open to new ideas

.changeset/rare-cobras-poke.md
AllanPazRibeiro: my bad, I am not fully comfortable with this change set structure so I may set this as a major by mistake

tassoevan (approved):

PR #31008, opened 2023-11-17
regression: Use `this.getLicense` to instead of `this._license` when calculating shouldPreventAction map

KevLehman:
NBJ-540


ee/packages/license/src/license.ts
```diff
@@ -145,6 +145,7 @@ export class LicenseManager extends Emitter<LicenseEvents> {
 
 	private invalidateLicense(): void {
 		this._valid = false;
+		this.states.clear();
```
KevLehman, line 148: @ggazzo shouldn't the call on "clearLicenseData" do the same?

PR #31048, opened 2023-11-22
chore: remove pre-push git hooks

debdutdeb:

Removes the pre push hooks that runs variety of tests before pushing to remote. 

## Further comments

Existing clones, run `rm .git/hooks/husky*` to remove existing hooks.

### DIFF
--- a/.husky/pre-push
+++ b/.husky/pre-push
@@ -1,4 +1,0 @@
-[ "$HUSKY" = "0" ] && echo "skipping push hook" && exit 0
-
-yarn lint && \
-yarn testunit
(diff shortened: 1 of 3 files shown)


KevLehman (approved):

geekgonecrazy (approved):
Makes for a very bad developer UX.  Once people learn how they end up skipping it.  I’m in favor of it going away.  

People have to wait for this and then it to happen again in CI.  So why wouldn’t they skip it?

Not to mention this is heavy and not everyone has resources to spare

rodrigok (approved):

sampaiodiego (approved):
there are some references left on some workflow files, but we can remove them later if needed

PR #31050, opened 2023-11-22
chore: Reduce amount of logs on info level

rodrigok:


MarcosSpessatto (dismissed):

packages/logger/src/getPino.ts
```diff
@@ -19,9 +19,9 @@ function logMethod(this: Logger, args: unknown[], method: any): void {
 const mainPino = pino({
 	hooks: { logMethod },
 	customLevels: {
-		http: 35,
-		method: 35,
-		subscription: 35,
+		http: 20,
```
sampaiodiego, line 22: do we want to "downgrade" these to DEBUG level? INFO is the correct level IMO

rodrigok: I makes the other logs almost unusable in the middle of lots of HTTP calls. For me HTTP call info is more for a debug purpose than to see relevant information rather than only warns and errors

sampaiodiego: there is a specific level for errors and warnings, if you want to see only them.

regarding requests logs I guess it is a matter of opinion.. we got no complaints what so ever when we added them to INFO level a while ago.. if I remember correctly it was a costumer request to show them without the noise (and performance degradation) of DEBUG level..

also another point, not sure if this can be considered a breaking change..

rodrigok: Maybe we should have another level like "trace" to have what we call "debug" today. But I agree, it's almost a breaking change

sampaiodiego (approved):

PR #31096, opened 2023-11-28
regression: seats cap contextual bar

hugocostadev:

Replace old reached seats cap modal with contextual bar

image

## Further comments

NBJ-563


apps/meteor/ee/client/views/admin/users/UserPageHeaderContentWithSeatsCap.tsx
```diff
@@ -68,7 +50,7 @@ const UserPageHeaderContentWithSeatsCap = ({ activeUsers, maxActiveUsers }: User
 					{t('New_user')}
 				</Button>
 				{isCreateUserDisabled && (
-					<Button is='a' href={manageSubscriptionUrl} target='_blank' rel='noopener noreferrer' primary>
+					<Button mis={8} is='a' href={manageSubscriptionUrl} target='_blank' rel='noopener noreferrer' primary>
```
juliajforesti, line 53: When a button is actually an external link, prefer:
- adding an `onClick` to the button with `useExternalLink` hook rather than `Button is='a' (...)`
- add `role='link` for semantics

Doing that, you won't need to manually add the margin - it will receive from `ButtonGroup` correctly.

apps/meteor/client/views/admin/users/AdminUserUpgrade.tsx
```diff
@@ -0,0 +1,36 @@
...
+				<States>
+					<StatesIcon name='warning' />
+					<StatesTitle>{t('Seat_limit_reached')}</StatesTitle>
+					<StatesSubtitle>{t('Seat_limit_reached_Description')}</StatesSubtitle>
+				</States>
+			</ContextualbarScrollableContent>
+			<ContextualbarFooter>
+				<ButtonGroup stretch>
+					<Button onClick={() => router.navigate('/admin/users')}>{t('Cancel')}</Button>
+					<Button primary onClick={() => openExternalLink(manageSubscriptionUrl)}>
```
juliajforesti, line 27: add `role='link` for semantics

juliajforesti (changes requested):

apps/meteor/ee/client/views/admin/users/UserPageHeaderContentWithSeatsCap.tsx
hugocostadev: Thanks for your review! It's fixed now

juliajforesti (commented):
Can you vertically centralize the empty state on the contextual bar? That's the standard...

PR #31102, opened 2023-11-29
chore!: remove unused oauth setting

hugocostadev:

The settings  `Accounts_OAuth_Custom_Button_Label_Color` , `Accounts_OAuth_Custom_Button_Color` , `*_button_label_color` and `*_button_color`  was not being used anymore for oauth buttons

## Further comments

https://rocketchat.atlassian.net/browse/SUP-395


elfajriharsuda (approved):

ggazzo (changes requested):
how about a migration to delete? otherwise old instances will show the setting with no translations

hugocostadev:
> how about a migration to delete? otherwise old instances will show the setting with no translations

Good idea! I'll work on that! Thanks 🚀

apps/meteor/server/startup/migrations/v305.ts
```diff
@@ -0,0 +1,18 @@
...
+	version: 305,
+	async up() {
+		const customOauthServicesButtonColors = await Settings.find(
+			{ _id: /Accounts_OAuth_.+button.+color$/ },
+			{ projection: { _id: 1 } },
+		).toArray();
+
+		for await (const setting of customOauthServicesButtonColors) {
+			await Settings.removeById(setting._id);
+		}
```
ggazzo, line 16: why not just a `deleteMany`?

ggazzo (changes requested):

apps/meteor/server/startup/migrations/v305.ts
hugocostadev: Much better indeed! Thanks

ggazzo (dismissed):

debdutdeb (dismissed):

ggazzo (approved):

PR #31128, opened 2023-12-01
fix: Mentioning discussion displays its ID instead of name

abhinavkrin:

Fix: Mentioning discussions are appearing as ID

## Issue(s)

If we mention a discussion, it is showing the discussion internal ID and not the mention
name.
## Steps to test or reproduce

1. Create a discussion
2. Join the discussion and send a message.
3. Open a chat
4. Type # and the name of the discussion

TC-878


packages/core-typings/src/IMessage/IMessage.ts
```diff
@@ -124,6 +124,10 @@ export type MessageMention = {
 	_id: string;
 	name?: string;
 	username?: string;
+<<<<<<< Updated upstream
+=======
+	fname?: string; // incase of channel mentions
+>>>>>>> Stashed changes
```
abhinavkrin, line 130: @yash-rajpal I think this might be a mistake. It would cause build to fail.

yash-rajpal: Ooops, how did it get there 😓

gabriellsh (approved):

tassoevan (approved):

PR #31134, opened 2023-12-01
fix: connected users search on spotlight for user mentions

yash-rajpal:
## Further comments

TC-1050
TC-1052


matheusbsilva137 (dismissed):

matheusbsilva137 (approved):

elfajriharsuda (approved):

apps/meteor/server/lib/spotlight.js
```diff
@@ -237,7 +237,7 @@ export class Spotlight {
 		}
 
 		if (users.length === 0 && canListOutsiders) {
-			const exactMatch = await Users.findOneByUsernameIgnoringCase(text, {
+			const exactMatch = await Users.findOneByUsernameIgnoringCase(usernames, {
```
matheusbsilva137, line 240:
```suggestion
		if (users.length === 0 && canListOutsiders && text) {
			const exactMatch = await Users.findOneByUsernameIgnoringCase(text, {
```

If I correctly understood the flow for this function, this block is supposed to retrieve exact matches for what the user has typed -- so we should keep using the `text` param here. Also, `findOneByUsernameIgnoringCase` first param should be a string, not an array (such as `usernames`).
Just checking if text is defined should avoid the error thrown in this model method, so all the rest will work fine.

matheusbsilva137 (changes requested):

matheusbsilva137 (approved):

PR #30551, opened 2023-10-02
feat: subscription page

hugocostadev:
## Proposed changes (including videos or screenshots)

New Manage subscription page to allow admin users check usage of some services and manage the subscription of the workspace

**Changes:**

**Front-end:**
- new page, route, admin item and Cards components for the new `Manage subscription` admin page
- New `FramedIcon` component in `ui-client` package, should be moved to fuselage in the future
- Added `Box` component props to `Card` components to give more flexibility on styling card elements due to design specs
- new `useIsSelfHosted.ts` hook to check if server is running on cloud or not
- changes to `useLicense.ts` hook to use the new `license.info` endpoint
- new `getDaysLeft.ts` util function to check quantity of days given a specific date
- new `getPlanName.ts` util function to retrieve the plan of the license given some condition related to activeModules, this function is necessary to prevent code repetition
- fix `useAdministrationItems.spec.ts` unit test
- moved `useActiveConnections` hook to `hooks` folder to be reused in other places
 - new GO links
 - renamed `useAnalyticsObject.ts` hook to `useStatistics.ts` to better describe its usage and moved to more appropiate folder `hooks`
 - new i18n translations
 - new custom hook `useCheckoutURL.ts` to get the new endpoint `https://billing.rocket.chat/v2/checkout`

**Back-end:**
- new endpoint `billing.checkoutUrl` to get the checkout url provided by cloud
- new function `getCheckoutUrl.ts` to fetch data from new billing service

image
image

TODO: 
- [x] Integrate `license.info` endpoint instead of `license.get`
- [x] Add MAC data to `MACCard` and `CountMACCard` (waiting https://github.com/RocketChat/Rocket.Chat/pull/30222 to be merged)
- [x] implement or remove Mobile Push Notification Card
- [x] make requested changes from UX review

## Further comments

[NBJ-289](https://rocketchat.atlassian.net/browse/NBJ-289)

[NBJ-289]: https://rocketchat.atlassian.net/browse/NBJ-289?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #31153, opened 2023-12-04
fix: New `custom-roles` license module isn't properly checked

matheusbsilva137:

 - Check the new `custom-roles` license module in order to enable the custom roles feature (on the server and client);
 - Update the "role deletion" modal's text to consider the new plans (this modal is displayed for Community, Starter and Pro plans before deleting a role).

## Issue(s)

Introduced by #30551

## Steps to test or reproduce

Go to **Workspace > Permission > New Role**, the behavior of this button and editing roles should follow what is defined for each license (this feature is only enabled for the Enterprise plan). The `roles.create` and `roles.update` endpoints should also be blocked for non-Enteprise workspaces.

## Further comments

NBJ-576


gabriellsh (approved):
lgtm!

PR #31164, opened 2023-12-06
fix: mobile users receives two push notifications when a direct call is started

pierre-lehnen-rc:
## Further comments

VC-158


apps/meteor/app/lib/server/lib/sendNotificationsOnMessage.js
```diff
@@ -48,12 +48,13 @@ export const sendNotification = async ({
...
-					status: 1,
-					statusConnection: 1,
-					username: 1,
+					'active': 1,
+					'emails': 1,
+					'language': 1,
+					'status': 1,
+					'statusConnection': 1,
+					'username': 1,
+					'settings.preferences.enableMobileRinging': 1,
```
sampaiodiego, line 57: you should add this field to the lookup as well, this is only a fallback

add here:
https://github.com/RocketChat/Rocket.Chat/blob/40e5d4d0a0c62f95f517d2364a38f1e99efaf828/apps/meteor/app/lib/server/lib/sendNotificationsOnMessage.js#L191

sampaiodiego (changes requested):

sampaiodiego (approved):

PR #31167, opened 2023-12-06
fix: Don't try to listen to voip stream when not enabled

MartinSchoeler:
fixes this annoying issue 
image

### DIFF
--- a/apps/meteor/ee/client/hooks/useVoipClient.ts
+++ b/apps/meteor/ee/client/hooks/useVoipClient.ts
@@ -39,23 +39,25 @@ export const useVoipClient = (): UseVoipClientResult => {
 	const isEE = useHasLicenseModule('voip-enterprise');
 	const voipEnabled = settingVoipEnabled && voipConnectorEnabled;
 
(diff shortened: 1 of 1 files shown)


apps/meteor/ee/client/hooks/useVoipClient.ts
```diff
@@ -39,23 +39,25 @@ export const useVoipClient = (): UseVoipClientResult => {
...
-			subscribeToNotifyLoggedIn(`voip.statuschanged`, (enabled: boolean): void => {
-				setVoipConnectorEnabled(enabled);
-			}),
-		[setResult, setVoipConnectorEnabled, subscribeToNotifyLoggedIn],
-	);
+	useEffect(() => {
+		if (!voipEnabled || !user) {
+			return;
+		}
+		return subscribeToNotifyLoggedIn(`voip.statuschanged`, (enabled: boolean): void => {
```
ggazzo, line 46: what about checking if you are agent ?

elfajriharsuda (approved):

PR #31172, opened 2023-12-06
chore: change e2e placeholder

heitortanoue:

Change `E2E_message_encrypted_placeholder` in i18n to "_This message is end to end encrypted because of the room settings at the time it was sent, you won't be able to access it_".

## Further comments

PROD-200


dougfabris:
@heitortanoue is it still a valid PR?

PR #31181, opened 2023-12-07
fix: Notifications for thread messages are sent even after thread has been read

matheusbsilva137:

 - Clear notification queue after they thread messages have been read by a user;

## Steps to test or reproduce

Start a thread with two or more users, and keep it open for both of them while no new message is sent in the main room.
Send a thread message with user A, then read it with user B (by keeping the thread open or opening the thread after the message is sent).
**Current behavior:** a push notification is sent after some minutes;
**Expected behavior:** no push notifications should be sent since the messages have been read (the queue can be checked in the `rocketchat_notification_queue` db collection). Notifications should be cleared when using `getThreadMessages` or `readThreads` methods.

## Further comments

SUP-386


heitortanoue (dismissed):

sampaiodiego (changes requested):
I think the actual solution should be calling `await NotificationQueue.clearQueueByUserId(uid);` from `readThread` function:
https://github.com/RocketChat/Rocket.Chat/blob/b82de9b1bd72eaf3fffcd79f6a0d5d09f5e57eb9/apps/meteor/app/threads/server/functions.ts#L61

the way it works we just clear the whole queue if the user is active in the system.. we don't actually care of what is in the queue

sampaiodiego:
> > I think the actual solution should be calling `await NotificationQueue.clearQueueByUserId(uid);` from `readThread` function
> 
> Ohhh agreed 🤔 that's by the definition of how our notifications should work 😬 not checking "scope" Thanks!

yes, exactly.. if the user has made any action within the application we just wipe out the notification queue..

the idea is that since the user is active using the application we don't need to bother him by sending push or email notifications.

sampaiodiego (approved):

MarcosSpessatto (changes requested):
I think we must do the same here, don't we? https://github.com/RocketChat/Rocket.Chat/blob/develop/apps/meteor/ee/server/local-services/message-reads/service.ts#L11

Ideally we should unify everything in a single place, but we can do that in another PR.

PR #31189, opened 2023-12-07
fix: remove @rocket.chat/license from ddp-client deps

ggazzo:


ee/packages/license/src/definition/ILicenseTag.ts
```diff
@@ -1,4 +1 @@
-export interface ILicenseTag {
-	name: string;
-	color: string;
-}
+export { ILicenseTag } from '@rocket.chat/core-typings';
```
sampaiodiego, line 1: don't you think that re-exporting everything from the license package again can lead to the same issue? someone importing the typings from the license package instead of the core-typings package.

I'd say to let them there only.

sampaiodiego (changes requested):

PR #31191, opened 2023-12-07
fix: Updating ComposerPopupProvider to accept entities with prid

Gustrb:
[BUG] - Mentioning discussions are appearing as ID [WIP]

If we mention a discussion, it is showing the discussion internal ID and not the mention
name.

Expected Behavior:
1. It should show the name of the discussion
Actual Behavior:
1. It's appearing as an ID

## Steps to test or reproduce

Steps to reproduce:
1. Create a discussion
2. Open a chat
3. Type # and the name of the discussion


Gustrb:
I just found out that even though the text is now correct, it is not mentioning the discussion correctly, does anyone where does this mention processing part happens?

elfajriharsuda (approved):

PR #31204, opened 2023-12-11
fix: Remove non-monitor users from monitors list before saving unit

KevLehman:


lmauromb (approved):

sampaiodiego (approved):
very happy to see tests being added along with the fix 🤗

PR #31211, opened 2023-12-11
refactor: convert canned response callback

sampaiodiego:
## Issue(s)

ARCH-1252


apps/meteor/ee/server/hooks/message/BeforeSaveCannedResponse.ts
```diff
@@ -0,0 +1,106 @@
...
+
+	async replacePlaceholders({ message, room, user }: { message: IMessage; room: IRoom; user: IUser }): Promise<IMessage> {
+		// If the feature is disabled, return the message as is
+		if (!BeforeSaveCannedResponse.enabled) {
+			return message;
+		}
+
+		if (!settings.get('Canned_Responses_Enable')) {
+			return message;
+		}
```
KevLehman, line 66: hmmm, wouldn't be better to use stick with one? Either use setting on constructor & listen to changes or use the setting directly 👀

apps/meteor/ee/server/hooks/message/BeforeSaveCannedResponse.ts
```diff
@@ -0,0 +1,106 @@
...
+				return user;
+			}
+
+			return this.getUser(agentId);
+		};
+
+		message.msg = Object.keys(placeholderFields).reduce((messageText, field) => {
+			const placeholderConfig = placeholderFields[field as keyof typeof placeholderFields];
+
+			const from = placeholderConfig.from === 'agent' ? getAgent(agentId) : this.getVisitor(room.v._id);
```
KevLehman, line 97: no await? 👀

sampaiodiego: yeah, I realized that when writing tests 😬

apps/meteor/ee/server/hooks/messages/BeforeSaveCannedResponse.ts
```diff
@@ -0,0 +1,116 @@
...
+		from: 'agent',
+		dataKey: 'name',
+	},
+	'agent.email': {
+		from: 'agent',
+		dataKey: 'emails[0].address',
+	},
+};
+
+export class BeforeSaveCannedResponse {
```
KevLehman, line 31: Should we have like a base class for these actions? 👀

apps/meteor/ee/server/hooks/messages/BeforeSaveCannedResponse.ts
```diff
@@ -0,0 +1,116 @@
...
+	static enabled = false;
+
+	private getSetting: <T extends SettingValue = SettingValue>(id: ISetting['_id']) => T;
+
+	constructor({ getSetting }: { getSetting: <T extends SettingValue = SettingValue>(id: ISetting['_id']) => T }) {
```
KevLehman, line 36: type :trollface: 

Was thinking what's better, having to type the object or initializing it as predicate read-only in the constructor 🤔

apps/meteor/ee/server/hooks/messages/BeforeSaveCannedResponse.ts
```diff
@@ -0,0 +1,116 @@
...
+		message,
+		room,
+		user,
+	}: {
+		message: IMessage;
+		room: IRoom;
+		user: Pick<IUser, '_id' | 'username' | 'name' | 'emails' | 'language'>;
+	}): Promise<IMessage> {
+		// If the feature is disabled, return the message as is
+		if (!BeforeSaveCannedResponse.enabled) {
```
KevLehman, line 70: If it's disabled by setting, we won't be able to run it as the enabled prop will never update 😬

apps/meteor/ee/server/hooks/messages/BeforeSaveCannedResponse.ts
sampaiodiego: yes. I was thinking about this.. I couldn't think of a standard yet though

apps/meteor/ee/server/hooks/messages/BeforeSaveCannedResponse.ts
sampaiodiego: I had a hard time trying to find the best solution for this..

I assumed on startup I could check the setting, which would return `false` if no EE license is applied.. then added a listener to license module to be able to enable the feature..

but I still check the setting every time on line 74, which may not be needed btw

apps/meteor/ee/server/hooks/messages/BeforeSaveCannedResponse.ts
sampaiodiego: I don't know as well.. I wanted to decouple this as much as possible from the rest, but I couldn't decouple from the setting so I decided to go this way but definitely not something I'm happy with.

maybe here I can have only the static property to say it is enabled or not and control it from outside, like having the setting and license listener somewhere else in the omnichannel service.. wdyt?

KevLehman: That is def an option, and will help with the testability 🤔

What if all "callbacks" have this static prop of "enabled", like you suggest, with a setter that prints a debug log when the value is changed? That way we can at least track why a new callback is not working, while decoupling that event logic from here 👀

apps/meteor/ee/tests/unit/server/hooks/messages/BeforeSaveCannedResponse.tests.ts
```diff
@@ -0,0 +1,169 @@
...
+		const message = await canned.replacePlaceholders({
+			message: createMessage('{{agent.name}}'),
+			room: createRoom(),
+			user: createUser(),
+		});
+
+		return expect(message).to.have.property('msg', '{{agent.name}}');
+	});
+
+	it('should do nothing if the message is from a visitor', async () => {
```
sampaiodiego, line 88: this is a new behavior I introduced but think makes sense.. wdyt @KevLehman ?

apps/meteor/ee/server/hooks/message/BeforeSaveCannedResponse.ts
sampaiodiego: I just changed to control the feature completely from "outside" ;)

apps/meteor/ee/tests/unit/server/hooks/messages/BeforeSaveCannedResponse.tests.ts
KevLehman: Makes sense, we don't allow canned responses from visitors 👀

apps/meteor/ee/server/hooks/messages/BeforeSaveCannedResponse.ts
sampaiodiego: oh, I ended up not creating the setter nor adding a log.. but changed to rely only on the `enabled` flag

KevLehman (approved):

apps/meteor/ee/tests/unit/server/hooks/messages/BeforeSaveCannedResponse.tests.ts
sampaiodiego: it is weird though, the implementation replaces the placeholder for every message, not only the ones where canned responses were used.. 🤷

KevLehman: Yeah, that's because we don't have a way to differentiate if a canned response was used or not in a message, since the message comes from the client directly (as a normal message) :sadblob:

PR #31231, opened 2023-12-13
refactor: convert MAC check callback to class

sampaiodiego:
## Issue(s)

ARCH-1253


apps/meteor/server/services/messages/hooks/BeforeSaveCheckMAC.ts
```diff
@@ -0,0 +1,24 @@
+import { MeteorError, Omnichannel } from '@rocket.chat/core-services';
+import type { IRoom } from '@rocket.chat/core-typings';
+import { isOmnichannelRoom, type IMessage } from '@rocket.chat/core-typings';
+
+export class BeforeSaveCheckMAC {
+	async isWithinLimits({ message, room }: { message: IMessage; room: IRoom }): Promise<void> {
```
KevLehman, line 6: should we keep the "skip edited messages" if? 👀

sampaiodiego: image

sampaiodiego: it is there actually:

https://github.com/RocketChat/Rocket.Chat/pull/31231/files#diff-6d79e082eda977d9389a7d87f152ba71fc3566c53696d7be838b4893e33af933R139

apps/meteor/server/services/messages/service.ts
```diff
@@ -134,6 +138,7 @@ export class MessageService extends ServiceClassInternal implements IMessageServ
 
 		if (!this.isEditedOrOld(message)) {
 			await Promise.all([
+				this.checkMAC.isWithinLimits({ message, room: _room }),
```
sampaiodiego, line 141: since we're doing the "is edited" check here, I removed from the "callback code".

I think this block of code should be moved up, before the other actions.. it may have been put here due to requirement of a specific order.. need to take a look

KevLehman: thx man

apps/meteor/server/services/messages/hooks/BeforeSaveCheckMAC.ts
KevLehman: image

KevLehman (approved):

Issue #25940, opened 2022-06-21
Users custom status drops on change presence status


ankar84:
### Description:

Users custom status drops on change presence status

### Steps to reproduce:

1. Click avatar
2. Choose **Custom status**
3. Set some text status
4. Change presence status (for example online => away)
5. Check custom status is dropped

### Expected behavior:

Custom status should remain despite of presence status.
I'm not 100% sure, but on a pre 4.8 versions that was true.

### Actual behavior:

![custom-status01](https://user-images.githubusercontent.com/4023037/174724294-fde82f10-3dd3-4a1a-b4c7-991cc9acf7cf.gif)

### Server Setup Information:

- Version of Rocket.Chat Server: 4.8.1
- Operating System: OL8
- Deployment Method: docker
- Number of Running Instances: 20
- DB Replicaset Oplog: Enabled
- NodeJS Version: v14.18.3
- MongoDB Version: 4.4.14

### Client Setup Information

- Desktop App or Browser Version: Chrome 100
- Operating System: Windows 10

### Additional context

I guess that on a previous versions Custom status remain despite presence status changes

ankar84:
5.3.5 still have that issue

casalsgh:
@hardikbhatia777 is this something you might want to be assigned?

VanshulB:
Hey @casalsgh I would like to work on this issue. Found [useStatusText](https://github.com/RocketChat/Rocket.Chat/blob/develop/apps/meteor/client/sidebar/header/hooks/useStatusItems.tsx#L32), but not sure if it's the file containing the issue. Is there a way to check if the user has set up a custom status cause the function `isDefaultStatus()` always returns true? Also, it should be `user.statusText` instead of `status.name` for getting custom status ? (cause the `status` object doesn't seem to have `name` property in it)?

hardikbhatia777:
> @hardikbhatia777 is this something you might want to be assigned?

Sorry, I did not get a notification for this post. I'll look into it and get back to you.

ggazzo:
@casalsgh this feature was implemented like that by design, its ok if we decide/want to change it, but not a bug.  

and the changes and descriptions should reflect this decision

casalsgh:
FYI to all here -> Gazzo had the latest and feature is working as designed. Per our Product squad we are looking to change behaviour in the future as follows:
"Save the text and status in localstorage. When clicking ‘edit custom status' just fill out the fields with what we have saved there." 

[...]

PR #31237, opened 2023-12-13
fix: Store custom status in localStorage

hardikbhatia777:

Added the ability to store custom status in local storage. That way the custom status won't have to be re-typed every time it is changed to 'online', 'away' etc. as requested by @casalsgh in #25940 

https://github.com/RocketChat/Rocket.Chat/assets/98163873/11c391b5-7037-4be5-a9c2-59563694821d

## Issue(s)

#25940

## Steps to test or reproduce

1. Change status to online / away / offline
2. Set a custom status
3. Set your status back to online / away / offline
4. Try to change to a custom status. Your last custom status will already be entered in the text box, so all you have to do is press save or edit it if you wish to.


hardikbhatia777:
I'm unsure why the codecov/project test is failing. I'll appreciate some feedback about why this is happening so I can avoid it in the future.

hardikbhatia777:
I'm unsure why the codecov/project test is failing. I'll appreciate some feedback about why this is happening so I can avoid it in the future. Do I need to write unit tests for the status feature?

gabriellsh (changes requested):
Please use the 'useLocalStorage' hook

hardikbhatia777:
@gabriellsh I've made the necessary changes.

baldbull69 (approved):

apps/meteor/client/sidebar/header/EditStatusModal.tsx
```diff
@@ -31,6 +33,9 @@ const EditStatusModal = ({ onClose, userStatus, userStatusText }: EditStatusModa
 		if (statusText && statusText.length > USER_STATUS_TEXT_MAX_LENGTH) {
 			return setStatusTextError(t('Max_length_is', USER_STATUS_TEXT_MAX_LENGTH));
 		}
+		if (statusText && statusText.length < USER_STATUS_TEXT_MAX_LENGTH) {
+			setCustomStatus(e.currentTarget.value);
```
gabriellsh, line 37: This code will never run, because there's an exactly equal if statement above that returns the fuction.

Also, it is preferable that you save the custom status only after it is successfully saved. You can move this to the `handleSaveStatus` callback below, inside the `trycatch` block (after the `setUserStatus` function call)

gabriellsh (changes requested):

apps/meteor/client/sidebar/header/EditStatusModal.tsx
hardikbhatia777: Thank you for the review and guidance :) I've made the changes now! 

PS. The if statement was there to handle edge cases (status is longer than 120 chars), it was the exact opposite condition than the one above. In hindsight, I should've probably used an else statement.

janainaCoelhoRocketchat:
I tried to test this, but for me, it is still happening. 
[screen-capture (1).webm](https://github.com/RocketChat/Rocket.Chat/assets/105796517/39815584-4d7b-404e-ac01-7294a5f31fda)

hardikbhatia777:
Hi @janainaCoelhoRocketchat 

That is the correct behaviour change as described by @casalsgh in the issue tagged in the initial description. 

Right now, the custom status is never stored and is forgotten every time the status is changed to online/busy/offline/away, as a result the text box is empty.

With this PR, the last status remains saved (locally) and you can quickly hit save to return back to that custom status ("TESSTEST" in your case), so the video is actually giving the correct output :)

hardikbhatia777:
Hi @janainaCoelhoRocketchat 


image
image

The original issue posted was not actually to be adressed as the feature was designed to be this way. However, as per the new wanted behaviour in Rocket.chat as described further down in the comments of #25940, the feature is working fine in the PR :)

Right now the status is forgotten as soon as you change it to Away/Offline/Online/Busy and the next time you click "Custom Status", the text box is completely empty. 

With this PR, the last custom status will be saved (as described in the issue thread) and will be presented, so all you have to do is hit save, no need to retype the whole thing as it is already present as the initial value of the textbox :)

@casalsgh can you please confirm that this is the behaviour change that RC is looking for?

janainaCoelhoRocketchat:
@hardikbhatia777 thanks for the explanation

casalsgh:
Thanks for the contribution @hardikbhatia777 ! Glad to see it merged and will soon be out on a new release (most likely 6.6.0)

PR #31249, opened 2023-12-15
chore: add aria-label to Select Inputs at Engagement Dashboard

hugocostadev:

Due to accessibility concerns, it's necessary to add aria-label to Inputs that does not contains visible label


dougfabris (approved):
🚀

PR #31270, opened 2023-12-18
fix: not being able to access or create rooms with join code

heitortanoue:

Added `joinCode` to the room settings and fetch it if not passed to the `Room.join()` function.

## Steps to test or reproduce

- Create a channel

- Click on the icon to edit the channel

- Activate “Password to access”

- Fill in a password

- Save

An issue also happens when trying to join a protected room, even if you fill in the correct password it returns:
`{"success":false,"error":"error-invalid-subscription"}`

## Further comments

TC-1060


KevLehman (commented):
just passing by to ask for some api tests :trollface:

apps/meteor/server/services/room/service.ts
```diff
@@ -106,14 +106,19 @@ export class RoomService extends ServiceClassInternal implements IRoomService {
...
+			if (!room.joinCode) {
+				fetchedJoinCode = (
+					(await Rooms.findOneById(room._id, {
+						projection: { joinCode: 1 },
+					})) as IRoomWithJoinCode
+				).joinCode;
+			}
+
+			if ((!joinCode || joinCode !== fetchedJoinCode) && !(await Authorization.hasPermission(user._id, 'join-without-join-code'))) {
+				throw new MeteorError('error-code-required', 'Code required', { method: 'joinRoom' });
```
pierre-lehnen-rc, line 120: Instead of loading the joinCode to memory, just do a query by room id and join code to see if the record exists.

pierre-lehnen-rc (changes requested):
Please fix the existing tests for this endpoint. There are currently two tests for this kind of room:

test 1: `should succeed when joining code-needed channel without join code and with join-without-join-code permission` -> OK
test 2: `should succeed when joining code-needed channel with join code` -> this one should've caught this bug, but it's running with the admin credentials so the code is never even used.

Also add another test:
`should fail when joining code-needed channel without join code`

apps/meteor/app/channel-settings/server/methods/saveRoomSettings.ts
dougfabris, line 1: It's not necessary to add this param, it was a mistake on the frontend side that will be fixed here: #31277
You can remove it!

dougfabris (changes requested):

.changeset/stale-monkeys-yell.md
```diff
@@ -0,0 +1,5 @@
+---
+"@rocket.chat/meteor": patch
+---
+
+fix: not being able to access or create rooms with join code
```
hugocostadev, line 5: This message should be user friendly, check it out the docs and edit, please
https://developer.rocket.chat/contribute-to-rocket.chat/modes-of-contribution/participate-in-rocket.chat-development/development-workflow#4.-adding-changeset-to-your-pull-request

hugocostadev (changes requested):

.changeset/stale-monkeys-yell.md
heitortanoue: thanks for the heads up!

apps/meteor/server/services/room/service.ts
```diff
@@ -106,14 +106,24 @@ export class RoomService extends ServiceClassInternal implements IRoomService {
...
+
+			const isCorrectJoinCode = !!(await Rooms.findOne(
+				{
+					_id: room._id,
+					joinCode,
+				},
+				{
+					projection: { _id: 1 },
+				},
+			));
```
pierre-lehnen-rc, line 122: Move the query to a new function inside the model.

pierre-lehnen-rc (changes requested):
All good otherwise, just make sure to always keep query objects contained to the model class.

matheusbsilva137 (approved):

pierre-lehnen-rc (approved):

elfajriharsuda (approved):

sampaiodiego (approved):

dougfabris (approved):

hugocostadev (approved):

PR #31272, opened 2023-12-19
chore: remove Messages model dependency of Rooms model

sampaiodiego:

Previously the `Message` model used to interact directly with the `Room` model, which is not ideal, since models shouldn't know about the business logic of the application.

This change also helped finding some places where multiple changes to the same room object were being performed one right after the other, so I improved that as well by performing a single operation.


ggazzo (dismissed):

KevLehman (dismissed):

packages/model-typings/src/models/IRoomsModel.ts
```diff
@@ -239,7 +239,7 @@ export interface IRoomsModel extends IBaseModel<IRoom> {
 	incUsersCountById(rid: string, inc: number): Promise<UpdateResult>;
 	incUsersCountNotDMsByIds(rids: string[], inc: number): Promise<Document | UpdateResult>;
 	setLastMessageById(rid: string, lastMessage: IRoom['lastMessage']): Promise<UpdateResult>;
-	resetLastMessageById(rid: string, lastMessage?: IMessage | null): Promise<UpdateResult>;
+	resetLastMessageById(rid: string, lastMessage?: IMessage | null, msgCounter?: number): Promise<UpdateResult>;
```
pierre-lehnen-rc, line 242:
```suggestion
	resetLastMessageById(rid: string, lastMessage?: IMessage | null, msgCountDelta?: number): Promise<UpdateResult>;
```
Just trying to make the param's purpose clearer

pierre-lehnen-rc (changes requested):

pierre-lehnen-rc (approved):

PR #31277, opened 2023-12-19
fix: Can't create public channels with joinCode

dougfabris:

It also verifies the permission `join-without-join-code` to show `ComposerJoinWithPassword`


apps/meteor/client/views/room/composer/ComposerJoinWithPassword.tsx
hugocostadev, line 1: I don't know if could be in another place the error, but I set the pwd to `12345` and is giving me wrong password, tried to change multiple times... 

did't have time to explore further

image

hugocostadev (changes requested):

apps/meteor/client/views/room/composer/ComposerJoinWithPassword.tsx
dougfabris: it is another issue related and will be fixed here: #31270

hugocostadev (approved):

PR #31299, opened 2023-12-21
fix: Loading state for `Marketplace` related lists

tiagoevanp:

Brings back the loading state of marketplace lists with `Skeleton` component


apps/meteor/client/providers/AppsProvider.tsx
```diff
@@ -133,9 +133,18 @@ const AppsProvider: FC = ({ children }) => {
...
+					value: { apps: store.data?.[1] || [] },
+				},
+				marketplaceApps: {
+					phase: store.isLoading ? AsyncStatePhase.LOADING : AsyncStatePhase.RESOLVED,
+					value: { apps: store.data?.[0] || [] },
+				},
+				privateApps: {
+					phase: store.isLoading ? AsyncStatePhase.LOADING : AsyncStatePhase.RESOLVED,
+					value: { apps: store.data?.[2] || [] },
+				},
```
Dnouv, line 147: How about extracting the common logic into a function, and destructing the array with more descriptive variables, something like, below:

```ts
const getAppState = (apps) => ({
  phase: store.isLoading ? AsyncStatePhase.LOADING : AsyncStatePhase.RESOLVED,
  value: { apps: apps || [] },
});

const [marketplaceAppsData, installedAppsData, privateAppsData] = store.data || [];

<AppsContext.Provider
  children={children}
  value={{
    installedApps: getAppState(installedAppsData),
    marketplaceApps: getAppState(marketplaceAppsData),
    privateApps: getAppState(privateAppsData),
    reload: async () => {
      await Promise.all([queryClient.invalidateQueries(['marketplace'])]);
    },
  }}
/>
```

tiagoevanp: Thanks for the suggestion, nice catch! <3

Dnouv (approved):

casalsgh:
Tested/validated by Tiago and Devanshu, consider this as QA tested since we do not have a QA person to focus on this rn.

PR #31306, opened 2023-12-22
fix: Search Function Not Displaying Messages with Attached Images

### LINKED ISSUE #31156: Search Function Not Displaying Messages with Attached Images
### Description:

Hello, I noticed a potential bug within the RocketChat platform, where messages, while having a certain keyword, are not being shown in the search results if there is an image attached. This behavior is observed both on the RocketChat Server v6.4.6, v6.5.0 the RocketChat Electron Client v3.9.9, and also when accessed through a web browser.

### Steps to reproduce:

1. Go into any group chat on RocketChat.
2. Post a message with a particular keyword and attach an image to the message.
3. Use the search bar to look for that particular keyword.
4. Notice that the message with the attached image does not appear in the search results, even though it contains the searched keyword.

### Expected behavior:

The message with the keyword and the attached image should appear in the search results.

### Actual behavior:

Messages with attached images are not showing up in the keyword search results.

### Server Setup Information:

- Version of Rocket.Chat Server:  6.4.6, 6.5.0
- Operating System: Linux
- Deployment Method: docker

### Client Setup Information

- Desktop App or Browser Version: RocketChat.Electron: 3.9.9, Google Chrome: 119

### Additional context

This issue is causing difficulties finding important messages embedded with images. Any help regarding this discrepancy would be deeply appreciated.

### Screenshots
screen1
screen3

Sayan4444:

BEFORE THE FIX
https://github.com/RocketChat/Rocket.Chat/assets/112304873/f8c3ccfb-32e7-48b3-820c-e229ca16e72c

AFTER THE FIX
https://github.com/RocketChat/Rocket.Chat/assets/112304873/f1093541-446c-4ade-a247-d4299087b79a

REASON BEHIND CHANGES

No text-indexing was done in collection 'rocketchat_message' field attachments[].description and thats why any search related to attachments was not working. In commit 1, I fixed that. But mongoDB allows one text-index per collection, so we need to remove the previous index to introduce the new one which covers both fields "msg,attachments[].description". This is of commit2. If there is a way to add fields in existing index ,commit 2 can be removed.

## Issue(s)

This should close issues :-
fix #31156 

## Steps to test or reproduce

1. Upload a picture or video and add description with it
2. Click on the search bar within the chat and try searching for words you added in the description


Sayan4444:
@dougfabris Hi is there any update for this PR? as I feel this issue is important to be resolved

Sayan4444:
Hello, @dougfabris I am aware that you have been quite busy, but could you provide me an update on the issue's status?

hugocostadev (changes requested):
Hey @Sayan4444 thanks again for your contribution...

There is no need to dropIndex in there, It's best to create a migration to accomplish that. 

Also, check lint errors, there are some

hugocostadev:
No, you have to drop the index, but it's better to do in a migration file, otherwise every restart is going to drop the index.
Please, check this example: https://github.com/RocketChat/Rocket.Chat/blob/develop/apps/meteor/server/startup/migrations/v304.ts , after creating the new migration add to https://github.com/RocketChat/Rocket.Chat/blob/develop/apps/meteor/server/startup/migrations/index.ts

Sayan4444:
@hugocostadev I have done the required changes. I just want to know that exactly when are these starup migrations up functions get called? I thought they run every time I start a dev server but I was wrong

hugocostadev:
> @hugocostadev I have done the required changes. I just want to know that exactly when are these starup migrations up functions get called? I thought they run every time I start a dev server but I was wrong

The migrations are called in every update/restart, BUT we store the last migration execution to do not run the same migration more than once... we just run the migration that is bigger than `lastMigration`.

Check the files: `Migrations.ts`, `xrun.ts`, `migrations.ts`

Sayan4444:
@hugocostadev absolutely understood. I hope the changes made by me solves the issue

hugocostadev:
Since this PR drop the biggest collection in the database, we will release this fix in the next major. (7.0)

Sayan4444:
Hi @hugocostadev do I need to make any more changes?

PR #31309, opened 2023-12-23
 fix: Visibility of attachments in channel when 'Send to Channel' flag is used in thread

### LINKED ISSUE #29909: Images uploaded to threads are not shown in channel message
### Description:

Uploading an image to a thread while using the _"Also send to channel"_, **the message does not show up** (completely missing) in the channel.

This is a huge problem as users don't know what to look for because they only got notification but can't see or click on the message. Inside the thread view, it is displayed fine.

Checked the markup code and it seems the message row is absent (so it not just hidden, it's not rendered).
Also checked file upload settings and nothing seems off/wrong because it works fine just doesn't show up.

### Steps to reproduce:

1. send a message
2. reply to it in thread
3. upload an image to the thread while using _"Also send to channel"_ option

### Expected behavior:

the message (name of the image) shown in the channel like in pre-6.0 versions

### Actual behavior:

the message is missing

### Server Setup Information:

- Version of Rocket.Chat Server: 6.2.10 (latest)
- Operating System: Debian 11
- Deployment Method: docker (official image)
- DB Replicaset Oplog: enabled
- NodeJS Version: 14.21.3
- MongoDB Version: 6.0.8 / wiredTiger

### Client Setup Information

- Desktop App or Browser Version: Desktop App: 3.9.5 (latest); Browser: Chrome 115 (latest)

### Additional context

* GridFS is used for file upload
* This problem happened after the upgrade from 5.4 to 6.0, then persist ever since (6.1, 6.2), so it's not a new bug, long running issue.

### Relevant logs:

There is no error, warning or any other suspicious relevant log (the image upload succeeds and works fine, only the render is broken).

Ruchir28:

In threads, the 'tshow' flag, which indicates if a message should also appear in the main channel, wasn't being correctly forwarded to subsequent methods. This led to an issue where attachments sent with the 'also send to channel' option enabled were not visible in threads. This PR addresses the problem by ensuring the 'tshow' flag is passed along properly. As a result, messages with attachments now correctly display in the channel as 'Message with attachment'.

## Issue(s)

closes #29909 

## Steps to test or reproduce

1. send a message
2. reply to it in thread
3. upload an image/file or send a audio or video message to the thread while using "Also send to channel" option

Before fix : 
message isn't visible in the channel

After fix : 
message visible in thread as "Message With Attachment" and on clicking opens up the thread and point to the attached file

### DIFF
--- a/apps/meteor/app/api/server/v1/rooms.ts
+++ b/apps/meteor/app/api/server/v1/rooms.ts
@@ -3,6 +3,7 @@ import type { IRoom } from '@rocket.chat/core-typings';
 import { Messages, Rooms, Users } from '@rocket.chat/models';
 import type { Notifications } from '@rocket.chat/rest-typings';
 import { isGETRoomsNameExists } from '@rocket.chat/rest-typings';
+import Ajv from 'ajv';
 import { Meteor } from 'meteor/meteor';
 
 import { isTruthy } from '../../../../lib/isTruthy';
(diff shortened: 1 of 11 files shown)


Ruchir28:
Hey @hugocostadev  , can you review this

PR #31315, opened 2023-12-24
chore: prohibit development on windows

debdutdeb:
## Issue(s)

We don't support development on windows natively.

## Further comments

Use wsl2.

### DIFF
--- a/package.json
+++ b/package.json
@@ -43,6 +43,10 @@
 		"node": "14.21.3",
 		"npm": "Use yarn instead"
 	},
+	"os": [
+		"linux",
+		"darwin"
+	],
 	"packageManager": "yarn@3.5.0",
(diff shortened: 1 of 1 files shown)


sampaiodiego (approved):
oh, this is interesting, I didn't know about this..

just wondering if we should set it to `!win32` though.. wdyt?

debdutdeb:
well .. I think that's fine. by default if a platform isn't there, shouldn't allow to run yarn or anything, which would be the same as adding !win32. But I made a mistake actually. This should be present in all package.jsons not just this.

debdutdeb:
huh? it isn't working in our repo .. might workspaces not support this field? 😭

elfajriharsuda (approved):

PR #31318, opened 2023-12-25
fix: Attachments not collapsing when using incoming webhook

### LINKED ISSUE #29276: Message attachments not collapsed when 'collapsed: true' provided in v6.2.0
### Description:

We recently upgraded to to v6.2.0 of RocketChat, however it seems like in this release message attachments do not initially appear collapsed when `collapsed: true` is provided in the message request body.

### Steps to reproduce:

1. Using curl as below to make a request containing a test message with an example message attachment containing fields:
`curl -H "X-Auth-Token: my-token" \
     -H "X-User-Id: my-id" \
     -H "Content-type:application/json" \
     https://localhost:3000/api/v1/chat.postMessage \
     -d '{"channel":"#general","emoji":":smirk:","text":"Sample message","attachments":[{"collapsed":true,"color":"#abcdef","fields":[{"title":"test 1","value":"test 1 value"},{"title":"test 2","value":"test 2 value"}],"text":"test text","title":"test title","title_link":"https://youtube.com","ts":"2016-12-09T16:53:06.761Z"}]}'`
2. observe message in specified channel and it is not collapsed.

### Expected behavior:
In previous releases message attachments with `collapsed: true` would appear as below:
Screenshot 2023-05-17 at 19 56 07

### Actual behavior:

In v6.2.0 attachments are appearing expanded even if `collapsed: true` is provided:
Screenshot 2023-05-17 at 19 56 22

### Server Setup Information:

- Version of Rocket.Chat Server: 6.2.0
- Operating System: Debian GNU/Linux
- Deployment Method: Helm
- NodeJS Version: v14.21.3
- MongoDB Version: 6.0.5

hardikbhatia777:

Right now there has been a long-standing issue where attachments don't collapse even when `"collapsed": true` is specified. Collapse always initializes as null (doesn't consider the value from the incoming webhook).

Before: (regardless of specifying collapsed: true or collapsed: false)
image

After:

https://github.com/RocketChat/Rocket.Chat/assets/98163873/a2fcc6ad-853c-48fe-970a-291885d2af0d

## Issue(s)

fix #29276 

## Steps to test or reproduce

1. Create webhook
2. Try to send a message containing an attachment
3. Specify "collapsed": true
4. The attachment should be collapsed but it's not.
5. Specify "collapsed": false
6. The attachment is not collapsed.

## Further comments

`curl -H "X-Auth-Token: my-token" \ -H "X-User-Id: my-id" \ -H "Content-type:application/json" \ https://localhost:3000/api/v1/chat.postMessage \ -d '{"channel":"#general","emoji":":smirk:","text":"Sample message","attachments":[{"collapsed":true,"color":"#abcdef","fields":[{"title":"test 1","value":"test 1 value"},{"title":"test 2","value":"test 2 value"}],"text":"test text","title":"test title","title_link":"https://youtube.com","ts":"2016-12-09T16:53:06.761Z"}]}'` 

This command can be used to test the modification


debdutdeb:
hm this may be one fix .. but probably not the ideal one. IIRC when i looked at this, there was an inconsistency in how collapsed was passed at different stages vs how it was read. Just a comment for now

hardikbhatia777:
Fair enough, I was sorta hesitant to use nullish operator but it seemed like the only option at the time. Turns out the whole collapsed field wasn't needed in the first place.

I think the inconsistency came from 'collapsed' being declared explicitly as a parameter when it wasn't really required. The `collapsed` value is taken as input in [sendMessage.js](https://github.com/RocketChat/Rocket.Chat/blob/85effe240343c143803d3315035d90512e79a8e5/apps/meteor/app/lib/server/functions/sendMessage.ts#L119) as a sort of intrinsic value for every attachment (just like every other argument), however it is redeclared in the [Attachments.tsx](https://github.com/RocketChat/Rocket.Chat/blob/85effe240343c143803d3315035d90512e79a8e5/apps/meteor/client/components/message/content/Attachments.tsx#L7) file.

Long story short, the type defined in Attachments.tsx need not redeclare `collapsed`. This should be the optimal solution imo, check out the new commit whenever you have time @debdutdeb

hugocostadev (changes requested):
I read your considerations and I agree with your statment... 

You can remove the prop from the type declaration and also from the <QuoteAttachment />  component that is passing the `collapsed` prop to <Attachment />  

Quote component here: https://github.com/RocketChat/Rocket.Chat/blob/develop/apps/meteor/client/components/message/content/attachments/QuoteAttachment.tsx#L71

hardikbhatia777:
Perfect :) I've made the changes @hugocostadev

hugocostadev (approved):

elfajriharsuda (approved):

janainaCoelhoRocketchat:
tested

Issue #30793, opened 2023-10-27
Dropping a file from another browser window creates two upload dialogs


paulchen:
### Description:

When dragging a file from another Chromium window and dropping it into a Rocket.Chat channel, the file upload dialog is shown twice.

### Steps to reproduce:

1. Open a Rocket.Chat channel in one Chromium window.
2. Open the image to be drag-and-dropped in another Chromium window.
3. Drag the image and drop it into the Rocket.Chat channel.

### Expected behavior:

One file upload dialog is opened.

### Actual behavior:

Two file upload dialogs are opened, one overlapping the other:

https://github.com/RocketChat/Rocket.Chat/assets/404840/ceee0db6-b7c9-424a-9e78-07f6af2c165d

### Server Setup Information:

- Version of Rocket.Chat Server: 6.4.4
- Deployment Method: Docker
- Number of Running Instances: 1
- MongoDB Version: 6.0.11

### Client Setup Information

I was able to reproduce this with the following clients:

- Windows 10
  - Google Chrome Version 118.0.5993.118 (Offizieller Build) (64-Bit)
  - Microsoft Edge Version 118.0.2088.76 (Offizielles Build) (64-Bit)
  - Electron Client 3.9.9
- Arch Linux
  - Chromium Version 118.0.5993.117 (Offizieller Build) Arch Linux (64-Bit)

I was unable to reproduce this in Firefox 119.0, neither on Windows 10 nor on Arch Linux.

paulchen:
> This issue cant be reproduced, can you recheck if it exists?

I just successfully reproduced it using
 - Rocket.Chat 6.5.1
 - Arch Linux, Chromium Version 120.0.6099.129 (Offizieller Build) Arch Linux (64-Bit)

Sayan4444:
@paulchen very very thank you for the confirmation. It helped me to fix the exsisting problem. It will be fixed as soon as this PR gets merged

PR #31332, opened 2023-12-28
fix: Dropping a file from another browser window creates two upload dialogs

Sayan4444:
BEFORE THE FIX
https://github.com/RocketChat/Rocket.Chat/assets/112304873/a6f0faf1-cd9a-480d-bbe1-c60044d10205

AFTER THE FIX
https://github.com/RocketChat/Rocket.Chat/assets/112304873/d59fb5c2-9fcd-40fc-9630-8386a0712520

REASON BEHIND CHANGES
We are comparing the files by their size in bytes and checking for any duplicate files as it is very hard to find files of same size.We can also use nodejs to check but it takes a bit longer for long files.If you want the nodejs version please do tell me.

Issue(s)
This should close issues :-
closes https://github.com/RocketChat/Rocket.Chat/issues/30793

Steps to test or reproduce
Open a Rocket.Chat channel in one Chromium window.
Open the image to be drag-and-dropped in another Chromium window.
Drag the image and drop it into the Rocket.Chat channel.


hugocostadev:
Did you understand why it's creating two files?  I think that we need to fix this problem instead of resolving this issue in our application

Sayan4444:
Yes I have @hugocostadev.
image
The primary cause of multiple files being created for (some type of files) is this commented-out code(commented by me for now for demonstration purpose). The program attempts to send a get request to each src of img contained in the dropped file as soon as the if condition is satisfied. A duplicate file is not produced if the URL returns an error stating that a direct request is blocked.
Example

Screenshot 2024-01-04 235343

However, if the request is made, a new file containing the identical picture is created and added to the files array that previously included the image, resulting in the creation of a duplicate file.

The problematic code has been commented out, and practically all cases for drag and drop runs normally. Would you kindly let me know if I missed any cases and how do you want me to approach the situation with the duplicate files?

Sayan4444:
@hugocostadev can you check from this website,
https://tenor.com/ 
and revert me back?

hugocostadev:
For me does not work but a windows user tested and it was able to reproduce, I'll accept your solution! Thanks

hugocostadev (approved):

hugocostadev (approved):

PR #31336, opened 2023-12-28
fix: Cron job for clearing OEmbed cache isn't working

### LINKED ISSUE #29434: Automatic Oembed cache cleaning does not work
### Description:

We noticed on multiple RC installs that `rocketchat_oembed_cache` collection is never cleaned up by the automatic process of RC. The default cache expiration time is 30 days, and this value is not respected.

This collection is used as a cache for generating previews in the messages. It seems to contain full original web pages of the links in the messages, hence it can grow quite big on the old RC setups.

### Steps to reproduce:

1. Dump the database and mark down the size of the file `rocketchat_oembed_cache.bson`.
2. Log into database and run `db.rocketchat_oembed_cache.find().pretty()`. Mark down the time of some entry in the beginning of the output. (The last entry will do too as the output limiter cuts it early anyway.)
3. Press `Clear` cache button under `Admin -> Settings -> Message -> Embed Cache Expiration Days`
4. After a while you'll get an error message. Don't mind that as the button seems to work regardless.
5. Run the `db.rocketchat_oembed_cache.find().pretty()` in the database again and make note of time stamps.
6. Dump the database again and make note of the new size of the file `rocketchat_oembed_cache.bson`.

### Expected behavior:

There should not be more than 30 days worth of content in the oembed cache when the `Admin -> Settings -> Message -> Embed Cache Expiration Days` settings has been set to 30 days.

### Actual behavior:

The `Embed Cache Expiration Days` setting is not respected. The database contains preview source data from far longer time period.

The manual operation using the `Clear` button works, although it outputs an error message `e.params is not iterable (cannot read property)`.

In our case, after the manual clearing of the table using the `Clear` button, the size of the collection dropped from 4.6 GB to 178 MB.

### Server Setup Information:

- Version of Rocket.Chat Server: 6.2.1
- Operating System: RHEL8
- Deployment Method: tar
- DB Replicaset Oplog: Yes
- NodeJS Version: 14.21.3
- MongoDB Version: 5.0

### Client Setup Information

- Desktop App or Browser Version: Any

### Additional context

Looking at the timestamps of the uncleared cache, the cleaning process of oembed cache stopped working a year ago in 2022 April/May.

matheusbsilva137:

 - Fix error running cron job for clearing OEmbed cache (we shouldn't check for the `uid` in this case);
 - Fix `data.params undefined` error when clearing OEmbed cache (eg with the "Clear" button in the settings).

## Issue(s)

Fixes #29434

## Steps to test or reproduce

Send some messages containing links with previews, then go to **Workspace > Settings > Message** and check the value for the **Embed Cache Expiration Days** setting. A cron job to erase all OEmbed cache registries older than `Embed Cache Expiration` days should be erased from the database (there should be no document in the `rocketchat_oembed_cache collection` with an `updatedAt` field older than defined in the setting).
The cron job runs everydat at 2:24am (this can be changed in [this line](https://github.com/RocketChat/Rocket.Chat/pull/31336/files#diff-21161a95fbf3394faa4f2af4add5e90d8eeeb279694d3107c86f834c41059e99R6) for test purposes -- use it as `* * * * *` to run every minute, for example).

## Further comments

SUP-425


apps/meteor/server/methods/OEmbedCacheCleanup.ts
```diff
@@ -21,12 +28,10 @@ Meteor.methods<ServerMethods>({
...
 		}
 
-		const date = new Date();
-		const expirationDays = settings.get<number>('API_EmbedCacheExpirationDays');
-		date.setDate(date.getDate() - expirationDays);
-		await OEmbedCache.removeAfterDate(date);
+		await executeClearOEmbedCache();
 		return {
 			message: 'cache_cleared',
+			params: [],
```
KevLehman, line 34: If this `params` is needed, pls add it to the return type of method

apps/meteor/server/methods/OEmbedCacheCleanup.ts
```diff
@@ -12,6 +12,13 @@ declare module '@rocket.chat/ui-contexts' {
 	}
 }
 
+export const executeClearOEmbedCache = async () => {
+	const date = new Date();
+	const expirationDays = settings.get<number>('API_EmbedCacheExpirationDays');
+	date.setDate(date.getDate() - expirationDays);
+	await OEmbedCache.removeBeforeDate(date);
```
KevLehman, line 19: Is the await necessary? You may just return the promise as well and let the caller await for it (since it's already async)

apps/meteor/server/methods/OEmbedCacheCleanup.ts
matheusbsilva137: Good catch, done!
Thanks

apps/meteor/server/methods/OEmbedCacheCleanup.ts
matheusbsilva137: Good idea, done!

heitortanoue (approved):

elfajriharsuda (approved):

apps/meteor/client/views/admin/settings/inputs/ActionSettingInput.tsx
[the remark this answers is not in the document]
matheusbsilva137: This fixes the `data.params undefined` error when hitting the "Clear cache" button, so I'd say so :thinking: 
We were missing the `params` field in the message returned by this action. My original fix was just to add it there, but after talking to @MarcosSpessatto we thought it'd be better to fix it in a more general way (in the client)

yash-rajpal (approved):

sampaiodiego (approved):

PR #31338, opened 2023-12-29
fix: Unexpected behaviour of the room-opened event in iFrames

### LINKED ISSUE #30782: Unexpected behaviour of the "room-opened" event in iFrames
### Description:

We are planning on using the "room-opened" event to notify the parent window of our website in case the user selects a room from the sidebar. The first time you select a specific room, the event triggers, but on every subsequent selection of a previously opened room, the event only triggers when closing the room or when selecting a room you have previously not opened.

### Steps to reproduce:

1. Select room x from the sidebar inside a Rocket.Chat iFrame => Event is fired.
2. Select another room y => Event is fired.
3. Select room x again => Event is not fired here.
4. Now when you select room y again, the "room-opened" event is fired, but with the event data of room x instead of y.

### Expected behavior:

I would expect the room-opened event to fire every time you select a room from the sidebar and show the event data of the selected room.

### Actual behavior:

Selecting a previously opened room only triggers when closing the room. In case you open room x twice and then open a room that has not been opened before (after the latest browser refresh), you will neither get notified of room x opening nor closing.

Is this the intended behaviour? To my understanding, the room-opened event should trigger every time you open a room regardless of if it was open before from the time the last browser refresh happened.

### Server Setup Information:

- Version of Rocket.Chat Server: 6.4.2
- Operating System: linux
- Deployment Method: docker
- NodeJS Version: 14.21.3 - x64
- MongoDB Version: 7.0.2

### Client Setup Information

- Desktop App or Browser Version: Browser Version
- Operating System: affects any operating system

### LINKED ISSUE #31327: The "room-opened" event is not fired in iframe under special conditions
Description:
Steps to reproduce:

In the Windows Chrome browser in Iframe Rocket Chat, if you open a chat room and go to another chat room, after returning to the previous chat, you will see that this event is not sent ("room-opened"), because maybe that page has already been loaded, if not created Please switch between the two chats several times. You can see that this event is sometimes not fired

Expected behavior:
This event is fired every time you navigate between two chat room

Server Setup Information:
Version of [Rocket.Chat](https://rocket.chat/) Server: 6.5.0
Operating System: ubuntu 22.04
Deployment Method: tar
Number of Running Instances: 1
NodeJS Version: 14.21.3
MongoDB Version: 5

Client Setup Information
Chrome Version 120.0.6099.71 (Official Build) (64-bit)
Operating System: windows 11

Sayan4444:
BEFORE THE FIX
https://github.com/RocketChat/Rocket.Chat/assets/112304873/724db234-465f-431d-a164-e2a1e26123c4

AFTER THE FIX
https://github.com/RocketChat/Rocket.Chat/assets/112304873/d4e4f9a5-e300-4851-a888-14a4c3c24381

REASON BEHIND CHANGES
Before I introduced these changes, we were adding an eventlistner to "room-opened" event and a function passing the value of room object as a argument. This function is hit when 'room is opened'. Next time when a new room-is opned,we were suppose to remove the event listner and add the new one with the new room object. But sometimes before adding the new event-listner with the new room object, the previous function gets fired, since that function had the old room object,we are seeing the mis-match.

Since this useOpenRoom will be hit every time a room changes, we dont need to specifically set a event listner.

Issue(s)
This should close issues :-
closes https://github.com/RocketChat/Rocket.Chat/issues/30782
closes #31327 

Steps to test or reproduce
Select room x from the sidebar inside a Rocket.Chat iFrame => Event is fired.
Select another room y => Event is fired.
Select room x again => Event is not fired here.
Now when you select room y again, the "room-opened" event is fired, but with the event data of room x instead of y.


gabriellsh:
Hi! Thanks for contributing!

Please, take a look at the comment I made in the referenced issue: https://github.com/RocketChat/Rocket.Chat/issues/31327#issuecomment-1877206842

We'll probably have to think of another approach to this, since I believe it should be treated as an improvement and not a fix

Sayan4444:
@gabriellsh ...so should I then close my PR or I can help on any ways?

urosTeo:
Hi, aret there any news regarding this?

Issue #31341, opened 2023-12-29
Minimize the media by clicking outside the image, instead of using the minimize arrow.


Sanskar711:
### Steps to reproduce:

1.  Login to rocket chat.
2. Go to any Community chat.
3. Click on any media 

### Expected behavior:

Clicking on area nearby should close the image

### Actual behavior:

![issue1](https://github.com/RocketChat/Rocket.Chat/assets/97801615/3b436bf4-bca1-41df-a4fd-72d486c1aaf6)

I think this feature will improve user experience significantly

Sayan4444:
@Sanskar711 I hope this PR may solve this issue. If you find something wrong or wanted to convey something else, please revert me back

PR #31344, opened 2023-12-29
fix: close image gallery on backdrop click

Sayan4444:
BEFORE THE FIX
https://github.com/RocketChat/Rocket.Chat/assets/112304873/c3948a9b-21c0-40f9-8f4c-97e79909a698

AFTER THE FIX
https://github.com/RocketChat/Rocket.Chat/assets/112304873/c7a48b70-4951-43e3-9e35-a879107d61de

REASON BEHIND CHANGES
There are far better ways to implement this but with this configuration,this is the best way to implement this feature

Issue(s)
This should close issues :-
closes https://github.com/RocketChat/Rocket.Chat/issues/31341

Steps to test or reproduce

Login to rocket chat.
Go to any Community chat.
Click on any media

### DIFF
--- a/apps/meteor/client/components/ImageGallery/ImageGallery.tsx
+++ b/apps/meteor/client/components/ImageGallery/ImageGallery.tsx
@@ -109,10 +109,10 @@ const ImageGallery = () => {
 	return createPortal(
 		<FocusScope contain restoreFocus autoFocus>
 			<Box className={swiperStyle}>
-				<div className='swiper-container'>
+				<div className='swiper-container' onClick={onClose}>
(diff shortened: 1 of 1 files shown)


juliajforesti:
> Instead of doing that I would create a new function and check if the element that was clicked (overlay) contains the class that I want, in that case, `swiper-container`.
> 
> WDYT @juliajforesti ?

That selector might not work, since there's another layer of container on top (`swiper-zoom-container`).
IMO it would be confusing to check if the element contains that class to conditionally close... 

I like the current solution :)

juliajforesti (approved):
@Sayan4444 thank you for your contribution!

Issue #27924, opened 2023-02-01
No error message when uploading a file that is not allowed


shevchukma:
### Description:

When sending some rare file types (for example [.har](https://en.wikipedia.org/wiki/HAR_(file_format))) that are not allowed to be sent, there is no message that this type of file is forbidden to be sent.

### Steps to reproduce:

1. Set "Accepted Media Types" setting in "File Upload"  section. For example `image/*,application/pdf,text/plain,application/msword`
2. Try sending a rare file type like .har
3. The file is not loaded into the chat, and there is no error message.

### Expected behavior:

An error appears like this:
![image](https://user-images.githubusercontent.com/104189570/216084439-ff079cc1-7c24-4aa3-8461-2b2a86b8f170.png)

### Actual behavior:

The message does not appear, the file is not loaded, the user does not understand what the problem is.

### Server Setup Information:

- Version of Rocket.Chat Server: 5.4.1
- Operating System: Oracle Linux Server release 8.6
- Deployment Method: docker
- Number of Running Instances: 20
- DB Replicaset Oplog: yes
- NodeJS Version: v14.19.3
- MongoDB Version: 5.0.14 

### Client Setup Information

- Desktop App or Browser Version: Chrome 108
- Operating System: Win10

### Additional context

HAR file for example https://gist.github.com/ericduran/6330201

### Relevant logs:

In the browser console there are messages like this:
`a86fdcee20cb2c4da8e5ac8c08c3c5f8b6199d9f.js?meteor_js_resource=true:5 fuselage: 0.31.22
a86fdcee20cb2c4da8e5ac8c08c3c5f8b6199d9f.js?meteor_js_resource=true:1583 MessageType.render is deprecated. Use MessageType.message instead. livechat_webrtc_video_call
registerType @ a86fdcee20cb2c4da8e5ac8c08c3c5f8b6199d9f.js?meteor_js_resource=true:1583
a86fdcee20cb2c4da8e5ac8c08c3c5f8b6199d9f.js?meteor_js_resource=true:1046 strict mode: use allowUnionTypes to allow union type keyword at "#/properties/value" (strictTypes)
w @ a86fdcee20cb2c4da8e5ac8c08c3c5f8b6199d9f.js?meteor_js_resource=true:1046
a86fdcee20cb2c4da8e5ac8c08c3c5f8b6199d9f.js?meteor_js_resource=true:1583 Uncaught (in promise) undefined
[...]

PR #31346, opened 2023-12-30
fix: No error message when uploading a file that is not allowed

Sayan4444:

BEFORE THE FIX
https://github.com/RocketChat/Rocket.Chat/assets/112304873/68e9d8d0-3ef5-411a-93e5-d009ff97c331

AFTER THE FIX
https://github.com/RocketChat/Rocket.Chat/assets/112304873/73c161a2-3658-4b76-8276-e38dbfc6f177

REASON BEHIND CHANGES
If the type of file which we want to upload is not clear then we should not allow it to be uploaded in the server

Issue(s)
This should close issues :-
closes https://github.com/RocketChat/Rocket.Chat/issues/27924

Steps to test or reproduce

1. Login to rocket chat.
2. Set "Accepted Media Types" setting in "File Upload" section. For example image/*,application/pdf,text/plain,application/msword
3. Try sending a rare file type like .har
4. The file is not loaded into the chat, and there is no error message.

### DIFF
--- a/apps/meteor/client/lib/chats/flows/uploadFiles.ts
+++ b/apps/meteor/client/lib/chats/flows/uploadFiles.ts
@@ -46,7 +46,7 @@ export const uploadFiles = async (chat: ChatAPI, files: readonly File[], resetFi
 					imperativeModal.close();
 					uploadNextFile();
 				},
-				invalidContentType: Boolean(file.type && !fileUploadIsValidContentType(file.type)),
+				invalidContentType: !(file.type && fileUploadIsValidContentType(file.type)),
 			},
 		});
 	};
--- a/.changeset/lemon-shrimps-draw.md
+++ b/.changeset/lemon-shrimps-draw.md
@@ -1,0 +1,5 @@
+---
+"@rocket.chat/meteor": patch
+---
+
+Fixed error message when uploading a file that is not allowed


hugocostadev (dismissed):
LGTM 🚀

hugocostadev (approved):

IgorOhrimenko:
This message broke the uploading of any files with an unknown type:
https://github.com/RocketChat/Rocket.Chat/issues/31692
Set the default value in the settings:
изображение
and create an empty file with no extension or an .ovpn extension, and you will receive an error.

emoxam:
"If the type of file which we want to upload is not clear then we should not allow it to be uploaded in the server" - We don't need it!

IgorOhrimenko:
Default value is `Leave it blank for accepting all media types.`
I leave it blank and I want to uploading `qwe.ovpn` file and I cannot. Why?

Sayan4444:
@IgorOhrimenko the problem is that previously when u try to upload something whose file type cant be determined,  not only the file was not uploaded but no error was given....thus this commit throws the error to let the end user know that this file cant be uploaded in the backend

emoxam:
Ok. How to upload OVPN file ?

roman-bergman:
I check mimetype for *.ovpn file. It has next type: text/plain
ovpn mimetype

I added text/plain to Accepted Media Types and but i get the same error: "media types not accepted".
I tried reboot my docker container and vm. It didn't produce any results.

emoxam:
ovpn file is text/plain
the same has txt files..

Can we turn off media type checking at all ?

DustpaN-Spb:
Look like a last change is a error
Because
Was:				invalidContentType: Boolean(file.type && !fileUploadIsValidContentType(file.type)),
Now:			invalidContentType: !(file.type && fileUploadIsValidContentType(file.type)),
If we can't detrminate a MIME type of a file (file.type) it will return a false and all not recognised types will be not work at all even if "Accepted Media Types" is empty.
Please revert

emoxam:
> Look like a last change is a error Because Was: invalidContentType: Boolean(file.type && !fileUploadIsValidContentType(file.type)), Now: invalidContentType: !(file.type && fileUploadIsValidContentType(file.type)), If we can't detrminate a MIME type of a file (file.type) it will return a false and all not recognised types will be not work at all even if "Accepted Media Types" is empty. Please revert https://forums.rocket.chat/t/cant-upload-ovpn-file/19455

Where it was ? What path and file you are talking about ? Who should revert it ?

DustpaN-Spb:
In this merge in tab "Files changed" you can see a changed files.
Author of a commit should do it?
I do not know how all work here.

apps/meteor/client/lib/chats/flows/uploadFiles.ts
DustpaN-Spb, line 1: If we can't detrminate a MIME type of a file (file.type) it will return a false and all not recognised types will be not work at all even if "Accepted Media Types" is empty.
Please revert

richardturnerfos:
How do I manage the mime types to allow my files?

D34DC3N73R:
I'd rather have no error message when uploading file types that are not allowed, than not be able to upload unrecognized MIME types when no restrictions are set. This should be reverted.

emoxam:
> I'd rather have no error message when uploading file types that are not allowed, than not be able to upload unrecognized MIME types when no restrictions are set. This should be reverted.

Totally agree!

Issue #30710, opened 2023-10-20
Link enclosed in parentheses is not clickable, but link preview is shown


paulchen:
### Description:

When enclosing a link in parentheses (without spaces between parentheses and the URL), a link preview will be generated (meaning the URL is recognized), but the URL itself will not be clickable.

### Steps to reproduce:

1. Navigate to any channel.
2. Write a message containing `(https://rocket.chat/)` (make sure to include the parentheses!) and submit it.

### Expected behavior:

Either no preview is shown and the link is not clickable; or a preview is shown and the link is clickable.

### Actual behavior:

A link preview is shown, but the link itself is not clickable:

![image](https://github.com/RocketChat/Rocket.Chat/assets/404840/dceb9b3e-c245-418a-8cdc-5eea3b3890e8)

### Server Setup Information:

- Version of Rocket.Chat Server: 6.4.2
- Deployment Method: Docker
- Number of Running Instances: 1
- MongoDB Version: 6.0.11

### Client Setup Information

The issue seems to be client-agnostic. However, I verified it with:

- Browser Version:
  - Chromium Version 118.0.5993.70 (Offizieller Build) Arch Linux (64-Bit)
  - Firefox 118.0.2 (64-bit)
- Operating System: Arch Linux

### Additional context

When searching existing issues regarding parentheses and links, you will stumble upon #13149. However, that issue describes a different problem that occurs when using Markdown to create a link.

sheikhaffan21:
Can I give it a try?

Sayan4444:
This issue is hapenning due a function called 'parse' which is a node_module. So a PR to the code-base cannot fix this issue. If you want to me to elaborate how and why I came to this conclusion, please do tell me. @paulchen 

https://github.com/RocketChat/Rocket.Chat/assets/112304873/ed67c004-589d-483f-aee9-7ac1c892a91d

This video shows the 'type' file and the .js file which is the main code is also present under the same directory inside node_moduels.

jarommadsen:
@Sayan4444 it is not an excuse to blame a dependency for a problem with RocketChat. If a dependency doesn't suite the needs of the project, it should be replaced, forked, or updated.

In this case though, `message-parser` is a `Rocket.Chat` package: https://github.com/RocketChat/Rocket.Chat/tree/9faf1ec91aaf16691d3e637b81e075d032b77778/packages/message-parser

ggazzo:
[...]

PR #31355, opened 2024-01-02
fix: Link enclosed in parentheses is not clickable, but link preview is shown

Devansht420:

BEFORE THE FIX:
image

AFTER THE FIX:
image

REASON BEHIND CHANGES:
Earlier when a url was contained in parentheses or apostrophes i wasnt clickable and still showed a preview of the link to fix this i made the changes in the regex so that now to doesnt show the preview if the url isnt passed properly.

## Issue(s)
This should close issues :-
closes [#30710](https://github.com/RocketChat/Rocket.Chat/issues/30710)
## Steps to test or reproduce
Navigate to any channel.
Write a message containing `(https://rocket.chat/)` (make sure to include the parentheses!) and submit it.


apps/meteor/lib/getMessageUrlRegex.ts
```diff
@@ -1,2 +1,2 @@
 export const getMessageUrlRegex = (): RegExp =>
-	/([A-Za-z]{3,9}):\/\/([-;:&=\+\$,\w]+@{1})?([-A-Za-z0-9\.]+)+:?(\d+)?((\/[-\+=!:~%\/\.@\,\w]*)?\??([-\+=&!:;%@\/\.\,\w]+)?(?:#([^\s\)]+))?)?/g;
+/(?:\(|'|^)(\w+:\/\/[-A-Za-z0-9\.]+(:\d+)?(\/[-\+=&?:%#\.\/\w]*)?)(?:\)|'|$)|(?:\(|'|^)(https?:\/\/[-A-Za-z0-9\.]+(:\d+)?(\/[-\+=&?:%#\.\/\w]*)?)(?:\)|'|$)/g;
```
hugocostadev, line 2: This regex blocks preview for links with spaces in it's boundaries...

image

hugocostadev (changes requested):

Devansht420:
Hey @hugocostadev I'll look into it, will try to come up with some better changes. Will update you.

hugocostadev:
> Hey @hugocostadev I'll look into it, will try to come up with some better changes. Will update you.

Thanks a lot @Devansht420 ! Keep making your amazing contributions! 🚀

hugocostadev:
> Hey @hugocostadev , I did some research and found out that the parsing is done via a grammar file in Fuselage, shouldn't the correct fix be by updating the file here -> https://github.com/RocketChat/fuselage/blob/91e502a0c43c90d589bd96039b5a671a21336876/packages/message-parser/src/grammar.pegjs#L243 ?
> 
> The regex is only used for generating preview, which I believe is working anyways.

But the issue is the preview... The link shouldn't be clickable in this scenario

Devansht420:
@hugocostadev , should i rectify the regex so that it solves the issue and the problem that you pointed out or should i work on a different approach?

Devansht420:
Hey @hugocostadev , I've pushed the changes in the regex, I think now its working fine. Please let me know your thoughts.

Devansht420:
hey @hugocostadev any update?
i just replaced spaces in the tab to an actual tab in the last commit to fix the lint but now its showing errors in CI which was not happening earlier.

PR #31361, opened 2024-01-03
fix: Mute/UnMute users in read-only channels

### LINKED ISSUE #27818: mute/unmute user on readonly room
### Description:

On readonly room the user action menu always shows `mute user` even that all users are mute 
### Steps to reproduce:

1. Create new room
2. add a user to it
3. set the room readonly
4. room members action menu does not show unmute user

### Expected behavior:

1. if user is muted show `unmute user`
2. if user is unmuted show `mute user`
### Actual behavior:

room members action menu does not show unmute user
### Server Setup Information:

- Version of Rocket.Chat Server:  5.4.1
- Operating System: ubuntu server 
- Deployment Method: docker
- DB Replicaset Oplog: 
- NodeJS Version: v14.19.3
- MongoDB Version: 5.0.14

### Client Setup Information

- Desktop App or Browser Version: both
- Operating System: [Screencast from 01-23-2023 07:29:50 PM.webm](https://user-images.githubusercontent.com/29951360/214120888-7daf4555-ff76-464c-a4b8-18d8afe4514c.webm)

yash-rajpal:
## Issue(s)

Fixes #27818

## Further comments

TC-813


apps/meteor/client/views/room/hooks/useMuteUserMutation.ts
```diff
@@ -0,0 +1,53 @@
...
+	const dispatchToastMessage = useToastMessageDispatch();
+	const t = useTranslation();
+	const queryClient = useQueryClient();
+
+	const dispatchMessage = isUserMuted ? 'User__username__unmuted_in_room__roomName__' : 'User__username__muted_in_room__roomName__';
+
+	const roomName = room?.t && escapeHTML(roomCoordinator.getRoomName(room.t, room));
+
+	return useMutation(
+		['muteUser', 'unmuteUser'],
```
gabriellsh, line 19: I think roomId should maybe be part of the dependency array. WDYT?

gabriellsh (changes requested):

apps/meteor/client/views/room/hooks/useMuteUserMutation.ts
yash-rajpal: Yes, it is a good idea to include deps in query key, but in this case we don't care about the returned result nor caching it, so didn't thought of making query key unique.

gabriellsh (dismissed):
LGTM!

apps/meteor/client/views/room/hooks/useMuteUserMutation.ts
tassoevan: @yash-rajpal Mutation keys are optional, AFAIK. Couldn't we just omit it?

apps/meteor/app/api/server/v1/users.ts
```diff
@@ -1240,6 +1241,27 @@ API.v1.addRoute(
...
+	'users.isMuted',
+	{ authRequired: true },
+	{
+		async get() {
+			const user = await getUserFromParams(this.queryParams);
+			const room = await Rooms.findOne(this.queryParams.roomId);
+
+			if (!room) {
+				throw new Error('invalid-room');
+			}
```
matheusbsilva137, line 1254: Could you please use `ajv` to validate this endpoints' params instead?
(Check `isUsersInfoParamsGetProps` for reference)

packages/rest-typings/src/v1/users.ts
```diff
@@ -367,6 +367,12 @@ export type UsersEndpoints = {
 	'/v1/users.deleteOwnAccount': {
 		POST: (params: { password: string; confirmRelinquish?: boolean }) => void;
 	};
+
+	'/v1/users.isMuted': {
+		GET: (params: { userId?: string; username?: string; roomId: string }) => {
```
matheusbsilva137, line 372: Please update this too after doing what's needed to use ajv
(Check how `UsersInfoParamsGet` is used in this file for reference)

apps/meteor/app/api/server/v1/users.ts
```diff
@@ -1240,6 +1241,27 @@ API.v1.addRoute(
 	},
 );
 
+API.v1.addRoute(
+	'users.isMuted',
```
matheusbsilva137, line 1245: I'm wondering if we should really implement this as a `users` endpoint or into somewhere else :thinking: What made me think about this is that we do not send a `roomId` in any other user related endpoint
This does looks ok to me as it is, but maybe having something such as `rooms.isMuted` or `subscriptions.isMuted` would look better since this is not related to the user in a global context. Would like to hear others' opinions, though
cc @MarcosSpessatto @gabriellsh

matheusbsilva137 (changes requested):

apps/meteor/app/api/server/v1/users.ts
yash-rajpal: Yes, makes sense, I was also skeptical about this. 

Moving this endpoint to Subscription makes more sense, I'll do that. Thanks!

apps/meteor/client/views/room/hooks/useMuteUserMutation.ts
yash-rajpal: Yep, makes no sense to have query key in such case, removed it. Thanks Tasso!

apps/meteor/app/api/server/v1/users.ts
yash-rajpal: Done!

apps/meteor/app/api/server/v1/users.ts
yash-rajpal: Done!

packages/rest-typings/src/v1/dm/im.ts
```diff
@@ -53,7 +53,7 @@ export type ImEndpoints = {
 
 	'/v1/im.members': {
 		GET: (params: DmMemberProps) => PaginatedResult<{
-			members: Pick<IUser, '_id' | 'status' | 'name' | 'username' | 'utcOffset'>[];
+			members: (Pick<IUser, '_id' | 'status' | 'name' | 'username' | 'utcOffset'> & { isMuted: boolean })[];
```
matheusbsilva137, line 56:
```suggestion
			members: Pick<IUser, '_id' | 'status' | 'name' | 'username' | 'utcOffset'>[];
```
Since DM members can't be muted, I'm not sure if we should have this new field here. I guess the `im.members` and `dm.members` endpoints haven't been changed also, so rolling back this change should be fine

packages/rest-typings/src/v1/rooms.ts
```diff
@@ -423,6 +423,40 @@ const GETRoomsNameExistsSchema = {
...
+					type: 'string',
+				},
+			},
+			required: ['username', 'roomId'],
+			additionalProperties: false,
+		},
+	],
+};
+
+export const isRoomsIsUserMuted = ajv.compile<RoomsIsUserMuted>(RoomsIsUserMutedSchema);
```
matheusbsilva137, line 458:
```suggestion
export const isRoomsIsUserMutedProps = ajv.compile<RoomsIsUserMuted>(RoomsIsUserMutedSchema);
```
Just a small detail, but I think that's the naming pattern.

packages/rest-typings/src/v1/rooms.ts
```diff
@@ -572,4 +606,10 @@ export type RoomsEndpoints = {
 			discussions: IRoom[];
 		}>;
 	};
+
+	'/v1/rooms.isUserMuted': {
+		GET: (params: { userId: string; roomId: string } | { username: string; roomId: string }) => {
```
matheusbsilva137, line 611:
```suggestion
		GET: (params: RoomsIsUserMuted) => {
```
You can use the type defined above.

matheusbsilva137 (changes requested):
I'd also suggest adding some end-to-end tests for the new endpoint. Two test cases (true/false) should be enough.

apps/meteor/app/api/server/v1/rooms.ts
```diff
@@ -635,3 +639,36 @@ API.v1.addRoute(
...
+
+			if (!subscription) {
+				throw new Error('error-invalid-subscription');
+			}
+
+			const roomCanMute = await roomCoordinator.getRoomDirectives(room.t).allowMemberAction(room, RoomMemberActions.MUTE, user._id);
+
+			if (!roomCanMute) {
+				return API.v1.failure();
+			}
```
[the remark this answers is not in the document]
yash-rajpal: Makes sense. Done!

apps/meteor/app/api/server/v1/groups.ts
```diff
@@ -704,12 +709,9 @@ API.v1.addRoute(
 	{ authRequired: true },
 	{
 		async get() {
-			const findResult = await findPrivateGroupByIdOrName({
-				params: this.queryParams,
-				userId: this.userId,
-			});
+			const room = await getRoomFromParams(this.queryParams, { muted: 1, unmuted: 1 });
```
matheusbsilva137, line 712: Oops, this is dangerous :grimacing: 
`getRoomFromParams` misses some checks we have in `findPrivateGroupByIdOrName` (such as checking if the user has acces to the room or if it is archived). I'd suggest to keep using `findPrivateGroupByIdOrName`, but add an optional `roomFields` param to it so that you can send `{ muted: 1, unmuted: 1 }`
It should have just the same effect

packages/rest-typings/src/v1/rooms.ts
```diff
@@ -423,6 +423,40 @@ const GETRoomsNameExistsSchema = {
...
+	type: 'object',
+	oneOf: [
+		{
+			properties: {
+				userId: {
+					type: 'string',
+				},
+				roomId: {
+					type: 'string',
+				},
```
matheusbsilva137, line 438:
```suggestion
				userId: {
					type: 'string',
					minLength: 1,
				},
				roomId: {
					type: 'string',
					minLength: 1,
				},
```
I think it'd be good not to accept empty strings. We handle this case internally, but since ajv can check it, then why not? (this also avoids calling `Rooms.findOne` searching for an empty `_id`, which returns `undefined` for us but I think that's something to avoid)
Sorry for not suggesting this before, though

packages/rest-typings/src/v1/rooms.ts
```diff
@@ -423,6 +423,40 @@ const GETRoomsNameExistsSchema = {
...
+			additionalProperties: false,
+		},
+		{
+			properties: {
+				username: {
+					type: 'string',
+				},
+				roomId: {
+					type: 'string',
+				},
```
matheusbsilva137, line 450:
```suggestion
				username: {
					type: 'string',
					minLength: 1,
				},
				roomId: {
					type: 'string',
					minLength: 1,
				},
```
Same here

matheusbsilva137 (changes requested):

apps/meteor/app/api/server/v1/groups.ts
matheusbsilva137: This made me find out we have no tests for trying to get members from a private group someone doesn't have access to :grimacing:

packages/rest-typings/src/v1/rooms.ts
yash-rajpal: Nice idea, don't worry about it. :)

PR #31363, opened 2024-01-03
fix: remove reason when password change is set to false

tapiarafael:
Remove the `requirePasswordChangeReason` when the field `requirePasswordChange` is set to false.

## Further comments

[SUP-444](https://rocketchat.atlassian.net/browse/SUP-444)

[SUP-444]: https://rocketchat.atlassian.net/browse/SUP-444?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/meteor/app/lib/server/functions/saveUser.js
```diff
@@ -416,8 +416,11 @@ export const saveUser = async function (userId, userData) {
 		updateUser.$set.language = userData.language;
 	}
 
-	if (typeof userData.requirePasswordChange !== 'undefined') {
-		updateUser.$set.requirePasswordChange = userData.requirePasswordChange;
+	console.log({ userData });
```
MarcosSpessatto, line 419:
```suggestion
```

MarcosSpessatto (changes requested):
Can we add a test case to cover this?

tapiarafael:
> Can we add a test case to cover this?

Not for this case, since we cannot simulate the `requirePasswordChangeReason` being populated to check if it was deleted when the `requirePasswordChange` is set to false :disappointed:

MarcosSpessatto (dismissed):

apps/meteor/app/lib/server/functions/saveUser.js
```diff
@@ -416,8 +416,9 @@ export const saveUser = async function (userId, userData) {
 		updateUser.$set.language = userData.language;
 	}
 
-	if (typeof userData.requirePasswordChange !== 'undefined') {
-		updateUser.$set.requirePasswordChange = userData.requirePasswordChange;
+	updateUser.$set.requirePasswordChange = userData?.requirePasswordChange || false;
```
pierre-lehnen-rc, line 419: This will cause the flag to be turned off anytime the user is saved unless it is explicitly turning the flag on. Is that really the intended behavior?

tapiarafael: As far as I checked, everytime the user is saved the `requirePasswordChange` is sent (along with some other fields), so that wouldn't be a problem. But if you worry about unintended change, we can keep as it was and just add a validation to delete the `requirePasswordChangeReason` if the `requirePasswordChange` is set to false.

pierre-lehnen-rc: That would be the best approach, thanks!

MarcosSpessatto (approved):

apps/meteor/app/lib/server/functions/saveUser.js
```diff
@@ -418,6 +418,9 @@ export const saveUser = async function (userId, userData) {
 
 	if (typeof userData.requirePasswordChange !== 'undefined') {
 		updateUser.$set.requirePasswordChange = userData.requirePasswordChange;
+		if (!userData.requirePasswordChange) {
+			updateUser.$unset.requirePasswordChangeReason = 1;
```
pierre-lehnen-rc, line 422: $unset may still be undefined here.

pierre-lehnen-rc (changes requested):
Please add a test that covers this part of the function - even if you can't validate this specific bug in an api test, we would at least have a test running the functionality

tapiarafael:
> Please add a test that covers this part of the function - even if you can't validate this specific bug in an api test, we would at least have a test running the functionality

I didn't find a way to test it.
E2E tests can't cover this scenario, and I couldn't write a unit test since the `saveUser.js` file has many dependencies.

pierre-lehnen-rc:
> > Please add a test that covers this part of the function - even if you can't validate this specific bug in an api test, we would at least have a test running the functionality
> 
> @pierre-lehnen-rc, I didn't find a way to test it. E2E tests can't cover this scenario, and I couldn't write a unit test since the `saveUser.js` file has many dependencies.

E2E **can** cover this, just create an user with the reason set, then write a test that checks if the reason is removed when the flag is set to false.

pierre-lehnen-rc (dismissed):

MarcosSpessatto (approved):

MarcosSpessatto (changes requested):
Can we add a changeset?

pierre-lehnen-rc (approved):

MarcosSpessatto (approved):

matheusbsilva137 (approved):

PR #31372, opened 2024-01-04
ci: update ubuntu version on CI

MarcosSpessatto:
Extracted from #29112


.github/workflows/ci.yml
```diff
@@ -82,7 +82,7 @@ jobs:
 
   notify-draft-services:
     name: 🚀 Notify external services - draft
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-22.04
```
sampaiodiego, line 85: any reason for just this one being updated (other than it was just this one updated on the Meteor 3.0 PR 😛 )?

there are other using the 20.04 version, like:
https://github.com/RocketChat/Rocket.Chat/blob/47331bacc3e789db6d4cdd88afb271dc6c3eddb5/.github/workflows/ci.yml#L484

and some others using `ubuntu-latest` that would be good to have a fixed version as well, like:
https://github.com/RocketChat/Rocket.Chat/blob/47331bacc3e789db6d4cdd88afb271dc6c3eddb5/.github/workflows/ci.yml#L22

sampaiodiego (changes requested):

PR #31373, opened 2024-01-04
regression: `AppRow` bundleIn verification

juliajforesti:


dougfabris (approved):

geekgonecrazy:
We should potentially alert title as this one effected 6.5 not just develop.  So is more of a fix then regression

PR #31375, opened 2024-01-04
chore: services changes on lifecycle methods + throw if svc is not available

MarcosSpessatto:
Extracted from #29112

### DIFF
--- a/packages/core-services/src/LocalBroker.ts
+++ b/packages/core-services/src/LocalBroker.ts
@@ -8,6 +8,8 @@ import type { IBroker, IBrokerNode } from './types/IBroker';
 import type { ServiceClass, IServiceClass } from './types/ServiceClass';
 
 export class LocalBroker implements IBroker {
+	private started = false;
+
 	private methods = new Map<string, (...params: any) => any>();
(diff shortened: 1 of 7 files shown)


packages/core-services/src/LocalBroker.ts
```diff
@@ -73,6 +73,7 @@ export class LocalBroker implements IBroker {
 
 			this.methods.set(`${namespace}.${method}`, i[method].bind(i));
 		}
+		void instance.started();
```
sampaiodiego, line 76: we should call this only after the broker has been started as well

sampaiodiego (changes requested):

packages/core-services/src/LocalBroker.ts
MarcosSpessatto: That makes sense, I've added a flag to know if the broker has already started. To add more context, the idea here and the problem it has is that the `started()` method within the services is not being called after the broker has already started, so if you start the broker and create a service after that, this was not called.

sampaiodiego (approved):

PR #31383, opened 2024-01-05
fix!: hide archived rooms in room.autocomplete endpoint

heitortanoue:

Issue with the create option discussion, the list when you choose the Channel or Group is showing Archived Teams/groups/channels

## Steps to test or reproduce

- Create at least one channel/group/team and then archive them

- Choose to create a discussion and then you will be able to pick this archived channel from the drop down suggestion list for parent channel

## Further comments

SUP-286


apps/meteor/server/models/raw/Rooms.ts
```diff
@@ -294,7 +294,10 @@ export class RoomsRaw extends BaseRaw<IRoom> implements IRoomsModel {
 				},
 			],
 			prid: { $exists: false },
-			$and: [{ $or: [{ federated: { $exists: false } }, { federated: false }] }],
+			$and: [
+				{ $or: [{ federated: { $exists: false } }, { federated: false }] },
+				{ $or: [{ archived: { $exists: false } }, { archived: false }] },
+			],
```
matheusbsilva137, line 300:
```suggestion
			$and: [{ federated: { $ne: true } }, { archived: { $ne: true } }],
```
Your suggestion works, but I think it's more concise and readable like this.

matheusbsilva137 (changes requested):

matheusbsilva137 (commented):
Also, I think we should consider this a **breaking change** since this changes the output of the `rooms.autocomplete.channelAndPrivate` endpoint. I'll approve it for now since it looks fine to me, but I'll mark it as a major change.
Please tell me if you agree, @MarcosSpessatto

matheusbsilva137 (dismissed):

MarcosSpessatto:
> Also, I think we should consider this a **breaking change** since this changes the output of the `rooms.autocomplete.channelAndPrivate` endpoint. I'll approve it for now since it looks fine to me, but I'll mark it as a major change. Please tell me if you agree, @MarcosSpessatto

Yes, I agree. Let's keep this as a breaking change then. 👍

MarcosSpessatto (approved):

.changeset/three-avocados-search.md
```diff
@@ -0,0 +1,5 @@
+---
+"@rocket.chat/meteor": major
+---
+
+Fixed archived rooms returned in `rooms.autocomplete.channelAndPrivate` endpoint
```
ggazzo, line 5: the description does not explain to me what changes for this to be a break change

ggazzo (changes requested):

ggazzo (approved):

PR #31388, opened 2024-01-05
chore: remove RedHat stuff

sampaiodiego:


geekgonecrazy (approved):
if anyone from redhat wants to reach out and have a conversation about this please do.  The integration broke with the changes to the registry to quay.io so seems like now we don't trigger an image but we push an image?

We've reached out but haven't been able to get in contact.  So for now agreed to clean this up and remove it.

debdutdeb (approved):

PR #31411, opened 2024-01-10
fix: better error handling in reportMessage function

Dnouv:

- updated access checks for app users. Now, if an app user reports a message, we won't check for room access. This should make the process smoother for app users. 
- unique and more informative error if the room is not valid
## Further comments

- Currently, no visual feedback is in place; it only handles errors in API calls and calls via Rocket.Chat Apps.Engine Moderation Bridge

[AECO-471](https://rocketchat.atlassian.net/browse/AECO-471)

[AECO-471]: https://rocketchat.atlassian.net/browse/AECO-471?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.changeset/quiet-trainers-jog.md
+++ b/.changeset/quiet-trainers-jog.md
@@ -1,0 +1,5 @@
+---
+'@rocket.chat/meteor': minor
+---
+
+optimized message reporting for improved error handling. Additionally, app users can now report messages without room access checks.
--- a/apps/meteor/server/lib/moderation/reportMessage.ts
+++ b/apps/meteor/server/lib/moderation/reportMessage.ts
@@ -25,11 +25,18 @@ export const reportMessage = async (messageId: IMessage['_id'], description: str
 		throw new Error('error-invalid-user');
 	}
 
+	const isAppUser = user?.type === 'app';
+
 	const { rid } = message;
 	// If the user can't access the room where the message is, report that the message id is invalid
 	const room = await Rooms.findOneById(rid);
-	if (!room || !(await canAccessRoomAsync(room, { _id: uid }))) {
-		throw new Error('error-invalid-message_id');
+
+	if (!room) {
+		throw new Error('error-invalid-room');
+	}
+
+	if (!isAppUser && !(await canAccessRoomAsync(room, { _id: uid }))) {
+		throw new Error('error-not-allowed');
 	}
 
 	const reportedBy = {


AllanPazRibeiro (commented):
Could record the screen showing how the UI deals with these errors? it can be a gif (I like the [peek](https://github.com/phw/peek) or any)

I am asking because I don't know how it could be for the end user

AllanPazRibeiro (approved):
Given the absence of visual feedback at present, it appears to be primarily an enhancement in error handling. LGTM

debdutdeb (changes requested):
Why? :)

Looking at the task, yes sure the error isn't a good one. But is this a good idea? 

App users are still users. While there is some weirdness with how it's handled and how it sort of ignore a lot of permissions in different places, accessing a room via membership is a clear indication of what app / user has access to the private content vs what does not.

E.g. i know of various private rooms in open server, i can update the app, then add the room name to the setting, then it can log every private message from there. 

Understand what I mean? :) 

I'm putting this CR in place to avoid merging immediately. I'd like to hear some thoughts from other folks like @sampaiodiego @d-gubert maybe others (pls mention forward)

Dnouv:
Thank you for your insightful comment, @debdutdeb . I completely understand your concerns about security and ensuring that app users have the appropriate access rights.

However, I am afraid it is out of the scope of this PR. This might be an improvement in how we trigger and read the messages in Apps.Engine. But I got your point that allowing App users to bypass `reportMessage` without room access is similar, but is uniform with the other access. 

This is an excellent idea to consider in the improvement of the security and access aspects of Apps.Engine.

Maybe additional checks in the call here, 
https://github.com/RocketChat/Rocket.Chat/blob/0da74237d4cc9ab283a89238bca9579a46072dbe/apps/meteor/app/lib/server/functions/sendMessage.ts#L281-L282


> Why? :)

On the Moderation App, the App Admins can select which room they want to moderate. (Only the private rooms they have access to will appear in the settings; other rooms to which the App Admin does not have access won't show up, so this acts as sort of an agreement to read room messages).
Once we get the message in the App, process it, and it is flagged, the App heads to reporting the message, but it fails because the App user is not in the channel and thus has no access. 

This is technically incorrect since the App admin already configured the private and public channels they want to moderate, and if they were to add in each channel the App user, that might be bad for UX.(I might be wrong; I am no UX expert 🙈).

I look forward to your feedback.

Thank you

debdutdeb:
> However, I am afraid it is out of the scope of this PR.

I disagree. Since the pr is adding such a behaviour into the product, it definitely puts it under scope. 

If thw goal is to avoid unclear error message, let's just make that unclear. Not add a passthrough behaviour in, that indirectly solves the problem. 

> On the Moderation App, the App Admins can select which room they want to moderate. (Only the private rooms they have access to will appear in the settings; other rooms to which the App Admin does not have access won't show up, so this acts as sort of an agreement to read room messages).
Once we get the message in the App, process it (only messages of the selected rooms), and it is flagged, the App heads to reporting the message, but it fails because the App user is not in the channel and thus has no access.

I understand your point, but we're still assuming there won't be any leaks anywhere in the middle to cause huge leaks through apps. Maybe I'm thinking about it too much? But to me a passthrough like this, I don't feel like a great idea. 

This also puts burden into our app reviewers in reviewing logic to make sure leaks aren't happening (e.g. through slashcommands), which is .. I can't think of a word right now. 

> This is technically incorrect since the App admin already configured the private and public channels they want to moderate, and if they were to add in each channel the App user, that might be bad for UX.(I might be wrong; I am no UX expert 🙈).

I can't comment on ux. I don't feel like it needs to be ux discussion either. The app can simply join the room configured by itself. Private, can't. Which is beautiful imo. 

Do private rooms require moderation to begin with? What is a "private" room? 

If a private room really needs moderation like this, adding a user account manually isn't a huge deal. Because then it'll be just as easy to disable moderation of that room by the room's moderator, by simply kicking the bot. No admin intervention needed. 

This is what we follow for the koko app and its out of office notifications per departments as well, btw.

PR #31417, opened 2024-01-10
feat: Update autotranslate on language preference change

Gustrb:

Now every time the user updates their language to a different language and the `AutoTranslate_AutoEnableOnJoinRoom` is set, we need to update all of his subscriptions to reflect the new AutoTranslate language.

## Steps to test or reproduce

Follow [this](https://docs.rocket.chat/setup-and-configure/advanced-workspace-management/google-cloud) to enable autotranslate:

1. Join a room
2. Mark the autotranslate checkbox (inside the kebab menu)
3. Change the language setting to one that is different from the room
4. Go back to the room and check if all the next messages are translated to the users new language

## Further comments

TC-1076


matheusbsilva137 (approved):
Congrats on your first PR :rocket: :partying_face: 
LGTM. I see we do not have an index for `u._id: 1, autoTranslate: 1`, but I'd say it's ok to move on as it is since we already have an index on the `autoTranslate` field (let's wait for a review from Architecture)

MarcosSpessatto (approved):

.changeset/sixty-bananas-sit.md
```diff
@@ -0,0 +1,6 @@
+---
+"@rocket.chat/meteor": patch
+"@rocket.chat/model-typings": patch
```
matheusbsilva137, line 3:
```suggestion
"@rocket.chat/meteor": minor
"@rocket.chat/model-typings": minor
```
Since this is a new feature (and not a fix), we should mark this as minor changes instead

matheusbsilva137 (changes requested):

matheusbsilva137 (approved):

apps/meteor/server/models/raw/Subscriptions.ts
```diff
@@ -603,6 +603,21 @@ export class SubscriptionsRaw extends BaseRaw<ISubscription> implements ISubscri
 		return this.updateOne(query, update);
 	}
 
+	updateAllAutoTranslateLanguagesByUserId(userId: IUser['_id'], language: string): Promise<UpdateResult | Document> {
+		const query = {
+			'u._id': userId,
+			'autoTranslate': true,
```
ggazzo, line 609: I would like to ask, what happens if I set a different language for a specific room? Will I lose my configuration? Are we heading in a direction where language (based on user preference) is the way to go? If so, I wouldn't save the language in the subscription anymore, but use the user's preference to determine the language.

ggazzo (dismissed):

apps/meteor/server/models/raw/Subscriptions.ts
Gustrb: Yes, whenever you update your user's language, all the subscription's `autoTranslateLanguage` would be set to be the new user's language. I guess using the user's preference would be better, but I am not sure whether the feature to set a specific language for a specific channel is altogether not useful to keep, and I believe it would be a lot more complicated to have some kind of hierarchy of languages to choose instead of just using the Subscription value always.

I guess we could do something like this, to define such hierarchy:
```
if subscription.autoTranslate && subscription.autoTranslateLanguage
  translateMessages(subscription.autoTranslateLanguage)
else
  translateMessages(user.language)
```

But I am not sure if this would be preferable, we could also have a setting for this, but I am not sure, what do you think @matheusbsilva137

matheusbsilva137: > I would like to ask, what happens if I set a different language for a specific room? Will I lose my configuration? 

Yes, you would. That's the behavior we planned for this case.

> Are we heading in a direction where language (based on user preference) is the way to go?

Not exactly. That will only be true when the new `AutoTranslate_AutoEnableOnJoinRoom` setting is enabled. This setting has been introduced recently, and its idea is to turn on auto-translate automatically when a user who has a language preference different from the workspace's default joins a room. We're introducing this new behavior now so that the auto-translate language is always synced with the language preference **when this setting is enabled**.

matheusbsilva137: @gustrb liked your approach, that'd allow users to set custom translation preferences on rooms and when that's not defined we'd just use their language preference.

Still, generally speaking, I'm not sure it's really worth changing the approach we already have since I can't think of any practical benefits :thinking: When the language preference is changed, we'd still need to run through the user's subscription (in this case, to remove the `autoTranslateLanguage` field) AND we'd even need to add a preference "query" to check their language.

In my perspective, since this is a specific case for when a setting is turned on, I don't believe it's worth creating a whole new flow for that (also, I can't see any performance improvements on it). Would like to hear @ggazzo thoughts about it, though.

apps/meteor/server/models/raw/Subscriptions.ts
```diff
@@ -603,6 +603,21 @@ export class SubscriptionsRaw extends BaseRaw<ISubscription> implements ISubscri
 		return this.updateOne(query, update);
 	}
 
+	updateAllAutoTranslateLanguagesByUserId(userId: IUser['_id'], language: string): Promise<UpdateResult | Document> {
+		const query = {
```
KevLehman, line 607: We May need an index for the query.

apps/meteor/server/methods/saveUserPreferences.ts
```diff
@@ -169,6 +172,10 @@ export const saveUserPreferences = async (settings: Partial<UserPreferences>, us
 		if (Array.isArray(settings.highlights)) {
 			await Subscriptions.updateUserHighlights(user._id, settings.highlights);
 		}
+
+		if (settings.language && oldLanguage !== settings.language && rcSettings.get('AutoTranslate_AutoEnableOnJoinRoom')) {
+			await Subscriptions.updateAllAutoTranslateLanguagesByUserId(user._id, settings.language);
+		}
```
KevLehman, line 178: About this one, can we try to update other methods on this setImmediate to use a single query when possible?

This callback does a bunch of heavy ops that could potentially update a ton of records, so if we can make them go on the fewest possible calls, that would be ideal.

You can do those changes with some tests after this one is merged and before the release is cut, wdyt?

Gustrb: Yea, that sounds good

MarcosSpessatto (approved):

KevLehman (approved):

apps/meteor/server/models/raw/Subscriptions.ts
rodrigok: IMO the best solution would be:
1. Have a user's preference for the language
2. Have a subscription's preference for the language (which by default is "inherited" from the user's preference, or in other words, it says to use the user's preference setting
3. The check for the language to use when processing the translation checks the subscription setting, it not set or set to inherit it uses the user's setting
4. A button to reset all subscription defined language to "inherited"

That way we prevent an update on many records if the user changes his language, which can be heavy and can happen by mistake.

What do you guys think?

scuciatto: In think it is a good approach. As we talked earlier today, let's bring these improvements to another task.

ggazzo (approved):

PR #31423, opened 2024-01-11
feat: Increase the limit of members that can be added using the autojoin feature in a team's channel

Gustrb:
There existed an implicit limit of 50 members that would be included whenever a a channel with auto join was created inside a team, this task ties the limit to the API_Users_Limit setting, so it is configurable.

## Steps to test or reproduce

This task requires a team with over 1000 members, so all of these members should be created and associated with the team
1. Create a channel from the team
2. Open the kebab-menu and mark the auto join box
All the members should've been added correctly
3. Add one more user to the team
4. Create a new channel in the team
5. Mark the auto join box
Now the new user should not have been added since there is a hard limit of 1000 members

## Further comments

[SUP-34]

[SUP-34]: https://rocketchat.atlassian.net/browse/SUP-34?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/meteor/server/services/team/service.ts
```diff
@@ -473,7 +473,8 @@ export class TeamService extends ServiceClassInternal implements ITeamService {
 		await Rooms.setTeamDefaultById(rid, isDefault);
 
 		if (room.teamDefault) {
-			const teamMembers = await this.members(uid, room.teamId, true, undefined, undefined);
+			const MAX_NUMBER_OF_AUTO_JOIN_MEMBERS = 1000;
```
matheusbsilva137, line 476: Even though this is the only constant being used in this file, I'd suggest adding this to a `constants.ts` file. You can create one in `apps/meteor/server/services/team/lib/constants.ts`

matheusbsilva137 (changes requested):

apps/meteor/server/services/team/service.ts
Gustrb: great idea!

MarcosSpessatto (commented):
Shouldn't we add a warning (already mentioned in the task) to let the channel's creator know that a limit does exist (Even if this limitation already exists)? Otherwise, the creator could (wrongly) think that all team members will join automatically.

Gustrb:
For sure @MarcosSpessatto, just found out there were comments in JIRA.
I will add the popup

apps/meteor/client/views/teams/contextualBar/channels/RoomActions.js
```diff
@@ -81,10 +81,25 @@ const RoomActions = ({ room, reload }) => {
 	const menuOptions = useMemo(() => {
 		const AutoJoinAction = async () => {
 			try {
-				await updateRoomEndpoint({
+				const { numberOfMembers } = await updateRoomEndpoint({
 					roomId: rid,
 					isDefault: !room.teamDefault,
 				});
+
+				const MAX_NUMBER_OF_AUTO_JOIN_MEMBERS = 1000;
```
Gustrb, line 89: Is there a constants file I could put this one?

apps/meteor/app/api/server/v1/teams.ts
```diff
@@ -242,8 +242,9 @@ API.v1.addRoute(
 			const canUpdateAny = !!(await hasPermissionAsync(this.userId, 'view-all-team-channels', team.roomId));
 
 			const room = await Team.updateRoom(this.userId, roomId, isDefault, canUpdateAny);
+			const numberOfMembers = await TeamMember.countDocuments({ teamId: team._id });
```
yash-rajpal, line 245: Here to get the total number of team-members, instead of querying the database again, you can return the total members from the above `Team.updateRoom` function call. The above function already has that info.

apps/meteor/server/services/team/service.ts
```diff
@@ -473,7 +474,7 @@ export class TeamService extends ServiceClassInternal implements ITeamService {
 		await Rooms.setTeamDefaultById(rid, isDefault);
 
 		if (room.teamDefault) {
-			const teamMembers = await this.members(uid, room.teamId, true, undefined, undefined);
+			const teamMembers = await this.members(uid, room.teamId, true, { offset: 0, count: MAX_NUMBER_OF_AUTO_JOIN_MEMBERS });
```
yash-rajpal, line 477: Here it also returns the total count of members, we can return this total members count from this function and use above.

apps/meteor/client/views/teams/contextualBar/channels/RoomActions.js
yash-rajpal: Yes, we can share the same constant on client and server by putting it in some `meteor/lib` folder context.

Any code inside `lib` folder is available on both client and server, for example take a look at - `apps/meteor/lib/videoConference/constants.ts` file

yash-rajpal (changes requested):

apps/meteor/server/services/team/service.ts
```diff
@@ -472,12 +483,16 @@ export class TeamService extends ServiceClassInternal implements ITeamService {
...
+		if (isDefault) {
+			const teamMembers = await this.members(uid, room.teamId, true, { offset: 0, count: MAX_NUMBER_OF_AUTO_JOIN_MEMBERS });
 
 			for await (const m of teamMembers.records) {
 				await addUserToRoom(room._id, m.user, user);
 			}
+			return {
+				room,
+				numberOfMembers: teamMembers.total,
+			};
```
matheusbsilva137, line 495:
```suggestion
```
I don't think we need this new `numberOfMembers` field to be returned here. You can use `room.usersCount` instead, we already store this info in the room.  By doing that, you also won't need multiple return types for this method.

packages/core-services/src/types/ITeamService.ts
```diff
@@ -57,6 +57,11 @@ export interface IListRoomsFilter {
 	allowPrivateTeam: boolean;
 }
 
+export interface ISetDefaultRoomForTeamUpdateInfo {
+	room: IRoom;
+	numberOfMembers: number;
+}
+
```
matheusbsilva137, line 64:
```suggestion
```
Since the room already stores this info, we should not need this new type too.

apps/meteor/client/views/teams/contextualBar/channels/RoomActions.js
```diff
@@ -81,10 +82,24 @@ const RoomActions = ({ room, reload }) => {
 	const menuOptions = useMemo(() => {
 		const AutoJoinAction = async () => {
 			try {
-				await updateRoomEndpoint({
+				const { numberOfMembers } = await updateRoomEndpoint({
```
matheusbsilva137, line 85: Try using `usersCount` here instead.

.changeset/funny-plums-protect.md
matheusbsilva137, line 1: chore PRs should not have changesets since they're not intended to be shown in the release notes. I think it'd be ok to tag this PR as a `feat, though

matheusbsilva137 (changes requested):

apps/meteor/server/services/team/service.ts
Gustrb: You're right, the approach I took also has a subtle problem I just found out, that the channel could have more users than the number of members in the team, so the number would be messed up.

apps/meteor/server/services/team/service.ts
```diff
@@ -472,8 +473,8 @@ export class TeamService extends ServiceClassInternal implements ITeamService {
 		room.teamDefault = isDefault;
 		await Rooms.setTeamDefaultById(rid, isDefault);
 
-		if (room.teamDefault) {
-			const teamMembers = await this.members(uid, room.teamId, true, undefined, undefined);
+		if (isDefault) {
+			const teamMembers = await this.members(uid, room.teamId, true, { offset: 0, count: MAX_NUMBER_OF_AUTO_JOIN_MEMBERS });
 
 			for await (const m of teamMembers.records) {
 				await addUserToRoom(room._id, m.user, user);
```
matheusbsilva137, line 480:
```suggestion
				(await addUserToRoom(room._id, m.user, user)) && room.usersCount++;
```

I've just found out that this endpoint is returning an old (=incorrect) `usersCount` 
We return the room that has been fetched before, so it is not counting the users that have just been added. So that we don't need to search it again, we can increment the value as I suggested above
Please update the end-to-end tests to cover this

apps/meteor/client/views/teams/contextualBar/channels/RoomActions.js
```diff
@@ -81,10 +82,24 @@ const RoomActions = ({ room, reload }) => {
 	const menuOptions = useMemo(() => {
 		const AutoJoinAction = async () => {
 			try {
-				await updateRoomEndpoint({
+				const { room: updatedRoom } = await updateRoomEndpoint({
 					roomId: rid,
 					isDefault: !room.teamDefault,
 				});
+
+				if (!room.teamDefault) {
```
matheusbsilva137, line 90:
```suggestion
				if (updatedRoom.teamDefault) {
```
Maybe this way it's easier to read.

matheusbsilva137 (changes requested):

apps/meteor/packages/rocketchat-i18n/i18n/en.i18n.json
```diff
@@ -5003,6 +5003,8 @@
   "Team_Add_existing_channels": "Add Existing Channels",
   "Team_Add_existing": "Add Existing",
   "Team_Auto-join": "Auto-join",
+  "Team_Auto-join_exceeded_user_limit": "Auto-join has a limit of {{limit}} members, #{{channelName}} now has that {{numberOfMembers}} members",
```
matheusbsilva137, line 5006:
```suggestion
  "Team_Auto-join_exceeded_user_limit": "Auto-join has a limit of {{limit}} members, #{{channelName}} now has {{numberOfMembers}} members",
```

matheusbsilva137 (changes requested):

matheusbsilva137 (dismissed):

ggazzo (changes requested):

apps/meteor/lib/team/constants.ts
```diff
@@ -0,0 +1 @@
+export const MAX_NUMBER_OF_AUTO_JOIN_MEMBERS = 1000;
```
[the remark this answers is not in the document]
Gustrb: I wonder a bit, performance was raised as a concern in this task, should we give the user full control about the value considering this?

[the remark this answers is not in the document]
Gustrb: @milton-rucks 's commentary on the ticket:

> We don’t want to invest effort in it. In 3 years that we have had the teams feature, that was the only complaint about this limitation. I completely agree that changing the code to be more resilient/fast would be a better approach, but I don’t think it’s worth the investment. So, the solution that we had previously discussed with Henriques would be to increase the limit.

apps/meteor/client/views/teams/contextualBar/channels/RoomActions.js
```diff
@@ -81,10 +82,25 @@ const RoomActions = ({ room, reload }) => {
...
 		const AutoJoinAction = async () => {
 			try {
-				await updateRoomEndpoint({
+				const { room: updatedRoom } = await updateRoomEndpoint({
 					roomId: rid,
 					isDefault: !room.teamDefault,
 				});
+
+				if (updatedRoom.teamDefault) {
+					const messageType = updatedRoom.usersCount > MAX_NUMBER_OF_AUTO_JOIN_MEMBERS ? 'info' : 'success';
```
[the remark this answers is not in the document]
ggazzo: @Gustrb  I think these two numbers should go together (be the same), and the message could just say that the number has exceeded the limit of users in the room. I agree that the name of the setting (API_User_Limit) is so bad, but it's what we use nowadays. maybe we could rename/move to another place

matheusbsilva137 (approved):

PR #31439, opened 2024-01-12
chore: reorganize server imports

ggazzo:


apps/meteor/server/main.ts
```diff
@@ -1,85 +1,37 @@
...
-import './routes/health';
-import './routes/i18n';
-import './routes/timesync';
-import './routes/userDataDownload';
-import './stream/stdout';
 
+import './stream/stdout';
 import './features/EmailInbox/index';
+
+import '../imports/personal-access-tokens/server/api/methods';
```
sampaiodiego, line 28: should this import be moved to the file `apps/meteor/server/methods/index.ts`?

sampaiodiego (approved):

PR #31450, opened 2024-01-15
i18n: Update th-TH.i18n.json - Fix typos, wrong translations, inconsistencies; Add more translations

bact:

- Use country names and spellings according to Thailand Ministry of Foreign Affairs / Office of the Royal Society of Thailand
- Remove "ประเทศ" (country) prefixes from some countries' translation, to make it consistent with other countries in the same file (that their translation doesn't have that prefix)
- Fix "Togo" (country) translation (now translated literally as "to go") and "Turkey" (now translated literally as Turkey the bird)
- Translate "Domain"
- Fix "ลิงค์" to "ลิงก์"


bact:
(The best way to have translations for language and territory names is probably just use ones from Unicode CLDR https://cldr.unicode.org/index/downloads )

PR #31455, opened 2024-01-15
feat(UiKit): Introduce users/channels selectors

tiagoevanp:

### DIFF
--- a/packages/fuselage-ui-kit/src/elements/UsersSelectElement/UsersSelectElement.tsx
+++ b/packages/fuselage-ui-kit/src/elements/UsersSelectElement/UsersSelectElement.tsx
@@ -1,0 +1,169 @@
+import { MultiSelectFiltered, Box, Chip } from '@rocket.chat/fuselage';
+import { useDebouncedValue } from '@rocket.chat/fuselage-hooks';
+import { useEndpoint } from '@rocket.chat/ui-contexts';
+import type * as UiKit from '@rocket.chat/ui-kit';
+import { useQuery } from '@tanstack/react-query';
+import type { ReactElement } from 'react';
+import { memo, useState, useCallback, useMemo } from 'react';
+
+import { useStringFromTextObject } from '../../hooks/useStringFromTextObject';
+import { useUiKitState } from '../../hooks/useUiKitState';
+import type { BlockProps } from '../../utils/BlockProps';
+import AutocompleteOptions, {
+  OptionsContext,
+} from './UserAutoCompleteMultipleOptions';
+
+type UsersSelectElementProps = BlockProps<UiKit.UsersSelectElement>;
+
+type UserAutoCompleteOptionType = {
+  name: string;
+  username: string;
+  _federated?: boolean;
+};
+
+type UserAutoCompleteOptions = {
+  [k: string]: UserAutoCompleteOptionType;
+};
+
+const matrixRegex = new RegExp('@(.*:.*)');
+
+const UsersSelectElement = ({
+  block,
+  context,
+}: UsersSelectElementProps): ReactElement => {
+  const [{ loading, value, error }, action] = useUiKitState(block, context);
+  const fromTextObjectToString = useStringFromTextObject();
+  const [filter, setFilter] = useState('');
+  const [selectedCache, setSelectedCache] = useState<UserAutoCompleteOptions>(
+    {}
+  );
+
+  const debouncedFilter = useDebouncedValue(filter, 500);
+  const getUsers = useEndpoint('GET', '/v1/users.autocomplete');
+
+  const { data } = useQuery(
+    ['users.autocomplete', debouncedFilter],
+    async () => {
+      const users = await getUsers({
+        selector: JSON.stringify({ term: debouncedFilter }),
+      });
+      const options = users.items.map(
+        (item): [string, UserAutoCompleteOptionType] => [item.username, item]
+      );
+
+      // Add extra option if filter text matches `username:server`
+      // Used to add federated users that do not exist yet
+      if (matrixRegex.test(debouncedFilter)) {
+        options.unshift([
+          debouncedFilter,
+          {
+            name: debouncedFilter,
+            username: debouncedFilter,
+            _federated: true,
+          },
+        ]);
+      }
+
+      return options;
+    },
+    { keepPreviousData: true }
+  );
+
+  const options = useMemo(() => data || [], [data]);
+
+  const onAddUser = useCallback(
+    (username: string): void => {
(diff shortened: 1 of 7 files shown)


gabriellsh (changes requested):
I actually had several other changes to request, but there's a bigger issue that I noticed upon further. Some (if not all) of the components created for the fuselage-ui-kit package, are copies of other existing components that are already in use. Even some comments have been copied over, as well as API calls (some have minor changes).

This feels a little strange to me, as we should not be copying implementation like this. Even more so because there are some implementation flaws on these components (like irrelevant useMemos and some weird logic) that was not modified/removed/improved in any sort of form. Could you explain the reasoning behind this? Is there any way we could improve this by reusing the same components?

Also, are we expecting that federated users will be shown in apps? I think this should be discussed since federated users have some limitations on how they work compared to regular users.

Btw, I know this is already a little too critic, but the name chosen for the components `Multi....Multiple`, doesn't it seem a little... redundant?

tiagoevanp:
In fact, I was copying core implementations, thanks to raise the points where my eyes passed by without notice some of the dumb errors... I will correct them ASAP. Another important thing to mention is: I know that we don't have a perfect autocomplete/select component to use, and we have a lot of different implementations of it in so many parts of the code. I already have created a different task to create a separated package for users/channel selectors who should centralize all... like `ui-selects`, or something else! I received a negative from another teammate because there is more than a simple work on doing this, regarding fuselage work for select components as well 😬. So, my first option to make some progress was the copycat technique, and wait for future improvements on this side.

packages/fuselage-ui-kit/src/elements/UsersSelectElement/MulitUserAutoCompleteMultipleOption.tsx
```diff
@@ -0,0 +1,38 @@
...
+import { UserAvatar } from '@rocket.chat/ui-avatar';
+import type { ReactElement } from 'react';
+
+type UserAutoCompleteMultipleOptionProps = {
+  label: {
+    _federated?: boolean;
+  } & Pick<IUser, 'username' | 'name'>;
+};
+
+const UserAutoCompleteMultipleOption = ({
```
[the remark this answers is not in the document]
tiagoevanp: fixed!

packages/fuselage-ui-kit/src/elements/UsersSelectElement/MulitUserAutoCompleteMultipleOption.tsx
```diff
@@ -0,0 +1,38 @@
...
+      {...props}
+      data-qa-type='autocomplete-user-option'
+      avatar={<UserAvatar username={username || ''} size='x20' />}
+      icon={_federated ? 'globe' : undefined}
+      key={username}
+      label={
+        (
+          <>
+            {name || username}{' '}
+            {!_federated && <OptionDescription>({username})</OptionDescription>}
```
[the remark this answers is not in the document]
tiagoevanp: fixed!

packages/fuselage-ui-kit/src/elements/UsersSelectElement/UsersSelectElement.tsx
```diff
@@ -0,0 +1,64 @@
...
+
+const UsersSelectElement = ({ block, context }: UsersSelectElementProps) => {
+  const [{ loading, value }, action] = useUiKitState(block, context);
+
+  const [filter, setFilter] = useState('');
+  const debouncedFilter = useDebouncedValue(filter, 1000);
+
+  const data = useUsersData({ filter: debouncedFilter });
+
+  const options = useMemo(() => data || [], [data]);
```
gabriellsh, line 26: You can remove this memo by returning `options || []` from the useQuery call itself.

packages/fuselage-ui-kit/src/elements/UsersSelectElement/MultiUsersSelectElement.tsx
```diff
@@ -0,0 +1,80 @@
...
+        <Option data-qa-type='autocomplete-user-option' key={value} {...props}>
+          <OptionAvatar>
+            <UserAvatar username={value} size='x20' />
+          </OptionAvatar>
+          <OptionContent>
+            {label} <OptionDescription>({value})</OptionDescription>
+          </OptionContent>
+        </Option>
+      )}
+      options={data || []}
```
gabriellsh, line 74:
```suggestion
      options={data}
```

packages/fuselage-ui-kit/src/elements/UsersSelectElement/UsersSelectElement.tsx
gabriellsh:
```suggestion
```

packages/fuselage-ui-kit/src/elements/UsersSelectElement/hooks/useUsersData.ts
```diff
@@ -0,0 +1,32 @@
...
+        selector: JSON.stringify({ term: filter }),
+      });
+      const options = users.items.map(
+        (item): UserAutoCompleteOptionType => ({
+          value: item.username,
+          label: item.name || item.username,
+        })
+      );
+
+      return options;
```
gabriellsh, line 26:
```suggestion
      return options || [];
```

gabriellsh (changes requested):

.changeset/cuddly-cycles-nail.md
```diff
@@ -0,0 +1,6 @@
+---
+"@rocket.chat/fuselage-ui-kit": minor
+"@rocket.chat/ui-kit": patch
```
MarcosSpessatto, line 3:
```suggestion
"@rocket.chat/ui-kit": minor
```

gabriellsh (changes requested):
Please try to add unit tests as we discussed in DM ;)

packages/fuselage-ui-kit/src/elements/UsersSelectElement/UserSelectElement.spec.tsx
```diff
@@ -0,0 +1,99 @@
...
+  });
+
+  it('should open the user selector', async () => {
+    const input = await screen.findByRole('textbox');
+    input.focus();
+
+    expect(await screen.findByRole('listbox')).toBeInTheDocument();
+  });
+
+  it('should filter users', async () => {
```
gabriellsh, line 71: Same here

gabriellsh (changes requested):

gabriellsh (approved):

MarcosSpessatto (approved):

PR #31487, opened 2024-01-18
fix: ensure sessionId on Sessions documents

MarcosSpessatto:

[Jira](https://rocketchat-poc.atlassian.net/browse/SM-65)


sampaiodiego (changes requested):
what specific query is this index for? I didn't find one in the model class.. I wonder if this is for an "old" query we don't use anymore.

sampaiodiego (approved):

PR #31497, opened 2024-01-19
chore: don't `ignoreUndefined`

sampaiodiego:

This was set to `true` by default from Meteor due to an inconsistency with `minimongo` but it actually caused a lot of issues for us in the past.. since we don't care too much about this inconsistencies I'm proposing we change the default value. Ref.: https://github.com/meteor/meteor/issues/6051 https://github.com/meteor/meteor/pull/9444 https://www.mongodb.com/docs/drivers/node/current/fundamentals/bson/undefined-values/

## Issue(s)

[ARCH-1447]

## Further comments

[ARCH-1447]: https://rocketchat.atlassian.net/browse/ARCH-1447?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/meteor/tests/unit/app/apps/server/rooms.tests.ts
+++ b/apps/meteor/tests/unit/app/apps/server/rooms.tests.ts
@@ -112,9 +112,9 @@ describe('The AppMessagesConverter instance', () => {
 			expect(rocketchatRoom).to.have.property('_id', appRoom.id);
 			expect(rocketchatRoom).to.have.property('name', appRoom.slugifiedName);
 			expect(rocketchatRoom).to.have.property('sysMes', appRoom.displaySystemMessages);
+			expect(rocketchatRoom).to.have.property('msgs', 0);
 			expect(rocketchatRoom).to.have.property('_updatedAt', appRoom.updatedAt);
 
-			expect(rocketchatRoom).to.not.have.property('msgs');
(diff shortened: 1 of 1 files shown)


sampaiodiego:
🎂

KevLehman (dismissed):
I'll leave the symbolic approve for when this PR is finally opened.

ggazzo (dismissed):

KevLehman (approved):

pierre-lehnen-rc (commented):
Before merging I suggest we double check every query that uses the `$exists` operator to ensure that it's not checking for something that will start existing as undefined.

sampaiodiego:
> Before merging I suggest we double check every query that uses the `$exists` operator to ensure that it's not checking for something that will start existing as undefined.

I agree but I'm afraid I'm not able to do such an extensive check 🙈

KevLehman (approved):

MarcosSpessatto (dismissed):

KevLehman (approved):

MarcosSpessatto (approved):

KevLehman (approved):

ggazzo (approved):

PR #31525, opened 2024-01-24
fix: show only relevant userInfoActions for mentioned non-members

### LINKED ISSUE #31712: Admin can remove user even if user is already removed from the room
### Description:

When an admin removes a user from a room, the option to remove the user persists for the admin, even after the user has been successfully removed. This creates confusion and a potential for unintended actions.

### Steps to reproduce:

1. Log in to Rocket.Chat as an administrator.
2. Navigate to a specific room where multiple users are present.
3. Remove a user from the room.
4. Verify that the user has been successfully removed from the room.
5. Observe that the option to remove the same user is still available for the admin.

### Expected behavior:

The option to remove a user should not be available for the admin after the user has been successfully removed from the room. This ensures clarity and prevents unintentional actions.

### Actual behavior:

The option to remove the user persists for the admin even after successfully removing the user from the room.

### Server Setup Information:

- Version of Rocket.Chat Server: 6.7.0
- Operating System: macOS
- NodeJS Version: 14.21.3
- MongoDB Version: 6.0.3

### Client Setup Information

- Desktop App or Browser Version: chrome 121
- Operating System: macOs

### Pull request corresponds to this issue 
#31525 

### Additional context

https://github.com/RocketChat/Rocket.Chat/assets/125182653/71c4bf94-7e4e-484c-b734-e7488ab57e31

abhinavkrin:

This pull request addresses an issue in the user mention functionality within Rocket.Chat channels. Previously, when a user mentioned another user who was not a member of the channel, the User Info card still displayed options like 'Remove User', 'Set as Leader', 'Set as Moderator', etc., which were not applicable for non-members. This fix introduces a check to determine if the mentioned user is a member of the channel/group. If the user is not a member, it now shows an option to 'Add User' to the channel and omits inappropriate options for non-members.
1. Implemented an API to verify if the mentioned user is a member of the channel/group/dm. 
2. Display 'Add User' option if the mentioned user is not a member of the channel.
3. Hide options like 'Remove User', 'Set as Leader', 'Set as Moderator', etc., for non-members.

https://github.com/RocketChat/Rocket.Chat/assets/15830206/ee84cace-8bf8-4030-b1cf-8bf509e47b4b

## Issue(s)

When a user mentions another user who was not a member of the channel, the User Info card still displays options like 'Remove User', 'Set as Leader', 'Set as Moderator', etc., which are not applicable for non-members
Also
Fixes #31712 
## Steps to test or reproduce

1. Mention a user who is not a member of the channel.
2. Observe the options available in the User Info card.
3. 'Add User' option should be visible while options like 'Remove User', 'Set as Leader', 'Set as Moderator' should not be available for non-members.

## Further comments

[TC-1079](https://rocketchat.atlassian.net/browse/TC-1079)

[TC-1079]: https://rocketchat.atlassian.net/browse/TC-1079?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

#### Justification for New Endpoint `room.isMember`
We considered using `channels.members/groups.members` with filters for membership checks but faced issues:
1. **User Matching**: Regex filters might return similar, not exact, usernames, causing inaccuracies.
2. **Pagination**: If results exceed `LIMIT_PER_PAGE`, the target user might not appear on the first page, requiring extra API calls.
3. **Performance**: Regex searches are slower than direct comparisons.
To ensure precise and efficient user membership verification, we introduced the `room.isMember` endpoint, circumventing the limitations of existing endpoints and ensuring a more reliable and performant solution.


apps/meteor/app/api/server/v1/channels.ts
```diff
@@ -1094,6 +1094,38 @@ API.v1.addRoute(
...
+	{ authRequired: true },
+	{
+		async get() {
+			check(
+				this.queryParams,
+				Match.ObjectIncluding({
+					username: String,
+					roomId: String,
+				}),
+			);
```
debdutdeb, line 1108: use ajv

apps/meteor/app/api/server/v1/groups.ts
```diff
@@ -747,6 +747,38 @@ API.v1.addRoute(
...
+	{ authRequired: true },
+	{
+		async get() {
+			check(
+				this.queryParams,
+				Match.ObjectIncluding({
+					username: String,
+					roomId: String,
+				}),
+			);
```
debdutdeb, line 761: ajv

apps/meteor/app/api/server/v1/im.ts
```diff
@@ -353,6 +353,44 @@ API.v1.addRoute(
...
+	},
+	{
+		async get() {
+			check(
+				this.queryParams,
+				Match.ObjectIncluding({
+					username: String,
+					roomId: String,
+				}),
+			);
```
debdutdeb, line 369: ajv

debdutdeb (changes requested):
could we not use subscription.getOne endpoint instead?

abhinavkrin:
> could we not use subscription.getOne endpoint instead?

Hey @debdutdeb, sorry for delayed reply.
The use-case is different here, I have created endpoints to check if another user is member of a channel/group/member or not.

debdutdeb:
Right. But I don't think you need all these endpoints. Membership is a subscription. So one endpoint should be sufficient, like `subscription.exists?uid=..&rid=...`. Furthermore, nobody can leave a dm, so a case where you need to check for user exists in a dm or not shouldn't come up in this context.

debdutdeb:
https://github.com/RocketChat/Rocket.Chat/blob/ed451251a3abfd64b26c9bdc8d18e9b84754a1f3/apps/meteor/server/services/authorization/canAccessRoom.ts#L55-L62

debdutdeb:
also idk if client caches the memberlist, might be a good idea to utilize if it does.

abhinavkrin:
> Right. But I don't think you need all these endpoints. Membership is a subscription. So one endpoint should be sufficient, like `subscription.exists?uid=..&rid=...`. Furthermore, nobody can leave a dm, so a case where you need to check for user exists in a dm or not shouldn't come up in this context.

Thanks for the input @debdutdeb. I did not know about subscriptions earlier.
Indeed, I created an endpoint `subscriptions.exists?roomId=...&username=....` to check a user's membership for which I am checking whether a subscription exists for them or not. I have applied appropriate validation so that unauthorized users cannot access this endpoint.

abhinavkrin:
> also idk if client caches the memberlist, might be a good idea to utilize if it does.

This was my initial approach, but the member list functionality does a search and receives a paginated result. So there is no guarantee that the required username would be present in the result or not.

apps/meteor/app/api/server/v1/subscriptions.ts
```diff
@@ -103,3 +107,35 @@ API.v1.addRoute(
...
+				Users.findOneByIdOrUsername(userId || username),
+			]);
+			if (!room?._id) {
+				return API.v1.failure('error-room-not-found');
+			}
+			if (!user?._id) {
+				return API.v1.failure('error-user-not-found');
+			}
+
+			if (room.broadcast && !(await hasPermissionAsync(this.userId, 'view-broadcast-member-list', room._id))) {
```
MarcosSpessatto, line 131: I'm not sure if this is really necessary 🤔 @sampaiodiego can you help on this one?

apps/meteor/app/api/server/v1/subscriptions.ts
```diff
@@ -103,3 +107,35 @@ API.v1.addRoute(
 		},
 	},
 );
+
+API.v1.addRoute(
+	'subscriptions.exists',
```
MarcosSpessatto, line 112: I was thinking of moving this to a new `rooms.isMember`, this would make more sense. WDYT?

apps/meteor/app/api/server/v1/subscriptions.ts
```diff
@@ -103,3 +107,35 @@ API.v1.addRoute(
...
+			if (!user?._id) {
+				return API.v1.failure('error-user-not-found');
+			}
+
+			if (room.broadcast && !(await hasPermissionAsync(this.userId, 'view-broadcast-member-list', room._id))) {
+				return API.v1.unauthorized();
+			}
+
+			if ((await canAccessRoomAsync(room, this.user)) || (await canAccessRoomIdAsync(room._id, this.userId))) {
+				return API.v1.success({ exists: !!(await Subscriptions.countByRoomIdAndUserId(room._id, user._id)) });
```
MarcosSpessatto, line 136:
```suggestion
				return API.v1.success({ exists: (await Subscriptions.countByRoomIdAndUserId(room._id, user._id)) > 0 });
```

apps/meteor/app/api/server/v1/subscriptions.ts
```diff
@@ -103,3 +107,35 @@ API.v1.addRoute(
...
+
+API.v1.addRoute(
+	'subscriptions.exists',
+	{
+		authRequired: true,
+		validateParams: isSubscriptionsExistsProps,
+	},
+	{
+		async get() {
+			const { username, roomId, userId, roomName } = this.queryParams;
```
MarcosSpessatto, line 119: I'd say we can accept only roomId and userId, which is good enough and we can keep it simpler. WDYT?

apps/meteor/app/api/server/v1/subscriptions.ts
```diff
@@ -103,3 +107,35 @@ API.v1.addRoute(
 		},
 	},
 );
+
+API.v1.addRoute(
```
MarcosSpessatto, line 111: Can we add API E2E tests for this new endpoint?

MarcosSpessatto (changes requested):

apps/meteor/app/api/server/v1/subscriptions.ts
```diff
@@ -103,3 +107,35 @@ API.v1.addRoute(
...
+			}
+			if (!user?._id) {
+				return API.v1.failure('error-user-not-found');
+			}
+
+			if (room.broadcast && !(await hasPermissionAsync(this.userId, 'view-broadcast-member-list', room._id))) {
+				return API.v1.unauthorized();
+			}
+
+			if ((await canAccessRoomAsync(room, this.user)) || (await canAccessRoomIdAsync(room._id, this.userId))) {
```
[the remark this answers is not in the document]
debdutdeb: Don't you think using canAccessRoom here mayb be overdoing it? I think just checking for membership should be enough through subscriptions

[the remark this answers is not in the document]
gabriellsh: There are some cases that the user can be in a room but does not the `view-p-room` or `view-c-room` permission.

apps/meteor/client/views/room/hooks/useUserInfoActions/useUserInfoActions.ts
```diff
@@ -40,6 +42,14 @@ export const useUserInfoActions = (
 	reload?: () => void,
 	size = 2,
 ): { actions: [string, UserInfoAction][]; menuActions: any | undefined } => {
+	const { data, refetch } = useMemberExists(useMemo(() => ({ rid, username: user.username || '' }), [rid, user.username]));
```
gabriellsh, line 45: no need to use `useMemo` here.

apps/meteor/client/views/hooks/useMemberExists.ts
```diff
@@ -0,0 +1,16 @@
+import { useEndpoint } from '@rocket.chat/ui-contexts';
+import { useQuery } from '@tanstack/react-query';
+
+type MembersExistsOptions = {
+	rid: string;
+	username: string;
+	refresh?: boolean;
+};
+
+export const useMemberExists = (options: MembersExistsOptions) => {
```
gabriellsh, line 10:
```suggestion
export const useMemberExists = (rid: string, username: string) => {
```

apps/meteor/client/views/room/hooks/useUserInfoActions/actions/useAddUserAction.tsx
```diff
@@ -0,0 +1,89 @@
...
+import { getRoomDirectives } from '../../../lib/getRoomDirectives';
+import type { UserInfoAction } from '../useUserInfoActions';
+
+export const useAddUserAction = (
+	user: Pick<IUser, '_id' | 'username'>,
+	rid: IRoom['_id'],
+	reload?: () => void,
+): UserInfoAction | undefined => {
+	const t = useTranslation();
+	const room = useUserRoom(rid);
```
gabriellsh, line 26: Use `useRoom` instead.

apps/meteor/client/views/room/hooks/useUserInfoActions/useUserInfoActions.ts
```diff
@@ -40,6 +42,14 @@ export const useUserInfoActions = (
...
+	const memberChangeReload = useCallback(async () => {
+		await reload?.();
+		await refetch();
+	}, [reload, refetch]);
+	const isMember = !!useMemo(() => data?.exists, [data?.exists]);
```
gabriellsh, line 50: Also no need to use useMemo here.

apps/meteor/client/views/room/hooks/useUserInfoActions/actions/useAddUserAction.tsx
```diff
@@ -0,0 +1,89 @@
...
+	);
+
+	const userCanAdd =
+		room && user && isRoomFederated(room)
+			? Federation.isEditableByTheUser(currentUser || undefined, room, subscription)
+			: hasPermissionToAddUsers;
+
+	const { roomCanInvite } = getRoomDirectives({ room, showingUserId: uid, userSubscription: subscription });
+
+	const addUsersToRooms = useMethod('addUsersToRoom');
```
gabriellsh, line 49: You should use an api endpoint for this instead of method.

apps/meteor/client/views/hooks/useMemberExists.ts
```diff
@@ -0,0 +1,16 @@
...
+	const checkMember = useEndpoint('GET', '/v1/subscriptions.exists');
+
+	return useQuery(['roomMembershipCheck', options.rid, options.username], () =>
```
gabriellsh, line 13:
```suggestion
	return useQuery(['subscriptions/exists', options.rid, options.username], () =>
```

gabriellsh (changes requested):

apps/meteor/app/api/server/v1/subscriptions.ts
abhinavkrin: Yes this makes sense.

apps/meteor/app/api/server/v1/subscriptions.ts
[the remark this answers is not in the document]
debdutdeb: I missed the check is for the caller, and not the user that is being checked. Ignore my comment pls :)

apps/meteor/app/api/server/v1/subscriptions.ts
abhinavkrin: I thought this endpoint could also be used by community members for their own use case. So I gave some more parameters like in some other endpoints.

apps/meteor/client/views/room/hooks/useUserInfoActions/actions/useAddUserAction.tsx
abhinavkrin: useRoom
It seems like the UserCard is not rendered inside of Room context.
useUserRoom eliminates this error.

apps/meteor/app/api/server/v1/subscriptions.ts
abhinavkrin: Hey @MarcosSpessatto I used this check because this same check was implemented in channels.members for member list. I thought if this check is done to retrieve member list, then it should also be implemented while checking if a user is a member of a room or not.
https://github.com/RocketChat/Rocket.Chat/blob/ab077230dac052e753c456054dc9829d4c93a8bc/apps/meteor/app/api/server/v1/channels.ts#L1060

apps/meteor/app/api/server/v1/subscriptions.ts
abhinavkrin: Hey @MarcosSpessatto I used this check because this same check was implemented in channels.members for member list. I thought if this check is done to retrieve member list, then it should also be implemented while checking if a user is a member of a room or not.
https://github.com/RocketChat/Rocket.Chat/blob/ab077230dac052e753c456054dc9829d4c93a8bc/apps/meteor/app/api/server/v1/channels.ts#L1060

apps/meteor/app/api/server/v1/subscriptions.ts
MarcosSpessatto: But the community users can still use the endpoint, I think the idea is just to keep it simpler, like avoid to validate if we have one parameter or another, but not both at the same time.

apps/meteor/app/api/server/v1/rooms.ts
```diff
@@ -635,3 +635,37 @@ API.v1.addRoute(
...
+
+API.v1.addRoute(
+	'rooms.isMember',
+	{
+		authRequired: true,
+		validateParams: isRoomsIsMemberProps,
+	},
+	{
+		async get() {
+			const { username, roomId, userId, roomName } = this.queryParams;
```
MarcosSpessatto, line 647: Even if we are doing this already in other endpoints, let's try to keep this simple and use only `userId` and `roomId` it will simplify even the tests we need to do. Also, we should use the `findRoomByIdOrName` function here as it already validates whether the room is archived.

Besides that, since we accepting DM rooms here as well, this could be a problem to search for room names as DMs don't have a name.

packages/rest-typings/src/v1/rooms.ts
```diff
@@ -423,6 +423,27 @@ const GETRoomsNameExistsSchema = {
...
+		roomName: { type: 'string' },
+		userId: { type: 'string' },
+		username: { type: 'string' },
+	},
+	oneOf: [
+		{ required: ['roomId', 'userId'] },
+		{ required: ['roomName', 'userId'] },
+		{ required: ['roomId', 'username'] },
+		{ required: ['roomName', 'username'] },
+	],
```
MarcosSpessatto, line 441: Let's try to simplify this and use only `userId` and `roomId`

apps/meteor/tests/end-to-end/api/09-rooms.js
```diff
@@ -1707,4 +1707,230 @@ describe('[Rooms]', function () {
...
+			}).end((err, res) => {
+				testDM = res.body.room;
+				done();
+			});
+		});
+		after(async () => {
+			await Promise.all([
+				closeRoom({ type: 'c', roomId: testChannel._id }),
+				closeRoom({ type: 'p', roomId: testGroup._id }),
+				closeRoom({ type: 'd', roomId: testDM._id }),
```
MarcosSpessatto, line 1769: Let's delete the room instead of just closing it.

apps/meteor/tests/end-to-end/api/09-rooms.js
```diff
@@ -1707,4 +1707,230 @@ describe('[Rooms]', function () {
...
+				type: 'd',
+				username: testUser2.username,
+				credentials: testUser1Credentials,
+			}).end((err, res) => {
+				testDM = res.body.room;
+				done();
+			});
+		});
+		after(async () => {
+			await Promise.all([
```
MarcosSpessatto, line 1766: We don't need this await, we can just return the Promise.all directly.

apps/meteor/tests/end-to-end/api/09-rooms.js
```diff
@@ -1707,4 +1707,230 @@ describe('[Rooms]', function () {
...
+					roomId: testChannel._id,
+					userId: fakeUserId,
+				})
+				.expect(400)
+				.expect((res) => {
+					expect(res.body).to.have.property('success', false);
+					expect(res.body).to.have.property('error', 'error-user-not-found');
+				})
+				.end(done);
+		});
```
MarcosSpessatto, line 1806: That's why I requested to remove `roomName` and `username` from the accepted parameters, otherwise, we need to include those test cases here as well, since we are only validating for invalid `roomId` and `userId` here.

apps/meteor/tests/end-to-end/api/09-rooms.js
```diff
@@ -1707,4 +1707,230 @@ describe('[Rooms]', function () {
...
+
```
MarcosSpessatto, line 1807: Can we add a test case to ensure the broadcast validation as well?

MarcosSpessatto (changes requested):

apps/meteor/tests/end-to-end/api/09-rooms.js
```diff
@@ -1707,4 +1707,297 @@ describe('[Rooms]', function () {
...
+		it('create a direct message room', (done) => {
+			createRoom({
+				type: 'd',
+				username: testUser2.username,
+				credentials: testUser1Credentials,
+			}).end((err, res) => {
+				testDM = res.body.room;
+				done();
+			});
+		});
```
MarcosSpessatto, line 1764: Since we don't have any assertion here and it's just a setup step, I think this would fit better by using a `before` instead of `it`, also, we should use `await` here instead of `done()` as you did in the user's creation.

MarcosSpessatto (changes requested):

MarcosSpessatto (dismissed):

gabriellsh (approved):

apps/meteor/packages/rocketchat-i18n/i18n/en.i18n.json
```diff
@@ -6296,4 +6297,4 @@
   "Seat_limit_reached": "Seat limit reached",
   "Seat_limit_reached_Description": "Your workspace reached its contractual seat limit. Buy more seats to add more users.",
   "Buy_more_seats": "Buy more seats"
-}
\ No newline at end of file
+}
```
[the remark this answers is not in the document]
abhinavkrin: Yeah, this is auto written by vs code. I need to figure out how to disable it.

apps/meteor/packages/rocketchat-i18n/i18n/en.i18n.json
```diff
@@ -6293,4 +6294,4 @@
...
```
abhinavkrin, line 6297:
```suggestion
}```

File: apps/meteor/tests/end-to-end/api/09-rooms.js [hunk 25]
```diff
@@ -1707,4 +1707,294 @@ describe('[Rooms]', function () {
...
+				})
+				.expect(400)
+				.expect((res) => {
+					expect(res.body).to.have.property('success', false);
+					expect(res.body).to.have.property(
+						'error',
+						'The required "roomId" or "roomName" param provided does not match any channel [error-room-not-found]',
+					);
+				})
+				.end(done);
```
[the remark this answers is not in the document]
matheusbsilva137: I think you're now just missing the return statement for every  request, @abhinavkrin

apps/meteor/tests/end-to-end/api/09-rooms.js
```diff
@@ -1707,4 +1707,280 @@ describe('[Rooms]', function () {
 				.end(done);
 		});
 	});
+	describe('/rooms.isMember', async () => {
```
matheusbsilva137, line 1710:
```suggestion
	describe('/rooms.isMember', () => {
```
I believe you don't need this async here

apps/meteor/tests/end-to-end/api/09-rooms.js
```diff
@@ -1707,4 +1707,280 @@ describe('[Rooms]', function () {
...
+		const testGroupName = `group.test.${Date.now()}-${Math.random()}`;
+
+		let testUser1;
+		let testUser2;
+		let testUserNonMember;
+		let testUser1Credentials;
+		let testUserNonMemberCredentials;
+
+		before(async () => {
+			[testUser1, testUser2, testUserNonMember] = await Promise.all([createUser(), createUser(), createUser()]);
```
matheusbsilva137, line 1728: I'm still not so sure of why these tests are failing (and they're also flaky)
Maybe try not using `Promise.all` here specifically. I'm wondering if maybe their usernames are coliding due to the timestamp (since createUser uses `Date.now()` to define each username)

matheusbsilva137 (changes requested):

apps/meteor/tests/end-to-end/api/09-rooms.js
abhinavkrin: Yes, It might be due to Date.now() in usernames as I was using promise.all for creating users. I have updated the code to create user sequentially

MarcosSpessatto (dismissed):

matheusbsilva137 (changes requested):
I was just about to approve it, but I'm thinking that maybe a better approach would be to take the chance of our next major to tweak the `subscriptions.getOne` endpoint instead of adding a new one (`rooms.isMember`). This would be a breaking change.

My suggestion is to discard the new endpoint and improve `subscriptions.getOne` instead to return an `API.v1.failure('subscription-not-found')` when no matching subscription is returned. Currently, our `subscriptions.getOne` returns `{"subscription":null,"success":true}` for not found subscriptions, which doesn't look like the best design/approach to me.

Sorry for only suggesting this now :grimacing: at least I think you can port your new tests to this endpoint (and you'll also need to fix some existing ones that expect this `null` value). Please let me know what you think

cc @MarcosSpessatto

matheusbsilva137 (dismissed):
I'll approve this PR since I think this new endpoint does make sense. Let's work on the improvements for `subscriptions.getOne` on another PR aimed to the 7.0 release

apps/meteor/client/views/room/hooks/useUserInfoActions/useUserInfoActions.ts
```diff
@@ -40,6 +42,14 @@ export const useUserInfoActions = (
 	reload?: () => void,
 	size = 2,
 ): { actions: [string, UserInfoAction][]; menuActions: any | undefined } => {
+	const { data, refetch, isSuccess: membershipCheckSuccess } = useMemberExists(rid, user.username as string);
```
yash-rajpal, line 45: This `useUserInfoActions` hook can be called from 2 places - 

- Room Members Contextual Bar
- User Card on Mentions and stuff.

If it is being called from Room Members Contextual Bar, it will be called for each and every user in the list. It will inturn call the `rooms.isMember` endpoint for each and every user in the list, while we already know that if the user is present in room members contextual bar, user is already part of the room. In this case we are unnecessarily calling the `rooms.isMember` endpoint for each user.

This flow only makes sense when we call this hook from UserCard or some other place, where we are not user about user's subscription to the room.

I would suggest to move up this check to outside this hook, to it's parent and this hook should receive this info as a prop, whether user is part of room or not, based on this prop, we can return the list of user actions.

abhinavkrin: Hey @yash-rajpal  Thanks for noticing this. I agree we should not call the endpoint when we already know if the user is a member.
To fix this I have added an override prop for the useMemberExists hook which will prevent api call when an override value is provided. And when calling this hook in memberlist, i am passing the override value and hence preventing unnecessary api calls.

apps/meteor/client/views/hooks/useMemberExists.ts
```diff
@@ -0,0 +1,10 @@
+import { useEndpoint } from '@rocket.chat/ui-contexts';
+import { useQuery } from '@tanstack/react-query';
+
+export const useMemberExists = (roomId: string, username: string, override?: boolean) => {
```
MarcosSpessatto, line 4: I think we should use an object for parameters here, so we can use named parameters instead of providing booleans to the function, it makes it harder to understand. What about a `{ ..., isMember: boolean }`?

MarcosSpessatto (changes requested):

apps/meteor/client/views/hooks/useMemberExists.ts
abhinavkrin: Yes. It makes sense.

abhinavkrin: @MarcosSpessatto Should I also change parameters for useUserInfoActions to an object? It will make the code more understandable.

apps/meteor/client/views/hooks/useMemberExists.ts
```diff
@@ -0,0 +1,12 @@
+import { useEndpoint } from '@rocket.chat/ui-contexts';
+import { useQuery } from '@tanstack/react-query';
+
+type UseMemberExistsProps = { roomId: string; username: string; isMember?: boolean };
+
+export const useMemberExists = ({ roomId, username, isMember }: UseMemberExistsProps) => {
```
yash-rajpal, line 6: While this would work, but imo this doesn't follow best design principles and best practices.

We are sending the info whether a user is part of a room or not to a hook which is supposed to give us this info.

Why don't we move this hook call to component above in the tree, like `UserCardWithData` where it would return us this info which can be used when calling the `userInfoAction` hook. 🤔

yash-rajpal (changes requested):

yash-rajpal (approved):
LGTM!

apps/meteor/client/views/room/UserCard/UserCardWithData.tsx
```diff
@@ -25,6 +26,9 @@ const UserCardWithData = ({ username, rid, onOpenUserInfo, onClose }: UserCardWi
 	const showRealNames = Boolean(useSetting('UI_Use_Real_Name'));
 
 	const { data, isLoading } = useUserInfoQuery({ username });
+	const { data: isMemberData, refetch, isSuccess: membershipCheckSuccess } = useMemberExists({ roomId: rid, username });
+
+	const isMember = (isMemberData?.exists as boolean) && membershipCheckSuccess;
```
MarcosSpessatto, line 31: We should not just ignore the errors or the error-handling mechanism provided by useQuery. We must handle cases like when this is still loading, or if an error happens. Otherwise, this can create even flaky test cases because while the component is still loading we are showing some options, when the request is done we show other ones instead, making the screen flicker.

MarcosSpessatto (changes requested):
Also, let's create some unit tests for the front-end side. We already covered all the BE behavior, but let's also guarantee the FE side.

apps/meteor/client/views/room/UserCard/UserCardWithData.tsx
```diff
@@ -24,7 +25,16 @@ const UserCardWithData = ({ username, rid, onOpenUserInfo, onClose }: UserCardWi
...
+	const { data, isLoading: isUserInfoLoading } = useUserInfoQuery({ username });
+	const {
+		data: isMemberData,
+		refetch,
+		isSuccess: membershipCheckSuccess,
+		isLoading: isMembershipStatusLoading,
+	} = useMemberExists({ roomId: rid, username });
+
+	const isLoading = isUserInfoLoading || isMembershipStatusLoading;
+	const isMember = (isMemberData?.exists as boolean) && membershipCheckSuccess;
```
MarcosSpessatto, line 37: If you first check for success I believe you don't need to cast this. @gabriellsh can you help us here?

MarcosSpessatto (changes requested):

apps/meteor/app/api/server/v1/rooms.ts
```diff
@@ -670,6 +676,36 @@ API.v1.addRoute(
...
+					params: { roomId },
+				}) as Promise<IRoom>,
+				Users.findOneByIdOrUsername(userId || username),
+			]);
+
+			if (!user?._id) {
+				return API.v1.failure('error-user-not-found');
+			}
+
+			if (await canAccessRoomAsync(room, this.user)) {
```
MarcosSpessatto, line 699:
```suggestion
			if (await canAccessRoomAsync(room, { _id: this.userId })) {
```

apps/meteor/tests/e2e/user-card-info-in-room.spec.ts
```diff
@@ -0,0 +1,56 @@
...
+		const mentionSpan = page.locator(`span[title="Mentions user"][data-uid="${Users.user1.data.username}"]`);
+		await mentionSpan.click();
+
+		await expect(page.locator('button[title="Add to room"]')).not.toBeVisible();
+
+		await page.locator('div[aria-label="User card actions"] button[title="More"]').click();
+
+		await expect(page.locator('label[data-key="Remove from room"]')).toBeVisible();
+		await expect(page.locator('label[data-key="Set as leader"]')).toBeVisible();
+		await expect(page.locator('label[data-key="Set as moderator"]')).toBeVisible();
```
MarcosSpessatto, line 40: For this to be true, the user must be the room owner (or at least they must have the necessary permission), right? 🤔  I belive being a member only doesn't mean you can add leaders, removing people from the room, etc...

If so, we need to put this in the test title `should show correct userinfo actions for the room owner`, and create another test cases for when the user is not the owner/don't have permissions for removing from room, set as leader, etc.

MarcosSpessatto (changes requested):

apps/meteor/tests/e2e/user-card-info-in-room.spec.ts
abhinavkrin: Yes, you are right :)
Moving forward I have added tests for both when viewed by room owner and when viewed by normal member.
I have also split the test test files into two  in order to prevent the timeout issue that results in flaky tests.

Also, the action button can be visible in the card or it could be hidden under the more button. So, in order to prevent flaky tests that might arise in future, I have added check for both the cases.

MarcosSpessatto (dismissed):

matheusbsilva137 (changes requested):
Could you please fix the conflicts too?

apps/meteor/tests/end-to-end/api/09-rooms.ts
```diff
@@ -2859,4 +2859,286 @@ describe('[Rooms]', () => {
...
+				});
+		});
+
+		it('should return success with isMember=true if given username is a member of the channel', () => {
+			return request
+				.get(api('rooms.isMember'))
+				.set(testUser1Credentials)
+				.query({
+					roomId: testChannel._id,
+					userId: testUser2.username,
```
MarcosSpessatto, line 3000:
```suggestion
					username: testUser2.username,
```

apps/meteor/tests/end-to-end/api/09-rooms.ts
```diff
@@ -2859,4 +2859,286 @@ describe('[Rooms]', () => {
...
+				});
+		});
+
+		it('should return success with isMember=true if given username is a member of the group', () => {
+			return request
+				.get(api('rooms.isMember'))
+				.set(testUser1Credentials)
+				.query({
+					roomId: testGroup._id,
+					userId: testUser2.username,
```
MarcosSpessatto, line 3045:
```suggestion
					username: testUser2.username,
```

apps/meteor/tests/end-to-end/api/09-rooms.ts
```diff
@@ -2859,4 +2859,286 @@ describe('[Rooms]', () => {
...
+				});
+		});
+
+		it('should return success with isMember=true if given username is a member of the DM', () => {
+			return request
+				.get(api('rooms.isMember'))
+				.set(testUser1Credentials)
+				.query({
+					roomId: testDM._id,
+					userId: testUser2.username,
```
MarcosSpessatto, line 3105:
```suggestion
					username: testUser2.username,
```

MarcosSpessatto (changes requested):

MarcosSpessatto (approved):

matheusbsilva137 (approved):

MarcosSpessatto (approved):

KevLehman (approved):

apps/meteor/client/views/room/hooks/useUserInfoActions/actions/useAddUserAction.tsx
```diff
@@ -0,0 +1,95 @@
+import type { IRoom, IUser } from '@rocket.chat/core-typings';
```
ggazzo, line 1: just keep in mind this is not a fix, you are adding an extra feature here

ggazzo (approved):

PR #31527, opened 2024-01-24
chore: use single message event

sampaiodiego:

- removed the event `message.sent` in favor of the old one `watch.messages` that is now used for either when a message comes from Db or emitted manually
- removed the function `listenToMessageSentEvent` since we now have a single event to listen, there is no need to decide which one to listen to
- removed the callback `broadcastCallback` from the `broadcastMessage` function since it can now broadcast to a single event


KevLehman (dismissed):

apps/meteor/server/modules/watchers/lib/messages.ts
```diff
@@ -12,41 +13,48 @@ const getUserNameCached = mem(
...
+	}
+
+	return message;
+}
 
-		void broadcastCallback(message);
+// TODO once the broadcast from file apps/meteor/server/modules/watchers/watchers.module.ts is removed
+// this function can be renamed to broadcastMessage
+export async function broadcastMessageFromData({ id, data }: { id: IMessage['_id']; data?: IMessage }): Promise<void> {
+	console.log('broadcastMessageFromData', dbWatchersDisabled);
```
pierre-lehnen-rc, line 51:
```suggestion
```

pierre-lehnen-rc (changes requested):

apps/meteor/server/modules/watchers/lib/messages.ts
```diff
@@ -12,41 +13,48 @@ const getUserNameCached = mem(
...
+	if (!dbWatchersDisabled) {
+		return;
+	}
```
pierre-lehnen-rc, line 54: If db watchers is disabled, it won't trigger any event anymore, but the `message.sent` event worked without the watchers.

pierre-lehnen-rc (changes requested):

apps/meteor/server/modules/watchers/lib/messages.ts
pierre-lehnen-rc, line 54:
```suggestion
	// if db watchers are active, the event will be triggered automatically so we don't need to broadcast it here.
	if (!dbWatchersDisabled) {
		return;
	}
```

Adding a comment here to just to clarify the difference

ggazzo (approved):

PR #31528, opened 2024-01-24
fix: Coerce values on appearance endpoint from setting data types

KevLehman:


apps/meteor/app/livechat/imports/server/rest/appearance.ts
```diff
@@ -52,8 +52,39 @@ API.v1.addRoute(
 				throw new Error('invalid-setting');
 			}
 
+			const dbSettings = await Settings.find({ _id: { $in: validSettingList } }, { projection: { _id: 1, value: 1, type: 1 } }).toArray();
+
+			const settingsToSave: typeof this.bodyParams = [];
+			dbSettings.forEach((dbSetting) => {
```
MarcosSpessatto, line 58: I believe we could use the map loop here instead of creating a new one 👀

apps/meteor/app/livechat/imports/server/rest/appearance.ts
```diff
@@ -52,8 +52,39 @@ API.v1.addRoute(
...
+				const setting = settings.find(({ _id }) => _id === dbSetting._id);
+				if (!setting || dbSetting.value === setting.value) {
+					return;
+				}
+
+				switch (dbSetting?.type) {
+					case 'boolean':
+						settingsToSave.push({
+							_id: dbSetting._id,
+							value: coerceValue(setting?.value, 'boolean'),
```
MarcosSpessatto, line 68: At this point, we already know the data type, Is there any reason to create the function to coerce multiple data types, instead of comparing it here directly? e.g `(value === 'true' || value === true)`

MarcosSpessatto (changes requested):

apps/meteor/app/livechat/imports/server/rest/appearance.ts
KevLehman: You know, I always forget maps exist

apps/meteor/app/livechat/imports/server/rest/appearance.ts
KevLehman: idk, I just liked the function. Initially this copied the multiple data types the settings can have, but then I realized i only needed string|bool|int

apps/meteor/app/livechat/imports/server/rest/appearance.ts
```diff
@@ -52,8 +53,37 @@ API.v1.addRoute(
 				throw new Error('invalid-setting');
 			}
 
+			const dbSettings = await Settings.find({ _id: { $in: validSettingList } }, { projection: { _id: 1, value: 1, type: 1 } }).toArray();
```
pierre-lehnen-rc, line 56: Avoid making queries from outside of the model, instead create a new function in the model that runs the query you need.

pierre-lehnen-rc (commented):
Keep in mind `parseInt` will turn invalid values into `NaN` - I guess you probably want a fallback for that.

ggazzo (changes requested):

apps/meteor/app/livechat/imports/server/rest/appearance.ts
```diff
@@ -62,3 +90,20 @@ API.v1.addRoute(
...
 	},
 );
+
+function coerceInt(value: string | number | boolean): number {
+	if (typeof value === 'number') {
+		return value;
+	}
+
+	if (typeof value === 'boolean') {
+		return 0;
```
MarcosSpessatto, line 100: Are we okay with assuming 0 when an invalid value is provided as a number? If so, can we add some test cases to cover this and the `NaN` cases?

MarcosSpessatto (changes requested):

MarcosSpessatto (approved):

lmauromb (approved):

sampaiodiego (approved):

PR #31530, opened 2024-01-24
chore: Remove endpoints dependency on meteor/check

matheusbsilva137:
## Further comments

TC-1090


apps/meteor/app/api/server/v1/permissions.ts
```diff
@@ -40,13 +39,9 @@ API.v1.addRoute(
 
 API.v1.addRoute(
 	'permissions.update',
-	{ authRequired: true },
+	{ authRequired: true, permissionsRequired: ['access-permissions'] },
```
ggazzo, line 42: not related please move to a different pr

apps/meteor/app/api/server/v1/channels.ts
```diff
@@ -947,7 +944,12 @@ API.v1.addRoute(
 
 API.v1.addRoute(
 	'channels.list',
-	{ authRequired: true },
+	{
+		authRequired: true,
+		permissionsRequired: {
+			GET: { permissions: ['view-c-room', 'view-joined-room'], operation: 'hasAny' },
```
ggazzo, line 950: same

ggazzo (changes requested):
since this is supposed to be huge because there is no other way to go, avoid different changes, stick with meteor/check removal, if you can split into commits/pr each endpoint would be great also

matheusbsilva137:
Closed in favor of #31579 and #31576
Thanks for the reviews, @ggazzo ! Please check the new PRs later

PR #31538, opened 2024-01-25
feat: introduce new methods to fetch rooms on Livechat Bridge

lmauromb:

- Add new method on the Livechat Bridge to fetch the current numbers of open chats by a provided AgentID
- After related: https://github.com/RocketChat/Rocket.Chat.Apps-engine/pull/710 is merged we would need to bump the apps.engine version package.

## Issue(s)

- https://rocketchat.atlassian.net/browse/PROSVC-55


apps/meteor/app/apps/server/bridges/livechat.ts
```diff
@@ -135,6 +135,18 @@ export class AppLivechatBridge extends LivechatBridge {
 		return true;
 	}
 
+	protected async findRoomsByAgentId(agentId: string, appId: string): Promise<ILivechatRoom[]> {
+		this.orch.debugLog(`The App ${appId} is looking for livechat rooms. for agent ${agentId}`);
```
KevLehman, line 139:
```suggestion
		this.orch.debugLog(`The App ${appId} is looking for livechat rooms associated with agent ${agentId}`);
```

apps/meteor/app/apps/server/bridges/livechat.ts
```diff
@@ -135,6 +135,18 @@ export class AppLivechatBridge extends LivechatBridge {
 		return true;
 	}
 
+	protected async findRoomsByAgentId(agentId: string, appId: string): Promise<ILivechatRoom[]> {
```
KevLehman, line 138: Since this is a find, can you create also a function that returns just the count? In case we need it

KevLehman (changes requested):

apps/meteor/app/apps/server/bridges/livechat.ts
lmauromb: performance wise, we prefer to do this on the DB instead of just using `length` right?

KevLehman: Yup, less data to send to the app and to keep in mem

apps/meteor/server/models/raw/LivechatRooms.ts
```diff
@@ -2264,6 +2264,17 @@ export class LivechatRoomsRaw extends BaseRaw<IOmnichannelRoom> implements ILive
 		);
 	}
 
+	countOpenByAgent(userId: string, extraQuery: Filter<IOmnichannelRoom> = {}) {
```
KevLehman, line 2267: extraQuery is optional on model but required here 🤔 

(Can you update `findOpenByAgent` too? :please: )

KevLehman (approved):

apps/meteor/app/apps/server/bridges/livechat.ts
```diff
@@ -135,6 +135,28 @@ export class AppLivechatBridge extends LivechatBridge {
...
+
+		const result = await LivechatRooms.findOpenByAgent(agentId).toArray();
+
+		return Promise.all((result as unknown as ILivechatRoom[]).map((room) => this.orch.getConverters()?.get('rooms').convertRoom(room)));
+	}
+
+	protected async countRoomsByAgentId(agentId: string, appId: string): Promise<number> {
+		this.orch.debugLog(`The App ${appId} is counting livechat rooms associated with agent ${agentId}`);
+
+		if (!agentId) {
```
MarcosSpessatto, line 153: Wouldn't be better to throw an error if the agent id was not provided?

apps/meteor/app/apps/server/bridges/livechat.ts
```diff
@@ -135,6 +135,28 @@ export class AppLivechatBridge extends LivechatBridge {
 		return true;
 	}
 
+	protected async findRoomsByAgentId(agentId: string, appId: string): Promise<ILivechatRoom[]> {
+		this.orch.debugLog(`The App ${appId} is looking for livechat rooms associated with agent ${agentId}`);
+
+		if (!agentId) {
```
MarcosSpessatto, line 141: Wouldn't be better to throw an error if the agent id was not provided?

apps/meteor/app/apps/server/bridges/livechat.ts
```diff
@@ -135,6 +135,28 @@ export class AppLivechatBridge extends LivechatBridge {
...
+			return [];
+		}
+
+		const result = await LivechatRooms.findOpenByAgent(agentId).toArray();
```
MarcosSpessatto, line 145: We could use `.map` here directly instead of using `.toArray()`. Wdyt?

MarcosSpessatto (changes requested):

apps/meteor/app/apps/server/bridges/livechat.ts
lmauromb: Is it better to do the `map` instead of `toArray` ? I'm not familiar with the tradeoffs of each option.

.changeset/tiny-toes-fail.md
```diff
@@ -0,0 +1,5 @@
+---
+'@rocket.chat/meteor': patch
+---
+
+Added new methods `findRoomsByAgentId` and `countRoomsByAgentId` to livechat bridge
```
d-gubert, line 5:
```suggestion
Introduced new methods for Rocket.Chat Apps to interact with livechat: `findRoomsByAgentId` and `countRoomsByAgentId`
```

Just a more "user friendly" message, mentioning more explicitly what feature is impacted

d-gubert:
Just a note for the future: whenever updating the Apps-Engine, you can run a `yarn up @rocket.chat/apps-engine@x.y.z` at the root of the project - this will update the dependency in all packages inside the monorepo to the same version.

Of course, this also works for any other package that is used throughout the monorepo :)

d-gubert (dismissed):

MarcosSpessatto (dismissed):

ggazzo (dismissed):
please rewrite the title

KevLehman (approved):

MarcosSpessatto (approved):

.changeset/tiny-toes-fail.md
```diff
@@ -0,0 +1,5 @@
+---
+'@rocket.chat/meteor': patch
```
MarcosSpessatto, line 2:
```suggestion
'@rocket.chat/meteor': minor
```

MarcosSpessatto (approved):

KevLehman (approved):

apps/meteor/app/apps/server/bridges/livechat.ts
```diff
@@ -138,6 +138,27 @@ export class AppLivechatBridge extends LivechatBridge {
 		return true;
 	}
 
+	protected async findRoomsByAgentId(agentId: string, appId: string): Promise<ILivechatRoom[]> {
```
d-gubert, line 141: I've just noticed that the implementation includes some filtering that is not clear by the method's name, i.e. `findRoomsByAgentId` does not convey that the method will filter by `{open: true}` - which the method `findOpenByAgent` makes more clear.

So my suggestion here would be to rename the methods in the Apps-Engine to something more similar to `findOpenRoomsByAgentId` and add some comments to the methods explaining what filters it will apply, so it is clearer from the Apps-Engine perspective.

d-gubert (changes requested):

apps/meteor/app/apps/server/bridges/livechat.ts
lmauromb: @d-gubert can you help me with the release for that change: https://github.com/RocketChat/Rocket.Chat.Apps-engine/pull/735

lmauromb: ready!

d-gubert (approved):

MarcosSpessatto (approved):

KevLehman (approved):

ggazzo (approved):

PR #31539, opened 2024-01-25
refactor: Delay queue-worker start until first work is received

KevLehman:
## Further comments

Also changes the queue-worker processing from 1 to 5 secs


ee/packages/omnichannel-services/src/OmnichannelTranscript.ts
```diff
@@ -122,6 +125,11 @@ export class OmnichannelTranscript extends ServiceClass implements IOmnichannelT
 			return;
 		}
 
+		if (!(await queueService.isQueueStarted())) {
```
sampaiodiego, line 128: should this logic be inside of `queueService`? what happens if `queueService.queueWork` is called from somewhere else without calling `registerWorks`? 😬

KevLehman: hmmmm, good question

ee/packages/omnichannel-services/src/QueueWorker.ts
```diff
@@ -117,6 +120,10 @@ export class QueueWorker extends ServiceClass implements IQueueWorkerService {
 	// This is a "generic" job that allows you to call any service
 	async queueWork<T extends Record<string, unknown>>(queue: Actions, to: string, data: T): Promise<void> {
 		this.logger.info(`Queueing work for ${to}`);
+		if (!this.queueStarted) {
+			this.registerWorkers();
```
sampaiodiego, line 124: although `registerWorkers` is `async` it doesn't have any `await` in it, but just in case, should we `await` here?

KevLehman: No. It was async cause at the beginning, I created the indexes here. Then removed the async and made it return a promise.

I saw it today returning a promise and made it async...

Cycle of life

ee/packages/omnichannel-services/src/QueueWorker.ts
```diff
@@ -17,12 +17,15 @@ export class QueueWorker extends ServiceClass implements IQueueWorkerService {
 
 	private logger: Logger;
 
+	public queueStarted = false;
```
KevLehman, line 20:
```suggestion
	private queueStarted = false;
```

PR #31552, opened 2024-01-26
chore: include thread messages when sorting by activity

heitortanoue:
Threads are now being considered in the ordering. Modified the `notifyUsersOnMessage` to update the room's `lastMessage` and `lm` fields when sending messages.

## Steps to test or reproduce

- On the sidebar, click on the Display icon and set it as:- Display: Extended- Sort by: Activity- Group by: Unread, favorites, and Types

- Open DM’s with different users

- Send and receive messages in the DM

- Send messages inside threads in the chats

## Further comments

TC-1083

### DIFF
--- a/apps/meteor/app/lib/server/lib/notifyUsersOnMessage.js
+++ b/apps/meteor/app/lib/server/lib/notifyUsersOnMessage.js
@@ -177,14 +177,13 @@ export async function notifyUsersOnMessage(message, room) {
 		return message;
 	}
 
-	// if message sent ONLY on a thread, skips the rest as it is done on a callback specific to threads
-	if (message.tmid && !message.tshow) {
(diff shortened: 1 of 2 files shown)


apps/meteor/app/lib/server/lib/notifyUsersOnMessage.js
```diff
@@ -177,15 +177,14 @@ export async function notifyUsersOnMessage(message, room) {
 		return message;
 	}
 
+	// Update all the room activity tracker fields
```
ggazzo, line 180: The threads were made to not bother those who don't follow them, if we start updating lastmessage everyone will have their room modified, the behavior is correct

ggazzo (changes requested):

PR #31559, opened 2024-01-27
REFACTOR Update package.json to include newer version of yarn and node.

wolfiex:


ggazzo:
Thanks for  your effort, but not just about changing these values. Need to upgrade the meteor framework, but they are not ready yet.

Also pay attention we use conventional commits and should be used on titles

PR #31563, opened 2024-01-28
Update emoji picker to use joypixels

dhruvsinghal7494:
Updated the references.

updated the old emojione with the new joypixels in the codebase, includes updating library names, CSS filenames, class names, and any other reference

## Issue(s)
emojione is not working

## Steps to test or reproduce
 can check on chat app

## Further comments
Please merge it


sampaiodiego:
hi @dhruvsinghal7494 , thx for your contribution.. unfortunately looks like you simply did a search & replace of all emojione references and changed them to joypixels which doesn't work.. for example the package "joypixels" doesn't exist as it is referenced on the file below:
https://github.com/RocketChat/Rocket.Chat/blob/d8d4b1695cb1a9cf1869add4e5a71f05ab47ca36/apps/meteor/package.json#L330

there is also a licensing situation with joypixels we need to get sorted out, it is not simply an update.. 

we appreciate your contribution but we cannot merge this one though.

PR #31576, opened 2024-01-29
chore: Use more `permissionsRequired` in endpoints instead of manually checking

matheusbsilva137:

 - Use `permissionsRequired` as much as possible in all endpoints (the permissions checks left couldn't be transferred to `permissionsRequired` since they require scope or may not imply in an unauthorized error)

## Further comments

TC-1090
https://rocketchat.atlassian.net/browse/CORE-307


apps/meteor/app/api/server/v1/channels.ts
```diff
@@ -954,7 +951,12 @@ API.v1.addRoute(
 
 API.v1.addRoute(
 	'channels.list',
-	{ authRequired: true },
+	{
+		authRequired: true,
+		permissionsRequired: {
+			GET: { permissions: ['view-c-room', 'view-joined-room'], operation: 'hasAny' },
```
MarcosSpessatto, line 957: Didn't see any use of `view-joined-room` 🤔

apps/meteor/app/api/server/v1/roles.ts
```diff
@@ -90,7 +90,7 @@ API.v1.addRoute(
 
 API.v1.addRoute(
 	'roles.getUsersInRole',
-	{ authRequired: true },
+	{ authRequired: true, permissionsRequired: ['view-other-user-channels'] },
```
MarcosSpessatto, line 93: This is not the same permission as before 🤔

apps/meteor/app/api/server/v1/users.ts
```diff
@@ -452,6 +443,7 @@ API.v1.addRoute(
 	{
 		authRequired: true,
 		queryOperations: ['$or', '$and'],
+		permissionsRequired: ['view-d-room'],
```
MarcosSpessatto, line 446: Did we forget to remove the manual check from this endpoint?

MarcosSpessatto (changes requested):
Let's consider this a breaking change since we are changing the error messages that users calling the endpoints could rely on.

Also, I saw you changed the existing endpoints to match the new error message, can we create a test case for the permission check for all the endpoints (including the ones without tests)? This way we can ensure our change is respecting the permissions.

matheusbsilva137:
I'll create a PR for each set of endpoints. This way, it'll be easier to review this and implement tests
Tks for the suggestion, @MarcosSpessatto

matheusbsilva137:
I'll create a PR for each set of endpoints. This way, it'll be easier to review this and implement tests
Tks for the suggestion, @MarcosSpessatto

apps/meteor/app/api/server/v1/channels.ts
matheusbsilva137: It's inside `if (!hasPermissionToSeeAllPublicChannels) {`
The user must have at least one of the permissions

PR #31579, opened 2024-01-30
chore!: Remove `meteor/check` package dependency on all v1 endpoints

matheusbsilva137:

 - Remove all `meteor/check` dependencies on v1 endpoints (by using `validateParams` or throwing errors instead).

## Further comments

The only endpoint left using `check` is `livechat/room`. There's a comment there saying that ajv doesn't support what's needed for this case, so maybe we should think of another strategy here.

TC-1090
https://rocketchat.atlassian.net/browse/CORE-314


MarcosSpessatto (changes requested):
Let's split this PR, creating one PR for each file, since we need to add tests to cover every change the PR which is already big, will be even bigger.

PR #31591, opened 2024-01-31
regression: Engagement dashboard crashing 

yash-rajpal:
## Further comments

[TC-1094](https://rocketchat.atlassian.net/browse/TC-1094?atlOrigin=eyJpIjoiMjk2MjM3ZDcwYjU0NGVmZjk4NzM3N2NmNWIwMzM1YmQiLCJwIjoiaiJ9)

[TC-1094]: https://rocketchat.atlassian.net/browse/TC-1094?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


hugocostadev (approved):

rique223 (approved):
I believe it would be good to revisit this problem in the future and see which specific part of the new useEndpointAction hook caused this but since this is only happening in the engagement dashboard, we are close to release and useEndpoint fits better in this place I'll approve it.

PR #31593, opened 2024-01-31
regression: Moderation console translations

rique223:

Fixed a problem in the moderation console translations in which inexistent translation keys were being used to translate the Reports column of the users table tab and the success toast of the deactivate user action.

Demo gif:
moderation_consoles_translations

## Further comments

[TC-1098](https://rocketchat.atlassian.net/browse/TC-1098)

[TC-1098]: https://rocketchat.atlassian.net/browse/TC-1098?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


hugocostadev (dismissed):

dougfabris:
why not rebasing it instead of creating a new one?

PR #31600, opened 2024-02-01
fix: Looking at the user's permission before rendering the 'Start Call' button on UserInfo

Gustrb:
This solves an akward mismatch of the permissions, since we cannot start a call if we don't have the according permissions, but the button shows nonetheless, which is misleading.

Looking at the workspace setting called: VideoConf_Enable_DMs
Also looking at the user permission called: call-management
For a user to be able to call someone both of those should be enabled

## Steps to test or reproduce

1. Remove the Call Management permission from the user;
2. Settings > Conference call > Enable in direct messages = Disabled;
3. Open a new DM with the user
4. Click on the 'User Info' icon

Expected Behavior: It should not show the 'Start Call' button

if either of 'Call Management' or 'Enable in direct messages' are disabled the button won't appear, but if both of those are set the 'Start Call' button should appear and work correctly

## Further comments

SUP-470


yash-rajpal (approved):
LGTM!

ggazzo:
this is a `fix:` not a `feat:`

PR #31602, opened 2024-02-01
regression: Moderation console translations

rique223:

Fixed a problem with the moderation console translations in which inexistent translation keys were being used to translate the Reports column of the users table tab and the success toast of the deactivate user action.

Demo gif:
moderation_consoles_translations

## Further comments

[TC-1098](https://rocketchat.atlassian.net/browse/TC-1098?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ)

[TC-1098]: https://rocketchat.atlassian.net/browse/TC-1098?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/meteor/client/views/admin/moderation/ModerationConsoleTable.tsx
+++ b/apps/meteor/client/views/admin/moderation/ModerationConsoleTable.tsx
@@ -102,7 +102,7 @@ const ModerationConsoleTable: FC = () => {
 				{t('Moderation_Report_date')}
 			</GenericTableHeaderCell>,
 			<GenericTableHeaderCell key='reports' direction={sortDirection} active={sortBy === 'count'} onClick={setSort} sort='count'>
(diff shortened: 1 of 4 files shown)


hugocostadev (approved):

dougfabris:
Why you didn't rebase the previous PR instead of creating a new one?

rique223:
> Why you didn't rebase the previous PR instead of creating a new one?

The first approach that came to my mind was a cherry-pick since it was something I've seen in the past and had knowledge about, so it would be faster to do it, also I've never done a rebase in a situation like this.

PR #31619, opened 2024-02-02
chore: close non existing channel

heitortanoue:
User should be able to close a subscription to an non existing channel 

## Steps to test or reproduce

Create a channel

Open Compass MongoDB

Connect to the database

Search for “rocketchat_room”

Drop the collection

Try to access the channel you’ve createdIt will show the message “Room not found”

Hide the channel

### Expected Behavior
It should hide the channel

## Further comments

[TC-883](https://rocketchat.atlassian.net/browse/TC-883?atlOrigin=eyJpIjoiOWRjZTRiM2RlNjU1NGZlZDljZWExZmQwMzEzZGJlZWQiLCJwIjoiaiJ9)

[TC-883]: https://rocketchat.atlassian.net/browse/TC-883?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/meteor/app/api/server/v1/channels.ts
```diff
@@ -565,22 +565,36 @@ API.v1.addRoute(
...
+				roomName = findResult.name;
+
+				if (findResult._id) {
+					roomId = findResult._id;
+				}
+			} catch (e) {}
+
+			if (!roomId) {
+				return API.v1.failure('Could not find the channel or any subscription linked to it');
+			}
```
matheusbsilva137, line 585: Please use `validateParams` in the endpoint definition instead to make sure all params are correctly sent.

apps/meteor/app/api/server/v1/channels.ts
```diff
@@ -565,22 +565,36 @@ API.v1.addRoute(
...
+			try {
+				const findResult = await findChannelByIdOrName({
+					params: this.bodyParams,
+					checkedArchived: false,
+				});
+				roomName = findResult.name;
+
+				if (findResult._id) {
+					roomId = findResult._id;
+				}
```
matheusbsilva137, line 580:
```suggestion
				const roomName = this.bodyParams.roomName || findResult.name;
				const roomId = this.bodyParams.roomName || findResult._id;
```
We usually avoid using `let`. Also, you don't need to check if `_id` is defined since this is checked inside `findChannelByIdOrName`, which returns an `IRoom` object.
By applying these changes, you can delete the `let` definitions above.

.changeset/sixty-countries-fix.md
matheusbsilva137, line 1: I think I'd mark this PR as a chore, since this is not really intended behavior or a real bug fix :thinking: 
So you could remove the changeset

matheusbsilva137 (changes requested):

apps/meteor/app/api/server/v1/channels.ts
```diff
@@ -565,22 +565,36 @@ API.v1.addRoute(
...
+			} catch (e) {}
```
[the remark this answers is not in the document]
heitortanoue: That is because `findChannelByIdOrName` throws an error when the room is not found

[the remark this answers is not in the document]
matheusbsilva137: Ohhh, got it. Still, in this case I think we'd like to throw the error as it is in this function, so we don't need a try/catch to operate on the (possibly) returned error.
Also, if we decided to use the `catch` statement to return a different error, I believe that could be considered a breaking change -- since someone could expect the error as it is thrown by `findChannelByIdOrName`

apps/meteor/app/api/server/v1/channels.ts
heitortanoue: Here as the code is inside the `try` block, it wouldn't be possible to access the `roomName` and `roomId` from the outside, would it? About the `_id`, that is because if the room collection was dropped, the IRoom would not be found 🤔

matheusbsilva137: Since roomId and roomName are not used in the try/catch blockm you can still avoid using let by defining roomId and roomName after the search. So you could do something like:
`const roomId = findResult._id || this.bodyParams.roomId;` (after and outside of the try/catch)

packages/rest-typings/src/v1/channels/ChannelsCloseProps.ts
```diff
@@ -0,0 +1,28 @@
...
+		},
+		{
+			type: 'object',
+			properties: {
+				roomName: { type: 'string' },
+			},
+			required: ['roomName'],
+			additionalProperties: false,
+		},
+	],
```
matheusbsilva137, line 25:
```suggestion
	type: 'object',
	properties: {
		roomId: { type: 'string' },
		roomName: { type: 'string' },
	},
	oneOf: [{ required: ['roomId'] }, { required: ['roomName'] }],
	additionalProperties: false,
```
I believe it's easier to read like this.

apps/meteor/app/api/server/v1/channels.ts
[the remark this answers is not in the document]
matheusbsilva137: Please use `findByTypeAndNameOrId` instead of `findChannelByIdOrName` so that you don't need  this empty try/catch, which is not accept by our lint rules.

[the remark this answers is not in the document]
matheusbsilva137: Lint won't allow you to have this empty catch block.
Maybe you could try using the `findOneByIdOrName` method we already have. You'd just need to add this check after it:
```
if (room.t !== 'c' && room.t !== 'l') {
	throw new Meteor.Error('error-room-not-found', 'The required "roomId" or "roomName" param provided does not match any channel');
}
```

apps/meteor/app/api/server/v1/channels.ts
```diff
@@ -562,25 +563,35 @@ API.v1.addRoute(
...
 	'channels.close',
-	{ authRequired: true },
+	{ authRequired: true, validateParams: isChannelsCloseProps },
 	{
 		async post() {
-			const findResult = await findChannelByIdOrName({
-				params: this.bodyParams,
-				checkedArchived: false,
+			const receivedRoomId = 'roomId' in this.bodyParams ? this.bodyParams.roomId : undefined;
+			const receivedRoomName = 'roomName' in this.bodyParams ? this.bodyParams.roomName : undefined;
```
matheusbsilva137, line 570:
```suggestion
			const receivedRoomId = this.bodyParams.roomId;
			const receivedRoomName = this.bodyParams.roomName;
```
Won't this have the same effect? :thinking:

packages/rest-typings/src/v1/channels/ChannelsCloseProps.ts
matheusbsilva137, line 1: Can we also add a new end-to-end test to validate this? Just by not sending any param and checking if it returns an error

matheusbsilva137 (changes requested):

apps/meteor/app/api/server/v1/channels.ts
heitortanoue: This approach gives a type error such as `roomId does not exist on {roomName: string}` 🙁

matheusbsilva137: Ohhhhh, that's because you're using `oneOf` in the rest-typing definition
Try using `anyOf` instead. This will allow both params to be sent together
Sorry for the wrong suggestion there :confused:

PR #31632, opened 2024-02-03
fix: admin workspace item activation

YuvRaj6392:
In this modification, the isActive condition is adjusted to check if the currentPath exactly matches the path or if it starts with the path/. This should help ensure that the workspace item is correctly activated based on the provided paths.

### DIFF
--- a/apps/meteor/client/components/Sidebar/SidebarNavigationItem.tsx
+++ b/apps/meteor/client/components/Sidebar/SidebarNavigationItem.tsx
@@ -28,7 +28,7 @@ const SidebarNavigationItem: FC<SidebarNavigationItemProps> = ({
 	badge: Badge,
 }) => {
 	const path = pathSection;
-	const isActive = !!path && currentPath?.includes(path as string);
+	const isActive = !!path && (currentPath === path || currentPath?.startsWith(`${path}/`));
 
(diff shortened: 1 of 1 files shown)


alihamza1221:
> In this modification, the isActive condition is adjusted to check if the currentPath exactly matches the path or if it starts with the path/. This should help ensure that the workspace item is correctly activated based on the provided paths.

Hy, do it work for what is being wanted I have said that in the pr that `path` does not contain `workspace` but it's an item 
maybe due to some updates route from admin/workspace maybe changed to admin/info 
so, if we check as path with the currentpath than it will never active admin/workspace

I hope you can understand, I have put in hours to know what's going on

alihamza1221:
> In this modification, the isActive condition is adjusted to check if the currentPath exactly matches the path or if it starts with the path/. This should help ensure that the workspace item is correctly activated based on the provided paths.

Also in your pr you have used `path === currentPath` which will be true only if current path === path but 
what if rocket chat later on decide to make other sub routes inside the currentpath in that case the items that we used to come at that place willbeomce inactive creating the issue again. 
And can you explain why you have used startsWith(method) infact its also like match all as path and currentpath 
values are like admin/workspace, admin/info what is the logic here?

PR #31651, opened 2024-02-05
fix: livechat agent status set to available after reactivation

abhinavkrin:
This PR introduces a fix to the livechat agent status handling within Rocket.Chat. Previously, when a deactivated user, assigned as a livechat agent, was reactivated, their status in the livechat system was automatically set to 'Available'. This behavior was unintended as it could lead to conversations being assigned to agents who might not be ready to handle them. With this update, reactivated users will retain the 'Not Available' status until manually set to 'Available' by an admin or the agent themselves, ensuring better control and management of agent availability.

https://github.com/RocketChat/Rocket.Chat/assets/15830206/d1c053a8-be66-407a-8c6d-2b83bda53c34

## Issue(s)

This PR addresses the issue where livechat agent status was incorrectly set to 'Available' upon the reactivation of a deactivated user. Deactivated users should not be made available when they get activated.
## Steps to test or reproduce

To test this fix, follow these steps:
1. As an admin, go to Admin Panel > Users and deactivate a user who is also a livechat agent.
2. Observe in Omnichannel > Agents that the agent's status is correctly set to 'Not Available'.
3. Reactivate the user in Admin Panel > Users.
4. Check the status of the user in Omnichannel > Agents.
## Further comments

[CORE-46](https://rocketchat.atlassian.net/browse/CORE-46)

[CORE-46]: https://rocketchat.atlassian.net/browse/CORE-46?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.changeset/nice-ducks-shout.md
```diff
@@ -0,0 +1,5 @@
+---
+'@rocket.chat/meteor': patch
+---
+
+Fixed auto-availability of reactivated livechat agents; they now stay 'Not Available' until manually set
```
MarcosSpessatto, line 5:
```suggestion
Fixed auto-availability of reactivated livechat agents; they now stay 'Not Available' until manually set to 'Available'
```

apps/meteor/app/livechat/server/hooks/afterUserActions.ts
```diff
@@ -37,16 +37,8 @@ const handleDeactivateUser = async (user: IUser) => {
 	}
 };
 
-const handleActivateUser = async (user: IUser) => {
-	if (isAgent(user) && user.username) {
-		await LivechatTyped.addAgent(user.username);
```
MarcosSpessatto, line 42: Is losing some of the behaviors inside the addAgent method (e.g setOperator, dispatch some callbacks) the expected result here? I understand the task asks for not to set the User as Available/Not available but what about the other behaviors?

P.S. I'm not saying this is the correct behavior as well, we should not run `livechat.onNewAgentCreated` after changing a user status, but we were already, so might places be relying on that. 😞

MarcosSpessatto (changes requested):

apps/meteor/app/lib/server/functions/setUserActiveStatus.ts
```diff
@@ -90,10 +90,6 @@ export async function setUserActiveStatus(userId: string, active: boolean, confi
 
 	await Users.setUserActive(userId, active);
 
-	if (active && !user.active) {
-		await callbacks.run('afterActivateUser', user);
```
abhinavkrin, line 94: Hey @MarcosSpessatto , When I removed this part of code. I could not find any other use case of 'afterActivateUser' callback.
afterActivateUser
Did a search on this callback, and found only one usecase (which i removed).
So I thought of completely removing this callback. Am I missing something :thinking: ?

MarcosSpessatto: No, you are right, if it's not being used anymore, we can remove that. I was talking about the other behaviors, like, previously whenever the agent changed its status the callback was dispatched and the `await LivechatTyped.addAgent(user.username)` was called, inside this method we do have some logic and by removing this we also would not execute that code anymore, I'm just asking if losing those behaviors is the expected behavior since we can have some collateral effects by doing that.

MarcosSpessatto: I believe this is a good idea. 👍

MarcosSpessatto (dismissed):

apps/meteor/app/livechat/server/lib/LivechatTyped.ts
```diff
@@ -1682,6 +1682,14 @@ class LivechatClass {
 		return false;
 	}
 
+	async afterAgentUserActivated(user: IUser) {
+		if (!user.roles.includes('livechat-agent')) {
+			throw new Meteor.Error('invalid-user-role');
```
matheusbsilva137, line 1687:
```suggestion
			throw new Error('invalid-user-role');
```
Couldn't we just throw a normal error here to avoid using more Meteor?

matheusbsilva137 (changes requested):

matheusbsilva137 (approved):

MarcosSpessatto (approved):

lmauromb (approved):

PR #31655, opened 2024-02-05
test: make chat api tests fully independent

MarcosSpessatto:
https://rocketchat.atlassian.net/browse/CORE-150


apps/meteor/tests/end-to-end/api/05-chat.js
```diff
@@ -2167,13 +2169,21 @@ describe('[Chat]', function () {
...
+			).body.discussion;
 		});
 
+		after(() =>
+			Promise.all([
+				deleteRoom({ type: 'c', roomId: testChannel._id }),
+				request
+					.post(api('rooms.delete'))
+					.set(credentials)
+					.send({ roomId: testChannel._id })
```
sampaiodiego, line 2181: I guess this room is being deleted twice?

apps/meteor/tests/end-to-end/api/05-chat.js
```diff
@@ -785,44 +791,40 @@ describe('[Chat]', function () {
...
-						.get(api('chat.getMessage'))
-						.set(credentials)
-						.query({
-							msgId: ytEmbedMsgId,
-						})
-						.expect('Content-Type', 'application/json')
-						.expect(200)
-						.expect((res) => {
-							expect(res.body).to.have.property('message').to.have.property('urls').to.be.an('array').that.is.not.empty;
+				request
```
sampaiodiego, line 794: not sure about removing the `setTimeout` here.. the oembed action runs in a `afterSaveMessage` callback and it depends on how fast the CI can make the request and also how fast YT can respond to the request.. 

I have to admit though I'm shocked the test didn't fail.. but maybe removing the timeout can make it flaky

apps/meteor/tests/end-to-end/api/05-chat.js
```diff
@@ -785,44 +791,40 @@ describe('[Chat]', function () {
...
-						.get(api('chat.getMessage'))
-						.set(credentials)
-						.query({
-							msgId: imgUrlMsgId,
-						})
-						.expect('Content-Type', 'application/json')
-						.expect(200)
-						.expect((res) => {
-							expect(res.body).to.have.property('message').to.have.property('urls').to.be.an('array').that.is.not.empty;
+				request
```
sampaiodiego, line 814: same concern as above.. 

I wonder though if we should change both these end-to-end tests for some unit tests.. what we need is to check if the oembed code is doing it's thing.. relying on external services (like on the youtube link above) is prone to fail.. wdyt?

sampaiodiego (changes requested):

apps/meteor/tests/end-to-end/api/05-chat.js
MarcosSpessatto: Agreed, I'll remove (skip) these tests in this PR and I will create that in another PR, wdyt?

sampaiodiego (dismissed):

sampaiodiego (approved):

PR #31660, opened 2024-02-05
fix: Quote attachments are not deleted on message update

matheusbsilva137:

 - Remove old quote attachments from messages on every update.

## Steps to test or reproduce

There are two main test cases:
1. Update a message to remove all of its quotes: no quote attachment should be displayed within the message after the update;
2. Update a message to remove some of its quotes: only quote attachments related to links that are still in the message should be displayed.

This can be tested through the UI **OR** using the `updateMessage` method or `chat.update` endpoint.

## Further comments

SUP-411


packages/core-typings/src/IMessage/MessageAttachment/MessageQuoteAttachment.ts
```diff
@@ -13,4 +13,5 @@ export type MessageQuoteAttachment = {
 	attachments?: Array<MessageQuoteAttachment>; // TODO this is cauising issues to define a model, see @ts-expect-error at apps/meteor/app/api/server/v1/channels.ts:274
 } & MessageAttachmentBase;
 
-export const isQuoteAttachment = (attachment: MessageAttachment): attachment is MessageQuoteAttachment => 'message_link' in attachment;
+export const isQuoteAttachment = (attachment: MessageAttachment): attachment is MessageQuoteAttachment =>
```
MarcosSpessatto, line 16: Can we add a unit test to cover this?

apps/meteor/server/services/messages/hooks/BeforeSaveJumpToMessage.ts
```diff
@@ -71,6 +71,10 @@ export class BeforeSaveJumpToMessage {
 			useRealName: boolean;
 		};
 	}): Promise<IMessage> {
+		if (message?.attachments) {
```
MarcosSpessatto, line 74: I'd say we should verify if this is an edited message here since this case only applies to edited messages.

apps/meteor/server/services/messages/hooks/BeforeSaveJumpToMessage.ts
```diff
@@ -71,6 +71,10 @@ export class BeforeSaveJumpToMessage {
...
+			// Do not keep old quote attachments since they may not still be linked to the message
+			message.attachments = message.attachments?.filter((attachment) => !isQuoteAttachment(attachment));
```
MarcosSpessatto, line 76: In my opinion, it would be clearer if we could create a private function with a proper name for this operation. 

From what I understood we are "resetting" the attachments at this point, removing all the quotes and keeping the other kind of attachments, right? So, if we have new URLs(quotes) to parse it will reparse below this point, otherwise(if the existing quotes were deleted) it will return the message without the old (quote) attachment.

What do you think?

apps/meteor/tests/end-to-end/api/05-chat.js
```diff
@@ -1129,6 +1129,44 @@ describe('[Chat]', function () {
...
+				.expect('Content-Type', 'application/json')
+				.expect(200)
+				.expect((res) => {
+					expect(res.body).to.have.property('success', true);
+					expect(res.body).to.have.nested.property('message.msg', `Testing quotes ${quotedMsgLink}`);
+					expect(res.body.message).to.have.property('attachments').that.is.an('array').that.has.lengthOf(1);
+				});
+		});
+
+		it('should erase old quote attachments when updating a message', async () => {
```
MarcosSpessatto, line 1153: Can we add a test case for when we are updating or replacing or both the existing quotes of a message?

apps/meteor/tests/end-to-end/api/24-methods.js
```diff
@@ -1955,6 +1955,7 @@ describe('Meteor.methods', function () {
 
 	describe('[@updateMessage]', () => {
 		let rid = false;
+		let roomName = false;
```
MarcosSpessatto, line 1958: Any reason to define `false` here? 🤔

apps/meteor/tests/end-to-end/api/24-methods.js
```diff
@@ -2067,6 +2069,72 @@ describe('Meteor.methods', function () {
...
+				.expect(200)
+				.expect((res) => {
+					expect(res.body).to.have.property('message').that.is.an('object');
+					expect(res.body.message).to.have.property('msg', `${quotedMsgLink} updated`);
+					expect(res.body.message).to.have.property('attachments').that.is.an('array').that.has.lengthOf(1);
+					expect(res.body.message.attachments[0]).to.have.property('message_link').that.is.a('string');
+				});
+		});
+
+		it('should remove a quote attachment from a message', async () => {
```
MarcosSpessatto, line 2107: Can we add a test case for when we are updating or replacing or both the existing quotes of a message?

MarcosSpessatto (changes requested):
Also, can we add more (unit) test cases to cover your changes?

https://github.com/RocketChat/Rocket.Chat/blob/develop/apps/meteor/tests/unit/server/services/messages/hooks/BeforeSaveJumpToMessage.tests.ts#L51

apps/meteor/tests/end-to-end/api/24-methods.js
matheusbsilva137: No, that's just me copy+pasting :see_no_evil:

apps/meteor/server/services/messages/hooks/BeforeSaveJumpToMessage.ts
matheusbsilva137: I did that change and got a test unit error. It happened because we were expecting this function to also not duplicate quotes from message attachments that have already been provided (on the first time a message is sent, not only on edit)

Also, I'm thinking we should **always** remove existing quote attachments and rebuild them here. **Reason:** someone could provide a quote attachment in a message and skip all the checks we have in this file :grimacing: So my decision was to always remove quote attachments and make sure only the ones provided in `urls` are created, wdyt?

packages/core-typings/src/IMessage/MessageAttachment/MessageQuoteAttachment.ts
matheusbsilva137: Done!

MarcosSpessatto (approved):

apps/meteor/server/services/messages/hooks/BeforeSaveJumpToMessage.ts
MarcosSpessatto: I think that works if the URLs are sent whenever we edit a message. Once we have a test covering this decision, we are good to go.

sampaiodiego (approved):

PR #31665, opened 2024-02-06
feat: Removes hardcoded prune message limit in favor of setting

### LINKED ISSUE #21305: Deleting Messages deletes max 2000 messages
### Description:

Deleting Messages deletes max 2000 messages

### Steps to reproduce:

1. click on the channel in question
2. click on the triple dot in the upper right corner
3. click delete messages
4. select older-than
5. choose 2021-01-01
6. click the big fat red DELETE button

### Expected behavior:

Messages before 2021-01-01 are deleted

### Actual behavior:

2000 Messages are deleted.

### Server Setup Information:

- Version of Rocket.Chat Server: 3.12.1
- Operating System: Linux
- Deployment Method: snap
- DB Replicaset Oplog: I dont know what this is
- NodeJS Version: 12.18.4
- MongoDB Version: 3.6.14

### Client Setup Information

- Desktop App or Browser Version: desktop 3.0.3 
- Operating System: MacOS

### Additional context

none that I am aware of.

### Relevant logs:

no relevant logs found.

dougfabris:
## Issue(s)

Closes #21305


gabriellsh (changes requested):
I have a couple of concerns about this:

- If I pass no limit to the API, all messages will be deleted, disregarding the setting.
- This could be considered a breaking change. I suggest we make the limit setting be disabled if passing `0` as a value.

It seems a little strange that I can pass a value to limit the pruning, OR deleting all messages, but no inbetween. I think this setting would make more sense if it was the limit even if I didn't pass the parameter to the API (e.g if no limit is provided, use the setting value).

apps/meteor/app/api/server/v1/rooms.ts
```diff
@@ -264,6 +264,10 @@ API.v1.addRoute(
 				return API.v1.failure('Body parameter "oldest" is required.');
 			}
 
+			if (limit && limit > settings.get<number>('Prune_message_limit')) {
```
MarcosSpessatto, line 267: This should be the upper limit, if the user provides a limit greater than the setting it should be blocked. On the other hand, if the user doesn't pass any limit we should use the setting limit, unless it's not defined, in this case, we can delete all messages.

apps/meteor/tests/end-to-end/api/09-rooms.js
```diff
@@ -462,6 +462,25 @@ describe('[Rooms]', function () {
 					expect(res.body).to.have.property('count', 0);
 				});
 		});
+		it('should return error if limit is higher than setting', async () => {
+			await updateSetting('Prune_message_limit', 1);
```
MarcosSpessatto, line 466: Can we set `after` finish the tests this value back to the defaults, please?

MarcosSpessatto (changes requested):

PR #31671, opened 2024-02-07
test: make rooms api test fully independent

MarcosSpessatto:
https://rocketchat.atlassian.net/browse/CORE-154


apps/meteor/tests/data/rooms.helper.js
```diff
@@ -54,14 +54,14 @@ function actionRoom({ action, type, roomId }) {
...
 				roomId,
 			})
 			.end(resolve);
 	});
 }
 
-export const deleteRoom = ({ type, roomId }) => actionRoom({ action: 'delete', type, roomId });
+export const deleteRoom = ({ type, roomId, overrideCredentials = credentials }) => actionRoom({ action: 'delete', type, roomId, overrideCredentials });
 
 export const closeRoom = ({ type, roomId }) => actionRoom({ action: 'close', type, roomId });
```
sampaiodiego, line 67: this seems to not be used anymore, as it would not even work since it is not providing any credentials.. should it be removed?

sampaiodiego (changes requested):

PR #31676, opened 2024-02-07
test: make assets api tests fully independent

MarcosSpessatto:


apps/meteor/tests/end-to-end/api/14-assets.js
```diff
@@ -1,34 +1,25 @@
...
-			.set(credentials)
-			.send({ permissions: [{ _id: 'manage-assets', roles: ['admin'] }] })
-			.expect('Content-Type', 'application/json')
-			.expect(200)
-			.expect((res) => {
-				expect(res.body).to.have.property('success', true);
-			})
-			.end(done);
-	});
+	before(() => updatePermission('manage-assets', ['admin']));
```
sampaiodiego, line 13: looks like the tests don't even touch on this permission, maybe this is not required at all?

sampaiodiego (changes requested):

apps/meteor/tests/end-to-end/api/14-assets.js
MarcosSpessatto: They were updating the setting. I just removed the "manual" update to a `before/after` approach.

```
it('giving "manage-assets" permission to user', (done) => {
		request
			.post(api('permissions.update'))
			.set(credentials)
			.send({ permissions: [{ _id: 'manage-assets', roles: ['admin'] }] })
			.expect(200)
			.expect((res) => {
			})
			.end(done);
	});
```

sampaiodiego: true.. my bad

sampaiodiego (approved):

PR #31698, opened 2024-02-08
feat: Improve File Upload modal and endpoint

matheusbsilva137:

 - Accept new `altDescription` param in the `rooms.upload` endpoint. This can be used by screen readers to describe images;
 - Improve File Upload modal to have a new Message Box component in it, which allows users to mention other users os rooms easily;
 - Send File Upload text content in the `msg` field instead of `description` (this allows mentions to be processed and rendered correctly).

## Further comments

CORE-16


apps/meteor/client/views/room/composer/messageBox/TextMessageBox.tsx
```diff
@@ -0,0 +1,201 @@
...
+	const messageComposerRef = useRef<HTMLElement>(null);
+	const shadowRef = useRef<HTMLDivElement>(null);
+
+	const storageID = `messagebox_${rid}-fileupload`;
+	const callbackRef = useCallback(
+		(node: HTMLTextAreaElement) => {
+			if (node === null || composer) {
+				return;
+			}
+			setComposer(createComposerAPI(node, storageID));
```
matheusbsilva137, line 86: The composer is set, but never cleared. Since there's no "send" button in this component, this action must be triggered bu its parent component (the upload modal, in this case). I'm not sure what would be the best way to tackle this :thinking: do we do something similar anywhere else?

apps/meteor/client/views/room/modals/FileUploadModal/FileUploadModal.tsx
```diff
@@ -101,11 +112,28 @@ const FileUploadModal = ({
...
 							<Field>
-								<FieldLabel>{t('Upload_file_description')}</FieldLabel>
+								<FieldLabel>{t('Message')}</FieldLabel>
 								<FieldRow>
-									<TextInput value={description} onChange={handleDescription} placeholder={t('Description')} ref={ref} />
+									<Box is='section' display='flex' flexDirection='column' flexGrow={1} flexShrink={1} flexBasis='auto' height='full'>
+										<ComposerPopupProvider room={room}>
+											<TextMessageBox showFormattingTips rid={room._id} onChange={handleMessage} />
+										</ComposerPopupProvider>
+									</Box>
```
matheusbsilva137, line 136: The popup component is not showing **over** the other fields in this modal, it's actually making the modal grow for it to be displayed:

Captura de tela de 2024-02-08 16-34-10

Am I missing something?

apps/meteor/client/views/room/composer/messageBox/TextMessageBox.tsx
matheusbsilva137, line 1: I could manage to use the emoji picker here (by calling it directly, not using the chat object), but it was being displayed under the modal :thinking: so I decided to remove it for now. Is this an easy fix or would we need a bigger refactor?

matheusbsilva137 (commented):
Changes on this PR are manily FE related. If we decide to move on with this PR, I need some FE help to improve it (left some comments explaining what's missing)
cc @gabriellsh

rodrigok:
@matheusbsilva137 lets discuss these changes in relation to the e2ee ones.

PR #31701, opened 2024-02-08
fix: Correct display of discussion names in sidebar search

abhinavkrin:
This fix alters the spotlight search endpoint method to include the `prid` (parent room ID) for rooms. This adjustment ensures that the client correctly identifies discussions and displays their names instead of IDs in the sidebar search results. This change resolves the issue where discussion IDs were displayed when the "Allow Special Characters in Room Names" setting was disabled, aligning the behavior with the search results when the setting is enabled.

## Issue(s)

Fixes issue where discussion names were incorrectly displayed as IDs in sidebar search results under specific conditions.
## Steps to test or reproduce

  1. Disable "Allow Special Characters in Room Names".
  2. Create a discussion inside a public channel.
  3. Search for the discussion by name in the sidebar with a different user account.
## Further comments

[SUP-410](https://rocketchat.atlassian.net/browse/SUP-410)

[SUP-410]: https://rocketchat.atlassian.net/browse/SUP-410?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


matheusbsilva137 (dismissed):

MarcosSpessatto (changes requested):
Can we add a test case to cover this scenario?

abhinavkrin:
> Can we add a test case to cover this scenario?

Hey @MarcosSpessatto I have added test for this scenario. It tests discussion search in both cases when `UI_Allow_room_names_with_special_chars` is `true` or `false`

MarcosSpessatto (approved):

PR #31720, opened 2024-02-09
fix: New users are not automatically inserted in auto-join rooms inside default teams

Gustrb:

Earlier, whenever we had auto join teams, when the user joined, we would not look at the auto-join channels inside of it.
The idea is, that when we are going to create the subscriptions between the team and the user, we check whether the team has a default channel, if so, fetch those and add the user to it as well

## Steps to test or reproduce

1. Convert the general room into a team
2. Create a new channel inside the general team
3. Set the new channel as auto-join
4. Create a new user

Expected Behavior: The user should be added to both the team and the channel
Current behavior: The user is being added only to the team

## Further comments

CORE-58


apps/meteor/app/lib/server/functions/addUserToDefaultChannels.ts
```diff
@@ -10,6 +10,23 @@ export const addUserToDefaultChannels = async function (user: IUser, silenced?:
...
+		for await (const teamRoom of teamRooms) {
+			const defaultTeamRooms = await Rooms.findDefaultRoomsForTeam(teamRoom.teamId).toArray();
+
+			const defaultTeamRoomsThatWereNotAlreadyAdded = defaultTeamRooms.filter((channel) => !roomsThatAreGoingToBeJoined.has(channel._id));
+
+			defaultTeamRoomsThatWereNotAlreadyAdded.forEach((channel) => roomsThatAreGoingToBeJoined.add(channel._id));
+			// Add the channels to the defaultRooms list
+			defaultRooms.push(...defaultTeamRoomsThatWereNotAlreadyAdded);
+		}
+	}
```
MarcosSpessatto, line 28: We could extract this (new) team functionality to a new function, what do you think?

MarcosSpessatto (changes requested):
Would it be possible to add tests to cover this fix? It could be API or even UI...

apps/meteor/tests/end-to-end/api/01-users.js
```diff
@@ -277,6 +290,75 @@ describe('[Users]', function () {
...
+
+			it('should create a subscription for the user in the default channels if the joinDefaultChannels is true', async () => {
+				await request
+					.get(api('subscriptions.getOne'))
+					.set(credentials)
+					.query({ roomId: defaultTeamId })
+					.expect('Content-Type', 'application/json')
+					.expect(200)
+					.expect((res) => {
+						expect(res.body).to.have.property('success', true);
```
MarcosSpessatto, line 324: Can we ensure the room id as well?

apps/meteor/tests/end-to-end/api/01-users.js
```diff
@@ -277,6 +290,75 @@ describe('[Users]', function () {
...
+					});
+			});
+
+			it('should create a subscription for the user in all the auto join channels of the team', async () => {
```
MarcosSpessatto, line 328: Can we create the cases? Do you think it makes sense?

- Create a non-default team and create an auto-join room there (not sure if it's possible though)
- Create a default team with a channel that is not auto-join
- Create a non-default team with a channel that is also not auto-join

apps/meteor/tests/end-to-end/api/01-users.js
```diff
@@ -277,6 +290,75 @@ describe('[Users]', function () {
...
+				const user2Credentials = await login(user2.username, password);
+
+				await request
+					.get(api('subscriptions.getOne'))
+					.set(user2Credentials)
+					.query({ roomId: group.body.group._id })
+					.expect('Content-Type', 'application/json')
+					.expect(200)
+					.expect((res) => {
+						expect(res.body).to.have.property('success', true);
```
MarcosSpessatto, line 358: Can we ensure roomId too?

apps/meteor/tests/end-to-end/api/01-users.js
```diff
@@ -277,6 +290,75 @@ describe('[Users]', function () {
...
+							teamId: defaultTeamId,
+							topic: '',
+						},
+					});
+				await request.post(api('teams.updateRoom')).set(credentials).send({
+					roomId: group.body.group._id,
+					isDefault: true,
+				});
+
+				const user2 = await createUser({ joinDefaultChannels: true });
```
MarcosSpessatto, line 348: Can we delete this user? Using `after` would be the ideal.

MarcosSpessatto (changes requested):

apps/meteor/tests/end-to-end/api/01-users.js
matheusbsilva137: Done! Thanks for the suggestions

apps/meteor/tests/data/rooms.helper.js
```diff
@@ -65,3 +65,27 @@ function actionRoom({ action, type, roomId }) {
...
+	return request
+		.post(api('rooms.saveRoomSettings'))
+		.set(credentials)
+		.send({
+			rid: roomId,
+			default: true,
+			favorite: {
+				defaultValue: true,
+				favorite: false
+			}
```
MarcosSpessatto, line 79: We should not favorite the room here since the function name is saying it's just making it as default. I think it would be better to have two different functions or create a saveRoomSettings function and accept the desired parameters.

apps/meteor/tests/data/rooms.helper.js
```diff
@@ -65,3 +65,27 @@ function actionRoom({ action, type, roomId }) {
...
+};
+
+export const unsetDefaultRoom = ({ roomId }) => {
+	return request
+		.post(api('rooms.saveRoomSettings'))
+		.set(credentials)
+		.send({
+			rid: roomId,
+			default: false,
+		});
```
MarcosSpessatto, line 90: We can reuse the same endpoint call here maybe?

apps/meteor/tests/end-to-end/api/01-users.js
```diff
@@ -277,6 +278,165 @@ describe('[Users]', function () {
...
+						readonly: false,
+						members: [],
+						extraData: {
+							broadcast: false,
+							encrypted: false,
+							teamId: defaultTeamId,
+							topic: '',
+						},
+					});
+				group = body.group;
```
MarcosSpessatto, line 314: Can we use the existing `createRoom` function here?

apps/meteor/tests/end-to-end/api/01-users.js
```diff
@@ -277,6 +278,165 @@ describe('[Users]', function () {
...
+					deleteRoom({ roomId: group._id, type: 'p' }),
+					deleteTeam(credentials, teamName),
+					deleteUser(user),
+					deleteUser(user2),
+					deleteUser(user3),
+					deleteUser(userNoDefault),
+				]),
+			);
+
+			it('should NOT create create subscriptions to non default teams or rooms', async () => {
```
MarcosSpessatto, line 328:
```suggestion
			it('should NOT create subscriptions to non default teams or rooms even if joinDefaultChannels is true', async () => {
```

apps/meteor/tests/end-to-end/api/01-users.js
```diff
@@ -277,6 +278,165 @@ describe('[Users]', function () {
...
+							broadcast: false,
+							encrypted: false,
+							teamId: defaultTeamId,
+							topic: '',
+						},
+					});
+				group = body.group;
+			});
+
+			after(async () =>
```
MarcosSpessatto, line 317:
```suggestion
			after(() =>
```

MarcosSpessatto (changes requested):

MarcosSpessatto (approved):

matheusbsilva137 (approved):

PR #31731, opened 2024-02-11
new approach for live composer

imf-ali:


imf-ali:
Hey @hugocostadev and @tassoevan, can you check this approach if this is going in right direction? Would love your feedback. Its still in progress

tassoevan:
I couldn't make it work, for some reason.

PR #31745, opened 2024-02-13
fix: #31722 roomName is not getting update 

### LINKED ISSUE #31722: Unable to change room names
### Description:

On 2 snap installations I cannot change the names of existing rooms. When I edit the room name a message pops up saying it was successful, though it does not actually update the name. They are both on the current version 6.6.0

### Steps to reproduce:

1. Go to workspace
2. Go to Rooms
3. Edit Room name
4. Change does not reflect

### Expected behavior:

It should update the room name on the front end

### Actual behavior:

The room name does not update 

### Server Setup Information:

- Version of Rocket.Chat Server: 6.6.0
- Operating System: Ubuntu
- Deployment Method: snap
- Number of Running Instances: 2

Pritam-sd-dev:
## Issue(s)

closes #31722

### DIFF
--- a/apps/meteor/client/views/admin/rooms/EditRoom.tsx
+++ b/apps/meteor/client/views/admin/rooms/EditRoom.tsx
@@ -97,7 +97,7 @@ const EditRoom = ({ room, onChange, onDelete }: EditRoomProps) => {
 	const handleArchive = useArchiveRoom(room);
 
 	const handleUpdateRoomData = useEffectEvent(async ({ isDefault, roomName, favorite, ...formData }) => {
-		const data = getDirtyFields(formData, dirtyFields);
+		const data = getDirtyFields({ isDefault, roomName, favorite, ...formData }, dirtyFields);
 		delete data.archived;
 		delete data.favorite;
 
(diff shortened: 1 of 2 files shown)


apps/meteor/client/views/admin/rooms/EditRoom.tsx
```diff
@@ -97,7 +97,7 @@ const EditRoom = ({ room, onChange, onDelete }: EditRoomProps) => {
 	const handleArchive = useArchiveRoom(room);
 
 	const handleUpdateRoomData = useEffectEvent(async ({ isDefault, roomName, favorite, ...formData }) => {
-		const data = getDirtyFields(formData, dirtyFields);
```
Pritam-sd-dev, line 100: since roomName is not passed due to which edit_room_name functionality was not working

apps/meteor/client/views/admin/rooms/EditRoom.tsx
```diff
@@ -97,7 +97,7 @@ const EditRoom = ({ room, onChange, onDelete }: EditRoomProps) => {
...
+		const data = getDirtyFields({ isDefault, roomName, favorite, ...formData }, dirtyFields);
```
yash-rajpal, line 100: API doesn't accept `isDefault` field and that case is taken into account below, so this can cause errors when this field is update, I would suggest that we remove `isDefault` from here.

yash-rajpal, line 100: Case for updating `roomName` is also handled below, having it here would override that. I would suggest instead of overriding the roomName, better delete roomName from `data` object.

Pritam-sd-dev:
@yash-rajpal 
thanks for correcting me, i got the point that why you have overriden my change

PR #31748, opened 2024-02-14
fix: wrapped quote message links in the main box instead of the time box

### LINKED ISSUE #31747: [BUG]: Quoted message link is wrapped with time text instead of the entire box.
### Description:

Currently, when quoting a message and sending it, the message link gets wrapped inside the **time text/box**, which is inappropriate behavior. It doesn't make much sense. Instead, the quoted message should be wrapped over the **complete box** so that if anyone clicks on that box, they will be scrolled towards that message easily.

### Steps to reproduce:

1. Open Rocket.Chat.
2. Send a message, for example, "Hi."
3. Quote that message and reply with some message, for instance, "Hello."
4. After quoting, you will notice that the wrapped message link is with the time text, and when you click on it, you will be scrolled towards the quoted message.

### Expected behavior:

The entire box should be wrapped with the message link so that the user can click on the box and be scrolled towards the message.

### Actual behavior:

The quoted message is wrapped in time text, which is not appropriate.

https://github.com/RocketChat/Rocket.Chat/assets/78961432/f8affd73-6ccd-4dbd-8a6e-27955e01b995

### Server Setup Information:

- Rocket.Chat Server Version: 6.7.0-develop
- Operating System: Ubuntu 22.04.3 LTS 
- Deployment Method: Local development server deployment
- NodeJS Version: 14.21.4 - x64
- MongoDB Version: 6.0.3

### Client Setup Information

- Desktop App or Browser Version: Microsoft Edge Version 120.0.2210.133 (Official build) (64-bit)
- Operating System: Ubuntu 22.04.3 LTS

Spiral-Memory:

The issue was that quote links were wrapped in the time box. I removed the message link from there and set the color as 'hint' as it was default to render in both cases. Then, I wrapped the message link around the entire `<AttachmentContent>` component for better accessibility. I set the `textDecoration` as none. Now, the user can click on the main box directly, and they will be scrolled to the quoted message.

## Issue(s)
Closes #31747 

https://github.com/RocketChat/Rocket.Chat/assets/78961432/c21f36f9-b838-40d6-b495-16739d153c91

## Steps to test or reproduce

The steps to reproduce are mentioned in the issue I created, #31747

## Further comments
NA


Spiral-Memory:
Hey @dougfabris , can you please check if the problem is genuine and if the fix is right? Let me know if anything needs to be changed. Thanks!

Spiral-Memory:
Hey @dougfabris  @hugocostadev  , can you please check if the problem is genuine and if the fix is right? Let me know if anything needs to be changed. Thanks!

Spiral-Memory:
@hugocostadev @juliajforesti @dougfabris Could any of you please review this PR? I believe this issue is significant and needs to be addressed. If any changes are required, kindly let me know.

PR #31750, opened 2024-02-14
fix: Don't show join default channels option for edit user form 

yash-rajpal:

When we create new users, we have to option to add these new users to default channels.
We were also showing this option on edit users panel, which doesn't make any sense, so this PR removes this option.
## Further comments

SUP-423


Gustrb (approved):

.changeset/real-bobcats-train.md
```diff
@@ -0,0 +1,6 @@
+---
+'@rocket.chat/rest-typings': patch
+'@rocket.chat/meteor': patch
+---
+
+Don't show Join default channels option on edit user form and stop accepting this param on API
```
MarcosSpessatto, line 6: We can't just remove the parameter right now in a sudden manner, it's considered a breaking change.

packages/rest-typings/src/v1/users/UsersUpdateParamsPOST.ts
```diff
@@ -16,7 +16,6 @@ export type UsersUpdateParamsPOST = {
 		nickname?: string;
 		statusText?: string;
 		roles?: string[];
-		joinDefaultChannels?: boolean;
```
MarcosSpessatto, line 19: We can't just remove the parameter right now in a sudden manner, it's considered a breaking change. You can still keep the client-side changes though.

MarcosSpessatto (changes requested):

packages/rest-typings/src/v1/users/UsersUpdateParamsPOST.ts
yash-rajpal: You are right, I'll revert this.

MarcosSpessatto (changes requested):
I think we could add a UI test to cover this behavior. WDYT?

MarcosSpessatto (approved):

PR #31757, opened 2024-02-14
chore: Remove references to EE code from the app bridges

pierre-lehnen-rc:
## Further comments

With the new converter type definitions, several type errors were detected - fixing them would require a breaking change on the apps-engine, so for now I added typecasts to keep the current behavior in place. I added a comment with the `#AppsEngineTypes` tag to all of those typecasts.

To minimize the potential risk of those typecasts, I typed everything in two steps: First I assign things to a typed constant using the correct types, then I do a typecast to the type that is currently expected. For example:

```ts
const promise: Promise<IUser | undefined> = getUser();
return promise as Promise<IUser>;
```

This way, if the `getUser` function signature is modified the typecast will not suppress any new type errors.


apps/meteor/app/apps/server/converters/threads.ts
```diff
@@ -101,6 +103,7 @@ export class AppThreadsConverter {
 			groupable: 'groupable',
 			token: 'token',
 			blocks: 'blocks',
+			// #ToDo: #AppsEngineTypes - bug? - This "room" object won't be properly mapped to the app object
```
pierre-lehnen-rc, line 106: Possible bug on apps-engine: Room object being passed to the map won't be copied to the resulting object

apps/meteor/app/apps/server/bridges/messages.ts
```diff
@@ -44,30 +46,33 @@ export class AppMessageBridge extends MessageBridge {
...
 
-		const convertedMsg = await this.orch.getConverters()?.get('messages').convertAppMessage(message);
-		const convertedUser = await this.orch.getConverters()?.get('users').convertById(user.id);
+		// #TODO: #AppsEngineTypes - Remove explicit types and typecasts once the apps-engine definition/implementation mismatch is fixed.
+		const convertedMsg: IMessage | undefined = await this.orch.getConverters()?.get('messages').convertAppMessage(message);
+		const convertedUser: IAppsUser | undefined = await this.orch.getConverters()?.get('users').convertById(user.id);
 
-		await deleteMessage(convertedMsg, convertedUser);
+		// #TODO: #AppsEngineTypes - deleteMessage expects an IUser but is receiving an IAppsUser - this is probably a bug
+		await deleteMessage(convertedMsg as IMessage, convertedUser as unknown as IUser);
```
pierre-lehnen-rc, line 72: This one is probably a bug; The deleteMessage function expects an IUser but is receiving an IAppsUser.

apps/meteor/app/apps/server/bridges/persistence.ts
```diff
@@ -21,7 +20,8 @@ export class AppPersistenceBridge extends PersistenceBridge {
 			throw new Error('Attempted to store an invalid data type, it must be an object.');
 		}
 
-		return this.orch.getPersistenceModel().insertOne({ appId, data });
+		// #TODO: #AppsEngineTypes - Remove explicit types and typecasts once the apps-engine definition/implementation mismatch is fixed.
+		return this.orch.getPersistenceModel().insertOne({ appId, data }) as Promise<unknown> as Promise<string>;
```
pierre-lehnen-rc, line 24: cast from InsertOneResult to string.

apps/meteor/app/apps/server/bridges/persistence.ts
```diff
@@ -35,7 +35,8 @@ export class AppPersistenceBridge extends PersistenceBridge {
 			throw new Error('Attempted to store an invalid data type, it must be an object.');
 		}
 
-		return this.orch.getPersistenceModel().insertOne({ appId, associations, data });
+		// #TODO: #AppsEngineTypes - Remove explicit types and typecasts once the apps-engine definition/implementation mismatch is fixed.
+		return this.orch.getPersistenceModel().insertOne({ appId, associations, data }) as Promise<unknown> as Promise<string>;
```
pierre-lehnen-rc, line 39: cast from InsertOneResult to string.

apps/meteor/app/apps/server/bridges/persistence.ts
```diff
@@ -125,6 +126,7 @@ export class AppPersistenceBridge extends PersistenceBridge {
 			associations,
 		};
 
-		return this.orch.getPersistenceModel().update(query, { $set: { data } }, { upsert });
+		// #TODO: #AppsEngineTypes - Remove explicit types and typecasts once the apps-engine definition/implementation mismatch is fixed.
+		return this.orch.getPersistenceModel().update(query, { $set: { data } }, { upsert }) as Promise<unknown> as Promise<string>;
```
pierre-lehnen-rc, line 130: cast from UpdateResult to string.

pierre-lehnen-rc (commented):
Commenting some stuff that are probably bugs in the current implementation

apps/meteor/app/apps/server/bridges/messages.ts
d-gubert: Yes, looks like a potential bug. It might work now due to both types sharing some properties, but that's just luck.

apps/meteor/app/apps/server/bridges/persistence.ts
d-gubert: Definitely a problem. I'm guessing at some point `insertOne` returned just the inserted id? Either that or another function was being used here and changed.

apps/meteor/app/apps/server/bridges/persistence.ts
d-gubert: https://github.com/RocketChat/Rocket.Chat/pull/31757/files/3238d4dee1e327cc1cc960092cfdc27d6b800390#r1492815143

ggazzo (approved):

PR #31761, opened 2024-02-14
fix: UI fixed, link also visible along with preview of the link 

Aadarshprajapati777:
This pull request addresses formatting and visibility issues in the OEmbedPreviewContent  component. Specifically, it ensures that all content, including the URL and associated metadata, is displayed properly and in a visually appealing manner. Additionally, adjustments have been made to the layout to improve spacing and alignment for better readability.
Now, link of the preview image also showing along with the preview 

Before Fix

Screenshot from 2024-02-15 02-21-06

After fix

Screenshot from 2024-02-15 02-09-27
Screenshot from 2024-02-15 02-06-55


jsdgdo:
Hey @Aadarshprajapati777 , we appreciate your dedication on improving Rocket.Chat. While this is certainly a way to go, it's a tad unnecessary. The URL is already being displayed in the body of the message (and I know it could also be a markdown link), but displaying it in the embed only adds visual clutter and deters from the main elements of the link: the video title, and where it's from. We think the current design works well for us, it works not only for video but for all links. 

Thanks again!

gabriellsh:
Thanks for your contribution @Aadarshprajapati777 ! I understand where you're coming from with this, but as pointed out in the previous comment, we'll keep the current design for now.

Also, when opening this kind of PR, you should not put `fix:` in the title, since the code is not fixing an issue, but improving/introducing a new feature. This helps us filter and categorize PRs better.

I'll be closing this PR due to the above reason. Thanks for your comprehension. Please, don't let this discourage you from contributing, we really appreciate your effort.

Aadarshprajapati777:
hey @jsdgdo , the current design is good but if we sent multiple links then it becomes very difficult to find the content related to that link, if the link is also visible along with the content preview then it becomes easier to search

Issue #31749, opened 2024-02-14
Third-Party Login only works for user with "Manage Oauth Apps"-permission.


mbreslein-thd:
### Description:

After I installed the Update to 6.6.0 (from 6.4.8) users without the "Manage Oauth Apps"-permission can't Log into apps that use RocketChat as Oauth povider. Users without that permission get this error: 
![image](https://github.com/RocketChat/Rocket.Chat/assets/89986535/c90e9ff2-a63d-488d-b347-8ef422463e2e)

In the Logs, this message appears: 
```json
{
"level":35,
"time":"2024-02-14T10:56:11.856Z",
"pid":4621,
"hostname":"[redacted]",
"name":"API",
"method":"GET",
"url":"/api/v1/oauth-apps.get?clientId=[redacted]",
"userId":"[redacted]",
"userAgent":"Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/17.3 Safari/605.1.15",
"host":"[redacted]",
"referer":"https://[redacted]/oauth/authorize?response_type=code&client_id=[redacted]&redirect_uri=[redacted]",
"remoteIP":"[redacted]",
"status":403,
"responseTime":1
} 
```
With a user that has the permission, the status-code is 200 instead.

### Steps to reproduce:

1. Set up a third-party app under Administration->Workspace->Third-party login
2. Try to login in the app with a user that doesn't have the "Manage Oauth Apps"-permission. -> won't work.
3. Configure the user to have that permission
4. Try logging in again -> works.

### Expected behavior:

What happened in 6.4.8 and previously. Users can log in using oauth without that permission.

### Actual behavior:

Users need to have the permission to manage oauth apps, which isn't something they should be able to.

### Server Setup Information:

- Version of Rocket.Chat Server: 6.6.0
- Operating System: Ubuntu 20.04.6 LTS
- Deployment Method: "Deploy with ubuntu"
- Number of Running Instances: 1
- NodeJS Version: 14.18.3
- MongoDB Version: 4.4.25

### Client Setup Information

- Desktop App or Browser Version: any
- Operating System: any

### Relevant logs:
see description.

mbreslein-thd:
This line seems to be the culprit: https://github.com/RocketChat/Rocket.Chat/blob/develop/apps/meteor/app/api/server/v1/oauthapps.ts#L30

[...]

PR #31771, opened 2024-02-16
fix: login with third-party apps wasn't working without "Manage Oauth Apps"-permission

mbreslein-thd:
This change was made in https://github.com/RocketChat/Rocket.Chat/commit/5bb039037015d7d4cd3dcd255ac89e63dbe5c84c and broke third-party apps. (see #31749)

Make third-party apps work again without giving users permission to edit/add them.

## Issue(s)

#31749


AndrewChau:
Hi @sampaiodiego , when can this fix be merged?

apps/meteor/app/api/server/v1/oauthapps.ts
```diff
@@ -27,10 +27,6 @@ API.v1.addRoute(
 	{ authRequired: true, validateParams: isOauthAppsGetParams },
 	{
 		async get() {
-			if (!(await hasPermissionAsync(this.userId, 'manage-oauth-apps'))) {
```
Gustrb, line 30: I don't think that removing the permission altogether is a good idea
we should probably just have a different one for managing oauth apps and actually fetching them

.changeset/spotty-dingos-push.md
```diff
@@ -0,0 +1,5 @@
+---
+"@rocket.chat/meteor": patch
+---
+
+fix: login with third-party apps wasn't working without "Manage Oauth Apps"-permission
```
Gustrb, line 5: please reword the changeset so it is not just the commit text, you couldve written something like:
"Fixed an issue where users were not able to login using OAuth without having the 'manage-oauth-apps' permissions" or something like that

Gustrb:
is it possible to add some tests validating that the login with third party apps is working even without the 'manage-oauth-apps' permission?

apps/meteor/app/api/server/v1/oauthapps.ts
mbreslein-thd: It's been without any permission checks forever before this change was done.
I just reverted that change to get it back to a working state with minimal effort.

I'm proposing to fix it the easy way now and add an additional oauth-fetch-permission later.

mbreslein-thd:
> is it possible to add some tests validating that the login with third party apps is working even without the 'manage-oauth-apps' permission?

It would probably be a good idea to add that check where the old one was:
https://github.com/RocketChat/Rocket.Chat/pull/31771/files#diff-54d2e43cb32ca66b463865e1af3547e2b1d5fdf9460c9c2153725b10e962167eL78

I don't have any experience with the used testing-software, so it's probably better if someone else does that.

apps/meteor/app/api/server/v1/oauthapps.ts
Gustrb: yea yea I see, but I don't believe this is a good approach overall, this route should be guarded, if it wasn't in the past it probably was a mistake.
but I believe the permission the see an app should not be the same as the permission to manage it though.

I think I can try to push this PR through if we add a new permission, I can help you on that if you need it

Gustrb:
> > is it possible to add some tests validating that the login with third party apps is working even without the 'manage-oauth-apps' permission?
> 
> It would probably be a good idea to add that check where the old one was: https://github.com/RocketChat/Rocket.Chat/pull/31771/files#diff-54d2e43cb32ca66b463865e1af3547e2b1d5fdf9460c9c2153725b10e962167eL78
> 
> I don't have any experience with the used testing-software, so it's probably better if someone else does that.

About the testing, I took a closer look, and, altough I think we should add a testing flow to validate the login is working properly, it would take a lot of effort and I believe we don't even have the testing capabilities to test that now.
So I think we can just add the permission to fetch OAuth apps and keep the old tests (just changing the permission name)

apps/meteor/app/api/server/v1/oauthapps.ts
mbreslein-thd: If you can add that permission for me I'd be happy to add the check back in with the new permission.

Gustrb:
Just pushed the new permission to the branch, please update the tests :P

mbreslein-thd:
> Just pushed the new permission to the branch, please update the tests :P

Done: https://github.com/RocketChat/Rocket.Chat/pull/31771/commits/ffd96c382f3aacb3f9b8b5ff8da35b8c976a8343

Gustrb (approved):

apps/meteor/app/authorization/server/constant/permissions.ts
```diff
@@ -56,6 +56,7 @@ export const permissions = [
 	{ _id: 'manage-own-outgoing-integrations', roles: ['admin'] },
 	{ _id: 'manage-own-incoming-integrations', roles: ['admin'] },
 	{ _id: 'manage-oauth-apps', roles: ['admin'] },
+	{ _id: 'view-oauth-apps', roles: ['admin', 'owner', 'moderator', 'user'] },
```
matheusbsilva137, line 59:
```suggestion
	{ _id: 'view-oauth-apps', roles: ['admin', 'user'] },
```
Since this new permission is meant for global scope(and not rooms scope), we should keep only roles that fit this scope

.changeset/spotty-dingos-push.md
```diff
@@ -0,0 +1,5 @@
+---
+"@rocket.chat/meteor": patch
+---
+
+Fixed an issue where users were not able to login using OAuth without having the 'manage-oauth-apps' permissions
```
matheusbsilva137, line 5:
```suggestion
Fixed an issue where users were not able to login using OAuth Apps without having the 'manage-oauth-apps' permissions
```

packages/i18n/src/locales/en.i18n.json
```diff
@@ -3407,6 +3407,8 @@
   "manage-livechat-units_description": "Permission to manage omnichannel units",
   "manage-oauth-apps": "Manage OAuth Apps",
   "manage-oauth-apps_description": "Permission to manage the server OAuth apps",
+  "view-oauth-apps_description": "Permission to view the server OAuth apps",
+  "view-oauth-apps": "View OAuth Apps",
```
matheusbsilva137, line 3411: I'm wondering if we can have another name for this setting so as to make it clear that this is needed for users to login. I'll check this internally and bring some suggestion later

matheusbsilva137 (changes requested):

.changeset/spotty-dingos-push.md
Gustrb: I also think we should also note the new permission here on the changeset as well

packages/i18n/src/locales/en.i18n.json
Gustrb: Any suggestions @matheusbsilva137 ?

matheusbsilva137:
Hey @mbreslein-thd  !
I brought the solution proposed by this PR for the consideration of our team (internally) and we considered that we must have a restriction regarding the info returned by the endpoint, since some of the info returned by it are sensitive. Due to this decision, I created #32986, which we'll use to fix this issue
Thanks for contributing and for bringing this issue to us. Please check #32986 to be aware of the next steps regarding this fix (we'll push this so that this fix can be included in 6.12)

PR #31788, opened 2024-02-19
fix: some apps-engine bridges are receiving incompatible data

pierre-lehnen-rc:


apps/meteor/app/apps/server/bridges/persistence.ts
```diff
@@ -21,7 +22,10 @@ export class AppPersistenceBridge extends PersistenceBridge {
 			throw new Error('Attempted to store an invalid data type, it must be an object.');
 		}
 
-		return this.orch.getPersistenceModel().insertOne({ appId, data });
+		return this.orch
+			.getPersistenceModel()
+			.insertOne({ appId, data })
+			.then(({ insertedId }: InsertOneResult) => insertedId || appId);
```
d-gubert, line 28: I'm wondering if returning `appId` in this situation is ideal. This could mislead the app in the sense that it might try to use that ID to remove the record later on - with that in mind I'd think returning an empty string would do less harm. But in any case the `appId` is not going to be a relevant return value in this context.

Is there an instance where the `insertedId` will not be available? Are we forcing the mongo server to assign `_id` values instead of the driver?

d-gubert (changes requested):

d-gubert (approved):

PR #31807, opened 2024-02-21
fix: Remove username from muted/unmuted list on user removal from room

yash-rajpal:


matheusbsilva137 (changes requested):
LGTM, but can you please add some end-to-end tests to cover this? My suggestion would be to check the `channels.info` and `groups.info` endpoints after removing someone from a room.

PR #31811, opened 2024-02-22
chore: Deprecate mute/unmute meteor methods and convert to endpoint

yash-rajpal:
## Further comments

CORE-92


apps/meteor/tests/end-to-end/api/09-rooms.js
```diff
@@ -1707,4 +1707,131 @@ describe('[Rooms]', function () {
...
 	});
+
+	describe('/rooms.muteUser', () => {
+		let testChannel;
+
+		before('create a channel', async () => {
+			const result = await createRoom({ type: 'c', name: `channel.test.${Date.now()}-${Math.random()}` });
+			testChannel = result.body.channel;
+		});
+
```
Gustrb, line 1718: I think we should delete the resources after so there is no issue with the order of the tests

```javascript
after(() => deleteRoom({ type: 'c', roomId: testChannel._id }));
```

apps/meteor/tests/end-to-end/api/09-rooms.js
```diff
@@ -1707,4 +1707,131 @@ describe('[Rooms]', function () {
...
+		it('should invite rocket.cat user to room', async () => {
```
Gustrb, line 1719: those tests don't have to be async, a promise is returned from the .expect chain already

apps/meteor/tests/end-to-end/api/09-rooms.js
```diff
@@ -1707,4 +1707,131 @@ describe('[Rooms]', function () {
...
+				.send({
+					rid: testChannel._id,
+					readOnly: true,
+				})
+				.expect('Content-Type', 'application/json')
+				.expect(200)
+				.expect((res) => {
+					expect(res.body).to.have.property('success', true);
+				});
+
```
Gustrb, line 1788: Here we can delete the room as well

```
after(() => deleteRoom({ type: 'c', roomId: testChannel._id }));
```

apps/meteor/tests/end-to-end/api/09-rooms.js
```diff
@@ -1707,4 +1707,131 @@ describe('[Rooms]', function () {
...
+					username: 'rocket.cat',
+				})
+				.expect('Content-Type', 'application/json')
+				.expect(200)
+				.expect((res) => {
+					expect(res.body).to.have.property('success', true);
+					expect(res.body).to.have.nested.property('channel.name', testChannel.name);
+				});
+		});
+
```
Gustrb, line 1803: Here we can delete the room as well

```suggestion
after(() => deleteRoom({ type: 'c', roomId: testChannel._id }));
```

apps/meteor/tests/end-to-end/api/09-rooms.js
```diff
@@ -1707,4 +1707,131 @@ describe('[Rooms]', function () {
...
+	});
+
+	describe('/rooms.unmuteUser', () => {
+		let testChannel;
+
+		before('create a channel', async () => {
+			const result = await createRoom({ type: 'c', name: `channel.test.${Date.now()}-${Math.random()}` });
+			testChannel = result.body.channel;
+
+			await request
```
Gustrb, line 1776: Can't both of those requests be wrapped in a Promise.all?

There should be a small speed gain by not doing them synchronously

apps/meteor/tests/end-to-end/api/09-rooms.js
yash-rajpal: Nice catch!

matheusbsilva137 (changes requested):

apps/meteor/tests/end-to-end/api/09-rooms.js
```diff
@@ -1707,4 +1707,144 @@ describe('[Rooms]', function () {
...
+					expect(res.body).to.have.property('success', true);
+					expect(res.body).to.have.nested.property('channel.name', testChannel.name);
+				});
+		});
+
+		after(async () => {
+			await deleteRoom({ type: 'c', roomId: testChannel._id });
+		});
+
+		it('should unmute the rocket.cat user in read-only room', (done) => {
```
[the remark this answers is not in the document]
Gustrb: I gave the suggestion, but I was not clear on what I was thinking
I mean earlier just to take the async and return the promise instead of awaiting for it
(not sure it matters a lot though)

apps/meteor/server/methods/muteUserInRoom.ts
```diff
@@ -80,6 +81,8 @@ export const muteUserInRoom = async (fromId: string, data: { rid: IRoom['_id'];
 
 Meteor.methods<ServerMethods>({
 	async muteUserInRoom(data) {
+		methodDeprecationLogger.method('muteUserInRoom', '7.0.0');
```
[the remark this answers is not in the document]
MarcosSpessatto: Yup, you are right, the idea is to give the users at least one major version of warnings before removing stuff. So, we should use 8.0.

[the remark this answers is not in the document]
yash-rajpal: Yep, will shift warning for 8.0.0

apps/meteor/tests/end-to-end/api/09-rooms.js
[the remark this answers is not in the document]
yash-rajpal: Makes sense, will revert this commit. No worries Gustavo. :)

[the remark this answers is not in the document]
matheusbsilva137: Ohhhh got it now. You're right, we could keep functions `async` but just not await what's inside them when there's only one request. Looks cleaner

matheusbsilva137 (approved):

MarcosSpessatto (dismissed):

sampaiodiego:
since 7.0 was delayed, should we target 7.0 to remove the Meteor methods? 😬

matheusbsilva137:
> since 7.0 was delayed, should we target 7.0 to remove the Meteor methods? 😬

Wouldn't it still be too close to deprecate them on 6.8 and remove them on the next version?

sampaiodiego:
> > since 7.0 was delayed, should we target 7.0 to remove the Meteor methods? 😬
> 
> Wouldn't it still be too close to deprecate them on 6.8 and remove them on the next version?

yes, a little bit.. BUT these are Meteor methods (or the realtime API) which is something we want to get rid of asap

ggazzo (approved):

ggazzo (approved):

sampaiodiego (approved):

PR #31822, opened 2024-02-23
fix: Remove users upsert behavior

matheusbsilva137:

 - Remove user upsert behavior from `users.update` endpoint (by making `_id` field required);
 - Deprecate `insertOrUpdateUser` Meteor method;
 - Fix `users.update` and `users.create` params check to accept the `settings` field, which was already expected by them internally (this allow users to set their language preference, for example);
 - Fix `users.update` params check to accept the `language` field, which was already expected by its internal flow.

## Further comments

[CORE-79](https://rocketchat.atlassian.net/browse/CORE-79?atlOrigin=eyJpIjoiN2QwYmYyNGEyYzMwNDAzMmE5OWUzMWM0YjQ2MDc3MjUiLCJwIjoiaiJ9)

[CORE-79]: https://rocketchat.atlassian.net/browse/CORE-79?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.changeset/popular-goats-teach.md
+++ b/.changeset/popular-goats-teach.md
@@ -1,0 +1,7 @@
+---
+"@rocket.chat/meteor": minor
+"@rocket.chat/core-typings": minor
+"@rocket.chat/rest-typings": minor
+---
+
+Added support to configure the `settings` field on `users.update` and `users.create` endpoints
--- a/.changeset/popular-jars-drum.md
+++ b/.changeset/popular-jars-drum.md
@@ -1,0 +1,5 @@
+---
+"@rocket.chat/meteor": patch
+---
+
+Deprecate `insertOrUpdateUser` Meteor method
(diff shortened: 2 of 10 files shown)


MarcosSpessatto (changes requested):
I'd suggest creating different PRs for each of the purposes, one `fix` for each fix we are doing here ("accept the `settings` field" and another one for "accept the language field").

Then we should create a PR for the deprecation method (`insertOrUpdateUser`), and finally another PR for the "removal of upsert behavior".

matheusbsilva137:
Thanks for the suggestion, @MarcosSpessatto 
Closing this in favor of #31889 , #31888 , #31886 and #31884

PR #31833, opened 2024-02-26
fix: client ignoring offline presence updates

sampaiodiego:


apps/meteor/app/notifications/client/lib/Presence.ts
```diff
@@ -13,8 +13,5 @@ type args = [username: string, statusChanged?: UserStatus, statusText?: string];
 export const STATUS_MAP = [UserStatus.OFFLINE, UserStatus.ONLINE, UserStatus.AWAY, UserStatus.BUSY, UserStatus.DISABLED];
 
 Meteor.StreamerCentral.on('stream-user-presence', (uid: string, [username, statusChanged, statusText]: args) => {
-	if (!statusChanged) {
-		return;
-	}
```
debdutdeb, line 18: what is statusChanged? int? the indexing on the STATUS_MAP says that but typing says it's basically a string? the diff makes more sense if is an int?

debdutdeb (approved):

apps/meteor/app/notifications/client/lib/Presence.ts
debdutdeb: I wasn't looking at HEAD :P

PR #31840, opened 2024-02-27
feat: Encrypt file descriptions on E2EE rooms

yash-rajpal:

Encrypt the attachment descriptions at the time of uploading the attachment, and then saving the message as E2EE pending, so that this message and file descriptions can be decrypted upon receiving.
## Steps to test or reproduce

1. Configure E2EE
2. Create an E2EE room
3. Send any file attachment with description, file descriptions should be encrypted

Incase, E2EE is not setup, e2ee placeholder text will be shown in place of encrypted file descriptions.

## Further comments

[E2EE-2]

[E2EE-2]: https://rocketchat.atlassian.net/browse/E2EE-2?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.changeset/great-moles-rest.md
```diff
@@ -0,0 +1,5 @@
+---
+'@rocket.chat/meteor': patch
```
MarcosSpessatto, line 2:
```suggestion
'@rocket.chat/meteor': minor
```

apps/meteor/client/lib/chats/uploads.ts
```diff
@@ -55,14 +56,27 @@ const send = async (
 	]);
 
 	try {
+		// Assuming that only description is available in case of attachments
+		const message = description
+			? await onClientBeforeSendMessage({
+					_id: id,
+					msg: description,
```
MarcosSpessatto, line 63: I don't think we should do this, it would be clearer if we add the support for encrypting file descriptions in the functions instead of using the `msg` field instead.

apps/meteor/client/components/message/variants/room/RoomMessageContent.tsx
```diff
@@ -56,15 +58,17 @@ const RoomMessageContent = ({ message, unread, all, mention, searchText }: RoomM
...
 				</>
 			)}
 
 			{normalizedMessage.blocks && (
 				<UiKitMessageBlock rid={normalizedMessage.rid} mid={normalizedMessage._id} blocks={normalizedMessage.blocks} />
 			)}
 
-			{!!normalizedMessage?.attachments?.length && <Attachments id={message.files?.[0]._id} attachments={normalizedMessage.attachments} />}
+			{!!normalizedMessage?.attachments?.length && (
+				<Attachments id={message.files?.[0]._id} attachments={normalizedMessage.attachments} isMessageEncrypted={isMessageEncrypted} />
```
MarcosSpessatto, line 70:
```suggestion
				<Attachments id={message.files?.[0]?._id} attachments={normalizedMessage.attachments} isMessageEncrypted={isMessageEncrypted} />
```

apps/meteor/app/e2e/client/rocketchat.e2e.ts
```diff
@@ -418,15 +418,31 @@ class E2E extends Emitter {
 			return message;
 		}
 
-		const data = await e2eRoom.decrypt(message.msg);
+		const attachmentDescription = message.attachments?.[0]?.description;
```
MarcosSpessatto, line 421: We don't support multiple files yet, but we will support that soon. Do you think it makes sense to prepare this to support that?

apps/meteor/app/e2e/client/rocketchat.e2e.ts
```diff
@@ -418,15 +418,31 @@ class E2E extends Emitter {
...
+
+		const textToDecrypt = attachmentDescription || message.msg;
```
MarcosSpessatto, line 423: I don't think we should decrypt one or the other, I'd say we should keep decrypting the msg.msg AND add the ability to decrypt file descriptions as well.

MarcosSpessatto (changes requested):

MarcosSpessatto (changes requested):
Also, can we add a test case to cover this?

gabriellsh (dismissed):
I think FE looks ok, only concern is the prop drilling for `isMessageEncrypted` prop. There's 2 possibilities that we can think about in this case.

1 - If it is absolutely necessary, then I'd sugges making a context even if only for this 1 property.
2 - If the message is encrypted, attachment will also be encrypted, so we don't need to render the attachment and we just show the encrypted message before entering that whole attachment structure.

Another observation is that we are currently working in transforming attachment descriptions into the `msg` field of a message. Descriptions are on the way of becoming `alt` description of attachments (or title).

Wdyt?

MarcosSpessatto (commented):
@yash-rajpal can we check the typecheck breaking?

apps/meteor/app/file-upload/server/methods/sendFileMessage.ts
```diff
@@ -177,6 +185,8 @@ export const sendFileMessage = async (
 			groupable: Match.Optional(Boolean),
 			msg: Match.Optional(String),
 			tmid: Match.Optional(String),
+			t: Match.Optional(String),
+			e2e: Match.Optional(String),
```
MarcosSpessatto, line 189: Can we create an API test case for this? Just to ensure E2EE properties are being saved correctly.

apps/meteor/tests/e2e/e2e-encryption.spec.ts
MarcosSpessatto, line 1: Can we create some extra test cases:
 - Send a file with an encrypted description in a thread message
 - Try to simulate a message and check if the placeholder is showing: `E2E_message_encrypted_placeholder` (not sure if it's possible through E2E tests though)

apps/meteor/client/components/message/content/attachments/structure/AttachmentDescription.tsx
MarcosSpessatto, line 1: Is it possible to unit-test this component?

MarcosSpessatto (changes requested):

apps/meteor/client/components/message/content/attachments/structure/AttachmentDescription.tsx
yash-rajpal: Sure!

apps/meteor/app/file-upload/server/methods/sendFileMessage.ts
yash-rajpal: Done!

apps/meteor/tests/e2e/e2e-encryption.spec.ts
yash-rajpal: Done!

MarcosSpessatto (approved):

apps/meteor/client/components/message/content/attachments/structure/AttachmentDescription.tsx
MarcosSpessatto: It has been covered by E2E tests instead.

ggazzo (approved):

PR #31844, opened 2024-02-27
fix: make lifecycle methods part of inherited class in federation services

debdutdeb:
- fix: make lifecycle methods part of inherited class in federation services
- chore: add changeset

### DIFF
--- a/apps/meteor/server/services/federation/service.ts
+++ b/apps/meteor/server/services/federation/service.ts
@@ -330,4 +330,12 @@ export class FederationService extends AbstractBaseFederationService implements 
 		await federationService.initialize();
 		return federationService;
 	}
+
+	public async stopped(): Promise<void> {
+		return super.stopped();
+	}
+
+	public async created(): Promise<void> {
+		return super.created()
+	}
 }
--- a/apps/meteor/server/services/federation/service.ts
+++ b/apps/meteor/server/services/federation/service.ts
@@ -336,6 +336,6 @@ export class FederationService extends AbstractBaseFederationService implements 
 	}
 
 	public async created(): Promise<void> {
-		return super.created()
+		return super.created();
 	}
 }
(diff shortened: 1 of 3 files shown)


debdutdeb:
@MarcosSpessatto quickest one i could think, maybe we could just move them without calling super? i didn't see them used anywhere else but just to be sure.

apps/meteor/server/services/federation/service.ts
```diff
@@ -330,4 +330,12 @@ export class FederationService extends AbstractBaseFederationService implements
 		await federationService.initialize();
 		return federationService;
 	}
+
+	public async stopped(): Promise<void> {
+		return super.stopped();
+	}
+
+	public async created(): Promise<void> {
+		return super.created()
```
debdutdeb, line 339:
```suggestion
		return super.created();
```

MarcosSpessatto:
> @MarcosSpessatto quickest one i could think, maybe we could just move them without calling super? i didn't see them used anywhere else but just to be sure.

You are right, we can move them, I don't know why I've put this on the base class 🤦‍♂️

MarcosSpessatto (changes requested):
Can we follow your suggestion and move them lifecycle methods to the service class instead? I think it would be clearer.

debdutdeb:
actually looks like I can't that easily - 
this
https://github.com/RocketChat/Rocket.Chat/blob/9c06e1d6803c59088c1436dc04f0ae734f0e3772/apps/meteor/ee/server/local-services/federation/service.ts#L138
calls this
https://github.com/RocketChat/Rocket.Chat/blob/9c06e1d6803c59088c1436dc04f0ae734f0e3772/apps/meteor/server/services/federation/service.ts#L300
and is then shared with ce federation service.

MarcosSpessatto (approved):

PR #31863, opened 2024-02-29
feat: Setting to disable markdown cache

gabriellsh:

When on, markdown should be processed on the client
## Issue(s)

[CORE-122](https://rocketchat.atlassian.net/browse/CORE-122)
## Further comments

[CORE-122]: https://rocketchat.atlassian.net/browse/CORE-122?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.changeset/nice-comics-type.md
+++ b/.changeset/nice-comics-type.md
@@ -1,0 +1,6 @@
+---
+"@rocket.chat/meteor": patch
+"@rocket.chat/i18n": patch
+---
+
+added a setting to ignore markdown cache on clients.
--- a/.changeset/nice-comics-type.md
+++ b/.changeset/nice-comics-type.md
@@ -1,6 +1,6 @@
 ---
-"@rocket.chat/meteor": patch
-"@rocket.chat/i18n": patch
+"@rocket.chat/meteor": minor
+"@rocket.chat/i18n": minor
 ---
 
 added a setting to ignore markdown cache on clients.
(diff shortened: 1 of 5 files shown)


.changeset/nice-comics-type.md
```diff
@@ -0,0 +1,6 @@
+---
+"@rocket.chat/meteor": patch
+"@rocket.chat/i18n": patch
```
matheusbsilva137, line 3:
```suggestion
"@rocket.chat/meteor": minor
"@rocket.chat/i18n": minor
```
(this should be marked as minor since this PR is not a fix)

matheusbsilva137 (changes requested):

.changeset/nice-comics-type.md
gabriellsh: Thanks I did not pay attention to this. Updated

matheusbsilva137 (approved):

yash-rajpal (approved):

ggazzo (changes requested):
I don't understand why this should be a setting, if the problem will happen I will always want to keep this setting on. So if I will always want this setting active, a setting that is always active is useless. Just like when we activated it, and ignoring the backend fiel,  removing the md field forever (from the back) seems to be the most correct option.

also this will not help any mobile problems (they use the same parser)


The problem lives in the backend and that is where it should be dealt with, this way

PR #31881, opened 2024-03-04
fix: Multiple type of BH making bots not available on opening

KevLehman:
https://rocketchat.atlassian.net/browse/SUP-468


MarcosSpessatto (changes requested):
Would it be possible to add a test case for this?

apps/meteor/tests/end-to-end/api/livechat/19-business-hours.ts
```diff
@@ -308,7 +308,7 @@ describe('LIVECHAT - business hours', function () {
 			defaultBusinessHour = await getDefaultBusinessHour();
 
 			// close default business hour
-			await openOrCloseBusinessHour(defaultBusinessHour, true);
+			await openOrCloseBusinessHour(defaultBusinessHour, false);
```
KevLehman, line 311:
```suggestion
			await openOrCloseBusinessHour(defaultBusinessHour, true);
```

apps/meteor/tests/end-to-end/api/livechat/19-business-hours.ts
```diff
@@ -872,8 +887,8 @@ describe('LIVECHAT - business hours', function () {
 		it('should verify if agent becomes unavailable to take chats when user is activated, if business hour is inactive', async () => {
 			await openOrCloseBusinessHour(defaultBH, false);
 
-			await setUserActiveStatus(agent._id, false);
 			await setUserActiveStatus(agent._id, true);
+			await setUserActiveStatus(agent._id, false);
```
MarcosSpessatto, line 891: Do we need this? Isn't this making the agent inactive instead of active?

apps/meteor/tests/data/livechat/department.ts
```diff
@@ -57,11 +57,23 @@ new Promise((resolve, reject) => {
...
-export const createDepartmentWithAnOnlineAgent = async (): Promise<{department: ILivechatDepartment, agent: {
-	credentials: IUserCredentialsHeader;
+type OnlineAgent = {
 	user: WithRequiredProperty<IUser, 'username'>;
-}}> => {
-    const { user, credentials } = await createAnOnlineAgent();
+	credentials: IUserCredentialsHeader;
+};
+
+export const createDepartmentWithAnOnlineAgent = async (overrideUser?: OnlineAgent): Promise<{department: ILivechatDepartment, agent: OnlineAgent }> => {
```
MarcosSpessatto, line 65: I think it would be better to create another function for this, something like `createDepartmentUsingExistingOnlineAgent`. WDYT?

apps/meteor/tests/end-to-end/api/livechat/19-business-hours.ts
```diff
@@ -321,6 +321,21 @@ describe('LIVECHAT - business hours', function () {
...
+
+			const { department } = await createDepartmentWithAnOnlineAgent({ user: bot, credentials: creds });
+
+			await createCustomBusinessHour([department._id], false);
+
+			const latestAgent: ILivechatAgent = await getMe(creds);
+			expect(latestAgent).to.be.an('object');
+			expect(latestAgent.openBusinessHours).to.be.an('array').of.length(0);
+			expect(latestAgent.statusLivechat).to.be.equal(ILivechatAgentStatus.AVAILABLE);
+		});
```
MarcosSpessatto, line 338: Shouldn't we create a test case with the BH feature off, and then turn it on and check if it's behaving properly? According to what we have in the changeset, it looks like this only happens when the feature is turned on. Or is this already covered with the above test case?

MarcosSpessatto (changes requested):

apps/meteor/tests/end-to-end/api/livechat/19-business-hours.ts
KevLehman, line 891:
```suggestion
			await setUserActiveStatus(agent._id, false);
			await setUserActiveStatus(agent._id, true);
```

MarcosSpessatto (approved):

sampaiodiego (approved):

PR #31883, opened 2024-03-04
fix: generate png thumbnail previews for svg files

abhinavkrin:

Fixed an issue where SVG thumbnails were incorrectly handled as PNGs, causing preview errors. Now, the system correctly identifies and presents the generated PNG thumbnails, allowing for proper SVG image previews.
## Steps to test or reproduce

Outline the steps to test or reproduce the PR here.
1. Go to File Upload > Blocked Media Types, leave the field blank.
2. Upload any .svg image file.
3. The SVG image should now preview correctly without reverting to a retry screen.

## Further comments

[SUP-495](https://rocketchat.atlassian.net/browse/SUP-495)

[SUP-495]: https://rocketchat.atlassian.net/browse/SUP-495?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


MarcosSpessatto (changes requested):
We could add a test case to cover this. We can get the message and see if there is one of the files with the desired extension.

apps/meteor/app/file-upload/server/lib/FileUpload.ts
```diff
@@ -308,9 +308,18 @@ export const FileUpload = {
...
 		const image = await store._store.getReadStream(file._id, file);
 
-		const transformer = sharp().resize({ width, height, fit: 'inside' });
-
-		const result = transformer.toBuffer({ resolveWithObject: true }).then(({ data, info: { width, height } }) => ({ data, width, height }));
+		let transformer = sharp().resize({ width, height, fit: 'inside' });
+
+		if (file.type === 'image/svg+xml') {
+			transformer = transformer.png();
+			file.name = (file.name || '').replace(/\.svg$/i, '.png');
```
debdutdeb, line 315: IMO we shouldn't do this. Just add the extension if you want for clarity. as the extensions mean nothing, we shouldn't touch the original file name either.

apps/meteor/app/file-upload/server/lib/FileUpload.ts
```diff
@@ -308,9 +308,18 @@ export const FileUpload = {
...
+		}
+		const result = transformer.toBuffer({ resolveWithObject: true }).then(({ data, info: { width, height, format } }) => ({
+			data,
+			width,
+			height,
+			thumbFile: { ...file, type: mime.lookup(format) || file?.type || '' } as IUpload,
```
debdutdeb, line 321: Idk what thumbFile exactly is, but if it's the "thumbnail" file, it shouldn't fall back to original file's type. That's confusing. While falling back to empty isn't perfect either, but at least we can visually fallback to the original file's mimetype.

debdutdeb:
Also, I'm seeing a lot of `||`s ..not all just from this pr, but at what point we should just deny the process instead of adding fallback?

for example, what are the chances of a file getting uploaded without a filename, shouldn't we fail the action if there is no name? Then wouldn't have to fallback onto an empty string there.

abhinavkrin:
> Also, I'm seeing a lot of `||`s ..not all just from this pr, but at what point we should just deny the process instead of adding fallback?
> 
> for example, what are the chances of a file getting uploaded without a filename, shouldn't we fail the action if there is no name? Then wouldn't have to fallback onto an empty string there.

It seems we cannot upload a file without a name. So, I can remove the fallback

apps/meteor/app/file-upload/server/lib/FileUpload.ts
```diff
@@ -308,9 +308,20 @@ export const FileUpload = {
...
 
-		const transformer = sharp().resize({ width, height, fit: 'inside' });
+		let transformer = sharp().resize({ width, height, fit: 'inside' });
 
-		const result = transformer.toBuffer({ resolveWithObject: true }).then(({ data, info: { width, height } }) => ({ data, width, height }));
+		if (file.type === 'image/svg+xml') {
+			transformer = transformer.png();
+			if (file.name) {
+				file.name = `${file.name}.png`;
+			}
```
MarcosSpessatto, line 317: Is there any reason we need to change the file name? 🤔 This can lead to undesired behaviors like `filename.svg.png`
Would it be a problem to keep the filename as is?

If this is extremely necessary, let's return a new property then, like `thumbFilename` instead, it would be clearer.

apps/meteor/app/file-upload/server/lib/FileUpload.ts
```diff
@@ -308,9 +308,20 @@ export const FileUpload = {
...
+		}
+		const result = transformer.toBuffer({ resolveWithObject: true }).then(({ data, info: { width, height, format } }) => ({
+			data,
+			width,
+			height,
+			thumbFile: { ...file, type: mime.lookup(format) || '' } as IUpload,
```
MarcosSpessatto, line 323: Same here, I think we should not return the property thumbFile spreading the file inside of it, and change the type.

Following the previous comment, I think we could change the function return type to something like 

```
{
   result,
   thumbFileName,
   thumbFileType,
}
```

apps/meteor/app/file-upload/server/methods/sendFileMessage.ts
```diff
@@ -64,8 +64,8 @@ export const parseFileIntoMessageAttachments = async (
 			attachment.image_preview = await FileUpload.resizeImagePreview(file);
 			const thumbResult = await FileUpload.createImageThumbnail(file);
 			if (thumbResult) {
-				const { data: thumbBuffer, width, height } = thumbResult;
-				const thumbnail = await FileUpload.uploadImageThumbnail(file, thumbBuffer, roomId, user._id);
+				const { data: thumbBuffer, width, height, thumbFile } = thumbResult;
+				const thumbnail = await FileUpload.uploadImageThumbnail(thumbFile, thumbBuffer, roomId, user._id);
```
MarcosSpessatto, line 68: And finally, following my previous comments, here we can change the first parameter to instead of providing an `IUpload` we can provide an object like: `{ thumbFileName, thumbFileType, originalFileId }`.

It will improve the readability since it does not change the file directly inside the functions.

MarcosSpessatto (changes requested):

apps/meteor/app/file-upload/server/lib/FileUpload.ts
abhinavkrin: Hey @MarcosSpessatto  When I download the image on clicking "save image" in chrome, the thumb file is downloaded which will have `.svg` extension. The user computer will try to open this image as an svg file whereas it is a PNG file.
Therefore I attached a .png file.

abhinavkrin: hey @MarcosSpessatto I found that it was not necessary to add extension here. So I have removed  code to change filename.

apps/meteor/app/file-upload/server/methods/sendFileMessage.ts
abhinavkrin: Hey @MarcosSpessatto I have added the required changes.

MarcosSpessatto (approved):

debdutdeb (approved):
code lgtm. didn't test it. :+1:

PR #31888, opened 2024-03-04
feat: Accept `language` field in `users.update` endpoint

matheusbsilva137:

- Accept `language` field in `users.update` endpoint -- this allows users to set language preferences and is already handled internally.

## Further comments

[CORE-79](https://rocketchat.atlassian.net/browse/CORE-79)

[CORE-79]: https://rocketchat.atlassian.net/browse/CORE-79?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/rest-typings/src/v1/users/UsersUpdateParamsPOST.ts
+++ b/packages/rest-typings/src/v1/users/UsersUpdateParamsPOST.ts
@@ -23,6 +23,7 @@ export type UsersUpdateParamsPOST = {
 		verified?: boolean;
 		customFields?: Record<string, unknown>;
 		status?: string;
+		language?: string;
 	};
 	confirmRelinquish?: boolean;
 };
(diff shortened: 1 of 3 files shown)


packages/rest-typings/src/v1/users/UsersUpdateParamsPOST.ts
```diff
@@ -106,6 +107,10 @@ const UsersUpdateParamsPostSchema = {
 					type: 'string',
 					nullable: true,
 				},
+				language: {
+					type: 'string',
```
MarcosSpessatto, line 111: Would it make sense to validate the language instead of accepting any arbitrary value?

PR #31890, opened 2024-03-04
fix: `livechat/visitor` not updating custom fields some times

KevLehman:
https://rocketchat.atlassian.net/browse/SUP-488


apps/meteor/app/livechat/server/api/v1/visitor.ts
```diff
@@ -68,16 +68,41 @@ API.v1.addRoute('livechat/visitor', {
...
+				})
+					.map(async (field) => {
+						const customField = customFields.find((f) => f.key === field._id);
+						if (!customField) {
+							return;
+						}
+
+						const { key, value, overwrite } = customField;
+						// TODO: Change this to Bulk update
+						if (!(await VisitorsRaw.updateLivechatDataByToken(token, key, value, overwrite))) {
```
MarcosSpessatto, line 87: This is returning an `UpdateResult` from the DB, isn't it? If so, this will be always true.

apps/meteor/app/livechat/server/api/v1/visitor.ts
```diff
@@ -68,16 +68,41 @@ API.v1.addRoute('livechat/visitor', {
...
+
+			const processedKeys = await Promise.all(
+				await LivechatCustomField.findByIdsAndScope<Pick<ILivechatCustomField, '_id'>>(keys, 'visitor', {
+					projection: { _id: 1 },
+				})
+					.map(async (field) => {
+						const customField = customFields.find((f) => f.key === field._id);
+						if (!customField) {
+							return;
+						}
```
MarcosSpessatto, line 83: Do we really need this? 🤔 

Is there any case that the values from the db will not match the values provided by the user? I'm asking because we are fetching that by custom field keys, in theory, this will never fail.

To me it would make more sense to check the keys.length against the length of the array returned from the DB, if the keys are greater than the values returned from the DB, then yes, we could assume we have some key in the array, that is not present on the DB.

Maybe I'm missing something.

apps/meteor/app/livechat/server/api/v1/visitor.ts
```diff
@@ -68,16 +68,41 @@ API.v1.addRoute('livechat/visitor', {
...
+			if (processedKeys.length !== keys.length) {
+				LivechatTyped.logger.warn({
+					msg: 'Some custom fields were not processed',
+					visitorId,
+					missingKeys: keys.filter((key) => !processedKeys.includes(key)),
+				});
+			}
+
+			if (errors.length > 0) {
+				throw new Meteor.Error('error-updating-custom-fields', `Error updating custom fields: ${errors.join(', ')}`);
```
MarcosSpessatto, line 105: Can we throw a regular error instead?

MarcosSpessatto (changes requested):

apps/meteor/app/livechat/server/api/v1/visitor.ts
KevLehman: This is for getting the value that the user sent. We fetch the CF from the DB and that will give us the value that's already stored on the DB. Then we find that on the array of values from the user and use that value.

`find` returns T | null so the if is necessary.

apps/meteor/app/livechat/server/api/v1/visitor.ts
KevLehman: Yep, sadly. There's a case where it can return boolean | void iirc. But left this one exactly like this cause I plan to move this whole thing to a single batch update, so left the current behavior.

MarcosSpessatto: I think this can't return `void` unless we provide a callback to the Mongo driver. If this is possible to return void somehow, please update that on the model method.

MarcosSpessatto (changes requested):

apps/meteor/app/livechat/server/api/v1/visitor.ts
KevLehman: Missed on the `void` part. Func returns `Document | Boolean | UpdateResult`

While the type assertions again could work, I plan to refactor this one so any changes to the underlying endpoint won't reflect the new state.

MarcosSpessatto: But refactoring this to a bulk update, what would be considered an error then?

apps/meteor/app/livechat/server/api/v1/visitor.ts
```diff
@@ -68,16 +68,42 @@ API.v1.addRoute('livechat/visitor', {
...
-				}
-				const { key, value, overwrite } = field;
-				if (customField.scope === 'visitor' && !(await VisitorsRaw.updateLivechatDataByToken(token, key, value, overwrite))) {
-					return API.v1.failure();
-				}
+		if (customFields && Array.isArray(customFields) && customFields.length > 0) {
+			const keys = customFields.map((field) => field.key);
+			const errors: string[] = [];
+
+			const processedKeys = await Promise.all(
```
MarcosSpessatto, line 75: Do we need to have this Promise.all?

MarcosSpessatto (changes requested):

apps/meteor/app/livechat/server/api/v1/visitor.ts
KevLehman: Sadly yes because of the .map on the cursor 😬

MarcosSpessatto: But we are already `await`ing it (`await LivechatCustomField.findByIdsAndScope`), there is no need for the Promise.all

MarcosSpessatto (changes requested):

apps/meteor/app/livechat/server/api/v1/visitor.ts
KevLehman: Interesting. I tested and the async map always returned a Promise[] 🤔 

but apparently it works both ways lol

KevLehman: I pushed and rememberd why:
`.map.toArray` returns a `Promise<X>` type, when you await, then you get `X`
In this case, since the `map` is doing an async operation, the result `X` is of type `Promise<string | undefined>[]`.

That means, we do get an array of promises 😬 because of the map. It apparently works if we ignore the types, which is weird. I got the same behavior and everything apparently worked. BUT, TS won't allow it cause it's actually a promise array.

I'll leave it like that for now

apps/meteor/tests/end-to-end/api/livechat/09-visitors.ts
MarcosSpessatto, line 1: Can we add a test case to cover when `overwrite=false`?

MarcosSpessatto (changes requested):

apps/meteor/tests/end-to-end/api/livechat/09-visitors.ts
KevLehman: I mean, I can. That logic didn't change tho, so testing that would be a bit OOS.

MarcosSpessatto (approved):

sampaiodiego (approved):

PR #31892, opened 2024-03-05
fix: wrapped quote message links in the main box instead of the time box

### LINKED ISSUE #31747: [BUG]: Quoted message link is wrapped with time text instead of the entire box.

Arin13-03:
…box#31748

The problem arose from quote links being enclosed within the time box. I addressed this by removing the message link from there and setting the color to 'hint', which is the default for rendering in both cases. Additionally, I enclosed the message link around the entire <AttachmentContent> component to enhance accessibility. I also set the textDecoration to none. As a result, users can now click directly on the main box and be scrolled to the quoted message.

## Issue(s)

Closes #31747

## Further comments

NA

### DIFF
--- a/apps/meteor/client/components/message/content/attachments/QuoteAttachment.tsx
+++ b/apps/meteor/client/components/message/content/attachments/QuoteAttachment.tsx
@@ -42,7 +42,8 @@ export const QuoteAttachment = ({ attachment }: QuoteAttachmentProps): ReactElem
 
 	return (
 		<>
-			<AttachmentContent className={quoteStyles} width='full'>
+			<AttachmentContent className={quoteStyles} width='full'
+			{...(attachment.message_link && { is: 'a', href: attachment.message_link, style: { textDecoration: 'none' } })}>
 				<AttachmentDetails
 					is='blockquote'
 					borderRadius='x2'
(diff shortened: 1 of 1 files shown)


dougfabris:

Closing this PR because it's stale now, we decide to move with this approach #41091

---

PR #31896, opened 2024-03-05
fix: Validate rooms are not taken before processing by queue

KevLehman:
https://rocketchat.atlassian.net/browse/SUP-492


.changeset/strange-jars-invent.md
```diff
@@ -0,0 +1,5 @@
+---
+"@rocket.chat/meteor": patch
+---
+
+fix: Validate rooms are not taken before processing by queue
```
MarcosSpessatto, line 5: Can we explain this more instead of just the PR title?

MarcosSpessatto (dismissed):
Is it possible to add a test to cover this scenario?

KevLehman:
About the test, I don't think so. Queue is used (right now) only on very specific scenarios that are not available through testing easily (one of them is when a chat is received after MAC limit is reached, etc).

The only way to get this to happen is to get a room to an extremely invalid state where the conversation is open, assigned but queued (inquiry) and then the queue to pick up that one, having an agent that's available and can take the chat 😬

.changeset/strange-jars-invent.md
```diff
@@ -0,0 +1,5 @@
+---
+"@rocket.chat/meteor": patch
+---
+
+fix: Validate rooms are not taken before processing by queue. This will prevent an issue that caused a room, that's on an invalid state, to be re-processed by the queue worker, assigning it again to another user despite being already assigned to one. This happens when a room's inquiry gets to an state where it desyncs from the room object. Room is taken & served while inquriy is still queued. This fix will also reconciliate both when something like this happens: whenever the queue picks a chat that's already taken, it will update it's inquiry object to reflect that and avoid processing again.
```
MarcosSpessatto, line 5:
```suggestion
fix: Validate rooms are not taken before processing by queue. This will prevent an issue that caused a room, that's on an invalid state, to be re-processed by the queue worker, assigning it again to another user despite being already assigned to one. This happens when a room's inquiry gets to an state where it desyncs from the room object. Room is taken & served while inquiry is still queued. This fix will also reconciliate both when something like this happens: whenever the queue picks a chat that's already taken, it will update it's inquiry object to reflect that and avoid processing again.
```

MarcosSpessatto (approved):

PR #30236, opened 2023-08-31
feat: New user panel: First phase

rique223:
## Jira epic: [WM-51](https://rocketchat.atlassian.net/browse/WM-51?atlOrigin=eyJpIjoiMjRhOGEyY2Y2MWEzNDQyNzk3Mjk3NDYxZmUwZmEzMmEiLCJwIjoiaiJ9)

Demo video: https://www.loom.com/share/210fcb90a4724ae39d59c8405b645814?sid=c66bd484-935a-403e-aa48-c0bfa7ecbfae

This PR describes the first phase of a complete refactor of the users page by:

- Dividing the page in 4 tabs:
    * All: Lists all users and changes the status column to `Registration status`;
    * Pending: Lists users that either havent been activated by an admin yet or that have been activated but did not login for the first time, has a `Pending action` column that lists which action the given user is pending and added two new buttons, `Activate` so that the admin can quickly activate an user without having to access inner menus and the `Resend welcome email` so that the user can resend the welcome email and remind users that they can login for the first time;
    * Active: Lists the active users (Users that have been activated and logged in for the first time);
    * Deactivated: Lists the deactivatedf users (Users that have logged in for the first time but for some reason where deactivated);

- Implementing a multi select roles filter;

- Refactoring the contextual bar by:
    * Changing the order of the inputs to a more logical one;
    * Adding the reason for joining info when the `Manually activate new users` setting is enabled;

- Refactoring the New user contextual bar by:
    * Changing the order of the inputs to a more logical one;
    * Changing the `verified` toggle label to `Mark email as verified` and adding an info icon that explains the functionality of that toggle;
    * Dividing the `Set random password and send by email` toggle into two radios, one for setting the password randomly and the other to set the password manually;
    * Hidding the password field behind the new set random password radios and only showing it when the `Set manually` radio has been checked;
    * Creating a `Confirm password` input;
    * Implementing the `PasswordVerifier` component in those password fields;
    * Implementing a `Hide additional fields` at the end of the contextual bar;
    * Changing the `Reset` and `Save` buttons to a single `Add user` button;
    * Changing the `invalid email address` error message to `Please ensure that the email address is valid`;
    * Implementing a post-save confirmation page that allows the admin to create more than one user in sequence;
    * Implementing verifications to make sure email related form inputs are disabled when the SMTP is not configured;
    * Implementing a verification to disable the `Mark email as verified` input and to warn the user that email verification is not needed in case the `Require email verification to login` setting is disabled;
[...]

PR #31898, opened 2024-03-05
feat: Create back-end for the new users panel page

rique223:

This pull request introduces the necessary backend elements for the new users page. These include:

1. New endpoints with JSON schema validation for their payloads:

   - **api/v1/users.listByStatus**: Returns a paginated and/or filtered list of users with the following query parameters:
     ```typescript
     status: 'active' | 'deactivated'; // The type of user based on their status to be returned by the endpoint;
     hasLoggedIn: boolean; // If true, only returns users who have logged in for the firs time;
     type: string; // The type of users to be returned, e.g.: 'bot', 'user', 'app', etc;
     roles?: string[]; // A list of user roles; the returned users list will be filtered based on this list
     searchTerm?: string; // A search term; the returned users list will only include users with this term in their email, name, or username
     sort?: string; // The sorting order of the returned users list
     count?: number; // Number of items in the returned users list
     offset?: number; // The index at which the returned users list will start, helping to define the current page for pagination
     ```

   - **api/v1/users.sendWelcomeEmail**: Sends the welcome email template to the provided email address with the following query parameters:
     ```typescript
     email: string; // The email address of the user to receive the welcome message
     ```

2. Created API tests for the aforementioned endpoints.
3. Improved typing for the `users.ts` rest-typings.
## Further comments

Related to: https://github.com/RocketChat/Rocket.Chat/pull/30236
Jira: [WM-51](https://rocketchat.atlassian.net/browse/WM-51)


apps/meteor/app/api/server/v1/users.ts
```diff
@@ -555,6 +559,129 @@ API.v1.addRoute(
 	},
 );
 
+API.v1.addRoute(
+	'users.listByStatus',
+	{
+		authRequired: true,
+		validateParams: isUsersListStatusProps,
+		permissionsRequired: ['view-d-room', 'view-outside-room'],
```
matheusbsilva137, line 567:
```suggestion
		permissionsRequired: ['view-d-room'],
```
I believe only this permission is required for everyone. `view-outside-room` should be checked conditionally based on the `API_Apply_permission_view-outside-room_on_users-list` settings (as in `users.list`), please add this too

apps/meteor/app/api/server/v1/users.ts
```diff
@@ -555,6 +559,129 @@ API.v1.addRoute(
...
+					break;
+				case 'pending':
+					match = {
+						lastLogin: { $exists: false },
+						type: { $nin: ['bot', 'app'] },
+					};
+					projection.reason = 1;
+					break;
+				default:
+					throw new Meteor.Error('invalid-params', 'Invalid status parameter');
```
matheusbsilva137, line 625:
```suggestion
```
If ajv type validation is correct, then you won't need this :thinking:

apps/meteor/app/api/server/v1/users.ts
```diff
@@ -555,6 +559,129 @@ API.v1.addRoute(
...
+					{ username: { $regex: escapeRegExp(searchTerm), $options: 'i' } },
+					{ name: { $regex: escapeRegExp(searchTerm), $options: 'i' } },
+				],
+			};
+
+			if (roles?.length && !roles.includes('all')) {
+				match = {
+					...match,
+					roles: { $in: roles },
+				};
```
matheusbsilva137, line 643:
```suggestion
				match.roles = { $in: roles };
```
This should work, right? Looks cleaner to me

apps/meteor/app/api/server/v1/users.ts
```diff
@@ -555,6 +559,129 @@ API.v1.addRoute(
...
+
+			if (sort?.status) {
+				actualSort.active = sort.status;
+			}
+
+			if (sort?.name) {
+				actualSort.nameInsensitive = sort.name;
+			}
+
+			let match: Filter<IUser>;
```
matheusbsilva137, line 599:
```suggestion
			const match: Filter<IUser> = {};
```
We generally avoid using `let` :grimacing: Please change this to `const` and assign each property individually below (instead of setting the whole object).

apps/meteor/app/api/server/v1/users.ts
```diff
@@ -555,6 +559,129 @@ API.v1.addRoute(
...
+					...match,
+				},
+				{
+					sort: actualSort,
+					skip: offset,
+					limit: count,
+					projection,
+				},
+			);
+			const [users, total] = await Promise.all([cursor.toArray(), totalCount]);
```
matheusbsilva137, line 657: Maybe you can move all of the work done by this endpoint (up until the projection) to a new function that returns `users` and `total`.
Check the implementation of `findUsersToAutocomplete`, maybe you could try something similar

apps/meteor/app/api/server/v1/users.ts
```diff
@@ -555,6 +559,129 @@ API.v1.addRoute(
...
+	'users.listByStatus',
+	{
+		authRequired: true,
+		validateParams: isUsersListStatusProps,
+		permissionsRequired: ['view-d-room', 'view-outside-room'],
+	},
+	{
+		async get() {
+			const { offset, count } = await getPaginationItems(this.queryParams);
+			const { sort, fields } = await this.parseJsonQuery();
```
matheusbsilva137, line 572: The `fields` params is deprecated for all endpoints, so I'm not sure if we should keep adding it to new endpoints :grimacing: 
If you don't plan to use it (which I think is the case), please remove it.

matheusbsilva137 (changes requested):

apps/meteor/app/api/server/v1/users.ts
rique223: Is there a reason for it to be checked specificaly in the body of the function instead of in the permissions array? I remember there was a discussion and at the end it was decided to use it like it is right now.

matheusbsilva137: Yep, using the `permissionsRequired` array is better overall
But I _believe_ you don't mean to make the `view-outside-room` permission mandatory for this endpoint in this case. Reason: by checking the `users.list` endpoint, I found out that the `view-outside-room` permission is only checked conditionally (which I guess that is the behavior you'd like to mirror). 
If you keep it in the array, then this permission would be required even if the setting is turned off 

This is what we do in `users.list`:
```js
if (
	settings.get('API_Apply_permission_view-outside-room_on_users-list') &&
	!(await hasPermissionAsync(this.userId, 'view-outside-room'))
) {
	return API.v1.unauthorized();
}
```

apps/meteor/server/methods/sendWelcomeEmail.ts
```diff
@@ -0,0 +1,49 @@
...
+
+	if (!user) {
+		throw new Meteor.Error('error-invalid-user', 'Invalid user', {
+			method: 'sendWelcomeEmail',
+		});
+	}
+
+	try {
+		let html = '';
+		Mailer.getTemplate('Accounts_UserAddedEmail_Email', (template) => {
```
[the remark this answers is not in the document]
matheusbsilva137: Hmmm got it, so it makes sense to use the same setting. We can create new ones when the invitation flow is implemented

apps/meteor/server/methods/sendWelcomeEmail.ts
```diff
@@ -0,0 +1,49 @@
...
+			html = template;
+		});
+
+		await Mailer.send({
+			to: email,
+			from: settings.get('From_Email'),
+			subject: settings.get('Accounts_UserAddedEmail_Subject'),
+			html,
+		});
```
matheusbsilva137, line 42: You're missing the `data` param in this object. This is used to replace user's `password`, `name` or `username` in the email content
My suggestion is to use the `_sendUserEmail` function [here](https://github.com/RocketChat/Rocket.Chat/blob/7396045d45e57861a076a3e25122cd0295d432b8/apps/meteor/app/lib/server/functions/saveUser.js#L42). Export it and use it here, you won't even need this try/catch

apps/meteor/app/api/server/v1/users.ts
```diff
@@ -555,6 +559,129 @@ API.v1.addRoute(
...
+				count: users.length,
+				offset,
+				total,
+			});
+		},
+	},
+);
+
+API.v1.addRoute(
+	'users.sendWelcomeEmail',
```
matheusbsilva137, line 670: Shouldn't this be protected by some permission? I don't think anyone should be able to send welcome e-mails to other pending users.

matheusbsilva137 (changes requested):

apps/meteor/app/api/server/v1/users.ts
rique223: Makes sense, will fix

apps/meteor/server/methods/sendWelcomeEmail.ts
rique223: Since this is an unforeseen problem in the flow of this feature I'll keep the current implementation and pass this information to product so they can decide.

apps/meteor/app/api/server/lib/users.ts
```diff
@@ -119,3 +119,95 @@ export function getNonEmptyQuery<T extends IUser>(query: Mongo.Query<T> | undefi
 
 	return { ...defaultQuery, ...query };
 }
+
+type findPaginatedUsersByStatusProps = {
```
aleksandernsilva, line 123: Small nitpick, don't we usually capitalize the first letter for types?

aleksandernsilva (changes requested):

apps/meteor/app/api/server/lib/users.ts
rique223: You are correct, it was a little oversight on my side, will fix asap

aleksandernsilva (dismissed):

packages/rest-typings/src/v1/users.ts
```diff
@@ -110,6 +112,11 @@ export type UserPresence = Readonly<
 
 export type UserPersonalTokens = Pick<IPersonalAccessToken, 'name' | 'lastTokenPart' | 'bypassTwoFactor'> & { createdAt: string };
 
+export type PickedUser = Pick<
```
matheusbsilva137, line 115:
```suggestion
export type PickedUser = Pick<
```
Good catch on finding the missing properties here. Can we maybe define it with a more descriptive name? Maybe `UserWithDefaultInfo` or `UserListInfo`

matheusbsilva137 (changes requested):

matheusbsilva137 (approved):
The idea of this PR LGTM, even though this is not the final state we aim for `users.listByStatus` 
The `users.listByStatus` endpoint implemented here is improved in #31987

aleksandernsilva (dismissed):

matheusbsilva137 (changes requested):
Can we bring the BE changes from #31987 related to these new endpoints to this PR?

matheusbsilva137 (approved):

aleksandernsilva (dismissed):

tassoevan (approved):

aleksandernsilva (approved):

PR #31899, opened 2024-03-05
chore: function patches

pierre-lehnen-rc:

A suggestion of one possible way to solve the situation where EE functions need to be called from the FOSS code.

I added support for specifying conditions in which the patch should be used; Alternatively I could add a way to remove patches from the stack instead.


apps/meteor/app/api/server/helpers/getMatrixInstances.ts
```diff
@@ -0,0 +1,4 @@
+import { makeFunction } from '@rocket.chat/patch-coordinator';
+import type { BrokerNode } from 'moleculer';
+
+export const getMatrixInstances = makeFunction(async (): Promise<BrokerNode[]> => []);
```
geekgonecrazy, line 4: why are we calling these matrixInstances?  this naming doesn't make sense ?  Wouldn't this almost literally be getRocketChatInstances?  

I mean I know it was this way already.  But if refactoring and moving it around.  Would be nice if reading it semantically made more sense

pierre-lehnen-rc: changed it.

MarcosSpessatto (approved):

PR #31900, opened 2024-03-05
refactor: Improve queue startup process & reliability

KevLehman:
Related to https://rocketchat.atlassian.net/browse/SUP-492
Can be merged after https://github.com/RocketChat/Rocket.Chat/pull/31896

## Further comments

Note: the goal of this one is to stop using `methodName`.

methodName was being updated on a `watch` call. However, the watch call was happening _after_ the UI already tried to fetch the config for the "new" routing method. At this point, methodName still had the old value, so when the endpoint was hit, it was returning the old routing method config.

By using the settings now, we use the most recent value, preventing these cases from happening.


apps/meteor/app/livechat/server/lib/RoutingManager.ts
```diff
@@ -65,47 +65,27 @@ export const RoutingManager: Routing = {
...
 		if (shouldPreventQueueStart) {
 			logger.error('Monthly Active Contacts limit reached. Queue will not start');
 			return;
 		}
-
 		void (await Omnichannel.getQueueWorker()).shouldStart();
 	},
 
+	isMethodSet() {
+		return settings.get<string>('Livechat_Routing_Method') !== '';
```
MarcosSpessatto, line 79: By changing this we are not using `methodName` anymore. Is this the expected behavior? If so, let's remove that variable. I think here we need to revisit it too: https://github.com/RocketChat/Rocket.Chat/blob/1b50c634c9b035efe899cfedd27c2589bb3bc2f9/apps/meteor/app/livechat/server/lib/RoutingManager.ts#L96

apps/meteor/app/livechat/server/lib/RoutingManager.ts
```diff
@@ -65,47 +65,27 @@ export const RoutingManager: Routing = {
...
 
 	getMethod() {
-		if (!this.methodName) {
-			throw new Meteor.Error('error-routing-method-not-set');
-		}
-		if (!this.methods[this.methodName]) {
-			throw new Meteor.Error('error-routing-method-not-available');
-		}
-		return this.methods[this.methodName];
+		return this.methods[settings.get<string>('Livechat_Routing_Method')];
```
MarcosSpessatto, line 88: I believe we should keep the validation here, what if for some reason the selected method is not registered here?

apps/meteor/server/modules/listeners/listeners.module.ts
```diff
@@ -209,6 +209,10 @@ export class ListenersModule {
 		});
 
 		service.onEvent('watch.inquiries', async ({ clientAction, inquiry, diff }): Promise<void> => {
+			if (settings.get('Livechat_Routing_Method') !== 'Manual_Selection') {
+				return;
+			}
+
```
MarcosSpessatto, line 215: Let's keep this change to another PR, looks unrelated to the purpose of this PR. In the next PR we can explain better what is the purpose of not publishing inquiries if the routing method is equal to 'Manual_Selection'.

apps/meteor/server/services/omnichannel/queue.ts
```diff
@@ -135,6 +147,10 @@ export class OmnichannelQueue implements IOmnichannelQueue {
 		// This is a precaution to avoid taking the same inquiry multiple times. It should not happen, but it's a safety net
 		if (roomFromDb?.servedBy) {
 			queueLogger.debug(`Inquiry ${inquiry._id} already taken by agent ${roomFromDb.servedBy._id}. Skipping`);
+
+			// Reconciliation step: if we find an inquiry that is already taken, but somehow was processed again by the queue
+			// We'll update its status so it doesnt happen again.
+			await LivechatInquiry.takeInquiry(inquiry._id);
```
MarcosSpessatto, line 153: Let's create another PR for this change as well, since it's a behavior change.

MarcosSpessatto (changes requested):

apps/meteor/server/modules/listeners/listeners.module.ts
KevLehman: Hmmm, refactor is for the whole queue. And sending inquiries to clients are also part of the queue.

Not sending them when the algo is not manual selection is a perf improvement for the streaming of inquiries, since when the algo is not manual, we don't even show them on the client.
When the algorithm changes, UI fetches all the inquiries again so the cached collection in the UI is emptied. So all of these events are just sent for nothing.

apps/meteor/server/services/omnichannel/queue.ts
KevLehman: This is part of the refactor 😬 
Behavior change yes, but part of the queue refactor.

MarcosSpessatto (approved):

lmauromb (approved):

PR #31914, opened 2024-03-06
chore: create RoomBody useHandleUnread hook

ggazzo:


gabriellsh (changes requested):
debouncedReadMessageRead wasn't removed from RoomBody. (Line 237)

PR #31918, opened 2024-03-06
feat(UiKit): Channels select

tiagoevanp:

This PR introduces new input components to UiKit:

Multiple Channel Select
Single Channel Select

## Further comments

[CORE-60](https://rocketchat.atlassian.net/browse/CORE-60)

[CORE-60]: https://rocketchat.atlassian.net/browse/CORE-60?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.changeset/strong-humans-bow.md
```diff
@@ -0,0 +1,6 @@
+---
+"@rocket.chat/fuselage-ui-kit": patch
+"@rocket.chat/ui-kit": patch
```
MarcosSpessatto, line 3:
```suggestion
"@rocket.chat/fuselage-ui-kit": minor
"@rocket.chat/ui-kit": minor
```

packages/fuselage-ui-kit/src/elements/ChannelsSelectElement/ChannelsSelectElement.tsx
```diff
@@ -0,0 +1,83 @@
...
+  context,
+}: ChannelsSelectElementProps) => {
+  const [{ value, loading }, action] = useUiKitState(block, context);
+
+  const [filter, setFilter] = useState('');
+  const filterDebounced = useDebouncedValue(filter, 300);
+
+  const data = useChannelsData({ filter: filterDebounced });
+
+  const options = useMemo(
```
gabriellsh, line 24: This should be either calculated in the useQuery hook fetch function, or passed as a funtion to `select` property of useQuery.

packages/fuselage-ui-kit/src/elements/ChannelsSelectElement/MultiChannelsSelectElement.tsx
```diff
@@ -0,0 +1,77 @@
...
+  context,
+}: MultiChannelsSelectProps) => {
+  const [{ value, loading }, action] = useUiKitState(block, context);
+
+  const [filter, setFilter] = useState('');
+  const filterDebounced = useDebouncedValue(filter, 300);
+
+  const data = useChannelsData({ filter: filterDebounced });
+
+  const options = useMemo(
```
gabriellsh, line 24: same case here

packages/fuselage-ui-kit/src/elements/ChannelsSelectElement/hooks/useChannelsData.ts
```diff
@@ -0,0 +1,46 @@
...
+
+  const { data } = useQuery(
+    ['rooms.autocomplete.channelAndPrivate', filter],
+    () => getRooms(generateQuery(filter)),
+    {
+      keepPreviousData: true,
+    }
+  );
+
+  const result = data?.items.map(({ fname, name, _id, avatarETag, t }) => ({
```
gabriellsh, line 38: This map should be inside the fetch funtion or as `select` argument of the useQuery Hook. Btw, why are you mapping it twice? (once here and once more in the components). Why not already map the information in the structure that the elements will consume?

packages/fuselage-ui-kit/src/surfaces/FuselageSurfaceRenderer.tsx
```diff
@@ -528,4 +530,42 @@ export abstract class FuselageSurfaceRenderer extends UiKit.SurfaceRenderer<Reac
...
+    context: UiKit.BlockContext,
+    index: number
+  ): ReactElement | null {
+    if (context === UiKit.BlockContext.FORM) {
+      return (
+        <ChannelsSelectElement
+          block={block}
+          context={context}
+          index={index}
+          surfaceRenderer={this}
```
gabriellsh, line 545: More of a question, but, I see these values are passed to the element, but not used in them. Is there a reason for this?

gabriellsh (changes requested):

packages/fuselage-ui-kit/src/surfaces/FuselageSurfaceRenderer.tsx
tiagoevanp: To be honest, I have the same question as you... I was following the pattern of the type definition itself. If you look at the type definition, there is a necessity to pass them in BlockProps... in some cases like in ActionBlocks surfaceRenderer is used... maybe this is some of the things we need to revalidate in a refactor.

packages/fuselage-ui-kit/src/elements/ChannelsSelectElement/hooks/useChannelsData.ts
gabriellsh: > inside the fetch funtion or as select argument of the useQuery Hook

This still hasn't been implemented. I'll leave this comment here for history purposes

packages/fuselage-ui-kit/src/elements/ChannelsSelectElement/MultiChannelsSelectElement.spec.tsx
```diff
@@ -0,0 +1,132 @@
...
+
+    expect(await screen.findByRole('listbox')).toBeInTheDocument();
+  });
+
+  it('should filter channels', async () => {
+    const input = (await screen.findByRole('textbox')) as HTMLInputElement;
+
+    await userEvent.type(input, 'Channel 2', { delay: null });
+
+    mockUseChannelsData.mockReturnValueOnce(
```
gabriellsh, line 88: This here is a little concerning, because the filter behaviour is defined in the hook. By mocking the return value, it is not a real test of the filter.

IMO we can remove this filter test as it is not really testing anything. One option would be to mock the request response but since the filtering is done on the server, we'd still have the same problem (we're not really testing anything relevant).

gabriellsh (changes requested):

gabriellsh (approved):

MarcosSpessatto (approved):

PR #31925, opened 2024-03-07
fix: call `stopped` when destroying service

debdutdeb:


debdutdeb:
eh nawt ready

debdutdeb:
note https://github.com/RocketChat/Rocket.Chat/blob/70edbb1112fc738e224ac736883103b3c6e858f3/apps/meteor/ee/server/startup/services.ts#L36-L38

PR #31926, opened 2024-03-07
chore: Remove references to EE code from the app events

pierre-lehnen-rc:


sampaiodiego (commented):
how can we make sure there is no code running during the startup **before** `registerOrchestrator` is actually called? I wonder if having `Apps?.something` all over the place can introduce silence bugs where we think it should be defined already but it is not.. what is your take on this?

pierre-lehnen-rc:
> how can we make sure there is no code running during the startup **before** `registerOrchestrator` is actually called? I wonder if having `Apps?.something` all over the place can introduce silence bugs where we think it should be defined already but it is not.. what is your take on this?

I guess I can replace it with a proxy that do something if the orchestrator is not registered, but in theory there won't be any app event triggered before the server is fully loaded.

sampaiodiego (approved):

PR #31958, opened 2024-03-11
fix: Force logout live clients on E2EE keys reset

yash-rajpal:
## Further comments

E2EE-10

### DIFF
--- a/.changeset/eighty-pans-joke.md
+++ b/.changeset/eighty-pans-joke.md
@@ -1,0 +1,5 @@
+---
+'@rocket.chat/meteor': minor
+---
+
+Force logout live clients on E2EE keys reset
--- a/apps/meteor/server/lib/resetUserE2EKey.ts
+++ b/apps/meteor/server/lib/resetUserE2EKey.ts
@@ -3,6 +3,7 @@ import { Meteor } from 'meteor/meteor';
 
 import * as Mailer from '../../app/mailer/server/api';
 import { settings } from '../../app/settings/server';
+import { Notifications } from '../../app/notifications/server';
 import { i18n } from './i18n';
 import { isUserIdFederated } from './isUserIdFederated';
 
@@ -63,6 +64,9 @@ export async function resetUserE2EEncriptionKey(uid: string, notifyUser: boolean
 	if (isUserFederated) {
 		throw new Meteor.Error('error-not-allowed', 'Federated Users cant have TOTP', { function: 'resetTOTP' });
 	}
+
+	// force logout the live sessions
+	Notifications.notifyUser(uid, 'force_logout');
 
 	await Users.resetE2EKey(uid);
 	await Subscriptions.resetUserE2EKey(uid);


apps/meteor/tests/e2e/e2e-encryption.spec.ts
```diff
@@ -212,4 +212,32 @@ test.describe.serial('e2e-encryption', () => {
...
+		await expect(page.locator('role=button[name="Login"]')).toBeVisible();
+		await expect(anotherClientPage.locator('role=button[name="Login"]')).toBeVisible();
+
+		await expect(page.locator('role=banner')).toContainText('Your session was ended on this device, please log in again to continue.');
+		await expect(anotherClientPage.locator('role=banner')).toContainText('Your session was ended on this device, please log in again to continue.');
+
+		await anotherClientPage.close();
+
+		// inject initial data, so that tokens are restored after forced logout
+		await injectInitialData();
```
KevLehman, line 241: Should be on an `after` call (afterall maybe)

yash-rajpal: Makes sense.

.changeset/eighty-pans-joke.md
```diff
@@ -0,0 +1,5 @@
+---
+'@rocket.chat/meteor': minor
+---
+
+Force logout live clients on E2EE keys reset
```
KevLehman, line 5: Should this be a minor? Isn't it more like a patch?

btw, can you explain a bit more your fix in here? 🤗

yash-rajpal: Done!

packages/i18n/src/locales/en.i18n.json
```diff
@@ -3269,7 +3269,7 @@
   "Log_Trace_Subscriptions_Filter": "Trace subscription filter",
   "Log_Trace_Subscriptions_Filter_Description": "The text here will be evaluated as RegExp (`new RegExp('text')`). Keep it empty to show trace of every call.",
   "Log_View_Limit": "Log View Limit",
-  "Logged_Out_Banner_Text": "Your workspace admin ended your session on this device. Please log in again to continue.",
+  "Logged_Out_Banner_Text": "Your session was ended on this device, please log in again to continue.",
```
MarcosSpessatto, line 3272: We shouldn't change the content of this key, since it's being used in other places like the Device management page where the admin can log out users.

MarcosSpessatto (changes requested):

packages/i18n/src/locales/en.i18n.json
yash-rajpal: This was an intentional change after discussion with design team, idea is that same force_logout banner is also used for different purposes, so we should make this text more generic.

MarcosSpessatto (dismissed):

KevLehman (dismissed):

apps/meteor/server/lib/resetUserE2EKey.ts
```diff
@@ -64,6 +65,9 @@ export async function resetUserE2EEncriptionKey(uid: string, notifyUser: boolean
 		throw new Meteor.Error('error-not-allowed', 'Federated Users cant have TOTP', { function: 'resetTOTP' });
 	}
 
+	// force logout the live sessions
+	Notifications.notifyUser(uid, 'force_logout');
```
GleidsonDaniel, line 69:
```suggestion
	Notifications.notifyUser(uid, 'force-logout');
```
the pattern used in other streams is dash and not underlining.

https://github.com/RocketChat/Rocket.Chat/blob/a7823c1b0901c510af5bfa994e7b3f96ee10dd91/ee/packages/ddp-client/src/legacy/types/SDKLegacy.ts#L77

GleidsonDaniel (changes requested):

apps/meteor/server/lib/resetUserE2EKey.ts
gabriellsh: There's already a stream set with this name.
```
{ key: `${string}/force_logout`; args: [] },
```
it's used here: apps/meteor/client/startup/forceLogout.ts

I think this was introduced together with the Device Management feature. Not sure we should change it now...

gabriellsh (dismissed):
Nice tests!

apps/meteor/server/lib/resetUserE2EKey.ts
yash-rajpal: Oops, I missed your comment Daniel.

Gabriel is right, this is not a new stream event, we can change the stream event but not sure, maybe we should discuss first with architecture team. 🤔

GleidsonDaniel: Ok, I understand that it was done like this in the past, but the value that will come from the stream will look like this:
`stream-force_logout`.
Strange isn't it?

As this is already closed for 6.8, I don't see any problems, but I think it's worth coming back to this later and following the previous pattern.

GleidsonDaniel (approved):

rodrigok (approved):

ggazzo (approved):

PR #31974, opened 2024-03-12
feat: Add contentDisposition option to file storages

jeanfbrito:

Add the option to change `contentDisposition` on the download url for file uploads.
So we can make GridFS and FileSystem work with PDF viewer on desktop.

## Issue(s)

[SUP-516](https://rocketchat.atlassian.net/browse/SUP-516)
[SUP-494](https://rocketchat.atlassian.net/browse/SUP-494)

## Further comments

[SUP-516]: https://rocketchat.atlassian.net/browse/SUP-516?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/meteor/app/file-upload/server/config/GridFS.ts
```diff
@@ -161,7 +161,7 @@ new FileUploadClass({
 	async get(file, req, res) {
 		file = FileUpload.addExtensionTo(file);
 
-		res.setHeader('Content-Disposition', `attachment; filename*=UTF-8''${encodeURIComponent(file.name || '')}`);
+		res.setHeader('Content-Disposition', `inline; filename*=UTF-8''${encodeURIComponent(file.name || '')}`);
```
MarcosSpessatto, line 164: Same as above.

apps/meteor/client/components/message/content/attachments/file/GenericFileAttachment.tsx
```diff
@@ -55,7 +55,12 @@ const GenericFileAttachment = ({
 					<MessageGenericPreviewContent
 						thumb={<MessageGenericPreviewIcon name='attachment-file' type={format || getFileExtension(title)} />}
 					>
-						<MessageGenericPreviewTitle externalUrl={getExternalUrl()} onClick={handleTitleClick} data-qa-type='attachment-title-link'>
+						<MessageGenericPreviewTitle
+							download={!!openDocumentViewer}
+							externalUrl={getExternalUrl()}
+							onClick={handleTitleClick}
+							data-qa-type='attachment-title-link'
+						>
```
MarcosSpessatto, line 63: I would put this change in another PR, specific to this change. So we can explain it separately, using a different changeset for that, so we can keep track of the change.

MarcosSpessatto (changes requested):
I think this is more a `feat` rather than a `fix`, also we need to add a changeset.

.changeset/sweet-kiwis-scream.md
```diff
@@ -0,0 +1,5 @@
+---
+"@rocket.chat/meteor": minor
+---
+
+ In the Desktop App, clicking on a message attachment link will now initiate a direct download of the attachment if not a PDF
```
KevLehman, line 5:
```suggestion
Clicking on a message attachment link on the Desktop App will now initiate a direct download of the attachment only when the attachment is not a PDF file
```

apps/meteor/app/file-upload/server/config/AmazonS3.ts
```diff
@@ -11,7 +11,17 @@ import '../../ufs/AmazonS3/server';
 
 const get: FileUploadClass['get'] = async function (this: FileUploadClass, file, req, res) {
 	const { query } = URL.parse(req.url || '', true);
-	const forceDownload = typeof query.download !== 'undefined';
+	let forceDownload = typeof query.download !== 'undefined';
+	if (!forceDownload) {
+		switch (query.contentDisposition) {
```
KevLehman, line 16: When query.contentDisposition is none (inline, attachment) forceDownload will be false. If that's fine, ignore this comment, otherwise add a default case.

apps/meteor/app/file-upload/server/config/GridFS.ts
```diff
@@ -159,9 +160,13 @@ new FileUploadClass({
 	name: 'GridFS:Uploads',
 
 	async get(file, req, res) {
+		const { query } = URL.parse(req.url || '', true);
 		file = FileUpload.addExtensionTo(file);
+		const contentDisposition = query.contentDisposition === 'inline' ? 'inline' : 'attachment';
 
-		res.setHeader('Content-Disposition', `attachment; filename*=UTF-8''${encodeURIComponent(file.name || '')}`);
+		console.log('query.contentDisposition', query.contentDisposition);
```
KevLehman, line 167:
```suggestion
```

apps/meteor/app/file-upload/server/config/GoogleStorage.ts
```diff
@@ -10,7 +10,17 @@ import '../../ufs/GoogleStorage/server';
 
 const get: FileUploadClass['get'] = async function (this: FileUploadClass, file, req, res) {
 	const { query } = URL.parse(req.url || '', true);
-	const forceDownload = typeof query.download !== 'undefined';
+	let forceDownload = typeof query.download !== 'undefined';
```
KevLehman, line 13: wdyt of making this logic a lib since we're using the exact same as AmazonS3? 👀

KevLehman (changes requested):

apps/meteor/app/file-upload/server/config/AmazonS3.ts
jeanfbrito: Its fine because its the way it works today.

apps/meteor/app/file-upload/server/config/GoogleStorage.ts
jeanfbrito: Dont know how to do that and dont know if is necessary to just two use cases.

MarcosSpessatto: I agree with @KevLehman, it doesn't need to be a lib, just a single function to make sure we have the logic in a single place. You can create a new file under `file-upload/server/config/helper.ts` or something like that.

apps/meteor/app/file-upload/server/config/GridFS.ts
```diff
@@ -159,9 +160,11 @@ new FileUploadClass({
 	name: 'GridFS:Uploads',
 
 	async get(file, req, res) {
+		const { query } = URL.parse(req.url || '', true);
 		file = FileUpload.addExtensionTo(file);
+		const contentDisposition = query.contentDisposition === 'inline' ? 'inline' : 'attachment';
 
-		res.setHeader('Content-Disposition', `attachment; filename*=UTF-8''${encodeURIComponent(file.name || '')}`);
+		res.setHeader('Content-Disposition', `${contentDisposition}; filename*=UTF-8''${encodeURIComponent(file.name || '')}`);
```
MarcosSpessatto, line 167: I think we can use the same idea as before here, and create a shared function like `getContentDisposition()`. Just to have the shared logic in a single place.

MarcosSpessatto (changes requested):
Also, tests will be delivered in a subsequent PR. Task: https://rocketchat.atlassian.net/browse/CORE-457

apps/meteor/app/file-upload/server/config/helper.ts
```diff
@@ -0,0 +1,26 @@
...
+			case 'attachment':
+				forceDownload = true;
+				break;
+		}
+	}
+	return forceDownload;
+};
+
+export const getContentDisposition = (req: http.IncomingMessage): string => {
+	const { query } = URL.parse(req.url || '', true) as any;
```
KevLehman, line 21:
```suggestion
	const { query } = URL.parse(req.url || '', true);
```

KevLehman (changes requested):

KevLehman (approved):

sampaiodiego (approved):

MarcosSpessatto (approved):

KevLehman (approved):

PR #31985, opened 2024-03-13
regression: Livechat resubscription

MartinSchoeler:
Fixes Issue with Livechat not re subscribing to streams after closing the chat, every time it was trying to subscribe using the same token every time

introduced here https://github.com/RocketChat/Rocket.Chat/pull/31919


MartinSchoeler:
As expected the tests are failing without the fix:

image

PR #31986, opened 2024-03-13
feat: Implement Cloud communication reliability 

Gustrb:
When a Workspace Registers with Cloud it gets a set of credentials it uses to generate access tokens and communicate with Cloud. Can read more about how this works in [ADR3](https://adr.rocket.chat/0003).

When a workspace owner needs to scale and deploys multiple Rocket.Chat instances to make up their deployment. Each workspace establishes a change stream request to mongo watching settings. Then updates its collection cache locally.

When a workspace goes to use an access token to talk to a cloud service it first evaluates if the token needs renewed. If it does it request a new one. Then updates a setting containing the access token. Then also another setting containing the expire.

The problem occures when latency enters the picture. If due to either longer mongo query or network condition the other workspaces don't see both of these changes happen and update its cache. An instance can be using an expired token instead of renewing because the setting cache hasn't been updated.

We've seen many cases of this happen and typically restarting workspace can resolve.

This PR moves all credential related data to the 'workspace_credentials' collection, and stop using settings.get to fetch the configs that can be changed on multiple server instances

## Steps to test or reproduce

Create a new workspace, on registering, it should create a new record inside the `rocketchat_workspace_credentials` collection, containing an empty date as `expirationDate`, an empty string as `accessToken` and an empty string as `scopes`.
To test everything else, you must use the cloud functions, not sure how to call them from RC, but you could enter the Marketplace screen, go to paid apps, open one of them, and check if there is a new token in the database

## Further comments

This implements: https://adr.rocket.chat/0061

CONN-5


Gustrb:
I think I should write a migration to move the `'workspace_public_key', 'workspace_registration_client_uri', 'workspace_id', 'workspace_name', 'cloud_workspace_access_token', 'cloud_workspace_access_token_expires_at'` values from the Settings collection into the workspace_credentials collection right? @geekgonecrazy

geekgonecrazy:
I think client is and client secret should stay in settings and use the settings get that bypasses cache.  Making the new collection only for access tokens

Gustrb:
Oooh, it makes a lot of sense now, will work on it now, thx

tassoevan (approved):

apps/meteor/ee/server/apps/communication/rest.ts
```diff
@@ -442,7 +455,11 @@ export class AppsRestApi {
 					}
 
 					const baseUrl = orchestrator.getMarketplaceUrl();
-					const workspaceId = settings.get<string>('Cloud_Workspace_Id');
+					const workspaceId = await Settings.getValueById('Cloud_Workspace_Id');
+
+					if (!workspaceId) {
```
ggazzo, line 460: Why do we use 2 different types error?

apps/meteor/ee/server/apps/communication/rest.ts
```diff
@@ -886,15 +903,16 @@ export class AppsRestApi {
 						headers.Authorization = `Bearer ${token}`;
 					}
 
-					const workspaceIdSetting = await Settings.findOneById('Cloud_Workspace_Id');
-					if (!workspaceIdSetting) {
+					const workspaceId = await Settings.getValueById('Cloud_Workspace_Client_Id');
+
+					if (!workspaceId) {
 						return API.v1.failure('No workspace id found');
```
ggazzo, line 909: > Why do we use 2 different types error?


like here

apps/meteor/server/settings/setup-wizard.ts
```diff
@@ -1322,28 +1322,6 @@ export const createSetupWSettings = () =>
...
-				type: 'date',
-				hidden: true,
-				readonly: true,
-				enableQuery: {
-					_id: 'Register_Server',
-					value: true,
-				},
-				secret: true,
-			});
-
```
ggazzo, line 1346: I think we should have a migration to remove this, since these are hidden we can do this in a different pr

apps/meteor/app/cloud/server/functions/removeLicense.ts
```diff
@@ -20,7 +20,8 @@ export async function removeLicense() {
 			throw new CloudWorkspaceAccessTokenEmptyError();
 		}
 
-		const workspaceRegistrationClientUri = settings.get<string>('Cloud_Workspace_Registration_Client_Uri');
+		const workspaceRegistrationClientUri = Settings.getValueById('Cloud_Workspace_Registration_Client_Uri');
```
ggazzo, line 23: I think I'm missing an await here, alto some test to make sure we have this setting set

.changeset/fifty-laws-sparkle.md
```diff
@@ -0,0 +1,8 @@
+---
+"@rocket.chat/meteor": patch
+"@rocket.chat/core-typings": patch
+"@rocket.chat/model-typings": patch
+"@rocket.chat/models": patch
+---
+
+Now all the cloud credentials are stored in its own collection, preventing stale tokens from being used
```
ggazzo, line 8: I think we could explain a bit better

The fact is not that a new collection prevents this, but not using the cache, which can have a minimal delay, is what fixes the problem

ggazzo (changes requested):

apps/meteor/app/cloud/server/functions/removeLicense.ts
Gustrb: are there any tests on these cloud functions?

apps/meteor/server/settings/setup-wizard.ts
Gustrb: Here it is: https://github.com/RocketChat/Rocket.Chat/pull/32122

apps/meteor/ee/server/models/raw/WorkspaceCredentials.ts
```diff
@@ -0,0 +1,60 @@
...
+			scopes: {
+				$all: [scope],
+				$size: 1,
+			},
+		};
+
+		await this.deleteOne(query);
+	}
+
+	async updateCredentialByScope(scope: string, accessToken: string, expirationDate: Date): Promise<void> {
```
matheusbsilva137, line 38:
```suggestion
	async createOrUpdateCredentialByScope(scope: string, accessToken: string, expirationDate: Date): Promise<void> {
```
Since we're performing an upsert here, it'd be good to make it clear in the method's name.

packages/core-typings/src/IWorkspaceCredentials.ts
```diff
@@ -0,0 +1,8 @@
+import type { IRocketChatRecord } from './IRocketChatRecord';
+
+export interface IWorkspaceCredentials extends IRocketChatRecord {
+	_id: string;
+	scopes: string[];
```
matheusbsilva137, line 5: This may be a dumb question, but do we really need scopes to be an array? :thinking: I see we're always setting AND searching it as an array with a single element

apps/meteor/ee/server/models/raw/WorkspaceCredentials.ts
```diff
@@ -0,0 +1,60 @@
...
+		};
+
+		const query: Filter<IWorkspaceCredentials> = {
+			scopes: {
+				$all: [scope],
+				$size: 1,
+			},
+		};
+
+		await this.updateOne(query, record, { upsert: true });
```
matheusbsilva137, line 54:
```suggestion
		return this.updateOne(query, record, { upsert: true });
```
I know we're not using the returned value here, but it looks like a better practice _to me_ to return it so we're ready to use it if we ever to need to validate something. Same for all other methods in this file
I'd like to hear what others think about this, though

.changeset/fifty-laws-sparkle.md
```diff
@@ -0,0 +1,8 @@
+---
+"@rocket.chat/meteor": patch
+"@rocket.chat/core-typings": patch
+"@rocket.chat/model-typings": patch
+"@rocket.chat/models": patch
```
matheusbsilva137, line 5:
```suggestion
"@rocket.chat/meteor": minor
"@rocket.chat/core-typings": minor
"@rocket.chat/model-typings": minor
"@rocket.chat/models": minor
```
Since this is a feat (not a patch), we should use minor instead.

matheusbsilva137 (changes requested):

packages/core-typings/src/IWorkspaceCredentials.ts
Gustrb: Leaving it as an array is more of future-proofing than anything, it is not necessary nowadays, I think it was a suggestion by @geekgonecrazy

apps/meteor/ee/server/models/raw/WorkspaceCredentials.ts
Gustrb: oh yea, for sure

Gustrb:
@AllanPazRibeiro I guess we should find a way to add tests for our cloud services, but I am not sure if it is possible now, do you have any idea on how could we do it?

apps/meteor/ee/server/models/raw/WorkspaceCredentials.ts
Gustrb: I agree, I don't think I should remove the return value just because I am not using it now, I'll add it, thanks

debdutdeb: That's a very good approach @matheusbsilva137 < 3

packages/core-typings/src/IWorkspaceCredentials.ts
debdutdeb: I agree. Lists are almost always better.

PR #31987, opened 2024-03-13
feat: New users page pending tab

rique223:

This PR introduces a new pending users tab to the users page and a set of new features:

1. The pending tab will list users that havent been activated yet or that have been activated but haven't logged in for the first time;
2. The pending tab contains a counter that shows the number of users to be activated;
3. A new "Pending action" column was introduced, this column displays which action the given user is pending to do/receive, be it activation or first time login;
4. A new "actions" column was introduced in the users table, this column contains a kebab menu similar to the contextual bar one and, when inside of the pending tab, two new buttons:
    - Activate: Activates the user, its a shortcut to opening the kebab menu/contextual bar;
    - Resend welcome email: Resends the welcome e-mail to the given user, this serves as a reminder for the user to login for the first time;
5. A new "Reason for joining" section has been introduced to the users page contextual bar, this section shows the message written by the given user when creating their account if the "Manually approve new users" setting was enabled;
6. A callout component was introduced to the page and it will appear once the workspace exceeds its seats cap;
7. Upon exceeding the seats cap the "Invite", "New User" and "Activate" buttons will now be disabled;
*----------------------------------------- Changes after initial review -------------------------------------------------*
8. A refactor of the responsiveness of the users table as well as a change to the useBreakpoints hook in order not to use magical numbers;
9. Introduced two new user-defined email validation methods for the AJV "format" field based on the emailValidator.ts:
    - basic_email: Simple validation that will only verify the existence of an @ symbol and characters before and after it;
    - rfc_email: More complex validation that aims at being more compliant with RFC 5322;
    This change solves an issue with the standard AJV email validation in which dotless emails were not being accepted.

Demo gif:
pending_users_tab

## Further comments

Blocked by: #31917 
Related to: https://github.com/RocketChat/Rocket.Chat/pull/30236
Jira: [WM-51](https://rocketchat.atlassian.net/browse/WM-51)


apps/meteor/client/views/admin/users/UsersTable/UsersTableRow.tsx
```diff
@@ -42,6 +53,71 @@ const UsersTableRow = ({ user, onClick, mediaQuery, tab }: UsersTableRowProps):
...
+			dispatchToastMessage({ type: 'error', message: t('Welcome_email_failed') });
+			return;
+		}
+
+		resendWelcomeEmail.mutateAsync({ email: emails[0].address });
+	};
+
+	const renderPendingButton = (): ReactElement =>
+		active ? (
+			<Button small secondary mie={8} onClick={handleResendWelcomeEmail}>
```
aleksandernsilva, line 112: Might be a good idea to disable the button while the request/mutation is pending 🤔

apps/meteor/client/views/admin/users/hooks/useSendWelcomeEmailMutation.ts
```diff
@@ -0,0 +1,25 @@
...
+export const useSendWelcomeEmailMutation = (): UseMutationResult<null, Error, useSendWelcomeEmailMutationProps> => {
+	const t = useTranslation();
+	const dispatchToastMessage = useToastMessageDispatch();
+
+	const sendWelcomeEmail = useEndpoint('POST', '/v1/users.sendWelcomeEmail');
+
+	return useMutation(
+		async ({ email }) => {
+			const result = await sendWelcomeEmail({ email });
+			return result;
```
aleksandernsilva, line 18: nitpick: `result` const is unnecessary

aleksandernsilva (changes requested):

aleksandernsilva (changes requested):
- I was not able to see the full text of the "Resend welcome email" button, even in full screen. This doesn't seem ideal 🤔 

Screenshot 2024-05-10 at 10 36 11

- It seems email validation when registering a user is different than the one required by the `users.sendWelcomeEmail` endpoint, causing it to return 400 in this scenario:

Screenshot 2024-05-10 at 10 46 26

Screenshot 2024-05-10 at 10 46 22

- There seems to be a z-index issue with the table headers, it is going over the contextual bar (see "Pending action"):

Screenshot 2024-05-10 at 10 43 48

apps/meteor/client/views/admin/users/UsersTable/UsersTableRow.tsx
rique223: I believe @dougfabris said something in one of the frontend weekly's about not disabling buttons anymore as this was bad for accessibility

aleksandernsilva (approved):

packages/rest-typings/src/v1/Ajv.ts
```diff
@@ -4,6 +4,13 @@ import addFormats from 'ajv-formats';
 const ajv = new Ajv({
 	coerceTypes: true,
 });
+
 addFormats(ajv);
 
+ajv.addFormat('basic_email', /^[^@]+@[^@]+$/);
+ajv.addFormat(
+	'rfc_email',
+	/^[a-zA-Z0-9.!#$%&'*+\/=?^_`{|}~-]+@[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?(?:\.[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?)*$/,
```
Gustrb, line 13: Where did this validation comes from?
I thought the correct RFC email validation regex was this one: https://pdw.ex-parrot.com/Mail-RFC822-Address.html

but I am not sure if we need to go this far though

rique223: I believe you have already found the email validator as you have resolved this conversation, but I will comment here for future reference. These regexes were extracted from the emailValidator.ts in the apps/meteor/lib folder, which was implemented in this PR: https://github.com/RocketChat/Rocket.Chat/pull/23816. Since these were approved by Tasso and Diego, I believe they accurately represent RC's view on email validation.

I am aware that more robust regular expressions exist, and I encountered several during my research on the impacts of this change. However, since we already have an in-house solution that is used throughout most of the system, I decided to maintain it for consistency's sake.

Gustrb: yea, I was looking into to see if we were being consistent,
once I found the email validator using the same regex to validate it I decided to delete, since I don't think we need to go over the top with e-mail validation since the actual only way for us to be 100% an e-mail is valid is to actually send stuff to it, also, the regex that is being used is good enough for all the cases I could've think about.

and since it is consistent there is no need to worry about this, good work :)

Gustrb (approved):

Gustrb (approved):

apps/meteor/client/views/admin/users/AdminUsersPage.tsx
```diff
@@ -104,13 +120,33 @@ const AdminUsersPage = (): ReactElement => {
 					</TabsItem>
 				</Tabs>
 				<PageContent>
+					{isSeatsCapExceeded && (
```
ggazzo, line 123: use the same logic used here `SubscriptionCalloutLimits`, also check with pms if there is a real need to introduce a new texts to explain the same thing

packages/core-typings/src/IUser.ts
```diff
@@ -191,7 +192,7 @@ export interface IRegisterUser extends IUser {
 }
 
 export const isRegisterUser = (user: IUser): user is IRegisterUser => user.username !== undefined && user.name !== undefined;
-export const isUserFederated = (user: Partial<IUser>) => 'federated' in user && user.federated === true;
+export const isUserFederated = (user: Partial<IUser> | Partial<Serialized<IUser>>) => 'federated' in user && user.federated === true;
```
ggazzo, line 195: `isUserFederated` is not actually an assertion function, this function is actually simple useless, and  it should not be used since the `federated` prop is defined on the `IUser` type, so just do `Boolean(user.federated)` instead

ggazzo (changes requested):

apps/meteor/client/views/admin/users/UsersTable/UsersTableRow.tsx
```diff
@@ -42,6 +54,71 @@ const UsersTableRow = ({ user, onClick, mediaQuery, tab }: UsersTableRowProps):
...
+		...(deleteUserAction && {
+			delete: { label: { label: deleteUserAction.label, icon: deleteUserAction.icon }, action: deleteUserAction.action },
+		}),
+	};
+
+	const handleResendWelcomeEmail = () => {
+		if (!emails?.length) {
+			dispatchToastMessage({ type: 'error', message: t('Welcome_email_failed') });
+			return;
+		}
```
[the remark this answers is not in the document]
rique223: Thist verification is avoiding undefined for the emails list because the emails field of IUser is optional, if it is not possible for it not to be set why is the field optional?

apps/meteor/client/views/admin/users/UsersTable/UsersTableRow.tsx
```diff
@@ -53,39 +130,68 @@ const UsersTableRow = ({ user, onClick, mediaQuery, tab }: UsersTableRowProps):
...
-									{`@${username}`}
-								</Box>
-							)}
+					{username && <UserAvatar size={isMobile || isLaptop ? 'x28' : 'x40'} username={username} etag={avatarETag} />}
+					<Box display='flex' flexGrow={1} flexShrink={1} flexBasis='0%' alignSelf='center' alignItems='center' withTruncatedText>
+						<Box mi={8}>
+							<UserStatus status={status || Status.OFFLINE} />
+						</Box>
+						<Box fontScale='p2' withTruncatedText>
+							{name || username}
```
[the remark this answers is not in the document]
rique223: Yes, but the problem here is that if I show any other column on mobile it will cause issues with the "Resend welcome email" due to the large width of it.

tassoevan (approved):

ggazzo (approved):

PR #31990, opened 2024-03-13
fix(Omnichannel): Open expanded view (galery mode) for image attachments from livechat

tiagoevanp:
## Further comments

[SUP-499](https://rocketchat.atlassian.net/browse/SUP-499)

[SUP-499]: https://rocketchat.atlassian.net/browse/SUP-499?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/meteor/app/livechat/server/methods/sendMessageLivechat.ts
```diff
@@ -26,7 +26,7 @@ declare module '@rocket.chat/ui-contexts' {
 }
 
 export const sendMessageLivechat = async ({
-	message: { token, _id, rid, msg, file, attachments },
+	message: { token, _id, rid, msg, file, files, attachments },
```
MarcosSpessatto, line 29: There is a place in the code, calling this function without all the properties, we need to take a look on that as well.

https://github.com/RocketChat/Rocket.Chat/blob/develop/apps/meteor/app/livechat/server/methods/sendMessageLivechat.ts#L83

MarcosSpessatto (changes requested):
Also, let's add a test to cover this scenario, if I understood correctly the problem is not to have the `files` property within the msg object. We can send a file message through the Livechat API and then get the message(`chat.getMessage`) to ensure the `files` are there.

apps/meteor/app/livechat/server/methods/sendMessageLivechat.ts
tiagoevanp: I think nobody is using this Meteor method anymore... I've found zero locals searching... Removing it from code!

tassoevan (dismissed):

apps/meteor/tests/data/livechat/rooms.ts
```diff
@@ -139,7 +149,10 @@ export const createManager = (overrideUsername?: string): Promise<ILivechatAgent
 			});
 	});
 
-export const makeAgentAvailable = async (overrideCredentials?: { 'X-Auth-Token': string | undefined; 'X-User-Id': string | undefined }): Promise<Response> => {
+export const makeAgentAvailable = async (overrideCredentials?: {
+	'X-Auth-Token': string | undefined;
+	'X-User-Id': string | undefined;
+}): Promise<Response> => {
```
MarcosSpessatto, line 155: Same as before.

apps/meteor/tests/data/livechat/rooms.ts
```diff
@@ -188,7 +201,7 @@ export const getLivechatRoomInfo = (roomId: string): Promise<IOmnichannelRoom> =
 
 /**
  * @summary Sends message as visitor
-*/
+ */
```
MarcosSpessatto, line 204:
```suggestion
*/
```

apps/meteor/tests/data/livechat/rooms.ts
MarcosSpessatto, line 1: Let's keep only the function you created (`uploadFIle`) here, and remove everything else related to formatting.

apps/meteor/tests/end-to-end/api/livechat/20-messages.ts
```diff
@@ -51,5 +52,17 @@ describe('LIVECHAT - messages', () => {
...
 			}
 		});
+
+		it('it should verify if visitor is receiving a message with a image attachment', async () => {
+			const {
+				room: { _id: roomId },
+				visitor: { token },
+			} = await startANewLivechatRoomAndTakeIt();
+
+			const imgMessage = await uploadFile(roomId, token);
```
MarcosSpessatto, line 62: I think it would be better if we could grab the message using this endpoint: https://github.com/RocketChat/Rocket.Chat/blob/develop/apps/meteor/app/api/server/v1/chat.ts#L110

MarcosSpessatto (changes requested):

apps/meteor/tests/end-to-end/api/livechat/20-messages.ts
tiagoevanp: ok, but the endpoint called by livechat is calling uploadFile...

apps/meteor/app/livechat/server/methods/sendMessageLivechat.ts
MarcosSpessatto: We still need to add the `files` property to the method, we removed it and added it back, but we didn't add the missing `files` property there. 😬

MarcosSpessatto (changes requested):

apps/meteor/app/livechat/server/methods/sendMessageLivechat.ts
tiagoevanp: added

MarcosSpessatto (dismissed):

tassoevan (dismissed):

.changeset/green-ways-tie.md
```diff
@@ -0,0 +1,5 @@
+---
+"@rocket.chat/meteor": patch
+---
+
+Fix Open expanded view (galery mode) for image attachments sent by livechat
```
KevLehman, line 5:
```suggestion
Fixed open expanded view (galery mode) for image attachments sent by livechat widget
```

apps/meteor/tests/data/livechat/rooms.ts
```diff
@@ -208,6 +209,21 @@ export const sendMessage = (roomId: string, message: string, visitorToken: strin
...
+	return new Promise((resolve, reject) => {
+		request
+			.post(api(`livechat/upload/${roomId}`))
+			.set({ 'content-type': '', 'x-visitor-token': visitorToken, ...credentials })
+			.attach('file', imgURL)
+			.end((err: Error, res: DummyResponse<IMessage>) => {
+				if (err) {
+					return reject(err);
+				}
+				resolve(res.body as unknown as IMessage);
```
KevLehman, line 222: Meteor methods return a non error response even for errors. Better check if it actually worked before converting the type

apps/meteor/tests/end-to-end/api/livechat/20-messages.ts
```diff
@@ -51,5 +52,17 @@ describe('LIVECHAT - messages', () => {
 				expect(quotedMessageAttachments).to.have.property('text').that.is.equal(agentMsgSentence);
 			}
 		});
+
+		it('it should verify if visitor is receiving a message with a image attachment', async () => {
```
KevLehman, line 56:
```suggestion
		it('should verify if visitor is receiving a message with a image attachment', async () => {
```

apps/meteor/tests/data/livechat/rooms.ts
```diff
@@ -208,6 +209,21 @@ export const sendMessage = (roomId: string, message: string, visitorToken: strin
 	});
 };
 
+export const uploadFile = (roomId: string, visitorToken: string): Promise<IMessage> => {
+	return new Promise((resolve, reject) => {
+		request
+			.post(api(`livechat/upload/${roomId}`))
+			.set({ 'content-type': '', 'x-visitor-token': visitorToken, ...credentials })
```
KevLehman, line 216: Do u need to set content type as empty?

apps/meteor/tests/end-to-end/api/livechat/20-messages.ts
```diff
@@ -51,5 +52,17 @@ describe('LIVECHAT - messages', () => {
...
+			const {
+				room: { _id: roomId },
+				visitor: { token },
+			} = await startANewLivechatRoomAndTakeIt();
+
+			const imgMessage = await uploadFile(roomId, token);
+
+			expect(imgMessage).to.have.property('files');
+			expect(imgMessage.files).to.be.an('array');
```
KevLehman, line 65: You can combine these 2 expects in one too.

But also would be cool if you checked for the array to contain what u want too

KevLehman (changes requested):

apps/meteor/tests/data/livechat/rooms.ts
KevLehman: Saw you added an if to another function but not to this one 🤔

apps/meteor/tests/end-to-end/api/livechat/20-messages.ts
```diff
@@ -51,5 +52,16 @@ describe('LIVECHAT - messages', () => {
...
+
+		it('should verify if visitor is receiving a message with a image attachment', async () => {
+			const {
+				room: { _id: roomId },
+				visitor: { token },
+			} = await startANewLivechatRoomAndTakeIt();
+
+			const imgMessage = await uploadFile(roomId, token);
+
+			expect(imgMessage).to.have.property('files').that.is.an('array');
```
KevLehman, line 64: Is there something else we can check on here? Like, the array including the actual file you sent and not just being an empty array 👀

KevLehman (approved):

PR #31992, opened 2024-03-14
refactor: mentions out of db watcher

ricardogarim:
This pull request implements mention notifications for users in rooms and private conversations without the need to observe document updates in MongoDB, specifically in the subscriptions collection, via change streams. With this new approach, whenever there's a new mention, an event will be triggered directly to the listener service, which will be responsible for distributing notifications directly to the DDP Streamer service, which in turn will notify the users subscribed to the new event. MongoDB document updates in subscriptions will still exist on the backend, but they will no longer be used as triggers to fire notifications.

The new events consist of only one format: `{ROOM_ID}/mention`, which can be subscribed through the `notify-user` stream.

**Current format - passing through DB Watcher:**
```mermaid
sequenceDiagram
    actor a as User A
    actor b as User B
    participant app as App
    participant mongodb as MongoDB
    participant dbwatcher as DB Watcher
    participant listener as Listener
    participant ddp as DDP Streamer

    a->>app: Send message w/ User B mention
    app->>mongodb: Update B subscription document
    mongodb->>dbwatcher: Listen for document updates
    dbwatcher->>listener: Send mention action for processing
    listener->>ddp: Send mention notification to WS handler
    ddp->>b: Send mention notification to User B
```

**Suggested approach - out of DB Watcher:**
```mermaid
sequenceDiagram
    actor a as User A
    actor b as User B
    participant app as App
    participant mongodb as MongoDB
    participant listener as Listener
    participant ddp as DDP Streamer

    par Notify mention directly to listener
        a->>app: Send message w/ User B mention
        app->>listener: Send mention action for processing
        listener->>ddp: Send mention notification to WS handler
        ddp->>b: Send mention notification to User B
    and Update subscription on DB
        app->>mongodb: Update B subscription document
    end
```

## Steps to test or reproduce
To use the new mention format, it is necessary to start the Rocket.Chat application by informing the flag `DISABLE_DB_WATCHERS=true`.

The use cases that need to be tested are:
- Mentioning users in a room.
- Mentioning aliases `@here` and `@all` in a room.

The expected behavior for both use cases is the same as the current behavior. When receiving a nominal mention, the user should notice, in the left sidebar listing rooms, a red icon with a number representing the number of mentions received. Additionally, the name of the room should be highlighted in bold. When receiving a general mention (here or all), the user should only see the room name in bold.

## Further comments
The client-side implementation seeks to make as few modifications as possible. Thus, the only modification occurs at the application startup, which will start subscribing to mention event for the user. Whenever any mention event is received, the client will update the existing subscription document on Minimongo. We chose to proceed in this manner to avoid refactoring the frontend at this time. It is not ruled out the possibility of implementing the new mention format on the frontend in the future, but for now, there is no need to do so.

Performance changes were measured using K6 load testing and from user's perspective there are no degradation as well as insignificant performance increase, but since there will no longer be a need to observe updates in MongoDB documents, overall health of the system grows. Also, the query to get users from channels doesnt affect the performance of the system as it looks for a indexed field and return just the needed infos.

Moreover, existing constraints such as the maximum number of users in a room for it to receive `@all` or `@here` mentions can be avaliated again. This is because, with the new approach, the server can send notifications directly to the DDP Streamer, eliminating the need for intermediaries and avoiding the generation of N events for each notified user, where N represents the total number of participants in each room.

Edit: In an update to our initial proposal, we have opted for a single stream for all mentions, rather than separate streams for user and room-related mentions. This decision was made due to the significant number of potential subscriptions at the startup of each client, as a user could be associated with multiple rooms, leading to numerous subscriptions. Under the new approach, all mentions are directed to the user stream, and the server manages the task of directing notifications to the appropriate recipients.

### DIFF
--- a/apps/meteor/app/lib/server/lib/notifyUsersOnMessage.ts
+++ b/apps/meteor/app/lib/server/lib/notifyUsersOnMessage.ts
@@ -20,36 +20,35 @@ function messageContainsHighlight(message: IMessage, highlights: string[]): bool
 }
 
 export async function getMentions(message: IMessage): Promise<IMentionCounter> {
-	if (!message.mentions) return { hasAllMention: false, hasHereMention: false, mentionIds: [] };
+	if (!message.mentions) return { toAll: false, toHere: false, mentionIds: [] };
 
-	const {
-		u: { _id: senderId },
-	} = message;
+	let toAll = false;
+	let toHere = false;
 
-	let hasAllMention = false;
-	let hasHereMention = false;
+	const mentions = message.mentions.reduce(
+		(acc, cur) => {
+			if (cur._id === message.u._id) return acc;
 
-	const mentions = message.mentions.reduce((acc, cur) => {
-		if (cur._id === senderId) return acc;
+			if (cur.type === 'user') {
+				acc.u.push(cur._id);
+			} else if (cur.type === 'team') {
+				acc.t.push(cur._id);
+			} else if (cur._id === 'all') {
+				toAll = true;
+				acc.c.push(message.rid);
+			} else if (cur._id === 'here') {
+				toHere = true;
+				acc.c.push(message.rid);
+			}
 
-		if (cur.type === 'user') {
-			acc.u.push(cur._id);
-		} else if (cur.type === 'team') {
-			acc.t.push(cur._id);
-		} else if (cur._id === 'all') {
-			hasAllMention = true;
-			acc.c.push(message.rid);
-		} else if (cur._id === 'here') {
-			hasHereMention = true;
-			acc.c.push(message.rid);
-		}
-
-		return acc;
-	}, {} as { c: string[]; u: string[]; t: string[] });
+			return acc;
+		},
+		{ c: [], u: [], t: [] } as { c: string[]; u: string[]; t: string[] },
+	);
 
 	const mentionIds = await beforeGetMentions(mentions.u, mentions.t, mentions.c);
 
-	return { hasAllMention, hasHereMention, mentionIds };
+	return { toAll, toHere, mentionIds };
 }
 
 type UnreadCountType = 'all_messages' | 'user_mentions_only' | 'group_mentions_only' | 'user_and_group_mentions_only';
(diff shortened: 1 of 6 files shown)


packages/core-services/src/events/Events.ts
```diff
@@ -53,6 +53,17 @@ type LoginServiceConfigurationEvent = {
...
 
+// TODO: Decided to keep this here for now, but it should be moved to a more appropriate place in the future
+export interface IMentionCounter {
+	toAll: boolean;
+	toHere: boolean;
+	mentionIds: string[];
+}
+
+export enum EventNames {
+	USER_MENTIONS = 'user.mention',
```
MarcosSpessatto, line 64: Would not it be clearer to have another value for the event, since we are not handling only user mentions through this event, but also `@all` and `@here`?

apps/meteor/app/lib/server/lib/notifyUsersOnMessage.ts
```diff
@@ -16,36 +19,36 @@ function messageContainsHighlight(message: IMessage, highlights: string[]): bool
...
-	}
+export async function getMentions(message: IMessage): Promise<IMentionCounter> {
+	if (!message.mentions) return { toAll: false, toHere: false, mentionIds: [] };
+
+	let toAll = false;
+	let toHere = false;
+
+	const mentions = message.mentions.reduce(
+		(acc, cur) => {
+			if (cur._id === message.u._id) return acc;
```
MarcosSpessatto, line 30: What about create a variable with the name of this condition? I'm not sure what does it mean 🤔 WDYT?

MarcosSpessatto (changes requested):

apps/meteor/app/lib/server/lib/notifyUsersOnMessage.ts
```diff
@@ -16,36 +19,36 @@ function messageContainsHighlight(message: IMessage, highlights: string[]): bool
...
+			}
 
-	const teamsMentions = mentions.filter((mention) => mention.type === 'team');
-	const filteredMentions = mentions
-		.filter((mention) => !mention.type || mention.type === 'user')
-		.filter(({ _id }) => _id !== senderId && !['all', 'here'].includes(_id))
-		.map(({ _id }) => _id);
+			return acc;
+		},
+		{ c: [], u: [], t: [] } as { c: string[]; u: string[]; t: string[] },
```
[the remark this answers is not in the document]
debdutdeb: I agree with Marcos :+1:

PR #31998, opened 2024-03-14
fix: Allow queue to fix inquiry status when processing items

KevLehman:
## Issue(s)

[CORE-200](https://rocketchat.atlassian.net/browse/CORE-200)
## Further comments

Introduce a new step to the queue worker: when an inquiry that's on an improper status is selected for processing, queue worker will first check its status and will attempt to fix it.
For example, if an inquiry points to a closed room, there's no point in processing, system will now remove the inquiry
If an inquiry is already taken, the inquiry will be updated to reflect the new status and clean the queue.

This prevents issues where the queue worker attempted to process an inquiry _forever_ because it was in an improper state.

[CORE-200]: https://rocketchat.atlassian.net/browse/CORE-200?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/meteor/tests/unit/server/services/omnichannel/queue.tests.ts
```diff
@@ -0,0 +1,462 @@
...
+		});
+		it('should remove inquiries from rooms that were closed', async () => {
+			const queue = new OmnichannelQueue();
+			await queue.reconciliation('closed', { roomId: 'rid', inquiryId: 'inquiryId' });
+
+			expect(models.LivechatInquiry.removeByRoomId.calledOnce).to.be.true;
+		});
+		it('should return true for any other case', async () => {
+			const queue = new OmnichannelQueue();
+			expect(await queue.reconciliation('random', { roomId: 'rid', inquiryId: 'inquiryId' })).to.be.true;
```
MarcosSpessatto, line 288: The function returns true no matter what happens. We could assert if the model's functions are not being called.

apps/meteor/tests/unit/server/services/omnichannel/queue.tests.ts
```diff
@@ -0,0 +1,462 @@
...
+
+			const queue = new OmnichannelQueue();
+			expect(await queue.processWaitingQueue('department1', mockedInquiry)).to.be.true;
+			expect(
+				queueLogger.debug.calledWith({
+					msg: 'Room from inquiry missing. Removing inquiry',
+					roomId: 'rid',
+					inquiryId: 'inquiryId',
+					step: 'reconciliation',
+				}),
```
MarcosSpessatto, line 328: I don't think we should check the log message here, we can check if `reconciliation()` has been called with the correct parameters instead.

apps/meteor/tests/unit/server/services/omnichannel/queue.tests.ts
```diff
@@ -0,0 +1,462 @@
...
+			).to.be.true;
+		});
+		it('should call takeInquiry when findOneById returns a room thats already being served', async () => {
+			models.LivechatRooms.findOneById.returns({ _id: 'rid', servedBy: { some: 'thing' } });
+
+			const queue = new OmnichannelQueue();
+			queue.reconciliation = Sinon.stub().returns(true);
+			expect(await queue.processWaitingQueue('department1', mockedInquiry)).to.be.true;
+			expect(queue.reconciliation.calledOnce).to.be.true;
```
MarcosSpessatto, line 337: Let's check if the function has been called with the expected parameters.

apps/meteor/tests/unit/server/services/omnichannel/queue.tests.ts
```diff
@@ -0,0 +1,462 @@
...
+		});
+		it('should call removeInquiry when findOneById returns a room that was closed', async () => {
+			models.LivechatRooms.findOneById.returns({ _id: 'rid', closedAt: new Date() });
+
+			const queue = new OmnichannelQueue();
+			queue.reconciliation = Sinon.stub().returns(true);
+			expect(await queue.processWaitingQueue('department1', mockedInquiry)).to.be.true;
+			expect(queue.reconciliation.calledOnce).to.be.true;
```
MarcosSpessatto, line 345: Same as before.

apps/meteor/tests/unit/server/services/omnichannel/queue.tests.ts
```diff
@@ -0,0 +1,462 @@
...
+			license.shouldPreventAction.reset();
+			queueLogger.debug.reset();
+		});
+
+		after(() => {
+			license.shouldPreventAction.reset();
+			queueLogger.debug.reset();
+		});
+
+		it('should return if service is not running', async () => {
```
MarcosSpessatto, line 385:
```suggestion
		it('should return undefined if service is not running', async () => {
```

apps/meteor/tests/unit/server/services/omnichannel/queue.tests.ts
```diff
@@ -0,0 +1,462 @@
...
+			queue.running = false;
+			expect(await queue.execute()).to.be.undefined;
+		});
+		it('should return undefined if license is over mac limits', async () => {
+			license.shouldPreventAction.returns(true);
+
+			const queue = new OmnichannelQueue();
+			queue.running = true;
+			expect(await queue.execute()).to.be.undefined;
+			expect(license.shouldPreventAction.calledOnce).to.be.true;
```
MarcosSpessatto, line 396: We could also assert the `running` variable here being turned to `false`.

apps/meteor/tests/unit/server/services/omnichannel/queue.tests.ts
```diff
@@ -0,0 +1,462 @@
...
+		it('should try to process a queue if license is not over mac limits', async () => {
+			license.shouldPreventAction.returns(false);
+
+			const queue = new OmnichannelQueue();
+			queue.running = true;
+			queue.nextQueue = Sinon.stub();
+			await queue.execute();
+
+			expect(queue.nextQueue.calledOnce).to.be.true;
+			expect(queueLogger.debug.calledOnce).to.be.true;
```
MarcosSpessatto, line 407: I think just the assertion with `calledWith` is good enough, isn't it?

apps/meteor/tests/unit/server/services/omnichannel/queue.tests.ts
```diff
@@ -0,0 +1,462 @@
...
+			queueLogger.info.reset();
+		});
+		it('should unlock all inquiries and set running to false', async () => {
+			const queue = new OmnichannelQueue();
+			queue.running = true;
+			await queue.stop();
+
+			expect(queue.running).to.be.false;
+			expect(models.LivechatInquiry.unlockAll.calledOnce).to.be.true;
+			expect(queueLogger.info.calledOnce).to.be.true;
```
MarcosSpessatto, line 458: Just calledWith is good enough.

MarcosSpessatto (changes requested):

apps/meteor/server/services/omnichannel/queue.ts
```diff
@@ -123,26 +126,74 @@ export class OmnichannelQueue implements IOmnichannelQueue {
 		void (routingSupportsAutoAssign ? this.start() : this.stop());
 	}
 
+	private async reconciliation(reason: 'closed' | 'taken' | 'missing', { roomId, inquiryId }: { roomId: string; inquiryId: string }) {
+		switch (reason) {
+			case 'closed': {
```
MarcosSpessatto, line 131: Also, I think we can use the same `case` for `closed` and `missing`, they are almost identical, just the log message is slightly different. WDYT?

apps/meteor/tests/unit/server/services/omnichannel/queue.tests.ts
KevLehman: Yeah, don't think that's something to worry tho.

apps/meteor/tests/unit/server/services/omnichannel/queue.tests.ts
```diff
@@ -0,0 +1,462 @@
...
+		it('should fetch active queues and set running to true', async () => {
+			const queue = new OmnichannelQueue();
+			queue.running = false;
+			queue.getActiveQueues = Sinon.stub().returns(['department1']);
+			queue.execute = Sinon.stub();
+			await queue.start();
+
+			expect(queue.running).to.be.true;
+			expect(queue.getActiveQueues.calledOnce).to.be.true;
+			expect(queueLogger.info.calledOnce).to.be.true;
```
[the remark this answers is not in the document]
KevLehman: Same

apps/meteor/tests/unit/server/services/omnichannel/queue.tests.ts
KevLehman: Same

apps/meteor/server/services/omnichannel/queue.ts
KevLehman: Message is different. They call the same funciton but this way we know what actually happened with a room by seeing the logs.

apps/meteor/tests/unit/server/services/omnichannel/queue.tests.ts
KevLehman: both are valid ways imho. but that check is not for assuring the function received the valid params, is for assuring that reconciliation got called at the right step.

Since missing & closed call the same func, the log is a good way to diff them.

MarcosSpessatto (approved):

MarcosSpessatto (approved):

PR #32002, opened 2024-03-15
test: forward messages

jessicaschelly:

Creation of test cases for the Forward message feature
- Forward text message to channel
- Forward text message to team
- Forward text message to direct message


apps/meteor/client/components/UserAndRoomAutoCompleteMultiple/UserAndRoomAutoCompleteMultiple.tsx
```diff
@@ -51,6 +51,7 @@ const UserAndRoomAutoCompleteMultiple = ({ value, onChange, ...props }: UserAndR
 	return (
 		<AutoComplete
 			{...props}
+			data-qa-type='user-and-room-auto-complete-input'
```
ggazzo, line 54: could you reach this field using role/name?

ggazzo (changes requested):

apps/meteor/client/components/UserAndRoomAutoCompleteMultiple/UserAndRoomAutoCompleteMultiple.tsx
dougfabris, line 54:
```suggestion
			name='userAndRoom'
```
Can we avoid adding this kind of prop?
Instead, you can pass a name for this component and get it by role and this name

e.g: `page.getByRole('textbox', { name: 'userAndRoom' })`

dougfabris (changes requested):

dougfabris (approved):
LGTM!

MarcosSpessatto (approved):

PR #32017, opened 2024-03-18
test: solve flaky edit unit test

ggazzo:
It seems that  opening the page and the unit creation step happened at the same time. As the page does not have the search input if there is no item the test would fail. So it was up to the destiny (network plus cpu) to decide what happened first


dougfabris (approved):
LGTM!

PR #32022, opened 2024-03-18
chore: register a dummy Read Receipts model in CE

pierre-lehnen-rc:
so that the model is always defined even when the EE code is not executed

## Issue(s)

FOSS-9


apps/meteor/server/models/dummy/ReadReceipts.ts
```diff
@@ -0,0 +1,58 @@
...
+	constructor() {
+		super('read_receipts');
+	}
+
+	findByMessageId(_messageId: string): FindCursor<ReadReceipt> {
+		return this.find({});
+	}
+
+	removeByUserId(_userId: string): Promise<DeleteResult> {
+		return this.deleteMany({});
```
sampaiodiego, line 17: this is scary  🙈 

my understanding is that dummy models will never be actually called, so should them only throw errors? having this operations here are way too scary

sampaiodiego (changes requested):

apps/meteor/server/models/dummy/ReadReceipts.ts
pierre-lehnen-rc: it's not an operation, it's just calling the mocked `deleteMany` that returns a compatible `DeleteResult` without doing anything.

pierre-lehnen-rc: The dummy model doesn't even have a mongo connection to operate on.

sampaiodiego: lol yeah I missed that 😛 my bad.

sampaiodiego (approved):

PR #32025, opened 2024-03-18
chore(livechat): stop calling twice closeChat 

ggazzo:
it seems that in extreme conditions (low cpu x slow network)
It is possible to call close chat more than once,
one caused by user interaction and the other by the message received via websocket. causing some navigation problems.


MartinSchoeler (approved):
That explains a lot lol

PR #32046, opened 2024-03-21
fix: `rooms.export` endpoint generates empty export when given an invalid date

matheusbsilva137:

 - Use oldest possible date as `dateFrom` when this param is not provided to the `rooms.export` endpoint;
 - Use current date as `dateTo` when this param is not provided to the `rooms.export` endpoint;
 - Added end-to-end tests for the export rooms as file feature.

## Further comments

[SUP-523](https://rocketchat.atlassian.net/browse/SUP-523)

[SUP-523]: https://rocketchat.atlassian.net/browse/SUP-523?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


abhinavkrin (approved):
LGTM!

KevLehman (changes requested):

apps/meteor/tests/end-to-end/api/09-rooms.js
```diff
@@ -1848,4 +1848,147 @@ describe('[Rooms]', function () {
...
+				})
+				.expect('Content-Type', 'application/json')
+				.expect(400)
+				.expect((res) => {
+					expect(res.body).to.have.property('success', false);
+					expect(res.body).to.have.property('errorType', 'error-invalid-format');
+				});
+		});
+
+		it('should succesfully export room as file', async () => {
```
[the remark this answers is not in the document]
matheusbsilva137: Ohhh, good catch
I thought of using a delay + our `Triggered_Emails_Count` setting, wdyt? :thinking:
So that we don't need a new flow for TEST_MODE=true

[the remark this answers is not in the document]
matheusbsilva137: I've just added some new end-to-end tests to validate the content of the exported file. Please check this too :hugs:

apps/meteor/client/views/room/contextualBar/ExportMessages/FileExport.tsx
```diff
@@ -31,10 +31,10 @@ const FileExport = ({ formId, rid, exportOptions, onCancel }: FileExportProps) =
 		[t],
 	);
 
-	const handleExport = ({ type, dateFrom, dateTo, format }: MailExportFormValues) => {
+	const handleExport = ({ dateFrom, dateTo, format }: MailExportFormValues) => {
```
tiagoevanp, line 34: If you are changing the props expected, you will need to change where this function is called to... But I think you can do that change in another PR, this is not important from what you are doing here, right?

tiagoevanp (changes requested):

apps/meteor/tests/unit/server/lib/dataExport/exportRoomMessagesToFile.spec.ts
```diff
@@ -0,0 +1,213 @@
...
+			_id: 'oNfuDGerMEiPoF7tq',
+			username: 'matheus.barbosa',
+			name: 'Matheus Barbosa Silva',
+		},
+		groupable: false,
+	},
+	{
+		_id: 'LmHvpSLmZfwcuNG9e',
+		rid: '65f7aa32d187cb7fe6779da3',
+		ts: new Date('2024-03-20T13:46:01.623Z'),
```
jessicaschelly, line 56: Is there a specific reason we're using these exact dates instead of relative dates? This might give the impression of arbitrary date selection

apps/meteor/client/views/room/contextualBar/ExportMessages/FileExport.tsx
matheusbsilva137: Actually, this is called directly by handleSubmit :thinking: Is there any way to make it not send the type param to this function? I thought of removing type from `MailExportFormValues`, but this is used to set default values.

apps/meteor/tests/unit/server/lib/dataExport/exportRoomMessagesToFile.spec.ts
matheusbsilva137: Not exactly, I just used an example response I had locally so that I didn't have to organize the data manually. I'm not sure about what would be the best approach, though :thinking:

If I had used random dates here, then I'd have to repeat the same process that is applied in the tested function to validate that the date is being correctly parsed to string. By using a fixed date, I don't need any magic and this can be checked as a string (as done [here](https://github.com/RocketChat/Rocket.Chat/pull/32046/files#diff-5d6db8c10f7d42233b9605b0caef5e44068dc875bd6c19eb33a8d4d3b1cdc23fR157) ). Which approach do you prefer?

apps/meteor/tests/unit/server/lib/dataExport/exportRoomMessagesToFile.spec.ts
```diff
@@ -0,0 +1,213 @@
...
+const messages = [
+	{
+		_id: '65f46ce9e162d4ca5e6aba5c',
+		t: 'uj',
+		rid: 'GENERAL',
+		ts: new Date('2024-03-15T15:44:41.889Z'),
+		msg: 'matheus.barbosa',
+		u: {
+			_id: 'oNfuDGerMEiPoF7tq',
+			username: 'matheus.barbosa',
```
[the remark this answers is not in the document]
matheusbsilva137: I've just added multiple names and usernames just to make sure we're not messing this up

apps/meteor/tests/unit/server/lib/dataExport/exportRoomMessagesToFile.spec.ts
```diff
@@ -0,0 +1,213 @@
...
+			i18n: {
+				t: stubs.translateKey,
+			},
+		},
+		'../../../app/settings/server': {
+			settings: stubs.settings,
+		},
+	});
+
+const messages = [
```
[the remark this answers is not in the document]
KevLehman: And maybe using something like "faker" for these data

apps/meteor/tests/unit/server/lib/dataExport/exportRoomMessagesToFile.spec.ts
jessicaschelly: Got it! In this case, maybe we could add to a variable, so it can be a little more clear when reading the test. WDYT? Asking since I saw it's used multiple times - It's just a personal preference but not an issue at all. 

e.g: `messageSentDate`

apps/meteor/app/api/server/v1/rooms.ts
```diff
@@ -581,12 +577,8 @@ API.v1.addRoute(
 				const { dateFrom, dateTo } = this.bodyParams;
 				const { format } = this.bodyParams;
 
-				if (!['html', 'json'].includes(format || '')) {
-					throw new Meteor.Error('error-invalid-format');
-				}
-
-				const convertedDateFrom = new Date(dateFrom || '');
-				const convertedDateTo = new Date(dateTo || '');
+				const convertedDateFrom = dateFrom ? new Date(dateFrom) : new Date(0);
```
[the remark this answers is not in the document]
matheusbsilva137: This made me curious too, but turns out `coerceTypes` will cast the data to `string`, since this is the `type` of this param. So we still need to convert it into a date.

apps/meteor/tests/unit/server/lib/dataExport/exportRoomMessagesToFile.spec.ts
```diff
@@ -0,0 +1,213 @@
...
+describe('Export - exportMessageObject', () => {
+	before(() => {
+		stubs.translateKey.returns('translated-placeholder-attachment');
+	});
+
+	it('should only stringify message object when exporting message as json', async () => {
+		const result = await exportMessageObject('json', messagesData[3]);
+
+		expect(result).to.be.a.string;
+		expect(result).to.equal(`{"msg":"${messagesData[3].msg}","username":"${messagesData[3].username}","ts":"2024-03-26T18:24:21.994Z"}`);
```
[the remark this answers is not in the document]
matheusbsilva137: I was trying to make sure no other key would be added. But stringifying it again should be enough, though
Tks for the suggestion

apps/meteor/tests/end-to-end/api/09-rooms.js
```diff
@@ -1848,4 +1848,230 @@ describe('[Rooms]', function () {
...
+				})
+				.expect('Content-Type', 'application/json')
+				.expect(400)
+				.expect((res) => {
+					expect(res.body).to.have.property('success', false);
+					expect(res.body).to.have.property('errorType', 'invalid-params');
+				});
+		});
+
+		it('should succesfully export room via email', async () => {
```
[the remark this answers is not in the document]
matheusbsilva137: There's no way to properly test this here. We just filter invalid messages out of the final export (we apply a query in the Messages model `$in: providedMessagesIds`, so there's no error to catch)

apps/meteor/tests/unit/server/lib/dataExport/exportRoomMessagesToFile.spec.ts
matheusbsilva137: Done! I ended up using `faker`, which looks even better. So now I don't reference a specific date anymore :partying_face: 
Thanks again for the suggestions

jessicaschelly (approved):
LGTM! 🥳

apps/meteor/tests/unit/server/lib/dataExport/exportRoomMessagesToFile.spec.ts
```diff
@@ -0,0 +1,147 @@
...
+
+		expect(result).to.be.a.string;
+		expect(result).to.equal(
+			`<p><strong>${messagesData[0].username}</strong> (${new Date(messagesData[0].ts).toUTCString()}):<br/>\n<i>${
+				messagesData[0].msg
+			}</i>\n</p>`,
+		);
+	});
+
+	it('should correctly reference file when exporting a message object with an attachment as html', async () => {
```
MarcosSpessatto, line 78: Can we add a test case for when the message type is not italic?

MarcosSpessatto (changes requested):

apps/meteor/tests/unit/server/lib/dataExport/exportRoomMessagesToFile.spec.ts
matheusbsilva137: Wouldn't that be the same as the plain text message object? :thinking: Or do you mean having a message that has a type, but not an italic one?

matheusbsilva137: I went with the second option and added a few more expects to make sure we're using the correct text

MarcosSpessatto (dismissed):

apps/meteor/tests/unit/app/apps/server/mocks/data/messages.data.js
```diff
@@ -105,3 +109,44 @@ export const appMessageInvalidRoomMock = {
...
+		file: {
+			_id: 'txt-file-id',
+			name: 'test.txt',
+			type: 'text/plain',
+			size: 29,
+			format: '',
+		},
+		attachments: [
+			{
+				type: 'file',
```
KevLehman, line 145: also this one doesn't have a title, is that good?

apps/meteor/client/views/room/contextualBar/ExportMessages/FileExport.tsx
tiagoevanp: ok, in this case... zero problems

apps/meteor/tests/unit/app/apps/server/mocks/data/messages.data.js
matheusbsilva137: Yep, I'm testing specifically what happens when no `title` is provided

apps/meteor/tests/unit/app/apps/server/mocks/data/messages.data.js
```diff
@@ -105,3 +109,43 @@ export const appMessageInvalidRoomMock = {
...
+				title: 'test.txt',
+				title_link: '/file-upload/txt-file-id/test.txt',
+			},
+		],
+	}),
+	createFakeMessage({
```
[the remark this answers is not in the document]
matheusbsilva137: Tks for suggesting this! Done

KevLehman (approved):
Nice!

KevLehman (approved):

tiagoevanp (dismissed):

jessicaschelly (dismissed):

matheusbsilva137:
Let's not consider this a breaking change since sending an incorrect date currently generates an empty export (which shouldn't be expected by anyone). So this is more of a fix/improvement than something breaking
cc @ggazzo

matheusbsilva137:
Closing this in favor of #32364 (rebasing this to include it in 6.9, this PR is kinda messed up for me to rebase it again)

PR #32062, opened 2024-03-25
fix: New messages export overwrites previous one from the same day when using Amazon S3

matheusbsilva137:

 - Include unique id in message export zip files' names to avoid replacement instead of insertion on external storage services (due to name collision).

## Steps to test or reproduce

Set Amazon S3 as the Storage Type and sequentially export multiple rooms as files.

**Current behavior:** by accessing the oldest export sent by email and clicking the link, the downloaded file is actually related to the latest export.

 **Expected behavior:** each downloaded file should contain its corresponding export, no overwrites should be made.

## Further comments

[SUP-522](https://rocketchat.atlassian.net/browse/SUP-522)

[SUP-522]: https://rocketchat.atlassian.net/browse/SUP-522?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/meteor/server/lib/dataExport/uploadZipFile.spec.ts
```diff
@@ -0,0 +1,101 @@
...
+					type: 'application/zip',
+					size: fileStat,
+				}),
+			),
+		).to.be.true;
+
+		expect(result).to.have.property('_id', randomId);
+		expect(result).to.have.property('name').that.is.a.string;
+		const fileName: string = result.name;
+		expect(fileName.endsWith(`${userId}-data-${randomId}.zip`));
```
MarcosSpessatto, line 75:
```suggestion
		expect(fileName.endsWith(`${userId}-data-${randomId}.zip`)).to.be.true;
```

apps/meteor/server/lib/dataExport/uploadZipFile.spec.ts
```diff
@@ -0,0 +1,101 @@
...
+					type: 'application/zip',
+					size: fileStat,
+				}),
+			),
+		).to.be.true;
+
+		expect(result).to.have.property('_id', randomId);
+		expect(result).to.have.property('name').that.is.a.string;
+		const fileName: string = result.name;
+		expect(fileName.endsWith(`${userId}-${randomId}.zip`));
```
MarcosSpessatto, line 99:
```suggestion
		expect(fileName.endsWith(`${userId}-${randomId}.zip`)).to.be.true;
```

apps/meteor/server/lib/dataExport/uploadZipFile.spec.ts
MarcosSpessatto: Fixing the assertion probably will make the test break since the user display name will not be equal to userId.

apps/meteor/server/lib/dataExport/uploadZipFile.spec.ts
MarcosSpessatto: Same as before.

apps/meteor/server/lib/dataExport/uploadZipFile.spec.ts
```diff
@@ -0,0 +1,101 @@
...
+
+describe('Export - uploadZipFile', () => {
+	const randomId = 'random-id';
+	const fileStat = 100;
+	const userName = 'John Doe';
+	const userId = 'user-id';
+	const filePath = 'random-path';
+
+	beforeAll(() => {
+		stubs.findOneUserById.returns({ name: userName });
```
MarcosSpessatto, line 47: Can we add different test cases for each of the options? Since the `userDisplayName` is directly related to what this function returns and it's also directly related to the filename creation, we should also add tests for when the user doesn't have `name` and `username`.

`userDisplayName = user.name || user.username || userId;`

MarcosSpessatto (changes requested):

.changeset/twelve-seas-battle.md
```diff
@@ -0,0 +1,5 @@
+---
+"@rocket.chat/meteor": patch
+---
+
+Added file id to message exports zip file names in order to avoid overwriting previous exports generated in the same day when using external storage services (such as Amazon S3)
```
debdutdeb, line 5: I think changeset doesn't have to mention things like ids and just get to wat was fixed.
```suggestion
Fixed an issue where old exports would get overwritten by new ones if generated on the same day, when using external storage services (such as Amazon S3)
```

matheusbsilva137: Thanks for the suggestion! That's much better

apps/meteor/server/lib/dataExport/uploadZipFile.spec.ts
matheusbsilva137: Done!

apps/meteor/server/lib/dataExport/uploadZipFile.spec.ts
matheusbsilva137: I was not setting the stubs correctly :grimacing: it's working now
Thanks for noticing this

apps/meteor/tests/unit/server/lib/dataExport/uploadZipFile.spec.ts
```diff
@@ -0,0 +1,153 @@
...
+					type: 'application/zip',
+					size: fileStat,
+				}),
+			),
+		).to.be.true;
+
+		expect(result).to.have.property('_id', randomId);
+		expect(result).to.have.property('name').that.is.a.string;
+		const fileName: string = result.name;
+		expect(fileName.endsWith(`${userName}-data-${randomId}.zip`));
```
MarcosSpessatto, line 77:
```suggestion
		expect(fileName.endsWith(`${userName}-data-${randomId}.zip`)).to.be.true;
```

apps/meteor/tests/unit/server/lib/dataExport/uploadZipFile.spec.ts
```diff
@@ -0,0 +1,153 @@
...
+					type: 'application/zip',
+					size: fileStat,
+				}),
+			),
+		).to.be.true;
+
+		expect(result).to.have.property('_id', randomId);
+		expect(result).to.have.property('name').that.is.a.string;
+		const fileName: string = result.name;
+		expect(fileName.endsWith(`${userName}-${randomId}.zip`));
```
MarcosSpessatto, line 101:
```suggestion
		expect(fileName.endsWith(`${userName}-${randomId}.zip`)).to.be.true;
```

apps/meteor/tests/unit/server/lib/dataExport/uploadZipFile.spec.ts
```diff
@@ -0,0 +1,153 @@
...
+					type: 'application/zip',
+					size: fileStat,
+				}),
+			),
+		).to.be.true;
+
+		expect(result).to.have.property('_id', randomId);
+		expect(result).to.have.property('name').that.is.a.string;
+		const fileName: string = result.name;
+		expect(fileName.endsWith(`${userUsername}-${randomId}.zip`));
```
MarcosSpessatto, line 126:
```suggestion
		expect(fileName.endsWith(`${userUsername}-${randomId}.zip`)).to.be.true;
```

apps/meteor/tests/unit/server/lib/dataExport/uploadZipFile.spec.ts
```diff
@@ -0,0 +1,153 @@
...
+					type: 'application/zip',
+					size: fileStat,
+				}),
+			),
+		).to.be.true;
+
+		expect(result).to.have.property('_id', randomId);
+		expect(result).to.have.property('name').that.is.a.string;
+		const fileName: string = result.name;
+		expect(fileName.endsWith(`${userId}-${randomId}.zip`));
```
MarcosSpessatto, line 151:
```suggestion
		expect(fileName.endsWith(`${userId}-${randomId}.zip`)).to.be.true;
```

MarcosSpessatto (changes requested):

MarcosSpessatto (approved):

KevLehman (approved):

PR #32063, opened 2024-03-25
fix: UI allowing to mark room as favorite despite room was not a `default` room

KevLehman:
https://rocketchat.atlassian.net/browse/SUP-514

- Update UI to disable favorite button when `default` is not true
- Add API tests for the favorite behavior
- Also UI tests

## Further comments

This was only a UI issue, BE was ignoring the change to non default rooms.


MarcosSpessatto (changes requested):
Can we create UI tests as well? Asking, since the majority of this problem is the UI not respecting the feature at all.

apps/meteor/tests/end-to-end/api/09-rooms.js
```diff
@@ -1847,5 +1847,64 @@ describe('[Rooms]', function () {
 					expect(res.body.room).to.have.property('fname', newDiscussionName);
 				});
 		});
+
+		it('should mark a room as favorite', async () => {
```
MarcosSpessatto, line 1851: Can we make the room as default inside this `it` (as well)? I am asking because we need to run the whole `describe` block for this to pass, otherwise, if https://github.com/RocketChat/Rocket.Chat/blob/5d04e9872c7614a3ba60f70b275b836c24e6424b/apps/meteor/tests/end-to-end/api/09-rooms.js#L1767 didn't run the test will fail. I mean, the result will be the same, just to make this independent.

apps/meteor/tests/end-to-end/api/09-rooms.js
```diff
@@ -1847,5 +1847,64 @@ describe('[Rooms]', function () {
...
+				.query({
+					roomId: testChannel._id,
+				})
+				.expect(200)
+				.expect((res) => {
+					expect(res.body).to.have.property('success', true);
+					expect(res.body).to.have.property('room').and.to.be.an('object');
+
+					expect(res.body.room).to.have.property('_id', testChannel._id);
+					expect(res.body.room).to.not.have.property('favorite');
```
MarcosSpessatto, line 1906: I wonder if it doesn't make sense to throw an error instead of failing silently, what do you think?

MarcosSpessatto (changes requested):

apps/meteor/tests/end-to-end/api/09-rooms.js
KevLehman: idk, since when should an "it" be independent from the describe? I can understand describes being separated, but `it`? 🤔

apps/meteor/tests/end-to-end/api/09-rooms.js
KevLehman: It makes absolute sense, but changing that would be a breaking change and so, we're stuck with this behavior until major 😞

MarcosSpessatto: That's correct, thanks for reminding me about this.

apps/meteor/tests/end-to-end/api/09-rooms.js
MarcosSpessatto: I mean, just to be more explicit this test needs the room as default, so we can have everything inside that it, but that's ok, not a big deal. 👍

MarcosSpessatto (approved):

dougfabris (approved):
LGTM! 
Approving on behalf of frontend!

PR #32066, opened 2024-03-25
test: `InitialData.insertAdminUserFromEnv`

KevLehman:
https://rocketchat.atlassian.net/browse/CORE-194

## Further comments

- Abstract logic for creating an admin user from environment variables into function
- Unit tests for newly created function


apps/meteor/tests/unit/server/startup/initialData.tests.ts
```diff
@@ -0,0 +1,235 @@
...
+		setPasswordAsync.reset();
+		settingsGet.reset();
+		process.env.ADMIN_PASS = 'pass';
+	});
+
+	it('should do nothing if process.env.ADMIN_PASS is empty', async () => {
+		process.env.ADMIN_PASS = '';
+		const result = await insertAdminUserFromEnv();
+		expect(getUsersInRole.called).to.be.false;
+		expect(result).to.be.undefined;
```
MarcosSpessatto, line 70:
```suggestion
```

I think this is unnecessary since the function doesn't return anything either way.

apps/meteor/tests/unit/server/startup/initialData.tests.ts
```diff
@@ -0,0 +1,235 @@
...
+	});
+	it('should do nothing if theres already an admin user', async () => {
+		getUsersInRole.returns({ count: () => 1 });
+
+		const result = await insertAdminUserFromEnv();
+		expect(getUsersInRole.called).to.be.true;
+		expect(validateEmail.called).to.be.false;
```
MarcosSpessatto, line 77: Instead of asserting if this function is called since it's inside another condition, it'd be better to test if the user was created or not (`Users.create()`). WDYT?

apps/meteor/tests/unit/server/startup/initialData.tests.ts
```diff
@@ -0,0 +1,235 @@
...
+		process.env.ADMIN_EMAIL = 'email';
+		getUsersInRole.returns({ count: () => 0 });
+		validateEmail.returns(false);
+		models.Users.create.returns({ insertedId: 'newuserid' });
+
+		const result = await insertAdminUserFromEnv();
+		expect(getUsersInRole.called).to.be.true;
+		expect(validateEmail.called).to.be.true;
+		expect(validateEmail.calledWith('email')).to.be.true;
+		expect(models.Users.create.called).to.be.true;
```
MarcosSpessatto, line 90: Let's check if this was called with the proper email address.

apps/meteor/tests/unit/server/startup/initialData.tests.ts
```diff
@@ -0,0 +1,235 @@
...
+	it('should ignore the admin email when another user already has it set', async () => {
+		process.env.ADMIN_EMAIL = 'email';
+		getUsersInRole.returns({ count: () => 0 });
+		validateEmail.returns(true);
+		models.Users.create.returns({ insertedId: 'newuserid' });
+		models.Users.findOneByEmailAddress.returns({ _id: 'someuser' });
+
+		await insertAdminUserFromEnv();
+
+		expect(models.Users.findOneByEmailAddress.getCall(0).firstArg).to.not.to.have.property('email', 'email');
```
MarcosSpessatto, line 127: Wouldn't it be better to check if the user is being created with the proper email instead? Like we are doing here: https://github.com/RocketChat/Rocket.Chat/pull/32066/files#diff-bae5a355e8a182e2302bf4cb73656f039b36c9dd87884218b724e3529e1a5e9eR177

apps/meteor/tests/unit/server/startup/initialData.tests.ts
```diff
@@ -0,0 +1,235 @@
...
+		models.Users.create.returns({ insertedId: 'newuserid' });
+		models.Users.findOneByEmailAddress.returns(undefined);
+
+		await insertAdminUserFromEnv();
+
+		expect(models.Users.create.getCall(0).firstArg)
+			.to.have.property('emails')
+			.to.deep.equal([{ address: 'email', verified: true }]);
+	});
+	it('should validate a username with setting UTF9_User_Names_Validation when process.env.ADMIN_USERNAME is set', async () => {
```
MarcosSpessatto, line 156:
```suggestion
	it('should validate a username with setting UTF8_User_Names_Validation when process.env.ADMIN_USERNAME is set', async () => {
```

apps/meteor/tests/unit/server/startup/initialData.tests.ts
KevLehman: There's some tests for that case below. These ones is just to check that both these functions were not called when the env is not set.
(Actually when env is not set it shouldn't call anything, so i just picked these 2 from inside the if)

apps/meteor/tests/unit/server/startup/initialData.tests.ts
KevLehman: That's the test on L129. This one only checks that when ADMIN_EMAIL is set to any value, it will try to validate it.

L129 covers the email being valid and saved by users.create

apps/meteor/tests/unit/server/startup/initialData.tests.ts
KevLehman: ops, wrong stub

apps/meteor/tests/unit/server/startup/initialData.tests.ts
KevLehman: Yup, this is just to test that the function actually ran and returned, as there's only one return.

Kinda a dummy check, but also ensures the function just ran and didn't throw

apps/meteor/tests/unit/server/startup/initialData.tests.ts
KevLehman: :kek:

apps/meteor/tests/unit/server/startup/initialData.tests.ts
```diff
@@ -0,0 +1,235 @@
...
+		validateEmail.reset();
+		addUserRolesAsync.reset();
+		models.Users.create.reset();
+		models.Users.findOneByEmailAddress.reset();
+		setPasswordAsync.reset();
+		settingsGet.reset();
+		process.env.ADMIN_PASS = 'pass';
+	});
+
+	it('should do nothing if process.env.ADMIN_PASS is empty', async () => {
```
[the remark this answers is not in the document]
KevLehman: I think this env bypasses the password policy check 🙈 it sets the password direclty to the user, so anything can be set. I'll take a look but i'm pretty sure the above its true.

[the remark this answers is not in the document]
KevLehman: Yep, env ignores policy check :kek:

jessicaschelly (approved):

MarcosSpessatto (approved):

PR #32073, opened 2024-03-26
fix: Surface elements (Modals, Contextual Bars) fail to respond when triggered by app submit events

AllanPazRibeiro:

**Purpose:**
This PR addresses a specific issue in the app's UI behavior, particularly in the modal and contextual bar interfaces used for updates.

**Problem:**
Previously, the update modal/contextual bar did not consistently close upon the submission of updates, leading to potential confusion and a disjointed user experience.

**Solution:**
To resolve this, the PR introduces a refinement in the logic controlling the modal/contextual bar. It now retrieves the interaction type involved in the update process. This enhancement allows the system to accurately determine if and when the modal/contextual bar should close after a submission event.

## Issue(s)

- https://rocketchat.atlassian.net/browse/CORE-138
- https://rocketchat.atlassian.net/browse/SUP-459

## Steps to test or reproduce

1. Install (or create one) an app  that updates a modal/contextual
2. Trigger the action that would update the surface
3. It should update the surface instead of just closing it

## Further comments

Here is an app to test this issue:

[test-surface-changes_0.0.1.zip](https://github.com/RocketChat/Rocket.Chat/files/14932874/test-surface-changes_0.0.1.zip)
[test-surface-changes_0.0.1-not-compiled.zip](https://github.com/RocketChat/Rocket.Chat/files/14932889/test-surface-changes_0.0.1-not-compiled.zip)


apps/meteor/app/ui-message/client/ActionManager.ts
```diff
@@ -71,34 +71,41 @@ export class ActionManager implements IActionManager {
...
-		]).finally(() => {
+			});
+
+			const interaction = await sdk.rest.post(`/apps/ui.interaction/${appId}`, {
+				...userInteraction,
+				triggerId,
+			});
+
+			return this.handleServerInteraction(interaction);
+		} finally {
```
MarcosSpessatto, line 94: I'm unsure if this new implementation has the same behavior as the old one. The previous implementation used `Promise.race` to add a manual timeout to the HTTP call, with the new version it seems the `void` makes the promise result ignored and also makes the code wait for the HTTP call to be fulfilled, which in the previous version this was not true, the method finishes the execution when the first promise is fulfilled, ignoring completely the other one.

- What happens if the `reject`ion happens indeed? Would the method still throw an error to the caller function?
- Will the error being thrown to the upper layers and when the HTTP call finishes, also make the value be returned (since now we expect a value from the method)?
- Is try/finally block identical to what Promise.race/finally does speaking about the error being thrown to the upper layers?

Finally, since even the previous implementation adds some complexity and unpredictability to it, can we add a unit test (could be just for this method) just to make sure the method is doing what we expect?

MarcosSpessatto (changes requested):
Asking this, but I'm not sure if it's possible, can we add a UI test to cover this?

MarcosSpessatto (changes requested):
Would it be possible to add some UI test to ensure this behavior?

tassoevan (approved):

gabriellsh (dismissed):
Can you change the title of this PR to something more descriptive? Otherwise looking good!

Will there be tests?

MarcosSpessatto (approved):

tassoevan (approved):

PR #32083, opened 2024-03-27
chore: rename API method unauthorized method to forbidden

ggazzo:


apps/meteor/app/api/server/api.ts
```diff
@@ -612,18 +636,29 @@ export class APIClass<TBasePath extends string = ''> extends Restivus {
 									throw new Meteor.Error('invalid-params', validatorFunc.errors?.map((error: any) => error.message).join('\n '));
 								}
 							}
-							if (
-								shouldVerifyPermissions &&
-								(!this.userId ||
+							if (shouldVerifyPermissions) {
+								if (!this.userId) {
+									if (applyBreakingChanges) {
+										throw new Meteor.Error('error-unauthorized', 'You must be logged in to do this');
```
MarcosSpessatto, line 642: Shouldn't it be `error-forbidden` here?

apps/meteor/app/api/server/api.ts
```diff
@@ -678,18 +713,26 @@ export class APIClass<TBasePath extends string = ''> extends Restivus {
...
+							result = ((e: any) => {
+								switch (e.error) {
+									case 'error-too-many-requests':
+										return API.v1.tooManyRequests(typeof e === 'string' ? e : e.message);
+									case 'unauthorized':
+									case 'error-unauthorized':
+										if (applyBreakingChanges) {
+											return API.v1.unauthorized(typeof e === 'string' ? e : e.message);
+										}
+										return API.v1.forbidden(typeof e === 'string' ? e : e.message);
```
MarcosSpessatto, line 725: Isn't this in a reversed order?

MarcosSpessatto (changes requested):

apps/meteor/app/api/server/api.ts
ggazzo: It's a bit complex, but the PR is exactly about these changes.

You need to keep 2 or 3 things in mind at the same time:

- We used to call the unauthorized method (which would be correct) but passed the status code for forbidden. -- With the renaming of methods, even though it's incorrect, the natural approach is to start calling forbidden to avoid any breaking changes.

MarcosSpessatto (approved):

PR #32086, opened 2024-03-27
fix!: api login should not suggest which credential is wrong

ggazzo:
Apparently at some point the `_runLoginHandlers` function became asynchronous, thwarting our efforts to obfuscate what the reason for the login rejection was.

Instead of returning `Incorrect Password` its going to return `Unauthorized` as expected


KevLehman (commented):
You think that change from `Incorrect pass` to `Unauthorized` could be viewed as a breaking change?

ggazzo:
> You think that change from `Incorrect pass` to `Unauthorized` could be viewed as a breaking change?

it could be and we could hours discuss about. there is no way to fix it without changing the payload, so there is no other way to patch. so far its planned to be released only at 7.0 (already a major).

so the discussion if its or not a break is a bit subjective. until now, being released only on 7.0 its a break.  if someone decides to patch this, restoring the old behavior, so its a regular patch, at least the changeset will not complain about it

It is worth mentioning that this was once the default behavior, and unintentionally changed.

KevLehman (dismissed):

apps/meteor/definition/externals/meteor/accounts-base.d.ts
```diff
@@ -54,6 +54,14 @@ declare module 'meteor/accounts-base' {
 
 		const _accountData: Record<string, any>;
 
+		interface AccountsServerOptions {
+			ambiguousErrorMessages?: boolean;
+			restrictCreationByEmailDomain?: string | (() => string);
+			forbidClientAccountCreation?: boolean | undefined;
+		}
+
+		export const _options: AccountsServerOptions;
```
tiagoevanp, line 63: export a private const?

apps/meteor/app/lib/server/lib/loginErrorMessageOverride.ts
```diff
@@ -0,0 +1,16 @@
...
+import { Accounts } from 'meteor/accounts-base';
+import { Meteor } from 'meteor/meteor';
+
+const { _runLoginHandlers } = Accounts;
+
+Accounts._options.ambiguousErrorMessages = true;
+Accounts._runLoginHandlers = async function (methodInvocation, options) {
+	const result = await _runLoginHandlers.call(Accounts, methodInvocation, options);
+
+	if (result.error instanceof Meteor.Error) {
```
MarcosSpessatto, line 11: All errors here should be considered as a User not found? Asking because we do have some examples like this one: https://github.com/RocketChat/Rocket.Chat/blob/develop/apps/meteor/app/apple/server/loginHandler.ts#L36-L38
returning other things as errors.

Also what about regular `Error` (not Meteor ones), should this be considered a problem?

apps/meteor/definition/externals/meteor/accounts-base.d.ts
ggazzo: I didnt get your point. it starts with `_` but we use on our code, so its not private.
thats the reason I'm declaring this  prop actually to be used outside.

apps/meteor/app/lib/server/lib/loginErrorMessageOverride.ts
ggazzo: I was expecting at least ONE test telling me that we use some error for a different case.

ggazzo: since now we are considering this as a break change 🤷🏻‍♂️ no need to worry about

sampaiodiego (approved):

PR #32104, opened 2024-04-01
test: forward messages with files

jessicaschelly:

Creation of automated tests for Files on Forward messages

## Issue(s)
Closes: https://rocketchat.atlassian.net/browse/CORE-229


apps/meteor/tests/e2e/message-actions.spec.ts
MarcosSpessatto, line 1: Just curious:
 - Should we create a test case to send a video as well? 
 - I saw we are forwarding messages to channels only, would there be any difference if we forward them to groups, DMs, teams, etc? Asking because I saw some test cases before these forwarding to different kinds of rooms too

jessicaschelly: - Yes, I forgot about the videos! I will do the tests for it! Thank you for pointing it out

- I think it shares a lot of similarities (e.g opening the modal, forwarding, showing in the room) and I didn't want to add a lot of tests for the same thing, but maybe I could create a test named "forward pdf file to rooms" and each step be to forward to a group, dm and such. I'm not sure if that would be the right approach, but let me know what you think and if you have some suggestions!

MarcosSpessatto: I agree with you, we don't need to test cases for the same thing. Let's keep it as-is for now.

MarcosSpessatto (approved):

apps/meteor/tests/e2e/message-actions.spec.ts
```diff
@@ -150,5 +150,65 @@ test.describe.serial('message-actions', () => {
 		await poHomeChannel.sidenav.openChat(ADMIN_CREDENTIALS.username);
 		await expect(poHomeChannel.content.lastUserMessage).toContainText(message)
 	})
+
+	test('expect forward text file to channel', async () => {
+		const filename = 'any_file.txt'
+		await poHomeChannel.content.sendFileMessage(filename);
+		await poHomeChannel.content.btnModalConfirm.click();
+		await expect(poHomeChannel.content.lastUserMessage).toContainText(filename)
```
matheusbsilva137, line 158: There are some missing `;` in this file :grimacing: 
Not really an issue, but maybe it's better to have it and avoid errors in case we enable lint here at some point

matheusbsilva137 (changes requested):

apps/meteor/tests/e2e/message-actions.spec.ts
jessicaschelly: Thank you for pointing it out! Fixed now!

matheusbsilva137 (approved):

PR #32105, opened 2024-04-01
regression(fuselage-ui-kit): Use default translation namespace for `-core` apps

tassoevan:

Core apps differ from regular apps by the ID suffix `-core`. Before this change, it'd be assumed that any app might have a translation namespace available, named `app-<app ID>`. This is simple not true for core apps, as they use default namespaces (e.g. `core`) only. The absence of a core app translation namespace would cause UIKit blocks to be suspended ("blink") while a request to server for translations on the expected namespace is going on.
## Issue(s)

[CORE-227](https://rocketchat.atlassian.net/browse/CORE-227)
## Further comments

[CORE-227]: https://rocketchat.atlassian.net/browse/CORE-227?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


gabriellsh (dismissed):

MarcosSpessatto (changes requested):
Can a unit-test be added to cover this?

PR #32106, opened 2024-04-02
fix: Image gallery moves to older images when clicking on the right arrow button

matheusbsilva137:

 - Invert the image gallery's direction so that older images are displayed when clicking on the left arrow button;
 - Add e2e tests for the image gallery. 

## Steps to test or reproduce

I suggesting using more than 5 images to test so that pagination is applied.

**Before:**

https://github.com/RocketChat/Rocket.Chat/assets/36537004/6056e896-4176-4af1-95b1-c68e0a91daaa

**After:**

https://github.com/RocketChat/Rocket.Chat/assets/36537004/632a4dfd-079c-49fc-9799-6c41fe58b67e

## Further comments

[SUP-419](https://rocketchat.atlassian.net/browse/SUP-419)

[SUP-419]: https://rocketchat.atlassian.net/browse/SUP-419?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


gabriellsh (changes requested):

apps/meteor/client/components/ImageGallery/ImageGallery.tsx
```diff
@@ -140,8 +140,8 @@ export const ImageGallery = ({ images, onClose, loadMore }: { images: IUpload[];
 					<Swiper
 						ref={swiperRef}
 						navigation={{
-							nextEl: '.rcx-swiper-next-button',
-							prevEl: '.rcx-swiper-prev-button',
+							nextEl: '.rcx-swiper-prev-button',
+							prevEl: '.rcx-swiper-next-button',
```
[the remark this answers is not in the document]
matheusbsilva137: The way you implemented it looks a lot better. Thanksss
I just had to bring back `direction: 'rtl'` to fix the direction of the animations, please check it again

apps/meteor/tests/e2e/image-gallery.spec.ts
```diff
@@ -0,0 +1,73 @@
...
+	let poHomeChannel: HomeChannel;
+	let targetChannel: string;
+	// Using more than 5 images so that new images need to be loaded by the gallery
+	const imageNames = ['number1.png', 'number2.png', 'number3.png', 'number4.png', 'number5.png', 'number6.png'];
+
+	test.beforeAll(async ({ api }) => {
+		targetChannel = await createTargetChannel(api);
+	});
+
+	test.beforeEach(async ({ page }) => {
```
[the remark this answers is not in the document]
matheusbsilva137: I had to tweak some stuff to do that in beforeAll (where using page is not allowed), but it worked. Tests are now parallel

apps/meteor/tests/e2e/image-gallery.spec.ts
```diff
@@ -0,0 +1,75 @@
...
+	test.afterAll(async ({ api }) => {
+		expect((await api.post('/channels.delete', { roomName: targetChannel })).status()).toBe(200);
+	});
+
+	test('expect successfully move to older images by using the left arrow button', async () => {
+		await poHomeChannel.content.lastUserMessage.getByRole('img').click();
+
+		/* eslint-disable no-await-in-loop */
+		for (let i = 0; i < imageNames.length - 1; i++) {
+			await expect(poHomeChannel.content.nextSlideButton).toBeEnabled();
```
MarcosSpessatto, line 45: Can we ensure we are indeed moving to the older images? I mean, the description says it fixes the problem of presenting the images in a reversed way, not sure if this is possible though, but maybe ensuring it by filename? 🤔

MarcosSpessatto (changes requested):

apps/meteor/tests/e2e/image-gallery.spec.ts
matheusbsilva137: The way I'd retrieve them to do that would be kinda attached to the order too (since the swiper labels them by their indexes/order), but I think that's better than not validating it at all
Tks for the suggestion, I'll do that

matheusbsilva137: Done! Found a way to do that not attached to the order :partying_face:

gabriellsh (approved):

MarcosSpessatto (approved):

tiagoevanp (changes requested):
The e2e test added is failing.

matheusbsilva137:
Moving this to the next release since I'll need to refactor this PR. The Swiper we use is taking longer to load images when the dir prop is passed to it 🤷

MarcosSpessatto (approved):

matheusbsilva137 (commented):
FYI I decided to keep the e2e tests serial since the `beforeAll` process is heavy. It takes more time than the actual tests, so I can't see any real benefits on making the e2e tests parallel in this case
cc @gabriellsh @tiagoevanp

matheusbsilva137 (commented):
The `direction: 'rtl'` which I had to use to fix animations' direction is causing images to take longer to load for some reason. That's due to an issue on the swiper component (which has been reported [here](https://github.com/nolimits4web/swiper/issues/7493), but it is yet to be fixed)
So I decided to work on another approach (inverting the array instead of using the direction prop) so that we can have animations working properly. Can you please let me know what you think of this new approach? @gabriellsh @MarcosSpessatto

MarcosSpessatto:
> The `direction: 'rtl'` which I had to use to fix animations' direction is causing images to take longer to load for some reason. That's due to an issue on the swiper component (which has been reported [here](https://github.com/nolimits4web/swiper/issues/7493), but it is yet to be fixed) So I decided to work on another approach (inverting the array instead of using the direction prop) so that we can have animations working properly. Can you please let me know what you think of this new approach? @gabriellsh @MarcosSpessatto

It looks like the issue has been closed and the developers say the bug is not with the lib itself. But since you already provided a solution by reversing the array, LGTM.

MarcosSpessatto (approved):

tiagoevanp (approved):

PR #32121, opened 2024-04-03
test: fix `should edit name of targetChannel` flaky test

MartinSchoeler:
Exactly when the test was going to `fill()` the searchbox, the room finished loading and set the focus to the message composer, causing the search term to be written there. This should make it so it waits for the room to be loaded (also tests if the header component updated properly)

[CORE-239]

[CORE-239]: https://rocketchat.atlassian.net/browse/CORE-239?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/meteor/tests/e2e/page-objects/fragments/home-content.ts
```diff
@@ -307,4 +307,8 @@ export class HomeContent {
 	findSystemMessage(text: string): Locator {
 		return this.page.locator(`[data-qa-type="system-message-body"] >> text="${text}"`);
 	}
+	
+	get infoHeaderName(): Locator {
+		return this.page.locator('.rcx-room-header').getByRole('heading');
+	}
```
ggazzo, line 313:
```suggestion
```

PR #32122, opened 2024-04-03
feat: Migrate cloud credentials to the new collection

Gustrb:
This PR is a children of: #31986 so it must be merged before this one gets merged.

Since all the cloud credentials data must be stored inside the `rocketchat_workspace_credentials` collection, we must migrate all the values from inside the `settings` collection to their new home.

## Further comments

CONN-5


.changeset/fifty-laws-sparkle.md
```diff
@@ -0,0 +1,8 @@
+---
+"@rocket.chat/meteor": patch
+"@rocket.chat/core-typings": patch
+"@rocket.chat/model-typings": patch
+"@rocket.chat/models": patch
```
matheusbsilva137, line 5:
```suggestion
"@rocket.chat/meteor": minor
"@rocket.chat/core-typings": minor
"@rocket.chat/model-typings": minor
"@rocket.chat/models": minor
```
Since this is a feat (not a fix), we should use minor instead

apps/meteor/server/startup/migrations/v305.ts
matheusbsilva137, line 1: I believe we're trying to only add migrations in the next major release :grimacing: 
Is there any chance we could move this to the `release-7.0.0` branch?
cc @sampaiodiego

matheusbsilva137 (changes requested):

PR #32129, opened 2024-04-04
chore: new stats for statsCollector

hugocostadev:

Adding new workspaces statistics.

## Further comments

I took the opportunity to simplify the storybook that uses IStats to avoid having a big object and just sending what's being used

https://rocketchat.atlassian.net/browse/CONN-115


apps/meteor/app/statistics/server/lib/statistics.ts
```diff
@@ -185,11 +185,13 @@ export const statistics = {
...
-		statsPms.push(
-			LivechatCustomField.col.count().then((count) => {
-				statistics.totalCustomFields = count;
-			}),
-		);
+		statsPms.push((statistics.totalCustomFields = await LivechatCustomField.countDocuments({})));
+
+		// Number of public custom fields
+		statsPms.push((statistics.totalLivechatPublicCustomFields = await LivechatCustomField.countDocuments({ public: true })));
+
```
tassoevan, line 192: I think is out of scope, but I'd at least write a TODO comment asking for writing a new method on `LivechatCustomField` to get this information in a single query instead of two `$count` aggregations.

tassoevan (approved):

tassoevan (approved):

Gustrb:
Nice work!!
-511 +45

while adding new stats is amazing 😄

apps/meteor/app/statistics/server/lib/statistics.ts
Gustrb: I agree, it is probably quite simple but I don't think we need to overcomplicate this task

Gustrb (approved):

PR #32131, opened 2024-04-04
fix!: Update statuses for stuck Omnichannel rooms via migration

KevLehman:
https://rocketchat.atlassian.net/browse/CORE-238

## Further comments

Migration will do 3 steps:
- Remove the inquiries of closed rooms
- Update the status of inquiries that are queued, but the related room is already served
- Remove inquiries that point to a room that no longer exists

Since at the time a room is closed the inquiry is removed, we'll assume rooms that are already closed have all the data they require set, so we're only removing the inquiry to fix the status and unblock processing.

Queue will do the same steps for each of the inquiries it touches. Generally speaking, unless someone updates directly to this version or has been using manual selection all the time, this migration will do nothing.

[Merge when 7.0 is happening]

### DIFF
--- a/apps/meteor/server/startup/migrations/v305.ts
+++ b/apps/meteor/server/startup/migrations/v305.ts
@@ -1,0 +1,51 @@
+import { LivechatInquiryStatus } from '@rocket.chat/core-typings';
+import { LivechatInquiry, LivechatRooms } from '@rocket.chat/models';
+import type { UpdateResult } from 'mongodb';
+
+import { addMigration } from '../../lib/migrations';
+
+addMigration({
+	version: 305,
+	name: 'Remove Omnichannel stuck rooms caused by improper room closing or queue processing',
+	async up() {
+		// This migration aims to remove all the rooms that were left in an imporper state due to the closing process
+		// not being completed or the queue not updating them properly.
+		// Process is as follows:
+		// 1. Find all the closed rooms that still have an open inquiry
+		// 2. Remove the inquiry
+		// 1. Find all the inquiries that point to an invalid rooms
+		// 2. Remove the inquiry
+		// 1. Find all the inquiries that are queued but have any "takenAt" date
+		// 2. Remove the inquiry
+
+		// Closed rooms with open inquiries
+		const closedRooms = await LivechatRooms.find({ closedAt: { $exists: true } }, { projection: { _id: 1 } }).toArray();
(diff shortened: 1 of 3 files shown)


apps/meteor/server/startup/migrations/v305.ts
```diff
@@ -0,0 +1,57 @@
...
+		// 2. Remove the inquiry
+		// 1. Find all the inquiries that point to an invalid rooms
+		// 2. Remove the inquiry
+		// 1. Find all the inquiries that are queued but have any "takenAt" date
+		// 2. Remove the inquiry
+		// 1. Find all closed rooms
+		// 2. Remove subscriptions associated with them
+
+		// Closed rooms with open inquiries
+		const closedRooms = await LivechatRooms.find({ closedAt: { $exists: true } }, { projection: { _id: 1 } }).toArray();
```
geekgonecrazy, line 24: This query takes a long time for cases with a lot of omnichannel conversations.  Example with 1 million.

apps/meteor/server/startup/migrations/v305.ts
```diff
@@ -0,0 +1,57 @@
...
+		const { deletedCount } = await LivechatInquiry.deleteMany({ rid: { $in: closedRooms.map((room) => room._id) } });
```
geekgonecrazy, line 25: With same case above this could fail because of the size of the array being returned

apps/meteor/server/startup/migrations/v305.ts
```diff
@@ -0,0 +1,57 @@
...
+
+		console.log(`[Migration] Removed ${deletedCount} inquiries from closed rooms`);
+
+		// Queued inquiries pointing to a room thats served
+		const openAndServedRooms = await LivechatRooms.find({ servedBy: { $exists: true } }, { projection: { _id: 1 } }).toArray();
```
geekgonecrazy, line 30: Again will take a long time.  While servedBy does exist as index too with a lot of records this still takes a long time when has a lot of records

geekgonecrazy (changes requested):
This will take a really long time to get responses from mongo on these queries for large omnichannel use cases

KevLehman:
Yep, that's a known thing, as i said over there, that's why this is still a draft 👀

KevLehman:
Closing as no longer needed

PR #32139, opened 2024-04-05
chore: change cloud timeouts

heitortanoue:

Currently all calls to cloud.rocket.chat from Rocket.Chat in the sync process have a 3s timeout.  We need to bump to 5s to give us a bit of wiggle room.  If we don’t respond fast enough the sync fails.  While we keep working on making our response faster we need Rocket.Chat to be a bit more tolerant

## Further comments

CONN-109


Gustrb (approved):
looks good to me :p

sampaiodiego (approved):

PR #32141, opened 2024-04-05
chore: Deprecate `channels.images` in favor of `rooms.images`

KevLehman:
https://rocketchat.atlassian.net/browse/CORE-231

## Further comments

Deprecate `channels.images` in favor of `rooms.images`. `Rooms` endpoints are more broad and should interact with all types of rooms. `Channels` on the other hand are specific to public channels.
This change is to keep the semantics and conventions of the endpoints


MarcosSpessatto (dismissed):

apps/meteor/app/api/server/v1/channels.ts
```diff
@@ -806,7 +806,7 @@ API.v1.addRoute(
 
 API.v1.addRoute(
 	'channels.images',
-	{ authRequired: true, validateParams: isChannelsImagesProps },
+	{ authRequired: true, validateParams: isRoomsImagesProps, deprecationVersion: '7.0.0' },
```
sampaiodiego, line 809: looks like the `addRoute` API doesn't support the "info" param from `apiDeprecationLogger.endpoint` so there is no way to recommend a new endpoint since this one is deprecated.. 

I can see two options in this case:

- We add the support to this API to pass additional info, it could be something like:
```ts
API.v1.addRoute(
	'channels.images',
	{ authRequired: true, validateParams: isRoomsImagesProps, deprecationInfo: 'Use /api/v1/rooms.images instead.', deprecationVersion: '7.0.0' },
     // ...
);
```
- Or we don't use this API and call `apiDeprecationLogger.endpoint` direct from the endpoint's body:
```ts

API.v1.addRoute(
	'channels.images',
	{ authRequired: true, validateParams: isRoomsImagesProps },
	{
		async get() {
			// ...
			apiDeprecationLogger.endpoint(this.request.route, '7.0.0', this.response, 'Use /v1/rooms.images instead.');
			// ...
		}
	}
);
```

what do you think? I think either are fine.

sampaiodiego (changes requested):

apps/meteor/app/api/server/v1/channels.ts
KevLehman: Option 2 was fine, went with that one 😬

apps/meteor/app/api/server/v1/channels.ts
```diff
@@ -806,7 +806,7 @@ API.v1.addRoute(
 
 API.v1.addRoute(
 	'channels.images',
-	{ authRequired: true, validateParams: isChannelsImagesProps },
+	{ authRequired: true, validateParams: isRoomsImagesProps, deprecationVersion: '7.0.0', deprecationInfo: 'Use /v1/rooms.images instead' },
```
KevLehman, line 809:
```suggestion
	{ authRequired: true, validateParams: isRoomsImagesProps, deprecationVersion: '7.0.0', deprecationInfo: 'Use /v1/rooms.images instead.' },
```

sampaiodiego (approved):
sharing here something Kevin mentioned to me that would be a good idea to change the "deprecation API" to receive the new endpoint as a param and then compose the log message by it self, so we're sure all deprecations are following the same pattern..

this will be accomplished in a follow up PR though 😉 :right:

MarcosSpessatto (approved):

dougfabris (approved):
LGTM! Approving it on behalf of frontend team

Issue #31920, opened 2024-03-06
Decimal number can't be converted into italic format.


Himanshu-kiran:
**How to reproduce?**
->Simply try to put any number and try to play with changing format.
->You can see any number can be converted to italic but decimal number or text with point can't be.

**Actual behaviour:**
Number is getting shown with underscore.

**Expected:**
Decimal number should also be converted to italic
**Additional:**
![image](https://github.com/RocketChat/Rocket.Chat/assets/133533054/d34a4836-1aa8-4c80-8f4b-0b53148d3840)

Himanshu-kiran:
Hey @Sing-Li ,I want to allow text with period(.) to be italicised. 
But changing regex in markdown file is also not useful  anyway:
![image](https://github.com/RocketChat/Rocket.Chat/assets/133533054/26896df2-1799-447d-9403-63bc5d31ec49)
Can you please help me finding real code//file responsible for italic or bold formatting.

Visbhavesh:
I wanted to work in thus issue please Assign this issue to me.

PR #32146, opened 2024-04-06
fix: Italic for decimal numbers and strings with one dot ( #31920)

joanamarquees:

The conversion of decimal numbers to italics wasn't working. After testing several italic conversions, I realised that the problem occurred when putting numbers with just a dot in the middle or even words containing a dot. I then realised that this was because the text was being parsed according to another rule included in the grammar.pegjs file.

After looking at various solutions, I was able to come up with a proposal that solves the initial problem. The rule blocking the conversion was the AutoLink, so now the autoLink verifies if it is supposed to be an italic conversion or if it is a link.

## Issue(s)

 #31920

## Steps to test or reproduce

Steps already described on the issue page.


joanamarquees:
@sampaiodiego @ggazzo I would like someone to check this pull request, if possible. Thank you

PR #32152, opened 2024-04-08
fix: cleaning the room filter between navigation on the admin section

AllanPazRibeiro:

**Resolve Room Filter Persistence Issue with Local Storage**

- Addressed a bug where room filter settings were only partially resetting during admin page navigation.
- Now, all filters are completely cleared.

## Issue(s)

https://rocketchat.atlassian.net/browse/CORE-217

## Steps to test or reproduce

1. Go to admin page
2. Go to Room section
3. Change the filter
4. Navigate to another section
5. Go back to the Room filter, the filter should be cleaned

## Further comments

https://github.com/RocketChat/Rocket.Chat/assets/35040806/054ee962-1903-483b-bddd-c7f02189eff9


.changeset/brown-lobsters-join.md
```diff
@@ -0,0 +1,7 @@
+---
+'@rocket.chat/meteor': patch
+---
+
+Update: Fixed Room Filter Persistence in Admin Navigation
```
KevLehman, line 5:
```suggestion
```

apps/meteor/client/providers/SearchFiltersProvider/SearchFilterProvider.spec.tsx
```diff
@@ -0,0 +1,87 @@
+/* eslint-disable no-await-in-loop */
```
KevLehman, line 1: Why this if you can use `for await (of)` loop?

apps/meteor/client/providers/SearchFiltersProvider/SearchFilterProvider.spec.tsx
```diff
@@ -0,0 +1,87 @@
...
+
+		const firstCheckbox = await screen.findByLabelText(initialTypes[0].text);
+		userEvent.click(firstCheckbox);
+
+		const updatedFirstCheckbox = await screen.findByLabelText(initialTypes[0].text);
+		expect(updatedFirstCheckbox).toBeChecked();
+	});
+
+	test('handles searchText and types in context correctly', async () => {
+		const initialTypes: OptionProp[] = [
```
KevLehman, line 69:
```suggestion
		const initialTypes = [
```

apps/meteor/client/views/admin/rooms/RoomsTable.tsx
```diff
@@ -16,41 +15,43 @@ import {
...
 
 	const { sortBy, sortDirection, setSort } = useSort<'name' | 't' | 'usersCount' | 'msgs' | 'default' | 'featured'>('name');
 	const { current, itemsPerPage, setItemsPerPage, setCurrent, ...paginationProps } = usePagination();
-	const searchText = useDebouncedValue(roomFilters.searchText, 500);
+	const searchText = useDebouncedValue(searchFilters.searchText, 500);
 
 	const query = useDebouncedValue(
 		useMemo(() => {
-			if (searchText !== prevRoomFilterText.current) {
+			const filtersChanged =
```
KevLehman, line 39: Just checking if it's good that when prevRoomFilters & searchFilters have different order but same keys, this to be `true`

apps/meteor/tests/e2e/admin-room.spec.ts
```diff
@@ -0,0 +1,117 @@
...
+	});
+	test('should display the Rooms Table', async ({ page }) => {
+		await expect(page.locator('[data-qa-type="PageHeader-title"]')).toContainText('Rooms');
+		await expect(page.locator(`[qa-room-name="${privateRoom}"]`)).toBeVisible();
+		await expect(page.locator(`[qa-room-name="${channel}"]`)).toBeVisible();
+	});
+
+	test('should filter room by name', async ({ page }) => {
+		const input = page.locator('input[type="text"]');
+		console.log('🚀 ~ test ~ channel:', channel);
```
KevLehman, line 34:
```suggestion
```

apps/meteor/tests/e2e/admin-room.spec.ts
```diff
@@ -0,0 +1,117 @@
...
+		const privateOption = page.locator('text=Private channels');
+
+		await privateOption.waitFor();
+		await privateOption.click();
+
+		await expect(page.locator(`[qa-room-name="${privateRoom}"]`)).toBeVisible();
+	});
+
+	test('should filter rooms by type and name', async ({ page }) => {
+		const input = page.locator('input[type="text"]');
```
KevLehman, line 55: any possibility this ends up falling in other input?

apps/meteor/tests/e2e/admin-room.spec.ts
```diff
@@ -0,0 +1,117 @@
...
+		await privateOption.click();
+
+		await expect(page.locator(`[qa-room-name="${privateRoom}"]`)).toBeVisible();
+	});
+
+	test('should be empty in case of the search does not find any room', async ({ page }) => {
+		const input = page.locator('input[type="text"]');
+
+		await input.click();
+		await input.fill('Wrong-channel');
```
KevLehman, line 75: [nitpick] If someone creates a `wrong-channel` on a prior test, then this will return value and will fail 😬 
would suggest to use a more unique value, like a random value

KevLehman (changes requested):

apps/meteor/client/providers/SearchFiltersProvider/SearchFilterProvider.spec.tsx
tiagoevanp: or follow this pattern instead of `as [type]`. I think is better to have the type described before the object itself. Many languages has the type as a first thing...

apps/meteor/client/providers/SearchFiltersProvider/SearchFilterProvider.tsx
```diff
@@ -0,0 +1,16 @@
+import type { FC, ReactNode } from 'react';
+import React, { useState } from 'react';
+
+import type { SearchFilters } from '../../contexts/SearchFilterContext';
+import { SearchFilterContext } from '../../contexts/SearchFilterContext';
+
+type SearchFilterProviderProps = {
+	children: ReactNode;
```
tiagoevanp, line 8: AFAIK: FC already have children described on it 🤔... there is no necessity of it here...

apps/meteor/tests/e2e/admin-room.spec.ts
tiagoevanp: same thought... how about to use a name for the input and search for it instead of type?

tiagoevanp (changes requested):
I'm thinking here, you have one single filter to use in the context. If I want to use the same context for another screen, I will kill the previous filter with the new one state, right? Is this the desired behavior? 🤔

apps/meteor/tests/e2e/admin-room.spec.ts
tiagoevanp: you can use faker API... with a unic ID concatenated

tiagoevanp (changes requested):

tiagoevanp (approved):

apps/meteor/tests/e2e/admin-room.spec.ts
```diff
@@ -0,0 +1,116 @@
...
+
+		channel = await createTargetChannel(api);
+		privateRoom = await createTargetPrivateChannel(api);
+	});
+	test('should display the Rooms Table', async ({ page }) => {
+		await expect(page.locator('[data-qa-type="PageHeader-title"]')).toContainText('Rooms');
+	});
+
+	test('should filter room by name', async ({ page }) => {
+		const input = page.locator('[data-qa-id="AdminRoomSearchInput"]');
```
jessicaschelly, line 32: This locator already exists inside `apps/meteor/tests/e2e/page-objects/admin.ts`
You could call it like this after initializing:

```
await admin.inputSearchRooms.click();
await admin.inputSearchRooms.fill(channel);
```

apps/meteor/tests/e2e/admin-room.spec.ts
```diff
@@ -0,0 +1,116 @@
...
+		await expect(page.locator('text=Private Channel').first()).toBeVisible();
+	});
+
+	test('should filter rooms by type and name', async ({ page }) => {
+		const input = page.locator('[data-qa-id="AdminRoomSearchInput"]');
+
+		await input.click();
+		await input.fill(privateRoom);
+
+		const dropdown = page.locator('[data-qa-id="AdminRoomDropdownInput"]');
```
jessicaschelly, line 58: You could add this locator inside

`apps/meteor/tests/e2e/page-objects/admin.ts`

apps/meteor/tests/e2e/admin-room.spec.ts
```diff
@@ -0,0 +1,116 @@
...
+		await privateOption.waitFor();
+		await privateOption.click();
+
+		await expect(page.locator('text=No results found')).toBeVisible();
+	});
+
+	test('should filter rooms by type and name and keep the filter after changing section', async ({ page }) => {
+		const input = page.locator('[data-qa-id="AdminRoomSearchInput"]');
+
+		await input.click();
```
jessicaschelly, line 90: Can you check if the click is necessary before the fill()? I think we could remove it

jessicaschelly (changes requested):

apps/meteor/client/views/admin/rooms/RoomsTableFilters.tsx
```diff
@@ -78,21 +92,22 @@ const RoomsTableFilters = ({ setFilters }: { setFilters: Dispatch<SetStateAction
...
 					alignItems='center'
 					placeholder={t('Search_rooms')}
 					addon={<Icon name='magnifier' size='x20' />}
 					onChange={handleSearchTextChange}
-					value={text}
+					value={prevRoomFilters.current.searchText}
 				/>
 			</Box>
-			<Box minWidth='x224' m='x4'>
+			<Box minWidth='x224' m='x4' data-qa-id='AdminRoomDropdownInput'>
```
jessicaschelly, line 104: It would be good to use roles in this case, maybe @dougfabris could help us on the best way to do it here

apps/meteor/tests/e2e/admin-room.spec.ts
```diff
@@ -0,0 +1,116 @@
...
+		await privateOption.click();
+
+		await expect(page.locator(`[qa-room-name="${privateRoom}"]`)).toBeVisible();
+	});
+
+	test('should be empty in case of the search does not find any room', async ({ page }) => {
+		const input = page.locator('[data-qa-id="AdminRoomSearchInput"]');
+
+		await input.click();
+		const wrongChannel = faker.string.alpha(10);
```
jessicaschelly, line 73: We could change the name to "nonExistingChannel" since the channel is not "wrong". This might help clarifies that the channel does not exist rather than being inherently "wrong."

jessicaschelly (changes requested):

apps/meteor/tests/e2e/admin-room.spec.ts
```diff
@@ -0,0 +1,108 @@
...
+		await admin.inputSearchRooms.fill(privateRoom);
+
+		await admin.dropdownFilterRoomType.click();
+
+		const privateOption = page.locator('text=Private channels');
+
+		await privateOption.waitFor();
+		await privateOption.click();
+
+		await page.goto('/admin/info');
```
jessicaschelly, line 93: Is this step necessary? or could it just go directly to /rooms?

apps/meteor/client/views/admin/rooms/RoomsTableFilters.tsx
dougfabris: yeah, first thing I can see, this select is not navigable by keyboard at all. I would suggest adding at least a button role on the anchor and adding the proper button keyboard event. With the role I believe we can get the locator using `locator.getByRole` instead of the custom `data-qa-id` prop

.changeset/brown-lobsters-join.md
```diff
@@ -0,0 +1,5 @@
+---
+'@rocket.chat/meteor': patch
```
MarcosSpessatto, line 2:
```suggestion
'@rocket.chat/meteor': minor
```

apps/meteor/tests/e2e/admin-room.spec.ts
```diff
@@ -0,0 +1,114 @@
...
+		const selectDropdown = await admin.dropdownFilterRoomType('Rooms (1)');
+		await expect(selectDropdown).toBeVisible();
+
+		await selectDropdown.click();
+
+		const isChecked = await privateOption.isChecked();
+
+		expect(isChecked).toBe(true);
+
+		await expect(admin.inputSearchRooms).toHaveValue(privateRoom);
```
MarcosSpessatto, line 112: Can we assert if the rooms in the list were filtered properly together with the filter dropdown?

MarcosSpessatto (changes requested):

jessicaschelly (approved):

apps/meteor/tests/e2e/admin-room.spec.ts
```diff
@@ -0,0 +1,103 @@
...
+		const privateOption = page.locator('text=Private channels');
+
+		await privateOption.waitFor();
+		await privateOption.click();
+
+		await page.goto('/admin/info');
+
+		await page.goto('/admin/rooms');
+
+		const selectDropdown = await admin.dropdownFilterRoomType('Rooms (1)');
```
MarcosSpessatto, line 100: Would not it be better to check if the default (All rooms) was displayed?

apps/meteor/.meteorMocks/index.ts
MarcosSpessatto, line 1: I think we are not using these mocks anymore, are we? 🤔

MarcosSpessatto (changes requested):

apps/meteor/tests/unit/client/views/admin/rooms/RoomsTable.spec.tsx
```diff
@@ -0,0 +1,114 @@
+/* eslint-disable no-restricted-properties */
```
KevLehman, line 1:
```suggestion
```

apps/meteor/tests/unit/client/views/admin/rooms/RoomsTable.spec.tsx
```diff
@@ -0,0 +1,114 @@
...
+	it('renders no results when data is empty', async () => {
+		useQueryStub.returns({
+			isLoading: false,
+			data: {
+				rooms: [],
+			},
+			error: undefined,
+			isSuccess: true,
+		});
+		render(<RoomsTable reload={{ current: sinon.fake() }} />, { wrapper: createQueryClientWrapper() });
```
KevLehman, line 82: Since you're passing fakes to the calls, maybe you can assert those are being called on the proper places too? 👀

apps/meteor/tests/unit/client/views/admin/rooms/RoomsTable.spec.tsx
```diff
@@ -0,0 +1,114 @@
...
+
+jest.mock('../../../../../../client/views/admin/rooms/RoomRow', () => {
+	return ({ someProp }: any) => (
+		<tr data-id='RoomRow'>
+			<td>MockedRoomRow {someProp}</td>
+		</tr>
+	);
+});
+
+const createQueryClientWrapper = () => {
```
KevLehman, line 23: This would be useful as a lib for other tests to also use 👀

MarcosSpessatto (approved):

jessicaschelly (dismissed):

apps/meteor/client/views/admin/rooms/RoomsTable.tsx
```diff
@@ -121,7 +121,7 @@ const RoomsTable = ({ reload }: { reload: MutableRefObject<() => void> }): React
 				<GenericTable>
 					<GenericTableHeader>{headers}</GenericTableHeader>
 					<GenericTableBody>
-						<GenericTableLoadingTable headerCells={mediaQuery ? 6 : 3} />
+						<GenericTableLoadingTable data-id='RoomGenericTableLoadingTable' headerCells={mediaQuery ? 6 : 3} />
```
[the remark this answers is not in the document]
gabriellsh: we should not add `data-id` or similar things for test purposes only.

MarcosSpessatto (approved):

apps/meteor/.meteorMocks/index.ts
```diff
@@ -1,5 +1,164 @@
...
+
+		equals(key, value) {
+			const currentValue = this.get(key);
+			return currentValue === value;
+		}
+
+		all() {
+			return Object.fromEntries(this.dict);
+		}
+	},
```
gabriellsh, line 163: I'm very unsure about this. I don't have a suggestion here unfortunately but I don't think it's correct to do it. I'm thinking this is to avoid `Meteor` sub-imports from breaking the tests?

apps/meteor/client/views/admin/rooms/RoomsTableFilters.tsx
```diff
@@ -78,6 +88,7 @@ const RoomsTableFilters = ({ setFilters }: { setFilters: Dispatch<SetStateAction
 		>
 			<Box minWidth='x224' display='flex' m='x4' flexGrow={2}>
 				<TextInput
+					data-testid='AdminRoomSearchInput'
```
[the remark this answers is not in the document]
dougfabris: Indeed, it's possible to query this input by role

packages/ui-client/src/components/MultiSelectCustom/MultiSelectCustom.tsx
```diff
@@ -105,8 +105,9 @@ export const MultiSelectCustom = ({
 		<Box display='flex' flexGrow={1} position='relative'>
 			<MultiSelectCustomAnchor
 				ref={reference}
-				onClick={toggleCollapsed as any}
 				collapsed={collapsed}
+				onClick={() => toggleCollapsed(!collapsed)}
+				onKeyDown={(e) => (e.code === 'Enter' || e.code === 'Space') && toggleCollapsed(!collapsed)}
```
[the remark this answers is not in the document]
dougfabris: @gabriellsh in order to allow the `MultiSelectCustom` be accessible by keyboard we're adding the `toggleCollapse` action to the keyboard event as well

packages/ui-client/src/components/MultiSelectCustom/MultiSelectCustomAnchor.tsx
```diff
@@ -1,66 +1,44 @@
...
-			pb={10}
-			pi={16}
-			color={isDirty ? Palette.text['font-default'].toString() : Palette.text['font-annotation'].toString()}
-			className={inputStyle}
+			className={['rcx-input-box__wrapper', customStyle].filter(Boolean)}
 			{...props}
 		>
 			{isDirty ? `${t(selectedOptionsTitle)} (${selectedOptionsCount})` : t(defaultTitle)}
-			<Icon name='chevron-down' fontSize='x20' color='hint' />
+			<Icon name={collapsed ? 'chevron-up' : 'chevron-down'} fontSize='x20' color='hint' />
```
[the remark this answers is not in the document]
dougfabris: @gabriellsh This component was not accessible and hasn't any focus state, this change helps with that

packages/ui-client/src/components/MultiSelectCustom/MultiSelectCustom.tsx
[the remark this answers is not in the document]
MarcosSpessatto: Wouldn't it be better to add this behavior in another PR if it's not related to the fix itself?

packages/ui-client/src/components/MultiSelectCustom/MultiSelectCustomAnchor.tsx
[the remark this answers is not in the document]
MarcosSpessatto: Wouldn't it be better to add this behavior in another PR if it's not related to the fix itself?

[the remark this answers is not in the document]
MarcosSpessatto: Wouldn't it be better to add this behavior in another PR if it's not related to the fix itself?

packages/ui-client/src/components/MultiSelectCustom/MultiSelectCustom.tsx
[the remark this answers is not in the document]
MarcosSpessatto: Wouldn't it be better to add this behavior in another PR if it's not related to the fix itself?

MarcosSpessatto (changes requested):

dougfabris:
@MarcosSpessatto All those changes was made in order to be able to get good locators for the requested e2e tests. Despite is not related directly I think it somehow related since we're improving the filters tests. So if we decide to remove those changes we'll have to add tests with bad locators. So IMO we have two paths: 
- Add the filters changes **AND** add good locators
- Remove the filters changes **AND** add bad locators

MarcosSpessatto:
> **Related to the `MultiSelectCustom`** @MarcosSpessatto All those changes was made in order to be able to get good locators for the requested e2e tests. Despite it's not related directly I think it somehow related since we're improving the filters tests. So if we decide to remove those changes we'll have to add tests with bad locators. So IMO we have two paths:
> 
> * Add the filters changes **AND** add good locators
> * Remove the filters changes **AND** add bad locators

Agree with your approach guys, we can go ahead.  Thanks for the explanation.

MarcosSpessatto (dismissed):

apps/meteor/client/views/admin/rooms/RoomsTable.tsx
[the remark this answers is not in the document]
gabriellsh: This is a bad practice for tests.
https://playwright.dev/docs/best-practices#best-practices
https://testing-library.com/docs/queries/about/#priority

Basically you should always emulate how a user would select elements on a page, or maybe a screen reader. By adding these ids you're directly selecting the element that could not be accessible otherwise.

apps/meteor/client/views/admin/rooms/RoomsTableFilters.tsx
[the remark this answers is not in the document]
gabriellsh: This should also be removed, as it hasn't been yet

apps/meteor/tests/e2e/admin-room.spec.ts
```diff
@@ -0,0 +1,103 @@
...
+	test('should filter rooms by type and name', async ({ page }) => {
+		await admin.inputSearchRooms.fill(privateRoom);
+
+		const dropdown = await admin.dropdownFilterRoomType();
+		await dropdown.click();
+
+		const privateOption = page.locator('text=Private channels');
+
+		await privateOption.waitFor();
+		await privateOption.click();
```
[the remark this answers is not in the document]
gabriellsh: So? Did you try it?

packages/mock-providers/src/MockedQueryClientWrapper.tsx
```diff
@@ -0,0 +1,12 @@
+import { QueryClient, QueryClientProvider } from '@tanstack/react-query';
+import React from 'react';
+
+export const MockedQueryClientWrapper = () => {
```
gabriellsh, line 4: I don't understand why did you rename this function?

gabriellsh (changes requested):

gabriellsh (approved):
LGTM

MarcosSpessatto (approved):

PR #32159, opened 2024-04-09
fix!: api login should not suggest which credential is wrong

ggazzo:


ZhaoQi99:
@ggazzo Can you optimize the frontend message? It may confuse users when an existing user enters the wrong password.

PR #32164, opened 2024-04-09
chore: Change `deprecationVersion` prop in favor of new `deprecation` prop

KevLehman:
## Further comments

New `deprecation` prop has the following format
```
deprecation?: { version: string, alternatives?: string[] }
```
This allow devs to pass a list of alternative endpoints that they can use instead of the deprecated one.

Alternatives is an array thinking on the cases where an endpoint may be separated onto 2 different endpoints/handlers. Not so common, but possible


apps/meteor/app/api/server/api.helpers.ts
```diff
@@ -101,3 +102,10 @@ export function checkPermissions(options: { permissionsRequired?: PermissionsReq
 	// If reached here, options.permissionsRequired contained an invalid payload
 	return false;
 }
+
+export function parseDeprecation(methodThis: any, { alternatives, version }: { version: string; alternatives?: string[] }): void {
+	const infoMessage = alternatives?.length
+		? `Please use the alternative(s): ${alternatives?.length === 1 ? alternatives[0] : alternatives?.join(',')}`
```
sampaiodiego, line 108: I guess we need an space in the beginning 🤔  

and since at this is point we're sure it is an array, we can just use `.join()` 😉 
```suggestion
		? ` Please use the alternative(s): ${alternatives.join(',')}`
```

sampaiodiego (changes requested):

sampaiodiego (approved):

aleksandernsilva (approved):

PR #32170, opened 2024-04-10
fix: Language priority order not respected when changing server setting

### LINKED ISSUE #32008: User/Global translation not working most of the time
### Description:

User selected translation not working most of the time (defaults to english).

### Steps to reproduce:

Change default language (user or global) to something other than english (polish in my case).

Do reload the page a few times - it's rendered in english.

### Expected behavior:

Proper language displayed.

### Actual behavior:

Even though language file is being loaded - UI is rendered with English.

### Server Setup Information:
Official docker setup - 6.6.3

### Client Setup Information

- Desktop App or Browser Version: Browser
- Operating System: Docker

gabriellsh:

Stop forcing language switch when changing setting. Priority should be:
User Preference > Browser Language > Server Language

In most cases Server language is just a fallback, and that will mostly happen in Desktop browser. If the user has not set a language in his preferences, notifications will respect the server's setting.

image

## Issue(s)

[SUP-493](https://rocketchat.atlassian.net/browse/SUP-493)
Closes #32008
## Steps to test or reproduce

Set a language in the user preferences
Go to admin > General > Language and change to another one
The language will switch to the one set on the setting instead of keeping the user preference.

## Further comments

[SUP-493]: https://rocketchat.atlassian.net/browse/SUP-493?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/meteor/tests/e2e/translations.spec.ts
```diff
@@ -0,0 +1,51 @@
...
+
+        expect((await setUserPreferences(api, { language: 'pt-BR' })).status()).toBe(200);
+
+        await expect(page.locator('h2')).toHaveText('Bem-vindo ao Rocket.Chat');
+    });
+
+    test.describe('Browser language', () => {
+        test.use({ locale: 'pt-BR' })
+        test('expect to respect browser language', async ({ page }) => {
+            // await expect(page.locator('html')).toHaveAttribute('lang', 'pt-BR');
```
MarcosSpessatto, line 35:
```suggestion
```

apps/meteor/tests/e2e/translations.spec.ts
```diff
@@ -0,0 +1,51 @@
...
+            await expect(page.locator('h2')).toHaveText('Bem-vindo ao Rocket.Chat');
+        });
+    })
+
+    test.describe.skip('Server language', () => {
+        test.use({ locale: '' })
+        test('expect to respect server language', async ({ page, api }) => {
```
MarcosSpessatto, line 42: Can we add 2 (more) test cases to ensure the page is not changing the language when the user has a language preference and another one for when they have a language in their browser too?

Asking that because I remember the mentioned bug happened because the whole page was translated whenever the user changed the server language.

MarcosSpessatto (changes requested):

apps/meteor/tests/e2e/translations.spec.ts
gabriellsh: I can add the tests, but that was not a "bug" since the code was specifically implemented to change the languages when switching the setting (and only happens through the UI). I removed that bit of code. Also, I had tried to add tests for the server language, but I cannot test the server language since playwright's browser will always have a locale set, and there's no way to change that.

I actually removed this one in the latest commits.

MarcosSpessatto (approved):

PR #32176, opened 2024-04-11
feat: Apps-Engine method to read multiple messages from a room

Dnouv:

Added a new method, `getMessages` to fetch messages from a specific room. This method accepts a `roomId` and an `options` object as parameters. The `options` object can contain `limit, skip, and sort` properties to control the query. The method uses the `Messages.find` function to fetch the messages from the database and converts them using a message raw converter obtained from the orch object.

Since a room can have a large number of messages, this method returns a maximum of 100 messages in a single call, to read more messages one can make another call passing the `skip` value. The default sort order is the descending order of timestamp (meaning the most recent messages will be returned), which can be overridden by passing the `sort` parameter.

Added a new converter `convertMessageRaw` it removes a few fields from the following (reason: [comment](https://github.com/RocketChat/Rocket.Chat/pull/32176#discussion_r1607241913) ):
```
sender: {
  id: string,
  username: string,
  name?: string
},
editor: {
  id: string,
  username: string
},
room: {
  id: string
}
```

Additionally, this new method removes the `delete` since the delete operator is generally slow and can cause performance issues. It modifies the underlying object, which can lead to deoptimization in JavaScript engines.

## Steps to test or reproduce

To test as well as reproduce the issue and solution, try running the below in the App.
```ts
// before this PR
const roomMessage = await read
                .getRoomReader()
                .getMessages(message.room.id)

// after this PR 
const roomMessage = await read
                .getRoomReader()
                .getMessages(message.room.id, {
                    limit: 3,
                    sort: { createdAt: 'asc' },
                    skip: 1,
                });
```
Output:
(Before this PR):
- Error method not implemented

(After this PR):
- Array of messages

You can pass your own query to filter the messages you want to retrieve. The available options are:

- `limit`: The number of messages to retrieve. (Default: 100 and Max: 100)
- `skip`: The number of messages to skip.
- `sort`: The sort order of the messages. e.g. `{ ts: -1 }` will sort the messages by timestamp in descending order.

## Further comments

- [ADR](https://adr.rocket.chat/0063) {For Internal Use}
- [JIRA-28](https://rocketchat.atlassian.net/browse/SAFE-28)

<details>
<summary>Execution Stats on a room with 3M messages</summary>

```
{
  explainVersion: '1',
  queryPlanner: {
    namespace: 'meteor.rocketchat_message',
    indexFilterSet: false,
    parsedQuery: {
      '$and': [
        {
          rid: {
            '$eq': 'GENERAL'
          }
        },
        {
          _hidden: {
            '$not': {
              '$eq': true
            }
          }
        },
        {
          t: {
            '$not': {
              '$exists': true
            }
          }
        }
      ]
    },
    queryHash: '19FB9CF7',
    planCacheKey: 'E315B320',
    maxIndexedOrSolutionsReached: false,
    maxIndexedAndSolutionsReached: false,
    maxScansToExplodeReached: false,
    winningPlan: {
      stage: 'LIMIT',
      limitAmount: 100,
      inputStage: {
        stage: 'FETCH',
        filter: {
          '$and': [
            {
              _hidden: {
                '$not': {
                  '$eq': true
                }
              }
            },
            {
              t: {
                '$not': {
                  '$exists': true
                }
              }
            }
          ]
        },
        inputStage: {
          stage: 'IXSCAN',
          keyPattern: {
            rid: 1,
            ts: 1,
            _updatedAt: 1
          },
          indexName: 'rid_1_ts_1__updatedAt_1',
          isMultiKey: false,
          multiKeyPaths: {
            rid: [],
            ts: [],
            _updatedAt: []
          },
          isUnique: false,
          isSparse: false,
          isPartial: false,
          indexVersion: 2,
          direction: 'forward',
          indexBounds: {
            rid: [
              '["GENERAL", "GENERAL"]'
            ],
            ts: [
              '[MinKey, MaxKey]'
            ],
            _updatedAt: [
              '[MinKey, MaxKey]'
            ]
          }
        }
      }
    },
    rejectedPlans: [
      {
        stage: 'LIMIT',
        limitAmount: 100,
        inputStage: {
          stage: 'FETCH',
          filter: {
            '$and': [
              {
                t: {
                  '$not': {
                    '$exists': true
                  }
                }
              },
              {
                _hidden: {
                  '$not': {
                    '$eq': true
                  }
                }
              }
            ]
          },
          inputStage: {
            stage: 'IXSCAN',
            keyPattern: {
              rid: 1,
              t: 1,
              'u._id': 1
            },
            indexName: 'rid_1_t_1_u._id_1',
            isMultiKey: false,
            multiKeyPaths: {
              rid: [],
              t: [],
              'u._id': []
            },
            isUnique: false,
            isSparse: false,
            isPartial: false,
            indexVersion: 2,
            direction: 'forward',
            indexBounds: {
              rid: [
                '["GENERAL", "GENERAL"]'
              ],
              t: [
                '[null, null]'
              ],
              'u._id': [
                '[MinKey, MaxKey]'
              ]
            }
          }
        }
      },
      {
        stage: 'LIMIT',
        limitAmount: 100,
        inputStage: {
          stage: 'FETCH',
          filter: {
            '$and': [
              {
                _hidden: {
                  '$not': {
                    '$eq': true
                  }
                }
              },
              {
                t: {
                  '$not': {
                    '$exists': true
                  }
                }
              }
            ]
          },
          inputStage: {
            stage: 'IXSCAN',
            keyPattern: {
              rid: 1,
              tcount: 1
            },
            indexName: 'rid_1_tcount_1',
            isMultiKey: false,
            multiKeyPaths: {
              rid: [],
              tcount: []
            },
            isUnique: false,
            isSparse: false,
            isPartial: false,
            indexVersion: 2,
            direction: 'forward',
            indexBounds: {
              rid: [
                '["GENERAL", "GENERAL"]'
              ],
              tcount: [
                '[MinKey, MaxKey]'
              ]
            }
          }
        }
      }
    ]
  },
  executionStats: {
    executionSuccess: true,
    nReturned: 100,
    executionTimeMillis: 4,
    totalKeysExamined: 100,
    totalDocsExamined: 100,
    executionStages: {
      stage: 'LIMIT',
      nReturned: 100,
      executionTimeMillisEstimate: 0,
      works: 101,
      advanced: 100,
      needTime: 0,
      needYield: 0,
      saveState: 0,
      restoreState: 0,
      isEOF: 1,
      limitAmount: 100,
      inputStage: {
        stage: 'FETCH',
        filter: {
          '$and': [
            {
              _hidden: {
                '$not': {
                  '$eq': true
                }
              }
            },
            {
              t: {
                '$not': {
                  '$exists': true
                }
              }
            }
          ]
        },
        nReturned: 100,
        executionTimeMillisEstimate: 0,
        works: 100,
        advanced: 100,
        needTime: 0,
        needYield: 0,
        saveState: 0,
        restoreState: 0,
        isEOF: 0,
        docsExamined: 100,
        alreadyHasObj: 0,
        inputStage: {
          stage: 'IXSCAN',
          nReturned: 100,
          executionTimeMillisEstimate: 0,
          works: 100,
          advanced: 100,
          needTime: 0,
          needYield: 0,
          saveState: 0,
          restoreState: 0,
          isEOF: 0,
          keyPattern: {
            rid: 1,
            ts: 1,
            _updatedAt: 1
          },
          indexName: 'rid_1_ts_1__updatedAt_1',
          isMultiKey: false,
          multiKeyPaths: {
            rid: [],
            ts: [],
            _updatedAt: []
          },
          isUnique: false,
          isSparse: false,
          isPartial: false,
          indexVersion: 2,
          direction: 'forward',
          indexBounds: {
            rid: [
              '["GENERAL", "GENERAL"]'
            ],
            ts: [
              '[MinKey, MaxKey]'
            ],
            _updatedAt: [
              '[MinKey, MaxKey]'
            ]
          },
          keysExamined: 100,
          seeks: 1,
          dupsTested: 0,
          dupsDropped: 0
        }
      }
    }
  },
  command: {
    find: 'rocketchat_message',
    filter: {
      rid: 'GENERAL',
      _hidden: {
        '$ne': true
      },
      t: {
        '$exists': false
      }
    },
    limit: 100,
    '$db': 'meteor'
  },
  serverParameters: {
    internalQueryFacetBufferSizeBytes: 104857600,
    internalQueryFacetMaxOutputDocSizeBytes: 104857600,
    internalLookupStageIntermediateDocumentMaxSizeBytes: 104857600,
    internalDocumentSourceGroupMaxMemoryBytes: 104857600,
    internalQueryMaxBlockingSortMemoryUsageBytes: 104857600,
    internalQueryProhibitBlockingMergeOnMongoS: 0,
    internalQueryMaxAddToSetBytes: 104857600,
    internalDocumentSourceSetWindowFieldsMaxMemoryBytes: 104857600,
    internalQueryFrameworkControl: 'trySbeRestricted'
  },
  ok: 1,
  '$clusterTime': {
    clusterTime: Timestamp({ t: 1721412111, i: 1 }),

  },
  operationTime: Timestamp({ t: 1721412111, i: 1 })
}
```

</details>


apps/meteor/app/apps/server/bridges/rooms.ts
```diff
@@ -102,6 +102,34 @@ export class AppRoomBridge extends RoomBridge {
 		return this.orch.getConverters()?.get('users').convertById(room.u._id);
 	}
 
+	protected async getMessages(
+		roomId: string,
+		appId: string,
```
d-gubert, line 107: Please put the `appId` as the last parameter for convention

d-gubert (changes requested):

apps/meteor/app/apps/server/bridges/rooms.ts
Dnouv: Actually, the `options` is optional, and TS complains if we keep optional before a required param 😕

.changeset/polite-foxes-repair.md
```diff
@@ -0,0 +1,5 @@
+---
+'@rocket.chat/meteor': minor
+---
+
+Added the `getMessages` method in the Apps Engine, this method will allow Apps to read room messages. You can pass your own query to filter the messages you want to retrieve. With flexibility of viewing up to 100 messages at once, skipping over content you’ve already seen, and sorting messages to find what you need quickly.
```
d-gubert, line 5:
```suggestion
Added a method to the Apps-Engine that allows apps to read multiple messages from a room
```

d-gubert (approved):

apps/meteor/app/apps/server/bridges/rooms.ts
```diff
@@ -102,6 +102,34 @@ export class AppRoomBridge extends RoomBridge {
...
+			skip,
+			sort,
+		};
+
+		const messages = await Messages.findVisibleByRoomId(roomId, messageQueryOptions).toArray();
+
+		const messageConverter = this.orch.getConverters()?.get('messages');
+		if (!messageConverter) {
+			throw new Error('Message converter not found');
+		}
```
debdutdeb, line 128: if you are checking for this I'd assume there are chances of this happening? last I remember there wasn't. If required, put this before the model call, let's help the db as much as we can ;)

debdutdeb (changes requested):

apps/meteor/app/apps/server/bridges/rooms.ts
```diff
@@ -102,6 +102,34 @@ export class AppRoomBridge extends RoomBridge {
...
+		appId: string,
+	): Promise<IMessage[]> {
+		this.orch.debugLog(`The App ${appId} is getting the messages of the room: "${roomId}"`);
+		const { limit, skip, sort = { ts: -1 } } = options;
+
+		const messageQueryOptions = {
+			limit: Math.min(limit, 100),
+			skip,
+			sort,
+		};
```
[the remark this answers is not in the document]
Dnouv: Nice catch! I think it makes sense to handle just the `limit` in this case; as for `skip,` it won't affect the expected results. In the long run, we should look into a solution across all the bridge methods.

apps/meteor/app/apps/server/bridges/rooms.ts
Dnouv: Done, thank you!

apps/meteor/app/apps/server/bridges/rooms.ts
```diff
@@ -102,6 +102,38 @@ export class AppRoomBridge extends RoomBridge {
...
+			limit: number;
+			skip?: number;
+			sort?: Record<string, 1 | -1>;
+		},
+		appId: string,
+	): Promise<IMessage[]> {
+		this.orch.debugLog(`The App ${appId} is getting the messages of the room: "${roomId}"`);
+		if (typeof options.limit !== 'number') {
+			options.limit = 100;
+		}
```
debdutdeb, line 117: You'd still have to check if limit is NaN since `typeof NaN == 'number'`, `Math.min(NaN, 100)` would still return NaN.

```suggestion
		if (typeof options.limit !== 'number' || Number.isNaN(options.limit)) {
			options.limit = 100;
		}
```

debdutdeb (changes requested):

apps/meteor/app/apps/server/bridges/rooms.ts
debdutdeb: Hm actually, there is the special case of 0 as well. Math.min(0, 100) means 0 means all documents again. Negative limit shouldn't be a big deal but might want to not allow that. AFAIK it's driver dependent.

debdutdeb:
```suggestion
		if (typeof options.limit !== 'number' || Number.isNaN(options.limit) || options.limit <= 0) {
			options.limit = 100;
		}
```

debdutdeb: Let's also make sure we carry this over to the unread messages pr 🙏🏼

debdutdeb: -Infinity < 0 is true, but I didn't know isFinite is a thing. Thanks.

debdutdeb: -Infinity < 0 is true. But didn't know isFinite is a thing. Thanks.

apps/meteor/app/apps/server/bridges/rooms.ts
```diff
@@ -102,6 +102,58 @@ export class AppRoomBridge extends RoomBridge {
...
+			rid: roomId,
+			_hidden: { $ne: true },
+			t: { $exists: false },
+		};
+
+		const cursor = Messages.find(query, messageQueryOptions);
+
+		const messagePromises: Promise<IMessage>[] = [];
+		await cursor.forEach((message) => {
+			messagePromises.push(messageConverter.convertMessage(message));
```
d-gubert, line 151: I think we will need to rethink how we fetch the messages here. The message converter itself does at least 2 more queries to the DB - 3 if the message has been edited - which can hit pretty hard if the workspace is very active.

if we have 100 messages returned by mongo, we're looking at an average of 400 other requests sent via the network :grimacing: 

@Dnouv do you have any ideas how we could circumvent this?

The one thing I can think of is to simply passthrough the queried fields (`room`, `sender`, `editor`) without querying for the full data, leaving it to the app developer to do the queries if they need to.

The message converter, for reference https://github.com/RocketChat/Rocket.Chat/blob/33cd125419f254f5093b3de9f68ab13f534e3a28/apps/meteor/app/apps/server/converters/messages.js#L39-L74

d-gubert (changes requested):

apps/meteor/app/apps/server/bridges/rooms.ts
Dnouv: Added a new converter `convertMessageRaw` it removes a few fields from the following:
```
sender: {
  id: string,
  username: string,
  name?: string
},
editor: {
  id: string,
  username: string
},
room: {
  id: string
}
```

Thank you!

apps/meteor/app/apps/server/converters/messages.js
```diff
@@ -14,6 +14,59 @@ export class AppMessagesConverter {
...
+			alias: 'alias',
+			file: 'file',
+			customFields: 'customFields',
+			groupable: 'groupable',
+			token: 'token',
+			blocks: 'blocks',
+			room: getRoom,
+			editor: getEditor,
+			attachments: getAttachments,
+			sender: getSender,
```
d-gubert, line 64: IMO those fields do not need fancy conversions, we can just pass them through, i.e.:
```
roomId: 'rid',
editor: 'editedBy',
sender: 'u',
```

And then they require a new type on the apps-engine like `IMessageRaw`. This is important so it doesn't create the expectation on the app developer that the `room` field will have all the info because it is the same name and type of `IMessage.room`

d-gubert (changes requested):

apps/meteor/app/apps/server/converters/messages.js
```diff
@@ -14,6 +14,59 @@ export class AppMessagesConverter {
...
+			return undefined;
+		}
+
+		const { rid, editedBy, attachments, u, ...message } = msgObj;
+
+		const getRoom = () => ({ id: rid });
+
+		const getEditor = () => (editedBy ? { id: editedBy._id, username: editedBy.username } : undefined);
+
+		const getAttachments = async () => this._convertAttachmentsToApp(attachments);
```
d-gubert, line 28: Whew, just checked and `_convertAttachmentsToApp` doesn't do any queries to the DB :relieved:

packages/apps/src/converters/IAppMessagesConverter.ts
```diff
@@ -10,4 +10,6 @@ export interface IAppMessagesConverter {
 	convertAppMessage(message: undefined | null): Promise<undefined>;
 	convertAppMessage(message: IAppsMessage): Promise<IMessage | undefined>;
 	convertAppMessage(message: IAppsMessage | undefined | null): Promise<IMessage | undefined>;
+	convertMessageRaw(message: IMessage): Promise<IAppsMessage>;
+	converMessageRaw(message: IMessage | undefined | null): Promise<IAppsMessage | undefined>;
```
d-gubert, line 14:
```suggestion
	convertMessageRaw(message: IMessage | undefined | null): Promise<IAppsMessage | undefined>;
```

d-gubert, line 14:
```suggestion
	convertMessageRaw(message: IMessage | undefined | null): Promise<IAppsMessage | undefined>;
```

Dnouv: Thank you, totally missed the "t"!

apps/meteor/app/apps/server/converters/messages.js
Dnouv: Yes, it doesn't make any DB calls.

d-gubert (dismissed):

apps/meteor/app/apps/server/bridges/rooms.ts
```diff
@@ -102,6 +102,55 @@ export class AppRoomBridge extends RoomBridge {
...
+			sort?: Record<string, 1 | -1>;
+		},
+		appId: string,
+	): Promise<IMessageRaw[]> {
+		this.orch.debugLog(`The App ${appId} is getting the messages of the room: "${roomId}"`);
+
+		let { limit, skip = 0, sort } = options;
+
+		if (!Number.isFinite(limit) || limit < 1) {
+			limit = 100;
```
ggazzo, line 119: if the numbers are invalid why not throw some error instead?

ggazzo (changes requested):

apps/meteor/app/apps/server/bridges/rooms.ts
Dnouv: We can but we didn't.

There is going to be a default error from TS and compiler when the Numbers are invalid unless someone asserts something like `limit: "string" as unknown as Number` which was intentional. So taking the approach of default behavior and returning the max limit `100`.

apps/meteor/app/apps/server/bridges/rooms.ts
```diff
@@ -102,6 +102,55 @@ export class AppRoomBridge extends RoomBridge {
...
+	protected async getMessages(
+		roomId: string,
+		options: {
+			limit: number;
+			skip?: number;
+			sort?: Record<string, 1 | -1>;
+		},
+		appId: string,
+	): Promise<IMessageRaw[]> {
+		this.orch.debugLog(`The App ${appId} is getting the messages of the room: "${roomId}"`);
```
KevLehman, line 114: This one should login the pagination options passed as well, so we can track what the app was doing

apps/meteor/app/apps/server/bridges/rooms.ts
```diff
@@ -102,6 +102,55 @@ export class AppRoomBridge extends RoomBridge {
...
+			sort,
+		};
+
+		const query = {
+			rid: roomId,
+			_hidden: { $ne: true },
+			t: { $exists: false },
+		};
+
+		const cursor = Messages.find(query, messageQueryOptions);
```
KevLehman, line 147: The description talks about `findVisibleByRoomId` but here it's just a normal find 👀

apps/meteor/app/apps/server/bridges/rooms.ts
tassoevan: Forget about the compiler, since the app developer still can just use JS. Why we can't throw a runtime error here? Do we have a policy of being merciful at logic/calculation errors at production? If that's so, at least we should log warnings for the developer at development mode (if there is one). Otherwise, throwing a runtime error is a warranty for future changes not being blocked by this kind of leniency.

debdutdeb: Agree with Gazzo and Tasso. Not to clutter the comment section, but adding a different perspective - we at SRE have had tons of problems due to no-logs. One, had literally yesterday (not core related). There needs to be a 1:1 correlation between what we see and what we get, exceptions with verbosity for easier detection. IOW we should be able to figure out some path from reading logs, get some idea.

A log, or throw error is good. Throw error might be better as then we just don't compute anything for wrong code/calculation.

Dnouv: Thank you for the review.

No, we don't have a specific policy of being merciful. 😄 

The `limit` is optional on the Apps Engine end. You can just call `getMessages(roomed),`, and it will return "100" messages. So even if you call with a `NaN` we just return the default.

I think logging the events would be a nice idea.

apps/meteor/app/apps/server/bridges/rooms.ts
Dnouv: Thank you! Nice idea.

apps/meteor/app/apps/server/bridges/rooms.ts
Dnouv: Done, thank you!

apps/meteor/app/apps/server/bridges/rooms.ts
d-gubert: @debdutdeb @tassoevan those are good points. There were no discussions or hard stances on not throwing an error at this point, I think @Dnouv just didn't do it this way.

apps/meteor/app/apps/server/bridges/rooms.ts
```diff
@@ -102,6 +102,55 @@ export class AppRoomBridge extends RoomBridge {
...
+
+		if (!Number.isFinite(limit) || limit < 1) {
+			limit = 100;
+		}
+
+		if (!Number.isFinite(skip) || skip < 0) {
+			skip = 0;
+		}
+
+		if (!sort || typeof sort !== 'object') {
```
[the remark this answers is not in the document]
Dnouv: That's a good idea, but maybe not for the first iteration. Let's keep this in the ADR as a future improvement.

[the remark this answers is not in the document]
debdutdeb: I will even propose the question, a **raw** sort, is that something we even need?

right now, somebody to **really* sort, again would have to read core code to know the shorthand field names, which kinda is confusing when app engines' types have a more intuitive version (u vs user, e.g.).

from my perspective, if sort must be like that (a raw mongo sort payload) it should also be "converted", giving an intuitive interface for developers, not a mixture, and obviously Kev's suggestion included. Can't have random queries potentially slowing things down, and giving us a hard time troubleshooting. This sort thing if is too hard to do right now, could actually be a future improvement, which will be non-breaking.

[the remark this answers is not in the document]
d-gubert: Having to send use rocketchat's fields on the sort is indeed a problem, creates confusion and makes it harder to improve without breaking. I think @KevLehman's suggestion is a good solution.

@Dnouv maybe we can break the sort down to 2 fields:
```
{
  //..... other props
  sort: {
    field: 'createdAt',
    direction: 'asc' | 'desc',
  },
}
```

Using the direction with a `1 | -1` can be natural to mongodb users, but definitely not for people who don't.

`createdAt` in the apps-engine is `ts` in rocketchat-land, is and indexed field and based on the indices I've seen seems to be the only to field to make sense in this context. @KevLehman @debdutdeb thoughts?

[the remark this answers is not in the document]
Dnouv: Sounds like a solution to me. 🤔 

Maybe we can simplify it a bit and have it as a part of Apps Engine, not as a rocketchat-land converter (that way we don't have to check again and do those error/warning throw on the core side), wdyt @d-gubert :
```
{
  //..... other props
  sort: {
   'createdAt': 'asc' | 'desc',
  },
}
```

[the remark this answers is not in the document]
KevLehman: I'm happy with the approach. I think `ts` is the only field that makes sense in the context as well, but others may be added if needed so the structure should allow this without major code changes 🥳 

About the `promise.all` thing that Deb mentioned, I think that deserves its own thread so if he agrees, @debdutdeb you can move that comment to a top level one and we can resolve this one, wdyt?

apps/meteor/app/apps/server/bridges/rooms.ts
Dnouv: Removed the checks, since we make sure on Apps Engine side to return a valid number. I will note the decision on ADR as well. Thanks!

apps/meteor/app/apps/server/bridges/rooms.ts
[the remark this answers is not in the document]
debdutdeb: Sounds good.

apps/meteor/app/apps/server/bridges/rooms.ts
debdutdeb: This isn't resolved yet, moving the same code to apps engine makes no difference

https://github.com/RocketChat/Rocket.Chat.Apps-engine/pull/770/files#diff-b4205e5b20daca4b90bb2ad8f137fce587773eea26bf35fb7df28196a327ea52R33-R46

it's still neither logging nor throwing error. What difference is it making?

Dnouv:
Depends on: https://github.com/RocketChat/Rocket.Chat.Apps-engine/pull/770

apps/meteor/app/apps/server/bridges/rooms.ts
Dnouv: We log what `options` are passed. https://github.com/RocketChat/Rocket.Chat/blob/f35fa9a037fc4698f419eb4d4e6a43ef50089698/apps/meteor/app/apps/server/bridges/rooms.ts#L114

> What difference is it making?

On the bridge code (rc side) we know the `limit` will always be a valid number. We were limiting the value to `100` on Apps Engine's accessor if no `limit` is passed from `user` so extended it further to validate a number as well.

Dnouv: I previously used the way to go with default value instead of throwing an error because we do similar for API calls: https://github.com/RocketChat/Rocket.Chat/blob/fd2aee8f8839b02bf14decd44911a5dc9766e7f6/apps/meteor/app/api/server/helpers/getPaginationItems.ts#L6 

In my opinion, it provides a better user experience.

Dnouv: We throw an error now, 
```
if (typeof options.limit !== 'undefined' && (!Number.isFinite(options.limit) || options.limit > 100)) {
            throw new Error(`Invalid limit provided. Expected number <= 100, got ${options.limit}`);
        }
```

apps/meteor/app/apps/server/bridges/rooms.ts
```diff
@@ -103,6 +105,35 @@ export class AppRoomBridge extends RoomBridge {
...
+			throw new Error('Message converter not found');
+		}
+
+		const messageQueryOptions: FindOptions<ICoreMessage> = {
+			limit,
+			skip,
+			sort,
+		};
+
+		const query = {
```
KevLehman, line 124: Just being sure we have an index covering this and it won't get to the `rid` index

Dnouv: Yes, we have an index that covers this query. The existing compound index `{ rid: 1, t: 1, 'u._id': 1 }` is being used effectively for this query.

The `executionStatus` for a sample Query.
```
    winningPlan: {
      stage: 'FETCH',
      filter: {
        '$and': [
          {
            t: {
              '$not': {
                '$exists': true
              }
            }
          },
          {
            _hidden: {
              '$not': {
                '$eq': true
              }
            }
          }
        ]
      },
      inputStage: {
        stage: 'IXSCAN',
        keyPattern: {
          rid: 1,
          t: 1,
          'u._id': 1
        },
        indexName: 'rid_1_t_1_u._id_1',
        isMultiKey: false,
        multiKeyPaths: {
          rid: [],
          t: [],
          'u._id': []
        },
        isUnique: false,
        isSparse: false,
        isPartial: false,
        indexVersion: 2,
        direction: 'forward',
        indexBounds: {
          rid: [
            '["GENERAL", "GENERAL"]'
          ],
          t: [
            '[null, null]'
          ],
          'u._id': [
            '[MinKey, MaxKey]'
          ]
        }
      }
    },
```

Dnouv: Execution time on >3M messages room.

```
executionStats: {
    executionSuccess: true,
    nReturned: 100,
    executionTimeMillis: 4,
    totalKeysExamined: 100,
    totalDocsExamined: 100,
```

That was fast. Thanks to Kev's nice script, it was easy to fill a room with messages.

d-gubert (approved):

KevLehman (approved):

tassoevan (approved):

PR #32180, opened 2024-04-11
test: `channel-management` e2e tests improvements

dougfabris:


apps/meteor/tests/e2e/channel-management.spec.ts
```diff
@@ -57,69 +57,81 @@ test.describe.serial('channel-management', () => {
...
-	// FIXME: bad assertion
-	test('should add "user1" to "targetChannel"', async () => {
+	test('should add user1 to targetChannel', async () => {
 		await poHomeChannel.sidenav.openChat(targetChannel);
 		await poHomeChannel.tabs.btnTabMembers.click();
 		await poHomeChannel.tabs.members.showAllUsers();
 		await poHomeChannel.tabs.members.addUser('user1');
 
-		await expect(poHomeChannel.toastSuccess).toBeVisible();
+		await expect(poHomeChannel.getSystemMessageByText('added user1')).toBeVisible();
```
jessicaschelly, line 66: Maybe we could add an extra validation to ensure the user is in the room, WDYT? 
e.g: expect that the user is in the channel list

apps/meteor/tests/e2e/channel-management.spec.ts
```diff
@@ -57,69 +57,81 @@ test.describe.serial('channel-management', () => {
...
-	});
-
-	test.fixme('should mute "user1"', async () => {
+	test('should mute user1', async () => {
 		await poHomeChannel.sidenav.openChat(targetChannel);
 		await poHomeChannel.tabs.btnTabMembers.click();
 		await poHomeChannel.tabs.members.showAllUsers();
 		await poHomeChannel.tabs.members.muteUser('user1');
+
+		await expect(poHomeChannel.getSystemMessageByText('muted user1')).toBeVisible();
```
jessicaschelly, line 75: Same here, we could ensure that the user can't actually write in the channel after being muted

apps/meteor/tests/e2e/channel-management.spec.ts
```diff
@@ -57,69 +57,81 @@ test.describe.serial('channel-management', () => {
...
 
 
-	test.fixme('should set "user1" as moderator', async () => {
+	test('should set "user1" as moderator', async () => {
 		await poHomeChannel.sidenav.openChat(targetChannel);
 		await poHomeChannel.tabs.btnTabMembers.click();
 		await poHomeChannel.tabs.members.showAllUsers();
 		await poHomeChannel.tabs.members.setUserAsModerator('user1');
+
+		await expect(poHomeChannel.getSystemMessageByText('set user1 as moderator')).toBeVisible();
```
jessicaschelly, line 94: I would suggest doing the same thing here, an extra validation that he has "moderator" powers, but I'm not sure how we could easily do that besides opening a new browser with the user and checking the channel 😁 so I'll just leave this "half" suggestion

jessicaschelly (approved):

ggazzo (approved):

PR #32186, opened 2024-04-11
chore: Increase `agenda` time in Omnichannel EE jobs to 1 minute

KevLehman:
## Further comments

This is to reduce the frequency of these actions. These 3 schedulers were left behind when we moved away from synced-cron to our own cron package. Other task will be created to move these schedulers to use it in the future.

### DIFF
--- a/apps/meteor/ee/app/livechat-enterprise/server/lib/AutoCloseOnHoldScheduler.ts
+++ b/apps/meteor/ee/app/livechat-enterprise/server/lib/AutoCloseOnHoldScheduler.ts
@@ -33,6 +33,7 @@ class AutoCloseOnHoldSchedulerClass {
 			mongo: (MongoInternals.defaultRemoteCollectionDriver().mongo as any).client.db(),
 			db: { collection: SCHEDULER_NAME },
 			defaultConcurrency: 1,
+			processEvery: '1 minute',
 		});
 
 		await this.scheduler.start();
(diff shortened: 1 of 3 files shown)


MarcosSpessatto:
Since we are skipping the test covering this, we will create unit tests to cover the whole class in a subsequent PR.

MarcosSpessatto (approved):

ggazzo (changes requested):
There is no way this could be a patch release, There is no problem being solved here, at most we are masking something worse. I don't believe we should release this as a fix.

Furthermore, the title points out to something different.

Skipping a test is not ideal, you could make this modifiable via an envvar

KevLehman:
Problem being solved is that the default timing for OM jobs is 5 secs, while the default timer for any other Agenda job (via @rocket.chat/cron) is 1 minute. These items were left when refactoring from old jobs to /cron package.

While I'd agree there's no "problem" being solved, the sole fact these jobs are going to the db every 5s (which also caused problems in the past when we first migrated to /cron) is problematic by itself.

What title would you suggest? Title says we're increasing agenda time, maybe that part is not clear but what we're actually doing is increasing the agenda processing time.

About the test, yeah, skipping is not ideal, but since this is targetted for an older version, we decided to not to add more stuff in here and just skip for now. Idea is to make these jobs configurable in the future, just not now.

wdyt?

ggazzo:
I didn't understand the part about targeting an old version, as it is pointing to develop

about the title, `chore`, generally does not indicate fix, but some internal modification. Doesn't sound like a real fix

maybe something like fix: Omnichannel Jobs running every 5 seconds

KevLehman:
All right, updated PR title. I actually didn't notice I put this one as a chore. Fixed now.

The part about and old version is from a customer that's getting bombed by this. When we first moved to `cron` package, we had the same issue of DB getting bombed by agenda and so the fix was to update it to run every min. These 3 jobs were left behind on that fix (probably since the fix was spot on Open and we don't have these running there) and so the same issue persisted for these, and now a customer is having problems because of it (not all the problems are tied to these, but fixing them would alliviate a bit)

ggazzo (approved):

PR #32194, opened 2024-04-12
feat: Apps-Engine method for reading and counting unread room messages for a user

Dnouv:

- Method to fetch room unread messages based on `userId` and `roomId`
- Method to fetch the count of unread messages based on `userId` and `roomId`

## Steps to test or reproduce

```ts
const userId = "W9Ms3XifoMwWemv4Q"
const roomId = "GENERAL"
const unreadMsg = await read
            .getMessageReader()
            .getUnreadByRoomAndUser(roomId, userId);

// call with custom limit and sort order
        const limitedunreadMsg = await read
           .getRoomReader()
            .getUnreadByUser(
               roomId,
                userId,
                { sort: { createdAt: "asc" }, limit: 10 }
            )


// Get total unread messages count
const totalUnread = await read
                .getUserReader()
                .getUserUnreadMessageCount(
                    roomId,
                    userId
                );
```

## Further comments

- [JIRA-29](https://rocketchat.atlassian.net/browse/SAFE-29)
- [RCAI-13](https://rocketchat.atlassian.net/browse/RCAI-13)


Dnouv:
Removing milestone `6.8` since I can see this getting blocked by https://github.com/RocketChat/Rocket.Chat/pull/32176 (both PRs have the same approach)

Thank you!

.changeset/chilled-files-relate.md
```diff
@@ -0,0 +1,8 @@
+---
+'@rocket.chat/meteor': minor
+---
+
+Added two methods in the Apps Engine:
+
+1. A method to retrieve unread messages for a user from a specified room.
+2. A method to calculate the total number of unread messages for a user in a specified room.
```
d-gubert, line 8:
```suggestion
Added methods to the Apps-Engine that interact with unread messages
```

Please keep changeset messages short and concise. Detailed info should be put in dedicated documentation if necessary

apps/meteor/app/apps/server/bridges/rooms.ts
```diff
@@ -245,6 +245,64 @@ export class AppRoomBridge extends RoomBridge {
...
+
+		const messageConverter = this.orch.getConverters()?.get('messages');
+		if (!messageConverter) {
+			throw new Error('Message converter not found');
+		}
+
+		const [room, subscription] = await Promise.all([
+			Rooms.findOneById(roomId, { projection: { _id: 1 } }),
+			Subscriptions.findOneByRoomIdAndUserId(roomId, userId, { projection: { ls: 1 } }),
+		]);
```
d-gubert, line 259: This is an interesting approach, I'm only concerned about "wasted" cases, i.e. if the room does not exist, a subscription to it would not exist as well - which means the query for a subscription is useless.

@MarcosSpessatto do you have any thoughts on this?

d-gubert (changes requested):

apps/meteor/app/apps/server/bridges/rooms.ts
Dnouv: Some stats for better tradeoff decision (for a count of 136 unread messages).

Concurrent approach:
```
Execution Times (excluding first run): 1.965ms, 5.788ms, 2.717ms, 1.631ms, 2.808ms
Average = (1.965 + 5.788 + 2.717 + 1.631 + 2.808) / 5 = 2.9818ms
```

Sequential Approach:
```
Execution Times (excluding first run): 9.735ms, 7.949ms, 4.642ms, 5.005ms, 7.303ms, 3.948ms, 4.997ms
Average = (9.735 + 7.949 + 4.642 + 5.005 + 7.303 + 3.948 + 4.997) / 7 = 6.2256ms
```

Concurrent approach is roughly twice as fast as the sequential approach.

d-gubert: Yes, it will be faster, I'm more concerned with resource consumption and high DB loads - cpu cycles are "wasted" waiting on the promise and it needlessly contributes to he workload of the database (in those specific cases). However, the queries you're doing use indexes and transfer little data.

I might be overthinking this, I'll just get a third opinion here just in case.

Dnouv: Well, interesting, it makes sense; I will use the sequential approach here.

d-gubert (dismissed):

apps/meteor/app/apps/server/bridges/rooms.ts
```diff
@@ -245,6 +245,60 @@ export class AppRoomBridge extends RoomBridge {
...
+
+		const sort: Sort = options.sort?.createdAt ? { ts: options.sort.createdAt } : { ts: 1 };
+
+		const cursor = Messages.findVisibleByRoomIdBetweenTimestampsNotContainingTypes(roomId, lastSeen, new Date(), [], {
+			...options,
+			sort,
+		});
+
+		const messages = await cursor.toArray();
+		return Promise.all(messages.map((msg) => messageConverter.convertMessageRaw(msg)));
```
debdutdeb, line 277: map could happen on the cursor iirc. 

return cursor.map(yourFunc).toArray();

apps/meteor/app/apps/server/bridges/rooms.ts
```diff
@@ -245,6 +245,60 @@ export class AppRoomBridge extends RoomBridge {
...
+		this.orch.debugLog(`The App ${appId} is getting the unread messages count of the room: "${roomId}" for the user: "${uid}"`);
+		const user = await Users.findOneById(uid, { projection: { _id: 1 } });
+		if (!user) {
+			throw new Error('User not found');
+		}
+
+		const room = await Rooms.findOneById(roomId, { projection: { _id: 1 } });
+		if (!room) {
+			throw new Error('Room not found');
+		}
```
debdutdeb, line 290: I wonder the value of these. 

sure these add specific errors, but wouldn't `no relation found between room and user` be sufficient too? let the dev check user/room themselves (maybe they could be doing that already and duplicating).

if the roomid or userid is wrong, there wouldnt be a subscription either. we aren't checking for any sort of permission here.

.changeset/chilled-files-relate.md
```diff
@@ -0,0 +1,5 @@
+---
+'@rocket.chat/meteor': minor
+---
+
+Added methods to the Apps-Engine that interact with unread messages
```
KevLehman, line 5: Changesets should be in present tense

(i didnt change this but somehow im the one enforcing it lel)

apps/meteor/app/apps/server/bridges/rooms.ts
```diff
@@ -245,6 +245,60 @@ export class AppRoomBridge extends RoomBridge {
...
 
+	protected async getUnreadByUser(roomId: string, userId: string, options: GetMessagesOptions, appId: string): Promise<Array<IMessageRaw>> {
+		this.orch.debugLog(`The App ${appId} is getting the unread messages for the user: "${userId}" in the room: "${roomId}"`);
+
+		const messageConverter = this.orch.getConverters()?.get('messages');
+		if (!messageConverter) {
+			throw new Error('Message converter not found');
+		}
+
+		const room = await Rooms.findOneById(roomId, { projection: { _id: 1 } });
```
KevLehman, line 256:
```suggestion
		const room = await Rooms.findOneById<Pick<...>>(roomId, { projection: { _id: 1 } });
```

apps/meteor/app/apps/server/bridges/rooms.ts
```diff
@@ -245,6 +245,60 @@ export class AppRoomBridge extends RoomBridge {
...
+
+		if (!room) {
+			throw new Error('Room not found');
+		}
+
+		const subscription = await Subscriptions.findOneByRoomIdAndUserId(roomId, userId, { projection: { ls: 1 } });
```
KevLehman, line 262:
```suggestion
		const subscription = await Subscriptions.findOneByRoomIdAndUserId<Pick<...>>(roomId, userId, { projection: { ls: 1 } });
```
if the func does not support rn, ignore this one

apps/meteor/app/apps/server/bridges/rooms.ts
```diff
@@ -245,6 +245,60 @@ export class AppRoomBridge extends RoomBridge {
...
+			sort,
+		});
+
+		const messages = await cursor.toArray();
+		return Promise.all(messages.map((msg) => messageConverter.convertMessageRaw(msg)));
+	}
+
+	protected async getUserUnreadMessageCount(roomId: string, uid: string, appId: string): Promise<number> {
+		this.orch.debugLog(`The App ${appId} is getting the unread messages count of the room: "${roomId}" for the user: "${uid}"`);
+		const user = await Users.findOneById(uid, { projection: { _id: 1 } });
```
KevLehman, line 282:
```suggestion
		const user = await Users.findOneById<Pick...>(uid, { projection: { _id: 1 } });
```

apps/meteor/app/apps/server/bridges/rooms.ts
KevLehman: I do think you can just receive a uid+rid and check on the subs collection directly. If no value is there, that means user or room are invalid, or that they have no relation between them, which should be enough

apps/meteor/app/apps/server/bridges/rooms.ts
```diff
@@ -245,6 +245,60 @@ export class AppRoomBridge extends RoomBridge {
...
+		}
+
+		const subscription = await Subscriptions.findOneByRoomIdAndUserId(roomId, uid, { projection: { ls: 1 } });
+
+		const lastSeen = subscription?.ls;
+		if (!lastSeen) {
+			return 0;
+		}
+
+		return Messages.countVisibleByRoomIdBetweenTimestampsNotContainingTypes(roomId, lastSeen, new Date(), []);
```
KevLehman, line 299: This will count system messages as "unread", and likely giving a different count from what we see on the UI.

apps/meteor/app/apps/server/bridges/rooms.ts
```diff
@@ -245,6 +245,60 @@ export class AppRoomBridge extends RoomBridge {
...
+		const subscription = await Subscriptions.findOneByRoomIdAndUserId(roomId, userId, { projection: { ls: 1 } });
+
+		const lastSeen = subscription?.ls;
+		if (!lastSeen) {
+			return [];
+		}
+
+		const sort: Sort = options.sort?.createdAt ? { ts: options.sort.createdAt } : { ts: 1 };
+
+		const cursor = Messages.findVisibleByRoomIdBetweenTimestampsNotContainingTypes(roomId, lastSeen, new Date(), [], {
```
KevLehman, line 271: This would return system messages as well as "unread". Is that a valid use case? 👀

KevLehman (changes requested):

apps/meteor/app/apps/server/bridges/rooms.ts
Dnouv: Yes it is thank you!

apps/meteor/app/apps/server/bridges/rooms.ts
Dnouv: We want the system messages.

apps/meteor/app/apps/server/bridges/rooms.ts
Dnouv: Thank you for your suggestion. I understand that mapping directly on the cursor can be more concise. However, I prefer the current approach for the following reasons:

1. Clarity
2. Readability

Please elaborate on the benefits you see in mapping directly on the cursor. Is it something standard we started to follow? I haven't come across any documentation mentioning this, so if it's more of a personal preference, I'd prefer to keep it as it is.

debdutdeb: we do this everywhere last I remember. cursor mapping means it's truely an iterator-like. not doubling the memory allocation.

https://github.com/RocketChat/Rocket.Chat/blob/d0490d3d84193b3a9ceec899673757209c14de3b/apps/meteor/app/apps/server/bridges/rooms.ts#L133-L135

you're also one of the folks following the same pattern in the past

apps/meteor/app/apps/server/bridges/rooms.ts
debdutdeb: Right ..I believe we've made this mistake of wanting and validating everything everywhere at every level - causing mostly performance degradation and unnecessary computation. something we're now trying to revert.

the "no relation" error may not be direct, but should suggest to the developer to do their own checks before if needed. keeping our api lean.

apps/meteor/app/apps/server/bridges/rooms.ts
Dnouv: Oh, my bad! But to be honest, I am not sure we did this everywhere; IMHO, it doesn't matter much.

apps/meteor/app/apps/server/bridges/rooms.ts
Dnouv: Yes it doesn't!

apps/meteor/app/apps/server/bridges/rooms.ts
```diff
@@ -245,6 +245,57 @@ export class AppRoomBridge extends RoomBridge {
...
+		this.orch.debugLog(`The App ${appId} is getting the unread messages for the user: "${uid}" in the room: "${roomId}"`);
+
+		const messageConverter = this.orch.getConverters()?.get('messages');
+		if (!messageConverter) {
+			throw new Error('Message converter not found');
+		}
+
+		const subscription = await Subscriptions.findOneByRoomIdAndUserId(roomId, uid, { projection: { ls: 1 } });
+
+		if (!subscription) {
```
KevLehman, line 258: Prefer a more generic error and log the error message as you want. But throwing a 2 lines error feels weird

apps/meteor/app/apps/server/bridges/rooms.ts
```diff
@@ -245,6 +245,57 @@ export class AppRoomBridge extends RoomBridge {
...
+		return Promise.all(messages.map((msg) => messageConverter.convertMessageRaw(msg)));
+	}
+
+	protected async getUserUnreadMessageCount(roomId: string, uid: string, appId: string): Promise<number> {
+		this.orch.debugLog(`The App ${appId} is getting the unread messages count of the room: "${roomId}" for the user: "${uid}"`);
+
+		const subscription = await Subscriptions.findOneByRoomIdAndUserId(roomId, uid, { projection: { ls: 1 } });
+
+		if (!subscription) {
+			throw new Error(
```
KevLehman, line 286: Same

KevLehman (changes requested):

KevLehman (approved):

d-gubert (approved):

PR #32196, opened 2024-04-12
test: ignore user messages

jessicaschelly:

Creation of automated tests for the ignore/unignore user inside a channel.

- Ignore user
- Unignore only one message
- Unignore user

closes: https://rocketchat.atlassian.net/browse/CORE-261


apps/meteor/tests/e2e/channel-management.spec.ts
```diff
@@ -76,6 +76,63 @@ test.describe.serial('channel-management', () => {
...
+
+		await user1Page.close();
+	});
+
+
+	test('should unignore "user1" messages', async ({ browser }) => {
+		await poHomeChannel.sidenav.openChat(targetChannel);
+		await poHomeChannel.tabs.btnTabMembers.click();
+		await poHomeChannel.tabs.members.showAllUsers();
+		await poHomeChannel.tabs.members.unignoreUser('user1');
```
MarcosSpessatto, line 121: Maybe we should send a message before this just to ensure the user is still ignored. 🤔

apps/meteor/tests/e2e/page-objects/fragments/home-flextab-members.ts
```diff
@@ -26,6 +26,18 @@ export class HomeFlextabMembers {
...
+		await this.page.locator('role=button[name="More"]').click();
+		await this.page.locator('role=menuitem[name="Ignore"]').click();
+	}
+
+	async unignoreUser(username: string) {
+		await this.page.locator(`[data-qa="MemberItem-${username}"]`).click();
+		await this.page.locator('role=button[name="More"]').click();
+		await this.page.locator('role=menuitem[name="Unignore"]').click();
+	}
+
```
MarcosSpessatto, line 40: Since the function body is almost identical we could extract that "shared" to a new private function, something like:

```
private async ignoreUserAction(action: string, username: string): {
        await this.page.locator(`[data-qa="MemberItem-${username}"]`).click();
	await this.page.locator('role=button[name="More"]').click();
	await this.page.locator(`role=menuitem[name="${action}"]`).click();
}

async ignoreUser(username: string) {
	await this.ignoreUserAction('Ignore', username);
}

async unignoreUser(username: string) {
	await this.ignoreUserAction('Unignore', username);
}
```

MarcosSpessatto (changes requested):

apps/meteor/tests/e2e/channel-management.spec.ts
jessicaschelly: Thank you! I added a new validation to ensure the user is still ignored and I added one that checks to see if the previous message is also being unignored

apps/meteor/tests/e2e/page-objects/fragments/home-flextab-members.ts
jessicaschelly: Awesome, thank you

MarcosSpessatto (approved):

PR #32197, opened 2024-04-12
feat(E2EE): Async E2EE keys exchange

yash-rajpal:

Async key distribution process for E2EE rooms. Earlier users needed to be online to receive and share and verify the group keys which caused a bottleneck in the process. Made this process async so that users now no longer need to be online at the same time.

Keys will be generated even when the users are offline and they can receive it when they come online.
## Steps to test or reproduce

- Add a few users in an E2EE room
- reset the E2EE for few users
- Log back in with those users (just to create their public keys)
- When the main users comes online in the E2EE room, it will generate the keys for all the users waiting in queue.
- When users come online they receive keys and E2EE rooms messages are decrypted.

## Further comments

[E2EE-13](https://rocketchat.atlassian.net/browse/E2EE-13)

[E2EE-13]: https://rocketchat.atlassian.net/browse/E2EE-13?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/meteor/app/e2e/server/functions/provideUsersSuggestedGroupKeys.ts
```diff
@@ -0,0 +1,56 @@
...
+
+	if (!roomIds) {
+		return;
+	}
+
+	for await (const roomId of roomIds) {
+		if (!(await canAccessRoomIdAsync(roomId, userId))) {
+			throw new Meteor.Error('error-invalid-room', 'Invalid room');
+		}
+	}
```
MarcosSpessatto, line 27: Maybe change to a Promise.all will be faster?

apps/meteor/app/e2e/server/functions/provideUsersSuggestedGroupKeys.ts
```diff
@@ -0,0 +1,56 @@
...
+	const completedRoomIds = (
+		await Promise.all(
+			roomIds.map(async (roomId) => {
+				const users = usersSuggestedGroupKeys[roomId];
+
+				const subscription = await Subscriptions.findOneByRoomIdAndUserId(roomId, userId);
+
+				if (!subscription) {
+					return;
+				}
```
MarcosSpessatto, line 38: We don't need to check this, we already checked the `canAccessRoomIdAsync`. We are good.

apps/meteor/server/lib/resetUserE2EKey.ts
```diff
@@ -54,6 +54,14 @@ const sendResetNotification = async function (uid: string): Promise<void> {
 	}
 };
 
+const addUserToE2EKeysWaitingQueue = async (uid: string): Promise<void> => {
+	const subscribedRoomIds = (await Subscriptions.findByUserId(uid, { projection: { rid: 1 } }).toArray()).map((sub) => sub.rid);
```
MarcosSpessatto, line 58: I think we don't need to do this. Let's set the data directly to rooms that are encrypted.

apps/meteor/server/lib/resetUserE2EKey.ts
```diff
@@ -54,6 +54,14 @@ const sendResetNotification = async function (uid: string): Promise<void> {
...
+
+	await Rooms.removeUserFromE2EEQueueByRoomIds(subscribedRoomIds, uid);
```
MarcosSpessatto, line 60: Also, we don't need this.

apps/meteor/server/models/raw/Rooms.ts
```diff
@@ -1970,4 +1970,71 @@ export class RoomsRaw extends BaseRaw<IRoom> implements IRoomsModel {
...
+		const query: Filter<IRoom> = {
+			_id: {
+				$in: roomIds,
+			},
+			encrypted: true,
+		};
+
+		const update: UpdateFilter<IRoom> = {
+			$push: {
+				usersWaitingForE2EKeys: { $each: [uid], $slice: -50 },
```
MarcosSpessatto, line 1984: We need to add the timestamp as well, not only the userId.

apps/meteor/app/e2e/client/rocketchat.e2e.ts
```diff
@@ -507,6 +517,65 @@ class E2E extends Emitter {
 
 		return message;
 	}
+
+	fetchUsersWithWaitingKeys() {
+		return sdk.rest.get('/v1/e2e.fetchUsersWaitingForGroupKey');
```
MarcosSpessatto, line 522: I'm wondering if we could do this on the client side only. For the rooms I am a participant, the field telling if users are waiting for the queue should be available on the client side, isn't it? This way we can save this call and not add an extra endpoint.

apps/meteor/app/api/server/v1/e2e.ts
MarcosSpessatto, line 1: We need to check if the feature is enabled(`E2E_Enable`).

apps/meteor/app/lib/server/functions/addUserToRoom.ts
```diff
@@ -127,5 +127,9 @@ export const addUserToRoom = async function (
 		await Team.addMember(inviter || userToBeAdded, userToBeAdded._id, room.teamId);
 	}
 
+	if (room.encrypted) {
```
MarcosSpessatto, line 130: Shouldn't we check if the setting for E2E enabled is true?

MarcosSpessatto (changes requested):
I still have some general comments:
 - Let's improve the PR description.
 - We need to add a changeset.
 - We can have some API tests to test at least half of the workflow, we can call the provideUsersGroupKeys endpoint with arbitrary values for keys, and see if the keys were set to the subscriptions. Of course, the test will not be fully complete, and test it using an UI test could potentially create a slow test, since the process is completely async.

apps/meteor/app/e2e/client/rocketchat.e2e.ts
MarcosSpessatto: I think we can keep your implementation here since it could be a problem having this field available on the client-side. Let's make sure this is not available there.

apps/meteor/app/e2e/client/rocketchat.e2e.room.js
```diff
@@ -308,16 +308,24 @@ export class E2ERoom extends Emitter {
 		// Encrypt generated session key for every user in room and publish to subscription model.
 		try {
 			const { users } = await sdk.call('e2e.getUsersOfRoomWithoutKey', this.roomId);
-			users.forEach((user) => this.encryptForParticipant(user));
+			users.forEach(async (user) => {
```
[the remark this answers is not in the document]
rodrigok: @MarcosSpessatto is there a way to prevent this via lint?

apps/meteor/app/e2e/client/rocketchat.e2e.room.js
```diff
@@ -308,16 +308,24 @@ export class E2ERoom extends Emitter {
...
+				if (!user?.e2e?.public_key) {
+					return;
+				}
+
+				const encryptedGroupKey = await this.encryptGroupKeyForParticipant(user.e2e.public_key);
+				// Key has been encrypted. Publish to that user's subscription model for this room.
+				await sdk.call('e2e.updateGroupKey', this.roomId, user._id, encryptedGroupKey);
```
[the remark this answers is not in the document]
yash-rajpal: Already done that, using a rest API for that.

apps/meteor/app/e2e/client/rocketchat.e2e.room.js
```diff
@@ -326,8 +334,8 @@ export class E2ERoom extends Emitter {
 		// Encrypt session key for this user with his/her public key
 		try {
 			const encryptedUserKey = await encryptRSA(userKey, toArrayBuffer(this.sessionKeyExportedString));
-			// Key has been encrypted. Publish to that user's subscription model for this room.
-			await sdk.call('e2e.updateGroupKey', this.roomId, user._id, this.keyID + Base64.encode(new Uint8Array(encryptedUserKey)));
+			// return the encrypted key
```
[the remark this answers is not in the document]
yash-rajpal: Just put it there to make note that the returned value after all computations is the final encrypted group key, will remove the comment and use a variable to better understand the code.

apps/meteor/app/e2e/client/rocketchat.e2e.ts
```diff
@@ -507,6 +517,65 @@ class E2E extends Emitter {
...
+		if (this.timeout) {
+			return;
+		}
+
+		this.timeout = setInterval(async () => {
+			const { usersWaitingForE2EKeys, hasMore } = await this.fetchUsersWithWaitingKeys();
+
+			if (!hasMore && this.timeout) {
+				clearTimeout(this.timeout);
+			}
```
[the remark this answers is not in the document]
yash-rajpal: In that case they should receive the key immediately using the old flow, using the streams when both clients are online.

apps/meteor/app/e2e/client/rocketchat.e2e.room.js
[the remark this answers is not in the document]
yash-rajpal: I did it purposely, I saw no reason for waiting for this. 🤔

apps/meteor/app/e2e/server/functions/fetchUsersWaitingForGroupKey.ts
```diff
@@ -0,0 +1,48 @@
...
+					if (!userIds || userIds.length === 0) {
+						return;
+					}
+
+					const users = (await Users.findByIdsWithPublicE2EKey(userIds, { projection: { e2e: 1 } }).toArray()).map((user) => ({
+						_id: user._id,
+						public_key: user.e2e?.public_key as string,
+					}));
+
+					return [room._id, users];
```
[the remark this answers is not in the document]
yash-rajpal: Done this in the updated flow, just FYI - we need to query the users collection to get the users with their public keys generated, as we need user's public key to encrypt the group E2EE key.

apps/meteor/server/lib/resetUserE2EKey.ts
yash-rajpal: But we don't know the subscribed rooms of the user.

Using this I'm getting the list of all subscribed roomIds, then setting the data for rooms which are encrypted out of these subscribed rooms.

Is there some better way to do this 🤔

apps/meteor/app/e2e/client/rocketchat.e2e.ts
```diff
@@ -507,6 +517,65 @@ class E2E extends Emitter {
...
+
+			if (!usersWaitingForE2EKeys) {
+				return;
+			}
+
+			const userKeysWithRooms = await this.getSuggestedE2EEKeys(usersWaitingForE2EKeys);
+
+			try {
+				await this.provideUsersSuggestedGroupKeys({ usersSuggestedGroupKeys: userKeysWithRooms });
```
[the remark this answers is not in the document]
yash-rajpal: Already doing that, when we fetch the list of rooms, it also returns us if more rooms are remaining.

If it returns that no more rooms are remaining, it resets the timer.

packages/rest-typings/src/v1/e2e.ts
```diff
@@ -87,6 +87,35 @@ const E2eSetRoomKeyIdSchema = {
...
+type E2EProvideUsersGroupKeyProps = {
+	usersSuggestedGroupKeys: Record<IRoom['_id'], { _id: IUser['_id']; key: string }[]>;
+};
+
+const E2EProvideUsersGroupKeySchema = {
+	type: 'object',
+	properties: {
+		usersSuggestedGroupKeys: {
+			type: 'object',
+			additionalProperties: {
```
KevLehman, line 99: shouldn't this be `properties`?

apps/meteor/app/e2e/server/functions/fetchUsersWaitingForGroupKey.ts
```diff
@@ -0,0 +1,46 @@
...
+	const roomIds = (await Subscriptions.findByUserId(userId, { projection: { rid: 1 } }).toArray()).map((sub) => sub.rid);
+
+	const cursor = Rooms.findRoomsByRoomIdsWithE2EEQueue(roomIds, { projection: { usersWaitingForE2EKeys: 1 } });
+
+	const count = await cursor.count();
+
+	const hasMore = count > NUMBER_OF_ROOMS;
+	const encryptedRoomIds = (await cursor.toArray()).map((room) => room._id);
+
+	const rooms = await Rooms.fetchRandomUsersFromE2EEQueue(encryptedRoomIds, NUMBER_OF_ROOMS, QUEUE_SIZE);
```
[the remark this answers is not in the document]
yash-rajpal: There can be many rooms, we are just getting random 10 rooms, not sure if mongo returns rooms randomly 🤔

apps/meteor/app/e2e/server/functions/fetchUsersWaitingForGroupKey.ts
```diff
@@ -0,0 +1,46 @@
...
+const NUMBER_OF_ROOMS = 10;
+const QUEUE_SIZE = 50;
+
+export async function fetchUsersWaitingForGroupKey(userId: IUser['_id']): Promise<{
+	usersWaitingForE2EKeys: Record<IRoom['_id'], { _id: IUser['_id']; public_key: string }[]>;
+	hasMore: boolean;
+}> {
+	const roomIds = (await Subscriptions.findByUserId(userId, { projection: { rid: 1 } }).toArray()).map((sub) => sub.rid);
+
+	const cursor = Rooms.findRoomsByRoomIdsWithE2EEQueue(roomIds, { projection: { usersWaitingForE2EKeys: 1 } });
```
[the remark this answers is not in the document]
yash-rajpal: Got it

packages/rest-typings/src/v1/e2e.ts
yash-rajpal: we don't have fixed properties, properties are roomIds which can be any random strings, let me check this.

yash-rajpal: I always thought properties was for fixed props.

apps/meteor/server/models/raw/Rooms.ts
```diff
@@ -1970,4 +1970,80 @@ export class RoomsRaw extends BaseRaw<IRoom> implements IRoomsModel {
...
+		const query: Filter<IRoom> = {
+			_id: {
+				$in: roomIds,
+			},
+			'usersWaitingForE2EKeys.userId': { $ne: uid },
+			encrypted: true,
+		};
+
+		const update: UpdateFilter<IRoom> = {
+			$push: {
```
MarcosSpessatto, line 1984: We should store 50 users at maximum, after that we need to rotate the users, removing the oldest ones to add the new ones.

apps/meteor/app/e2e/server/functions/fetchUsersWaitingForGroupKey.ts
MarcosSpessatto, line 22: We can't rely on the users in the queue only, what if more than 50 users are waiting for the keys?

We must retrieve the users based on the subscriptions (the ones that don't contain E2EKey (`findByRidWithoutE2EKey`)), get all the user ids, and then, from those users, find all users WITH the public key in place, so finally from those users we can extract a `$sample` of users.

MarcosSpessatto (changes requested):

apps/meteor/app/e2e/server/functions/provideUsersSuggestedGroupKeys.ts
```diff
@@ -0,0 +1,47 @@
+import type { IRoom, IUser } from '@rocket.chat/core-typings';
+import { Rooms, Subscriptions } from '@rocket.chat/models';
+import { Meteor } from 'meteor/meteor';
+
+import { isTruthy } from '../../../../lib/isTruthy';
+import { canAccessRoomIdAsync } from '../../../authorization/server/functions/canAccessRoom';
+
+export const provideUsersSuggestedGroupKeys = async (
+	userId: IUser['_id'],
+	usersSuggestedGroupKeys: Record<IRoom['_id'], { _id: IUser['_id']; key: string }[]>,
```
rodrigok, line 10: Use typing from `E2EProvideUsersGroupKeyProps`

packages/rest-typings/src/v1/e2e.ts
rodrigok: It should be in properties for sure

apps/meteor/server/models/raw/Rooms.ts
```diff
@@ -1970,4 +1970,80 @@ export class RoomsRaw extends BaseRaw<IRoom> implements IRoomsModel {
...
+
+		return this.updateMany(query, update);
+	}
+
+	findRoomsByRoomIdsWithE2EEQueue(roomIds: IRoom['_id'][], options?: FindOptions<IRoom>): FindCursor<IRoom> {
+		const query: Filter<IRoom> = {
+			_id: {
+				$in: roomIds,
+			},
+			usersWaitingForE2EKeys: {
```
rodrigok, line 2020:
```suggestion
			'usersWaitingForE2EKeys.0': {
```
I think this will only return the ones with at least one record. As it is it can return the ones with an empty array.

apps/meteor/app/e2e/server/functions/fetchUsersWaitingForGroupKey.ts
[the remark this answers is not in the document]
rodrigok: Why not use the sample here to return the random ids already? It can be limited to `NUMBER_OF_ROOMS +1`, so if the result is `> NUMBER_OF_ROOMS` just remove the last item from the array and define hasMore as true.

apps/meteor/server/models/raw/Rooms.ts
```diff
@@ -1970,4 +1970,80 @@ export class RoomsRaw extends BaseRaw<IRoom> implements IRoomsModel {
...
+					_id: {
+						$in: roomIds,
+					},
+					usersWaitingForE2EKeys: {
+						$exists: true,
+					},
+					encrypted: true,
+				}
+			},
+			{ $sample: { size: queueSize }},
```
rodrigok, line 2042: No sure this works as intended, it won't limit the number of users

apps/meteor/app/e2e/server/functions/fetchUsersWaitingForGroupKey.ts
```diff
@@ -0,0 +1,46 @@
...
+
+	const rooms = await Rooms.fetchRandomUsersFromE2EEQueue(encryptedRoomIds, NUMBER_OF_ROOMS, QUEUE_SIZE);
+
+	const usersWaitingForE2EKeys = Object.fromEntries(
+		(
+			await Promise.all(
+				rooms.map(async (room) => {
+					const userIds = room?.usersWaitingForE2EKeys?.map((item) => item.userId)?.filter((uid) => uid !== userId);
+
+					if (!userIds || userIds.length === 0) {
```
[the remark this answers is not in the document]
rodrigok: @KevLehman have to remove the `!` from your suggestion

apps/meteor/server/lib/resetUserE2EKey.ts
```diff
@@ -67,6 +67,10 @@ export async function resetUserE2EEncriptionKey(uid: string, notifyUser: boolean
 	await Users.resetE2EKey(uid);
 	await Subscriptions.resetUserE2EKey(uid);
 
+	// Add user to room waiting queue
+	const subscribedRoomIds = (await Subscriptions.findByUserId(uid, { projection: { rid: 1 } }).toArray()).map((sub) => sub.rid);
```
rodrigok, line 71: Should filter by encrypted rooms only

apps/meteor/server/models/raw/Rooms.ts
```diff
@@ -1970,4 +1970,80 @@ export class RoomsRaw extends BaseRaw<IRoom> implements IRoomsModel {
 
 		return this.updateOne(query, update);
 	}
+
+	addUserIdToE2EEQueueByRoomIds(roomIds: IRoom['_id'][], uid: IUser['_id']): Promise<Document | UpdateResult> {
+		const query: Filter<IRoom> = {
+			_id: {
+				$in: roomIds,
+			},
+			'usersWaitingForE2EKeys.userId': { $ne: uid },
```
rodrigok, line 1979: A index for this field needs to be considered

rodrigok (changes requested):

apps/meteor/app/e2e/server/functions/fetchUsersWaitingForGroupKey.ts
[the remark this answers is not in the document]
KevLehman: oops

apps/meteor/app/api/server/v1/e2e.ts
```diff
@@ -5,9 +5,12 @@ import {
 	ise2eSetUserPublicAndPrivateKeysParamsPOST,
 	ise2eUpdateGroupKeyParamsPOST,
 } from '@rocket.chat/rest-typings';
+import { isE2EProvideUsersGroupKeyProps } from '@rocket.chat/rest-typings/dist/v1/e2e';
```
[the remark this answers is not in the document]
yash-rajpal: Oops, the auto import did that and I didn't notice. Will correct.

KevLehman (changes requested):

packages/rest-typings/src/v1/e2e.ts
rodrigok: I was wrong, it's right to use additionalProperties in this case

apps/meteor/app/e2e/client/rocketchat.e2e.ts
```diff
@@ -96,6 +104,33 @@ class E2E extends Emitter {
...
+
+				if (!e2eRoom) {
+					return;
+				}
+
+				if (await e2eRoom.importGroupKey(sub.E2ESuggestedKey)) {
+					await e2e.acceptSuggestedKey(sub.rid);
+					e2eRoom.keyReceived();
+				} else {
+					console.warn('Invalid E2ESuggestedKey, rejecting', sub.E2ESuggestedKey);
```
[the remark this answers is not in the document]
rodrigok: It's a client log, no issue, and good for debug

apps/meteor/app/e2e/server/functions/handleSuggestedGroupKey.ts
```diff
@@ -23,6 +23,11 @@ export async function handleSuggestedGroupKey(
 
 	if (handle === 'accept') {
 		await Subscriptions.setGroupE2EKey(sub._id, suggestedKey);
+		await Rooms.removeUsersFromE2EEQueueByRoomIds([sub.rid], [userId]);
```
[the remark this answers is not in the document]
rodrigok: I think it's gooed to be in sequence so it won't remove users from queue if the key set fail

apps/meteor/server/models/raw/Rooms.ts
rodrigok: @KevLehman addToSet does not allow the usage of slice to limit the size of the array

apps/meteor/app/e2e/server/functions/fetchUsersWaitingForGroupKey.ts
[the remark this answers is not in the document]
rodrigok: @KevLehman because many users can be doing this operation at the same time so it try to prevent them to generate keys for the same users and collaborate to generate to more users at the same time if it happens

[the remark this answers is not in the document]
KevLehman: Thanks for helping my curiosity.

apps/meteor/client/startup/e2e.ts
```diff
@@ -38,26 +37,14 @@ Meteor.startup(() => {
...
 		if (!e2e.isReady()) {
 			offClientMessageReceived?.();
-			unsubNotifyUser?.();
-			unsubNotifyUser = undefined;
 			observable?.stop();
 			offClientBeforeSendMessage?.();
 			return;
 		}
 
-		unsubNotifyUser = sdk.stream('notify-user', [`${Meteor.userId()}/e2ekeyRequest`], async (roomId, keyId): Promise<void> => {
```
yash-rajpal, line 52: @KevLehman Why are we removing the old way of key exchange?

yash-rajpal: IMO we need both ways of key exchange, asynchronous (new way) and synchronous (old way).

apps/meteor/app/e2e/server/methods/setUserPublicAndPrivateKeys.ts
```diff
@@ -23,5 +23,8 @@ Meteor.methods<ServerMethods>({
 			private_key: keyPair.private_key,
 			public_key: keyPair.public_key,
 		});
+
+		const subscribedRoomIds = await Rooms.getSubscribedRoomIdsWithoutE2EKeys(userId);
+		await Rooms.addUserIdToE2EEQueueByRoomIds(subscribedRoomIds, userId);
```
yash-rajpal, line 28: I like this that we are only adding users after we have ensured that they have their E2EE key pair.

We should also remove users from waiting queue, when they reset their keys again, as they no longer have their keys.

KevLehman: Good one!

ee/packages/ddp-client/src/types/streams.ts
```diff
@@ -160,7 +160,6 @@ export interface StreamerEvents {
 			key: `${string}/webdav`;
 			args: [{ type: 'changed'; account: Partial<IWebdavAccount> } | { type: 'removed'; account: { _id: IWebdavAccount['_id'] } }];
 		},
-		{ key: `${string}/e2ekeyRequest`; args: [string, string] },
```
MarcosSpessatto, line 163: We cannot simply remove this, this would be a breaking change, it's okay NOT to use this on the client side anymore but not to remove it from the server side, probably mobile is relying on this stream.

MarcosSpessatto (changes requested):

apps/meteor/app/api/server/v1/e2e.ts
```diff
@@ -230,3 +234,38 @@ API.v1.addRoute(
 		},
 	},
 );
+
+API.v1.addRoute(
+	'e2e.fetchUsersWaitingForGroupKey',
```
MarcosSpessatto, line 239: I think we need to check if the e2ee feature is enabled

apps/meteor/app/api/server/v1/e2e.ts
```diff
@@ -230,3 +234,38 @@ API.v1.addRoute(
...
+
+			return API.v1.success({
+				usersWaitingForE2EKeys,
+			});
+		},
+	},
+);
+
+API.v1.addRoute(
+	'e2e.provideUsersSuggestedGroupKeys',
```
MarcosSpessatto, line 259: I think we need to check if the e2ee feature is enabled

apps/meteor/app/e2e/client/rocketchat.e2e.ts
```diff
@@ -134,6 +143,31 @@ class E2E extends Emitter {
...
+
+					if (!e2eRoom) {
+						return;
+					}
+
+					if (await e2eRoom.importGroupKey(sub.E2ESuggestedKey)) {
+						await e2e.acceptSuggestedKey(sub.rid);
+						e2eRoom.keyReceived();
+					} else {
+						console.warn('Invalid E2ESuggestedKey, rejecting', sub.E2ESuggestedKey);
```
MarcosSpessatto, line 162: It would be better to use the private `log` method

apps/meteor/app/e2e/server/functions/provideUsersSuggestedGroupKeys.ts
```diff
@@ -0,0 +1,35 @@
...
+		}
+
+		const usersWithSuggestedKeys = [];
+		for await (const user of usersSuggestedGroupKeys[roomId]) {
+			const sub = await Subscriptions.findOneByRoomIdAndUserId(roomId, user._id, { projection: { _id: 1 } });
+			if (!sub) {
+				continue;
+			}
+
+			await Subscriptions.setGroupE2ESuggestedKey(sub._id, user.key);
```
MarcosSpessatto, line 29: Wouldn't it be better to set it by room id and user id (filtering it), this way we could save one query, if the filter doesn't match, nothing happens

apps/meteor/server/models/raw/Rooms.ts
```diff
@@ -1978,4 +1978,140 @@ export class RoomsRaw extends BaseRaw<IRoom> implements IRoomsModel {
...
+			'usersWaitingForE2EKeys.0': {
+				$exists: true,
+			},
+			'encrypted': true,
+		};
+
+		return this.find(query, options);
+	}
+
+	async fetchRandomUsersFromE2EEQueue(roomIds: IRoom['_id'][], numberOfRooms: number, queueSize: number): Promise<IRoom[]> {
```
MarcosSpessatto, line 2098: This can be removed, we are not using it anymore.

apps/meteor/server/models/raw/Rooms.ts
```diff
@@ -1978,4 +1978,140 @@ export class RoomsRaw extends BaseRaw<IRoom> implements IRoomsModel {
...
+		const update: UpdateFilter<IRoom> = {
+			$pull: {
+				usersWaitingForE2EKeys: { userId: uid },
+			},
+		};
+
+		return this.updateMany(query, update);
+	}
+
+	findRoomsByRoomIdsWithE2EEQueue(roomIds: IRoom['_id'][], options?: FindOptions<IRoom>): FindCursor<IRoom> {
```
MarcosSpessatto, line 2084: This can be removed, we are not using it anymore.

packages/model-typings/src/models/IRoomsModel.ts
```diff
@@ -278,4 +278,10 @@ export interface IRoomsModel extends IBaseModel<IRoom> {
 	removeDirectRoomContainingUsername(username: string): Promise<DeleteResult>;
 	countDiscussions(): Promise<number>;
 	setOTRForDMByRoomID(rid: string): Promise<UpdateResult>;
+	addUserIdToE2EEQueueByRoomIds(roomIds: IRoom['_id'][], uid: IUser['_id']): Promise<Document | UpdateResult>;
+	getSubscribedRoomIdsWithoutE2EKeys(uid: IUser['_id']): Promise<IRoom['_id'][]>;
+	removeUsersFromE2EEQueueByRoomId(roomId: IRoom['_id'], uids: IUser['_id'][]): Promise<Document | UpdateResult>;
+	findRoomsByRoomIdsWithE2EEQueue(roomIds: IRoom['_id'][], options?: FindOptions<IRoom>): FindCursor<IRoom>;
+	fetchRandomUsersFromE2EEQueue(roomIds: IRoom['_id'][], numberOfRooms: number, queueSize: number): Promise<IRoom[]>;
```
MarcosSpessatto, line 285: This can be removed, we are not using it anymore.

MarcosSpessatto (changes requested):

apps/meteor/app/api/server/v1/e2e.ts
```diff
@@ -230,3 +235,46 @@ API.v1.addRoute(
...
+
+API.v1.addRoute(
+	'e2e.fetchUsersWaitingForGroupKey',
+	{
+		authRequired: true,
+		validateParams: isE2EFetchUsersWaitingForGroupKeyProps,
+	},
+	{
+		async get() {
+			if (!settings.get('E2E_Enabled')) {
```
KevLehman, line 247:
```suggestion
			if (!settings.get('E2E_Enable')) {
```

apps/meteor/app/api/server/v1/e2e.ts
```diff
@@ -230,3 +235,46 @@ API.v1.addRoute(
...
+
+API.v1.addRoute(
+	'e2e.provideUsersSuggestedGroupKeys',
+	{
+		authRequired: true,
+		validateParams: isE2EProvideUsersGroupKeyProps,
+	},
+	{
+		async post() {
+			if (!settings.get('E2E_Enabled')) {
```
KevLehman, line 271:
```suggestion
			if (!settings.get('E2E_Enable')) {
```

apps/meteor/app/e2e/client/rocketchat.e2e.ts
```diff
@@ -296,6 +330,8 @@ class E2E extends Emitter {
 		this.instancesByRoomId = {};
 		this.privateKey = undefined;
 		this.started = false;
+		this.timeout && clearTimeout(this.timeout);
```
KevLehman, line 333:
```suggestion
		this.timeout && clearInterval(this.timeout);
```

MarcosSpessatto (changes requested):

apps/meteor/app/api/server/v1/e2e.ts
```diff
@@ -230,3 +238,46 @@ API.v1.addRoute(
...
+API.v1.addRoute(
+	'e2e.fetchUsersWaitingForGroupKey',
+	{
+		authRequired: true,
+		validateParams: isE2EFetchUsersWaitingForGroupKeyProps,
+	},
+	{
+		async get() {
+			if (!settings.get('E2E_Enable')) {
+				return API.v1.success({ usersWaitingForE2EKeys: {} });
```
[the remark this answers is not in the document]
KevLehman: My idea was to prevent errors on the setInterval 🤔 but not sure if that's a valid concern as the e2e worker won't run unless e2e is enabled.

Other possibility is the interval running at the same time with the feature being disabled, which would make the interval trying to do a run and failing with an unhandled promise rejection. Do you think that's a valid concern?

KevLehman (approved):

MarcosSpessatto (approved):

apps/meteor/app/e2e/client/rocketchat.e2e.ts
```diff
@@ -610,6 +731,99 @@ class E2E extends Emitter {
...
+			).filter(isTruthy),
+		);
+	}
+
+	async getSample(roomIds: string[], limit = 3): Promise<string[]> {
+		if (limit === 0) {
+			return [];
+		}
+
+		const randomRoomIds = _.sampleSize(roomIds, ROOM_KEY_EXCHANGE_SIZE);
```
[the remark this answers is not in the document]
yash-rajpal: The ADR suggests to get rooms randomly, eventually all rooms would be processed so not a big deal.

MarcosSpessatto (approved):

KevLehman (approved):

hugocostadev (approved):

apps/meteor/app/api/server/v1/e2e.ts
```diff
@@ -188,6 +193,9 @@ API.v1.addRoute(
 	{
 		authRequired: true,
 		validateParams: ise2eUpdateGroupKeyParamsPOST,
+		deprecation: {
+			version: '8.0.0',
+		},
```
cardoso, line 198: Deprecated here

apps/meteor/app/e2e/client/rocketchat.e2e.room.js
```diff
@@ -336,8 +352,8 @@ export class E2ERoom extends Emitter {
 		// Encrypt session key for this user with his/her public key
 		try {
 			const encryptedUserKey = await encryptRSA(userKey, toArrayBuffer(this.sessionKeyExportedString));
-			// Key has been encrypted. Publish to that user's subscription model for this room.
-			await sdk.call('e2e.updateGroupKey', this.roomId, user._id, this.keyID + Base64.encode(new Uint8Array(encryptedUserKey)));
+			const encryptedUserKeyToString = this.keyID + Base64.encode(new Uint8Array(encryptedUserKey));
+			return encryptedUserKeyToString;
```
cardoso, line 356: Usage removed here

Issue #32124, opened 2024-04-03
Firebase Cloud Messaging (FCM) Legacy API deprecation


machadolucas:
### Description:

Rocket.Chat is currently using the `node-gcm` project to integrate with Firebase Cloud Messaging (FCM) for sending push notifications. According to Google, the FCM Legacy API which is used by `node-gcm` project is now deprecated and will stop working on [June 20, 2024](https://firebase.google.com/support/faq#fcm-23-deprecation).
After that date, push notifications will stop working on Rocket.Chat if integration with FCM is not updated.

### Steps to reproduce:

1. Getting a [reminder](https://firebase.google.com/support/faq#fcm-23-deprecation) from Google about the deprecation and that my white-labeled RocketChat app is using Legacy HTTP protocol and Legacy Server Keys.
2. Verify that RocketChat server code is using outdated integration with FCM.

### Expected behavior:

Code needs to be updated with another FCM integration that supports the new FCM API. There is an issue in the `node-gcm` project recommending migration to another lib: https://github.com/ToothlessGear/node-gcm/issues/370

### Actual behavior:

`node-gcm` project is installed in [package.json#L135](https://github.com/RocketChat/Rocket.Chat/blob/develop/apps/meteor/package.json#L135) and used in [gcm.ts#L3](https://github.com/RocketChat/Rocket.Chat/blob/develop/apps/meteor/app/push/server/gcm.ts#L3)

### Server Setup Information:

- Version of Rocket.Chat Server: all versions

Gustrb:
Hey @machadolucas thanks for raising this issue!

I'm currently working on implementing the new version using the FCM API instead of the old API, whilst keeping everything compatible with the legacy way of sending notifications.

The implementation is mostly done, I will link it here when I open the Pull Request.

PR #32208, opened 2024-04-15
feat: node-gcm deprecation

Gustrb:

Currently, we are using a library called [node-gcm](https://www.npmjs.com/package/node-gcm) to send native notifications via firebase. Unfortunately Firebase is deprecating a series of libraries related to notifications (see https://firebase.google.com/support/faq?hl=pt-br#fcm-23-deprecation), and, because of this, the library we used to use is not going to be updated, mostly in favor of using the [oficial firebase SDK](https://www.npmjs.com/package/firebase).
We discussed about the use of the oficial package, but the oficial package bundles up every firebase feature inside a huge library, even though we just use a comically small subset of the features (sending notifications).
Therefore, we decided on not using the Firebase SDK and just using HTTP requests to send native notifications.

## Backportability

This feature is tricky on this regard, because this API is going to stop working, so we need to add a feature flag (using a setting for it), because everyone who is using supported versions of Rocket.Chat need native notifications to work, so we add a flag called `Push_UseLegacy`, if it is set to true, we will use the old api (using `node-gcm`), if it is false, it is going to use our HTTP wrapper on the notification service.

## Issue(s)

#32124

## Steps to test or reproduce

First, it will be needed a way to connect a mobile device into the local server, you can choose whatever way you want, I am using a tunneling application called [ngrok](https://ngrok.com/).
So, first I will start the application, on port 3000 (default) and then I will start ngrok there, you can do it like this:
```sh
$ ngrok http http://localhost:3000
```

After running the command, you will have a terminal interface, and you will see the URL the application is forwarding your local webserver, copy that value or save it somewhere, it will be needed later!
(it is a weird URL, something like: https://9d8c-191-243-134-83.ngrok-free.app)

So now, to test it you will need a test build of the application, please @ me on open, I can send you the build I am using 😎 

Download the build and install it on your device.
Open the app, and it will ask you for the URL of the workspace, in this field, please, put the forwarded URL by ngrok.

Ok, so now, most of the config on the mobile side is done, now there are a few config's needed to enable the new notification provider:
1 - Go to settings>push
2 - Disable Push_UseLegacy
3 - Get the credentials JSON and paste it into Push_google_api_credentials

No we should be ready to go!
Just create a new user on your phone (don't have to, but it will be easier), and now, go on your workspace open in your computer, and ping the new user in the #general channel, you should get a notification on your phobe 

## Further comments

I've used https://firebase.google.com/docs/reference/fcm/rest/v1/projects.messages?hl=pt-br to type out the request type 😛 

CONN-73


apps/meteor/app/push/server/push.ts
```diff
@@ -138,6 +189,33 @@ class PushClass {
 		}
 	}
 
+	private async getFcmAuthorizationToken(): Promise<{ token: string; projectNumber: string }> {
+		const credentialsString = settings.get('Push_google_api_credentials');
+		if (!credentialsString) {
+			throw new Error('Push_google_api_credentials is not set');
+		}
+
+		const credentials = JSON.parse(credentialsString as string) as FCMCredentials;
```
pierre-lehnen-rc, line 198:
```suggestion
		const credentialsString = settings.get<string>('Push_google_api_credentials');
		if (!credentialsString) {
			throw new Error('Push_google_api_credentials is not set');
		}

		const credentials = JSON.parse(credentialsString) as FCMCredentials;
```

apps/meteor/server/settings/push.ts
```diff
@@ -107,6 +113,14 @@ export const createPushSettings = () =>
 				enableQuery: [],
 				secret: true,
 			});
+
+			await this.add('Push_google_api_credentials', '', {
+				type: 'code',
+				multiline: true,
+				enableQuery: [],
```
pierre-lehnen-rc, line 120: Use `enableQuery` based on `Push_UseLegacy`'s value for both this new setting and the gcm one.

apps/meteor/app/push/server/push.ts
pierre-lehnen-rc, line 198: Since the string comes from a setting, we should validate that the resulting object actually matches the FCMCredentials type.

pierre-lehnen-rc (changes requested):

apps/meteor/app/push/server/fcm.ts
```diff
@@ -0,0 +1,146 @@
...
+	fcm_options?: {
+		analytics_label?: string;
+	};
+};
+
+function fetchWithRetry(url: string, options: RequestInit, retries = 0): Promise<Response> {
+	const MAX_RETRIES = 5;
+
+	try {
+		return fetch(url, options);
```
pierre-lehnen-rc, line 53: Is it okay to enable retry for every possible error? I don't know if there's any error that can happen and still succesfully send the push - in that case this could cause the notification to be sent multiple times.

apps/meteor/app/push/server/push.ts
Gustrb: should I use ajv or Match for that?

apps/meteor/app/push/server/fcm.ts
Gustrb: I think I will be a little smarter on that one, there are a few error codes that we shouldn't retry.
Technically there is no error code that would make a message be sent multiple times, BUT, there are a few validation errors that no matter how many requests we send over, they won't work, so we can save a few API calls there

apps/meteor/app/push/server/push.ts
Gustrb: did it using ajv

pierre-lehnen-rc (dismissed):

pierre-lehnen-rc (approved):

.changeset/silent-dodos-doubt.md
```diff
@@ -0,0 +1,5 @@
+---
+"@rocket.chat/meteor": patch
```
rodrigok, line 2:
```suggestion
"@rocket.chat/meteor": minor
```

apps/meteor/server/settings/push.ts
```diff
@@ -19,6 +19,12 @@ export const createPushSettings = () =>
 			alert: 'Push_Setting_Requires_Restart_Alert',
 		});
 
+		await this.add('Push_UseLegacy', false, {
+			type: 'boolean',
+			public: true,
```
rodrigok, line 24:
```suggestion
```
Only used on server side

apps/meteor/app/push/server/fcm.ts
```diff
@@ -0,0 +1,159 @@
...
+
+	// then we will create the message for each token
+	return userTokens.map((token) => ({ message: { ...message, token } }));
+}
+
+export const sendFCM = function ({ userTokens, notification, _replaceToken, _removeToken, options }: NativeNotificationParameters): void {
+	// We don't use these parameters, but we need to keep them to keep the function signature
+	// TODO: Remove them when we remove the old sendGCM function
+	_replaceToken;
+	_removeToken;
```
rodrigok, line 129: Why these references?

rodrigok (changes requested):

apps/meteor/app/push/server/fcm.ts
Gustrb: Just to keep the interface compatible with the legacy `sendFcm` API
It should be removed whenever we delete that code (not sure when though, hope it lands on 7.0)

.changeset/silent-dodos-doubt.md
sampaiodiego: we may want to have this as a `patch` since we'll probably release in older versions as well

rodrigok (approved):

apps/meteor/app/push/server/fcm.ts
```diff
@@ -0,0 +1,159 @@
...
+	if (notification.notId != null) {
+		data.notId = notification.notId.toString();
+	}
+	if (notification.gcm?.style != null) {
+		data.style = notification.gcm?.style;
+	}
+
+	if (notification.contentAvailable != null) {
+		data['content-available'] = notification.contentAvailable.toString();
+	}
```
debdutdeb, line 104: I don't have any personal feud against `!=` over `!==`, but for consistency I'd ask to use `!==` or just simply `if (var) { ..`. At first glance I even typed out the comment `why checking for null, not being set I'd expect them to be undefined instead`, until noticed the coercion. I guess that is something. If the value could never be `null` explicitly, `!= null` is confusing when we use strict equals everywhere AFAIK. 

Depending on how serialized, you might even just `data.notId = notification.notId?.toString()` for instance. Depends on if serialization checks for truthness or existence (and also how receiver treats em, should be safe).

apps/meteor/app/push/server/fcm.ts
```diff
@@ -0,0 +1,159 @@
...
+	// We don't use these parameters, but we need to keep them to keep the function signature
+	// TODO: Remove them when we remove the old sendGCM function
+	_replaceToken;
+	_removeToken;
+
+	const tokens = typeof userTokens === 'string' ? [userTokens] : userTokens;
+	if (!tokens.length) {
+		logger.debug('sendFCM no push tokens found');
+		return;
+	}
```
debdutdeb, line 135: should this be a normal log or debug? if no tokens, notification isn't sent, wouldn't glancing at logs immediately and noticing the issue help?

apps/meteor/app/push/server/push.ts
```diff
@@ -138,6 +234,34 @@ class PushClass {
 		}
 	}
 
+	private async getFcmAuthorizationToken(): Promise<string> {
+		const credentialsString = settings.get<string>('Push_google_api_credentials');
+		if (!credentialsString) {
+			throw new Error('Push_google_api_credentials is not set');
+		}
+
+		const credentials = JSON.parse(credentialsString);
```
debdutdeb, line 243: try catch

apps/meteor/app/push/server/fcm.ts
```diff
@@ -0,0 +1,159 @@
...
+	const headers = {
+		'Content-Type': 'application/json',
+		'Authorization': `Bearer ${options.gcm.apiKey}`,
+		'access_token_auth': true,
+	} as Record<string, any>;
+
+	const credentialsString = settings.get<string>('Push_google_api_credentials');
+	const credentials = JSON.parse(credentialsString) as FCMCredentials;
+
+	const url = `https://fcm.googleapis.com/v1/projects/${credentials.project_id}/messages:send`;
```
debdutdeb, line 149: it will pass ajv validation even if `project_id` is an empty string afaik. with retry being in place, would be better to validate first that it isn't an empty string before starting that story.

apps/meteor/app/push/server/push.ts
```diff
@@ -2,17 +2,79 @@ import type { IAppsTokens, RequiredField, Optional, IPushNotificationConfig } fr
...
+
+export const FCMCredentialsValidationSchema = {
+	type: 'object',
+	properties: {
+		type: {
+			type: 'string',
+		},
+		project_id: {
+			type: 'string',
+		},
```
debdutdeb, line 45: maybe a validator function for my comment above

apps/meteor/app/push/server/push.ts
```diff
@@ -124,7 +199,28 @@ class PushClass {
...
+					},
+				};
+
+				sendFCM({
+					userTokens: app.token.gcm,
+					notification,
+					_replaceToken: this.replaceToken,
+					_removeToken: this.removeToken,
+					options: sendGCMOptions as RequiredField<PushOptions, 'gcm'>,
+				});
```
debdutdeb, line 222: You are sending an object, could the project id not be sent from here at the same time without breaking the other signature (like inside `options`)?

I ask because the setting (Push_google_api_credentials) is loaded in both places, at one place being validated, another, not (not an issue exactly since flow has the second time covered), both times deserialized while doesn't have to just for the project_id.

apps/meteor/app/push/server/push.ts
```diff
@@ -138,6 +234,34 @@ class PushClass {
 		}
 	}
 
+	private async getFcmAuthorizationToken(): Promise<string> {
+		const credentialsString = settings.get<string>('Push_google_api_credentials');
+		if (!credentialsString) {
```
debdutdeb, line 239: nit
```suggestion
		if (!credentialsString.trim()) {
```

debdutdeb (changes requested):

debdutdeb:
with 6.8 around the corner, my main concern is the retry flow

apps/meteor/app/push/server/fcm.ts
Gustrb: I agree, I think I copied over this code from sendGcm
but that old pattern is probably a leftover from when it was written in js

apps/meteor/app/push/server/fcm.ts
```diff
@@ -0,0 +1,184 @@
...
+	const retryAfter = response.headers.get('retry-after');
+	const retryAfterSeconds = retryAfter ? parseInt(retryAfter, 10) : 60;
+
+	if (response.status === 429) {
+		await new Promise((resolve) => setTimeout(resolve, retryAfterSeconds * 1000));
+		return fetchWithRetry(url, options, retries + 1);
+	}
+
+	if (response.status >= 500 && response.status < 600) {
+		const backoff = Math.pow(2, retries) * 1000;
```
debdutdeb, line 88: minimum is 10 seconds
> Minimum Interval: Do not immediately retry a failed request with FCM. Wait at least 10 seconds before retrying a failed request.

debdutdeb:
```suggestion
		const backoff = Math.pow(2, retries) * 10000;
```

Gustrb: oh yea, my bad

debdutdeb (approved):

apps/meteor/server/settings/push.ts
```diff
@@ -104,13 +109,36 @@ export const createPushSettings = () =>
 			});
 			await this.add('Push_gcm_api_key', '', {
```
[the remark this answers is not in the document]
Gustrb: I mean I think we could just keep it to avoid confusion, since the change is going to be kind of big already
and when the API reaches end of life we just remove it + sends some emails so people update their settings

ggazzo:
I'm not ignoring the conversations, but since but in theory was already approved  and if was not the ci failing , it would be merged by saturday ;) so if there is anything to improve please use the release candidate do to that

apps/meteor/app/push/server/fcm.ts
Lao-Ax: M... May I ask you to explain why we don't need them? As I see, this PR does not contain a code to clear the mongo database from invalid tokens. Unlike sendGcm does.

Is it irrelevant now?

Gustrb: Hey @Lao-Ax thanks for your comment, I was mistaken on that one, there is no need to remove the tokens if it failed, since it is impossible in the new API
But, it looks like the logic for replacing stale tokens is missing, I will open a quick PR to implement that, thanks again :)

Gustrb: It looks like there is no way to tell if we should just refresh the token or invalidate it
but the spec says that if we get a 404 we should remove it from the database, good catch!

PR #32210, opened 2024-04-15
fix(Omnichannel): `audio.stop() is not a function`

MartinSchoeler:


gabriellsh (changes requested):
As we discussed in DM, please transform this startup function into a hook and use it in the CustomSoundsProvider. This way we can remove yet another startup.

gabriellsh (dismissed):

apps/meteor/client/hooks/useContinuousSoundNotification.ts
```diff
@@ -0,0 +1,53 @@
...
+const query = { t: 'l', ls: { $exists: false }, open: true };
+export const useContinuousSoundNotification = () => {
+	const userSubscriptions = useUserSubscriptions(query);
+
+	const playNewRoomSoundContinuously = useSetting('Livechat_continuous_sound_notification_new_livechat_room');
+
+	const newRoomNotification = useUserPreference<string>('newRoomNotification') || '';
+	const audioVolume = useUserPreference<number>('notificationsSoundVolume');
+
+	const continuousCustomSoundId = `${newRoomNotification}-continuous`;
```
MarcosSpessatto, line 16: What happens if the newRoomNotification has its default value (`''`)? Could this `-continuous` id be a problem?

MarcosSpessatto (changes requested):
I'm not sure if it's possible, but would it be possible to add UI tests? If not, what about unit tests to test the new hook?

apps/meteor/client/hooks/useContinuousSoundNotification.ts
```diff
@@ -0,0 +1,53 @@
...
+
+	const volume = audioVolume ? Number((audioVolume / 100).toPrecision(2)) : 1;
```
KevLehman, line 18: I know this is a great improvement but if a user sets its volume to 1 that means he/she wont listen anything (or super low), which defeats the purpose of the "continuous sound notifs" setting 🤔 

Can we check if that new behavior is compatible with the setting's expected behavior?

If that's good, then consider that if a user sets his volume as 0, this condition will play it at full sound cause `0 ?` will be false

KevLehman (changes requested):

apps/meteor/client/hooks/useContinuousSoundNotification.ts
MartinSchoeler: We should be good, as the message notification sounds on omnnichannel use the volume setting, as for the `0 ?` issue, I've fixed that on a subsequent commit

KevLehman (dismissed):

MarcosSpessatto (dismissed):

MarcosSpessatto:
Unit tests will be delivered in a subsequent PR

KevLehman (approved):

gabriellsh (approved):

KevLehman (approved):

MarcosSpessatto (approved):

PR #32212, opened 2024-04-15
fix!: "Forgot Password" e-mail is not translated correctly

matheusbsilva137:

 - Add migration to update translation key used by the `Forgot_Password_Email` setting.

## Issue(s)

Introduced by #31343 

## Steps to test or reproduce

Start a workspace on 6.5 (or an older version), then update to the latest RC version. Then, check the e-mail received when triggering the "forgot password" flow.
**_Current behavior:_** the e-mail contains the text `Lets_get_you_new_one` (not correctly translated key).
**_Expected behavior:_** the e-mail should contain a translated text (such as "Let's get you a new one!") instead of a key.

## Further comments

[CORE-247](https://rocketchat.atlassian.net/browse/CORE-247)

[CORE-247]: https://rocketchat.atlassian.net/browse/CORE-247?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/meteor/server/startup/migrations/v309.ts
+++ b/apps/meteor/server/startup/migrations/v309.ts
@@ -1,4 +1,3 @@
-import type { ISetting } from '@rocket.chat/core-typings';
 import { Settings } from '@rocket.chat/models';
 
 import { addMigration } from '../../lib/migrations';
--- a/apps/meteor/server/startup/migrations/v309.ts
+++ b/apps/meteor/server/startup/migrations/v309.ts
@@ -1,3 +1,4 @@
+import type { ISetting } from '@rocket.chat/core-typings';
 import { Settings } from '@rocket.chat/models';
 
 import { addMigration } from '../../lib/migrations';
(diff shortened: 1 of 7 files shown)


gabriellsh (dismissed):

apps/meteor/server/startup/migrations/v309.ts
```diff
@@ -0,0 +1,24 @@
+import { Settings } from '@rocket.chat/models';
+
+import { settings } from '../../../app/settings/server';
```
MarcosSpessatto, line 3: Let's not use the cached setting here, we need to use the model instead

MarcosSpessatto (changes requested):

apps/meteor/server/startup/migrations/v309.ts
matheusbsilva137: Ooh, you're right :man_facepalming: fixed!

MarcosSpessatto (approved):

PR #32213, opened 2024-04-15
test: Unit tests for `AutoCloseOnHold` scheduler

KevLehman:
https://rocketchat.atlassian.net/browse/CORE-262

## Further comments

Unit tests for `AutoCloseOnHold` scheduler.

This installs `chai-datetime` package to assert the `when` parameters sent to the scheduler are close to the time we want it to be.


apps/meteor/ee/tests/unit/apps/livechat-enterprise/lib/AutoCloseOnHold.tests.ts
```diff
@@ -0,0 +1,246 @@
...
+			mockAgendaCancel.resetHistory();
+		});
+
+		it('should fail if scheduler has not been init', async () => {
+			const scheduler = new AutoCloseOnHoldSchedulerClass();
+
+			try {
+				await scheduler.unscheduleRoom('roomId');
+			} catch (e) {
+				expect(e).to.not.to.be.undefined;
```
tiagoevanp, line 148: `[].not.be.undefined`

image

KevLehman:
```suggestion
				expect(e).not.to.be.undefined;
```

apps/meteor/ee/tests/unit/apps/livechat-enterprise/lib/AutoCloseOnHold.tests.ts
```diff
@@ -0,0 +1,246 @@
...
+			mockAgendaCancel.resetHistory();
+		});
+
+		it('should fail if scheduler has not been init', async () => {
+			const scheduler = new AutoCloseOnHoldSchedulerClass();
+
+			try {
+				await scheduler.unscheduleRoom('roomId');
+			} catch (e) {
+				expect(e).not.to.be.undefined;
```
MarcosSpessatto, line 148: Wouldn't it be better if we slightly change the method checking if the scheduler was initialized and then throwing an error if it was not, instead of receiving some error that looks like some programmatic error?

I mean, it's clearer for the caller of this method to receive a validation error. WDYT?

apps/meteor/ee/tests/unit/apps/livechat-enterprise/lib/AutoCloseOnHold.tests.ts
```diff
@@ -0,0 +1,246 @@
...
+	describe('scheduleRoom', () => {
+		beforeEach(() => {
+			mockAgendaCancel.resetHistory();
+			mockAgendaDefine.resetHistory();
+		});
+
+		it('should schedule a room', async () => {
+			const scheduler = new AutoCloseOnHoldSchedulerClass();
+
+			await scheduler.init();
```
MarcosSpessatto, line 123: Can we add a test for when the scheduler has not initialized yet? And maybe add a check for that following the comment below.

MarcosSpessatto (changes requested):

MarcosSpessatto (approved):

aleksandernsilva (approved):

PR #32224, opened 2024-04-16
feat(messages): Allow Custom Fields in Messages

rodrigok:

Add two new settings to control the custom fields in the messages' object:
- `Message_CustomFields_Enabled`: Boolean, default false
- `Message_CustomFields`: String/Code/JSON, an [AJV](https://ajv.js.org/json-schema.html) schema validator

It allows a new property named `customFields` to be sent via message APIs. When enabled it will be validated based on the AJV schema defined.

The backend will force the schema type to "object" and prevent extra properties by forcing "additionalProperties" as false.

## Issue(s)
[CORE-274](https://rocketchat.atlassian.net/browse/CORE-274)

## Steps to test or reproduce

cURL for testing
```
curl --location 'http://localhost:3000/api/v1/chat.sendMessage' \
--header 'X-Auth-Token: ***' \
--header 'X-User-Id: ***' \
--data '{"message": {"rid":"GENERAL","msg":"Sample","customFields":{"priority": "high"}}}'
```

## Further comments

TODO
- [x] Tests
- [x] Changeset
- [ ] Docs
- [x] Task

[CORE-274]: https://rocketchat.atlassian.net/browse/CORE-274?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/meteor/server/settings/message.ts
```diff
@@ -385,4 +385,35 @@ export const createMessageSettings = () =>
...
+{
+	"properties": {
+		"priority": {
+			"type": "string",
+			"nullable": false,
+			"enum": ["low", "medium", "high"]
+		}
+	},
+	"required": ["priority"]
+}
```
MarcosSpessatto, line 404: Do we really need to have this as a default value?

apps/meteor/app/lib/server/lib/validateCustomMessageFields.ts
```diff
@@ -0,0 +1,37 @@
...
+	return ajv.compile({
+		...schema,
+		type: 'object',
+		additionalProperties: false,
+	});
+});
+
+export const validateCustomMessageFields = (
+	customFields: Record<string, any>,
+	messageCustomFieldsEnabled: boolean,
```
MarcosSpessatto, line 22: I think it would be better to provide an object for params to avoid this boolean param, it's clearer to understand. WDYT?

apps/meteor/app/lib/server/lib/validateCustomMessageFields.ts
```diff
@@ -0,0 +1,37 @@
+import Ajv from 'ajv';
+import mem from 'mem';
+
+const ajv = new Ajv();
+
+const customFieldsValidate = mem((customFieldsSetting: string) => {
+	const schema = JSON.parse(customFieldsSetting);
```
MarcosSpessatto, line 7: What happens if this throws an error?

apps/meteor/tests/unit/app/lib/server/lib/validateCustomMessageFields.tests.ts
```diff
@@ -0,0 +1,61 @@
...
+
+		it('should not allow to pass anything different from an object', () => {
+			const customFields = [1, 2];
+			const config = JSON.stringify({
+				type: 'array',
+				items: [{ type: 'integer' }, { type: 'integer' }],
+			});
+			expect(() => validateCustomMessageFields(customFields, true, config)).to.throw('Invalid custom fields config');
+		});
+
```
MarcosSpessatto, line 46: Can we add a test providing some invalid value to the config? Like a number, empty string, etc?

apps/meteor/server/settings/message.ts
```diff
@@ -385,4 +385,35 @@ export const createMessageSettings = () =>
...
 			public: true,
 		});
+
+		await this.add('Message_CustomFields_Enabled', false, {
+			type: 'boolean',
+		});
+		await this.add(
+			'Message_CustomFields',
+			`
+{
```
MarcosSpessatto, line 395: I'm not sure if it's possible, but can we validate this to ensure it's an object?

MarcosSpessatto (changes requested):

MarcosSpessatto (changes requested):
Let's add this new optional parameter here as well: https://github.com/RocketChat/Rocket.Chat/blob/develop/packages/rest-typings/src/v1/chat.ts#L15

Also, should we add that feature on message editing too?

apps/meteor/app/lib/server/lib/validateCustomMessageFields.ts
rodrigok: It throws the error that you can check in the unit test named `should not allow to pass invalid custom fields config`

apps/meteor/server/settings/message.ts
rodrigok: It's not possible. It validates if it's a valid JSON only

apps/meteor/tests/unit/app/lib/server/lib/validateCustomMessageFields.tests.ts
rodrigok: Done

apps/meteor/app/lib/server/lib/validateCustomMessageFields.ts
rodrigok: Done

apps/meteor/app/lib/server/functions/sendMessage.ts
```diff
@@ -172,6 +173,14 @@ export const validateMessage = async (message: any, room: any, user: any) => {
 	if (Array.isArray(message.attachments) && message.attachments.length) {
 		validateBodyAttachments(message.attachments);
 	}
+
+	if (message.customFields) {
+		validateCustomMessageFields(
+			message.customFields,
+			settings.get('Message_CustomFields_Enabled'),
```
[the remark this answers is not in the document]
rodrigok: Done

.changeset/four-eyes-sniff.md
```diff
@@ -0,0 +1,7 @@
+---
+"@rocket.chat/meteor": patch
+"@rocket.chat/core-typings": patch
+"@rocket.chat/i18n": patch
```
ggazzo, line 4:
```suggestion
"@rocket.chat/meteor": minor
"@rocket.chat/core-typings": minor
"@rocket.chat/i18n": minor
```

.changeset/four-eyes-sniff.md
```diff
@@ -0,0 +1,7 @@
...
+---
+
+feat(messages): Allow Custom Fields in Messages. API-only feature. It can be enabled and configured in Workspace Settings.
```
ggazzo, line 7:
```suggestion
Allow Custom Fields in Messages. API-only feature. It can be enabled and configured in Workspace Settings.
```

apps/meteor/app/api/server/v1/chat.ts
```diff
@@ -312,6 +312,7 @@ API.v1.addRoute(
 					roomId: String,
 					msgId: String,
 					text: String, // Using text to be consistant with chat.postMessage
+					customFields: Match.Maybe(Object),
```
MarcosSpessatto, line 315: Let's not forget to add this to the rest-typings as well.

apps/meteor/tests/end-to-end/api/05-chat.js
```diff
@@ -1101,6 +1101,237 @@ describe('[Chat]', function () {
 				})
 				.end(done);
 		});
+
+		describe('customFields', () => {
```
MarcosSpessatto, line 1105: Can we add a test for the `chat.update` endpoint too?

MarcosSpessatto (changes requested):

apps/meteor/app/api/server/v1/chat.ts
MarcosSpessatto: @rodrigok for the edit and for the `chat.postMessage` too.

apps/meteor/app/lib/server/lib/validateCustomMessageFields.ts
```diff
@@ -0,0 +1,41 @@
+import Ajv from 'ajv';
+import mem from 'mem';
+
+const ajv = new Ajv();
+
+const customFieldsValidate = mem((customFieldsSetting: string) => {
```
KevLehman, line 6: Just checking if its good to have this `mem` cached forever (default value of maxAge is infinity)

KevLehman: This would also leave the memoized return value there forever as mem won't clean it 👀

rodrigok: Why not? Since it's based on the key (schema)

apps/meteor/tests/end-to-end/api/05-chat.js
rodrigok: They are there already

apps/meteor/app/api/server/v1/chat.ts
MarcosSpessatto, line 315: The customFields property is still missing in the rest-typings definitions.

MarcosSpessatto (changes requested):

MarcosSpessatto (dismissed):

apps/meteor/app/lib/server/lib/validateCustomMessageFields.ts
KevLehman: I mean, the value will be valid depending on the cache and params.
 
What I was referring is that whn u don't pass a maxAge to mem, it doesn't cleanup the cache. So every time the setting value is changed and the new value is set on the cache, the old value will be just there. 

```
if (typeof maxAge === 'number') {
		// TODO: Drop after https://github.com/SamVerschueren/map-age-cleaner/issues/5
		// @ts-expect-error
		mapAgeCleaner(cache);
	}
```

rodrigok: Ok, got it, you are worried about memory consumption. I didn't think about that. Just sent a fix, thanks @KevLehman

KevLehman (approved):

ggazzo (approved):

PR #32225, opened 2024-04-16
refactor: `TeamsInfo` in favor of `RoomInfo`

dougfabris:


gabriellsh (changes requested):
I was doing your review but, maybe let's take this moment to do a more complete refactor, making the RoomInfo component purely visual, then we can introduce one component per room type. This way we avoid having mixed logic between teams and different types of channels.

PR #32230, opened 2024-04-16
fix: OTR messages not working when running microservices

KevLehman:
https://rocketchat.atlassian.net/browse/CORE-268

## Steps to test or reproduce

To test:
- Run `yarn turbo run ms` (pray for your computer to not explode)
- Wait for all services to be online
- Go to `localhost:4000` (notice we're not using `:3000` cause we want to access rocketchat through DDP Streamer service)
- Login with 2 users, attempt to start an OTR session. This part should work.
- After establishing, try to send a message from any of the users. Message will stay grayed out

## Further comments

Fixed a problem that caused OTR Session messages' to not being transmitted from one peer to another when running Rocket.Chat as microservices. This was caused by a legacy streamer that tried to use the websocket directly, which works on monolith but doesn't on microservices, cause these events are routed through DDP Streamer service.


ee/apps/ddp-streamer/src/Client.ts
```diff
@@ -98,7 +98,7 @@ export class Client extends EventEmitter {
 			this.ws.close(WS_ERRORS.CLOSE_PROTOCOL_ERROR, WS_ERRORS_MESSAGES.CLOSE_PROTOCOL_ERROR);
 		});
 
-		this.setMaxListeners(50);
+		this.setMaxListeners(200);
```
KevLehman, line 101: This fixes another problem which I didn't investigate why it happens. Node complains we put 51 listeners to this client. IDK, should i keep this change here?

KevLehman: (I know it's not a "fix" itself but yeah, you get the point)

apps/meteor/app/otr/server/methods/updateOTRAck.ts
```diff
@@ -17,6 +16,6 @@ Meteor.methods<ServerMethods>({
 			throw new Meteor.Error('error-invalid-user', 'Invalid user', { method: 'updateOTRAck' });
 		}
 		const acknowlegedMessage: IOTRMessage = { ...message, otrAck: ack };
-		notifications.streamRoomMessage.emit(message.rid, acknowlegedMessage);
+		void api.broadcast('otrAckUpdate', { roomId: message.rid, acknowlegedMessage });
```
KevLehman, line 19:
```suggestion
		void api.broadcast('otrAckUpdate', { roomId: message.rid, acknowledgeMessage });
```

apps/meteor/app/otr/server/methods/updateOTRAck.ts
```diff
@@ -17,6 +16,6 @@ Meteor.methods<ServerMethods>({
 			throw new Meteor.Error('error-invalid-user', 'Invalid user', { method: 'updateOTRAck' });
 		}
 		const acknowlegedMessage: IOTRMessage = { ...message, otrAck: ack };
```
KevLehman, line 18:
```suggestion
		const acknowledgeMessage: IOTRMessage = { ...message, otrAck: ack };
```

yash-rajpal (dismissed):
LGTM!

I know I'm not code-owner but since I worked on OTR, everything looks fine to me.

ee/apps/ddp-streamer/src/Client.ts
ggazzo: what needs to be invested is: is it correct to have more than 50 streams being listened ?

KevLehman: True. There's a task for that investigation. This change is more of a "shut up" but what I don't know is, should we shut up and then investigate? That would make customer "happy" for a while.

IDK. I'll remove this change from here for now.

PR #32237, opened 2024-04-17
fix: settings-related statistics not being updated according to license changes

aleksandernsilva:
Changes were made to the statistics to ensure that settings-related information is current.

Previously, some parts of the code were accessing setting information directly from the collection, bypassing license considerations.
To address this, the code was changed to utilize the `settings.get` util function. 

The util (`settings`) already has the logic built-in to deliver the correct setting value based on the active license, thus ensuring that the data is always up-to-date.

## Issue(s)
[CONN-134](https://rocketchat.atlassian.net/browse/CONN-134)

## Steps to test or reproduce
- Change the value of a premium only setting (Ex.: enable `Message_Read_Receipt_Enabled`)
- Call `/v1/statistics?refresh=true`
- Notice that the value for `readReceiptEnabled` is `true`
- Remove the license from the workspace (premium only settings should return to `invalidValue`)
- Call `/v1/statistics?refresh=true`
- Notice that the value from `readReceiptEnabled` which should be `false` is still `true`

## Further comments

[CONN-134]: https://rocketchat.atlassian.net/browse/CONN-134?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.changeset/popular-fishes-lay.md
```diff
@@ -0,0 +1,12 @@
+---
+"@rocket.chat/meteor": patch
+---
+
+**fix: settings-related statistics not being updated according to license changes**
```
sampaiodiego, line 5: pls focus on the changes (and how to present them) to the end user: https://developer.rocket.chat/contribute-to-rocket.chat/modes-of-contribution/participate-in-rocket.chat-development/development-workflow#id-4.-adding-changeset-to-your-pull-request
```suggestion
```

sampaiodiego (changes requested):

sampaiodiego (approved):

PR #32269, opened 2024-04-19
fix: Bypass trash when removing OTR system messages and read receipts

KevLehman:
https://rocketchat.atlassian.net/browse/CORE-297

## Further comments

System messages associated with an OTR conversation should be deleted and not stored on trash. By using the underlying mongo connection directly, we bypass our internal logic for trashing deleted messages


MarcosSpessatto (changes requested):
Can we add a changeset?

MarcosSpessatto (approved):

sampaiodiego (approved):

PR #32285, opened 2024-04-22
fix!: LDAP sync triggers multiple cron jobs in case an invalid sync interval is provided

matheusbsilva137:

 - Validate LDAP sync intervals before setting up their corresponding cron jobs;
 - Use LDAP sync interval setting's `packageValue`s as the fallback value in case the provided value is invalid;
 - Add a migration to update LDAP sync interval setting's `packageValue`s (from cron description to actual cron expressions, which is the type of value expected by RC since 6.3).

## Steps to test or reproduce

Go to **Settings > LDAP > Premium** and set an invalid Avatar Background Sync Interval or Auto Logout Interval.
**Current behavior:** cron jobs are triggered successively (there is no interval between them).
**Expected behavior:** cron jobs intervals should follow each setting's default value instead.

## Further comments

[CORE-273](https://rocketchat.atlassian.net/browse/CORE-273)

[CORE-273]: https://rocketchat.atlassian.net/browse/CORE-273?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.changeset/sixty-vans-grab.md
+++ b/.changeset/sixty-vans-grab.md
@@ -1,0 +1,6 @@
+---
+"@rocket.chat/meteor": major
+"@rocket.chat/i18n": patch
+---
+
+Fixed issue with LDAP sync triggering multiple cron jobs in case an invalid sync interval is provided
--- a/.changeset/sixty-vans-grab.md
+++ b/.changeset/sixty-vans-grab.md
@@ -1,6 +1,5 @@
 ---
 "@rocket.chat/meteor": major
-"@rocket.chat/i18n": patch
 ---
 
 Fixed issue with LDAP sync triggering multiple cron jobs in case an invalid sync interval is provided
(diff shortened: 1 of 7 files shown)


packages/i18n/package.json
MarcosSpessatto, line 1: Let's discard this unrelated change

apps/meteor/server/startup/migrations/v305.ts
```diff
@@ -0,0 +1,25 @@
+import { Settings } from '@rocket.chat/models';
+import { isValidCron } from 'cron-validator';
+
+import { settings } from '../../../app/settings/server';
```
MarcosSpessatto, line 4: We should not use the cached setting in migrations, let's use the model instead.

MarcosSpessatto (changes requested):

packages/i18n/package.json
matheusbsilva137: Done! This is being changed on install, though
I'll check if this change is already on develop

MarcosSpessatto (dismissed):

PR #32286, opened 2024-04-22
chore: Add alpine indication to test jobs that use it

d-gubert:

Add an Alpine indicator to test jobs that use it


sampaiodiego (changes requested):
image

I guess we can get rid of the `false` 😬

d-gubert:
> image
> I guess we can get rid of the `false` 😬

Definitely, this should be a draft PR anyway :joy:

sampaiodiego (approved):

PR #32289, opened 2024-04-22
fix: Server sending 2 notifications when `@all/@here` were used by a user without permissions

KevLehman:
https://rocketchat.atlassian.net/browse/SUP-549

## Further comments

A duplicate call to `api.broadcast` was causing that an error on `BeforeSavePreventMention` was being propagated twice to clients.

The global handler on `saveMessage` was updated to support using the `details` property coming from meteor errors allowing the original hook to just throw and allow the caller to handle the error accordingly.


MarcosSpessatto (changes requested):
Can we add a brief description to the PR?

MarcosSpessatto (approved):

dougfabris (approved):
Approving it on behalf of frontend team

PR #32311, opened 2024-04-24
fix: pdf

KevLehman:

### DIFF
--- a/ee/packages/pdf-worker/src/templates/ChatTranscript/components/MessageList.tsx
+++ b/ee/packages/pdf-worker/src/templates/ChatTranscript/components/MessageList.tsx
@@ -10,7 +10,7 @@ import { Quotes } from './Quotes';
 
 const styles = StyleSheet.create({
 	wrapper: {
-		marginBottom: 16,
+		paddingBottom: 16,
 		paddingHorizontal: 32,
 	},
 	message: {
@@ -20,7 +20,7 @@ const styles = StyleSheet.create({
 });
 
 export const MessageList = ({ messages, invalidFileMessage }: { messages: ChatTranscriptData['messages']; invalidFileMessage: string }) => (
-	<View>
+	<>
 		{messages.map((message, index) => (
 			<View style={styles.wrapper} key={index} wrap={false}>
 				{message.divider && <Divider divider={message.divider} />}
@@ -30,5 +30,5 @@ export const MessageList = ({ messages, invalidFileMessage }: { messages: ChatTr
 				{message.files && <Files files={message.files} invalidMessage={invalidFileMessage} />}
 			</View>
 		))}
-	</View>
+	</>
 );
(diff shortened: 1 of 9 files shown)


ee/packages/pdf-worker/src/templates/ChatTranscript/components/MessageList.tsx
```diff
@@ -19,16 +20,21 @@ const styles = StyleSheet.create({
 	},
 });
 
+const messageLongerThanPage = (message: string) => message.length > 1200;
```
MarcosSpessatto, line 23: How can we know that 1200 characters are longer than a page? 🤔

packages/i18n/package.json
```diff
@@ -25,7 +25,6 @@
 		"/dist"
 	],
 	"type": "module",
-
```
MarcosSpessatto, line 28: Can we dismiss this change?

MarcosSpessatto (changes requested):

packages/i18n/package.json
KevLehman: Oh, this is going automattic with yarn install 🤔 think where it originated

ee/packages/pdf-worker/src/templates/ChatTranscript/components/MessageList.tsx
KevLehman: Oh we actually don't 😬 

This was based on tests I performed where this was the "ideal" char limit based on the font + size + space available.

The max amount of chars a page with our settings allowed was around 1600. I'm here reducing it a little bit so it can "ideally" fit an image in the same page alongside big text.

Since all PDF sizing is standard for our case, as we dont allow customizations, this should be fine.

ee/packages/pdf-worker/src/worker.spec.ts
```diff
@@ -0,0 +1,73 @@
...
+		const newWorker = new PdfWorker('chat-transcript');
+
+		expect(newWorker).toBeInstanceOf(PdfWorker);
+		expect(newWorker.mode).toBe('chat-transcript');
+	});
+	it('should generate a pdf transcript for a big bunch of messages', async () => {
+		const stream = await pdfWorker.renderToStream({ data: bigConversationData });
+		const buffer = await streamToBuffer(stream);
+
+		expect(buffer).toBeTruthy();
```
MarcosSpessatto, line 41: In which case this would be falsy? (Same for the others) 🤔

MarcosSpessatto (changes requested):

ee/packages/pdf-worker/src/worker.spec.ts
KevLehman: actually this just checks that buffer "exists". If the generation fails, the process would just be stuck 😬 with no errors thrown so if we have a buffer here, is because we're good

MarcosSpessatto: But the `streamToBuffer` never returns a falsy value, right?

MarcosSpessatto (changes requested):

MarcosSpessatto (approved):

PR #32312, opened 2024-04-25
regression: Don't encrypt attachment description for disabled E2EE rooms

yash-rajpal:

Implemented proper checks before encrypting message on file upload flow.
## Steps to test or reproduce

1. Create an E2EE room, then disable E2EE in the room.
2. Send some attachment with description
3. The message is encrypted (should not be encrypted)

## Further comments

[E2EE-26](https://rocketchat.atlassian.net/browse/E2EE-26)

[E2EE-26]: https://rocketchat.atlassian.net/browse/E2EE-26?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/meteor/client/lib/chats/flows/uploadFiles.ts
+++ b/apps/meteor/client/lib/chats/flows/uploadFiles.ts
@@ -3,10 +3,12 @@ import { isRoomFederated } from '@rocket.chat/core-typings';
 import { Random } from '@rocket.chat/random';
 
 import { e2e } from '../../../../app/e2e/client';
+import { ChatRoom } from '../../../../app/models/client';
 import { fileUploadIsValidContentType } from '../../../../app/utils/client';
 import FileUploadModal from '../../../views/room/modals/FileUploadModal';
 import { imperativeModal } from '../../imperativeModal';
 import { prependReplies } from '../../utils/prependReplies';
+import { waitUntilFind } from '../../utils/waitUntilFind';
 import type { ChatAPI } from '../ChatAPI';
 
 export const uploadFiles = async (chat: ChatAPI, files: readonly File[], resetFileInput?: () => void): Promise<void> => {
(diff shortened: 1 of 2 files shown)


MarcosSpessatto (changes requested):
Can we create some test cases to cover this?

apps/meteor/client/lib/chats/flows/uploadFiles.ts
```diff
@@ -61,7 +63,11 @@ export const uploadFiles = async (chat: ChatAPI, files: readonly File[], resetFi
 						return;
 					}
 
-					const shouldConvertSentMessages = e2eRoom.shouldConvertSentMessages({ msg });
+					const subscription = await waitUntilFind(() => ChatRoom.findOne({ _id: room._id }));
```
gabriellsh, line 66: The `room` object declared on line 19 should be the same as this. Is it possible to use it instead?

gabriellsh (changes requested):

apps/meteor/client/lib/chats/flows/uploadFiles.ts
```diff
@@ -61,7 +61,9 @@ export const uploadFiles = async (chat: ChatAPI, files: readonly File[], resetFi
 						return;
 					}
 
-					const shouldConvertSentMessages = e2eRoom.shouldConvertSentMessages({ msg });
+					room.encrypted ? e2eRoom.resume() : e2eRoom.pause();
```
sampaiodiego, line 64: did we starting executing code like this? I've always suggested to do the following, it is much cleaner IMO.. ternary should be used only attributions..
```suggestion
					if (room.encrypted) {
						e2eRoom.resume();
					} else {
						e2eRoom.pause();
					}
```

sampaiodiego (changes requested):

apps/meteor/client/lib/chats/flows/uploadFiles.ts
debdutdeb: +1 from me. Goes for both ternaries and &&s, ||s (room.encrypted && e2eRoom.resume()). 

Also reduces more noise down the road in diffs.

yash-rajpal: Agreed. Done!

In my defense, I copied this piece of code from someplace else, hence the negligence. 😬 

Will keep this in mind moving forward, thanks!

apps/meteor/client/lib/chats/flows/uploadFiles.ts
yash-rajpal: Done!

yash-rajpal:
This PR solved a regression caused by encryption of file description - #31840 , we reverted the encryption of file description feature - #32324 with the idea of delivering full file encryption in next release, so this regression no longer makes sense as the main PR was reverted.

PR #32318, opened 2024-04-26
fix(pdf-transcript): Don't error out when trying to process an image bigger than NATS max payload

KevLehman:
https://rocketchat.atlassian.net/browse/CORE-398

## Further comments

Fixed error handling for files bigger than NATS max allowed payload. This should prevent PDFs from erroring out when generating from rooms that contain heavy images. Also added projections to user/rooms db calls


ee/packages/omnichannel-services/src/OmnichannelTranscript.ts
```diff
@@ -232,8 +232,20 @@ export class OmnichannelTranscript extends ServiceClass implements IOmnichannelT
...
+				} catch (e: unknown) {
+					this.log.error(`Failed to get file ${file._id}`, e);
+					// Push empty buffer so parser processes this as "unsupported file"
+					files.push({ name: file.name, buffer: null });
+
+					if ((e as Error).message === 'MAX_PAYLOAD_EXCEEDED') {
+						this.log.error(
+							`File is too big to be processed by NATS. See NATS config for allowing bigger messages to be sent between services`,
+						);
+					}
```
debdutdeb, line 247: should we just increase or have some sort of file-split mechanism to work around it? like send 2MB-2MB-2MB -> 6MB file?

debdutdeb: ps: can be done later, not a blocker for this pr imo.

KevLehman: That would be cool, but yeah, for now, this is what we have.

Would be nice to send some streams over nats or maybe allowing the services to get a buffer directly with some package, idk.

(Or just chunked buffers, that works too)

.changeset/silly-clocks-return.md
```diff
@@ -0,0 +1,7 @@
+---
+'@rocket.chat/omnichannel-services': patch
+'@rocket.chat/core-services': patch
+'@rocket.chat/meteor': patch
+---
+
+Fixed error handling for files bigger than NATS max allowed payload. This should prevent PDFs from erroring out when generating from rooms that contain heavy images. Also added projections to user/rooms db calls
```
debdutdeb, line 7:
```suggestion
Fixed error handling for files bigger than NATS max allowed payload. This should prevent PDFs from erroring out when generating from rooms that contain heavy images. 
```

debdutdeb (changes requested):

apps/meteor/server/services/translation/service.ts
```diff
@@ -22,7 +22,7 @@ export class TranslationService extends ServiceClassInternal implements ITransla
 	}
 
 	// Use translate when you want to translate to the user's language, or server's as a fallback
-	async translate(text: string, user: IUser): Promise<string> {
+	async translate(text: string, user: AtLeast<IUser, 'language'>): Promise<string> {
```
MarcosSpessatto, line 25: Let's dismiss this unrelated change, also, according to the function implementation the language is not required.

ee/packages/omnichannel-services/src/OmnichannelTranscript.ts
```diff
@@ -232,8 +232,20 @@ export class OmnichannelTranscript extends ServiceClass implements IOmnichannelT
...
-				files.push({ name: file.name, buffer: fileBuffer, extension: uploadedFile.extension });
+				try {
+					const fileBuffer = await uploadService.getFileBuffer({ file: uploadedFile });
+					files.push({ name: file.name, buffer: fileBuffer, extension: uploadedFile.extension });
+				} catch (e: unknown) {
+					this.log.error(`Failed to get file ${file._id}`, e);
+					// Push empty buffer so parser processes this as "unsupported file"
+					files.push({ name: file.name, buffer: null });
+
+					if ((e as Error).message === 'MAX_PAYLOAD_EXCEEDED') {
```
MarcosSpessatto, line 243: Ideally, we should not tie this to the NATS error message, but since we don't have an abstraction for this, we could put a TODO comment to refactor this when it's possible. WDYT?

ee/packages/omnichannel-services/src/OmnichannelTranscript.ts
```diff
@@ -275,7 +287,9 @@ export class OmnichannelTranscript extends ServiceClass implements IOmnichannelT
 		}
 		this.currentJobNumber++;
 		try {
-			const room = await LivechatRooms.findOneById(details.rid);
+			const room = await LivechatRooms.findOneById<Pick<IOmnichannelRoom, '_id' | 'v' | 'closedAt' | 'servedBy'>>(details.rid, {
+				projection: { _id: 1, v: 1, closedAt: 1, servedBy: 1 },
+			});
```
MarcosSpessatto, line 292: It seems this is not related to the issue, if so, we can open a new PR for this improvement if necessary.

ee/packages/omnichannel-services/src/OmnichannelTranscript.ts
```diff
@@ -346,11 +360,11 @@ export class OmnichannelTranscript extends ServiceClass implements IOmnichannelT
 
 	private async pdfFailed({ details, e }: { details: WorkDetailsWithSource; e: Error }): Promise<void> {
 		this.log.error(`Transcript for room ${details.rid} by user ${details.userId} - Failed: ${e.message}`);
-		const room = await LivechatRooms.findOneById(details.rid);
+		const room = await LivechatRooms.findOneById<Pick<IOmnichannelRoom, '_id'>>(details.rid, { projection: { _id: 1 } });
```
MarcosSpessatto, line 363: Same as before

ee/packages/omnichannel-services/src/OmnichannelTranscript.ts
```diff
@@ -346,11 +360,11 @@ export class OmnichannelTranscript extends ServiceClass implements IOmnichannelT
...
 		if (!room) {
 			return;
 		}
-		const user = await Users.findOneById(details.userId);
+		const user = await Users.findOneById<Pick<IUser, 'language' | '_id'>>(details.userId, { projection: { _id: 1, language: 1 } });
```
MarcosSpessatto, line 367: Same

ee/packages/omnichannel-services/src/OmnichannelTranscript.ts
```diff
@@ -369,7 +383,7 @@ export class OmnichannelTranscript extends ServiceClass implements IOmnichannelT
 
 	private async pdfComplete({ details, file }: { details: WorkDetailsWithSource; file: IUpload }): Promise<void> {
 		this.log.info(`Transcript for room ${details.rid} by user ${details.userId} - Complete`);
-		const user = await Users.findOneById(details.userId);
+		const user = await Users.findOneById<Pick<IUser, 'language' | '_id'>>(details.userId, { projection: { _id: 1, language: 1 } });
```
MarcosSpessatto, line 386: Same

debdutdeb (dismissed):

ee/packages/omnichannel-services/src/OmnichannelTranscript.ts
```diff
@@ -232,8 +232,20 @@ export class OmnichannelTranscript extends ServiceClass implements IOmnichannelT
...
 				}
 
-				const fileBuffer = await uploadService.getFileBuffer({ file: uploadedFile });
-				files.push({ name: file.name, buffer: fileBuffer, extension: uploadedFile.extension });
+				try {
+					const fileBuffer = await uploadService.getFileBuffer({ file: uploadedFile });
+					files.push({ name: file.name, buffer: fileBuffer, extension: uploadedFile.extension });
+				} catch (e: unknown) {
+					this.log.error(`Failed to get file ${file._id}`, e);
+					// Push empty buffer so parser processes this as "unsupported file"
```
MarcosSpessatto, line 240: I didn't find where this is being handled, can you show me where this is being handled as an unsupported file?

MarcosSpessatto (changes requested):

ee/packages/omnichannel-services/src/OmnichannelTranscript.ts
KevLehman: That happens inside `pdf-worker`. When a `null` buffer is provided, it'll render an "Unsupported file" box on the transcript.

ee/packages/omnichannel-services/src/OmnichannelTranscript.ts
```diff
@@ -346,7 +359,7 @@ export class OmnichannelTranscript extends ServiceClass implements IOmnichannelT
 
 	private async pdfFailed({ details, e }: { details: WorkDetailsWithSource; e: Error }): Promise<void> {
 		this.log.error(`Transcript for room ${details.rid} by user ${details.userId} - Failed: ${e.message}`);
-		const room = await LivechatRooms.findOneById(details.rid);
+		const room = await LivechatRooms.findOneById<Pick<IOmnichannelRoom, '_id'>>(details.rid, { projection: { _id: 1 } });
```
KevLehman, line 362:
```suggestion
		const room = await LivechatRooms.findOneById(details.rid);
```

ee/packages/omnichannel-services/src/OmnichannelTranscript.ts
sampaiodiego: moleculer supports Streams so in theory we could use them to transfer bigger files.. so after receiving the whole stream we could create a buffer from it again.. even better, we could change our APIs to support streams as well

KevLehman: That would be pretty cool! I can spend some time investigating that, since our current way of sending the whole buffer seems a bit overkill. We just do that with images but yeah.

MarcosSpessatto (approved):
Unit tests will be added in a subsequent PR.

PR #32329, opened 2024-04-29
fix: PDFs uploaded by "PDF transcript" feature were returnig 403 when attempting to download

KevLehman:
https://rocketchat.atlassian.net/browse/CORE-399

## Further comments

Added a new setting `Restrict files access to users who can access room` that controls file visibility. This new setting allows users that "can access a room" to also download the files that are there. This is specially important for users with livechat manager or monitor roles, or agents that have special permissions to view closed rooms, since this allows them to download files on the conversation even after the conversation is closed.
New setting is disabled by default and it is mutually exclusive with the setting `Restrict file access to room members` since this allows _more_ types of users to download files.


apps/meteor/app/file-upload/server/lib/FileUpload.ts
```diff
@@ -464,16 +464,30 @@ export const FileUpload = {
...
 			return true;
 		}
 
+		if (fileUploadRestrictedToMembers && !fileUploadRestrictToUsersWhoCanAccessRoom) {
+			const sub = await Subscriptions.findOneByRoomIdAndUserId(file.rid, user._id, { projection: { _id: 1 } });
+			if (!sub) {
+				return false;
+			}
+
+			return !!sub;
```
debdutdeb, line 484: 👀

KevLehman: It can be return true but idk... would it matter the extra cpu cycle to do !! 🤔

debdutdeb: Not that 🙈 I am not *that* critical lol

the if (!sub) is not needed since you are just returning boolean, and already doing !!sub

debdutdeb: which might also sound being critical, but I was confused at first, reading those lines. that's all.

MarcosSpessatto: Agree with Deb, we don't need this if else, we could just return the boolean, if you need something explicit maybe a `sub !== undefined`?

apps/meteor/tests/end-to-end/api/09-rooms.js
```diff
@@ -112,6 +115,14 @@ describe('[Rooms]', function () {
 				done();
 			});
 		});
+		it('create a private channel', async () => {
```
MarcosSpessatto, line 118: Let's delete this channel after all

apps/meteor/tests/end-to-end/api/09-rooms.js
```diff
@@ -239,6 +251,34 @@ describe('[Rooms]', function () {
 			await request.get(fileOldUrl).set(credentials).expect('Content-Type', 'image/png').expect(200);
 		});
 
+		it('should be able to get the file if not member but can access room if setting allows', async () => {
+			await updateSetting('FileUpload_Restrict_to_room_members', false);
+			await updateSetting('FileUpload_Restrict_to_users_who_can_access_room', true);
+
+			await request.get(fileNewUrl).set(credentials).expect('Content-Type', 'image/png').expect(200);
```
MarcosSpessatto, line 258: By using `credentials` the user is a room member, isn't it? Following the test description we should use `userCredentials` instead if we want a not member user, right?

apps/meteor/tests/end-to-end/api/09-rooms.js
```diff
@@ -239,6 +251,34 @@ describe('[Rooms]', function () {
...
+				.attach('file', imgURL)
+				.expect('Content-Type', 'application/json')
+				.expect(200);
+
+			const fileUrl = `/file-upload/${body.message.file._id}/${body.message.file.name}`;
+
+			await request.get(fileUrl).set(userCredentials).expect(403);
+		});
+
+		it('should respect the setting with less permissions when both are true', async () => {
```
MarcosSpessatto, line 275: Wouldn't it be clearer to explicitly set the `FileUpload_ProtectFiles` here?

ee/packages/omnichannel-services/src/OmnichannelTranscript.ts
MarcosSpessatto, line 1: Can we put this small refactor in a subsequent PR? It would be clearer if we keep only the fix here. WDYT?

MarcosSpessatto (changes requested):

ee/packages/omnichannel-services/src/OmnichannelTranscript.ts
KevLehman: Shared on DM: not a refactor, this change is part of the fix 😬

MarcosSpessatto (changes requested):

apps/meteor/app/file-upload/server/lib/FileUpload.ts
```diff
@@ -464,16 +464,26 @@ export const FileUpload = {
...
 		}
 
-		if (!settings.get('FileUpload_Restrict_to_room_members') || !file?.rid) {
+		if (!file?.rid) {
 			return true;
 		}
 
-		const subscription = await Subscriptions.findOneByRoomIdAndUserId(file.rid, user._id, { projection: { _id: 1 } });
+		const fileUploadRestrictedToMembers = settings.get<boolean>('FileUpload_Restrict_to_room_members');
+		const fileUploadRestrictToUsersWhoCanAccessRoom = settings.get<boolean>('FileUpload_Restrict_to_users_who_can_access_room');
```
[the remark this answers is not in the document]
KevLehman: If both are true, that would be an invalid state and the checks will return false as the code won't enter any of the ifs below

ee/packages/omnichannel-services/src/OmnichannelTranscript.ts
```diff
@@ -367,35 +360,74 @@ export class OmnichannelTranscript extends ServiceClass implements IOmnichannelT
...
+		transcriptText: string;
+	}): Promise<IUpload[]> {
+		return Promise.all(
+			roomIds.map((roomId) => {
+				return uploadService.uploadFile({
+					userId: details.userId,
+					buffer,
+					details: {
+						// transcript_{company-name}_{date}_{hour}.pdf
+						name: `${transcriptText}_${data.siteName}_${new Intl.DateTimeFormat('en-US').format(new Date())}_${
```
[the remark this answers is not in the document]
KevLehman: Well, there may be a problem if you're from an asiatic or non romanic country. There they would see the date in probably a format they're not used to see.
however this is only for the filename and such any UTF8 should work. (and it was like this so changing would be more confusing)

ee/packages/omnichannel-services/src/OmnichannelTranscript.ts
```diff
@@ -367,35 +360,74 @@ export class OmnichannelTranscript extends ServiceClass implements IOmnichannelT
...
+					file: transcriptFile,
 					message: {
 						// Translate from service
 						msg: await translationService.translateToServerLanguage('pdf_success_message'),
 					},
 				}),
 				// Send the file to the user who requested it, so they can download it
 				uploadService.sendFileMessage({
-					roomId: rid,
+					roomId: rocketCatFile.rid || '',
```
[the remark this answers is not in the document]
KevLehman: Actually the process should error out before this place if `rocketCatFile` is not valid, so this check was more to shutup typescript. I don't mind changing, but i also don't see much value on adding another if just because the types says `rid` could be undefined (which is not true for 99% of the uploads)

MarcosSpessatto (dismissed):

MarcosSpessatto (approved):

PR #32333, opened 2024-04-29
chore!: Improve permissions check on im endpoints

matheusbsilva137:

 - Improve permissions check on `im` endpoints by using the `permissionsRequired` property.

## Issue(s)

[CORE-340](https://rocketchat.atlassian.net/browse/CORE-340)

## Further comments

[CORE-340]: https://rocketchat.atlassian.net/browse/CORE-340?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/meteor/tests/end-to-end/api/04-direct-message.js
```diff
@@ -355,63 +377,55 @@ describe('[Direct Messages]', function () {
 	});
 
 	describe('/im.messages.others', () => {
```
KevLehman, line 379: Can you add a case for `endpoint disabled but user with permission`?

KevLehman (dismissed):

apps/meteor/tests/end-to-end/api/04-direct-message.js
matheusbsilva137: I believe this is what the `should fail when the endpoint is disabled` does already :thinking: 
I'll update its title to make this clear

MarcosSpessatto (approved):

PR #32343, opened 2024-04-30
chore!: Improve permissions check on permissions endpoints

matheusbsilva137:

 - Improve permissions check on `permissions` endpoints by using the `permissionsRequired` property.

## Issue(s)

[CORE-345](https://rocketchat.atlassian.net/browse/CORE-345)

## Further comments

[CORE-345]: https://rocketchat.atlassian.net/browse/CORE-345?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


MarcosSpessatto (approved):

apps/meteor/tests/end-to-end/api/11-permissions.js
```diff
@@ -54,6 +56,18 @@ describe('[Permissions]', function () {
 	});
 
 	describe('[/permissions.update]', () => {
+		let testUser;
+		let testUserCredentials;
+		before(async () => {
+			testUser = await createUser();
+			testUserCredentials = await login(testUser.username, password);
+			await updatePermission('access-permissions', ['admin']);
```
MarcosSpessatto, line 64: Let's just reset to default `after` all. I know the value will be the same but in case someone adds another test here changing this permission, we could mess up with the subsequent tests

MarcosSpessatto (changes requested):

MarcosSpessatto (approved):

PR #32345, opened 2024-04-30
fix: `BurgerMenu` wrong semantics and colors 

dougfabris:
## Further comments

[CORE-412]

[CORE-412]: https://rocketchat.atlassian.net/browse/CORE-412?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/meteor/client/components/BurgerMenu/BurgerMenuButton.tsx
```diff
@@ -1,35 +1,21 @@
...
-			position='relative'
-			marginInlineEnd={8}
-			className={css`
-				cursor: pointer;
-			`}
-			onClick={onClick}
-		>
-			<BurgerIcon open={open} />
+		<Box position='relative'>
+			<IconButton icon='burger-menu' small data-qa-id='burger-menu' aria-label={t('Open_sidebar')} marginInlineEnd={8} onClick={onClick} />
```
tiagoevanp, line 18: BurguerMenu looks like a sidebar button not a generic menu anymore, maybe is better to change the name of the component... WDYT?

tiagoevanp, line 18: `aria-label` cannot mutate to a close message when the sidebar is opened? why?

tiagoevanp (commented):
No tests for burguer menu button (click and behavior)?

apps/meteor/client/components/BurgerMenu/BurgerMenuButton.tsx
dougfabris: it was never a generic menu, not even a menu, so doesn't make sense using this name, I will change it!

yash-rajpal (approved):
LGTM!

PR #32354, opened 2024-04-30
chore!: Improve permissions check on licenses endpoints

matheusbsilva137:

 - Improve permissions check on `licenses` endpoints by using the `permissionsRequired` property.

## Issue(s)

[CORE-390](https://rocketchat.atlassian.net/browse/CORE-390)

## Further comments

[CORE-390]: https://rocketchat.atlassian.net/browse/CORE-390?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/meteor/ee/server/api/licenses.ts
```diff
@@ -8,13 +8,13 @@ import { hasPermissionAsync } from '../../../app/authorization/server/functions/
 
 API.v1.addRoute(
```
KevLehman, line 9: Shouldn't we remove this endpoint instead of updating it?

KevLehman (changes requested):

apps/meteor/ee/server/api/licenses.ts
matheusbsilva137: I thought about this too, but decided to updated this endpoint regardless of our plans to remove it in the next release. So that even if we delay the removal by any reason, this would have a better permission validation
wdyt? :thinking: I could remove this update too

MarcosSpessatto: I agree with Kevin, since it's already deprecated and the release that is planned for it to be removed is 7.0.0, we should remove it instead.

MarcosSpessatto (changes requested):

apps/meteor/ee/server/api/licenses.ts
matheusbsilva137: I won't touch this endpoint in this PR, then. Let's wait for another one to remove it

MarcosSpessatto (approved):

PR #32366, opened 2024-05-02
fix: add oplog metrics to stream-hub

sampaiodiego:

### DIFF
--- a/apps/meteor/server/database/DatabaseWatcher.ts
+++ b/apps/meteor/server/database/DatabaseWatcher.ts
@@ -228,7 +228,7 @@ export class DatabaseWatcher extends EventEmitter {
 		this.emit(collection, doc);
 	}
 
-	onDoc(callback: (event: { collection: string; doc: RealTimeData<IRocketChatRecord> }) => void): this {
+	onDocument(callback: (event: { collection: string; doc: RealTimeData<IRocketChatRecord> }) => void): this {
 		return super.on('doc', callback);
 	}
 
--- a/apps/meteor/server/database/DatabaseWatcher.ts
+++ b/apps/meteor/server/database/DatabaseWatcher.ts
@@ -228,11 +228,11 @@ export class DatabaseWatcher extends EventEmitter {
 		this.emit(collection, doc);
 	}
 
-	onDoc(callback: (event: { collection: string; doc: any }) => void): this {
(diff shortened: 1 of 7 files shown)


apps/meteor/server/database/DatabaseWatcher.ts
```diff
@@ -218,15 +223,16 @@ export class DatabaseWatcher extends EventEmitter {
...
-			collection,
-			op: doc.action,
-		});
+		this.emit('doc', { collection, doc });
 
 		this.emit(collection, doc);
 	}
 
-	on<T>(collection: string, callback: (event: RealTimeData<T>) => void): this {
+	onDoc(callback: (event: { collection: string; doc: RealTimeData<IRocketChatRecord> }) => void): this {
```
sampaiodiego, line 231: not a fan of `onDoc`.. it could also be `onDocument` or maybe it shouldn't exist? it exists basically to have typings on the subscribing.

apps/meteor/server/modules/watchers/watchers.module.ts
```diff
@@ -268,7 +268,7 @@ export function initWatchers(watcher: DatabaseWatcher, broadcast: BroadcastCallb
 		}
 	});
 
-	watcher.on<ISetting>(Settings.getCollectionName(), async ({ clientAction, id, data, diff }) => {
+	watcher.onCollection<ISetting>(Settings.getCollectionName(), async ({ clientAction, id, data, diff }) => {
 		if (diff && Object.keys(diff).length === 1 && diff._updatedAt) {
```
KevLehman, line 272: Random but I think this check should be on all listeners by default

apps/meteor/server/database/DatabaseWatcher.ts
KevLehman: Vote for `onDocument`, or also `onOplogEntry`

apps/meteor/server/modules/watchers/watchers.module.ts
sampaiodiego: hmm.. I'll take a look at it to see if it makes sense.. thx for pointing it out

apps/meteor/server/database/DatabaseWatcher.ts
sampaiodiego: renamed to `onDocument`

PR #32374, opened 2024-05-06
fix: Modals opened by apps won't close when `executeViewCloseHandler` is not implemented

KevLehman:
[rocketchat.atlassian.net/browse/CORE-401](https://rocketchat.atlassian.net/browse/CORE-401)

## Further comments

Fixes an issue with some apps that didn't implement executeViewCloseHandler. This causes opened modals to be open forever on UI. This is because when the UI attempts to close it, it calls the aforementioned handler, and since it didn't exist, apps engine errored out.

This returned an empty response to the UI, which ignored the response and continued to show the view.

### DIFF
--- a/.changeset/plenty-turkeys-move.md
+++ b/.changeset/plenty-turkeys-move.md
@@ -1,0 +1,15 @@
+---
+"@rocket.chat/meteor": patch
+"@rocket.chat/apps": patch
+"@rocket.chat/core-services": patch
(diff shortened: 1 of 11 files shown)


ggazzo (changes requested):
What does this bump do? It seems to me that it's a fix, could you add a changeset explaining what it fixes? otherwise people won't know what was fixed

ggazzo (approved):

PR #32376, opened 2024-05-06
fix: Moderators with `Bypass Time limits` permission cannot edit messages

KevLehman:
## Issue(s)

https://rocketchat.atlassian.net/browse/SUP-552
## Further comments

Fixes a problem with how the permission was being checked on the UI. Since `Moderator` is a role scoped to subscriptions, the call to `hasPermission` should have had the `scope` (which for a subscription, is the room).

Since this value was not being provided, it returned always false


KevLehman:
Note: the bypass actions are minimum 1min, so i decided to not to add UI tests on this one.

If you know a way i can do them, lmk 😄

MarcosSpessatto (changes requested):
We have more usage of this permission check, can we verify if the same fix needs to be applicable for the other cases?

https://github.com/RocketChat/Rocket.Chat/blob/develop/apps/meteor/client/methods/updateMessage.ts#L53
https://github.com/RocketChat/Rocket.Chat/blob/develop/apps/meteor/client/views/room/contextualBar/RoomFiles/hooks/useMessageDeletionIsAllowed.ts#L12

We also have the same check on the server side.
https://github.com/RocketChat/Rocket.Chat/blob/develop/apps/meteor/app/authorization/server/functions/canDeleteMessage.ts#L37
https://github.com/RocketChat/Rocket.Chat/blob/develop/apps/meteor/app/lib/server/methods/updateMessage.ts#L56

Shouldn't this be fixed there as well?

Finally, can we add some UI tests to ensure this behavior? And the same for API if there are no tests yet?

KevLehman:
Hey, fixed the other places, thanks 🤗 i didn't notice we had that much usages of this permission.

On the tests part, i placed a comment just before yours 😬 about testability. Basically, for "testing" that the feature works, the minimum time allowed for this setting is 1min, and honestly I don't want to "wait" for one minute on the CI for testing this.

Was thinking on some Units, but since none of the files have Units, if we go this way, I'd prefer to have them as a separate task, wdyt?

MarcosSpessatto (approved):

PR #32393, opened 2024-05-09
fix: Monitors dissapearing from Unit upon edit

KevLehman:

Introduced here: #30734

## Issue(s)

https://rocketchat.atlassian.net/browse/SUP-572
## Further comments

Fixed an issue that caused the UI to send an invalid _id for monitors that were already saved on the unit. This caused the BE to ignore these monitors (as the _id didn't exist) and save only the newly added items to the list (if any)

This was caused because on an old refactor, the destructuring of the `monitors` property changed from `monitorId` (on JS) to `_id` (in TS)

As the elements were coming from the `unit_monitors` collection instead of the `users` collection, here `_id` represented something totally different. When the BE eliminated the elements that were not valid monitors from the array, these were excluded (which is correct) and not saved.


apps/meteor/tests/e2e/omnichannel/omnichannel-units.spec.ts
```diff
@@ -117,6 +122,24 @@ test.describe('OC - Manage Units', () => {
...
+			await expect(poOmnichannelUnits.inputSearch).toBeVisible();
+			await poOmnichannelUnits.search(editedUnitName);
+			await expect(poOmnichannelUnits.findRowByName(editedUnitName)).toBeVisible();
+			await poOmnichannelUnits.findRowByName(editedUnitName).click();
+			await expect(poOmnichannelUnits.contextualBar).toBeVisible();
+
+			await expect(poOmnichannelUnits.inputMonitors).toHaveText(/user2/);
+			await expect(poOmnichannelUnits.inputMonitors).toHaveText(/user3/);
+		});
+
```
MarcosSpessatto, line 142: Can we add a new test case removing one user from the list (of 2), save it, and see if it behaves correctly?

MarcosSpessatto (dismissed):

apps/meteor/tests/e2e/omnichannel/omnichannel-units.spec.ts
dougfabris, line 1: this amount of assertion checking if the elements are visible is really necessary?

eg: checking if the contextualBar is visible. Theres a step to to click in a button on contextualBar, the test you fail if its no visible, so what's the purpose of the assertion? I don't see a real value using this approach, kinda repetitive in my point of view

dougfabris (changes requested):

apps/meteor/tests/e2e/omnichannel/omnichannel-units.spec.ts
KevLehman: Good question. IDK. I copied the pattern from the tests above it and adapted my checks 😬

KevLehman: BUT, removed the unneeded expects

dougfabris (dismissed):
LGTM!

KevLehman:
Hey @dougfabris i forgot to add a changeset before, mind to review this one again? 😬

dougfabris (approved):

MartinSchoeler (approved):

PR #32412, opened 2024-05-10
fix: Users in role list not rendering properly

dougfabris:


apps/meteor/client/views/admin/permissions/UsersInRole/UsersInRoleTable/UsersInRoleTable.tsx
```diff
@@ -1,79 +1,106 @@
...
-			{users.length > 0 && (
+			{isLoading && (
+				<GenericTable>
+					<GenericTableHeader>{headers}</GenericTableHeader>
+					<GenericTableBody>
+						<GenericTableLoadingTable headerCells={2} />
+					</GenericTableBody>
+				</GenericTable>
+			)}
+			{isSuccess && users && users.length > 0 && (
```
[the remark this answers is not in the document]
dougfabris: I can't, because we're converting some params. Take a look at line 54

apps/meteor/tests/e2e/administration.spec.ts
MarcosSpessatto, line 1: Shouldn't we add tests to cover the problems mentioned in the PR description? Create tests throwing error and the empty state, and make sure it shows the toast...

MarcosSpessatto (changes requested):

apps/meteor/tests/e2e/administration.spec.ts
dougfabris: I already created a test to check if its rendering the empty state on `should remove user1 as moderator of target channel`. It **shouldn't** show the toast, its annoying.
About the error state:  I'm not sure how simulate it during the e2e test since its the prevention state, it renders when something get wrong on the request.

MarcosSpessatto (approved):

PR #32425, opened 2024-05-13
feat: Allow different time units for retention policy

gabriellsh:

image

## Issue(s)

[CORE-302](https://rocketchat.atlassian.net/browse/CORE-302)
[CORE-303](https://rocketchat.atlassian.net/browse/CORE-303)

## Further comments

[CORE-302]: https://rocketchat.atlassian.net/browse/CORE-302?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/meteor/server/startup/migrations/v305.ts
```diff
@@ -0,0 +1,29 @@
...
+				await Settings.findByIds(maxAgeSettings).toArray()
+			).map(async ({ value, packageValue, _id }) => {
+				// if the package value isn't 30 the setting has already been updated
+				if (packageValue !== 30) {
+					return;
+				}
+
+				const newPackageValue = convertDaysToMs(30);
+				// make sure the check passes if the number is 0;
+				const newValue = Number(value) + 1 ? convertDaysToMs(Number(value)) : newPackageValue;
```
KevLehman, line 23:
```suggestion
				const newValue = Number(value) >= 0 ? convertDaysToMs(Number(value)) : newPackageValue;
```

KevLehman (changes requested):

apps/meteor/client/views/admin/settings/inputs/TimespanSettingInput.tsx
```diff
@@ -0,0 +1,136 @@
...
+	const t = useTranslation();
+
+	const [timeUnit, setTimeUnit] = useState<TIMEUNIT>(getHighestTimeUnit(Number(value)));
+	const [internalValue, setInternalValue] = useState<number>(msToTimeUnit(timeUnit, Number(value)));
+
+	const handleChange: FormEventHandler<HTMLInputElement> = (event) => {
+		const newValue = Math.max(1, Number(event.currentTarget.value)) || 1;
+
+		try {
+			setInternalValue(newValue);
```
Gustrb, line 87: I think we should validate the newvalue (in the server as well I guess), not sure how high it would have to be, but I guess if we put the value in days, and it is large enough we would have Infinity as the value in the retention policy, since we are doing this chain of multiplications: `24 * 60 * 60 * 1000`

gabriellsh: The max supported value is something along the lines of 169 years, or very close to it. I'll make sure it cannot be bigger than 100 years just to make sure

apps/meteor/server/startup/migrations/v305.ts
```diff
@@ -0,0 +1,29 @@
...
+const maxAgeSettings = ['RetentionPolicy_MaxAge_Channels', 'RetentionPolicy_MaxAge_Groups', 'RetentionPolicy_MaxAge_DMs'];
+
+const convertDaysToMs = (days: number) => days * 24 * 60 * 60 * 1000;
+addMigration({
+	version: 305,
+	name: 'Convert retention policy max age from days to milliseconds',
+	async up() {
+		await Promise.all(
+			(
+				await Settings.findByIds(maxAgeSettings).toArray()
```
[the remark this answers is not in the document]
gabriellsh: I had tried doing this but for some reason it didn't work, perhaps it was not IT the problem but something else I did wrong at the time. Thanks for fixing it

apps/meteor/server/startup/migrations/v305.ts
```diff
@@ -0,0 +1,22 @@
...
+
+const maxAgeSettings = ['RetentionPolicy_MaxAge_Channels', 'RetentionPolicy_MaxAge_Groups', 'RetentionPolicy_MaxAge_DMs'];
+
+const convertDaysToMs = (days: number) => days * 24 * 60 * 60 * 1000;
+addMigration({
+	version: 305,
+	name: 'Convert retention policy max age from days to milliseconds',
+	async up() {
+		// if the package value isn't 30 the setting has already been updated
+		await Settings.find({ _id: { $in: maxAgeSettings }, packageValue: 30 }, { projection: { _id: 1, value: 1 } })
```
MarcosSpessatto, line 13: I think it would be better to retrieve all the rows and store them in a variable, then use Promise.all to update all the values in parallel

apps/meteor/server/startup/migrations/v305.ts
MarcosSpessatto, line 1: @RocketChat/architecture are we okay about adding this migration here for a minor release?

MarcosSpessatto (changes requested):
Doing some local tests, looks like there's something weird going on with the feature.

1 - The callout seems wrong
image

2- Executing the branch locally, the values for the new settings look weird.
Screenshot 2024-05-17 151351

3 - It's not possible to use the same granularity if the user wants to override the global retention policy, we can use only days (old behavior)
Screenshot 2024-05-17 151731

4 - It seems the messages are not being pruned after the specified time (1 minute), I'm not sure if it's respecting what the callout says... Or if it's just not working.
image

gabriellsh:
@MarcosSpessatto 
> 1 - The callout seems wrong
Will be fixes
> 2 - Executing the branch locally, the values for the new settings look weird.
Seems like an issue with the migration. I'm suspecting the packageValue is changing before the migration runs, so I'll have to do something different there
> 3 - 3 - It's not possible to use the same granularity if the user wants to override the global retention policy, we can use only days (old behavior)
There will be a fix to the input value but this form specifically will stay as it is for now. It was not part of the scope as far as I'm aware
> 4 - It seems the messages are not being pruned after the specified time (1 minute), I'm not sure if it's respecting what the callout says... Or if it's just not working.
This depends on the cron running, which is managed through settings above the ones I changed. Both callout and warning will be changing to reflect the cron time.

apps/meteor/server/startup/migrations/v305.ts
gabriellsh: This was being done just like described. @KevLehman made some changes in commit [ee32132](https://github.com/RocketChat/Rocket.Chat/pull/32425/commits/ee321325787018fbb3efea786f516e423bd32057) that removed the Promise.all

KevLehman: I changed this to be this way because at max, this thing will process 3 entries. So doing them in parallel or a bulk write doesn't give any extra perf gain

apps/meteor/client/views/admin/settings/inputs/TimespanSettingInput.tsx
```diff
@@ -0,0 +1,122 @@
...
+		return TIMEUNIT.minutes;
+	}
+
+	const hours = msToTimeUnit(TIMEUNIT.hours, value);
+	if (hours % 24 !== 0) {
+		return TIMEUNIT.hours;
+	}
+
+	return TIMEUNIT.days;
+};
```
debdutdeb, line 26: is this correct? the name of the function isn't very clear to me, but **my guess is** it's trying to see the longest the number goes. 

I see two issues here,

first is what if in L15, minutes is literally 61? 

Second, this assumes two types, ints and floats, and assumes all conversions up to `%` are in ints. That is not true.

msToTimeUnit can return with a decimal part while integer part is > 0 (or multiplier for the respective unit).

imagine passing a huge number (milliseconds) that is divisible by 1000 (for second), 60 (for minute), 60 (hour), **plus 1**. 3600001.

When converted to minutes using your function, you get 60.<something>. Pass through the first branch and you get a false positive. 60.001 (e.g.) % 60 != 0, yes, but that doesn't mean the highest unit for that number is minutes, it is actually hour.

debdutdeb (dismissed):
didn't read the rest yet as this stood out immediately.

apps/meteor/client/views/admin/settings/inputs/TimespanSettingInput.tsx
gabriellsh: The input does not accept not-whole values, so you can’t put in `1,5 hours` or something similar. If you want to be more granular you should go to the previous time unit. This is more of a UX issue than implementation issue, if you still think this shouldn't be as it is, the design team should be included in the discussion.

IMO the way it is provides enough granularity without increasing the complexity of the UI too much. What you pointed out about 60 minutes being treated as an hour, and 61 minutes being treated as minutes is actually how I intended this to be. We do not save the selected time unit, just the time in milliseconds, so the purpose of the function is to get the highest time unit that is a whole number.

If you check [this line](https://github.com/RocketChat/Rocket.Chat/pull/32425/files#diff-18a2cccf8325476785643aba44e04dc1d395aa4eb9727f2b98e47b7e6d6566b6R18) of the unit tests you'll see that I tested the specific case you're pointing out.

apps/meteor/server/startup/migrations/xrun.ts
```diff
@@ -6,11 +8,49 @@ const { MIGRATION_VERSION = 'latest' } = process.env;
...
+	['RetentionPolicy_MaxAge_Channels', 'RetentionPolicy_TTL_Channels'],
+	['RetentionPolicy_MaxAge_Groups', 'RetentionPolicy_TTL_Groups'],
+	['RetentionPolicy_MaxAge_DMs', 'RetentionPolicy_TTL_DMs'],
+]);
+
+const moveRetentionSetting = async () => {
+	const convertDaysToMs = (days: number) => days * 24 * 60 * 60 * 1000;
+
+	const promises: Array<Promise<any>> = [];
+	await Settings.find({ _id: { $in: Array.from(maxAgeSettingMap.keys()) } }, { projection: { _id: 1, value: 1 } }).forEach(
```
ggazzo, line 21:
```suggestion
	await Settings.find({ _id: { $in: Array.from(maxAgeSettingMap.keys()) }, value: { $ne: -1  }}, { projection: { _id: 1, value: 1 } }).forEach(
```

ggazzo (changes requested):

apps/meteor/server/startup/migrations/xrun.ts
```diff
@@ -6,11 +8,48 @@ const { MIGRATION_VERSION = 'latest' } = process.env;
...
+	await Settings.find(
+		{ _id: { $in: Array.from(maxAgeSettingMap.keys()) }, value: { $ne: -1 } },
+		{ projection: { _id: 1, value: 1 } },
+	).forEach(({ _id, value }) => {
+		if (!maxAgeSettingMap.has(_id)) {
+			throw new Error(`moveRetentionSetting - Setting ${_id} equivalent does not exist`);
+		}
+
+		promises.push(
+			Settings.update(
```
KevLehman, line 30:
```suggestion
			Settings.updateOne(
```

apps/meteor/server/startup/migrations/xrun.ts
```diff
@@ -6,11 +8,48 @@ const { MIGRATION_VERSION = 'latest' } = process.env;
...
+const maxAgeSettingMap = new Map([
+	['RetentionPolicy_MaxAge_Channels', 'RetentionPolicy_TTL_Channels'],
+	['RetentionPolicy_MaxAge_Groups', 'RetentionPolicy_TTL_Groups'],
+	['RetentionPolicy_MaxAge_DMs', 'RetentionPolicy_TTL_DMs'],
+]);
+
+const moveRetentionSetting = async () => {
+	const convertDaysToMs = (days: number) => days * 24 * 60 * 60 * 1000;
+
+	const promises: Array<Promise<any>> = [];
```
KevLehman, line 20:
```suggestion
	const promises: Array<Promise<UpdateResult>> = [];
```

KevLehman (changes requested):

apps/meteor/client/views/room/hooks/useRetentionPolicy.ts
```diff
@@ -89,11 +89,11 @@ export const useRetentionPolicy = (
 		filesOnly: useSetting('RetentionPolicy_FilesOnly') as boolean,
 		doNotPrunePinned: useSetting('RetentionPolicy_DoNotPrunePinned') as boolean,
 		appliesToChannels: useSetting('RetentionPolicy_AppliesToChannels') as boolean,
-		maxAgeChannels: useSetting('RetentionPolicy_MaxAge_Channels') as number,
+		maxAgeChannels: useSetting('RetentionPolicy_TTL_Channels') as number,
```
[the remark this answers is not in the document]
gabriellsh: The casting here is necessary since useSetting even after passing the type parameter says it could be undefined. I won't fix this now because I'd have to refactor the entire file.

apps/meteor/client/lib/convertTimeUnit.ts
```diff
@@ -0,0 +1,58 @@
...
+const isValidTimespan = (timespan: number): boolean => {
+	if (Number.isNaN(timespan)) {
+		return false;
+	}
+
+	if (!Number.isFinite(timespan)) {
+		return false;
+	}
+
+	if (timespan < 0) {
```
MarcosSpessatto, line 16: Would a timespan of 0 be a valid value? 🤔

gabriellsh: It already is today. It'll delete every message once the cron runs.

MarcosSpessatto (dismissed):

dougfabris (approved):
@gabriellsh this is a really good and solid work. 
Really well coded and you added great tests to cover your work. 

My only concern here is related to the consistency of the feature. In my opinion we should have keep the same time unit possibilities when overriding retention policy in the rooms. But it seems a product decision to follow like this. 

Because of that, I'm approving it on behalf of frontend team!

ggazzo (approved):

PR #32428, opened 2024-05-13
fix: saveSettings endpoint allowing NaN to be stored on Int settings

MartinSchoeler:
[CORE-418]

Fixes issue that allowed empty strings to be saved in the int settings, causing an NaN value on the DB

[CORE-418]: https://rocketchat.atlassian.net/browse/CORE-418?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


matheusbsilva137 (changes requested):
Could you please add an API end-to-end test for this? I believe you could use the `settings/:_id` POST endpoint

.changeset/happy-windows-drum.md
```diff
@@ -0,0 +1,5 @@
+---
+"@rocket.chat/meteor": patch
+---
+
+Stops the backend from saving invalid values for int settings (NaN for example)
```
KevLehman, line 5: Fixed an issue that allowed `saveSettings` method to save `NaN` values on numeric settings.

KevLehman (changes requested):

apps/meteor/tests/end-to-end/api/24-methods.js
```diff
@@ -3108,4 +3108,27 @@ describe('Meteor.methods', function () {
 			});
 		});
 	});
+
+	describe('[@saveSettings]', () => {
+		it('should return an error when trying to save a "NaN" value', () => {
```
MarcosSpessatto, line 3113: Can we add other test cases? e.g. boolean, strings, null, undefined, and finally one ensuring we can save a number (int, float) value.

MarcosSpessatto (changes requested):

MarcosSpessatto (approved):

apps/meteor/app/lib/server/methods/saveSettings.ts
```diff
@@ -83,6 +83,12 @@ Meteor.methods<ServerMethods>({
 							break;
 						case 'int':
 							check(value, Number);
```
debdutdeb, line 85: shouldn't this fail if value is an empty string?

debdutdeb (changes requested):
the value is fetched from db, and looks like can possibly be NaN. this tells me the save action is what should always set empty values to zero, and not be handled as an exception for this specific method.

please help me if I'm missing something.

debdutdeb:
we don't need a migration for this either, each setting is tried (like in a court) to insert or update, you can add a validation there, compare with cached value and fix it. 

@MarcosSpessatto wdyt?

apps/meteor/app/lib/server/methods/saveSettings.ts
debdutdeb: but I am still curious, how can "value" be NaN for an empty string, after getting serialized.

debdutdeb: ok I see it, client doesn't just serialize as is.

according to ddp code, while not possible through UI, but is if somebody manually sends Infinity. 

https://github.com/meteor/meteor/blob/7e2abe9464c215aace47d20f737e59243a819871/packages/ejson/ejson.js#L145-L155

suggest using Number.isFinite instead, wdyt @MarcosSpessatto

KevLehman (approved):

apps/meteor/app/lib/server/methods/saveSettings.ts
debdutdeb: @KevLehman not sure if this is resolved yet 😬

KevLehman: I was purposefuly ignoring it :trollface: 

Being serious, that's a possibility 🤔 not sure if it's something that common (to have someone sending `infinity` on the endpoints, but guess it could happen if they have some automated way of doing things.

Bet it would hurt nobody to change it @MartinSchoeler (and it should work the same)

Also, you can add a new test case for sending Infinity or -Infinity to the endpoint

matheusbsilva137: Couldn't we use `isInteger`? This way floats, `NaN`, `Infinity` and `-Infinity` would be avoided :thinking:

apps/meteor/app/lib/server/methods/saveSettings.ts
```diff
@@ -83,6 +83,12 @@ Meteor.methods<ServerMethods>({
...
+							if (Number.isNaN(value)) {
```
matheusbsilva137, line 86:
```suggestion
							if (!Number.isInteger(value)) {
```

matheusbsilva137 (changes requested):

apps/meteor/app/lib/server/methods/saveSettings.ts
debdutdeb: didn't know isInteger is a thing. came to know about isFinite even couple days ago. Doesn't matter to me as long as it's covering all cases 👍🏼

apps/meteor/tests/end-to-end/api/24-methods.js
```diff
@@ -3108,4 +3108,27 @@ describe('Meteor.methods', function () {
...
+					}),
+				})
+				.expect(200)
+				.expect((res) => {
+					expect(res.body).to.have.property('success', true);
+					const parsedBody = JSON.parse(res.body.message);
+					expect(parsedBody).to.have.property('error');
+					expect(parsedBody.error).to.have.property('error', 'Invalid setting value NaN');
+				});
+		});
```
matheusbsilva137, line 3132:
```suggestion
		});

		it('should return an error when trying to save a "Infinity" value', () => {
			request
				.post(api('method.call/saveSettings'))
				.set(credentials)
				.send({
					message: JSON.stringify({
						msg: 'method',
						id: '13',
						method: 'saveSettings',
						params: [[{ _id: 'Message_AllowEditing_BlockEditInMinutes', value: { $InfNaN: 1 } }]],
					}),
				})
				.expect(200)
				.expect((res) => {
					expect(res.body).to.have.property('success', true);
					const parsedBody = JSON.parse(res.body.message);
					expect(parsedBody).to.have.property('error');
					expect(parsedBody.error).to.have.property('error', 'Invalid setting value Infinity');
				});
		});

		it('should return an error when trying to save a "-Infinity" value', () => {
			request
				.post(api('method.call/saveSettings'))
				.set(credentials)
				.send({
					message: JSON.stringify({
						msg: 'method',
						id: '13',
						method: 'saveSettings',
						params: [[{ _id: 'Message_AllowEditing_BlockEditInMinutes', value: { $InfNaN: -1 } }]],
					}),
				})
				.expect(200)
				.expect((res) => {
					expect(res.body).to.have.property('success', true);
					const parsedBody = JSON.parse(res.body.message);
					expect(parsedBody).to.have.property('error');
					expect(parsedBody.error).to.have.property('error', 'Invalid setting value -Infinity');
				});
		});
```

matheusbsilva137 (changes requested):

matheusbsilva137 (approved):

MarcosSpessatto (approved):

yash-rajpal (approved):

PR #32430, opened 2024-05-13
fix: OTR conversation closes automatically when any of the peers goes offline

### LINKED ISSUE #7974: OTR e-mail notifications: remove message payload
When absent and I receive E-Mail messages, the message I receive contains encrypted text. 
bildschirmfoto_2017-08-30_11-05-28

Instead, replace the message with a sentence "Encrypted Message". Like the desktop notifications already do.

KevLehman:
## Issue(s)

https://rocketchat.atlassian.net/browse/SUP-447
## Steps to test or reproduce

Fixes #7974
## Further comments

The fix involves adding a listener to user presence in 2 places:
- On acknowledge (this is, the user that starts the OTR session after the other party clicks on "yes" to start OTR)
- On stablishment (the user clicking on "yes" on the modal)

This will allow any of the parties to end the session when the other party goes offline.
This will also trigger when one of the parties logs out.

When one of the involved user logs out, we'll clear the `instancesByRoomId` object. This will allow the user to login to a clean sheet and open any OTR conversation again (after a refresh, but this is caused by another bug).


apps/meteor/app/otr/client/OTRRoom.ts
```diff
@@ -110,6 +111,36 @@ export class OTRRoom implements IOTRRoom {
 		}
 	}
 
+	// Starts listening to other user's status changes and end OTR if any of the Users goes offline
+	// this should be called in 2 places: on acknowledge (meaning user accepted OTR) or on establish (meaning user initiated OTR)
+	listenToUserStatus(userId: IUser['_id']): void {
```
yash-rajpal, line 116: Maybe we should use the `Presence` API on client side instead of directly listening to the stream.

For more context, please check the `usePresence` hook.

apps/meteor/app/otr/client/OTR.ts
```diff
@@ -24,6 +24,10 @@ class OTR implements IOTR {
 		this.instancesByRoomId[rid] = otrRoom;
 		return this.instancesByRoomId[rid];
 	}
+
+	closeAllInstances(): void {
```
yash-rajpal, line 28: During closing all instances of OTR, we should also call the cleanup functions of each of the OTR-rooms, in this case calling the `end` function for each room.

yash-rajpal (approved):

apps/meteor/app/otr/client/OTR.ts
KevLehman: Yeah, was calling .end here first but problem is that at some point end calls removeoldotrmessages which is an auth call but since onLogout happens after, theres no auth.

Wdyt of changing those hooks to just to internal cleaning? Like reset but with the rest of the calls from end.

apps/meteor/app/otr/client/OTRRoom.ts
```diff
@@ -110,6 +111,28 @@ export class OTRRoom implements IOTRRoom {
...
+		Presence.listen(userId, (event: any) => {
```
yash-rajpal, line 117: We should also detach this listener using `Presence.stop`

yash-rajpal, line 117: The event would be of type `UserPresence`

apps/meteor/app/otr/client/OTRRoom.ts
```diff
@@ -110,6 +112,34 @@ export class OTRRoom implements IOTRRoom {
...
+	// this should be called in 2 places: on acknowledge (meaning user accepted OTR) or on establish (meaning user initiated OTR)
+	listenToUserStatus(userId: IUser['_id']): void {
+		Presence.listen(userId, (event: UserPresence | undefined) => {
+			if (!event) {
+				return;
+			}
+			if (event.status === UserStatus.OFFLINE) {
+				console.warn(`OTR Room ${this._roomId} ended because ${userId} went offline`);
+				this.end();
+				Presence.stop(userId, () => {
```
yash-rajpal, line 125: Removing listener won't work this way, as will have to pass the reference of the function which was passed at the time of adding the listener.

apps/meteor/app/otr/client/OTR.ts
```diff
@@ -24,6 +24,16 @@ class OTR implements IOTR {
...
+		// Resets state, but doesnt emit events
+		// Other party should receive event and fire events
+		Object.values(this.instancesByRoomId).forEach((instance) => {
```
KevLehman, line 31: Note: we're not firing events because the `onLogout` callback is emited after the user has logged out. This means that emitting the event & calling `deleteAllOTRMessages` call won't work.

We considered with @yash-rajpal to create a new `beforeLogout` hook similar to `afterLogoutCleanup` callback we have already, that way, we could emit those events right before the user logs out (at least the deleteMessages)

yash-rajpal (approved):
LGTM!

PR #32431, opened 2024-05-14
fix: sidebar last message E2EE

hugocostadev:

If End-to-End Encryption is enabled and you delete the last message in a room the is updated to the last message without a type (t) attribute set which will be last message before the encryption was enabled for this room or no message if no unencrypted message is available.

If you send an encrypted message the lastMessage is updated correctly but on deletion the method Rooms.resetLastMessageById is called which will get the last message by calling Messages.getLastVisibleMessageSentWithNoTypeByRoomId. But an e2e message has the t set to "e2e" and will be ignored.

I did: 
- renamed `Messages.getLastVisibleMessageSentWithNoTypeByRoomId` function to `Messages.getLastVisibleUserMessageSentByRoomId`
- add a query to include e2e type of messages
- added a end-to-end test

## Steps to test or reproduce

1. Enable End-to-End Encryption in Settings
2. Start a new room (no matter which type)
3. Enable encryption for this room
4. Write multiple messages with different content
5. Last message will be shown on the left side (Expanded view)
6. Delete the last message
7. Reload the page if the message on the left side is not updating
8. Find the wrong last message in the preview on the left.

### Expected behavior:
The last message is shown nevermind if its an encrypted or unencrypted message

## Further comments

https://rocketchat.atlassian.net/browse/E2EE2-12


yash-rajpal (dismissed):
LGTM!

yash-rajpal (approved):

tassoevan (approved):

PR #32432, opened 2024-05-14
chore: remove message column on moderation console

csuadev:
This PR removes the message preview from the Moderation Console since it didn't wasn't useful for the admin and we need to lead them to click on the reports from a user to see all the messages from that specific user and check if any moderation is needed. 

Before:
image

After:
image

## Issue(s)
[https://rocketchat.atlassian.net/browse/CORE-4](https://rocketchat.atlassian.net/browse/CORE-4)


apps/meteor/client/views/admin/moderation/ModerationConsoleTableRow.tsx
```diff
@@ -22,6 +22,13 @@ const ModerationConsoleTableRow = ({ report, onClick, isDesktopOrLarger }: Moder
 		return room.fname || room.name;
 	});
 
+	const normalizeMessage = () => {
+		if (message.startsWith('[ ](')) {
```
MarcosSpessatto, line 26: Don't we have the whole message available here? If so, would it be better to use the existing `isQuoteAttachment` instead of checking the string? What if someone sends this exact text in the message?

MarcosSpessatto (changes requested):

apps/meteor/client/views/admin/moderation/ModerationConsoleTableRow.tsx
```diff
@@ -22,6 +22,13 @@ const ModerationConsoleTableRow = ({ report, onClick, isDesktopOrLarger }: Moder
 		return room.fname || room.name;
 	});
 
+	const normalizeMessage = () => {
+		if (message.startsWith('[ ](') && message.includes('\n')) {
```
MarcosSpessatto, line 26: I wonder if it would not be better to have regex here to do the recognization instead. @ggazzo can you give your input here please?

MarcosSpessatto (changes requested):

MarcosSpessatto (changes requested):
If the PR _fixes_ the problem, we need to change the PR title to `fix:` and add a changeset.

apps/meteor/tests/e2e/message-moderation.spec.ts
```diff
@@ -0,0 +1,63 @@
...
+      await expect(poModeration.toastSuccess).toBeVisible();
+    });
+
+    test('Admin can see the reported messages', async ({ page }) => {
+      await page.goto('/admin/moderation/messages');
+      await expect(poModeration.reportedMessagesTab).toBeVisible();
+      await expect(poModeration.findRowByName(quotedMessage)).toBeVisible();
+    });
+  });
+});
```
MarcosSpessatto, line 63: Also, I'm concerned with this... What happens if we report a quoted message with an image instead of text? Or a video? Or any thing different from a text? 🤔

MarcosSpessatto (changes requested):

apps/meteor/client/views/admin/moderation/ModerationConsoleTableRow.tsx
dougfabris: @csuadev if we're going to render the message, we should use the parser to render it properly. But since messages wasn't made to be render in a table I'm wondering if we should have this. Is it really valuable for the users? shouldn't only the context bar be responsible to show the message content since it has enough space to that? I think would be great to discuss it with the designers, since we do have a lot of corner cases rendering messages in a table.

apps/meteor/tests/e2e/page-objects/fragments/home-content.ts
```diff
@@ -159,6 +159,10 @@ export class HomeContent {
 		return this.page.locator('[data-qa-id="edit-message"]');
 	}
 
+	get btnOptionQuoteMessage(): Locator {
+		return this.page.getByRole('button', { name: 'Quote' });
+	}
+
```
MarcosSpessatto, line 165: I think we don't need this anymore...

MarcosSpessatto (changes requested):
Would it be possible to fix the indentation of the tests files?

MarcosSpessatto (dismissed):

dougfabris (approved):

PR #32444, opened 2024-05-15
fix: CAS user merge not working

pierre-lehnen-rc:

This PR fixes an issue introduced on 6.6 where the CAS login process failed to find existing users with the same user as the one currently logging in if they had been created by a different login process.

## Issue(s)

[SUP-576](https://rocketchat.atlassian.net/browse/SUP-576)

## Further comments

[SUP-576]: https://rocketchat.atlassian.net/browse/SUP-576?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


KevLehman (commented):
You think it's there a way to write tests for this one?

pierre-lehnen-rc:
> You think it's there a way to write tests for this one?

Not at this moment; I'll try to bundle a CAS container so we can write unit tests for it but this PR is meant to be merged into older releases as well so adding that here would only overcomplicate things.

KevLehman (approved):

PR #32452, opened 2024-05-17
fix: `QueueInactivityMonitor` not processing inquiries

KevLehman:
## Issue(s)

https://rocketchat.atlassian.net/browse/SUP-579
## Further comments

Apparently the callback that created the job to remove the items from the queue was checking a `createdAt` property that didn't even exists on types. Since the function was typed as `any`, this created the perfect place for this happen.

Types were updated, missing property check was removed (now we just check for _updatedAt) and changed the use of a global value since we have the value in scope from the `watch` call.


KevLehman:
Note: no tests for this as the minimum time for this job to run is one min and waiting for the CI to run after one min is no good

Will create a separate task to add unit tests

apps/meteor/ee/app/livechat-enterprise/server/hooks/afterInquiryQueued.ts
```diff
@@ -1,32 +1,28 @@
+import type { ILivechatInquiryRecord } from '@rocket.chat/core-typings';
 import moment from 'moment';
 
 import { settings } from '../../../../../app/settings/server';
 import { callbacks } from '../../../../../lib/callbacks';
 import { OmnichannelQueueInactivityMonitor } from '../lib/QueueInactivityMonitor';
 import { cbLogger } from '../lib/logger';
 
-let timer = 0;
+const withTimer = (timer: number) => {
```
MarcosSpessatto, line 9: I know you mentioned the task for creating the tests for this in a separated PR, but I think it would be good to have a unit test for this function at least. This way we can ensure the high order function is behaving as expected.

MarcosSpessatto (changes requested):

apps/meteor/ee/app/livechat-enterprise/server/hooks/afterInquiryQueued.ts
KevLehman: Added 😬

MarcosSpessatto (dismissed):

MartinSchoeler (approved):

MartinSchoeler (approved):

apps/meteor/ee/tests/unit/apps/livechat-enterprise/server/hooks/afterInquiryQueued.spec.ts
```diff
@@ -0,0 +1,120 @@
...
+	});
+
+	it('should call the callback on settings.watch with proper values', () => {
+		const func = settingStub.watch.getCall(0).args[1];
+
+		func(1);
+		expect(callbackStub.add.callCount).to.be.equal(1);
+
+		func(2);
+		expect(callbackStub.add.callCount).to.be.equal(2);
```
MarcosSpessatto, line 58: This counts how many times the callback is being called, but the callback itself is being added only to the first call, right? So it's kind of useless, right?

KevLehman: Not 100% useless, since we're checking it actually reached this point. Even when the original `callback.add` would do nothing, the code will call it anyways so this is a test of "when value is > 0, callbacks.add is called"

PR #32459, opened 2024-05-20
fix: prevent OTR in E2EE

hugocostadev:

Attempting to use OTR and E2E at the same time just renders a harsh.

## Steps to test or reproduce

1. Enable E2E for a direct message channel
2. Assemble an OTR session
3. Try to send a message

## Further comments

https://rocketchat.atlassian.net/browse/E2EE2-13


MarcosSpessatto (changes requested):
Would it be possible to add some UI tests to ensure the behavior?

KevLehman:
Random question from my ignorance: if the action is not allowed, shouldn't we at least gray out the button so users cannot click? 🤔 
(nt sure if u are already doing it and i cannot see)

MarcosSpessatto (approved):

PR #32470, opened 2024-05-22
refactor(client): Use subcomponents for `InfoPanel` and `UserInfo`

rique223:

Refactored all places that use `InfoPanel` and `UserInfo` components as objects and implemented subcomponents and barrel exports/imports for both of them.

Jira task: [CONN-268](https://rocketchat.atlassian.net/browse/CONN-268)
## Further comments

[CONN-268]: https://rocketchat.atlassian.net/browse/CONN-268?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


dougfabris (dismissed):
Do you mind changing this notation to all places that uses InfoPanel subcomponents, so we can remove the exported object from `apps/meteor/client/components/InfoPanel/index.ts`

also prefer using `import { subComponents } from '../InfoPanel/'`

Btw, why so many commits?

rique223:
> Do you mind changing this notation to all places that uses InfoPanel subcomponents, so we can remove the exported object from `apps/meteor/client/components/InfoPanel/index.ts`
> 
> also prefer using `import { subComponents } from '../InfoPanel/'`
> 
> Btw, why so many commits? 

Sure, will do. About the commits I am not entirely sure, this branch was checked out of the feat/new-user-panel-pending branch that was recently merged and when it happened git automatically rebased this branch to the develop and duplicated a lot of the commits that were merged in the other branch. I assumed that since the git diff is correct it wouldn't be a problem, but if there's anyway to remove those commits I'd like do know, I'll do a little research on this as well.

tassoevan (approved):

rique223:
> > Do you mind changing this notation to all places that uses InfoPanel subcomponents, so we can remove the exported object from `apps/meteor/client/components/InfoPanel/index.ts`
> > also prefer using `import { subComponents } from '../InfoPanel/'`
> > Btw, why so many commits?
> 
> Sure, will do. About the commits I am not entirely sure, this branch was checked out of the feat/new-user-panel-pending branch that was recently merged and when it happened git automatically rebased this branch to the develop and duplicated a lot of the commits that were merged in the other branch. I assumed that since the git diff is correct it wouldn't be a problem, but if there's anyway to remove those commits I'd like do know, I'll do a little research on this as well.

For Future Reference:

The problem Doug mentioned was caused by squash merging the feat/new-user-panel-pending branch. When a branch is squash merged, all the commits in the pull request (PR) are combined into a single commit on the base branch. Because of this, Git loses the ability to track the individual commits from the original branch. If a child branch was created from the feat/new-user-panel-pending branch before the squash merge, it still contains those individual commits. As a result, when the child branch is compared or a PR is created from it, Git will not recognize that those commits were already merged, leading to duplicate commit history in the PR just like happened here;

tassoevan (dismissed):

aleksandernsilva (dismissed):

apps/meteor/client/components/UserInfo/index.ts
```diff
@@ -1,13 +1,7 @@
...
 import UserInfoUsername from './UserInfoUsername';
 
-export default Object.assign(UserInfo, {
-	Action: UserInfoAction,
-	Avatar: UserInfoAvatar,
-	Info: InfoPanel.Text,
-	Label: InfoPanel.Label,
-	Username: UserInfoUsername,
-});
+export { UserInfo, UserInfoAction, UserInfoAvatar, InfoPanelText as UserInfoText, InfoPanelLabel as UserInfoLabel, UserInfoUsername };
```
tiagoevanp, line 7: Why not export directly? Like:

`export { default as UserInfo } from './UserInfo';`

And why InfoPanel components are importing to export with another name? How about to have one single generic component on top of both places used by the User and Panel? InfoText and InfoLabel?

tiagoevanp (changes requested):

tiagoevanp (approved):

tassoevan (approved):

aleksandernsilva (approved):

apps/meteor/client/components/InfoPanel/InfoPanel.stories.tsx
```diff
@@ -25,62 +35,37 @@ const fakeRoom = createFakeRoom();
...
+		</InfoPanelSection>
+		<InfoPanelSection>
 			<RetentionPolicyCallout room={fakeRoom} />
-		</InfoPanel.Section>
+		</InfoPanelSection>
 	</InfoPanel>
 );
 Default.storyName = 'InfoPanel';
-
-// export const Archived = () => <Contextualbar height={800}>
```
dougfabris, line 64: ??

dougfabris (changes requested):

dougfabris (approved):

PR #32471, opened 2024-05-22
feat: Use `application/octet-stream` as a fallback media type to avoid "Unknown media type" errors

matheusbsilva137:

 - Allow any file with an unknown media type (the ones not defined in IANA, Apache or nginx MIME types will be classified as `application/octet-stream` as default if the new setting is disabled) to be uploaded by using a `application/octet-stream` as the default/fallback MIME type;
 - Normalize how MIME types are identified throughout the application (by using the `mime` package in the back-end too)
 - Add alert to "Blocked Media Types" setting
Captura de tela de 2024-06-17 13-15-45

## Steps to test or reproduce

Access a room and try uploading a `drawio` file in it (such as the one available in the end-to-end tests added by this PR); **_Expected behavior:_** The file should be accepted and uploaded succesfully. Check the `rocketchat_uploads` DB collection to confirm the file has been uploaded with the `application/octet-stream` media type (default one for unknown media extensions);
**_Current behavior:_** the file isn't accepted and an "Unknown media type not accepted" error is thrown.

## Further comments

[SUP-531](https://rocketchat.atlassian.net/browse/SUP-531)

[SUP-531]: https://rocketchat.atlassian.net/browse/SUP-531?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/meteor/app/api/server/lib/getUploadFormData.ts
```diff
@@ -78,12 +80,13 @@ export async function getUploadFormData<
...
 				return returnError(new MeteorError('error-file-too-large'));
 			}
+			const mimeTypeFromFilename = mime.lookup(filename);
 
 			uploadedFile = {
 				file,
 				filename,
 				encoding,
-				mimetype,
+				mimetype: typeof mimeTypeFromFilename === 'string' ? mimeTypeFromFilename : mimetype,
```
MarcosSpessatto, line 89: What does this variable being a string mean? 🤔 In which cases this will not be a string?

apps/meteor/client/views/room/body/hooks/useFileUploadDropTarget.ts
```diff
@@ -55,7 +56,8 @@ export const useFileUploadDropTarget = (): readonly [
 		const uniqueFiles = getUniqueFiles();
 
 		const uploads = Array.from(uniqueFiles).map((file) => {
-			Object.defineProperty(file, 'type', { value: mime.lookup(file.name) });
+			const fileMimeType = mime.lookup(file.name);
+			Object.defineProperty(file, 'type', { value: fileMimeType ?? (rejectUnknownMediaTypes ? undefined : 'application/octet-stream') });
```
MarcosSpessatto, line 60: Any reason for doing this instead of copying the object with the new prop? 🤔

apps/meteor/client/views/room/composer/messageBox/MessageBoxActionsToolbar/hooks/useFileUploadAction.ts
```diff
@@ -25,8 +26,9 @@ export const useFileUploadAction = (disabled: boolean): GenericMenuItemProps =>
 		const handleUploadChange = async () => {
 			const { mime } = await import('../../../../../../../app/utils/lib/mimeTypes');
 			const filesToUpload = Array.from(fileInputRef?.current?.files ?? []).map((file) => {
+				const fileMimeType = mime.lookup(file.name);
 				Object.defineProperty(file, 'type', {
-					value: mime.lookup(file.name),
+					value: fileMimeType ?? (rejectUnknownMediaTypes ? undefined : 'application/octet-stream'),
```
MarcosSpessatto, line 31: Same as before

apps/meteor/server/settings/file-upload.ts
```diff
@@ -25,6 +25,11 @@ export const createFileUploadSettings = () =>
 			i18nDescription: 'FileUpload_MediaTypeBlackListDescription',
 		});
 
+		await this.add('FileUpload_UnknownMediaTypeProtection', true, {
```
MarcosSpessatto, line 28: I think it would be better to rename this to something like `Protect_...`

apps/meteor/tests/e2e/file-upload.spec.ts
```diff
@@ -54,4 +63,33 @@ test.describe.serial('file-upload', () => {
 		await expect(poHomeChannel.content.getFileDescription).toHaveText('lst_description');
 		await expect(poHomeChannel.content.lastMessageFileName).toContainText('lst-test.lst');
 	});
+
+	test('expect not to send drawio file when the "Reject unknown media types" setting is enabled', async ({ api, page }) => {
```
MarcosSpessatto, line 67: It would be better to use the setting name here since we have nothing mentioning this "Reject...", it could cause confusion

apps/meteor/tests/e2e/file-upload.spec.ts
```diff
@@ -54,4 +63,33 @@ test.describe.serial('file-upload', () => {
...
+		await setSettingValueById(api, 'FileUpload_UnknownMediaTypeProtection', true);
+
+		await page.reload();
+		await poHomeChannel.content.sendFileMessage('diagram.drawio');
+		await expect(poHomeChannel.content.btnModalConfirm).not.toBeVisible();
+	});
+
+	test('expect send drawio file succesfully when the "Reject unknown media types" setting is disabled', async ({ api, page }) => {
```
MarcosSpessatto, line 75: Same

apps/meteor/tests/e2e/file-upload.spec.ts
```diff
@@ -54,4 +63,33 @@ test.describe.serial('file-upload', () => {
...
+		await page.reload();
+		await poHomeChannel.content.sendFileMessage('diagram.drawio');
+		await poHomeChannel.content.descriptionInput.fill('drawio_description');
+		await poHomeChannel.content.btnModalConfirm.click();
+
+		await expect(poHomeChannel.content.getFileDescription).toHaveText('drawio_description');
+		await expect(poHomeChannel.content.lastMessageFileName).toContainText('diagram.drawio');
+	});
+
+	test('expect not to send drawio file when the "Reject unknown media types" setting is disabled, but the default media type is blocked', async ({ api, page }) => {
```
MarcosSpessatto, line 87: Same

apps/meteor/tests/end-to-end/api/09-rooms.js
MarcosSpessatto, line 1: Can we create API tests to cover this as well?

MarcosSpessatto (changes requested):
Let's rewrite the description please, we are mentioning ` Reject unknown media types`, but the name/description for the setting is completely different (which is good in my opinion)

apps/meteor/app/api/server/lib/getUploadFormData.ts
matheusbsilva137: `mime.lookup` may return a `string`, `string[]` or `undefined`
 - `undefined` in case no MIME type has been found for the extension in the DB
 - `string[]` will only happen in case we append an alternative MIME type to a extension (which we won't)
 So I'm just making sure we have the expected type here

apps/meteor/client/views/room/body/hooks/useFileUploadDropTarget.ts
matheusbsilva137: > Any reason for doing this instead of copying the object with the new prop? 🤔

I understand using `Object.defineProperty` avoids reassignments to this property, but I can't spot a specific reason for doing this here :thinking: I'll assign the property directly, then
Let's see what front-enders think of this too, I may be missing something

matheusbsilva137: > Any reason for doing this instead of copying the object with the new prop? 🤔

I don't think so :thinking: the objects in the files array are not reused by the function that calls `onFileDrop`
So I think it's ok to copy the object here instead of assigning a prop to it

apps/meteor/tests/end-to-end/api/09-rooms.js
matheusbsilva137: Done!

apps/meteor/app/api/server/lib/getUploadFormData.ts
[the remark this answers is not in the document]
matheusbsilva137: Hmm, good catch. Done!

apps/meteor/server/settings/file-upload.ts
matheusbsilva137: We decided not to have a setting anymore. Having a fallback media type will be the default behavior

apps/meteor/client/views/room/body/hooks/useFileUploadDropTarget.ts
matheusbsilva137: I tried returning `return { ...file, type: typeof fileMimeType === 'string' ? fileMimeType : 'application/octet-stream' };` instead of assigning the property but it caused some errors. So I assume that the client depends on the `file` objects that are being changed here
I'd say it has something to do with this file: `apps/meteor/client/lib/chats/flows/uploadFiles.ts`
So I kept this property definition as it was

matheusbsilva137: > also, what if lookup returns empty string. ?? is being used here

Fixed! Thanks for the heads-up

apps/meteor/client/views/room/body/hooks/useFileUploadDropTarget.ts
```diff
@@ -55,7 +55,8 @@ export const useFileUploadDropTarget = (): readonly [
 		const uniqueFiles = getUniqueFiles();
 
 		const uploads = Array.from(uniqueFiles).map((file) => {
-			Object.defineProperty(file, 'type', { value: mime.lookup(file.name) });
+			const fileMimeType = mime.lookup(file.name);
+			Object.defineProperty(file, 'type', { value: typeof fileMimeType === 'string' ? fileMimeType : 'application/octet-stream' });
```
MarcosSpessatto, line 59: Wouldn't it be better to create a helper function? It seems it's being used in multiple places...

apps/meteor/client/views/room/composer/messageBox/MessageBoxActionsToolbar/hooks/useFileUploadAction.ts
```diff
@@ -25,8 +25,9 @@ export const useFileUploadAction = (disabled: boolean): GenericMenuItemProps =>
 		const handleUploadChange = async () => {
 			const { mime } = await import('../../../../../../../app/utils/lib/mimeTypes');
 			const filesToUpload = Array.from(fileInputRef?.current?.files ?? []).map((file) => {
+				const fileMimeType = mime.lookup(file.name);
 				Object.defineProperty(file, 'type', {
-					value: mime.lookup(file.name),
+					value: typeof fileMimeType === 'string' ? fileMimeType : 'application/octet-stream',
```
MarcosSpessatto, line 30: Same

MarcosSpessatto (changes requested):

apps/meteor/client/views/room/body/hooks/useFileUploadDropTarget.ts
matheusbsilva137: Good idea, done!

MarcosSpessatto (approved):

MarcosSpessatto (approved):

Issue #31906, opened 2024-03-06
Media Type Not Accepted


DustpaN-Spb:
Look like a last change is a error
Because
Was:				invalidContentType: Boolean(file.type && !fileUploadIsValidContentType(file.type)),
Now:			invalidContentType: !(file.type && fileUploadIsValidContentType(file.type)),
If we can't detrminate a MIME type of a file (file.type) it will return a false and all not recognised types will be not work at all even if "Accepted Media Types" is empty.
Please revert
https://forums.rocket.chat/t/cant-upload-ovpn-file/19455

_Originally posted by @DustpaN-Spb in https://github.com/RocketChat/Rocket.Chat/issues/31346#issuecomment-1975862098_

Visbhavesh:
I wanted to work in thus issue please Assign this issue to me.

DustpaN-Spb:
I do not have a permissions to assign a issue.
Who can do it?

richardturnerfos:
I don't understand why more people aren't complaining. All sort of files are being rejected suddenly and we have no way of allowing them.

ggazzo:
Considering security, if we cannot identify the mime type, I believe we should block it, otherwise this validation is useless. Don't you agree?


feel free to work on a proposal, there is no need to sign anyone for this, however I believe that simply inverting the boolean is not a safe option

DustpaN-Spb:
If a user can't send a file in a custom format because the file doesn't have a mime type, it doesn't look right.
If the "Accepted Media Types" in the admin panel is empty, I think the user should be able to send any file and it is not security related.

unwnstr:
> Considering security, if we cannot identify the mime type, I believe we should block it, otherwise this validation is useless. Don't you agree?
> 
> feel free to work on a proposal, there is no need to sign anyone for this, however I believe that simply inverting the boolean is not a safe option

This is purely client only file type check. User can just edit js code in f12 browser debugger and remove file type check. So there is no security anyway.
But when your team is working with CAD files, or some code, it is really annoying to rename file extension every time you need to upload it.

reetp:
> Considering security, if we cannot identify the mime type, I believe we should block it, otherwise this validation is useless. Don't you agree?

Nope, I don't unless RocketChat can validate correctly, which it does not appear to do. Mime detection seems broken.

Desktop test:

`xdg-mime query filetype  OVPN_John.ovpn`

text/plain

Upload to Rocket:

> Media Type not accepted

[...]

Issue #31692, opened 2024-02-08
Can't upload ovpn file


emoxam:
Can't send ovpn file
### Steps to reproduce:

1. try to send ovpn file
2. see "media types not accepted"

### Expected behavior:

i can send it as it was before.

### Actual behavior:

"media types not accepted" error

### Server Setup Information:

- Version of Rocket.Chat Server:  installed:          6.6.0             (1602) 361MB -
- Operating System:  ubuntu 
- Deployment Method: snap
- Number of Running Instances: 1
- DB Replicaset Oplog: 
- NodeJS Version: 
- MongoDB Version:

### Client Setup Information

- Desktop App or Browser Version: 3.9.12
- Operating System: Windows

dolohow:
Same problem here with 6.6.0, can't send gpg files and in admin settings I am not blocking any file type. In previous version (6.5.x) it did work.

IgorOhrimenko:
Not only ovpn file does not upload. You can try create emty file with name `1` and gave the same problem "Media Type Not Accepted:"

roman-bergman:
I got the following mimetype for *.ovpn: text/plain.
I added this mimetype to "Accepted Media Types" but i get the same error: "media types not accepted".

See my comment https://github.com/RocketChat/Rocket.Chat/pull/31346#issuecomment-1966163365

IvanBadikovJoyrun:
Check `mimeTypes.ts` in the `app/utils/lib`. 

The extension of your file is not defined in the [mime-db](https://www.npmjs.com/package/mime-db?activeTab=code) repo, so the type of your file should become undefined during lookup in the `app/client/views/room/body/hooks/useFileUploadDropTarget.ts` file line 57. 

I guess you should define your own type in the `mimeTypes.ts` in `app/utils/lib` in order to make it work.

emoxam:
How to define it ?

roman-bergman:
> Перевірте `mimeTypes.ts`в `app/utils/lib`.
> 
> Розширення вашого файлу не визначено в [ репозиторії mime-db ](https://www.npmjs.com/package/mime-db?activeTab=code) , тому тип вашого файлу має стати невизначеним під час пошуку в `app/client/views/room/body/hooks/useFileUploadDropTarget.ts`рядок файлу 57.
> 
> Я думаю, вам слід визначити свій власний тип у `mimeTypes.ts`в `app/utils/lib`щоб це працювало.
[...]

PR #32478, opened 2024-05-23
fix: Accepted Media Types settings validation

hugocostadev:

The setting `Accepted Media Types` in the File Upload menu was not working as expected since an Empty setting or  wildcard `*` value should allow all types of files if they are not in the `Blocked` settings

## Issue(s)

fix https://github.com/RocketChat/Rocket.Chat/issues/31906
fix https://github.com/RocketChat/Rocket.Chat/issues/31692

## Steps to test or reproduce

1. Login to rocket chat.
2. Set "Accepted Media Types" setting in "File Upload"  with `*` or leave it empty
3. Try sending any file that is not in the Blocked List
4. The file is not uploaded

## Further comments

https://rocketchat.atlassian.net/browse/CONN-214


KevLehman (dismissed):
This one should have API tests

hugocostadev:
> This one should have API tests

I created tests for the function that I changed and that's being used in frontend and backend. 
I agree that would be good to have api tests for that endpoint, I can do it, but I have others priorities in the moment. So would take a while

rique223 (dismissed):

apps/meteor/app/utils/lib/restrictions.ts
```diff
@@ -17,37 +15,45 @@ const fileUploadMediaBlackList = function (customBlackList: string): string[] |
...
 	}
 
-	if (blackList && isTypeOnList(type, blackList)) {
+	if (whiteList && type && isTypeOnList(type, whiteList)) {
+		return true;
+	}
+
+	if (!type && whiteList) {
 		return false;
 	}
```
matheusbsilva137, line 52:
```suggestion
	if (whiteList) {
		return isTypeOnList(type, whiteList);
	}
```
Is believe this works just the same, doesn't it? Looks easier to read

matheusbsilva137 (changes requested):

apps/meteor/app/utils/lib/restrictions.ts
hugocostadev: It does, but I had to make some smalls adjustments to `isTypeOnList` func ... Thanks for your feedback

rique223 (approved):

matheusbsilva137 (approved):

PR #32484, opened 2024-05-23
refactor!: removed roomNameExists method 

abhinavkrin:

Removed roomNameExists method as it has been replaced by rooms.nameExists endpoint
## Further comments

[CORE-351](https://rocketchat.atlassian.net/browse/CORE-351)

[CORE-351]: https://rocketchat.atlassian.net/browse/CORE-351?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


MarcosSpessatto (changes requested):
Can we take a look at this test case? I think there's something weird since it's expecting an error when no room is found...

https://github.com/RocketChat/Rocket.Chat/blob/develop/apps/meteor/tests/end-to-end/api/09-rooms.js#L425

abhinavkrin:
> Can we take a look at this test case? I think there's something weird since it's expecting an error when no room is found...
> 
> https://github.com/RocketChat/Rocket.Chat/blob/develop/apps/meteor/tests/end-to-end/api/09-rooms.js#L425

I looked at it, the test was for an invalid roomId which should not be there as we dont accept a roomId for this endpoint.
So I changed this test to check for invalid params which makes the test more clear.
Also, I added a test to check if false is returned if roomName is not found.

apps/meteor/tests/end-to-end/api/09-rooms.js
```diff
@@ -422,7 +422,22 @@ describe('[Rooms]', function () {
...
+				})
+				.expect(200)
+				.expect((res) => {
+					expect(res.body).to.have.property('success', true);
+					expect(res.body).to.have.property('exists', false);
+				})
+				.end(done);
+		});
+
+		it('should return an error when send an invalid param', (done) => {
```
MarcosSpessatto, line 440:
```suggestion
		it('should return an error when the require parameter (roomName) is not provided', (done) => {
```

MarcosSpessatto (changes requested):

MarcosSpessatto (approved):

ggazzo (approved):

PR #32485, opened 2024-05-23
fix: Override retention policy missing ignore threads option

### LINKED ISSUE #18738: [3.6.0] Manual channel purging and the retention policy does not delete threads.
### Description:

Manual purging and the retention policy does not delete threads

### Steps to reproduce:

1. create a channel
2. write a msg, then create a answer as thread
3. trigger Prune Messages (make sure that Do not prune Threads is not active)
4. reload page on webbrowser with [F5] or in desktop app with CTRL+R

### Expected behavior:

Threads are gone/deleted

### Actual behavior:

Thread still there after reloading.

### Server Setup Information:

- Version of Rocket.Chat Server:  3.6.0
- Operating System: GNU/Debian 10 alias Buster
- Deployment Method: tar
- Number of Running Instances: 5
- DB Replicaset Oplog: Yes
- NodeJS Version: 12.16.1
- MongoDB Version: 4.0.20

### Client Setup Information

- Desktop App or Browser Version: Both (Webbrowser Firefox/Opera)
- Operating System: Windows / Linux

### Relevant logs:
`
Meteor ➔ method cleanRoomHistory -> userId: xxxxx, arguments: [{"roomId":"xxxxxx","latest":"9999-12-31T23:59:59.000Z","oldest":"0001-01-01T00:00:00.000Z","inclusive":false,"limit":2000,"excludePinned":false,"ignoreDiscussion":false,"filesOnly":false,"fromUsers":[],"ignoreThreads":false}]
`

### LINKED ISSUE #19731: Retention Policy not deleting Threads
### Description:

It seems like the retention policy doesn't delete threads even though administration is configured correctly.

**Our Retention Policy settings in Admin:**
image
image

### Steps to reproduce:

1.  Create new private room
2. Open Room-Info
3. Click Edit -> Prune
4. Tick "Automatically prune old messages" and "Override global retention policy"
5. Set "Maximum message age in days (default: 30)" to 1
6. Click "Save"
7. Write something in this Channel
8. Answer with Thread
9. Wait 1 day

### Expected behavior:

Thread gets deleted

### Actual behavior:

Thread doesn't get deleted.

### Server Setup Information:

- Version of Rocket.Chat Server: 3.9.0
- Deployment Method: docker

### LINKED ISSUE #22384: [3.14.4] Threads always ignored when a channel overrides overrides the global retention policy
### Description:
When a channel is configured to override the global retention policy, there is no option to set the "Do not prune Threads" option within the channel's own retention policy, and no default value is set for the `Room.retention.ignoreThreads` setting, therefore threads are always ignored during message pruning on a channel that overrides the global retention policy.

### Steps to reproduce:

1. Enable the retention policy at `/admin/RetentionPolicy` and ensure that the global policy does not apply to channels
2. Open the `Room information` panel within a channel (e.g. `/channel/general/channel-settings`)
3. Click `Edit` and enable `Automatically prune old messages` and `Override global retention policy`
4. **Observe that there is no `Do not prune Threads` option**
5. Post a message to the channel, then reply to the message in order to create a thread
6. Update the message timestamps (`ts` field in the MongoDB documents), so that they are older than the configured maximum age
7. Trigger message pruning
8. **Observe that the thread is not deleted, despite being older than the configured maximum age**

### Expected behavior:

1. There should be an option to configure the `Do not prune threads` retention policy setting on a per-channel basis
2. When the `Do not prune Threads` setting is disabled, threaded messages that are older than the maximum age should be deleted when the message pruning task runs

### Actual behavior:

1. There is no option to configure the `Do not prune threads` retention policy setting on a per-channel basis
2. Threaded messages are never deleted if a channel overrides the global policy

### Server Setup Information:

- Version of Rocket.Chat Server: 3.14.4
- Deployment Method: snap

### Additional context

I created a debug build to investigate this issue. I logged `room.retention` at the following location and found that `ignoreThreads` was `undefined` in all cases where a channel had a retention policy enabled: https://github.com/RocketChat/Rocket.Chat/blob/119028b17f71df8797a46234cfee7cc6c8ee9217/app/retention-policy/server/cronPruneMessages.js#L50

The above then calls `cleanRoomHistory()` and passes the `ignoreThreads` variable to it. The body of the following `if` statement is never executed for rooms that override the global policy, due to `ignoreThreads` always being `undefined`: https://github.com/RocketChat/Rocket.Chat/blob/157b882d38b34b7b79dea527316554fe33f70891/app/lib/server/functions/cleanRoomHistory.js#L42-L50

The `ignoreThreads` variable also gets passed to `Messages.removeByIdPinnedTimestampLimitAndUsers()` here: https://github.com/RocketChat/Rocket.Chat/blob/157b882d38b34b7b79dea527316554fe33f70891/app/lib/server/functions/cleanRoomHistory.js#L52

Again, the body of the following `if` statement is never executed because `ignoreThreads` is always `undefined`: https://github.com/RocketChat/Rocket.Chat/blob/157b882d38b34b7b79dea527316554fe33f70891/app/models/server/models/Messages.js#L970-L973

### Temporary Workaround

This bug can be resolved by logging on to the server, connecting to the MongoDB instance, and running the following commands:

```
use parties
db.rocketchat_room.update({"retention.enabled": true}, {"$set": {"retention.ignoreThreads":false}}, {"multi":true})
```

This sets the `retention.ignoreThreads` setting to false on all rooms/channels that have a retention policy enabled, which will result in threaded messages being deleted by the message pruning task (provided they are older than the maximum age).

dougfabris:

Adds the missing `ignoreThreads` param fixing the issue not allowing ignoring threads when overriding retention policy

## Issue(s)

Closes #19731
Closes #22384
Closes #18738

## Further comments

[CORE-487]

[CORE-487]: https://rocketchat.atlassian.net/browse/CORE-487?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/meteor/tests/e2e/retention-policy.spec.ts
```diff
@@ -171,6 +174,15 @@ test.describe.serial('retention-policy', () => {
 				await expect(poHomeChannel.tabs.room.getMaxAgeLabel('15')).toBeVisible();
 				await expect(poHomeChannel.tabs.room.inputRetentionMaxAge).toHaveValue('365');
 			});
+
+			test('should ignore threads be checked accordingly with the global default value', async () => {
```
MarcosSpessatto, line 178: Can we add a case overriding the value and ensuring the value is being overwritten?

MarcosSpessatto (changes requested):

MarcosSpessatto (approved):

PR #32487, opened 2024-05-24
chore: allow remove cache on application error

ggazzo:

### DIFF
--- a/apps/meteor/client/views/root/AppErrorPage.tsx
+++ b/apps/meteor/client/views/root/AppErrorPage.tsx
@@ -1,4 +1,4 @@
-import { Box, States, StatesIcon, StatesSubtitle, StatesTitle } from '@rocket.chat/fuselage';
+import { Box, States, StatesAction, StatesActions, StatesIcon, StatesSubtitle, StatesTitle } from '@rocket.chat/fuselage';
 import type { ErrorInfo, ReactElement } from 'react';
 import React from 'react';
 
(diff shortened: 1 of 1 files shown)


apps/meteor/client/views/root/AppErrorPage.tsx
```diff
@@ -15,6 +15,17 @@ const AppErrorPage = (_props: AppErrorPageProps): ReactElement => {
 				<StatesIcon name='error-circle' />
 				<StatesTitle>Application Error</StatesTitle>
 				<StatesSubtitle>The application GUI just crashed.</StatesSubtitle>
```
jeanfbrito, line 17: `GUI` for end users may be too technical

I suggest a more friendly text to make the user more comfortable in use the new feature:

Oops! Something Went Wrong
We encountered an unexpected error and need to reload the application.

What Happened:
The application experienced an issue that requires a reload to continue.

What to Do:
Click the button below to reload the application.
If the problem persists, please contact our support team.
[Reload Application]

We apologize for the inconvenience and appreciate your understanding.

rodrigok (approved):

apps/meteor/client/views/root/AppErrorPage.tsx
ggazzo: I understand and agree with your suggestion, but this is a UI improvement, there are several improvements (actually showing the error that occurred, etc.) and it is not the objective of the pr here, I am 100% in favor of improving this page, just not in this pr

PR #32493, opened 2024-05-24
feat: Improve Engagement Dashboard's "Channels" tab performance

matheusbsilva137:

 - Replace duplicated aggregation execution (which was used only to count documents) by a single request for counting all rooms;
 - Apply `limit` as soon as possible in the aggregation so as to carry less data to the next steps;
 - Only return channels, groups and DM rooms from this endpoint. Livechat rooms have been removed since we don't track their messages data.
 - Added a `hideRoomsWithNoActivity` param to the `engagement-dashboard/channels/list` endpoint, which enables a faster search by not returning rooms that had no message sent in the given period. This param is being used as `true` by default in our client, which will be the default behavior starting on 7.0 (when this param will be removed).

## Steps to test or reproduce

The Engagement Dashboard's "Channels" tab should keep working just like before, but faster when there's a huge amount of rooms in the workspace. Channels should always be listed sorted by their amount of messages when no `sort` param is provided.
What should be validated:
 - Rooms returned by the `engagement-dashboard/channels/list` endpoint should be sorted by their amount of new messages in the analyzed timespan (decreasingly);
 -  The `engagement-dashboard/channels/list` endpoint should accept a new boolean `hideRoomsWithNoActivity` param, which enables a faster search by removing rooms that had no messages sent in the analyzed timespan. This is specially relevant for workspaces that have a huge amount of rooms (this option is being used in the UI);
 - A deprecation warning log should be displayed whenever the `engagement-dashboard/channels/list` endpoint is called (so that users are aware that the new param will be deprecated);
 - Livechat rooms should not be returned by the `engagement-dashboard/channels/list` endpoint anymore.

## Further comments

[SUP-513](https://rocketchat.atlassian.net/browse/SUP-513)
[SUP-558](https://rocketchat.atlassian.net/browse/SUP-588)

**Aggregation explain (before):** [explain-aggregation-before.json](https://github.com/RocketChat/Rocket.Chat/files/15488767/explain-aggregation-before.json)

**Aggregation explain (after):** [explain-aggregation-after.json](https://github.com/RocketChat/Rocket.Chat/files/15488769/explain-aggregation-after.json)

**Aggregation explain (after and using hideRoomsWithNoActivity param):** [aggregation-explain-after-hide-rooms-with-no-activity.json](https://github.com/user-attachments/files/15876417/aggregation-explain-after-hide-rooms-with-no-activity.json)

[SUP-513]: https://rocketchat.atlassian.net/browse/SUP-513?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


KevLehman (changes requested):

KevLehman (commented):
༼ つ ◕_◕ ༽つ API tests

apps/meteor/ee/server/lib/engagementDashboard/channels.ts
```diff
@@ -43,15 +43,7 @@ export const findAllChannelsWithNumberOfMessages = async ({
...
-	const total =
-		(
-			await Rooms.countChannelsWithNumberOfMessagesBetweenDate({
-				start: convertDateToInt(start),
-				end: convertDateToInt(end),
-				startOfLastWeek: convertDateToInt(startOfLastWeek),
-				endOfLastWeek: convertDateToInt(endOfLastWeek),
-			}).toArray()
-		)[0]?.total ?? 0;
+	const total = await Rooms.countDocuments();
```
KevLehman, line 46:
```suggestion
	const total = await Rooms.countDocuments({});
```

apps/meteor/server/models/raw/Rooms.ts
```diff
@@ -531,22 +532,6 @@ export class RoomsRaw extends BaseRaw<IRoom> implements IRoomsModel {
...
-	}): AggregationCursor<{ total: number }> {
-		const aggregationParams = this.getChannelsWithNumberOfMessagesBetweenDateQuery(params);
-		aggregationParams.push({ $count: 'total' });
-
-		return this.col.aggregate<{ total: number }>(aggregationParams, {
-			allowDiskUse: true,
-			readPreference: readSecondaryPreferred(),
-		});
-	}
-
```
debdutdeb, line 549: adding to my previous comment, that is exactly what this method was doing, while the duplication wasn't good, adding $count there to existing makes sense?

geekgonecrazy:
@matheusbsilva137 since adjusting the aggregation as part of improving can you please attach the mongo explain of the query before the optimization and after?

ideally with some real amount of data being processed so we can get an idea if the query is going to utilize appropriate indexes and be any better.

apps/meteor/ee/server/lib/engagementDashboard/channels.ts
```diff
@@ -43,15 +43,7 @@ export const findAllChannelsWithNumberOfMessages = async ({
...
-	const total =
-		(
-			await Rooms.countChannelsWithNumberOfMessagesBetweenDate({
-				start: convertDateToInt(start),
-				end: convertDateToInt(end),
-				startOfLastWeek: convertDateToInt(startOfLastWeek),
-				endOfLastWeek: convertDateToInt(endOfLastWeek),
-			}).toArray()
-		)[0]?.total ?? 0;
+	const total = await Rooms.countDocuments({});
```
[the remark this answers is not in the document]
matheusbsilva137: > why not just add $count to existing aggregation, and take from there?

I could do that with a `$facet` in the existing aggregation, but since this `$limit` may be performed before the result is complete (in the "middle" of it), we'd carry an object like this starting on that point:

```
rooms: [ { room: {}, messages: 5, lastWeekMessages: []}, ...]
totalCount: 25
```
This would make us need to apply **an extra unwind** to get the data inside the (`rooms`) array, I'm not sure it's worth it.

matheusbsilva137 (commented):
I've just attached the old and new aggregations explains to the PR description.
The new aggregation uses the `t` index for rooms, while the old one doesn't use indexes at all (based on what Compass told me :nerd_face: )

cc @MarcosSpessatto @geekgonecrazy

KevLehman:
any possibility of trying the aggregation with more rooms? 🤔

geekgonecrazy:
Thanks for adding :pray: this sort of data is important.  Might be worth just creating a bunch of records.  We should make it easier to fake a lot of data for optimizations like this.  Having to wait on larger environments seems like adding in delays that we could avoid

apps/meteor/ee/server/lib/engagementDashboard/channels.ts
```diff
@@ -34,24 +88,18 @@ export const findAllChannelsWithNumberOfMessages = async ({
...
-	const total =
-		(
-			await Rooms.countChannelsWithNumberOfMessagesBetweenDate({
-				start: convertDateToInt(start),
-				end: convertDateToInt(end),
-				startOfLastWeek: convertDateToInt(startOfLastWeek),
-				endOfLastWeek: convertDateToInt(endOfLastWeek),
-			}).toArray()
-		)[0]?.total ?? 0;
+	const total = await Rooms.countDocuments({ t: { $in: roomTypes } });
```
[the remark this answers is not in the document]
matheusbsilva137: This method returns all rooms (with valid types) regardless of the date range provided (the range is only used to count messages, not for filtering rooms)

apps/meteor/ee/server/api/engagementDashboard/channels.ts
```diff
@@ -45,17 +45,19 @@ API.v1.addRoute(
...
 
-			const { start, end } = this.queryParams;
+			const { start, end, hideRoomsWithNoActivity } = this.queryParams;
 			const { offset, count } = await getPaginationItems(this.queryParams);
 
-			const { channels, total } = await findAllChannelsWithNumberOfMessages({
+			const { channels, total } = await findChannelsWithNumberOfMessages({
 				start: mapDateForAPI(start),
 				end: mapDateForAPI(end),
+				hideRoomsWithNoActivity: hideRoomsWithNoActivity === 'true' ?? false,
```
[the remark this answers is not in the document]
matheusbsilva137: Oops, fixed
Thank you!

apps/meteor/tests/end-to-end/api/34-engagement-dashboard.ts
```diff
@@ -0,0 +1,347 @@
...
+describe('[Engagement Dashboard]', function () {
+	this.retries(0);
+
+	const isEnterprise = Boolean(process.env.IS_EE);
+
+	before((done) => getCredentials(done));
+
+	before(() => updatePermission('view-engagement-dashboard', ['admin']));
+
+	before(() => updatePermission('view-engagement-dashboard', ['admin']));
```
MarcosSpessatto, line 20:
```suggestion
	after(() => updatePermission('view-engagement-dashboard', ['admin']));
```

MarcosSpessatto (approved):

apps/meteor/tests/end-to-end/api/34-engagement-dashboard.ts
matheusbsilva137: Oops :man_facepalming: thank you!

MarcosSpessatto (dismissed):

apps/meteor/server/models/raw/Analytics.ts
```diff
@@ -211,4 +211,117 @@ export class AnalyticsRaw extends BaseRaw<IAnalytic> implements IAnalyticsModel
...
+		startOfLastWeek: number;
+		endOfLastWeek: number;
+		options?: any;
+	}) {
+		const typeAndDateMatch = {
+			$match: {
+				'type': 'messages',
+				'room.t': { $in: types },
+				'date': { $gte: startOfLastWeek, $lte: end },
+			},
```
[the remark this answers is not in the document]
matheusbsilva137: We already have an index on the `date` field for this collection :partying_face: We're using it on this query
Is your suggestion to add a compound index for these 3 fields? I'll try and check if we have any improvement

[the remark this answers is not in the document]
KevLehman: Yeah, I'd suggest trying with some indexes and add the one that adds the best result with a `partialFilterExpresssion` on messages.

Maybe something like
```
room.t: 1, date: 1, partialFilterExpression: { type: 'messages'
```

and see if it gets better or not. 

Try with hints as well, as mongo would sometimes pick the index with the most data output so no docs are left out even if a better index is on the collection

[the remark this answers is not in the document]
matheusbsilva137: Done! Average time is much better now with the new index
We dont't need to have the `partialFilterExpression` here. `sparse` works just the same since only documents with `type: 'messages'` will have both fields.

matheusbsilva137:
While improving the indexes for this PR, I found out we had an index for `room._id` and `date` for the **Analytics** collection that was actually not indexing anything :cry: That's because we applied an incorrect `partialFilterExpression` to it 
Since we committed not to add migrations in minor releases, I decided to fix it and assign a name to the index (so that it is recreated and doesn't throw a duplicated index name error). We can remove the old/incorrect index in the next major

apps/meteor/server/models/raw/Analytics.ts
```diff
@@ -14,7 +14,11 @@ export class AnalyticsRaw extends BaseRaw<IAnalytic> implements IAnalyticsModel
 	}
 
 	protected modelIndexes(): IndexDescription[] {
-		return [{ key: { date: 1 } }, { key: { 'room._id': 1, 'date': 1 }, unique: true, partialFilterExpression: { type: 'rooms' } }];
+		return [
+			{ key: { date: 1 } },
+			{ key: { 'room._id': 1, 'date': 1 }, sparse: true, name: 'room-id-date-index' },
```
matheusbsilva137, line 19: While improving the indexes for this PR, I found out we had an index for `room._id` and `date` for the **Analytics** collection that was actually not indexing anything 😢 That's because we applied an incorrect `partialFilterExpression` to it (`type: 'rooms'` won't find any matching document, I guess we intended to have `type: 'messages'`)
Since we committed not to add migrations in minor releases, I decided to fix it and assign a name to the index (so that it is recreated and doesn't throw a duplicated index name error). We can remove the old/incorrect index in the next major

apps/meteor/server/models/raw/Analytics.ts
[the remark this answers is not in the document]
KevLehman: Nice!
About partial vs sparse: https://www.mongodb.com/docs/manual/core/index-partial/#comparison-with-sparse-indexes

apps/meteor/server/models/raw/Analytics.ts
KevLehman: We can put this on a separate PR

KevLehman: Question: is the index actually not used for any query? If so, I'd say we can leave it like that for now and do a migration to remove it targetting 7.0

And creating the new index alongside it

matheusbsilva137: Yes, it's not used anywhere since it has an incorrect `partialFilterExpression` that won't match any document. But our intention was to use it on `saveMessageSent`, which is part of the flow of this engagement dashboard tab, it inserts the data (that's why I thought of including this here since it's also a performance improvement)

apps/meteor/server/models/raw/Analytics.ts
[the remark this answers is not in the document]
matheusbsilva137: Ohhh, thanks. I'm now using a `partialFilterExpressions` instead
Their indexes are both the exact same size with my test data, but let's follow mongodb docs :pray:

apps/meteor/server/models/raw/Analytics.ts
matheusbsilva137: I've just removed this index improvement from this PR. I'll create another task for this
Please tell me in case you think this should have stayed here

.changeset/many-tables-love.md
```diff
@@ -0,0 +1,6 @@
+---
```
KevLehman, line 1: Curious: does this deserve its own changeset? 

I know excluding them ended up being part of the perf gains but not sure if we should have a separate entry for that, but i may be wrong

wdyt @MarcosSpessatto ?

matheusbsilva137: We have a customer asking for this specific change, that's why I decided to include this here
That's described in SUP-558

dougfabris (dismissed):
Approving it on behalf of front-end team

MarcosSpessatto (dismissed):

KevLehman (approved):

dougfabris (dismissed):

KevLehman (dismissed):

apps/meteor/ee/server/api/engagementDashboard/channels.ts
```diff
@@ -45,17 +46,28 @@ API.v1.addRoute(
...
 					count: Match.Maybe(String),
 				}),
 			);
 
-			const { start, end } = this.queryParams;
+			const { start, end, hideRoomsWithNoActivity } = this.queryParams;
 			const { offset, count } = await getPaginationItems(this.queryParams);
 
-			const { channels, total } = await findAllChannelsWithNumberOfMessages({
+			apiDeprecationLogger.deprecatedParameterUsage(
```
ggazzo, line 58: Isn't something missing here  like `hideRoomsWithNoActivity != undefined`?

matheusbsilva137: I had two different deprecation messages based on `hideRoomsWithNoActivity` being undefined or not, but we decided to merge them into one per [this request](https://github.com/RocketChat/Rocket.Chat/pull/32493#discussion_r1638648260)
The idea here is that even users who don't send the `hideRoomsWithNoActivity` param need to be warned that this endpoint's results will be changed in the next major (it will behave as `hideRoomsWithNoActivity=true` by default)

matheusbsilva137: Got it, so I'll only call this when the deprecated param is defined, then. However, even users who don't use this param will perceive a breaking change in 7.0

KevLehman (approved):

apps/meteor/ee/server/api/engagementDashboard/channels.ts
matheusbsilva137: Hey @ggazzo
I had a brief discussion with @sampaiodiego about this deprecation message and I believe we agreed that even users who don't use the new param should be warned that this endpoint will be broken on 7.0 (we considered the metrics collected from this function, but this is a relevant message). That's because not only the new param will be removed, but the "default" set of results will also be changed on 7.0. I decided to remove the condition to display the deprecation message so that all users are aware of the incoming change when they use the endpoint
I also considered one more option: not removing the new param at all, but the current behavior can be considered destructive -- so I believe that's something we wouldn't like to keep.

Please tell me what you think.

KevLehman (dismissed):

sampaiodiego (dismissed):

apps/meteor/ee/server/api/engagementDashboard/channels.ts
ggazzo: I think no one uses this endpoint so whatever, but it is something non-standard, and does not apply to any other special case, in other scenarios this would flood the logs and create more harm than good. I wouldn't like to see this being done elsewhere. let keep this in mind.

ggazzo: Maybe warn if the parameter is NOT used (since our UI will use it by default)

sampaiodiego (approved):

ggazzo (approved):

PR #32500, opened 2024-05-24
fix: User status when setting "Use REST instead of websocket for Meteor calls" is disabled

sampaiodiego:


apps/meteor/ee/server/startup/presence.ts
```diff
@@ -30,7 +30,7 @@ Meteor.startup(() => {
 	});
 
 	Accounts.onLogin((login: any): void => {
-		if (login.type !== 'resume' || !login.connection.id) {
+		if (!login.connection.id) {
```
debdutdeb, line 33: wasn;t this added to actually fix another issue? 👀

ggazzo (approved):

Issue #74, opened 2015-05-31
Fully test on Mobile and fix bugs


engelgabriel:

riaan53:
Hi,

Will this include cordova? Will you also support offline mode (which is a important for cordova) with ground db? 

I have tested with iOS cordova and it doesnt look bad, with a bit of tweaks here and there you can have full iOS and Android support - which will be fantastic :)

Regards,
Riaan

rodrigok:
I think that we can create a cordova version first and then start native versions.

marceloschmidt:
Can we close this?

engelgabriel:
yes. to generic.

PR #32501, opened 2024-05-25
refactor: EmailInbox out of DB Watcher

ricardogarim:
As per the updates mentioned in [PROJ-7], [SCA-11] and [ADR #74], **this pull request focuses on relocating `EmailInbox` model out of DB Watcher service.**

### Context

This modification enhances RocketChat's app by allowing it to directly call listeners through the `api.broadcast` global function, bypassing the need for MongoDB Change Stream data propagation.

This change offers better control over user notifications via Web Sockets, enabling more precise management of use-cases. Instead of notifying every database change, we can now send an `api.broadcast` call only when necessary, reducing overall network messages. Additionally, this contributes to the future removal of the DB Watcher deployment, optimizing resource utilization.

## Proposed changes

Key changes include:

- Conditionally incorporating **EmailInbox** entity import within DB watchers on application startup based on the `dbWatchersDisabled` flag.
- Enabling support for the following use cases to directly trigger `watch.emailInbox` listener event, subject to the `dbWatchersDisabled` flag.

<summary>Updated Use Cases</summary>

| Use Case | Route/Trigger | Notes |
|----------|---------------|-------|
| Insert or update email inbox | POST `/api/v1/email-inbox` |  |
| Delete email inbox | DELETE `/api/v1/email-inbox/:_id` |  |
| Reconnect to email inbox on IMAPInterceptor | IMAPInterceptor reconnect > selfDisable functions |  |

## Steps to test or reproduce

1. Start RocketChat's application with the `DISABLE_DB_WATCHERS` environment variable set to `true`.
2. Perform HTTP requests or simulate use cases listed in the above table.

## Further comments

To maintain consistency and avoid potential regressions, event names and signatures have been kept unchanged on both the client and app sides. This decision streamlines efforts and mitigates the risk of unintended consequences.

Additionally, since the EmailInbox model was empty, some new model functions have been created to accommodate specific needs when dealing with database operations.

[PROJ-7]: https://rocketchat.atlassian.net/browse/PROJ-7

[SCA-11]: https://rocketchat.atlassian.net/browse/SCA-11

[ADR #74]: https://adr.rocket.chat/0074


apps/meteor/server/models/raw/EmailInbox.ts
```diff
@@ -12,4 +12,32 @@ export class EmailInboxRaw extends BaseRaw<IEmailInbox> implements IEmailInboxMo
 	protected modelIndexes(): IndexDescription[] {
 		return [{ key: { email: 1 }, unique: true }];
 	}
+
+	async setDisabledById(id: IEmailInbox['_id']): Promise<ModifyResult<IEmailInbox>> {
+		return this.findOneAndUpdate({ _id: id, active: true }, { $set: { active: false } }, { returnDocument: 'after' });
+	}
+
+	async removeById(id: IEmailInbox['_id']): Promise<DeleteResult> {
```
sampaiodiego, line 20: `removeById` already exists on `BaseRaw`.. you can just remove this one..

https://github.com/RocketChat/Rocket.Chat/blob/de1960f4a0ce21e13919991023c88054afced4fd/apps/meteor/server/models/raw/BaseRaw.ts#L266

packages/model-typings/src/models/IEmailInboxModel.ts
sampaiodiego, line 1: that's really great that you create all this methods here.. thx <3

apps/meteor/server/models/raw/EmailInbox.ts
```diff
@@ -12,4 +12,32 @@ export class EmailInboxRaw extends BaseRaw<IEmailInbox> implements IEmailInboxMo
...
+
+	async create(emailInbox: IEmailInbox): Promise<InsertOneResult<IEmailInbox>> {
+		return this.insertOne(emailInbox);
+	}
+
+	async updateById(id: IEmailInbox['_id'], data: UpdateFilter<IEmailInbox>): Promise<ModifyResult<IEmailInbox>> {
+		return this.findOneAndUpdate({ _id: id }, data, { returnDocument: 'after', projection: { _id: 1 } });
+	}
+
+	findActiveEmailInboxes(): FindCursor<IEmailInbox> {
```
sampaiodiego, line 32: I think adding `EmailInboxes` is redundant.. you also didn't add the suffix to the other methods as well, so I guess we can get rid of it here..

```suggestion
	findActive(): FindCursor<IEmailInbox> {
```

apps/meteor/app/api/server/lib/emailInbox.ts
sampaiodiego, line 1: this file should be converted to it's own service instead of "standalone" functions.. can we track that in another task pls?

sampaiodiego (changes requested):

apps/meteor/app/api/server/v1/email-inbox.ts
```diff
@@ -97,11 +108,12 @@ API.v1.addRoute(
 				throw new Error('error-invalid-param');
 			}
 
-			const emailInboxes = await EmailInbox.findOneById(_id);
-			if (!emailInboxes) {
+			const { deletedCount } = await removeEmailInbox(_id);
+
+			if (!deletedCount) {
```
debdutdeb, line 113: <3 minus the redundant findOne

apps/meteor/server/models/raw/EmailInbox.ts
debdutdeb, line 1: should findoptions be parameters? generally they should, so people can pass especially whatever subset of projections they need, guess the real question is if that should be done in this PR or not. projections here seem to be specific to the usecases in layers above, and later on might need to touch these files again to add new ones.

apps/meteor/server/models/raw/EmailInbox.ts
```diff
@@ -12,4 +12,28 @@ export class EmailInboxRaw extends BaseRaw<IEmailInbox> implements IEmailInboxMo
...
+		return this.findOneAndUpdate({ _id: id, active: true }, { $set: { active: false } }, { returnDocument: 'after' });
+	}
+
+	async create(emailInbox: IEmailInbox): Promise<InsertOneResult<IEmailInbox>> {
+		return this.insertOne(emailInbox);
+	}
+
+	async updateById(id: IEmailInbox['_id'], data: UpdateFilter<IEmailInbox>): Promise<ModifyResult<IEmailInbox>> {
+		return this.findOneAndUpdate({ _id: id }, data, { returnDocument: 'after', projection: { _id: 1 } });
+	}
```
debdutdeb, line 26: typing is wrong, right?
```suggestion
	async updateById(id: IEmailInbox['_id'], data: UpdateFilter<IEmailInbox>): Promise<IEmailInbox['_id']> {
		return this.findOneAndUpdate({ _id: id }, data, { returnDocument: 'after', projection: { _id: 1 } });
	}
```

sampaiodiego (approved):

PR #32522, opened 2024-05-29
fix: prevent broadcasting of system messages when hide setting is enabled

AllanPazRibeiro:

This change ensures that system messages are not broadcasted when the setting to hide system messages is enabled. By checking the Hide_System_Messages configuration, the system will prevent unnecessary message broadcasts, improving message handling efficiency. This fix guarantees adherence to user-defined visibility settings for system messages.

## Issue(s)

[SUP-559](https://rocketchat.atlassian.net/browse/SUP-559)

## Steps to test or reproduce

1. Set the Hide_System_Messages setting to include specific message types.
2. Send a message of one of these types.
3. Verify that the message is not broadcasted.

## Further comments

This update will enhance the efficiency of message handling by preventing the broadcast of system messages when configured to do so. This helps in maintaining cleaner logs and better performance in scenarios where certain system messages are not needed.

[SUP-559]: https://rocketchat.atlassian.net/browse/SUP-559?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.changeset/grumpy-games-greet.md
```diff
@@ -0,0 +1,4 @@
+---
+'@rocket.chat/meteor': patch
+---
+This change ensures that system messages are not broadcasted when the specific setting to hide system messages is enabled.
```
KevLehman, line 4:
```suggestion
Changed streaming logic to prevent hidden system messages from being broadcasted through `stream-room-messages`.
```

apps/meteor/server/modules/watchers/lib/messages.ts
```diff
@@ -19,6 +19,19 @@ export async function getMessageToBroadcast({ id, data }: { id: IMessage['_id'];
 		return;
 	}
 
+	if (message.t) {
+		const hideSystemMessage = (await getSettingCached('Hide_System_Messages')) as MessageTypesValues[];
+
+		if (hideSystemMessage) {
+			const isMutedUnmuted = (messageType: string): boolean => {
+				return messageType === 'user-muted' || messageType === 'user-unmuted';
+			};
```
KevLehman, line 28: Move this to outside of the func, no need to recreate it everytime `getMessageToBroadcast` is called

apps/meteor/tests/unit/server/modules/watchers/lib/messages.spec.ts
```diff
@@ -0,0 +1,127 @@
...
+
+	it('should return undefined if message is hidden or imported', async () => {
+		messagesFindOneStub.resolves({ ...sampleMessage, _hidden: true });
+
+		const result = await getMessageToBroadcast({ id: '123' });
+
+		expect(result).to.be.undefined;
+	});
+
+	it('should hide message if type is in hideSystemMessage settings', async () => {
```
KevLehman: add cases for other types of system messages.

apps/meteor/server/modules/watchers/lib/messages.ts
```diff
@@ -19,6 +19,19 @@ export async function getMessageToBroadcast({ id, data }: { id: IMessage['_id'];
...
+			if (hideSystemMessage.includes(message.t) || (isMutedUnmuted(message.t) && hideSystemMessage.includes('mute_unmute'))) {
```
KevLehman, line 29: Curious, is this the only case where the setting value is different from the actual message type?

I refer to the message type being `user-muted` but the hide setting having `mute_unmute` 👀

KevLehman (changes requested):

apps/meteor/server/modules/watchers/lib/messages.ts
```diff
@@ -13,12 +13,26 @@ const getUserNameCached = mem(
 	{ maxAge: 10000 },
 );
 
+const isMutedUnmuted = (messageType: string): boolean => {
```
KevLehman, line 16: I see there's a `getHiddenSystemMessages` function with some similar logic. I wonder if we could unify both 🤔 

Like creating a `lib` that did this `mute_unmute` parsing on a single place (without importing the `settings.watch` from there)

wdyt?

apps/meteor/app/lib/server/lib/getHiddenSystemMessages.ts
```diff
@@ -1,26 +1,8 @@
...
-	hideMessagesOfTypeServer.clear();
-
-	hiddenTypes.forEach((item) => hideMessagesOfTypeServer.add(item));
-});
-
-// TODO probably remove on chained event system
-export function getHiddenSystemMessages(room: IRoom): MessageTypesValues[] {
-	return Array.isArray(room?.sysMes) ? room.sysMes : [...hideMessagesOfTypeServer];
-}
+export const getHiddenSystemMessages = async (room: IRoom): Promise<MessageTypesValues[]> => {
```
KevLehman, line 5: it's better to pass the setting value as a parameter, so the other places can use the normal settings.get while this one uses the call to the model

apps/meteor/server/lib/systemMessage/hideSystemMessage.ts
```diff
@@ -0,0 +1,23 @@
...
+	return messageType === 'user-muted' || messageType === 'user-unmuted';
+};
+
+const getSettingCached = mem(async (setting: string): Promise<SettingValue> => Settings.getValueById(setting), { maxAge: 10000 });
+
+export const getCachedHiddenSystemMessage = async () => {
+	return getSettingCached('Hide_System_Messages') as unknown as MessageTypesValues[];
+};
+
+export const shouldHideSystemMessage = async (messageType: MessageTypesValues): Promise<boolean> => {
```
KevLehman, line 15: receive hidden system messages setting as param

apps/meteor/tests/unit/server/modules/watchers/lib/messages.spec.ts
```diff
@@ -0,0 +1,224 @@
...
+			{
+				description: 'should hide message if type is new-owner',
+				message: { ...sampleMessage, t: 'new-owner' },
+				hideSystemMessages: ['new-owner'],
+				useRealName: false,
+				expectedResult: undefined,
+			},
+		];
+
+		testCases.forEach(({ description, message, hideSystemMessages, useRealName, expectedResult }) => {
```
KevLehman, line 164: iirc there's a test.each or it.each that does this exact thing.

If it's not there on mocha/chai, pls ignore this one

KevLehman (changes requested):

apps/meteor/app/lib/server/methods/getChannelHistory.ts
```diff
@@ -67,7 +67,9 @@ Meteor.methods<ServerMethods>({
 			throw new Meteor.Error('error-invalid-date', 'Invalid date', { method: 'getChannelHistory' });
 		}
 
-		const hiddenMessageTypes = getHiddenSystemMessages(room);
+		const hiddenSystemMessages = (await Settings.getValueById('Hide_System_Messages')) as MessageTypesValues[];
```
KevLehman, line 70: Use `settings.get`

I'm referring to the cached settings, not the settings model

apps/meteor/server/lib/systemMessage/hideSystemMessage.ts
```diff
@@ -0,0 +1,21 @@
...
+
+export const isMutedUnmuted = (messageType: string): boolean => {
+	return messageType === 'user-muted' || messageType === 'user-unmuted';
+};
+
+const getSettingCached = mem(async (setting: string): Promise<SettingValue> => Settings.getValueById(setting), { maxAge: 10000 });
+
+export const getCachedHiddenSystemMessage = async () => {
+	return getSettingCached('Hide_System_Messages') as unknown as MessageTypesValues[];
+};
```
KevLehman, line 13: Do you use these for something? on `lib/messages` L25 you could use these but you're re-creating a mem there.

If no use for these, remove

apps/meteor/app/lib/server/lib/getMemSettings.ts
```diff
@@ -0,0 +1,4 @@
+import { Settings } from '@rocket.chat/models';
+import mem from 'mem';
+
+export const getSettingCached = mem(Settings.getValueById, { maxAge: 10000 });
```
KevLehman, line 4: needed?

apps/meteor/app/lib/server/functions/loadMessageHistory.ts
```diff
@@ -29,7 +30,9 @@ export async function loadMessageHistory({
 		throw new Error('error-invalid-room');
 	}
 
-	const hiddenMessageTypes = getHiddenSystemMessages(room);
+	const hiddenSystemMessages = (await getSettingCached('Hide_System_Messages')) as MessageTypesValues[];
+
+	const hiddenMessageTypes = getHiddenSystemMessages(room, hiddenSystemMessages);
```
KevLehman, line 35: This doesn't work like the original version

First version when calling `getHiddenSystemMessages` and the room didn't have a sysMes property, it returned the list of hidden system messages, but already parsing the `mute_unmute` one to `user-muted` and `user-unmuted`. 

This one is just returning `mute-unmute`

can you take a look?

KevLehman (changes requested):

apps/meteor/app/lib/server/functions/loadMessageHistory.ts
```diff
@@ -29,7 +30,9 @@ export async function loadMessageHistory({
 		throw new Error('error-invalid-room');
 	}
 
-	const hiddenMessageTypes = getHiddenSystemMessages(room);
+	const hiddenSystemMessages = (await getSettingCached('Hide_System_Messages')) as MessageTypesValues[];
```
KevLehman, line 33: Use the normal settings.get

On almost anything that's not used by microservices we can use settings.get (from CachedSettings)

apps/meteor/app/lib/server/lib/getMemSettings.ts
KevLehman: Yeah. But since you can remove the usage of this on loadMessageHistory, you can remove the file and keep the declaration on messages as it was

apps/meteor/app/lib/server/methods/getChannelHistory.ts
```diff
@@ -67,7 +68,9 @@ Meteor.methods<ServerMethods>({
 			throw new Meteor.Error('error-invalid-date', 'Invalid date', { method: 'getChannelHistory' });
 		}
 
-		const hiddenMessageTypes = getHiddenSystemMessages(room);
+		const hiddenSystemMessages = settings.get('Hide_System_Messages') as MessageTypesValues[];
```
KevLehman, line 71: Curious: can you pass the type on the type parameter of settings.get? Otherwise just ignore this comment

KevLehman (changes requested):

KevLehman (approved):

PR #32523, opened 2024-05-29
[CHORE] increase allowed stale time

geekgonecrazy:
14 days before marked stale and 14 days in stale before closed.  Giving double the time before issues are closed out fully.


sampaiodiego (approved):
it's been a while since I've seen this working 🤔

PR #32525, opened 2024-05-29
chore: Deprecate `banners.getNew` endpoint

matheusbsilva137:

 - Deprecate `banners.getNew` endpoint (which wasn't properly deprecated even though we had a comment in the code)

## Further comments

[CORE-472](https://rocketchat.atlassian.net/browse/CORE-472)

Requested in #32509

[CORE-472]: https://rocketchat.atlassian.net/browse/CORE-472?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/meteor/app/api/server/v1/banners.ts
```diff
@@ -52,9 +52,8 @@ import { API } from '../api';
  */
 API.v1.addRoute(
 	'banners.getNew',
-	{ authRequired: true },
```
KevLehman, line 55: Just curious, is that the the real alternative? I see the endpoint returning one or all banners 👀 but the new one apparently just returns one.

If that's the case, we should put the endpoint to return more than one as an alternative

matheusbsilva137: Oops, you're right. bid can be undefined
I'll add `banners` as an alternative too

KevLehman (approved):

MarcosSpessatto (approved):

PR #32527, opened 2024-05-30
fix: `Accounts_LoginExpiration` being used differently on codebase

KevLehman:
## Issue(s)

https://rocketchat.atlassian.net/browse/SUP-574
## Further comments

Fixed an inconsistent evaluation of the `Accounts_LoginExpiration` setting over the codebase. In some places, it was being used as milliseconds while in others as days. Invalid values produced different results. A helper function was created to centralize the setting validation and the proper value being returned to avoid edge cases.
Negative values may be saved on the settings UI panel but the code will interpret any negative, NaN or 0 value to the default expiration which is 90 days.


packages/tools/src/getLoginExpiration.ts
```diff
@@ -0,0 +1,10 @@
+const ACCOUNTS_DEFAULT_LOGIN_EXPIRATION = 90;
+
+// Given a value, validates if it mets the conditions to be a valid login expiration.
+// Else, returns the default login expiration (which for Meteor is 90 days)
+export const getLoginExpiration = (expiry?: number) => {
+	if (expiry && typeof expiry === 'number' && !Number.isNaN(expiry) && expiry > 0) {
+		return expiry;
+	}
+	return ACCOUNTS_DEFAULT_LOGIN_EXPIRATION;
+};
```
debdutdeb, line 10: may I suggest renaming to getLoginExpiration**InDays**?
```suggestion
const ACCOUNTS_DEFAULT_LOGIN_EXPIRATION_IN_DAYS = 90;

// Given a value, validates if it mets the conditions to be a valid login expiration.
// Else, returns the default login expiration (which for Meteor is 90 days)
export const getLoginExpirationInDays = (expiry?: number) => {
	if (expiry && typeof expiry === 'number' && !Number.isNaN(expiry) && expiry > 0) {
		return expiry;
	}
	return ACCOUNTS_DEFAULT_LOGIN_EXPIRATION_IN_DAYS;
};
```

packages/tools/src/getLoginExpiration.ts
```diff
@@ -0,0 +1,10 @@
+const ACCOUNTS_DEFAULT_LOGIN_EXPIRATION = 90;
+
+// Given a value, validates if it mets the conditions to be a valid login expiration.
+// Else, returns the default login expiration (which for Meteor is 90 days)
+export const getLoginExpiration = (expiry?: number) => {
```
MarcosSpessatto, line 5: It would be nice to have the unit (days) in the function name, isn't it? WDYT?

packages/tools/src/getLoginExpiration.spec.ts
```diff
@@ -0,0 +1,26 @@
...
+		expect(getLoginExpiration(-1)).toBe(90);
+	});
+	it('should return 90 when value is undefined', () => {
+		expect(getLoginExpiration(undefined)).toBe(90);
+	});
+	it('should return 90 when value is not a number', () => {
+		// @ts-expect-error - Testing
+		expect(getLoginExpiration('90')).toBe(90);
+	});
+	it('should return the value passed when its valid', () => {
```
MarcosSpessatto, line 23: Do we need to add a validation for non-integer values?

apps/meteor/app/api/server/v1/users.ts
```diff
@@ -1023,8 +1024,10 @@ API.v1.addRoute(
 
 			const token = me.services?.resume?.loginTokens?.find((token) => token.hashedToken === hashedToken);
 
+			const loginExp = settings.get<number>('Accounts_LoginExpiration');
+
 			const tokenExpires =
-				(token && 'when' in token && new Date(token.when.getTime() + settings.get<number>('Accounts_LoginExpiration') * 1000)) || undefined;
+				(token && 'when' in token && new Date(token.when.getTime() + getLoginExpiration(loginExp) * 60 * 60 * 24 * 1000)) || undefined;
```
MarcosSpessatto, line 1030: Maybe create a helper function (inside the same package) to do this conversion? WDYT?

MarcosSpessatto (changes requested):

packages/tools/src/getLoginExpiration.spec.ts
KevLehman: 2 lines above, there's an string test 👀

packages/tools/src/getLoginExpiration.spec.ts
```diff
@@ -0,0 +1,35 @@
+import { getLoginExpirationInDays, getLoginExpirationInMs } from './getLoginExpiration';
+
+describe('getLoginExpiration', () => {
```
MarcosSpessatto, line 3:
```suggestion
describe('getLoginExpirationInDays', () => {
```

packages/tools/src/converter.ts
```diff
@@ -0,0 +1,3 @@
+export const convertFromDaysToMilliseconds = (days: number) => {
```
MarcosSpessatto, line 1: Can we add a unit test for it? Also, do we need to ensure integer only here? 🤔

MarcosSpessatto (changes requested):

packages/tools/src/converter.ts
```diff
@@ -0,0 +1,7 @@
...
+	if (typeof days !== 'number') {
```
MarcosSpessatto, line 2: Should we round it to the closest integer? What if someone provides a decimal here?

MarcosSpessatto (changes requested):

packages/tools/src/converter.ts
KevLehman: 3.1 days still has some equivalence on millis innit? :kek:

Good question, i can take a look, maybe putting an `IsInteger` would suffice, wdyt?

MarcosSpessatto (approved):

PR #32532, opened 2024-05-31
chore!: Remove `meteor/check` from `chat` endpoints

matheusbsilva137:

 - Remove `meteor/check` from `chat` endpoints by using the `validateParams` property.

## Issue(s)

[CORE-316](https://rocketchat.atlassian.net/browse/CORE-316)

## Further comments

**The intention for this PR is to improve code readability and make invalid params errors more standardized. No side effects are expected. We're considering this a breaking change since invalid params errors are being changed to follow the same standard when possible.**

[CORE-316]: https://rocketchat.atlassian.net/browse/CORE-316?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/rest-typings/src/v1/chat.ts
+++ b/packages/rest-typings/src/v1/chat.ts
@@ -255,8 +255,9 @@ export const isChatReportMessageProps = ajv.compile<ChatReportMessage>(ChatRepor
 
 type ChatGetThreadsList = PaginatedRequest<{
 	rid: IRoom['_id'];
-	type: 'unread' | 'following' | 'all';
+	type?: 'unread' | 'following' | 'all';
 	text?: string;
+	fields?: string;
 }>;
 
 const ChatGetThreadsListSchema = {
@@ -267,6 +268,7 @@ const ChatGetThreadsListSchema = {
 		},
 		type: {
 			type: 'string',
+			enum: ['following', 'unread', 'all'],
 			nullable: true,
 		},
 		text: {
@@ -279,6 +281,18 @@ const ChatGetThreadsListSchema = {
 		},
 		count: {
 			type: 'number',
+			nullable: true,
+		},
+		sort: {
+			type: 'string',
+			nullable: true,
+		},
+		query: {
+			type: 'string',
+			nullable: true,
+		},
+		fields: {
+			type: 'string',
 			nullable: true,
 		},
 	},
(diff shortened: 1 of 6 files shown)


apps/meteor/tests/end-to-end/api/05-chat.js
```diff
@@ -1341,6 +1341,88 @@ describe('[Chat]', function () {
...
+				})
+				.expect('Content-Type', 'application/json')
+				.expect(400)
+				.expect((res) => {
+					expect(res.body).to.have.property('success', false);
+					expect(res.body).to.have.property('errorType', 'invalid-params');
+				});
+		});
+
+		it('should fail updating a message if no room id is provided', () => {
```
MarcosSpessatto, line 1376: This is duplicated with the first one added in this file...

packages/rest-typings/src/v1/chat.ts
```diff
@@ -255,8 +255,9 @@ export const isChatReportMessageProps = ajv.compile<ChatReportMessage>(ChatRepor
 
 type ChatGetThreadsList = PaginatedRequest<{
 	rid: IRoom['_id'];
-	type: 'unread' | 'following' | 'all';
+	type?: 'unread' | 'following' | 'all';
```
MarcosSpessatto, line 258: I didn't find any mention of `all` in the code... Is this expected?

packages/rest-typings/src/v1/chat.ts
```diff
@@ -267,6 +268,7 @@ const ChatGetThreadsListSchema = {
 		},
 		type: {
 			type: 'string',
+			enum: ['following', 'unread', 'all'],
```
MarcosSpessatto, line 271: Same as before..

MarcosSpessatto (changes requested):

packages/rest-typings/src/v1/chat.ts
matheusbsilva137: all is not handled specifically because it works just as when type is not sent (no extra filter is applied). Do you think we should remove this option?

apps/meteor/tests/end-to-end/api/05-chat.js
matheusbsilva137: Oops, this should test the case where text is not provided
Tks for the heads up!

packages/rest-typings/src/v1/chat.ts
MarcosSpessatto: I think it's better to remove it, to avoid some confusion

MarcosSpessatto (changes requested):

MarcosSpessatto (dismissed):

ggazzo (approved):

ggazzo (approved):

ggazzo (approved):

PR #32540, opened 2024-06-03
refactor: Subscriptions out of DB Watcher

ricardogarim:
As per the updates mentioned in [PROJ-7], [SCA-13] and [ADR #74], **this pull request focuses on relocating `Subscriptions` model out of DB Watcher service.**

### Context

This modification enhances RocketChat's app by allowing it to directly call listeners through the `api.broadcast` global function, bypassing the need for MongoDB Change Stream data propagation.

This change offers better control over user notifications via Web Sockets, enabling more precise management of use-cases. Instead of notifying every database change, we can now send an `api.broadcast` call only when necessary, reducing overall network messages. Additionally, this contributes to the future removal of the DB Watcher deployment, optimizing resource utilization.

## Proposed changes

Key changes include:

- Conditionally incorporating **Subscriptions** entity import within DB watchers on application startup based on the `dbWatchersDisabled` flag.
- Enabling support for the following use cases to directly trigger `watch.subscriptions` listener event, subject to the `dbWatchersDisabled` flag.

<details>
<summary>Updated Use Cases</summary>

| Use Case                                   | Route/Trigger                                 | Notes                                                         |
|--------------------------------------------|----------------------------------------------|---------------------------------------------------------------|
| removeRolesByUserId                        | applyRoomRolesToUser; Roles raw models removeUserRoles  |                                                               |
| addRolesByUserId                           | applyRoomRolesToUser; Roles raw models addUserRoles     |                                                               |
| setAsReadByRoomIdAndUserId                 | post route 'subscriptions.read'; afterSaveMessage callback |                                                               |
| updateAllRoomTypesByRoomId                 | updateRoomType                               |                                                               |
| removeByRoomId                             | deleteRoom server functions (add api route too); LivechatClass.closeRoom; RocketChatRoomAdapter.removeDirectMessageRoom |  |
| incUnreadForRoomIdExcludingUserIds         | updateUsersSubscriptions                     |                                                               |
| setAlertForRoomIdExcludingUserId           | updateUsersSubscriptions                     |                                                               |
| setOpenForRoomIdExcludingUserId            | updateUsersSubscriptions                     |                                                               |
| updateNameAndFnameByRoomId                 | RocketChatRoomAdapter updateRoomName and updateDisplayRoomName |                        |
| setGroupE2EKey                             | method meteor 'e2e.updateGroupKey'; handleSuggestedGroupKey |                       |
| setGroupE2ESuggestedKey                    | method meteor 'e2e.updateGroupKey'           |                                                               |
| unsetGroupE2ESuggestedKey                  | handleSuggestedGroupKey                      |                                                               |
| setOnHoldByRoomId                          | OmnichannelEE > placeRoomOnHold              |                                                               |
| unsetOnHoldByRoomId                        | onCloseLivechat; omnichannelEE.resumeRoomOnHold |                                                          |
| updateUnreadAlertById                      | meteor method saveNotificationSettings       |                                                               |
| updateNotificationsPrefById                | meteor method saveNotificationSettings       |                                                               |
| updateHideMentionStatusById                | meteor method saveNotificationSettings       |                                                               |
| updateDisableNotificationsById             | meteor method saveNotificationSettings       |                                                               |
| clearAudioNotificationValueById            | meteor method saveNotificationSettings saveAudioNotificationValue |                           |
| updateAudioNotificationValueById           | meteor method saveNotificationSettings saveAudioNotificationValue |                           |
| updateMuteGroupMentions                    | meteor method saveNotificationSettings       |                                                               |
| updateHideUnreadStatusById                 | meteor method saveNotificationSettings       |                                                               |
| updateAutoTranslateLanguageById            | meteor method 'autoTranslate.saveSettings'   |                                                               |
| updateAutoTranslateById                    | meteor method 'autoTranslate.saveSettings'   |                                                               |
| removeByVisitorToken                       | LivechatClass cleanGuestHistory              |                                                               |
| changeDepartmentByRoomId                   | livechat helper updateChatDepartment         |                                                               |
| updateAllAutoTranslateLanguagesByUserId    | saveUserPreferences                          |                                                               |
| disableAutoTranslateByRoomId               | saveRoomEncrypted                            |                                                               |
| resetUserE2EKey                            | resetUserE2EEncriptionKey                    |                                                               |
| setAsUnreadByRoomIdAndUserId               | meteor method unreadMessages                 |                                                               |
| archiveByRoomId                            | ImportDataConverter.archiveRoomById; archiveRoom function |                                                         |
| unarchiveByRoomId                          | unarchiveRoom function                       |                                                               |
| updateNameAndAlertByRoomId                 | updateRoomName                               |                                                               |
| setCustomFieldsDirectMessagesByUserId      |                                              |                                                               |
| setFavoriteByRoomIdAndUserId               | meteor method toggleFavorite                 |                                                               |
| hideByRoomIdAndUserId                      | hideRoomMethod                               |                                                               |
| updateNameAndFnameById                     | updateGroupDMsName                           |                                                               |
| setUserUsernameByUserId                    | updateUsernameReferences                     |                                                               |
| updateFnameByRoomId                        | updateFName                                  |                                                               |
| updateDisplayNameByRoomId                  | registerContact; LivechatClass.saveRoomInfo  |                                                               |
| updateDirectNameAndFnameByName             | updateUsernameReferences                     |                                                               |
| incGroupMentionsAndUnreadForRoomIdExcludingUserId | incGroupMentions                          |                                                               |
| unsetBlockedByRoomId                       | unblockUser                                  |                                                               |
| setLastReplyForRoomIdAndUserIds            | updateThreadUsersSubscriptions               |                                                               |
| updateCustomFieldsByRoomId                 | saveRoomCustomFields                         |                                                               |
| setOpenForRoomIdAndUserIds                 | reply; updateThreadUsersSubscriptions        |                                                               |
| setAlertForRoomIdAndUserIds                | reply; updateThreadUsersSubscriptions        |                                                               |
| updateTypeByRoomId                         | saveRoomType                                 |                                                               |
| setBlockedByRoomId                         | meteor method blockUser                      |                                                               |
| incUserMentionsAndUnreadForRoomIdAndUserIds| updateUsersSubscriptions                     |                                                               |
| ignoreUser                                 | meteor method ignoreUser                     |                                                               |
| addRoleById                                | meteor methods addRoomLeader, addRoomModerator, and addRoomOwner |                               |
| removeRoleById                             | removeRoomLeader, removeRoomModerator, and removeRoomOwner |                                |
| setArchivedByUsername                      | deactivate > setArchivedByUsername; CROWD.syncDataToUser > setArchivedByUsername; meteor method setUserActiveStatus; ldapeemanager > disableMissingUsers and updateExistingUsers; ImportDataConverter > updateUser |      |
| updateUserHighlights                       | saveUserPreferences                          |                                                               |
| updateNotificationUserPreferences          | saveUserPreferences                          |                                                               |
| clearNotificationUserPreferences           | saveUserPreferences                          |                                                               |
| removeByUserId                             | deleteUser                                   |                                                               |
| createWithRoomAndUser                      |                                              |                                                               |
| createWithRoomAndManyUsers                 |                                              |                                                               |
| removeByRoomIdAndUserId                    |                                              |                                                               |
| removeByRoomIds                            |                                              |                                                               |
| addUnreadThreadByRoomIdAndUserIds          |                                              |                                                               |
| removeUnreadThreadByRoomIdAndUserId        |                                              |                                                               |
| openByRoomIdAndUserId                      | meteor method openRoom                       |                                                               |
| removeUnreadThreadsByRoomId                |                                              |                                                               |

</details>

## Steps to test or reproduce

1. Start RocketChat's application with the `DISABLE_DB_WATCHERS` environment variable set to `true`.
2. Perform HTTP requests or simulate use cases listed in the above table.

## Further comments

To maintain consistency and avoid potential regressions, event names and signatures have been kept unchanged on both the client and app sides. This decision streamlines efforts and mitigates the risk of unintended consequences.

[PROJ-7]: https://rocketchat.atlassian.net/browse/PROJ-7

[SCA-13]: https://rocketchat.atlassian.net/browse/SCA-13

[ADR #74]: https://adr.rocket.chat/0074


apps/meteor/app/autotranslate/server/methods/saveSettings.ts
```diff
@@ -53,13 +54,17 @@ Meteor.methods<ServerMethods>({
 					});
 				}
 
-				await Subscriptions.updateAutoTranslateById(subscription._id, value === '1');
+				(await Subscriptions.updateAutoTranslateById(subscription._id, value === '1')).modifiedCount &&
+					void notifyOnSubscriptionChangedById(subscription._id);
+
```
debdutdeb, line 59: we don't use shorthands like this, please use if(..) { ... } 🙈

apps/meteor/app/autotranslate/server/methods/saveSettings.ts
```diff
@@ -53,13 +54,17 @@ Meteor.methods<ServerMethods>({
...
 				if (!subscription.autoTranslateLanguage && options.defaultLanguage) {
-					await Subscriptions.updateAutoTranslateLanguageById(subscription._id, options.defaultLanguage);
+					(await Subscriptions.updateAutoTranslateLanguageById(subscription._id, options.defaultLanguage)).modifiedCount &&
+						void notifyOnSubscriptionChangedById(subscription._id);
```
debdutdeb, line 62: same

apps/meteor/app/autotranslate/server/methods/saveSettings.ts
```diff
@@ -53,13 +54,17 @@ Meteor.methods<ServerMethods>({
...
 				}
 				break;
 			case 'autoTranslateLanguage':
-				await Subscriptions.updateAutoTranslateLanguageById(subscription._id, value);
+				(await Subscriptions.updateAutoTranslateLanguageById(subscription._id, value)).modifiedCount &&
+					void notifyOnSubscriptionChangedById(subscription._id);
```
debdutdeb, line 67: same

debdutdeb (changes requested):
not adding same comment for all, but my first look :)

apps/meteor/app/e2e/server/functions/handleSuggestedGroupKey.ts
```diff
@@ -21,9 +23,12 @@ export async function handleSuggestedGroupKey(
 		throw new Meteor.Error('error-no-suggested-key-available', 'No suggested key available', { method });
 	}
 
-	if (handle === 'accept') {
-		await Subscriptions.setGroupE2EKey(sub._id, suggestedKey);
-	}
+	const e2eKeyUpdateResult =
```
sampaiodiego, line 26: this is changing the previous code behavior, where `unsetGroupE2ESuggestedKey` was always being called.

apps/meteor/app/federation/server/endpoints/dispatch.js
```diff
@@ -134,15 +139,24 @@ const eventHandlers = {
 				if (persistedSubscription) {
 					// Update the federation, if its not already set (if it's set, this is likely an event being reprocessed
 					if (!persistedSubscription.federation) {
-						await Subscriptions.updateOne({ _id: persistedSubscription._id }, { $set: { federation: subscription.federation } });
+						const { modifiedCount } = await Subscriptions.updateOne(
+							{ _id: persistedSubscription._id },
+							{ $set: { federation: subscription.federation } },
```
sampaiodiego, line 144: `federation` field is not on the list of `subscriptionFields`, so we don't need to notify when changing it.

apps/meteor/app/lib/server/functions/deleteUser.ts
```diff
@@ -98,7 +99,9 @@ export async function deleteUser(userId: string, confirmRelinquish = false, dele
 		const rids = subscribedRooms.map((room) => room.rid);
 		void notifyOnRoomChangedById(rids);
 
-		await Subscriptions.removeByUserId(userId); // Remove user subscriptions
+		// TODO: handle modifiedCount and/or return validity to call listener
```
sampaiodiego, line 102: missing TODO?

apps/meteor/app/lib/server/functions/setUserActiveStatus.ts
```diff
@@ -101,7 +108,10 @@ export async function setUserActiveStatus(userId: string, active: boolean, confi
 	}
 
 	if (user.username) {
-		await Subscriptions.setArchivedByUsername(user.username, !active);
+		const setArchivedResponse = await Subscriptions.setArchivedByUsername(user.username, !active);
+		if (setArchivedResponse.modifiedCount) {
+			void notifyOnSubscriptionChangedByUserIdAndRoomType(user._id, 'd');
```
sampaiodiego, line 113: this notify should be by name and not by `id`.. the reason is that we're deactivating `userA`, so we need to archive everyone else's subscriptions that had a DM with `userA`.. but this `notifyOnSubscriptionChangedByUserIdAndRoomType` is actually just notifying `userA` instead of everyone else.

apps/meteor/app/livechat/server/lib/Helper.ts
```diff
@@ -483,7 +497,11 @@ export const updateChatDepartment = async ({
 		LivechatRooms.changeDepartmentIdByRoomId(rid, newDepartmentId),
 		LivechatInquiry.changeDepartmentIdByRoomId(rid, newDepartmentId),
 		Subscriptions.changeDepartmentByRoomId(rid, newDepartmentId),
-	]);
+	]).then((data) => {
```
sampaiodiego, line 500: any special reason to use `.then`? if not pls don't =)

apps/meteor/ee/app/livechat-enterprise/server/hooks/onCloseLivechat.ts
```diff
@@ -20,7 +21,11 @@ const onCloseLivechat = async (params: LivechatCloseCallbackParams) => {
 		LivechatRooms.unsetOnHoldByRoomId(roomId),
 		Subscriptions.unsetOnHoldByRoomId(roomId),
 		AutoCloseOnHoldScheduler.unscheduleRoom(roomId),
-	]);
+	]).then((data) => {
```
sampaiodiego, line 24: pls don't use `then()` expect if really needed (like cannot convert a function to `async`)

apps/meteor/app/lib/server/lib/notifyListener.ts
```diff
@@ -501,3 +503,263 @@ export async function notifyOnUserChangeById({ clientAction, id }: { id: IUser['
 
 	void notifyOnUserChange({ id, clientAction, data: user });
 }
+
+export async function notifyOnSubscriptionChangedByUserAndRoomId(
```
sampaiodiego, line 507: any reason you chose to name this function like `ByUserAndRoomId`? the reason I ask is because usually (if not always) we use "room" first.. so this decision makes code look like this:
```ts
	const { modifiedCount } = await Subscriptions.hideByRoomIdAndUserId(rid, userId);
	if (modifiedCount) {
		void notifyOnSubscriptionChangedByUserAndRoomId(userId, rid);
	}
```

we can see the parameters are swiped.. they're correct, but I wonder if it would be better to have them in the same "position".. it would make it easier to look and prone to copy&paste bugs..

apps/meteor/app/lib/server/lib/notifyListener.ts
```diff
@@ -501,3 +503,263 @@ export async function notifyOnUserChangeById({ clientAction, id }: { id: IUser['
...
+}
+
+export async function notifyOnSubscriptionChangedByToken(token: string, clientAction: ClientAction = 'updated'): Promise<void> {
+	if (!dbWatchersDisabled) {
+		return;
+	}
+
+	const subscriptions =
+		clientAction === 'removed'
+			? Subscriptions.trashFind({ 'v.token': token }, { projection: subscriptionFields })
```
sampaiodiego, line 608: I don't know how often we'd notify removed records, maybe in this particular case we'd call more, since it can be more common to omnichannel use cases, but this comment may be relevant to every other notify of removed records: the "trash collection" doesn't have the same indexes as the "original" collections, and that's because there is only a single trash collection to store records from all other collections... so if the collection is somewhat full, having to query by different fields other than `_id` can be concerning.. 

the alternative would be to change wherever we call "notify removed" to first fetch all records that will be removed, remove them, and then notify, without having to perform a find in the trash collection.

not sure though, I'm open to discussion about this.

sampaiodiego (changes requested):
there are some inline comments... I've found these broken behaviors while testing:

- when sending a message in a channel, other members' subscription were not being marked as unread.. weirdly enough, if a message has a mention to someone, it works as expected
- change a channel to a group (from public to private) was not reflected in the UI


other than that, I'd ask to not use `.then()` when it is not needed, I've found it in a few different places.

sampaiodiego (approved):

apps/meteor/app/channel-settings/server/functions/saveRoomName.ts
```diff
@@ -24,10 +30,16 @@ const updateRoomName = async (rid: string, displayName: string, slugifiedRoomNam
...
 
-	return Promise.all([
+	const responses = await Promise.all([
 		Rooms.setNameById(rid, slugifiedRoomName, displayName),
 		Subscriptions.updateNameAndAlertByRoomId(rid, slugifiedRoomName, displayName),
 	]);
+
+	if (responses[1]?.modifiedCount) {
+		void notifyOnSubscriptionChangedByRoomId(rid);
+	}
```
KevLehman, line 40: A suggestion: we can run the notify attaching a `then` to the returned promise, so you don't have to await and then return
```
return Promise.all([
		Rooms.setNameById(rid, slugifiedRoomName, displayName),
		Subscriptions.updateNameAndAlertByRoomId(rid, slugifiedRoomName, displayName),
	]).then((p) => { void notify....() });

File: apps/meteor/app/channel-settings/server/functions/saveRoomEncrypted.ts [hunk 13]
```diff
@@ -27,7 +29,10 @@ export const saveRoomEncrypted = async function (rid: string, encrypted: boolean
 	}
 
 	if (encrypted) {
-		await Subscriptions.disableAutoTranslateByRoomId(rid);
+		const disableAutoTranslateResponse = await Subscriptions.disableAutoTranslateByRoomId(rid);
```
KevLehman, line 32:
```suggestion
		const { modifiedCount } = await Subscriptions.disableAutoTranslateByRoomId(rid);
```

apps/meteor/app/channel-settings/server/functions/saveRoomCustomFields.ts
```diff
@@ -3,21 +3,28 @@ import { Match } from 'meteor/check';
...
 			function: 'RocketChat.saveRoomCustomFields',
 		});
 	}
-	const ret = await Rooms.setCustomFieldsById(rid, roomCustomFields);
+
+	const response = await Rooms.setCustomFieldsById(rid, roomCustomFields);
 
 	// Update customFields of any user's Subscription related with this rid
-	await Subscriptions.updateCustomFieldsByRoomId(rid, roomCustomFields);
+	const updateCustomFields = await Subscriptions.updateCustomFieldsByRoomId(rid, roomCustomFields);
```
KevLehman, line 24:
```suggestion
	const { modifiedCount } = await Subscriptions.updateCustomFieldsByRoomId(rid, roomCustomFields);
```

apps/meteor/app/e2e/server/functions/handleSuggestedGroupKey.ts
```diff
@@ -31,4 +33,6 @@ export async function handleSuggestedGroupKey(
 	}
 
 	await Subscriptions.unsetGroupE2ESuggestedKey(sub._id);
+
+	void notifyOnSubscriptionChangedById(sub._id);
```
KevLehman, line 37: no check on modifiedCount?

apps/meteor/app/lib/server/functions/cleanRoomHistory.ts
```diff
@@ -83,7 +84,17 @@ export async function cleanRoomHistory({
...
-			await Subscriptions.removeUnreadThreadsByRoomId(rid, [...threads]);
+			const subscriptionIds: string[] = (
+				await Subscriptions.findUnreadThreadsByRoomId(rid, [...threads], { projection: { _id: 1 } }).toArray()
+			).map(({ _id }: { _id: string }) => _id);
+
+			const removedUnreadThreadsResponse = await Subscriptions.removeUnreadThreadsByRoomId(rid, [...threads]);
+
+			if (removedUnreadThreadsResponse.modifiedCount) {
+				for await (const id of subscriptionIds) {
+					await notifyOnSubscriptionChangedById(id);
```
KevLehman, line 95: Most of other calls are `void` promises, is there a reason to await for them here?

apps/meteor/app/lib/server/functions/closeLivechatRoom.ts
```diff
@@ -34,9 +35,12 @@ export const closeLivechatRoom = async (
 	}
 
 	if (!room.open) {
-		const subscriptionsLeft = await Subscriptions.countByRoomId(roomId);
-		if (subscriptionsLeft) {
-			await Subscriptions.removeByRoomId(roomId);
+		const { deletedCount } = await Subscriptions.removeByRoomId(roomId, {
```
KevLehman, line 38: The code removed here was part of a fix some time ago, there's a path of code that will do the removal on happy scenarios. This is for the scenario of a room being closed while still having subs, that's why it counted before removing, to avoid doing the delete query. 

(just making sure the same behavior is kept after this change)

apps/meteor/app/lib/server/functions/createRoom.ts
```diff
@@ -98,7 +102,13 @@ async function createUsersSubscriptions({
 		await Users.addRoomByUserIds(memberIds, room._id);
 	}
 
-	await Subscriptions.createWithRoomAndManyUsers(room, subs);
+	const { insertedIds } = await Subscriptions.createWithRoomAndManyUsers(room, subs);
+
+	if (Object.keys(insertedIds).length) {
+		for await (const insertedId of Object.values(insertedIds)) {
+			await notifyOnSubscriptionChangedById(insertedId, 'inserted');
```
KevLehman, line 109: Same, any reason for await this one?

apps/meteor/app/lib/server/functions/deleteRoom.ts
```diff
@@ -2,16 +2,24 @@ import { Messages, Rooms, Subscriptions } from '@rocket.chat/models';
...
+		},
+	});
+
 	await FileUpload.getStore('Avatars').deleteByRoomId(rid);
+
 	await callbacks.run('afterDeleteRoom', rid);
-	await Rooms.removeById(rid);
 
-	void notifyOnRoomChangedById(rid, 'removed');
+	(await Rooms.removeById(rid)).deletedCount && void notifyOnRoomChangedById(rid, 'removed');
```
KevLehman, line 24: Other places are using the `if (deletedCount)`, think we should keep the same standard here.

apps/meteor/app/lib/server/functions/relinquishRoomOwnerships.ts
```diff
@@ -1,17 +1,21 @@
...
+		Subscriptions.removeByRoomIds(rids, {
+			async onTrash(doc) {
+				void notifyOnSubscriptionChanged(doc, 'removed');
+			},
+		}),
 		Messages.removeByRoomIds(rids),
 		ReadReceipts.removeByRoomIds(rids),
 		Rooms.removeByIds(rids),
+		// no bulk deletion for files
+		...rids.map((rid) => FileUpload.removeFilesByRoomId(rid)),
```
KevLehman, line 18: IMHO, this change is not related to the watcher changes and should be in its own PR

apps/meteor/app/lib/server/lib/notifyListener.ts
```diff
@@ -467,3 +470,119 @@ export const notifyOnMessageChange = withDbWatcherCheck(async ({ id, data }: { i
...
+
+export const notifyOnSubscriptionChangedByRoomIdAndUserId = withDbWatcherCheck(
+	async (
+		rid: ISubscription['rid'],
+		uid: ISubscription['u']['_id'],
+		clientAction: Exclude<ClientAction, 'removed'> = 'updated',
+	): Promise<void> => {
+		const subscriptions = Subscriptions.findByUserIdAndRoomIds(uid, [rid], { projection: subscriptionFields });
+
+		for await (const subscription of subscriptions) {
```
KevLehman, line 488:
```suggestion
		for (const subscription of subscriptions) {
```

apps/meteor/app/lib/server/lib/notifyListener.ts
```diff
@@ -467,3 +470,119 @@ export const notifyOnMessageChange = withDbWatcherCheck(async ({ id, data }: { i
...
+		uid: ISubscription['u']['_id'],
+		notificationOriginField: keyof ISubscription,
+		originFieldNotEqualValue: 'user' | 'subscription',
+		clientAction: Exclude<ClientAction, 'removed'> = 'updated',
+	): Promise<void> => {
+		const subscriptions = Subscriptions.findByUserPreferences(uid, notificationOriginField, originFieldNotEqualValue, {
+			projection: subscriptionFields,
+		});
+
+		for await (const subscription of subscriptions) {
```
KevLehman, line 516:
```suggestion
		for (const subscription of subscriptions) {
```

KevLehman (changes requested):
In some places you destructure the returned objects and in others you access them via `.`, think would be nice to be consistent and stick to one way where this makes sense.

apps/meteor/app/lib/server/functions/closeLivechatRoom.ts
sampaiodiego: I removed the `count` since it is not really required .. if we want to remove them, we do so.. I kept the same behavior regarding returning an error or not by verifying if something was deleted or not (same as count > 0)

apps/meteor/app/channel-settings/server/functions/saveRoomName.ts
sampaiodiego: hmmm.. isn't mixing both a bit weird? not sure which one looks weirder :P 

```ts
	const responses = await Promise.all([
		Rooms.setNameById(rid, slugifiedRoomName, displayName),
		Subscriptions.updateNameAndAlertByRoomId(rid, slugifiedRoomName, displayName),
	]);

	if (responses[1]?.modifiedCount) {
		void notifyOnSubscriptionChangedByRoomId(rid);
	}

	return responses;
```

vs

```ts
	return Promise.all([
		Rooms.setNameById(rid, slugifiedRoomName, displayName),
		Subscriptions.updateNameAndAlertByRoomId(rid, slugifiedRoomName, displayName),
	]).then((ret) => {
		if (ret[1].modifiedCount) {
			void notifyOnSubscriptionChangedByRoomId(rid);
		}

		return ret;
	});
```

KevLehman: honestly, I liked the oplog :trollface: 

Yeah, this is a personal pref so not sure, they're function wise the same so go with the one that looks less ugly :/

apps/meteor/app/lib/server/lib/notifyListener.ts
sampaiodiego: this is because of the AsyncIterator.. but in fact, we don't need to use it

apps/meteor/app/lib/server/functions/closeLivechatRoom.ts
KevLehman: Thanks man!

sampaiodiego (approved):

KevLehman (dismissed):

sampaiodiego (approved):

PR #32541, opened 2024-06-03
fix: update cached setting immediately at the time of updating the db

debdutdeb:

Refreshing cache was depending on changestream, as the update in SettingRegistry rounds back through oplog to our watchers, then some listener.

Problem was that in such a case if anything asks for those setting values BEFORE oplog has got back (slow mongo, or slow instance/server) or we have processed the update, they'll get an old value, not the one intended. This becomes a problem when and if an initialization can't be changed on the fly, and restart will be required to use the right value.

The behavior would be noticed when the setting has already been inserted once with one value, before using OVERWRITE_SETTING_.+

**IIUC, now, we'd have to add listener callbacks (watch, watchMultiple, etc) BEFORE `settingRegistry.add`, so that the callbacks actually get called, as they are indeed changes to track and behave to, but then even inserts will be tracked**

We are discussing if this solution makes sense.

## Steps to test or reproduce

For a clearer visible example, remove Site_Url record from db, unset OVERWRITE_SETTING_Site_Url if you already have one, let the server start and insert the default record.

Now stop the server, set the OVERWRITE_.. start the server and notice the startup ascii banner showing old siteurl (:3000) not the new one.

## Further comments

[CORE-517](https://rocketchat.atlassian.net/browse/CORE-517)

[CORE-517]: https://rocketchat.atlassian.net/browse/CORE-517?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


MarcosSpessatto (changes requested):
Tested it locally using the use case you mentioned. I think I found a possible issue. I was able to reproduce following the step-by-step, but after running the steps, I started the server again without the OVERWRITE env var, and the value is not being updated anymore (to the default one). I'm not sure if this is the expected behavior though.

Also, we need to add a changeset to the PR.

MarcosSpessatto (approved):

PR #32549, opened 2024-06-03
chore!: Remove `meteor/check` from `custom-user-status` endpoints

matheusbsilva137:

 - Remove `meteor/check` from `custom-user-status` endpoints by using the `validateParams` property.

## Issue(s)

[CORE-318](https://rocketchat.atlassian.net/browse/CORE-318)

## Further comments

**The intention for this PR is to improve code readability and make invalid params errors more standardized. No side effects are expected. We're considering this a breaking change since invalid params errors are being changed to follow the same standard when possible.**

[CORE-318]: https://rocketchat.atlassian.net/browse/CORE-318?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/rest-typings/src/v1/customUserStatus.ts
```diff
@@ -35,14 +36,55 @@ const CustomUserStatusListSchema = {
 
 export const isCustomUserStatusListProps = ajv.compile<CustomUserStatusListProps>(CustomUserStatusListSchema);
 
+type CustomUserStatusCreateProps = { name: string; statusType?: UserStatus };
```
MarcosSpessatto, line 39: The validation expects statuses not 100% equal to this type as you can see here: https://github.com/RocketChat/Rocket.Chat/blob/develop/apps/meteor/app/user-status/server/methods/insertOrUpdateUserStatus.ts#L67

Are we good with this change?

packages/rest-typings/src/v1/customUserStatus.ts
```diff
@@ -35,14 +36,55 @@ const CustomUserStatusListSchema = {
...
+			enum: Object.values(UserStatus),
+		},
+	},
+	required: ['name'],
+	additionalProperties: false,
+};
+
+export const isCustomUserStatusCreateProps = ajv.compile<CustomUserStatusCreateProps>(CustomUserStatusCreateSchema);
+
+type CustomUserStatusUpdateProps = { _id: string; name: string; statusType?: UserStatus };
```
MarcosSpessatto, line 58: Same question as before

packages/rest-typings/src/v1/customUserStatus.ts
```diff
@@ -35,14 +36,55 @@ const CustomUserStatusListSchema = {
...
 
+type CustomUserStatusCreateProps = { name: string; statusType?: UserStatus };
+
+const CustomUserStatusCreateSchema = {
+	type: 'object',
+	properties: {
+		name: {
+			type: 'string',
+		},
+		statusType: {
```
MarcosSpessatto, line 47: nullable?

packages/rest-typings/src/v1/customUserStatus.ts
```diff
@@ -35,14 +36,55 @@ const CustomUserStatusListSchema = {
...
+const CustomUserStatusUpdateSchema = {
+	type: 'object',
+	properties: {
+		_id: {
+			type: 'string',
+		},
+		name: {
+			type: 'string',
+		},
+		statusType: {
```
MarcosSpessatto, line 69: nullable?

apps/meteor/tests/end-to-end/api/17-custom-user-status.js
```diff
@@ -40,4 +40,176 @@ describe('[CustomUserStatus]', function () {
 				.end(done);
 		});
 	});
+
+	describe('[/custom-user-status.create]', () => {
+		let customUserStatusId;
+
+		after(async () => {
```
MarcosSpessatto, line 47: Do we need this `async`?

apps/meteor/tests/end-to-end/api/17-custom-user-status.js
```diff
@@ -40,4 +40,176 @@ describe('[CustomUserStatus]', function () {
...
+					customUserStatusId = res.body.customUserStatus._id;
+				});
+		});
+	});
+
+	describe('[/custom-user-status.update]', () => {
+		let customUserStatusId;
+		const customUserStatusName = 'custom-status';
+
+		before(async () => {
```
MarcosSpessatto, line 116: Do we really need this `async`?

apps/meteor/tests/end-to-end/api/17-custom-user-status.js
```diff
@@ -40,4 +40,176 @@ describe('[CustomUserStatus]', function () {
...
+					expect(res.body).to.have.property('customUserStatus').that.is.an('object');
+					expect(res.body.customUserStatus).to.have.property('_id');
+					expect(res.body.customUserStatus).to.have.property('name', customUserStatusName);
+					expect(res.body.customUserStatus).to.have.property('statusType', 'online');
+					expect(res.body.customUserStatus).to.have.property('_updatedAt');
+					customUserStatusId = res.body.customUserStatus._id;
+				});
+		});
+
+		after(async () => {
```
MarcosSpessatto, line 136: Same

apps/meteor/tests/end-to-end/api/17-custom-user-status.js
MarcosSpessatto, line 1: We would need to ensure the logged user has this permission, just to avoid side effects: `manage-user-status`

MarcosSpessatto (changes requested):

packages/rest-typings/src/v1/customUserStatus.ts
matheusbsilva137: This is a required field (as you can check [here](https://github.com/RocketChat/Rocket.Chat/blob/5c18a1ff3af537de3a8af9e3536353e578069bea/apps/meteor/app/user-status/server/methods/insertOrUpdateUserStatus.ts#L68)). Since this is upposed to be break, I believe we can fix this (make it required even though it didn't use to be)

packages/rest-typings/src/v1/customUserStatus.ts
matheusbsilva137: This is a required field (as you can check [here](https://github.com/RocketChat/Rocket.Chat/blob/5c18a1ff3af537de3a8af9e3536353e578069bea/apps/meteor/app/user-status/server/methods/insertOrUpdateUserStatus.ts#L68)). Since this is upposed to be break, I believe we can fix this (make it required even though it didn't use to be)

packages/rest-typings/src/v1/customUserStatus.ts
matheusbsilva137: I've created a new `CustomizableUserStatus` type/enum to fix this

PR #32551, opened 2024-06-04
feat: E2EE warnings on search and audit panel

yash-rajpal:

While end-to-end encryption in Rocket.Chat provides enhanced privacy and security for communications, it inherently restricts certain functionalities, such as auditing messages, searching message content, and interacting with bots within encrypted rooms. Users may not be aware of these limitations, which can lead to confusion and a mismatch between user expectations and the platform's capabilities. Incorporating clear disclaimers into the user interface will help set accurate expectations and improve user satisfaction by transparently communicating the trade-offs involved with encryption.

image

image

## Further comments

[E2EE2-10](https://rocketchat.atlassian.net/browse/E2EE2-10)

[E2EE2-10]: https://rocketchat.atlassian.net/browse/E2EE2-10?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


dougfabris (changes requested):
I'm thinking we're adding unnecessary complexity in the user experience. Basically we're adding the support for finding the encrypted rooms and when they select, we warn them they can't proceed with what they want (audit feature). If the issue is that users are being frustrated because they don't know they can't use these features for e2e encryption we're failing on the explanation of our features. We should look for an alternative to warn them such as displaying these callouts based on e2e being enabled on the workspace, saying they won't be able to find the encrypted channels for example. Kinda similar what you're doing on search messages.

And this change definitely isn't a `chore`

yash-rajpal:
Thanks for the review doug, and I'm always having hard time figuring out if something is a `chore` or not 🤔 

But the idea is, encrypted rooms are still auditable, and encrypted rooms might contain some un-encrypted messages which can be audited. The warning is just to let the users know about the limitations. 

If we stop showing encrypted rooms on audit panel then the admin won't be able to see the un-encrypted messages which could still be audited.

packages/i18n/src/locales/en.i18n.json
```diff
@@ -1938,8 +1938,14 @@
...
   "Encrypted": "Encrypted",
   "Encrypted_channel_Description": "Messages are end-to-end encrypted, search will not work and notifications may not show message content",
+  "Encrypted_content_cannot_be_searched": "Encrypted content cannot be searched.",
   "Encrypted_key_title": "Click here to disable end-to-end encryption for this channel (requires e2ee-permission)",
   "Encrypted_message": "Encrypted message",
+<<<<<<< Updated upstream
+  "Encrypted_RoomType": "Encrypted {{roomType}}",
+=======
+  "Encrypted_message_preview_unavailable": "Encrypted message, preview unavailable",
+>>>>>>> Stashed changes
```
hugocostadev, line 1948: I guess it has a error here

hugocostadev (changes requested):
fix errors and implement tests

packages/i18n/src/locales/en.i18n.json
yash-rajpal: How'd that get here 🤔 

Will remove

hugocostadev:
Tests requirement dismissed due to the nature of this feature, does not worth to do e2e tests to check warning

hugocostadev (dismissed):

.changeset/friendly-months-attack.md
```diff
@@ -0,0 +1,6 @@
+---
+'@rocket.chat/i18n': patch
+'@rocket.chat/meteor': patch
```
MarcosSpessatto, line 3:
```suggestion
'@rocket.chat/i18n': minor
'@rocket.chat/meteor': minor
```

apps/meteor/client/lib/getRoomTypeTranslation.ts
```diff
@@ -0,0 +1,18 @@
+import type { IRoom } from '@rocket.chat/core-typings';
+
+import { t } from '../../app/utils/lib/i18n';
+
+export const getRoomTypeTranslation = (type: IRoom['t']) => {
```
MarcosSpessatto, line 5: What about adding a unit test to cover this function?

MarcosSpessatto (changes requested):

apps/meteor/client/lib/getRoomTypeTranslation.ts
yash-rajpal: Aligned this with Marcos, the function is pretty basic and simple, doesn't need a test for this function.

hugocostadev (approved):

apps/meteor/client/components/RoomAutoComplete/RoomAutoComplete.tsx
```diff
@@ -12,7 +13,11 @@ const generateQuery = (
 	selector: string;
 } => ({ selector: JSON.stringify({ name: term }) });
 
-type RoomAutoCompleteProps = Omit<ComponentProps<typeof AutoComplete>, 'filter'> & { scope?: 'admin' | 'regular' };
+type RoomAutoCompleteProps = Omit<ComponentProps<typeof AutoComplete>, 'filter'> & {
+	scope?: 'admin' | 'regular';
+	renderRoomIcon?: (props: { encrypted: IRoom['encrypted']; type: IRoom['t'] }) => ReactElement | null;
```
tassoevan, line 18: Assume this behavior will be absorbed by the component if the design team agrees, making the property pointless.

tassoevan (approved):

apps/meteor/client/components/RoomAutoComplete/RoomAutoComplete.tsx
yash-rajpal: Yes, incase we plan to show this in other places as well, this behavior will be implemented by the component itself. Thanks Tasso!

MarcosSpessatto (approved):

PR #32552, opened 2024-06-04
fix: Using the banner _id field as a fallback for the view.viewId field

Gustrb:

We were having issues with the seats taken banner, which would've been un-dismissable because it was expecting it to have a view.viewId field.
But since the banner was created before we had this, it was always empty.
This banner was removed in #31047
And it got into the 6.5.0 release onwards, so whenever someone hadn't dismissed the seatsTaken banner and upgraded their workspace, it would now be un-dismissable and would show a toast with the error 'invalid banner'

## Steps to test or reproduce

1. Create a workspace at 6.4.9
2. Go into the License.getMaxActiveUsers and change it to return some fixed amount (this will be easier to test)
3. Add the number of users that would correspond to 80% of that ammount
4. Restart the workspace (now there should be a seats taken banner)
5. Change the workspace version to 6.5.0 (or newer)
6. Close the seats taken banner

(I've had a few issues trying to run this test, my MongoDB would fail to start properly on 6.4.9. To solve this issue, I've deleted the contents of .meteor/local, which did the trick for me :) )

This should actually close the banner and add it to the dismissed banners collection inside MongoDB

## Further comments

I could've just added a new migration that would put the value of _id inside the view.viewId field, but since we already have adaptations to have legacy banners I don't think we should add a new migration just for this :P

JIRA Task: https://rocketchat.atlassian.net/browse/SUP-437


apps/meteor/server/services/banner/service.ts
```diff
@@ -68,6 +68,12 @@ export class BannerService extends ServiceClassInternal implements IBannerServic
 			.filter((banner) => !dismissed.has(banner._id))
 			.map((banner) => ({
 				...banner,
+				view: {
+					...banner.view,
+					// All modern banners should have a viewId, but we have old banners that were created without it
+					// such as the seatsTaken banner. In this case, we use the bannerId as the viewId
+					viewId: banner.view.viewId || banner._id,
+				},
```
KevLehman, line 76: Can we have unit tests of this func?

Gustrb: sure thing

PR #32590, opened 2024-06-12
chore!: Remove API_Use_REST_For_DDP_Calls setting

debdutdeb:
## Further comments

with the removal of most meteor methods, I don't see any particular reason to keep this setting around. This is a must for ms deployments anyway, no harm is just making it the only reality.


MarcosSpessatto (changes requested):
Also, we need to fix the lint problems: `cd apps/meteor && yarn eslint:fix`

PR #32608, opened 2024-06-14
test: Moderation Console E2E

csuadev:
Add e2e tests for the moderation console. 

## Issue(s)
[https://rocketchat.atlassian.net/browse/CORE-509](https://rocketchat.atlassian.net/browse/CORE-509)

### DIFF
--- a/apps/meteor/tests/e2e/moderation.spec.ts
+++ b/apps/meteor/tests/e2e/moderation.spec.ts
@@ -1,3 +1,5 @@
+import { faker } from '@faker-js/faker';
+
 import { Users } from './fixtures/userStates';
 import { Moderation, HomeChannel } from './page-objects';
 import { createTargetChannel, deleteChannel } from './utils';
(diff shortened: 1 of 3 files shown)


apps/meteor/tests/e2e/moderation.spec.ts
```diff
@@ -0,0 +1,55 @@
...
+		await deleteChannel(api, targetChannel);
+	});
+
+	test.describe('Message reporting', async () => {
+		test('should user be able to report a given message', async () => {
+			await poHomeChannel.sidenav.openChat(targetChannel);
+			await poHomeChannel.content.openLastMessageMenu();
+			await poModeration.reportMsgButton.click();
+			await poModeration.reportMessageReasonText.fill('Reason to report');
+			await poModeration.reportMessageReasonSubmit.click();
```
dougfabris, line 46: we do still need a assertion here, otherwise we're not testing anything...

dougfabris (changes requested):

apps/meteor/tests/e2e/moderation.spec.ts
dougfabris: what happens when an user reports a message successfully?

dougfabris: so this should be only a step for the assertion you're doing below! mix them up, then

MarcosSpessatto (changes requested):
It would be good to test all the behaviors in the moderation console. What do you think?

There are some available options for the admin, (dismiss, delete all messages deactivate user, etc). Do you think it would be possible to test them all? Or at least the ones we can simulate through the test environment...

Issue #32600, opened 2024-06-12
The latest version 6.9.1 not supported


lovekyleman:
As you can see the photo below 
![image](https://github.com/RocketChat/Rocket.Chat/assets/93461213/c522c23d-3cf3-499d-b836-a104ebdc4fb7)

LiloBzH:
Hello

Same issue here. With the last docker version, i have the same error.

lovekyleman:
> Hello
> 
> 
> 
> Same issue here. With the last docker version, i have the same error.
> 
> 

That's not good 😆😆

chotaire:
Ooookay!

![acb0f2a270447d88f124d912dbd83bef](https://github.com/RocketChat/Rocket.Chat/assets/9246549/71d973c0-f865-43d2-9d82-9ceec6481b5f)

![27424ed55b4a755e715b62d489002958](https://github.com/RocketChat/Rocket.Chat/assets/9246549/45193234-d211-4377-a56d-c952f15bce4d)

Gummikavalier:
Same in our test server too with 6.9.1.

No issues with 6.8.1.

ryanvito:
I've had mixed results. I've been able to restart the container and occasionally it shows up as supported. Not sure if the variability helps.
![image](https://github.com/RocketChat/Rocket.Chat/assets/4916594/8af32368-b817-4d79-a4ed-46d4ef40c1c7)

casalsgh:
Checking into it folks, will provide updates ASAP

lovekyleman:
> I've had mixed results. I've been able to restart the container and occasionally it shows up as supported. Not sure if the variability helps.
> 
> ![image](https://github.com/RocketChat/Rocket.Chat/assets/4916594/8af32368-b817-4d79-a4ed-46d4ef40c1c7)
> 
> 

Restart RC can't help here. Still not supported

thokich:
Hello

Same issue here. With the last docker version, i have the same error.

chotaire:
After reading https://github.com/RocketChat/Rocket.Chat/issues/31706 and seeing that nobody of the Rocket.Chat team actually ever really resolved this (or gave instructions on how to resolve this), I will cease updating our production Rocket.Chat instance until I see green light here that the issue has been resolved for everyone.

From my understanding this is not just a warning message. This would disconnect all mobile clients after max. 14 days if I read correctly. 

It is absolutely unbelievable how scared one must always be to update this piece of software.

[...]

PR #32610, opened 2024-06-14
fix: Supported Versions misbehaving

ggazzo:
https://github.com/RocketChat/Rocket.Chat/issues/32600
[CONN-254]

It seems that one of the conditions was wrong, always having the oldest sync

## Further comments

[CONN-254]: https://rocketchat.atlassian.net/browse/CONN-254?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/meteor/app/cloud/server/functions/supportedVersionsToken/supportedVersionsToken.ts
```diff
@@ -134,6 +135,31 @@ const getSupportedVersionsToken = async () => {
 		(response.success && response.result) || undefined,
 	);
 
+	SystemLogger.debug({
+		msg: 'Supported versions',
+		supportedVersionsFromBuild: JSON.stringify(supportedVersionsFromBuild.timestamp),
+		versionsFromLicense: JSON.stringify(versionsFromLicense?.supportedVersions?.timestamp),
+		response: JSON.stringify(response.success && response.result?.timestamp),
```
sampaiodiego, line 142: there is no need to call `JSON.stringify` since the logger will already do that before showing the log itself in case the log level is correct

apps/meteor/app/cloud/server/functions/supportedVersionsToken/supportedVersionsToken.ts
```diff
@@ -134,6 +135,31 @@ const getSupportedVersionsToken = async () => {
...
+			SystemLogger.info({
+				msg: 'Using supported versions from license',
+			});
+			break;
+		case response.success && response.result:
+			SystemLogger.info({
+				msg: 'Using supported versions from cloud',
+			});
+			break;
+	}
```
sampaiodiego, line 161: I guess this is not accurate logs? I mean, `supportedVersionsFromBuild` will always be defined, so it will probably always show `Using supported versions from build` ?

shouldn't this log be based on the return of `supportedVersionsChooseLatest`?

apps/meteor/jest.config.ts
```diff
@@ -33,6 +33,7 @@ const config: Config = {
 				'<rootDir>/app/livechat/server/business-hour/**/*.spec.ts?(x)',
 				'<rootDir>/app/livechat/server/api/**/*.spec.ts',
 				'<rootDir>/ee/app/authorization/server/validateUserRoles.spec.ts',
+				'<rootDir>/app/cloud/server/functions/supportedVersionsToken/**.spec.ts',
```
sampaiodiego, line 36: maybe for another PR but, how can we make this not needed? like adding a new test unit file without having to worry (too much) about it being picked up or not.

sampaiodiego (changes requested):

apps/meteor/jest.config.ts
ggazzo: sure

tassoevan (approved):

PR #32614, opened 2024-06-14
test: Fixing a weird problem happening when running the autojoin tests

Gustrb:

For some reason they would sometimes pass, and sometimes fail with weird error messages.
There were some weird after/afterEach in this test which I decided to remove to check what was happening and it looks like that did the trick :)


apps/meteor/tests/end-to-end/api/25-teams.js
```diff
@@ -1596,75 +1596,68 @@ describe('[Teams]', () => {
...
-						teamName: testTeam.name,
-					});
+				testUser1 = await createUser();
+				testUser2 = await createUser();
 			});
 
-			afterEach(() =>
-				Promise.all([deleteTeam(credentials, testTeam.name), deleteRoom({ roomId: createdRoom.body.channel._id, type: 'c' })]),
-			);
-
```
ggazzo, line 1631: this should continue in "after" if the test fails it will not remove the rooms

ggazzo (changes requested):

apps/meteor/tests/end-to-end/api/25-teams.js
Gustrb: oh yea, my bad

Gustrb:
The root cause was that sometimes users were being created with the same name, since the name is created through a Random Number Generator, #32634 is the correct solutions.
I assumed this PR passing all the tries was an indication that it solved the flakiness, but it appears to have been sheer (bad) luck.

PR #32615, opened 2024-06-14
fix: Livechat Monitors can't view departments they are not agents of

matheusbsilva137:

 - Send `onlyMyDepartments = false` param to the `livechat/departments` endpoints from the client (which will remove the agents filter that is currently being applied only to livechat-monitors);
 - Add end-to-end tests to make sure livechat-monitors can view and edit departments they are not agents of.

## Steps to test or reproduce

1. Login as a `livechat-monitor` (role) user;
2. Go to **Omnichannel > Departments** and create a new department (do not add the logged in monitor user as an agent to this department);

**_Current behavior:_** when redirected to the list after c reating the department, the monitor user won't be able to view the department they just created.
**_Expected behavior:_** the monitor should be able to view and edit departments that they are not agents of (as long as they have the `view-livechat-departments` permission) 

## Further comments

[SUP-601](https://rocketchat.atlassian.net/browse/SUP-601)

### DIFF
--- a/apps/meteor/client/views/omnichannel/departments/DepartmentsTable/DepartmentsTable.tsx
+++ b/apps/meteor/client/views/omnichannel/departments/DepartmentsTable/DepartmentsTable.tsx
@@ -40,7 +40,7 @@ const DepartmentsTable = ({ archived }: { archived: boolean }) => {
 
 	const query = useMemo(
 		() => ({
-			onlyMyDepartments: 'true' as const,
+			onlyMyDepartments: 'false' as const,
 			text: debouncedText,
 			sort: JSON.stringify({ [sortBy]: sortDirection === 'asc' ? 1 : -1 }),
 			...(current && { offset: current }),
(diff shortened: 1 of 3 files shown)


MartinSchoeler (changes requested):
Even if the monitor is an agent of a department, it still does not show up when calling the endpoint with `onlyMyDepartments: 'true'`  we should double check the endpoint and make sure nothing else that uses `onlyMyDepartments` has been broken

MartinSchoeler:
> The `onlyMyDepartments` param is misbehaving for monitors (since `livechat-monitor` is the only role for which this param takes effect). It's only returning departments from units that the monitor is an agent of -- for that I'll create another PR as soon as we have proper definition

I think it really depends on the desired behaviour for the monitors in the department page, if they are supposed to be able to see all departments then yeah we have two issues, but if they are supposed to only manage departments they belong to, then the only issue is in the endpoint, I would suggest getting some input on the subject with the product team

matheusbsilva137:
Let's move on with another approach: making the association of a department and a unit required in departments creation (for monitors) so that they can view it.

PR #32618, opened 2024-06-17
fix: login forgot password link opening new tab on mobile

csuadev:
Remove href on FieldLink component that was causing mobile browsers to open a new page when clicking on 'Forgot your password' link. Adding a cursor style and a role to maintain accesibility since the component by default doesn't have it. 

## Issue(s)
[https://rocketchat.atlassian.net/browse/CORE-245](https://rocketchat.atlassian.net/browse/CORE-245)

## Steps to test or reproduce
- Go to Rocket.Chat login on a mobile browser.
- Tap on 'Forgot your password'.


dougfabris (changes requested):
`FieldLink` is already an anchor element it doesn't need the `link` role. 
Also its wrong to force the style since the link will have cursor: pointer if it receives the `href` attribute. 
So, instead of adding a `onClick` we have to provide a valid href to this link. 

I would suggest creating a new route to the forgot password and getting from the `useRouter `

PR #32632, opened 2024-06-19
feat: Improve UX from "Action Buttons" in `RoomInfo`

tiagoevanp:

Improve the UX from "Action Buttons" inside the contextualbar in RoomInfo component

Before:
image

After:
image

## Further comments

[CORE-464](https://rocketchat.atlassian.net/browse/CORE-464?atlOrigin=eyJpIjoiZTZlOTBjNWJmZjg3NDZmYmJjMDcwYWNiNDlkOGFmZDAiLCJwIjoiaiJ9)

[CORE-464]: https://rocketchat.atlassian.net/browse/CORE-464?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


dougfabris:
if you're going to improve the UX it will change something to the end user, so it shouldn't be a `chore`

MarcosSpessatto:
> if you're going to improve the UX it will change something to the end user, so it shouldn't be a `chore`

Maybe a fix would fit better in this case, also adding a changeset.

MarcosSpessatto (approved):

apps/meteor/client/views/room/contextualBar/Info/RoomInfo/Actions.tsx
dougfabris, line 1: I think it is an unnecessary change, its ok to have it there in `RoomInfo`

apps/meteor/client/views/room/contextualBar/Info/RoomInfo/RoomActionsMenu.tsx
dougfabris, line 1: I think we can use the `GenericMenu` instead

apps/meteor/client/views/room/contextualBar/Info/hooks/useRoomActions.ts
```diff
@@ -8,75 +8,111 @@ import { useRoomHide } from './actions/useRoomHide';
 import { useRoomLeave } from './actions/useRoomLeave';
 import { useRoomMoveToTeam } from './actions/useRoomMoveToTeam';
 
-type RoomActions = {
+type UseRoomActionsOptions = {
```
dougfabris, line 11:
```suggestion
type RoomActionsProps = {
```

apps/meteor/client/views/room/contextualBar/Info/hooks/useRoomActions.ts
```diff
@@ -8,75 +8,111 @@ import { useRoomHide } from './actions/useRoomHide';
...
+							id: 'convert_channel_team',
+							name: t('Teams_convert_channel_to_team'),
+							icon: 'team' as const,
+							action: handleConvertToTeam,
+						},
+				  ]
+				: []),
+		];
+
+		if (memoizedActions.length <= size) {
```
dougfabris, line 98: **this shouldn't live here**, the hook should just return the actions! 
This rule to display the actions should live outside here

dougfabris (changes requested):

dougfabris:
@tiagoevanp if you're improving, it's not a fix 😞

apps/meteor/client/views/room/contextualBar/Info/RoomInfo/Actions.tsx
tiagoevanp: I don't think so, if we are returning JSX from a function... it would be another component...

apps/meteor/client/views/room/contextualBar/Info/hooks/useRoomActions.ts
tiagoevanp: Have you seen that is not embracing all the props? only the second one that fits better as options for the hook...

apps/meteor/client/views/room/contextualBar/Info/RoomInfo/Actions.tsx
dougfabris: so what about changing the same to a less generic such as `RoomInfoActions`

tiagoevanp: perfect idea!

apps/meteor/client/views/room/contextualBar/Info/hooks/useRoomActions.ts
dougfabris: `RoomActionsOptions` then I don't think Use fits for type naming since use is a convention to the hook name, its kinda weird

apps/meteor/client/views/room/contextualBar/Info/hooks/useRoomActions.ts
```diff
@@ -8,75 +8,111 @@ import { useRoomHide } from './actions/useRoomHide';
...
 	onClickEnterRoom?: () => void;
 	onClickEdit?: () => void;
+	resetState?: () => void;
 };
 
-export const useRoomActions = (room: IRoom, { onClickEnterRoom, onClickEdit }: RoomActions, resetState?: () => void) => {
+/**
+ *
+ * @param room
+ * @param param1
```
dougfabris, line 20:
```suggestion
 * @param param1
```
what is this?

apps/meteor/client/views/room/contextualBar/Info/hooks/useRoomActions.ts
tiagoevanp: I think this is more about a "naming conventions". I think it's easier to see that the type itself belongs to a hook, if it was a type exported everyone should presume the origin by the name.

apps/meteor/client/views/room/contextualBar/Info/hooks/useRoomActions.ts
tiagoevanp: fixing the docs... it should be good now!

apps/meteor/client/views/room/contextualBar/Info/hooks/useRoomActions.ts
tiagoevanp: Why not? We are avoiding some unnecessary computations with an "Early Return" with the right object values...

apps/meteor/client/views/room/contextualBar/Info/RoomInfo/RoomActionsMenu.tsx
tiagoevanp: Got it!

apps/meteor/client/views/room/contextualBar/Info/hooks/useRoomActions.ts
dougfabris: The computation here is minimum since the actions are memoized. 
Ideally the hook should do a isolated job, such as return all possible actions. 

In this case you're assuming every single time we want to use this hook we'll get the actions divided based on the very specific render condition on this view, not allowing us to render the actions differently or having to do some magic when trying another approach.

tiagoevanp: this hook is so specific that I think there is no necessity to do a different approach... and it's possible to adjust the division passing the `size` option to the hook.

dougfabris: I'm sorry, but I don't agree with it. I understand we can't keep using `useActionsSpread` since this hook was created using a different pattern, but IMO it should be handled separately. Because of that I can't approve it.

tassoevan: @dougfabris That split was a parameter coming from `useActionsSpread`, `size`. You're arguing the hook should not split it or provide a way to customize the value? If the former, it's kinda pointless, since we already split the actions for the rendering logic as `regular` e `danger`; if the latter, promoting the constant as a parameter is adding an accidental complexity, since there is no reuse of `useRoomActions` anywhere else _yet_.

tassoevan: > Ideally the hook should do a isolated job, such as return all possible actions.

That's why `useActionsSpread` existed, because it was carrying the rendering logic while `useRoomActions` was dealing only with building the data structure. Not sure if we gain anything splitting those again for use in a single place.

tassoevan: Splitting is fine, I'm just preaching the gospel: https://overreacted.io/the-wet-codebase/

apps/meteor/tests/e2e/page-objects/fragments/home-flextab-room.ts
```diff
@@ -11,8 +11,12 @@ export class HomeFlextabRoom {
 		return this.page.locator('role=button[name="Edit"]');
 	}
 
-	get btnDelete(): Locator {
-		return this.page.locator('role=button[name="Delete"]');
+	get kebab(): Locator {
```
dougfabris, line 14:
```suggestion
	get btnMore(): Locator {
```

apps/meteor/tests/e2e/page-objects/fragments/home-flextab-room.ts
```diff
@@ -11,8 +11,12 @@ export class HomeFlextabRoom {
...
+		return this.page.locator('role=button[name="More"]');
+	}
+
+	get itemDelete(): Locator {
```
dougfabris, line 18:
```suggestion
	get optionDelete(): Locator {
```

apps/meteor/tests/e2e/channel-management.spec.ts
```diff
@@ -178,7 +178,10 @@ test.describe.serial('channel-management', () => {
 		await user1Page.goto(`/channel/${targetChannel}`);
 		await user1Channel.waitForChannel();
 		await user1Channel.tabs.btnRoomInfo.click();
-		await expect(user1Channel.tabs.room.btnDelete).toBeVisible();
+
+		await user1Channel.tabs.room.kebab.click();
```
dougfabris, line 182:
```suggestion
		await user1Channel.tabs.room.btnMore.click();
```

apps/meteor/tests/e2e/channel-management.spec.ts
```diff
@@ -178,7 +178,10 @@ test.describe.serial('channel-management', () => {
...
+
+		await expect(user1Channel.tabs.room.itemDelete).toBeVisible();
```
dougfabris, line 184:
```suggestion
		await expect(user1Channel.tabs.room.optionDelete).toBeVisible();
```

dougfabris (changes requested):

dougfabris (approved):
LGTM!

dougfabris (approved):

dougfabris (approved):

PR #32634, opened 2024-06-19
test: fix flaky `Team join` api tests

ggazzo:

### DIFF
--- a/apps/meteor/tests/end-to-end/api/25-teams.js
+++ b/apps/meteor/tests/end-to-end/api/25-teams.js
@@ -1602,9 +1602,8 @@ describe('[Teams]', () => {
 			let testUser2;
 
 			before(async () => {
-				const [testUserResult, testUser1Result] = await Promise.all([createUser(), createUser()]);
-				testUser1 = testUserResult;
-				testUser2 = testUser1Result;
+				testUser1 = await createUser();
+				testUser2 = await createUser();
 			});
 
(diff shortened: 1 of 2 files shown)


Gustrb (approved):

apps/meteor/tests/end-to-end/api/25-teams.js
```diff
@@ -1602,9 +1602,8 @@ describe('[Teams]', () => {
 			let testUser2;
 
 			before(async () => {
-				const [testUserResult, testUser1Result] = await Promise.all([createUser(), createUser()]);
-				testUser1 = testUserResult;
-				testUser2 = testUser1Result;
+				testUser1 = await createUser();
+				testUser2 = await createUser();
```
matheusbsilva137, line 1606: Wouldn't it be a better option to leave this as it is and change `createUser` instead? :thinking: So that we avoid this to happen again in case someone else tries to create users with `Promise.all`
I see that `createUser` uses `Date.now()` to create the username, maybe we could add a `Math.random()` there too

matheusbsilva137 (changes requested):

apps/meteor/tests/end-to-end/api/25-teams.js
Gustrb: nice one!

matheusbsilva137 (approved):

PR #32648, opened 2024-06-20
chore!: remove deprecated meteor eraseRoom method

AllanPazRibeiro:
This PR removes the eraseRoom to use the room.delete endpoint

## Issue(s)

[CORE-350](https://rocketchat.atlassian.net/browse/CORE-350)

## Further comments

[CORE-350]: https://rocketchat.atlassian.net/browse/CORE-350?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.changeset/chilled-boats-sip.md
+++ b/.changeset/chilled-boats-sip.md
@@ -1,0 +1,5 @@
+---
+'@rocket.chat/meteor': patch
+---
+
+This adjustment removes the deprecated `eraseRoom` method. Moving forward, use the `room.delete` endpoint to delete rooms.
--- a/apps/meteor/app/api/server/v1/channels.ts
+++ b/apps/meteor/app/api/server/v1/channels.ts
@@ -24,6 +24,7 @@ import { Meteor } from 'meteor/meteor';
 
 import { isTruthy } from '../../../../lib/isTruthy';
 import { findUsersOfRoom } from '../../../../server/lib/findUsersOfRoom';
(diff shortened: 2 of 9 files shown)


abhinavkrin (changes requested):
I think we should also move away from using `Meteor.Error` in the `eraseRoom` function and instead use `Error` object. This will be a step from moving away from meteor. Wdys?

intead of
https://github.com/RocketChat/Rocket.Chat/blob/2d06f5e986edc5f5b6fb9c332009f7ec4f8b69b8/apps/meteor/server/methods/eraseRoom.ts#L14
using
```
throw new Error('error-invalid-room')
```

ggazzo:
> I think we should also move away from using `Meteor.Error` in the `eraseRoom` function and instead use `Error` object. This will be a step from moving away from meteor. Wdys?
> 
> intead of
> 
> https://github.com/RocketChat/Rocket.Chat/blob/2d06f5e986edc5f5b6fb9c332009f7ec4f8b69b8/apps/meteor/server/methods/eraseRoom.ts#L14
> 
> 
> using
> ```
> throw new Error('error-invalid-room')
> ```

Meteor error and Error are handled differently by the api, `Error` is not exposed to final users.

ggazzo (approved):

ggazzo (dismissed):

KevLehman (approved):

PR #32649, opened 2024-06-21
regression: fix retention flaky test

ggazzo:


gabriellsh (approved):
Let that one pass by, thanks a lot for the fix!

Issue #32387, opened 2024-05-08
Right click on a message to access options on the right under 3 doted menu items like, Star, edit/delete message or Quote & reaction etc


Ahmad-Ali-com:
It is not an issue but a suggestion if we could have this option in the coming versions to access the menu item under 3 dotted icon on the right or reaction & quote option accessible by right clicking on any message, that way using these features would be much more user friendly.

We can keep them where they are right now but additionally should be accessible with the ease of a simple right click on the message similar to FaceBook or WhatsApp chats. 

I can explain further if needed, feel free to reach out.

reetp:
Most people have expected results in a browser when they right click.

It might be something that would be considered in the Electron client - you can open a bug there

https://github.com/RocketChat/Rocket.Chat.Electron

joanamarquees:
Hello, I've already a solution to this, but with a twist. In the web version, I've implemented a double-click on messages to open the options menu. As you can see.


https://github.com/RocketChat/Rocket.Chat/assets/99995545/5f59067e-eff1-4b28-8e57-ff4f991f588d

Can I open a pull request for it?

Ahmad-Ali-com:
> Hello, I've already a solution to this, but with a twist. In the web version, I've implemented a double-click on messages to open the options menu. As you can see.
> Screencast.from.28-05-2024.11.47.14.webm
> 
> Can I open a pull request for it?

instead of double click, it should work with right click as double click is being used already to select the message's text also the menu should expand/appear right where you right click that way it could be user friendly

OrneyVortex:
I would like to work on this feature
It seems to be interesting

reetp:
See my comment above about expected results.

Gustrb:
I agree with @reetp, this feature would not give expected results on a browser, but it could be interesting in the electron app.

DevRish:
Hi RocketChat Team!

In accordance with the previous comments, I have implemented the requested feature for the electron app, without affecting the web app. However, since the electron app renders the web app itself in a webview, the changes had to be made in this repository itself. 
So, I didn't open a separate issue in the desktop app repository...
I have also implemented the feature after viewing and keeping in mind the reviews on the previous closed PRs...

Please check it out and provide your reviews, whenever possible. Thanks!

PR #32666, opened 2024-06-24
feat: Open messages menu by right clicking (#32387)

joanamarquees:

Due to a request, I decided to implement another option to open the message options menu. If you right-click on a message, the menu, which used to appear by clicking on the 3 dots in the corner of the message, now appears wherever you right-click on a message.

## Issue(s)

#32387

## Steps to test or reproduce

Steps already described here and on the issue page.


apps/meteor/client/components/message/variants/RoomMessage.tsx
```diff
@@ -62,6 +62,24 @@ const RoomMessage = ({
 
 	const messageRef = useJumpToMessage(message._id);
 
+	// Add a new function for handling double click events
+	const handleRightClick = async (e: React.MouseEvent) => {
+		e.preventDefault();
+
+		//get the button that oppens the menu
+		const moreButton = document.querySelector('[title="More"]');
```
ggazzo, line 70: I suspect this will only work when your language is English

ggazzo: Also, I think you should test where you are clicking, there are links and buttons inside the message where you might want to use the original right click

PR #32677, opened 2024-06-25
chore: Log apps runtime data on SIGUSR2

d-gubert:

Install a handler for the `SIGUSR2` os signal that prints debug data to the `stdout`

After the server is fully up and running, you can use `kill -s SIGUSR2 [pid]` or `kill -12 [pid]` for the server to print the information


d-gubert:
This needs to be better planned and structured

PR #32678, opened 2024-06-26
refactor(ui-kit): Add tests for types

tassoevan:

UIKit types are *not open for extension*, therefore we can't use `interface` declarations nor enjoy the advantage of the `extends` construct for them. This pull request add types for assert type shapes in compilation type.


dougfabris (approved):
LGTM!

PR #32682, opened 2024-06-27
feat: Allow managing association to business units on departments' creation and update

matheusbsilva137:

 - Added a new optional `departmentUnit` object param to `livechat:saveDepartment` Meteor method, `livechat/department` (**POST**) endpoint and `livechat/department` (**PUT**) endpoint;
     - Providing an empty object (`{}`) to this param removes the department from its associated unit (or just doesn't add the department to any unit, if provided on the creation of a new department). Providing `{ _id: undefined }` or `null` triggers the same effect;
     - If no value is provided to this new param, then the association between the department and its unit (if it exists) won't be changed, nor created or removed;
     - When providing a string `_id` field in the object (`{ _id: "unit-id" }`), the department will be associated to the unit with the specified `_id` **IF** the user has the permission to do so.
 - In order to use any of the updated endpoints, the user must have the `manage-livechat-departments` permission (this is already the current behavior), but there are additional conditions to associate a department with a specific unit when the new param is provided:
     - If the user is a **monitor** (role `livechat-monitor`), then they must be a supervisor of the specified unit. That is, they must be linked (as **monitor**) to the specified unit in order to add or remove departments from it;
     - If the user is not a monitor (such as an `admin` or `livechat-manager`), any unit can be associated to the department (on creation or update).
 - Providing any value different from a string to the `_id` field to the new param's object will trigger an error.
 - This is an **additional**/non-blocking operation. No errors are thrown (yet) even if the user tries to add a department to a unit they don't have the permission to manage.

## Steps to test or reproduce

Example request to the `livechat:saveDepartment` method (the new param is the last one that should be provided -- `{\"_id\":\"unit-id\"}` in the example):
```
curl --request POST \
  --url http://localhost:3000/api/v1/method.call/livechat%3AsaveDepartment \
  --header 'Accept: application/json' \
  --header 'X-Auth-Token: f4U57vG-drmhd8hZbKAT_9TTJqW48Y_VB1zle6HWiRf' \
  --header 'X-User-Id: WM8KyWfu6qXDCRZrq' \
  --data '{"message":"{\"msg\":\"method\",\"id\":\"15\",\"method\":\"livechat:saveDepartment\",\"params\":[null,{\"enabled\":false,\"name\":\"My new department5\",\"description\":\"\",\"showOnRegistration\":false,\"showOnOfflineForm\":false,\"requestTagBeforeClosingChat\":false,\"email\":\"my-new-testdep5@example.com\",\"chatClosingTags\":[],\"offlineMessageChannelName\":\"\",\"abandonedRoomsCloseCustomMessage\":\"\",\"waitingQueueMessage\":\"\",\"departmentsAllowedToForward\":[],\"fallbackForwardDepartment\":\"\",\"allowReceiveForwardOffline\":false},[],{\"_id\":\"unit-id\"}]}"}'
```
The command above triggers the creation of a new department associated with the unit with id `unit-id`. Change the first `null` param to a valid `department` string to edit it.

Example request to the `livechat/department` POST endpoint (check the new `departmentUnit` param):

```
curl --request POST \
  --url http://localhost:3000/api/v1/livechat/department \
  --header 'X-Auth-Token: f4U57vG-drmhd8hZbKAT_9TTJqW48Y_VB1zle6HWiRf' \
  --header 'X-User-Id: WM8KyWfu6qXDCRZrq' \
  --data '{
  "department": {
    "name": "Test23456",
    "enabled": true,
    "showOnOfflineForm": true,
    "showOnRegistration": true,
    "email": "bla@bla"
  },
  "departmentUnit": { "_id": "unit-id" }
}'
```

Example request to the `livechat/department/:id` PUT endpoint (check the new `departmentUnit` param -- providing an empty object to it will remove the department from its associated unit):

```
curl --request PUT \
  --url http://localhost:3000/api/v1/livechat/department/666cc15cd32e4e47b82332fd \
  --header 'X-Auth-Token: f4U57vG-drmhd8hZbKAT_9TTJqW48Y_VB1zle6HWiRf' \
  --header 'X-User-Id: WM8KyWfu6qXDCRZrq' \
  --data '{
  "department": {
    "name": "Test23456",
    "enabled": true,
    "showOnOfflineForm": true,
    "showOnRegistration": true,
    "email": "bla@bla"
  },
  "departmentUnit": {}
}'
```

## Further comments

[SUP-613](https://rocketchat.atlassian.net/browse/SUP-613)

[SUP-613]: https://rocketchat.atlassian.net/browse/SUP-613?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/meteor/tests/end-to-end/api/livechat/14-units.ts
```diff
@@ -409,4 +413,428 @@ import { IS_EE } from '../../../e2e/config/constants';
...
+				.expect(200)
+				.expect((res) => {
+					expect(res.body).to.have.property('success', true);
+					expect(res.body).to.have.property('department').that.is.an('object');
+					expect(res.body.department).to.have.property('name', departmentName);
+					expect(res.body.department).to.have.property('type', 'd');
+					expect(res.body.department).to.have.property('_id');
+					departmentId = res.body.department._id;
+				});
+		};
```
MarcosSpessatto, line 486: It would be better to move this to the helpers instead and return the departmentId.

apps/meteor/tests/end-to-end/api/livechat/14-units.ts
```diff
@@ -409,4 +413,428 @@ import { IS_EE } from '../../../e2e/config/constants';
...
+			unit = await createUnit(monitor1._id, monitor1.username, []);
+		});
+
+		after(async () => Promise.all([deleteUser(monitor1), deleteUser(monitor2), deleteUnit(unit)]));
+
+		afterEach(() => {
+			if (departmentId) {
+				return deleteDepartment(departmentId);
+			}
+		});
```
MarcosSpessatto, line 504: I suggest keeping the deletion inside of each `it` creating it, otherwise, since it's not applied for all `it`s it could cause confusion on when it should be removed.

apps/meteor/tests/end-to-end/api/livechat/14-units.ts
```diff
@@ -409,4 +413,428 @@ import { IS_EE } from '../../../e2e/config/constants';
 			await deleteUser(user);
 		});
 	});
+
+	const testDepartmentsInUnit = (unitId: string, unitName: string, numDepartments: number) => {
```
MarcosSpessatto, line 417: It would be better to have an object as a parameter here, mainly to avoid magic numbers when calling the function: `testDepartmentsInUnit('test', 'test', 0)`

apps/meteor/tests/end-to-end/api/livechat/14-units.ts
```diff
@@ -409,4 +413,428 @@ import { IS_EE } from '../../../e2e/config/constants';
...
+
+		it('should fail creating a department into an existing unit that a monitor does not supervise', async () => {
+			const departmentName = 'Test-Department-Unit2';
+
+			await createDepartmentInUnit(monitor2Credentials, departmentName, unit._id);
+			await testDepartmentsInUnit(unit._id, unit.name, 0);
+			await testDepartmentAncestors(departmentId, departmentName);
+		});
+
+		it('should succesfully create a department into an existing unit as a livechat manager', async () => {
```
MarcosSpessatto, line 530: Shouldn't we create a case for the `admin`s also?

apps/meteor/tests/end-to-end/api/livechat/14-units.ts
```diff
@@ -409,4 +413,428 @@ import { IS_EE } from '../../../e2e/config/constants';
...
+			await testDepartmentsInUnit(unit._id, unit.name, 1);
+			await testDepartmentAncestors(departmentId, departmentName, unit._id);
+		});
+
+		it('should succesfully create a department into an existing unit that a monitor supervises', async () => {
+			const departmentName = 'Test-Department-Unit3';
+
+			await createDepartmentInUnit(monitor1Credentials, departmentName, unit._id);
+
+			// Deleting a department currently does not decrease its unit's counter. We must adjust this check when this is fixed
```
MarcosSpessatto, line 543: Can you please open a task for it?

apps/meteor/tests/end-to-end/api/livechat/14-units.ts
```diff
@@ -409,4 +413,428 @@ import { IS_EE } from '../../../e2e/config/constants';
...
+
+	describe('[PUT] livechat/department', () => {
+		let monitor1: Awaited<ReturnType<typeof createUser>>;
+		let monitor1Credentials: Awaited<ReturnType<typeof login>>;
+		let monitor2: Awaited<ReturnType<typeof createUser>>;
+		let monitor2Credentials: Awaited<ReturnType<typeof login>>;
+		let unit: IOmnichannelBusinessUnit;
+		let department: ILivechatDepartment;
+
+		const updateDepartment = (userCredentials: Credentials, departmentId: string, updatedName: string, unitId?: string) => {
```
MarcosSpessatto, line 557: I'm almost sure we already have a helper function that does that.

apps/meteor/tests/end-to-end/api/livechat/14-units.ts
```diff
@@ -409,4 +413,428 @@ import { IS_EE } from '../../../e2e/config/constants';
...
+		});
+
+		it("should fail updating a department's unit when providing an invalid _id type in the department unit object", () => {
+			const updatedName = 'updated-department-name';
+			return request
+				.put(api(`livechat/department/${department._id}`))
+				.set(credentials)
+				.send({
+					department: { name: updatedName, enabled: true, showOnOfflineForm: true, showOnRegistration: true, email: 'bla@bla' },
+					departmentUnit: { _id: true },
```
MarcosSpessatto, line 613: What about when it provides a valid input (string) but for a non-existing unit ?

apps/meteor/tests/end-to-end/api/livechat/14-units.ts
```diff
@@ -409,4 +413,428 @@ import { IS_EE } from '../../../e2e/config/constants';
...
+				.post(methodCall('livechat:saveDepartment'))
+				.set(credentials)
+				.send({
+					message: JSON.stringify({
+						method: 'livechat:saveDepartment',
+						params: [
+							'',
+							{ name: 'Fail-Test', enabled: true, showOnOfflineForm: true, showOnRegistration: true, email: 'bla@bla' },
+							[],
+							{ _id: true },
```
MarcosSpessatto, line 769: Same as before

apps/meteor/tests/end-to-end/api/livechat/14-units.ts
```diff
@@ -409,4 +413,428 @@ import { IS_EE } from '../../../e2e/config/constants';
...
+			await createDepartmentWithLivechatMethod(monitor2Credentials, testDepartmentId, departmentName);
+			await testDepartmentsInUnit(unit._id, unit.name, 1);
+			await testDepartmentAncestors(testDepartmentId, departmentName, unit._id);
+			await deleteDepartment(testDepartmentId);
+		});
+
+		it('should succesfully create a department in a unit that a monitor supervises', async () => {
+			await createDepartmentWithLivechatMethod(monitor1Credentials, '', departmentName, unit._id);
+
+			// Deleting a department currently does not decrease its unit's counter. We must adjust this check when this is fixed
```
MarcosSpessatto, line 835: Same

apps/meteor/tests/end-to-end/api/livechat/14-units.ts
MarcosSpessatto, line 1: I think we should not encapsulate especially the assertion into functions, otherwise is really hard for the test readers to know what's going on. I agree we must have helper functions to create/update/delete stuff, but I think we must have the assertions inside each `it`. It's preferrable for the helper functions to interact with the API and return data, so we can make the assertion explicitly.

apps/meteor/ee/app/livechat-enterprise/server/hooks/manageDepartmentUnit.ts
```diff
@@ -0,0 +1,51 @@
+import type { ILivechatDepartment, IOmnichannelBusinessUnit } from '@rocket.chat/core-typings';
+import { LivechatDepartment, LivechatUnit } from '@rocket.chat/models';
+
+import { hasAnyRoleAsync } from '../../../../../app/authorization/server/functions/hasRole';
+import { callbacks } from '../../../../../lib/callbacks';
+import { getUnitsFromUser } from '../methods/getUnitsFromUserRoles';
+
+callbacks.add(
+	'livechat.manageDepartmentUnit',
+	async ({ userId, departmentId, unitId }) => {
```
MarcosSpessatto, line 10: Could we create unit tests for this? You don't need to simulate the callback executing in order to test that, it's acceptable to extract the callback function into a function and export it, so it helps you to test that.

MarcosSpessatto (changes requested):

apps/meteor/app/livechat/imports/server/rest/departments.ts
```diff
@@ -57,10 +57,18 @@ API.v1.addRoute(
 			check(this.bodyParams, {
 				department: Object,
 				agents: Match.Maybe(Array),
+				departmentUnit: Match.Maybe({ _id: Match.Optional(String) }),
```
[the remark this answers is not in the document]
matheusbsilva137: Since we're also allowing users to remove departments from units here, I thought the distinction between `undefined` and `null` could be misleading like this
I mean, in order to remove a department from a unit, users would need to provide `{}` or `{ _id: undefined }`, while if we didn't have an object, just `null` would need to remove the department from its unit. So I felt that having an object would avoid users to perform this action by mistake
wdyt? should we change this?

apps/meteor/ee/server/models/raw/LivechatUnit.ts
```diff
@@ -109,28 +108,12 @@ export class LivechatUnitRaw extends BaseRaw<IOmnichannelBusinessUnit> implement
...
 		for await (const departmentId of departmentsToSave) {
-			await LivechatDepartment.updateOne(
-				{ _id: departmentId },
-				{
-					$set: {
-						parentId: _id,
-						ancestors,
-					},
-				},
-			);
```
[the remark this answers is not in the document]
matheusbsilva137: I'm using this same update in the callback I created, that's why I decided to put it in a function so that it can be used here and reused in my callback. Do you think I should rewrite this update in my callback instead?

apps/meteor/ee/app/livechat-enterprise/server/hooks/manageDepartmentUnit.ts
```diff
@@ -0,0 +1,51 @@
...
+		const units = await getUnitsFromUser(userId);
+		const isLivechatManager = await hasAnyRoleAsync(userId, ['admin', 'livechat-manager']);
```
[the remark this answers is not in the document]
matheusbsilva137: This is being considered in `getUnitsFromUser`. This only gets the units that the user "has access to"
I'll improve the variable name to make this clear

apps/meteor/app/livechat/server/lib/LivechatTyped.ts
```diff
@@ -1879,16 +1879,20 @@ class LivechatClass {
 	 * @param {string|null} _id - The department id
 	 * @param {Partial<import('@rocket.chat/core-typings').ILivechatDepartment>} departmentData
 	 * @param {{upsert?: { agentId: string; count?: number; order?: number; }[], remove?: { agentId: string; count?: number; order?: number; }}} [departmentAgents] - The department agents
+	 * @param {{_id?: string}} [departmentUnit] - The department's unit id
 	 */
 	async saveDepartment(
+		userId: string,
```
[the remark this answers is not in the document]
matheusbsilva137: Hmmm I believe we don't need to. It's already validated by the Meteor method that calls this function (this function is also called by an endpoint, but we don't need validations for this flow)

packages/rest-typings/src/v1/omnichannel.ts
```diff
@@ -575,7 +575,8 @@ type POSTLivechatDepartmentProps = {
 		chatClosingTags?: string[];
 		fallbackForwardDepartment?: string;
 	};
-	agents: { agentId: string; count?: number; order?: number }[];
+	agents?: { agentId: string; count?: number; order?: number }[];
```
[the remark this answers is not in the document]
matheusbsilva137: Yes, it was intended. I believe we created the type incorrectly
We use meteor check in this endpoint and the validation we have there allows this param to be undefined (check this [here](https://github.com/RocketChat/Rocket.Chat/blob/eb5f93c8dc16eba7682381dbbb7b0dd458ad6903/apps/meteor/app/livechat/imports/server/rest/departments.ts#L59))
Also, we already have a fallback for when this is undefined (check this [here](https://github.com/RocketChat/Rocket.Chat/blob/eb5f93c8dc16eba7682381dbbb7b0dd458ad6903/apps/meteor/app/livechat/imports/server/rest/departments.ts#L62))
wdyt?

apps/meteor/tests/end-to-end/api/livechat/14-units.ts
matheusbsilva137: I'll use the `createDepartment` helper we already have, instead (and just add some new params to it). We should just update this existing helper function at some point so that params are passed through an object
Please tell me if you think I should add a new helper instead

apps/meteor/tests/end-to-end/api/livechat/14-units.ts
matheusbsilva137: We only have one for the playwright/e2e tests. I believe it's better to have a new one for the API tests, right?
I'll improve this function and move it to the helpers

apps/meteor/tests/end-to-end/api/livechat/14-units.ts
matheusbsilva137: We'd need a new callback for validating this case (since the code we need for this validation is in EE) :grimacing: . That's why I decided to keep the association to a unit as an "additional" action and not throw any errors in this case
Do you think we should add a callback to implement this validation?

apps/meteor/app/livechat/imports/server/rest/departments.ts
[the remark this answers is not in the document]
MarcosSpessatto: No, I think your idea is fine, just wanted to understand it.

apps/meteor/ee/server/models/raw/LivechatUnit.ts
```diff
@@ -109,28 +108,12 @@ export class LivechatUnitRaw extends BaseRaw<IOmnichannelBusinessUnit> implement
...
 			if (!departmentsToSave.includes(departmentId)) {
-				await LivechatDepartment.updateOne(
-					{ _id: departmentId },
-					{
-						$set: {
-							parentId: null,
-							ancestors: null,
-						},
-					},
-				);
```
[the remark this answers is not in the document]
MarcosSpessatto: Didn't see that, in this case it's ok to keep it then.

apps/meteor/tests/end-to-end/api/livechat/14-units.ts
MarcosSpessatto: You can update the existing one if possible.

apps/meteor/tests/end-to-end/api/livechat/14-units.ts
MarcosSpessatto: Hmm, I don't think so, implementing a new CB just to validate it, doesn't look a good idea.

apps/meteor/ee/app/livechat-enterprise/server/hooks/manageDepartmentUnit.ts
matheusbsilva137: Done! That made me notice we were decreasing the departments count in units even when an invalid unit id was provided by an admin/manager. That's fixed :partying_face: (no action is performed when an invalid unit id is provided now)

apps/meteor/tests/end-to-end/api/livechat/14-units.ts
matheusbsilva137: Sorry, the tests here were actually using an admin (even though I wrote livechat managers in the description :man_facepalming: )
I fixed all the descriptions (to mention admin instead). Since admins and livechat managers behave exactly the same in this flow, is it ok to test with only one of the roles? Or should I repeat the tests for the managers too?

apps/meteor/tests/end-to-end/api/livechat/14-units.ts
matheusbsilva137: Sure, done! [Here](https://rocketchat.atlassian.net/browse/CORE-541) it is

MarcosSpessatto (dismissed):

KevLehman (dismissed):

KevLehman (dismissed):

KevLehman (approved):

MarcosSpessatto (approved):

PR #32693, opened 2024-06-27
feat: create contact endpoint

tapiarafael:

Implement the entity contact (collection `rocketchat_livechat_contact`) to the omnichannel and an endpoint to be able to create a contact manually.

[POST] `api/v1/omnichannel/contacts`
request example:
```json
{
	"name": "Contact 1",
	"emails": [
		"contact1@example.com"
	],
	"phones": [
		"5511999999999"
	],
}
```
response example:
```json
{
	"contactId": "668830239ac5cb4b15229d6d",
	"success": true
}
```

## Steps to test or reproduce

1. create a contact using the new endpoint mentioned above;
2. check on the collection;

## Further comments

[SCI-33](https://rocketchat.atlassian.net/browse/SCI-33)

[SCI-33]: https://rocketchat.atlassian.net/browse/SCI-33?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/meteor/tests/end-to-end/api/livechat/contacts.ts
```diff
@@ -0,0 +1,80 @@
...
+				.send({
+					name: faker.person.fullName(),
+					emails: [faker.internet.email().toLowerCase()],
+					phones: [faker.phone.number()],
+				});
+
+			expect(res.body).to.have.property('success', false);
+			expect(res.body.error).to.be.equal('User does not have the permissions required for this action [error-unauthorized]');
+
+			await restorePermissionToRoles('view-l-room');
```
MarcosSpessatto, line 43: Wouldn't it be better to have this in an `after` step?

apps/meteor/tests/end-to-end/api/livechat/contacts.ts
MarcosSpessatto, line 1: Can we create test cases for every input validation? e.g. when the user does not provide a name, when it's an invalid input, when no emails were provided, etc.

apps/meteor/app/livechat/server/api/v1/contact.ts
```diff
@@ -82,3 +86,15 @@ API.v1.addRoute(
 		},
 	},
 );
+
+API.v1.addRoute(
+	'omnichannel/contacts',
+	{ authRequired: true, permissionsRequired: ['view-l-room'], validateParams: isPOSTOmnichannelContactsProps },
+	{
+		async post() {
+			const contactId = await createContact({ ...this.bodyParams, unknown: false });
```
MarcosSpessatto, line 95: Curious why `unknown` is always false here? 🤔

MarcosSpessatto (changes requested):

apps/meteor/app/livechat/server/api/v1/contact.ts
tapiarafael: When a contact is manually created, it will always be known. But in the future, when a new visitor automatically creates a contact (by openning a conversation), it will be an unknown contact.

apps/meteor/app/livechat/server/lib/Contacts.ts
```diff
@@ -152,3 +171,71 @@ export const Contacts = {
...
 	},
 };
+
+export async function createContact(params: CreateContactParams): Promise<string> {
+	const { name, emails, phones, customFields = {}, contactManager, channels, unknown } = params;
+
+	if (contactManager) {
+		const contactManagerUser = await Users.findOneById(contactManager);
+		if (!contactManagerUser) {
+			throw new Meteor.Error('error-contact-manager-not-found', `No user found with id ${contactManager}`);
```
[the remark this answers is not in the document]
tapiarafael: Wouldn't it be better to use the `MeteorError` from `core-services` instead to keep the error message pattern? Or is just the `Error` OK?

apps/meteor/tests/end-to-end/api/livechat/contacts.ts
```diff
@@ -0,0 +1,277 @@
...
+			const res = await request
+				.post(api('omnichannel/contacts'))
+				.set(credentials)
+				.send({
+					name: faker.person.fullName(),
+					emails: [faker.internet.email().toLowerCase()],
+					phones: [faker.phone.number()],
+				});
+
+			expect(res.body).to.have.property('success', false);
```
Gustrb, line 52: does the permission get restored if the assertion fails??

apps/meteor/app/livechat/server/lib/Contacts.ts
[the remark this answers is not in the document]
MarcosSpessatto: I'd say to use Error only unless you really need the Meteor's error structure, if that's the case you can use MeteorError

apps/meteor/tests/end-to-end/api/livechat/contacts.ts
tapiarafael: We restore the permission on the `after` function

packages/rest-typings/src/v1/omnichannel.ts
```diff
@@ -1211,6 +1211,50 @@ const POSTOmnichannelContactSchema = {
...
+		customFields: {
+			type: 'object',
+			nullable: true,
+		},
+		contactManager: {
+			type: 'string',
+			nullable: true,
+		},
+	},
+	required: ['name', 'emails', 'phones'],
```
MarcosSpessatto, line 1252: Is `phones` required or nullable? 🤔

packages/rest-typings/src/v1/omnichannel.ts
```diff
@@ -1211,6 +1211,50 @@ const POSTOmnichannelContactSchema = {
...
+	name: string;
+	emails: string[];
+	phones: string[];
+	customFields?: Record<string, unknown>;
+	contactManager?: string;
+};
+
+const POSTOmnichannelContactsSchema = {
+	type: 'object',
+	properties: {
```
MarcosSpessatto, line 1224: Aren't we forgetting `channels` here? 🤔

packages/rest-typings/src/v1/omnichannel.ts
```diff
@@ -1211,6 +1211,50 @@ const POSTOmnichannelContactSchema = {
 
 export const isPOSTOmnichannelContactProps = ajv.compile<POSTOmnichannelContactProps>(POSTOmnichannelContactSchema);
 
+type POSTOmnichannelContactsProps = {
```
MarcosSpessatto, line 1214: Same as below

apps/meteor/tests/end-to-end/api/livechat/contacts.ts
```diff
@@ -0,0 +1,277 @@
...
+	before((done) => getCredentials(done));
+
+	before(async () => {
+		await updateSetting('Livechat_enabled', true);
+		await updatePermission('create-livechat-contact', ['admin']);
+	});
+
+	after(async () => {
+		await restorePermissionToRoles('create-livechat-contact');
+		await updateSetting('Livechat_enabled', false);
```
MarcosSpessatto, line 21: The default value for this one is `true`, isn't it?

apps/meteor/tests/end-to-end/api/livechat/contacts.ts
```diff
@@ -0,0 +1,277 @@
...
+					contactManager: normalUser._id,
+				});
+
+			expect(res.body).to.have.property('success', false);
+			expect(res.body).to.have.property('error');
+			expect(res.body.error).to.be.equal('error-invalid-contact-manager');
+		});
+
+		it('should be able to create a new contact with a contact manager', async () => {
+			const livechatAgent = await createAgent();
```
MarcosSpessatto, line 93: Can we create a new user for this instead of using the admin one? And after all delete that new user?

apps/meteor/app/livechat/server/lib/Contacts.ts
```diff
@@ -152,3 +171,71 @@ export const Contacts = {
...
+		contactManager,
+		channels,
+		customFields,
+		unknown,
+	});
+
+	return insertedId;
+}
+
+async function validateCustomFields(customFields: Record<string, string | unknown>) {
```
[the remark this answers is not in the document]
MarcosSpessatto: What about this one? Can we do that?

MarcosSpessatto (changes requested):

apps/meteor/tests/end-to-end/api/livechat/contacts.ts
```diff
@@ -0,0 +1,277 @@
...
+						},
+					});
+
+				expect(res.body).to.have.property('success', false);
+				expect(res.body).to.have.property('error');
+				expect(res.body.error).to.be.equal('Invalid value for Custom Field 1 field');
+			});
+		});
+
+		describe('Fields Validation', () => {
```
MarcosSpessatto, line 182: Can we add a test case trying to send additional properties?

MarcosSpessatto (changes requested):

packages/rest-typings/src/v1/omnichannel.ts
tapiarafael: Actually, no. The channels will be set automatically when we create a contact. We will set the channels based on visitor interaction, and the user cannot manually set this property when creating a contact.

packages/rest-typings/src/v1/omnichannel.ts
tapiarafael: Good catch 👍

apps/meteor/app/livechat/server/lib/Contacts.ts
```diff
@@ -152,3 +171,68 @@ export const Contacts = {
...
+
+async function getAllowedCustomFields(): Promise<ILivechatCustomField[]> {
+	return LivechatCustomField.findByScope(
+		'visitor',
+		{
+			projection: { _id: 1, label: 1, regexp: 1, required: 1 },
+		},
+		false,
+	).toArray();
+}
```
MarcosSpessatto, line 212: If we will not reuse this, we could use the model call directly, wdyt? 🤔

MarcosSpessatto (approved):

.changeset/sixty-spoons-own.md
```diff
@@ -0,0 +1,9 @@
+---
+"@rocket.chat/meteor": minor
+"@rocket.chat/core-typings": minor
+"@rocket.chat/model-typings": minor
+"@rocket.chat/models": minor
+"@rocket.chat/rest-typings": minor
+---
+
+Introduced "create contacts" endpoint to the omnichannel
```
KevLehman, line 9:
```suggestion
Introduced "create contacts" endpoint to omnichannel
```

I'd confirm with a PM but I'm pretty sure it's not "the omnichannel" but just "omnichannel"

apps/meteor/app/livechat/server/api/v1/contact.ts
```diff
@@ -1,14 +1,18 @@
...
-import { Contacts } from '../../lib/Contacts';
+import { Contacts, createContact } from '../../lib/Contacts';
 
+/**
+ * @deprecated to create a contact, use the omnichannel/contacts endpoint
+ */
 API.v1.addRoute(
 	'omnichannel/contact',
-	{ authRequired: true, permissionsRequired: ['view-l-room'] },
+	{ authRequired: true, permissionsRequired: ['view-l-room'], deprecationVersion: '8.0.0' },
```
KevLehman, line 15: The `deprecationVersion` form is "deprecated" lol

Please use the object form
`deprecation: { version: xxx, alternatives: [''] }`

apps/meteor/app/livechat/server/lib/Contacts.ts
```diff
@@ -152,3 +171,68 @@ export const Contacts = {
 		return contactId;
 	},
 };
+
+export async function createContact(params: CreateContactParams): Promise<string> {
+	const { name, emails, phones, customFields = {}, contactManager, channels, unknown } = params;
+
+	if (contactManager) {
+		const contactManagerUser = await Users.findOneById(contactManager, { projection: { roles: 1 } });
```
KevLehman, line 179:
```suggestion
		const contactManagerUser = await Users.findOneById<Pick<IUser, 'roles'>>(contactManager, { projection: { roles: 1 } });
```

apps/meteor/app/livechat/server/lib/Contacts.ts
```diff
@@ -152,3 +171,68 @@ export const Contacts = {
...
+		}
+		if (!contactManagerUser.roles || !Array.isArray(contactManagerUser.roles) || !contactManagerUser.roles.includes('livechat-agent')) {
+			throw new Error('error-invalid-contact-manager');
+		}
+	}
+
+	const allowedCustomFields = await getAllowedCustomFields();
+	validateCustomFields(allowedCustomFields, customFields);
+
+	const { insertedId } = await LivechatContacts.insertOne({
```
KevLehman, line 191: IF this will be the "new" way of adding visitors/contacts, shouldn't we put a note on the visitors model so people is aware? 👀 (not sure if that's gonna happen on another PR)

KevLehman (changes requested):

apps/meteor/app/livechat/server/lib/Contacts.ts
tapiarafael: We're using this on the update endpoint. #32729

apps/meteor/app/livechat/server/lib/Contacts.ts
tapiarafael: Visitors and contacts are different things. Visitors will be created the same way they're today. This introduces the capability of creating the new entity called Contact

MarcosSpessatto (dismissed):

apps/meteor/app/livechat/server/lib/Contacts.ts
KevLehman: Interesting 🤔 what's the diff between them?

apps/meteor/app/livechat/server/lib/Contacts.ts
```diff
@@ -152,3 +171,68 @@ export const Contacts = {
...
+	const { name, emails, phones, customFields = {}, contactManager, channels, unknown } = params;
+
+	if (contactManager) {
+		const contactManagerUser = await Users.findOneById<Pick<IUser, 'roles'>>(contactManager, { projection: { roles: 1 } });
+		if (!contactManagerUser) {
+			throw new Error('error-contact-manager-not-found');
+		}
+		if (!contactManagerUser.roles || !Array.isArray(contactManagerUser.roles) || !contactManagerUser.roles.includes('livechat-agent')) {
+			throw new Error('error-invalid-contact-manager');
+		}
```
KevLehman, line 185: You can use `users.findOneAgentById` and remove the manual `livechat-agent` check here

KevLehman (changes requested):

apps/meteor/app/livechat/server/lib/Contacts.ts
tapiarafael: Contact is a new entity being introduced by the Single Contact ID project. The main idea is to have one entity that centralizes all the information of all channels that a customer can contact. It will "group" a bunch of visitors who represent the same person and allow the workspace to verify the identity of this contact. We can schedule a meeting with @renatobecker  if you want to know more about the project and its objectives.

KevLehman (dismissed):

MarcosSpessatto (approved):

Gustrb (dismissed):

ggazzo (approved):

tassoevan (approved):

apps/meteor/app/livechat/server/api/v1/contact.ts
```diff
@@ -1,14 +1,22 @@
...
+/**
+ * @deprecated to create a contact, use the omnichannel/contacts endpoint
+ */
 API.v1.addRoute(
 	'omnichannel/contact',
-	{ authRequired: true, permissionsRequired: ['view-l-room'] },
+	{
+		authRequired: true,
+		permissionsRequired: ['view-l-room'],
+		deprecation: { version: '8.0.0', alternatives: ['omnichannel/contacts'] },
```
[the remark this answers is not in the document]
tapiarafael: So, what are your suggestions here? Should we remove the deprecation warning and let both endpoints?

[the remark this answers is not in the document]
ggazzo: I would remove the message and leave this new endpoint only inside TEST_MODE

ggazzo (approved):

KevLehman (approved):

PR #32696, opened 2024-06-28
regression: Fix new user panel memory exceeded error

rique223:

Added allowDiskUse option to the users panel query which should solve situations in which the number of users of a workspace is so big that it exceeds the 100MB MongoDb memory threshold;

Jira task: [CONN-277](https://rocketchat.atlassian.net/browse/CONN-277)
## Further comments

[CONN-277]: https://rocketchat.atlassian.net/browse/CONN-277?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.changeset/dull-knives-complain.md
```diff
@@ -0,0 +1,5 @@
+---
+"@rocket.chat/meteor": patch
```
ggazzo, line 2: regressions dont need changesets

ggazzo (changes requested):

.changeset/dull-knives-complain.md
rique223: Got it. Will remove it. But looks like we have a documentation problem then:

tassoevan (approved):

PR #32707, opened 2024-07-01
fix: Agents can't leave omnichannel rooms that have already been closed

matheusbsilva137:

 - Allow users to leave omnichannel rooms after they have been closed (by the visitor). If a user tries to close a room that has already been closed AND they have a subscription in this room, we'll remove them from the room instead of throwing an error;
 - Display button to close livechat rooms when the user is still subscribed to the room and has any of the permissions (`close-livechat-room` or `close-others-livechat-room`) to close a room.

Due to some **race condition**, users could end up joining omnichannel rooms that have already been closed, which caused them to get stuck in there.

## Steps to test or reproduce

Since having a user stuck in an omnichannel room after it's been closed depends on a race condition, reproducing this issue may require manually inserting a subscription in the DB.
With a user that is stuck in an omnichannel/livechat room, try to close the room again through the UI or using the `livechat.closeRoomByUser` endpoint OR the `livechat:closeRoom` method. Both options currently throw an error in this case, but with this PR, the subscription is removed and a succesful result should be returned.

## Further comments

I decided not to add this change to the `livechat.closeRoom` endpoint since this one is meant to be used by visitors -- which are not subscribed to the room directly, only agents are.

[SUP-612](https://rocketchat.atlassian.net/browse/SUP-612)

[SUP-612]: https://rocketchat.atlassian.net/browse/SUP-612?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ggazzo:
Fix?

KevLehman:
Yep, should be fix.

Also, pls add a description of what this fixes (+ changeset). It should be clear which case of rooms getting stuck this fix clears 🤗

apps/meteor/tests/end-to-end/api/livechat/00-rooms.ts
```diff
@@ -64,6 +66,32 @@ const getSubscriptionForRoom = async (roomId: string, overrideCredential?: Crede
...
+			username: 'username',
+			name: 'name',
+		},
+	};
+	const connection = await MongoClient.connect(URL_MONGODB);
+
+	await connection.db().collection('rocketchat_subscription').insertOne(subscription);
+	await connection.close();
+};
+
```
MarcosSpessatto, line 94: Let's not open this possibility to API tests, I'm concerned if we do this, we will start to insert things directly into the DB to make it easier to test, I'd say we should try to reproduce through the API only, if not possible we could create a unit test instead of open a DB connection here.

MarcosSpessatto (changes requested):

.changeset/happy-peaches-nail.md
```diff
@@ -0,0 +1,5 @@
+---
+"@rocket.chat/meteor": patch
+---
+
+Fixed issue with livechat agents not being able to leave omnichannel rooms if joining after a room has been closed (due to race conditions)
```
KevLehman, line 5: You probably have to put here that if you get stuck the first time calling, then calling a 2nd time should fix the data

matheusbsilva137: I believe that's not exacty the case for this fix. From the reprorts we got, this only happens when the room has been closed by the visitor, not when the agent itself had already called it once before
I'll make this clear in the changeset

apps/meteor/app/livechat/server/methods/closeRoom.ts
```diff
@@ -60,6 +60,16 @@ Meteor.methods<ServerMethods>({
...
 
+		const subscription = await SubscriptionRaw.findOneByRoomIdAndUserId(roomId, userId, {
+			projection: {
+				_id: 1,
+			},
+		});
+		if (!room.open && subscription) {
+			await SubscriptionRaw.removeByRoomId(roomId);
+			return;
+		}
```
debdutdeb, line 71: this can be a single db call, removing find call, just removing in the if.

apps/meteor/tests/end-to-end/api/livechat/00-rooms.ts
matheusbsilva137: Hmmm it's not possible to reproduce this only with the API since dangling subscriptions are only generated by race conditions (and we don't have any endpoint that would allow us to add a subscription to a closed room) -- or at least this could turn into a flaky test if we try to reproduce the race condition.

I thought of unit tests at first, but there seemed to be so much to  mock that made me believe that wouldn't be the right path. But I could try this out now

matheusbsilva137: Done! I've just replaced the end-to-end tests by unit tests

MarcosSpessatto (approved):

KevLehman (approved):

apps/meteor/app/lib/server/functions/closeLivechatRoom.ts
```diff
@@ -0,0 +1,81 @@
...
+					requestData: Pick<NonNullable<IOmnichannelRoom['transcriptRequest']>, 'email' | 'subject'>;
+			  };
+	},
+): Promise<void> => {
+	const room = await LivechatRooms.findOneById(roomId);
+	if (!room || !isOmnichannelRoom(room)) {
+		throw new Error('error-invalid-room');
+	}
+
+	const subscription = await Subscriptions.findOneByRoomIdAndUserId(roomId, user._id, { projection: { _id: 1 } });
```
KevLehman, line 36: @matheusbsilva137 i was thinking: what if you're not the only one in the room? Rooms can be served just by one people but other users/managers can join to those.

Which means there's a possibility of a user "closing" the room but one of the subs (maybe not yours) is left behind. In that case this fix won't remove the sub for that other user because we're only checking if the caller has a sub. 

What if instead of checking like this, we count the subs and if > 0 then we clear them again? does it makes sense?

KevLehman (dismissed):

apps/meteor/app/lib/server/functions/closeLivechatRoom.ts
matheusbsilva137: Thanks for suggesting this. Done!

KevLehman (approved):

MartinSchoeler (approved):

PR #32711, opened 2024-07-02
test: Copy new EE test license from CI to unit tests

d-gubert:

Copy the new EE test license from CI to the licnese package unit tests


ggazzo:
What about using the env var and the hardcoded one as fallback?

d-gubert:
> What about using the env var and the hardcoded one as fallback?

did

rodrigok (dismissed):

tassoevan (approved):

PR #32712, opened 2024-07-02
regression: E2EE DM setup header toolbox items

yash-rajpal:

The E2EE room setup header was showing irrelevant toolbox items on DM rooms which were not applicable for DMs.
## Further comments

[E2EE2-45](https://rocketchat.atlassian.net/browse/E2EE2-45)

[E2EE2-45]: https://rocketchat.atlassian.net/browse/E2EE2-45?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/meteor/client/views/room/Header/RoomToolbox/RoomToolboxE2EESetup.tsx
```diff
@@ -4,19 +4,23 @@ import { useTranslation } from '@rocket.chat/ui-contexts';
...
 
 	const { tab } = toolbox;
 
 	const actions = useStableArray(
 		roomActionHooksForE2EESetup
 			.map((roomActionHook) => roomActionHook())
-			.filter((roomAction): roomAction is RoomToolboxActionConfig => !!roomAction),
+			.filter((roomAction): roomAction is RoomToolboxActionConfig => !!roomAction)
+			.filter((action) => !action.groups || action.groups.includes(getRoomGroup(room))),
 	);
```
gabriellsh, line 24: Any chance these filters could be done in a single `filter` call?

gabriellsh (changes requested):
The tests you wrote don't actually test the issue that is being fixed. It is a nice addition, but we need an end-to-end test for this, or:
- Extract the header items filter logic to a hook and unit test that hook
- Unit test `RoomToolboxE2EESetup` component

One of these 3 should be enough for this

apps/meteor/client/views/room/Header/RoomToolbox/RoomToolboxE2EESetup.tsx
yash-rajpal: Sure, I just did this way to improve readability

yash-rajpal: Done!

apps/meteor/tests/e2e/e2e-encryption.spec.ts
```diff
@@ -654,6 +654,68 @@ test.describe.serial('e2ee room setup', () => {
...
+		await poAccountProfile.securityE2EEncryptionResetKeyButton.click();
+
+		await page.locator('role=button[name="Login"]').waitFor();
+
+		await page.reload();
+
+		await page.locator('role=button[name="Login"]').waitFor();
+
+		await injectInitialData();
+		await restoreState(page, Users.admin);
```
gabriellsh, line 669: I'm a little lost here, why the `waitFor` calls and the page reload? Won't restoreState login with the admin token?

apps/meteor/tests/e2e/e2e-encryption.spec.ts
```diff
@@ -654,6 +654,68 @@ test.describe.serial('e2ee room setup', () => {
...
+
+		await page.goto('/home');
+
+		await page.locator('role=banner >> text="Save your encryption password"').waitFor();
+		await expect(page.locator('role=banner >> text="Save your encryption password"')).toBeVisible();
+
+		await poHomeChannel.sidenav.openNewByLabel('Direct message');
+		await poHomeChannel.sidenav.inputDirectUsername.click();
+		await page.keyboard.type('user2');
+		await page.waitForTimeout(1000);
```
gabriellsh, line 679: Instead of waitForTimeout cant we wait for the locator with the username to be found?

apps/meteor/tests/e2e/e2e-encryption.spec.ts
```diff
@@ -654,6 +654,68 @@ test.describe.serial('e2ee room setup', () => {
...
+		await page.keyboard.press('Enter');
+		await poHomeChannel.sidenav.btnCreate.click();
+
+		await expect(page).toHaveURL(`/direct/rocketchat.internal.admin.testuser2`);
+
+		await poHomeChannel.tabs.kebab.click({ force: true });
```
gabriellsh, line 685: Why force: true?

apps/meteor/tests/e2e/e2e-encryption.spec.ts
```diff
@@ -654,6 +654,68 @@ test.describe.serial('e2ee room setup', () => {
...
+		await expect(poHomeChannel.tabs.btnEnableE2E).toBeVisible();
+		await poHomeChannel.tabs.btnEnableE2E.click({ force: true });
```
gabriellsh, line 687: here too...

apps/meteor/tests/e2e/e2e-encryption.spec.ts
```diff
@@ -654,6 +654,68 @@ test.describe.serial('e2ee room setup', () => {
...
+
+		await poHomeChannel.content.encryptedRoomHeaderIcon.first().waitFor();
```
gabriellsh, line 689: Is there any other `encryptedRoomHeaderIcon` that could be visible at this point? If yes, won't that impact the tests? Because you're using `waitFor` after `first()`, so it could get any of the elements that show up first.

apps/meteor/tests/e2e/e2e-encryption.spec.ts
```diff
@@ -654,6 +654,68 @@ test.describe.serial('e2ee room setup', () => {
...
+		await expect(poHomeChannel.content.encryptedRoomHeaderIcon.first()).toBeVisible();
```
gabriellsh, line 690: Also this seems to be a duplicate locator, because on the line above we're already waiting for the selector to be found. You could do the expect on the same line, or save the locator in a variable for it to be used on this expect.

apps/meteor/tests/e2e/e2e-encryption.spec.ts
```diff
@@ -654,6 +654,68 @@ test.describe.serial('e2ee room setup', () => {
...
+
+		await page.locator('role=button[name="Save E2EE password"]').waitFor();
+		await expect(page.locator('role=button[name="Save E2EE password"]')).toBeVisible();
```
gabriellsh, line 693: This is the same case, 2 locators for the same use.

apps/meteor/tests/e2e/e2e-encryption.spec.ts
```diff
@@ -654,6 +654,68 @@ test.describe.serial('e2ee room setup', () => {
...
+
+		await poHomeChannel.tabs.btnE2EERoomSetupDisableE2E.waitFor();
+		await expect(poHomeChannel.tabs.btnE2EERoomSetupDisableE2E).toBeVisible();
```
gabriellsh, line 696: Here once again, same locator being made twice.

apps/meteor/tests/e2e/e2e-encryption.spec.ts
```diff
@@ -654,6 +654,68 @@ test.describe.serial('e2ee room setup', () => {
...
+		await expect(poHomeChannel.tabs.btnTabMembers).not.toBeVisible();
+		await expect(poHomeChannel.tabs.btnRoomInfo).not.toBeVisible();
+
+		await expect(poHomeChannel.content.inputMessage).not.toBeVisible();
+
+		await page.locator('role=button[name="Save E2EE password"]').click();
```
gabriellsh, line 702: This could be the same hypothetical variable I mentioned on comment on line 692. You already had 2 locators for it, maybe it's better to save it in a const so you only need to find the element once

apps/meteor/tests/e2e/e2e-encryption.spec.ts
```diff
@@ -654,6 +654,68 @@ test.describe.serial('e2ee room setup', () => {
...
+
+		e2eePassword = (await page.evaluate(() => localStorage.getItem('e2e.randomPassword'))) || 'undefined';
+
+		await expect(page.locator('role=dialog[name="Save your encryption password"]')).toBeVisible();
+		await expect(page.locator('#modal-root')).toContainText(e2eePassword);
```
gabriellsh, line 707: I think it's not correct for us to use `#modal-root` here. You should do the same in line 706 with role=dialog, and you can filter the elements further from it.

Getting the e2e password from localstorage is also a little weird, but I get it, so I think it's fine for now at least.

apps/meteor/tests/e2e/e2e-encryption.spec.ts
```diff
@@ -654,6 +654,68 @@ test.describe.serial('e2ee room setup', () => {
...
+
+		await page.locator('#modal-root >> button:has-text("I saved my password")').click();
```
gabriellsh, line 709: Here we could use role=dialog aswell I believe.

gabriellsh (changes requested):
Left some comments. Overall I'm wondering on the complexity of these tests, if it makes sense to implement all those checks on the same test evaluation.

In general I'd like to see more atomic tests. For ex: in this case you're testing the header buttons visibility, but there's a lot of other components of the feature being tested. I like to see that you included tests for the whole UX for the save password state, but it could be a little more organized.

What I mean is, a describe for this whole flow, and then a test clause for each behaviour you're testing.

Please let me know if you wanna discuss any of these topics or if you disagree with anything. All feedback is welcome.

apps/meteor/tests/e2e/e2e-encryption.spec.ts
yash-rajpal: Getting the e2ee password from local storage is just to ensure that the password is being saved correctly on localstorage and is also shown correctly to the user.

apps/meteor/tests/e2e/e2e-encryption.spec.ts
yash-rajpal: Makes sense.

apps/meteor/tests/e2e/e2e-encryption.spec.ts
yash-rajpal: Page reload was due to an issue, where the websocket connection was dying on logout and wasn't working as expected on logging back again without a reload.

This issue was recently solved, I'll remove this page reload call.

apps/meteor/tests/e2e/e2e-encryption.spec.ts
yash-rajpal: Will it be made twice in this case 🤔 

As, it is using the same instance of the class? 🤔 Not sure.

apps/meteor/tests/e2e/e2e-encryption.spec.ts
yash-rajpal: This was necessary sometime back, but works now without it. Will remove it.

yash-rajpal:
Implemented a E2E test to cover this case, but for some reason the test is always failing on CI, so removed the test from this PR, will implement the test in a separate PR and it would be tracked by [this task](https://rocketchat.atlassian.net/browse/CORE-543)

gabriellsh (approved):

PR #32715, opened 2024-07-02
regression: Edit encrypted file description not working

rodrigok:


yash-rajpal (approved):
Would be great if we can have some E2E tests for editing e2ee messages flow, but LGTM without it.

PR #32735, opened 2024-07-07
chore: add `eslint:fix:branch` to run only on files changed from develop

debdutdeb:


tassoevan:
Isn't it going to hurt rules from `@typescript-eslint`? Changing in one module can break a rule in another one unmodified. For instance, the change

```diff
- function doTask() {
+ async function doTask() {
  ...
}

export { doTask };
```

would break `@typescript-eslint/no-floating-promises` in other non-linted modules importing `doTask`.

PR #32739, opened 2024-07-08
fix: `sendFCM` function not using `@rocket.chat/server-fetch`

debdutdeb:
Use legacy notification provider setting is disabled in Workspace > Settings > Push, push notifications stop going to Android devices. PR Should fix this situation for using a proxy for it

## Issue(s)
https://rocketchat.atlassian.net/browse/SUP-616


Gustrb (approved):

apps/meteor/app/push/server/fcm.ts
```diff
@@ -54,6 +54,8 @@ type FCMError = {
 	};
 };
 
+type FetchOptions = NonNullable<Parameters<typeof fetch>[1]>;
```
KevLehman, line 57: IF this is the same type of FetchOptions the package receives, I feel we can export this one from the package itself and import here (as it may be used later)

debdutdeb: image

debdutdeb: Done

KevLehman (dismissed):

apps/meteor/app/push/server/fcm.ts
```diff
@@ -1,6 +1,6 @@
+import { serverFetch as fetch, ExtendedFetchOptions } from '@rocket.chat/server-fetch';
```
Gustrb, line 1: hey, could you split the type and the function imports??

debdutdeb:
```suggestion
import { serverFetch as fetch, type ExtendedFetchOptions } from '@rocket.chat/server-fetch';
```

Gustrb (approved):

pierre-lehnen-rc (approved):

PR #32742, opened 2024-07-08
fix: update cached setting immediately at the time of updating the db

debdutdeb:
## Issue(s)
https://rocketchat.atlassian.net/browse/CORE-545


apps/meteor/ee/tests/unit/server/settings/SettingsRegistry.spec.ts
KevLehman, line 1: settingsregistry is ce but your tests are on ee 🤔 expected?

debdutdeb: ouch no :p

apps/meteor/app/settings/server/SettingsRegistry.ts
```diff
@@ -73,7 +74,8 @@ const compareSettingsIgnoringKeys =
 			.filter((key) => !keys.includes(key as keyof ISetting))
 			.every((key) => isEqual(a[key as keyof ISetting], b[key as keyof ISetting]));
 
-const compareSettings = compareSettingsIgnoringKeys([
+// NOTE(debdut): not really "metadata" - but yeah. I don't have a better name
```
ggazzo, line 77: I don't think this is in any way relevant or useful to the codebase

apps/meteor/tests/unit/app/settings/server/functions/settings.tests.ts
```diff
@@ -3,20 +3,32 @@ import { beforeEach, describe, it } from 'mocha';
...
 		Settings.upsertCalls = 0;
+		Settings.data.clear();
 		process.env = {};
 	});
 
 	it('should not insert the same setting twice', async () => {
-		const settings = new CachedSettings();
-		Settings.settings = settings;
-		settings.initialized();
-		const settingsRegistry = new SettingsRegistry({ store: settings, model: Settings as any });
```
ggazzo, line 19: Let's keep each test with its instance? to ensure it does exactly what it's supposed to do without interference? you can also test the clear cases, but as a specific case

apps/meteor/app/settings/server/SettingsRegistry.ts
```diff
@@ -73,7 +74,8 @@ const compareSettingsIgnoringKeys =
...
+export const compareSettingsMetadata = compareSettingsIgnoringKeys([
```
ggazzo, line 78: keeping the old name would not interfere in the final result

ggazzo (changes requested):
IMO too many modifications to a simple regression. the fix/regression has by far more changes than the original fix.

To be honest, I can't understand exactly what error is being corrected. I didn't find explanations in the task or in the pull request. It seems to me that the original correction was not so correct?

The only thing that catches my attention is the 
```
const settingFromCodeOverwritten = overwriteSetting(settingFromCode);
const settingOverwrittenDefault = overrideSetting(settingFromCode);
```
and then
```
const updatedSettingAfterApplyingOverwritten = isOverwritten ? settingFromCodeOverwritten : settingOverwrittenDefault;
```

In my idea, both should have the same value

I would like to see a revert of what was done and the same correction again on top of the original code, this way the original code is completely modified, to the point that I no longer remember that we wanted to update the cache at the time of the overwrite

KevLehman (dismissed):

apps/meteor/app/settings/server/SettingsRegistry.ts
```diff
@@ -198,9 +213,13 @@ export class SettingsRegistry {
 			return;
 		}
 
-		await this.model.insertOne(updatedSettingAfterApplyingOverwrite); // no need to emit unless we remove the oplog
+		const settingOverwrittenDefault = overrideSetting(settingFromCode);
+
+		const setting = isOverwritten ? settingFromCodeOverwritten : settingOverwrittenDefault;
+
+		await this.model.insertOne(setting); // no need to emit unless we remove the oplog
```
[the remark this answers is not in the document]
debdutdeb: Apparently not, checked a while ago. I still left the comment for evidence.

apps/meteor/app/settings/server/SettingsRegistry.ts
```diff
@@ -138,53 +140,60 @@ export class SettingsRegistry {
...
+					settingStored.value !== settingStoredOverwritten.value && { value: settingStoredOverwritten.value }),
+			};
 
-			await this.saveUpdatedSetting(_id, updatedProps, removedKeys);
+			const { value: settingForCache } = await this.saveUpdatedSetting(_id, updatedProps, removedKeys);
 
-			this.store.set(updatedSettingAfterApplyingOverwrite);
+			if (!settingForCache) {
+				// unreachable
+				throw new Error('No document returned after setting was updated due to metadata change, something is wrong with code');
```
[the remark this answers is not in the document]
debdutdeb: suggestion? 🙈  this is unreachable, so just tried to be verbose

apps/meteor/app/settings/server/SettingsRegistry.ts
```diff
@@ -73,7 +74,7 @@ const compareSettingsIgnoringKeys =
 			.filter((key) => !keys.includes(key as keyof ISetting))
 			.every((key) => isEqual(a[key as keyof ISetting], b[key as keyof ISetting]));
 
-const compareSettings = compareSettingsIgnoringKeys([
+export const compareSettingsMetadata = compareSettingsIgnoringKeys([
```
debdutdeb, line 77:
```suggestion
export const compareSettings = compareSettingsIgnoringKeys([
```

apps/meteor/app/settings/server/SettingsRegistry.ts
```diff
@@ -138,53 +139,65 @@ export class SettingsRegistry {
...
 			validateSetting(settingFromCode._id, settingFromCode.type, settingFromCode.value);
 		} catch (e) {
 			IS_DEVELOPMENT && SystemLogger.error(`Invalid setting code ${_id}: ${(e as Error).message}`);
 		}
 
-		const { _id: _, ...settingProps } = settingFromCodeOverwritten;
+		const isOverwritten = settingFromCode !== settingFromCodeOverwritten || (settingStored && settingStored !== settingStoredOverwritten);
+
+		if (settingStored && !compareSettingsMetadata(settingStored, settingFromCodeOverwritten)) {
+			// we need to update the setting metadata here
```
debdutdeb, line 155:
```suggestion
		if (settingStored && !compareSettings(settingStored, settingFromCodeOverwritten)) {
```

apps/meteor/tests/unit/app/settings/server/functions/compareSettingsMetadata.tests.ts
```diff
@@ -0,0 +1,28 @@
...
+			const copiedSetting: any = { ...testSetting };
+
+			if (['ts', 'createdAt', '_updatedAt'].includes(key)) {
+				copiedSetting[key] = new Date();
+			} else {
+				copiedSetting[key] = 'random';
+			}
+
+			expect(compareSettingsMetadata(testSetting, copiedSetting)).to.be.true;
+		}),
```
debdutdeb, line 26:
```suggestion
import { compareSettings } from '../../../../../../app/settings/server/SettingsRegistry';
import { getSettingDefaults } from '../../../../../../app/settings/server/functions/getSettingDefaults';

const testSetting = getSettingDefaults({
	_id: 'my_dummy_setting',
	type: 'string',
	value: 'dummy',
});

describe('#compareSettings', () => {
	const ignoredKeys = ['value', 'ts', 'createdAt', 'valueSource', 'packageValue', 'processEnvValue', '_updatedAt'];

	ignoredKeys.forEach((key) =>
		it(`should return true if ${key} changes`, () => {
			const copiedSetting: any = { ...testSetting };

			if (['ts', 'createdAt', '_updatedAt'].includes(key)) {
				copiedSetting[key] = new Date();
			} else {
				copiedSetting[key] = 'random';
			}

			expect(compareSettings(testSetting, copiedSetting)).to.be.true;
		}),
```

debdutdeb:
@ggazzo @KevLehman https://github.com/RocketChat/Rocket.Chat/pull/32814

apps/meteor/app/settings/server/SettingsRegistry.ts
```diff
@@ -149,32 +150,51 @@ export class SettingsRegistry {
 
 		const isOverwritten = settingFromCode !== settingFromCodeOverwritten || (settingStored && settingStored !== settingStoredOverwritten);
 
-		const { _id: _, ...settingProps } = settingFromCodeOverwritten;
-
 		if (settingStored && !compareSettings(settingStored, settingFromCodeOverwritten)) {
-			const { value: _value, ...settingOverwrittenProps } = settingFromCodeOverwritten;
+			const { value: _value, ...settingOverwrittenProps } = settingFromCodeOverwritten; // settingStoredOverwrite not used since we need the updated props, that is in code not db
```
debdutdeb, line 154:
```suggestion
			const { value: _value, ...settingOverwrittenProps } = settingFromCodeOverwritten;
```

ggazzo:
I ran your new tests with the old code, and they passed. This makes me believe that we don't have a test covering exactly what you are fixing

ggazzo (approved):

PR #32743, opened 2024-07-08
fix: validate username before registering user

ricardogarim:
As per [CORE-544](https://rocketchat.atlassian.net/browse/CORE-544) and as discussed internally at [#rocketchat-dev](https://open.rocket.chat/channel/rocketchat-dev?msg=zGLpHpcKHLAJ65WBL), creating a new user with an invalid username (containing special characters) results in an error message, but the user is still created. This leads to an error when attempting to register again using the same email, as it is already registered.

To address this issue, I introduced a new method to validate usernames before registering users, ensuring that invalid usernames prevent user creation.

https://github.com/RocketChat/Rocket.Chat/assets/9621276/c6fa6408-bde5-4cf4-9c25-701b47d9a796

[CORE-544]: https://rocketchat.atlassian.net/browse/CORE-544?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/meteor/app/lib/server/functions/validateUsername.ts
```diff
@@ -0,0 +1,9 @@
+import { settings } from '../../../settings/server';
+
+export const validateUsername = (username: string): boolean => {
+	const settingsRegExp = settings.get('UTF8_User_Names_Validation');
+
+	const usernameRegExp = settingsRegExp ? new RegExp(`^${settingsRegExp}$`) : new RegExp('^[0-9a-zA-Z-_.]+$');
```
MarcosSpessatto, line 6: I think an invalid regex could throw an error here, right?

apps/meteor/app/lib/server/functions/validateUsername.ts
```diff
@@ -0,0 +1,9 @@
+import { settings } from '../../../settings/server';
+
+export const validateUsername = (username: string): boolean => {
```
MarcosSpessatto, line 3: Can we create a unit test for this?

MarcosSpessatto (changes requested):
I think there are more places that require this change to be applied (or at least reviewed)...

https://github.com/RocketChat/Rocket.Chat/blob/develop/apps/meteor/server/methods/registerUser.ts#L31 is one of them, but I think we should revisit all the places where users are being created, like LDAP, SAML, admin panel, etc.

apps/meteor/tests/unit/app/lib/server/functions/validateUsername.spec.ts
```diff
@@ -0,0 +1,95 @@
...
+import proxyquire from 'proxyquire';
+
+const proxySettings = {
+	settings: {
+		get: (key: string): string | null => {
+			if (key === 'UTF8_User_Names_Validation') {
+				// Default value set at apps/meteor/server/settings/general.ts
+				return '[0-9a-zA-Z-_.]+';
+			}
+			return null;
```
KevLehman, line 11: You can declare a stub for this (with sinon) and then use `stub.withArgs('UTF...').returns('')`

apps/meteor/tests/unit/app/lib/server/functions/validateUsername.spec.ts
```diff
@@ -0,0 +1,95 @@
...
+
+		it('should return true for a username with allowed special characters', () => {
+			const result = validateUsername('username-_.');
+			expect(result).to.be.true;
+		});
+	});
+
+	describe('with custom regex settings', () => {
+		beforeEach(() => {
+			proxySettings.settings.get = (key: string) => {
```
KevLehman, line 50: Use an stub

KevLehman (changes requested):
add a changeset

KevLehman (approved):

MarcosSpessatto (approved):

PR #32752, opened 2024-07-10
feat(Omnichannel): System messages in transcripts

yash-rajpal:

Currently we don't handle system messages in our transcripts, the system messages are shown as empty space in our transcripts.

This PR adds this support to handle system generated messages for both, email and PDF transcripts.

image
 
## Further comments

[CORE-494](https://rocketchat.atlassian.net/browse/CORE-494)

[CORE-494]: https://rocketchat.atlassian.net/browse/CORE-494?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/meteor/server/settings/omnichannel.ts
```diff
@@ -412,6 +412,13 @@ export const createOmniSettings = () =>
 			enableQuery: omnichannelEnabledQuery,
 		});
 
+		await this.add('Livechat_transcript_show_system_messages', false, {
```
KevLehman, line 415: Does this apply to both transcripts? (Email, pdf)

apps/meteor/server/models/raw/Messages.ts
```diff
@@ -424,14 +429,25 @@ export class MessagesRaw extends BaseRaw<IMessage> implements IMessagesModel {
...
 
-	findLivechatMessagesWithoutClosing(rid: IRoom['_id'], options?: FindOptions<IMessage>): FindCursor<IMessage> {
-		return this.find(
-			{
-				rid,
-				t: { $exists: false },
-			},
-			options,
-		);
+	findLivechatMessagesWithoutTypes(
```
KevLehman, line 432: Instead of creating a new func, can we reuse the `findVisibleByRoomIdNotContainingTypesBeforeTs` one?

KevLehman, line 432: Instead of creating a new func, can we reuse the `findVisibleByRoomIdNotContainingTypesBeforeTs` one?

ee/packages/pdf-worker/src/strategies/ChatTranscript.ts
```diff
@@ -10,8 +12,41 @@ export class ChatTranscript implements IStrategy {
...
 	}
 
-	private parserMessages(messages: PDFMessage[], dateFormat: string, timeAndDateFormat: string, timezone: string): PDFMessage[] {
-		return messages.map((message, index, arr) => {
+	private async parseSystemMessage(messages: PDFMessage[]): Promise<PDFMessage[]> {
+		return Promise.all(
+			messages.map(async (message: PDFMessage) => {
+				if (!message.t) {
+					return message;
+				}
```
KevLehman, line 20:
```suggestion
			messages.filter((message) => !!message.t).map(async (message: PDFMessage) => {
```
Filter outside the async to reduce promises created

ee/packages/pdf-worker/src/strategies/ChatTranscript.ts
```diff
@@ -10,8 +12,41 @@ export class ChatTranscript implements IStrategy {
...
+
+				const systemMessageDefinition = getSystemMessage(message.t);
+
+				if (!systemMessageDefinition) {
+					return message;
+				}
```
KevLehman, line 26: Same here, you can filter the ones we have definitions for and avoid pushing a promise that will return inmediately (for cases where systemMessageDefinition is undefined)

ee/packages/pdf-worker/src/strategies/ChatTranscript.ts
```diff
@@ -1,5 +1,7 @@
+import { Translation as translationService } from '@rocket.chat/core-services';
```
KevLehman, line 1: since this is a "pdf" package, i wonder if we should do the system message parsing on the omnichannel-transcript service instead of inside pdf-worker package.

ee/packages/pdf-worker/src/strategies/ChatTranscript.ts
```diff
@@ -86,7 +121,7 @@ export class ChatTranscript implements IStrategy {
 					.format('H:mm:ss')} ${data.timezone}`,
 			},
 			messages: Array.isArray(data.messages)
-				? this.parserMessages(data.messages, data.dateFormat as string, data.timeAndDateFormat as string, data.timezone as string)
+				? await this.parserMessages(data.messages, data.dateFormat as string, data.timeAndDateFormat as string, data.timezone as string)
```
KevLehman, line 124: as this is now async and the number of messages in a conversation could be really big, i wonder if we should start batching these 🤔 

wdyt @MarcosSpessatto ?

KevLehman (changes requested):

apps/meteor/server/settings/omnichannel.ts
yash-rajpal: Yes.

apps/meteor/server/models/raw/Messages.ts
yash-rajpal: I was thinking the same thing, but both functions have a small difference related to getting the closing message and fetching the messages before the timestamp of closing message, so I wasn't sure.

From what I could find (and my limited knowledge on Omnichannel), there could still be system messages after closing the chat, so we might want to show these messages in case of PDF transcript? 

Correct me, if I'm wrong.

yash-rajpal: Duplicate.

KevLehman: There should be no system message after closing the chat. The only message sent _after_ closing is the message with the PDF itself 👀 so we should be safe doing the same the email does, of defaulting to `new Date()` to get the messages up to that point.

KevLehman: sorry im dum didn't see this was dupe lol

ee/packages/pdf-worker/src/strategies/ChatTranscript.ts
MarcosSpessatto: For sure, I agree with that.

apps/meteor/app/livechat/server/lib/LivechatTyped.ts
```diff
@@ -631,10 +634,22 @@ class LivechatClass {
 				author = showAgentInfo ? message.u.name || message.u.username : i18n.t('Agent', { lng: userLanguage });
 			}
 
+			const isSystemMessage = MessageTypes.isSystemMessage(message);
+			const messageType = isSystemMessage && MessageTypes.getType(message);
+
+			// eslint-disable-next-line no-nested-ternary
```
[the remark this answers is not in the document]
yash-rajpal: Just that it's very readable and understandable this way, doing this logic in separate steps would make code messy and difficult to understand, IMO.

This is simple, as it either takes system message or `msg`

apps/meteor/server/models/raw/Messages.ts
yash-rajpal: Got it, will implement accordingly

yash-rajpal: I checked this.

Just to clarify, I didn't create a new function to fetch the messages, rather updated the existing function used.

Also was checking why we were already using different functions, I stumbled upon a PR (by you) #28165 , I see why this was done using a different strategy on email transcripts, while we need all messages on the pdf transcripts.

I don't see any issues using different functions for both, or maybe we can merge both the functions as pass `beforeTs` as an optional param?

yash-rajpal: Discussed over DM and decided to keep both functions as separate.

ee/packages/pdf-worker/src/strategies/ChatTranscript.ts
yash-rajpal: Not needed anymore, as made the process synchronous.

apps/meteor/server/models/raw/Messages.ts
```diff
@@ -389,6 +390,10 @@ export class MessagesRaw extends BaseRaw<IMessage> implements IMessagesModel {
 			query.t = { $nin: types };
 		}
 
+		if (!showSystemMessages) {
```
MarcosSpessatto, line 393: Shouldn't we validate this against `types`? I mean, it wouldn't make sense to have both at the same time, and if so, maybe we should guarantee that.

apps/meteor/server/models/raw/Messages.ts
```diff
@@ -424,14 +429,25 @@ export class MessagesRaw extends BaseRaw<IMessage> implements IMessagesModel {
...
+	): FindCursor<IMessage> {
+		const query: Filter<IMessage> = {
+			rid,
+		};
+
+		if (ignoredTypes.length > 0) {
+			query.t = { $nin: ignoredTypes };
+		}
+
+		if (!showSystemMessages) {
```
MarcosSpessatto, line 446: Same question as before, since this will override the previous condition.

ee/packages/omnichannel-services/src/OmnichannelTranscript.ts
```diff
@@ -159,22 +201,49 @@ export class OmnichannelTranscript extends ServiceClass implements IOmnichannelT
 		return quotes;
 	}
 
+	private getSystemMessage(message: MessageData): false | MessageData {
```
MarcosSpessatto, line 204: Why false? Wouldn't it be better to be MessageData | void?

MarcosSpessatto (changes requested):
Can we add a changeset? Also a description together with the image?

apps/meteor/server/models/raw/Messages.ts
yash-rajpal: This was done intentionally, explained in different comment why.

ee/packages/omnichannel-services/src/OmnichannelTranscript.ts
yash-rajpal: Just a design choice, I like to see `false` returned from the function as a confirmation that system message wasn't found. Hope this is fine, else will change to use `undefined`.

KevLehman (changes requested):

ee/packages/omnichannel-services/src/OmnichannelTranscript.ts
```diff
@@ -65,6 +87,10 @@ export class OmnichannelTranscript extends ServiceClass implements IOmnichannelT
 		this.worker = new PdfWorker('chat-transcript');
 		// eslint-disable-next-line new-cap
 		this.log = new loggerClass('OmnichannelTranscript');
+
+		// this is initialized as undefined and will be set when the first pdf is requested.
+		// if we try to initialize it at the start of the service using IIAFE, for some reason i18next doesn't return translations, maybe i18n isn't initialised yet
+		this.translations = undefined;
```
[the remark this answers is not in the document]
yash-rajpal: Nope, just being more imperative, to be more clear in code. Can remove this if required.

ee/packages/omnichannel-services/src/OmnichannelTranscript.ts
```diff
@@ -300,14 +399,16 @@ export class OmnichannelTranscript extends ServiceClass implements IOmnichannelT
 			const agent =
 				room.servedBy && (await Users.findOneAgentById(room.servedBy._id, { projection: { _id: 1, name: 1, username: 1, utcOffset: 1 } }));
 
+			const translations = this.translations ?? (await this.getAllTranslations());
+			this.translations = translations;
```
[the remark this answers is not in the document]
yash-rajpal: Will do, makes sense. 😬

ee/packages/omnichannel-services/src/OmnichannelTranscript.ts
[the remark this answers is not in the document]
KevLehman: Think we can remove so it gets init only once.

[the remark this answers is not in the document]
yash-rajpal: Done

ee/packages/pdf-worker/src/strategies/ChatTranscript.spec.ts
```diff
@@ -60,4 +60,10 @@ describe('Strategies/ChatTranscript', () => {
 			result.t('invalidKey');
 		}).toThrow('Translation not found for key: invalidKey');
 	});
+
+	it('should parse the system message', () => {
```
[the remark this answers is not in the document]
yash-rajpal: I guess this test case doesn't makes any sense here, since this service/class is responsible for just rendering the provided data, it doesn't contain the list of valid system messages.

Maybe we should test it on omnichannel-services where this parsing and logic happens.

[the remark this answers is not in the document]
yash-rajpal: Implemented unit tests for `OmnichannelTranscript` class and covered this case over there

apps/meteor/server/services/translation/service.ts
```diff
@@ -28,9 +28,22 @@ export class TranslationService extends ServiceClassInternal implements ITransla
...
+		const language = await this.getServerLanguageCached();
+
+		return Promise.all(
+			keys.map(async (key) => {
+				return {
+					key,
+					value: await this.translateText(key, language),
+				};
+			}),
+		);
```
KevLehman, line 47: Instead of this being a promise, this can call i18n directly and avoid the promises. wdyt?

KevLehman (dismissed):

KevLehman (approved):

ggazzo:
As your pr has changes related with performance, I don't think we need a service to deal with translations, I think everything could be moved to a lib and all services could use it, that would be much faster. what do you think?

yash-rajpal:
@ggazzo Do you mean moving all i18n logic to the package itself, so all packages can use?

If so, then that's a great improvement and we should definitely do that, but I think it is outside the scope of this PR and should be handled separately. Currently this PR is important for this release. Maybe this can be a future overall improvement. What do you think?

MarcosSpessatto (dismissed):

KevLehman:
Vote 3 for removing service. It was created when translations were handled by meteor. Now that the files are free of meteor we could move the actual i18n thing onto it.

Also vote of doing that on a separate PR 👀

yash-rajpal:
> you just explained exactly why I would keep your performance part in another PR, separate from the feature ;)

You are right, but earlier I implemented the feature without the performance part, but the implementation created performance issues as we were having to do a lot of i18n calls for each and every message, making entire process async (as getting translations was async). The performance part is just making this process synchronous using some tricks.

So, it is kinda needed, can't remove it for now to do in future. 

But for sure this translation service should be removed and should be handled by a lib, when doing this we can improve transcripts as well. What do you think?

ggazzo (dismissed):

PR #32755, opened 2024-07-10
fix: Missing `notify` when user uses `setStatus` endpoint

KevLehman:


KevLehman:
This was caused by a general websocket problem, notify was there

PR #32763, opened 2024-07-11
fix: AAC audio type not playing on web

rodrigok:
## Issue(s)
[NATIVE-592](https://rocketchat.atlassian.net/browse/NATIVE-592)

Fix regression from https://github.com/RocketChat/Rocket.Chat/pull/32471


diegolmello (dismissed):

apps/meteor/app/utils/lib/mimeTypes.ts
```diff
@@ -12,7 +12,11 @@ const getExtension = (param: string): string => {
 	return !extension || typeof extension === 'boolean' ? '' : extension;
 };
 
-const getMimeType = (fileName: string): string => {
+const getMimeType = (mimetype: string, fileName: string): string => {
+	if (mimetype && mimetype !== 'application/octet-stream') {
```
matheusbsilva137, line 16: I believe this could cause other issue :thinking: 

As an example, check the `ico` MIME type definition [above](https://github.com/RocketChat/Rocket.Chat/pull/32763/files#diff-e64159623d12c88eb4700ba6e62fc7a89517ee0603d165af21938f9fd899e582R7). We want it to be mapped to the `image/x-icon` MIME type. With the change you suggested here, we'll prefer the MIME type we get from busboy (which is not `application/octet-stream`, but actually `image/vnd.microsoft.icon`) -- that could hurt users that have blocked the `image/x-icon` MIME type :thinking:

Please check #32764 , I tried another approach that also fixes this issue

matheusbsilva137 (changes requested):

apps/meteor/app/utils/lib/mimeTypes.ts
rodrigok: @matheusbsilva137 can you review the code now?

apps/meteor/app/utils/lib/mimeTypes.ts
```diff
@@ -12,7 +14,14 @@ const getExtension = (param: string): string => {
 	return !extension || typeof extension === 'boolean' ? '' : extension;
 };
 
-const getMimeType = (fileName: string): string => {
+const getMimeType = (mimetype: string, fileName: string): string => {
+	// If the extension from the mimetype is different from the file extension, the file extension may be wrong
+	// so use the informed mimetype
```
matheusbsilva137, line 19: I believe we'd have another issue with this approach because `mimetype` is not always correct. We may receive `application/octet-stream` from it, which could mess things up

Let's use `lst` as an example (which I tried locally): [this line](https://github.com/RocketChat/Rocket.Chat/pull/32763/files#diff-b3f6491decaee72b3d324f149c851d11cb1288ec73b30cf1641baa9edefb397aR88) will provide `application/octet-stream` to this function when `fileName` is `test.lst`. Since `mime.extension('application/octet-stream')` returns `bin`, we'd use `application/octet-stream` as the MIME type for this file, which is incorrect

apps/meteor/app/utils/lib/mimeTypes.spec.ts
```diff
@@ -0,0 +1,83 @@
...
+	describe('getMimeType', () => {
+		for (const [extension, mimeType] of Object.entries(extensionToMimeType)) {
+			it(`should return the correct mimeType ${mimeType} for the given fileName file.${extension} passing the correct mimeType`, async () => {
+				expect(getMimeType(mimeType, `file.${extension}`)).to.be.eql(mimeType);
+			});
+		}
+
+		it('should return the correct mimeType for the given fileName', async () => {
+			for (const [extension, mimeType] of Object.entries(extensionToMimeType)) {
+				expect(getMimeType('application/unknown', `file.${extension}`)).to.be.eql(mimeType);
```
matheusbsilva137, line 71:
```suggestion
				expect(getMimeType('application/octet-stream', `file.${extension}`)).to.be.eql(mimeType);
```
I believe this incorrect line is what made these tests succeed. `mime.extension('application/unknown')` evaluates to `undefined`,  so we won't get [here](https://github.com/RocketChat/Rocket.Chat/pull/32763/files#diff-e64159623d12c88eb4700ba6e62fc7a89517ee0603d165af21938f9fd899e582R22). However, I believe you meant to have `application/octet-stream` here. In this case, `mime.extension('application/octet-stream')`  returns `bin`

matheusbsilva137 (changes requested):

apps/meteor/app/utils/lib/mimeTypes.ts
rodrigok: We need to trust on something. The mime type is much more reliable than extension.

apps/meteor/app/utils/lib/mimeTypes.spec.ts
rodrigok: No, it's right. If the mime is not in our database it should use the one from the extension

apps/meteor/app/utils/lib/mimeTypes.ts
```diff
@@ -4,6 +4,8 @@ mime.types.wav = 'audio/wav';
 mime.types.lst = 'text/plain';
 mime.define('image/vnd.microsoft.icon', { source: '', extensions: ['ico'] }, mime.dupAppend);
 mime.define('image/x-icon', { source: '', extensions: ['ico'] }, mime.dupAppend);
+mime.define('audio/aac', { source: '', extensions: ['aac'] }, mime.dupAppend);
+mime.types.aac = 'audio/aac';
```
matheusbsilva137, line 8:
```suggestion
mime.define('audio/aac', { source: '', extensions: ['aac'] }, mime.dupOverwrite);
```
This works just the same (the assignment in the second line works like an overwrite, which invalidates the first line)

apps/meteor/app/utils/lib/mimeTypes.spec.ts
matheusbsilva137: Just checked that we now have tests for when the mime is in our database, but it is default one. Looks good to me

matheusbsilva137 (changes requested):

apps/meteor/app/utils/lib/mimeTypes.spec.ts
matheusbsilva137: Just checked that we now have tests for when the mime is in our database, but it is default one. Looks good to me

apps/meteor/app/utils/lib/mimeTypes.ts
matheusbsilva137: I think it looks better with the changes you applied now, it should be ok to rely on the mime type (since this onde may be extracted by analyzing more info than just the file name) **when** it's not `application/octet-stream`

apps/meteor/app/utils/lib/mimeTypes.ts
rodrigok: Good one, done

MarcosSpessatto (approved):

yash-rajpal (approved):

PR #32764, opened 2024-07-11
fix: AAC audio type does not play on web

matheusbsilva137:

- Define `aac` files MIME type as `audio/aac` instead of `audio/x-aac` (which is not supported by our audio player);

## Issue(s)

Introduced by #32471

## Further comments

[NATIVE-592](https://rocketchat.atlassian.net/browse/NATIVE-592)


matheusbsilva137:
Closing in favor of #32763 (the changes suggested in this PR have been incorporated there)

PR #32766, opened 2024-07-11
fix: Reload page after redirect flow ends to avoid reactivity issues

KevLehman:
https://rocketchat.atlassian.net/browse/SUP-619

## Further comments

PR adds an additional step to the `redirect` flow of OAuth logins, refreshing the page to recover reactivity on the UI.

This should fix most user inconsistencies happening after a login via OAuth (redirect). This shouldn't affect any other type of login, like, for example, popup flow.


apps/meteor/client/meteorOverrides/login/oauth.ts
```diff
@@ -16,6 +18,37 @@ export const convertError = <T>(error: T): Accounts.LoginCancelledError | T => {
...
+	}
+	if (redirectRefreshState === 'complete' || service.loginStyle !== 'redirect') {
+		localStorage.removeItem('oauth_redirect_refresh_state');
+		return;
+	}
+
+	localStorage.setItem('oauth_redirect_refresh_state', 'pending');
+});
+
+Accounts.onLogin((login: any) => {
```
MarcosSpessatto, line 44: `ILoginAttempt` maybe?

MarcosSpessatto (changes requested):
Is there any possibility of adding automated tests to cover this? Asking because I'm not sure if we can simulate that...

Also, can we add a description to the PR?

apps/meteor/client/meteorOverrides/login/oauth.ts
KevLehman: Oh didn't know we had a type for this, let me check

KevLehman:
About testing, i don't think it's possible 😬 right now. I don't think we have any oauth login test 🤔and for reliability we would need an oauth server configured for the CI (like a keycloack server). Totally doable, but a project in itself

packages/core-typings/src/IMeteorError.ts
```diff
@@ -0,0 +1,22 @@
...
+			errorType: this.errorType,
+			error: this.error,
+			reason: this.reason,
+			message: this.message,
+			...(this.details && { details: this.details }),
+		};
+	}
+}
+
+export const isMeteorError = (error: any): error is MeteorError => error?.errorType === 'Meteor.Error';
```
MarcosSpessatto, line 22: Are we using this?

MarcosSpessatto (changes requested):

packages/core-typings/src/IMeteorError.ts
KevLehman: Not yet. Idea is to have another PR to move these types from core-services into core-typings, so i copied to whole file. I can remove as it's not used, won't affect. wdyt?

MarcosSpessatto: I'd say it's better not to add unrelated stuff here. Thanks.

MarcosSpessatto (approved):

ggazzo:
could you check if this https://github.com/RocketChat/Rocket.Chat/pull/32929 solves the case?
I was not aware this was a real issue, I though only test were facing it

KevLehman:
let me check

KevLehman:
Closing this for now as apparently linked PR fixed the issue 🎉

PR #32767, opened 2024-07-11
refactor:  `dispatchInquiryPosition` being called multiple times

ggazzo:

### DIFF
--- a/apps/meteor/app/livechat/server/lib/QueueManager.ts
+++ b/apps/meteor/app/livechat/server/lib/QueueManager.ts
@@ -14,15 +14,18 @@ import { Random } from '@rocket.chat/random';
 import { Match, check } from 'meteor/check';
 import { Meteor } from 'meteor/meteor';
 
+import { dispatchInquiryPosition } from '../../../../ee/app/livechat-enterprise/server/lib/Helper';
 import { callbacks } from '../../../../lib/callbacks';
 import {
 	notifyOnLivechatInquiryChangedById,
 	notifyOnLivechatInquiryChanged,
 	notifyOnSettingChanged,
 } from '../../../lib/server/lib/notifyListener';
+import { settings } from '../../../settings/server';
 import { createLivechatRoom, createLivechatInquiry } from './Helper';
 import { Livechat as LivechatTyped } from './LivechatTyped';
 import { RoutingManager } from './RoutingManager';
+import { getInquirySortMechanismSetting } from './settings';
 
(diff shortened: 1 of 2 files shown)


apps/meteor/app/livechat/server/lib/QueueManager.ts
```diff
@@ -52,6 +55,7 @@ export const queueInquiry = async (inquiry: ILivechatInquiryRecord, defaultAgent
 		// We'll queue these inquiries so when new license is applied, they just start rolling again
 		// Minimizing disruption
 		await saveQueueInquiry(inquiry);
+		await QueueManager.dispatchInquiryPosition(inquiry);
```
KevLehman, line 58: This should only happen when `settings.get('Livechat_waiting_queue')` is true (which is an EE setting)

apps/meteor/app/livechat/server/lib/QueueManager.ts
```diff
@@ -62,33 +66,48 @@ export const queueInquiry = async (inquiry: ILivechatInquiryRecord, defaultAgent
...
 		logger.debug(`Inquiry with id ${inquiry._id} is ready. Delegating to agent ${inquiryAgent?.username}`);
+		/**
+		 * There is no need to call QueueManager.dispatchInquiryPosition
+		 * because the method is called on afterTakeInquiry anyway
+		 *
+		 */
 		return RoutingManager.delegateInquiry(dbInquiry, inquiryAgent, undefined, room);
 	}
+
+	await QueueManager.dispatchInquiryPosition(dbInquiry);
```
KevLehman, line 77: Same here

apps/meteor/app/livechat/server/lib/QueueManager.ts
```diff
@@ -62,33 +66,48 @@ export const queueInquiry = async (inquiry: ILivechatInquiryRecord, defaultAgent
...
-}
+export const QueueManager = class {
+	/**
+	 * It should be used when a inquiry is moved to queue(created|enabled|etc)
+	 * mainly used to inform visitors when they value is set for the first time
+	 *
+	 * @param inquiry
+	 *
+	 */
+	static async dispatchInquiryPosition(inquiry: ILivechatInquiryRecord) {
```
KevLehman, line 88: This was "ee" code, I don't mind moving it but just making sure the change is intentional (ee => ce)

KevLehman (changes requested):

PR #32769, opened 2024-07-11
fix: Sidebar scrolling misclick issue on iOS

MartinSchoeler:

An outdated polyfill was causing misclick issues when scrolling the channels sidebar

## Further comments

[SUP-582]

[SUP-582]: https://rocketchat.atlassian.net/browse/SUP-582?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/meteor/client/polyfills/hoverTouchClick.ts
```diff
@@ -1,17 +0,0 @@
-import * as domEvents from '../lib/utils/domEvents';
-import { isIOsDevice } from '../lib/utils/isIOsDevice';
```
yash-rajpal, line 2: The helper `isIosDevice` is not being used anywhere else in the project, we should remove that as well.

yash-rajpal (changes requested):
LGTM other than the comment left.

apps/meteor/client/polyfills/hoverTouchClick.ts
tassoevan: Even better! More deleted lines!

MartinSchoeler: Done!

yash-rajpal (approved):

MartinSchoeler:
Putting this as draft since the issue seems more complex than initially estimated

dougfabris (changes requested):
fix missing changeset

PR #32779, opened 2024-07-12
feat: Trace api calls

rodrigok:
closes https://github.com/RocketChat/Rocket.Chat/pull/22335

- [x] Microservices support
- [x] Fix BaseRaw on MS
- [x] EnvVar to enable it
- [x] Error handling

## Issue(s)

https://rocketchat.atlassian.net/browse/ARCH-1300


sampaiodiego (dismissed):

KevLehman (dismissed):

rodrigok:
I can't approve, but LGMT

apps/meteor/app/api/server/api.ts
```diff
@@ -645,8 +646,25 @@ export class APIClass<TBasePath extends string = ''> extends Restivus {
...
+										method: this.request.method,
+										userId: this.userId,
+									},
+								},
+								async (span) => {
+									if (span) {
+										this.response.setHeader('X-Trace-Id', span.spanContext().traceId);
+									}
+									return (
+										(await DDP._CurrentInvocation.withValue(invocation as any, async () => originalAction.apply(this))) || API.v1.success()
```
MarcosSpessatto, line 664: Wouldn't it be useful to save also the response status code and payload(maybe)? 🤔

packages/tracing/src/index.ts
```diff
@@ -0,0 +1,92 @@
...
+			throw err;
+		} finally {
+			span.end();
+		}
+	};
+
+	if (optl) {
+		const activeContext = propagation.extract(context.active(), optl);
+
+		return tracer.startActiveSpan(name, options, activeContext, computeResult);
```
MarcosSpessatto, line 67: We must make sure this is defined, right? What if someone forgets to start it?

packages/tracing/src/traceDatabaseCalls.ts
MarcosSpessatto, line 1: I'd say it would be better not to have this MongoDB dependency here, we are already exporting all the tracing API, so we could just use that inside the process we want.

packages/tracing/src/traceDatabaseCalls.ts
```diff
@@ -0,0 +1,49 @@
+import { trace, context } from '@opentelemetry/api';
+import type { MongoClient } from 'mongodb';
+
+import { isTracingEnabled } from '.';
+
+const tracer = trace.getTracer('core');
```
MarcosSpessatto, line 6: Also, I'd not hardcode 'core' here, even if we are sharing the DB among the services.

packages/tracing/src/traceDatabaseCalls.ts
```diff
@@ -0,0 +1,49 @@
...
+			return;
+		}
+
+		const { span } = DurationStart.get(event.requestId);
+		DurationStart.delete(event.requestId);
+
+		span.end();
+	});
+
+	client.on('commandFailed', (event) => DurationStart.delete(event.requestId));
```
MarcosSpessatto, line 48: I think we need to end the span here, right?

packages/tracing/src/index.ts
```diff
@@ -0,0 +1,92 @@
...
+	fn: F,
+	optl?: unknown,
+): ReturnType<F> {
+	const currentSpan = trace.getSpan(context.active());
+
+	if (process.env.LOG_UNTRACED_METHODS) {
+		console.log(`No active span for ${name}`, new Error().stack);
+	}
+
+	return currentSpan ? tracerSpan(name, options, fn, optl) : fn();
```
MarcosSpessatto, line 85: Wouldn't it be better to throw an error if the current span doesn't exist instead of just executing the fn? The developer could think it is being tracked while nothing was done.

MarcosSpessatto (changes requested):

apps/meteor/app/api/server/api.ts
sampaiodiego: do you mean adding the return value to the trace span?

packages/tracing/src/index.ts
sampaiodiego: yeah, we can test if it is defined and provide a better error than just node breaking because `tracer` is `undefined`.. I'll do that

packages/tracing/src/index.ts
sampaiodiego: no because this is the only purpose of this function, create a new span in case there is one already created..

we can name it better and maybe document the behavior, but that's the idea.. so using in addition to the flag `LOG_UNTRACED_METHODS` it will help us finding places where we missing a `span` on the above context

apps/meteor/app/api/server/api.ts
MarcosSpessatto: yes

rodrigok: Status code makes sense, but the payload may expose sensitive information and be heavy

MarcosSpessatto (dismissed):

KevLehman (dismissed):

PR #32791, opened 2024-07-15
fix: Allow custom fields on messages to be updated even if `msg` prop is not updated (via API)

KevLehman:

Fixed a behavior when updating messages that prevented the `customFields` prop from being updated if there were no changes to the `msg` property. Now, `customFields` will be always updated on message update even if `msg` doesn't change

## Issue(s)

[SUP-621](https://rocketchat.atlassian.net/browse/SUP-621)
## Further comments

[SUP-621]: https://rocketchat.atlassian.net/browse/SUP-621?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/meteor/tests/end-to-end/api/chat.ts
```diff
@@ -1460,6 +1467,62 @@ describe('[Chat]', () => {
...
+				.expect('Content-Type', 'application/json')
+				.expect(200)
+				.expect((res) => {
+					expect(res.body).to.have.property('success', true);
+					expect(res.body).to.have.nested.property('message.msg', 'This message was edited via API 2');
+					expect(res.body.message).to.have.property('customFields').that.is.an('object').that.deep.equals({ test: 'test' });
+				});
+		});
+
+		it('should update message custom fields without msg', async () => {
```
KevLehman, line 1508:
```suggestion
		it('should update message custom fields without changes to msg', async () => {
```

apps/meteor/tests/end-to-end/api/chat.ts
```diff
@@ -1460,6 +1467,62 @@ describe('[Chat]', () => {
...
+		});
+
+		it('should update message custom fields without changes to msg', async () => {
+			await request
+				.post(api('chat.update'))
+				.set(credentials)
+				.send({
+					roomId: testChannel._id,
+					msgId: message._id,
+					text: 'This message was edited via API 2',
```
matheusbsilva137, line 1515: Can we have a test case for when `text` is `undefined`? Since you made it optional
btw do we really need/want to make it optional?

matheusbsilva137 (changes requested):

apps/meteor/tests/end-to-end/api/chat.ts
KevLehman: Good question. For me if you can update one without the another, then both should be optional. `msg` can be updated alone as well as `customFields`, so ended up making it optional.

Wonder if that could have any side effects but since making something optional is not breaking we shouldn't.

Gonna add the other test, thanks man!

KevLehman: Good one! At the end, the prop couldn't be optional without bigger code changes, so will be kept as required but at least it doesn't have to change now :its_something:

apps/meteor/app/lib/server/functions/updateMessage.ts
```diff
@@ -12,7 +12,7 @@ import { validateCustomMessageFields } from '../lib/validateCustomMessageFields'
 import { parseUrlsInMessage } from './parseUrlsInMessage';
 
 export const updateMessage = async function (
-	message: AtLeast<IMessage, '_id' | 'rid' | 'msg'>,
+	message: AtLeast<IMessage, '_id' | 'rid' | 'msg' | 'customFields'>,
```
MarcosSpessatto, line 15: Why do we need to have `customFields` here being required?

apps/meteor/app/lib/server/methods/updateMessage.ts
```diff
@@ -12,7 +12,11 @@ import { updateMessage } from '../functions/updateMessage';
 
 const allowedEditedFields = ['tshow', 'alias', 'attachments', 'avatar', 'emoji', 'msg', 'customFields', 'content'];
 
-export async function executeUpdateMessage(uid: IUser['_id'], message: AtLeast<IMessage, '_id' | 'rid' | 'msg'>, previewUrls?: string[]) {
+export async function executeUpdateMessage(
+	uid: IUser['_id'],
+	message: AtLeast<IMessage, '_id' | 'rid' | 'msg' | 'customFields'>,
```
MarcosSpessatto, line 17: Same

MarcosSpessatto (changes requested):

apps/meteor/app/lib/server/functions/updateMessage.ts
KevLehman: image

Custom fields should retain the `optional` part as atleast just "picks" the types from the base one.

MarcosSpessatto (changes requested):
Can we have a PR description?

MarcosSpessatto (dismissed):

MarcosSpessatto (approved):

matheusbsilva137 (approved):

PR #32793, opened 2024-07-15
feat: Video Conference persistent chat

pierre-lehnen-rc:

This new feature allows video conference provider apps to use a rocket.chat discussion to store the messages sent by their users during a conference call. The feature is enabled through a setting on rocket.chat workspace configuration, but it will only take effect if the video conf provider app registers itself as compatible with it.

When enabled, Rocket.Chat will automatically create a new discussion every time a conference is started and the id of this discussion will be made available to the app. Any user who joins the conference will automatically be subscribed to that discussion. The provider app is then responsible for actually sending any messages exchanged during the conference to the specified discussion.

## Issue(s)

[PPC-1](https://rocketchat.atlassian.net/jira/software/c/projects/PPC/issues/PPC-1)
[ADR 092](https://adr.rocket.chat/092)

## Further comments

Discussions created for conference calls will not be listed on the room's list of discussions. They can be accessed through the list of conference calls instead. A button to access the discussion is also available on the message sent when a conference is started.

Includes PR #32817 

[PPC-1]: https://rocketchat.atlassian.net/browse/PPC-1?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/meteor/server/services/video-conference/service.ts
```diff
@@ -686,13 +718,15 @@ export class VideoConfService extends ServiceClassInternal implements IVideoConf
 			},
 			providerName,
 		});
+
+		await this.runNewVideoConferenceEvent(callId);
+
+		await this.maybeCreateDiscussion(callId, user);
```
KevLehman, line 724: Would something be affected here if we turn these into a p.all?

apps/meteor/server/services/video-conference/service.ts
```diff
@@ -1034,4 +1070,124 @@ export class VideoConfService extends ServiceClassInternal implements IVideoConf
 		await this.runVideoConferenceChangedEvent(call._id);
 		await this.sendAllPushNotifications(call._id);
 	}
+
+	private async isPersistentChatEnabled(): Promise<boolean> {
```
KevLehman, line 1074: Does this need to be async?
```suggestion
	private isPersistentChatEnabled(): boolean {
```

tassoevan (approved):

apps/meteor/server/services/video-conference/service.ts
```diff
@@ -1034,4 +1070,124 @@ export class VideoConfService extends ServiceClassInternal implements IVideoConf
...
+
+		const call = await VideoConferenceModel.findOneById(callId, { projection: { users: 1, messages: 1 } });
+		if (!call) {
+			return;
+		}
+
+		await VideoConferenceModel.setDiscussionRidById(callId, rid);
+
+		if (room) {
+			await Promise.all(call.users.map(({ _id }) => this.addUserToDiscussion(room._id, _id)));
```
KevLehman, line 1176: wonder if this will present any problem on big calls (with a ton of participants). Would it be a premature opt to batch these in groups instead of sending all them at once?

apps/meteor/server/services/video-conference/service.ts
```diff
@@ -61,33 +69,41 @@ export class VideoConfService extends ServiceClassInternal implements IVideoConf
...
-			throw new Error('failed-to-load-own-data');
-		}
+			if (!room) {
+				throw new Error('invalid-room');
+			}
 
-		if (type === 'direct') {
-			if (!isRoomCompatibleWithVideoConfRinging(room.t, room.uids)) {
-				throw new Error('type-and-room-not-compatible');
+			const user = await Users.findOneById<IUser>(createdBy, {});
```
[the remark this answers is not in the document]
pierre-lehnen-rc: Yes, the full user is needed down the line for the sendMessage function, which is still typed as `any` for the user param; So it's hard to know exactly what fields it needs (but it's many of them as there are many callbacks and apps-engine events involved).

This query is also not new in this PR, this line only had whitespace changes.

apps/meteor/server/services/video-conference/service.ts
pierre-lehnen-rc: Yes, the event needs to run first so that apps may assign an existing ID to new conferences (there's a pexip setting where discussions will be reused); Rocket.Chat only creates a discussion if the app didn't assign one during that event.

apps/meteor/server/services/video-conference/service.ts
pierre-lehnen-rc: It did initially cause I was checking for more things inside of it but it doesn't need to be async anymore with the current code. I'll change it.

apps/meteor/server/services/video-conference/service.ts
pierre-lehnen-rc: I wondered about this as well but with the current use cases this function will always be called with an empty list of users and I was just adding this here for compatibility in case an app is changed down the line. I didn't want to over engineer it for something that may never really be needed.

apps/meteor/server/services/video-conference/service.ts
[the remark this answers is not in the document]
KevLehman: Yeah, assumed that but stopped reading on the sendMessage call. Thanks man!

apps/meteor/server/services/video-conference/service.ts
KevLehman: interesting, i thought this was kinda an "add all users from call to discussion" thing, if it's not something that will be used that often or with many users, then should be fine

d-gubert (approved):

apps/meteor/tests/end-to-end/apps/video-conferences.ts
```diff
@@ -233,6 +234,56 @@ describe('Apps - Video Conferences', () => {
...
 			});
+
+			it('should start a call successfully when using a provider that supports persistent chat', async function () {
+				if (!process.env.IS_EE) {
+					this.skip();
+					return;
+				}
+
+				await updateSetting('VideoConf_Default_Provider', 'persistentchat');
+				await updateSetting('VideoConf_Enable_Persistent_Chat', true);
```
MarcosSpessatto, line 245: Let's rever this back to defaults `after` all

apps/meteor/tests/end-to-end/apps/video-conferences.ts
```diff
@@ -233,6 +234,56 @@ describe('Apps - Video Conferences', () => {
...
+			});
+
+			it('should start a call successfully when using a provider that supports persistent chat with the feature disabled', async function () {
+				if (!process.env.IS_EE) {
+					this.skip();
+					return;
+				}
+
+				await updateSetting('VideoConf_Default_Provider', 'persistentchat');
+				await updateSetting('VideoConf_Enable_Persistent_Chat', false);
```
MarcosSpessatto, line 270: Same

apps/meteor/tests/end-to-end/apps/video-conferences.ts
```diff
@@ -284,64 +335,224 @@ describe('Apps - Video Conferences', () => {
 		});
 
 		describe('[/video-conference.info]', () => {
-			let callId: string | undefined;
+			describe('[basic provider]', () => {
+				let callId: string | undefined;
+
+				before(async () => {
+					await updateSetting('VideoConf_Default_Provider', 'test');
```
MarcosSpessatto, line 342: Let's revert this (even if it's the same value) `after` all

apps/meteor/tests/end-to-end/apps/video-conferences.ts
```diff
@@ -284,64 +335,224 @@ describe('Apps - Video Conferences', () => {
 		});
 
 		describe('[/video-conference.info]', () => {
-			let callId: string | undefined;
+			describe('[basic provider]', () => {
```
MarcosSpessatto, line 338: Basic means no support for persistent chat, right? If so, I think you can explicitly put that.

apps/meteor/tests/end-to-end/apps/video-conferences.ts
```diff
@@ -284,64 +335,224 @@ describe('Apps - Video Conferences', () => {
...
+			describe('[persistent chat provider]', () => {
+				let callId: string | undefined;
+
+				before(async () => {
+					if (!process.env.IS_EE) {
+						return;
+					}
+
+					await updateSetting('VideoConf_Default_Provider', 'persistentchat');
+					await updateSetting('VideoConf_Enable_Persistent_Chat', true);
```
MarcosSpessatto, line 388: Let's revert this `after` all

apps/meteor/tests/end-to-end/apps/video-conferences.ts
```diff
@@ -284,64 +335,224 @@ describe('Apps - Video Conferences', () => {
...
+			describe('[persistent chat provider with the feature disabled]', () => {
+				let callId: string | undefined;
+
+				before(async () => {
+					if (!process.env.IS_EE) {
+						return;
+					}
+
+					await updateSetting('VideoConf_Default_Provider', 'persistentchat');
+					await updateSetting('VideoConf_Enable_Persistent_Chat', false);
```
MarcosSpessatto, line 431: Same

apps/meteor/tests/end-to-end/apps/video-conferences.ts
```diff
@@ -284,64 +335,224 @@ describe('Apps - Video Conferences', () => {
...
+				before(async () => {
+					await updateSetting('VideoConf_Default_Provider', 'test');
+					const res = await request.post(api('video-conference.start')).set(credentials).send({
+						roomId,
+					});
+
+					callId = res.body.data.callId;
+				});
+
+				it('should load the video conference data successfully', async () => {
```
MarcosSpessatto, line 350: Shouldn't we include a test case with the basic provider AND the persistent chat setting enabled?

apps/meteor/tests/end-to-end/apps/video-conferences.ts
```diff
@@ -284,64 +335,224 @@ describe('Apps - Video Conferences', () => {
...
+							expect(res.body).to.have.a.property('_id').equal(callId);
+							expect(res.body).to.have.a.property('discussionRid').that.is.a('string');
+
+							const { discussionRid } = res.body;
+							expect(res.body).to.have.a.property('url').equal(`pchat/videoconference/${callId}/${discussionRid}`);
+						});
+				});
+			});
+
+			describe('[persistent chat provider with the feature disabled]', () => {
```
MarcosSpessatto, line 422: Let's add some test cases with the Discussion setting disabled as well please.

apps/meteor/tests/end-to-end/apps/video-conferences.ts
```diff
@@ -369,8 +581,13 @@ describe('Apps - Video Conferences', () => {
 						expect(call1).to.have.a.property('createdBy').that.is.an('object');
 						expect(call1.createdBy).to.have.a.property('_id').equal(credentials['X-User-Id']);
 						expect(call1.createdBy).to.have.a.property('username').equal(adminUsername);
+						expect(call1).to.not.have.a.property('discussionRid');
 
 						expect(call2).to.have.a.property('_id').equal(callId2);
+						expect(call2).to.not.have.a.property('discussionRid');
+
+						expect(call3).to.have.a.property('_id').equal(callId3);
+						expect(call3).to.have.a.property('discussionRid').that.is.a('string');
```
MarcosSpessatto, line 590: Is this the expected behavior? The list endpoint returns either type and not consider the setting for the persistent chat, right? Curious.

apps/meteor/tests/end-to-end/apps/video-conferences.ts
MarcosSpessatto, line 1: Can we have a test case to ensure the discussionRid is valid?

MarcosSpessatto (changes requested):

tassoevan (dismissed):

apps/meteor/tests/end-to-end/apps/video-conferences.ts
pierre-lehnen-rc: Yes, the setting is considered in the call creation, the list endpoint will return whatever data is there regardless of the setting value.

tassoevan (approved):

d-gubert (approved):

MarcosSpessatto (approved):

PR #32809, opened 2024-07-17
fix: deactivate federated internal users permanently with the external user

debdutdeb:
## Further comments

CORE-616


.changeset/new-mayflies-wait.md
```diff
@@ -0,0 +1,5 @@
+---
+'@rocket.chat/meteor': minor
```
MarcosSpessatto, line 2:
```suggestion
'@rocket.chat/meteor': patch
```

apps/meteor/server/methods/setUserActiveStatus.ts
```diff
@@ -17,20 +20,40 @@ Meteor.methods<ServerMethods>({
...
+		const { federated } = (await Users.findOneById<Pick<IUser, 'federated'>>(uid, { projection: { federated: 1 } })) || {};
+
+		if (federated) {
+			throw new Meteor.Error('error-not-allowed', 'Deactivating federated user is not allowed', { method: 'setUserActiveStatus' });
+		}
+
+		const remoteUser = await MatrixBridgedUser.getExternalUserIdByLocalUserId(userId);
+
+		if (remoteUser) {
+			if (active) {
```
MarcosSpessatto, line 46: Is this correct? If so, it seems the error message is wrong.

MarcosSpessatto (changes requested):

apps/meteor/server/methods/setUserActiveStatus.ts
debdutdeb: `active` here is from the payload, so "mark this user active again". 

https://github.com/RocketChat/Rocket.Chat/blob/8fc6ca8b4e09b38ea9c74dbed586b8bc3694cf09/apps/meteor/app/api/server/v1/users.ts#L104-L111

apps/meteor/server/methods/setUserActiveStatus.ts
MarcosSpessatto, line 1: I think we should move this entire validation to the `setUserActiveStatus` instead of the Meteor method, I didn't see it before 😞

MarcosSpessatto (changes requested):

apps/meteor/server/methods/setUserActiveStatus.ts
debdutdeb: Ah right

apps/meteor/app/lib/server/functions/setUserActiveStatus.ts
```diff
@@ -52,12 +53,30 @@ export async function setUserActiveStatus(userId: string, active: boolean, confi
 		return false;
 	}
 
+	if (user.active === active) {
+		return true;
+	}
```
MarcosSpessatto, line 58: Do we really to change this here?

MarcosSpessatto (changes requested):

apps/meteor/app/lib/server/functions/setUserActiveStatus.ts
debdutdeb: if I don't do it here, one level up in the meteor method I'd have to (more ideal) but then we'd be querying the db for the same user twice. or change function signature

debdutdeb: If I don't check, if someone hits the endpoint for an already active user, it will still show the error. let me see if I can move it around

debdutdeb: @MarcosSpessatto pls check :)

MarcosSpessatto (approved):

PR #32810, opened 2024-07-17
fix: bad word filtering not working

abhinavkrin:

Fixes bad word filtering not working issue. If we look into the DB record, we'll see the message has been edited, but the UI still shows the word. The issue was identified to be due to displaying the value in the `md` property, and not the `msg` property, and the `md` is not edited to filter the bad words.
This occurs due to bad word filter service running after the markdown parser.

https://github.com/user-attachments/assets/9eb246ff-a2d9-46d6-a4a2-47cec4601693

## Further comments

[SUP-624](https://rocketchat.atlassian.net/browse/SUP-624)

[SUP-624]: https://rocketchat.atlassian.net/browse/SUP-624?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


KevLehman (commented):
Can we add some api tests?

apps/meteor/tests/end-to-end/api/chat.ts
```diff
@@ -768,6 +769,44 @@ describe('[Chat]', () => {
 				.end(done);
 		});
 
+		describe('Bad words filter', () => {
+			before(async () => {
```
Gustrb, line 773: I guess we can do them both at once

```javascript
before(() => Promise.all([updateSetting('Message_AllowBadWordsFilter', true), updateSetting('Message_BadWordsFilterList', 'badword')]));
```

apps/meteor/tests/end-to-end/api/chat.ts
```diff
@@ -768,6 +769,44 @@ describe('[Chat]', () => {
...
+				await updateSetting('Message_AllowBadWordsFilter', true);
+				await updateSetting('Message_BadWordsFilterList', 'badword');
+			});
+
+			after(async () => {
+				await updateSetting('Message_AllowBadWordsFilter', false);
```
Gustrb, line 779: we can do both in a Promise.all here as well

apps/meteor/tests/end-to-end/api/chat.ts
```diff
@@ -768,6 +769,44 @@ describe('[Chat]', () => {
...
+				await updateSetting('Message_BadWordsFilterList', '');
+			});
+
```
Gustrb, line 782: Can't we add a test case in which we have more than one badword, to assure it works as well :P

apps/meteor/tests/end-to-end/api/chat.ts
```diff
@@ -1460,6 +1499,42 @@ describe('[Chat]', () => {
 					expect(res.body.message).to.have.property('attachments').that.is.an('array').that.has.lengthOf(0);
 				});
 		});
+
+		describe('Bad words filter', () => {
+			before(async () => {
+				await updateSetting('Message_AllowBadWordsFilter', true);
```
Gustrb, line 1505: I think we should be doing the before and the after settings update inside of a promise all, check the comment on the other test

Gustrb (approved):

KevLehman (approved):

MarcosSpessatto (approved):

PR #32820, opened 2024-07-17
feat: Allow admins to decide if email transcript should be sent always

KevLehman:

Added a new setting `Livechat_transcript_send_always` that allows admins to decide if email transcript should be sent all the times when a conversation is closed. This setting bypasses agent's preferences. For this setting to work, `Livechat_enable_transcript` should be off, meaning that visitors will no longer receive the option to decide if they want a transcript or not.

Note: When this setting is true, agent's preferences regarding email transcript will be disabled and ignored. The description of the preference was accommodated to let users know about this change. 
## Issue(s)

https://rocketchat.atlassian.net/browse/CORE-496


apps/meteor/app/livechat/server/lib/LivechatTyped.ts
```diff
@@ -326,11 +298,13 @@ class LivechatClass {
 
 		this.logger.debug(`DB updated for room ${room._id}`);
 
+		const transcriptRequested =
+			!!transcriptRequest || (!settings.get('Livechat_enable_transcript') && settings.get('Livechat_transcript_send_always'));
```
MarcosSpessatto, line 302: Wouldn't it be better to check the "always" one first?

apps/meteor/tests/unit/app/livechat/server/lib/parseTranscriptRequest.spec.ts
```diff
@@ -0,0 +1,124 @@
...
+});
+
+describe('parseTranscriptRequest', () => {
+	beforeEach(() => {
+		settingsGetMock.get.reset();
+		modelsMock.Users.findOneById.reset();
+		modelsMock.LivechatVisitors.findOneById.reset();
+	});
+
+	it('should return options when Livechat_enable_transcript setting is true', async () => {
```
MarcosSpessatto, line 30: I think it would be better to have the meaning for that also, like: "should return the object allowing the request etc etc". WDYT?

MarcosSpessatto: The same for all other cases

packages/i18n/src/locales/en.i18n.json
```diff
@@ -3269,6 +3269,8 @@
   "Livechat_email_transcript_has_been_requested": "The transcript has been requested. It may take a few seconds.",
   "Livechat_transcript_request_has_been_canceled": "The chat transcription request has been canceled.",
   "Livechat_transcript_sent": "Omnichannel transcript sent",
+  "Livechat_transcript_send_always": "Always send email transcript to visitors",
+  "Livechat_transcript_send_always_Description": "Send email transcript to visitors automatically without asking visitor. This setting bypasses agent's preferences regarding email transcripts.",
```
MarcosSpessatto, line 3273: Shouldn't it be visitors' preferences?

MarcosSpessatto (changes requested):

.changeset/shaggy-hats-raise.md
```diff
@@ -0,0 +1,5 @@
+---
+"@rocket.chat/meteor": patch
+---
+
+Added a new setting `Livechat_transcript_send_always` that allows admins to decide if email transcript should be sent all the times when a conversation is closed. This setting bypasses agent's preferences. For this setting to work, `Livechat_enable_transcript` should be off, meaning that visitors will no longer receive the option to decide if they want a transcript or not.
```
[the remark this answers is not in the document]
KevLehman: Oh, we have both. We have agent's preferences for sending transcripts. Visitors don't have a preference, they're asked for a transcript all the times when enable_transcript setting is on. So in this case, visitors are not asked and agents preferences are ignored.

apps/meteor/app/livechat/server/lib/LivechatTyped.ts
KevLehman: This is not checking, i'm not sure what we use this value for tbh but we pass `transcriptRequested: true|false` to the closing message.

In here, if we already have a transcript request (via the `transcriptRequest` object) then we assume it's true.

BUT, if we don't have a transcriptRequest object but the setting is enabled, we assume this as `true` as well because a transcript will be sent.

TLDR: that thing is just a flag and we can check any of those first.

MartinSchoeler (approved):

packages/i18n/src/locales/en.i18n.json
```diff
@@ -3270,6 +3270,8 @@
...
```
KevLehman, line 3274:
```suggestion
  "Livechat_transcript_send_always": "Always send conversation transcript to visitors via email",
  "Livechat_transcript_send_always_Description": "Once finished, send conversation transcript via email to visitors automatically, regardless of agent's preferences.",
```

MartinSchoeler (approved):

apps/meteor/app/livechat/server/lib/parseTranscriptRequest.ts
```diff
@@ -0,0 +1,59 @@
...
+		return options;
+	}
+
+	// send always is disabled, no changes
+	const sendAlways = settings.get<boolean>('Livechat_transcript_send_always');
+	if (!sendAlways) {
+		return options;
+	}
+
+	const visitorData = visitor || (await LivechatVisitors.findOneById(room.v._id));
```
matheusbsilva137, line 25:
```suggestion
	const visitorData = visitor || (await LivechatVisitors.findOneById(room.v._id, { projection: { visitorEmails: 1 }));
```
I believe we're missing a `projection` here

apps/meteor/tests/unit/app/livechat/server/lib/parseTranscriptRequest.spec.ts
```diff
@@ -0,0 +1,124 @@
...
+		expect(options.emailTranscript.requestData.requestedBy).to.be.deep.equal({ _id: '123' });
+	});
+
+	it('should return `options` param with `transcriptRequest` key attached when no user is passed, but theres an agent serving the room', async () => {
+		settingsGetMock.get.withArgs('Livechat_enable_transcript').returns(false);
+		settingsGetMock.get.withArgs('Livechat_transcript_send_always').returns(true);
+		modelsMock.Users.findOneById.returns({ _id: '123', username: 'kevsxxx', name: 'Kev' } as any);
+		modelsMock.LivechatVisitors.findOneById.returns({ visitorEmails: [{ address: 'abc@rocket.chat' }] } as any);
+
+		const options = await parseTranscriptRequest({ v: { _id: '123' } } as any, {} as any);
```
matheusbsilva137, line 99: Shouldn't the agent that is serving (as you mentioned in the test case description) the room be specified here? :thinking: in the first object passed to the function, I mean
Also, I believe it'd be better to use `createFakeRoom({ t: 'l', v: { _id: '123' } })` everywhere in this file instead. So that these tests are more flexible in case we ever need to check more properties within this function

apps/meteor/tests/unit/app/livechat/server/lib/parseTranscriptRequest.spec.ts
```diff
@@ -0,0 +1,124 @@
...
+		expect(options).to.have.property('emailTranscript').that.is.an('object');
+		expect(options.emailTranscript.requestData).to.have.property('email', 'abc@rocket.chat');
+		expect(options.emailTranscript.requestData).to.have.property('subject', '');
+		expect(options.emailTranscript.requestData.requestedBy).to.be.deep.equal({ _id: '123', username: 'kevsxxx', name: 'Kev' });
+	});
+
+	it('should return `options` param with `transcriptRequest` key attached when no user is passed, no agent is serving but rocket.cat is present', async () => {
+		settingsGetMock.get.withArgs('Livechat_enable_transcript').returns(false);
+		settingsGetMock.get.withArgs('Livechat_transcript_send_always').returns(true);
+		modelsMock.Users.findOneById.returns({ _id: 'rocket.cat', username: 'rocket.cat', name: 'Rocket Cat' } as any);
```
matheusbsilva137, line 110: Please check if `findOneById` has been called using 'rocket.cat' as the id

apps/meteor/tests/unit/app/livechat/server/lib/parseTranscriptRequest.spec.ts
```diff
@@ -0,0 +1,124 @@
...
+		expect(options).to.have.property('emailTranscript').that.is.an('object');
+		expect(options.emailTranscript.requestData).to.have.property('email', 'abc@rocket.chat');
+		expect(options.emailTranscript.requestData).to.have.property('subject', '');
+		expect(options.emailTranscript.requestData.requestedBy).to.be.deep.equal({ _id: '123' });
+	});
+
+	it('should return `options` param with `transcriptRequest` key attached when no user is passed, but theres an agent serving the room', async () => {
+		settingsGetMock.get.withArgs('Livechat_enable_transcript').returns(false);
+		settingsGetMock.get.withArgs('Livechat_transcript_send_always').returns(true);
+		modelsMock.Users.findOneById.returns({ _id: '123', username: 'kevsxxx', name: 'Kev' } as any);
```
matheusbsilva137, line 96: Please check if `findOneById` has been called using the agent id as the id

apps/meteor/tests/unit/app/livechat/server/lib/parseTranscriptRequest.spec.ts
```diff
@@ -0,0 +1,124 @@
...
+		modelsMock.LivechatVisitors.findOneById.returns({ visitorEmails: [{ address: 'abc@rocket.chat' }] } as any);
+
+		const options = await parseTranscriptRequest({ v: { _id: '123' } } as any, {} as any);
+		expect(options).to.be.deep.equal({});
+	});
+
+	it('should return `options` param with `transcriptRequest` key attached when user is passed', async () => {
+		settingsGetMock.get.withArgs('Livechat_enable_transcript').returns(false);
+		settingsGetMock.get.withArgs('Livechat_transcript_send_always').returns(true);
+		modelsMock.LivechatVisitors.findOneById.returns({ visitorEmails: [{ address: 'abc@rocket.chat' }] } as any);
```
matheusbsilva137, line 83: Please check if `findOneById` has been called using `123` (that is, `v._id`) as the id in this test case and all the ones above.

matheusbsilva137 (changes requested):

apps/meteor/client/components/Omnichannel/modals/CloseChatModal.tsx
```diff
@@ -77,7 +78,7 @@ const CloseChatModal = ({
 		({ comment, tags, transcriptPDF, transcriptEmail, subject }): void => {
 			const preferences = {
 				omnichannelTranscriptPDF: !!transcriptPDF,
-				omnichannelTranscriptEmail: !!transcriptEmail,
+				omnichannelTranscriptEmail: alwaysSendTranscript ? true : !!transcriptEmail,
```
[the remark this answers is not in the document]
KevLehman: Actually, dont know. Wanted to be more declarative here to put that when the setting is true, then it will always be true.

Open to change but feel kinda a pref thing

apps/meteor/tests/unit/app/livechat/server/lib/parseTranscriptRequest.spec.ts
KevLehman: No need since we know on L90 the function was called with the actual param (since that value is being returned, we check the returned value)

apps/meteor/tests/unit/app/livechat/server/lib/parseTranscriptRequest.spec.ts
KevLehman: Same as prev comment

apps/meteor/tests/unit/app/livechat/server/lib/parseTranscriptRequest.spec.ts
KevLehman: We get agent info from room.servedBy not from params.

On createFakeRoom, can change, not sure if worth it as the "rooms" here have only 2 or 3 props but will take a look

apps/meteor/tests/unit/app/livechat/server/lib/parseTranscriptRequest.spec.ts
KevLehman: L118 checks return value from this db call, implicitely testing the model returned the right value.

KevLehman: Note: i know the reason of checking on the actual param, but since each test is testing one code path feel checking for param usage unnecessary.

apps/meteor/tests/unit/app/livechat/server/lib/parseTranscriptRequest.spec.ts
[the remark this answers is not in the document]
MarcosSpessatto: I agree with that

MarcosSpessatto (changes requested):

apps/meteor/tests/unit/app/livechat/server/lib/parseTranscriptRequest.spec.ts
matheusbsilva137: Hmmm I get you, but I'm not sure I agree. I think it'd be good to have **distinct** `v._id` and `user._id` passed to the function in this test case so that we can be sure we're not messing thing up :thinking: wdyt? I know the returned value is correct, but did we get it from the correct source?

Also, since we're providing the `user` to the function in this test case, we should validate that the `Users` model function (`findOneById`) isn't even being called, actually (if it is, then we have a problem/unnecessary query to the DB)

apps/meteor/tests/unit/app/livechat/server/lib/parseTranscriptRequest.spec.ts
matheusbsilva137: I agree with you on the `LivechatVisitors` models , but not on the call to the `Users.findOneById` method :thinking: 
We could be providing 'rocket.cat' (that's what we expect) or anything else (such as undefined, idk) to this method. This test would pass regardless of the user `_id` we're trying to find :thinking: I believe it'd be good to validate the param in this specific method just to make sure we're not messing things up

packages/i18n/src/locales/en.i18n.json
```diff
@@ -6265,7 +6267,7 @@
   "Always_send_the_transcript_to_contacts_at_the_end_of_the_conversations": "Always send the transcript to contacts at the end of the conversations.",
   "Export_conversation_transcript_as_PDF": "Export conversation transcript as PDF",
   "Omnichannel_transcript_email": "Send chat transcript via email.",
-  "Accounts_Default_User_Preferences_omnichannelTranscriptEmail_Description": "Always send the transcript to contacts at the end of the conversations.",
+  "Accounts_Default_User_Preferences_omnichannelTranscriptEmail_Description": "Always send the transcript to contacts at the end of the conversations. This preference may be overriden by an admin setting.",
```
matheusbsilva137, line 6270: I believe we should delete all other translations that will be outdated with this change so that they're re-translated later (lingohub doesn't update lines, it just inserts new ones).

matheusbsilva137 (dismissed):

packages/i18n/src/locales/en.i18n.json
KevLehman: Yikes. How can we confirm that? Don't wanna remove all the other translations without eing sure they won't get updated otherwise 👀

apps/meteor/tests/unit/app/livechat/server/lib/parseTranscriptRequest.spec.ts
KevLehman: Yeah, but again, each test covers exactly one path of the code 👀 because we're tweaking the params that way. At least for this test, there's no way for it to go to another path so for the environment on the test, you're "sure" if it gets called, it's because you're receiving the param.

Never against validating things but really not sure if it's worth it. But then, it's 1 line so 🤷🏽 

i'll change it

MarcosSpessatto (dismissed):

apps/meteor/client/views/account/omnichannel/PreferencesConversationTranscript.tsx
```diff
@@ -1,24 +1,36 @@
...
-	const canSendTranscriptEmail = usePermission('send-omnichannel-chat-transcript');
+	const canSendTranscriptEmailPermission = usePermission('send-omnichannel-chat-transcript');
+	const canSendTranscriptEmail = canSendTranscriptEmailPermission && !alwaysSendEmailTranscript;
+	const canSendTranscriptEmailPref = watch('omnichannelTranscriptEmail');
 	const cantSendTranscriptPDF = !canSendTranscriptPDF || !hasLicense;
 
 	const omnichannelTranscriptPDF = useUniqueId();
 	const omnichannelTranscriptEmail = useUniqueId();
 
+	useEffect(() => {
```
MartinSchoeler, line 25: Let's not use useEffect here, if there is a need to be reactive to setting changes, we'll have to change the whole page

MartinSchoeler (changes requested):

MartinSchoeler (dismissed):

matheusbsilva137 (approved):

Issue #32812, opened 2024-07-17
Proxy external avatars


anicoa:
### Description/Actual behavior

If the setting Accounts_AvatarExternalProviderUrl is used currently the avatars are fetched client-side from external source. 

### Expected/wanted behavior:

In some situations direct client access is not wanted. The server should fetch avatars from external source and deliver the avatars to the client (proxying).

### Solution

A setting to proxy these images would be useful in different situations. A corresponding proxy setting is already implemented for different contexts (e.g. FileUpload-Webdav_Proxy_Avatars)

Update:
### Server Setup Information:

- Version of Rocket.Chat Server: 6.9.3
- License Type: Enterprise
- Number of Users: 25160
- Operating System: Linux
- Deployment Method: tar
- Number of Running Instances: 1
- DB Replicaset Oplog: 
- NodeJS Version: v14.21.3
- MongoDB Version: 5.0.26 / wiredTiger (oplog activated)

### Client Setup Information

- Desktop App or Browser Version: 3.9.4
- Operating System: Linux

anicoa:
Of course, i added the information. Thanks for the hint.

reetp:
Just helps us to know if it is old and unsupported or not :-)

I'll ask the team to look at this though it is more a NFR than a bug.

They may suggest as an EE user you contact support directly.

casalsgh:
@anicoa since you mentioned you have an enterprise license, note that it is a different queue for EE requests. You can check details on this link -> https://docs.rocket.chat/docs/support-center

Also, see it as a new feature request (NFR), so highly recommend opening a ticket and speaking with CSM assigned to your account or sales rep so that you get more visibility and can help with follow-ups.

dougfabris:
Solved here #32824]

PR #32824, opened 2024-07-17
feat: proxy avatars for Accounts_AvatarExternalProviderUrl

anicoa:
This PR adds the ability to proxy external avatars. Solves Issue #32812 

Internal ticket:  https://rocketchat.atlassian.net/browse/CONN-298


tassoevan (approved):

KevLehman (approved):

anicoa:
@tassoevan sorry if I get it wrong. But now - as the pr is merged now -  where can i activate the proxying of the external avatar url through the rocketchat server itself?  Can't find the feature I added the pr for in the admin section anymore ;)

tassoevan:
@anicoa We discussed internally if a setting made more sense than just **always** proxying all the avatars requests. As we didn't find any strong reasons for maintaining the old behavior, the setting was ditched while the new behavior was kept.

anicoa:
@tassoevan thanks for clarification and work. Found this right now in the code but could not reproduce the behaviour on a fresh updated 6.12  installation. Will try/investigate and report.

anicoa:
@tassoevan checked the functionality with one 6.12.1 and one 6.12 installation.

1. Set Accounts_AvatarExternalProviderUrl to https://www.google.com/{username}
2. Login
3. Developer-Tools show request to https://www.google.com/foobar

Anything else I have to consider?

anicoa:
@tassoevan we tested the functionality and it's working now. Thank you very much.

We have two suggestions we would/could also solve with a pr.

a) A backend option "Prefer local stored avatars" would be good for a smooth transition and also it opens the posibility to override a centraly stored avatar.

b) Error handling would be very good fetching the avatar from external URL. In case it returns HTTP 4xx or HTTP 5xx Rocket.Chat should generate avatar by first name/last name or by username.

Would an pr solving this two cases would be welcomed?

Thanks a lot
Nico

PR #32838, opened 2024-07-18
chore: block actions in federated rooms if federation configuration is invalid or it is disabled

debdutdeb:
## Further comments

CORE-524


apps/meteor/server/methods/addRoomOwner.ts
```diff
@@ -36,12 +37,18 @@ Meteor.methods<ServerMethods>({
 			});
 		}
 
-		if (!(await hasPermissionAsync(uid, 'set-owner', rid)) && !isRoomFederated(room)) {
+		const isFederated = isRoomFederated(room);
+
+		if (!(await hasPermissionAsync(uid, 'set-owner', rid)) && isFederated) {
```
MarcosSpessatto, line 42: `!isFederated` right? 🤔

apps/meteor/server/services/federation/infrastructure/rocket-chat/hooks/index.ts
```diff
@@ -48,6 +47,9 @@ export class FederationHooks {
 				if (!params?.user || !room) {
 					return;
 				}
+
+				throwIfFederationNotEnabledOrReady();
```
MarcosSpessatto, line 51: I think we should not validate this here because we run only validations here (e.g. someone trying to add a federated user to a non-federated room), that's why we were not checking if federation was enabled.

apps/meteor/server/services/federation/infrastructure/rocket-chat/hooks/index.ts
```diff
@@ -190,9 +185,16 @@ export class FederationHooks {
...
-		if (!data || !settings.get('Federation_Matrix_enabled')) {
+		if (!data) {
+			return;
+		}
+
+		if (!isFederationEnabled()) {
 			return;
 		}
+
+		throwIfFederationNotReady();
```
MarcosSpessatto, line 196: Are we checking and throwing here only if it's not ready yet because we were just (early) returning if federation is not enabled, right?

apps/meteor/server/services/federation/utils.ts
MarcosSpessatto, line 1: Let's type all the returning values for all functions here.

apps/meteor/server/services/room/service.ts
```diff
@@ -121,4 +122,20 @@ export class RoomService extends ServiceClassInternal implements IRoomService {
...
+		FederationCheck.blockIfFederationEnabledButNotReady(room);
+	}
+
+	async beforeNameChange(room: IRoom) {
+		FederationCheck.blockIfFederationEnabledButNotReady(room);
+	}
+
+	async beforeTopicChange(room: IRoom) {
+		FederationCheck.blockIfFederationEnabledButNotReady(room);
+	}
```
MarcosSpessatto, line 140: Let's type these functions

apps/meteor/server/services/room/hooks/FederationCheck.ts
```diff
@@ -0,0 +1,9 @@
+import { FederationMatrixInvalidConfigurationError, isFederationEnabled, isFederationReady } from '../../federation/utils';
+
+export class FederationCheck {
+	public static blockIfFederationEnabledButNotReady({ federated = false }: { federated?: boolean }) {
```
MarcosSpessatto, line 4: Can we unit-test this?

apps/meteor/server/services/messages/service.ts
```diff
@@ -209,4 +222,16 @@ export class MessageService extends ServiceClassInternal implements IMessageServ
...
+		if ((isMessageFromMatrixFederation(message) || isRoomFederated(room)) && isFederationEnabled() && !isFederationReady()) {
+			throw new FederationMatrixInvalidConfigurationError('Unable to react to message');
+		}
+	}
+
+	async beforeDelete(message: IMessage, room: IRoom) {
+		if ((isMessageFromMatrixFederation(message) || isRoomFederated(room)) && isFederationEnabled() && !isFederationReady()) {
+			throw new FederationMatrixInvalidConfigurationError('Unable to delete message');
+		}
+	}
```
MarcosSpessatto, line 236: It would be good to unit-test these.

apps/meteor/server/services/room/hooks/FederationCheck.ts
MarcosSpessatto, line 4: Wouldn't it be better to type the input as IRoom?

apps/meteor/server/services/federation/service.ts
```diff
@@ -332,6 +333,14 @@ export abstract class AbstractFederationService extends ServiceClassInternal {
...
 		}
 	}
+
+	public async beforeCreateRoom(room: Partial<IRoom>) {
+		if (!isRoomFederated(room)) {
+			return;
+		}
+
+		throwIfFederationNotEnabledOrReady();
+	}
```
MarcosSpessatto, line 343: It would be nice to unit-test test these as well

MarcosSpessatto (changes requested):

apps/meteor/app/channel-settings/server/functions/saveRoomTopic.ts
```diff
@@ -20,6 +20,10 @@ export const saveRoomTopic = async function (
 		});
 	}
 
+	const room = await Rooms.findOneById(rid);
+
+	await Room.beforeTopicChange(room!);
```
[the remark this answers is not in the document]
debdutdeb: we make sure rid is valid one layer up, but still need to query it here again since otherwise would have to change signature and lots of other calls. Since this is not a frequently called function, decided to not do a refactor-ish, and do it like this. wdyt?

apps/meteor/server/services/federation/infrastructure/rocket-chat/hooks/index.ts
debdutdeb: Yes

apps/meteor/server/services/messages/service.ts
debdutdeb: Let me centralize these checks, then we can add tests once for that module

apps/meteor/app/channel-settings/server/functions/saveRoomTopic.ts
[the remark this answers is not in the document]
MarcosSpessatto: We can keep it, but if you identified a refactor necessity, let's create a task to keep track of it.

apps/meteor/tests/unit/server/federation/utils.spec.ts
```diff
@@ -0,0 +1,76 @@
...
+
+		it('should throw if federation is enabled but configuration is invalid', () => {
+			settings.enabled = true;
+			expect(throwIfFederationNotEnabledOrNotReady).to.throw;
+		});
+
+		it('should not throw if both federation is enabled and configuration is valid', () => {
+			settings.enabled = true;
+			settings.ready = true;
+			throwIfFederationNotEnabledOrNotReady();
```
MarcosSpessatto, line 57: I think we need to assert here to ensure the function is not throwing

apps/meteor/tests/unit/server/federation/utils.spec.ts
```diff
@@ -0,0 +1,76 @@
...
+
+	describe('#throwIfFederationEnabledButNotReady', () => {
+		it('should throw if federation is enabled and configuration is invalid', () => {
+			settings.enabled = true;
+			settings.ready = false;
+
+			expect(throwIfFederationEnabledButNotReady).to.throw;
+		});
+
+		it('should not throw if federation is disabled', () => {
```
MarcosSpessatto, line 69: Same as before

MarcosSpessatto (changes requested):

MarcosSpessatto (approved):

PR #32842, opened 2024-07-19
refactor: move system message calls from sendMessage to saveSystemMessage

ricardogarim:
As per [OPI-19](https://rocketchat.atlassian.net/browse/OPI-19), some use cases that send internal live chat messages have been updated. These messages now use the `saveSystemMessage` method instead of `sendMessage`. This change maintains the same outcome but bypasses certain validations that increase execution time.

We believe these validations are unnecessary since internal messages are typically predefined (often hardcoded) and do not violate server message rules.

[OPI-19]: https://rocketchat.atlassian.net/browse/OPI-19?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


sampaiodiego (changes requested):

apps/meteor/app/livechat/server/lib/Helper.ts
```diff
@@ -141,10 +140,7 @@ export const createLivechatRoom = async <
 	}
 
 	await callbacks.run('livechat.newRoom', room);
-
-	// TODO: replace with `Message.saveSystemMessage`
-
-	await sendMessage(guest, { t: 'livechat-started', msg: '', groupable: false, token: guest.token }, room);
+	await Message.saveSystemMessageAndNotifyUser('livechat-started', rid, '', { _id, username }, { groupable: false });
```
sampaiodiego, line 143: looks like you removed `token` for some reason.. I'm suggesting removing `groupable` since `saveSystemMessageAndNotifyUser` already adds it
```suggestion
	await Message.saveSystemMessageAndNotifyUser('livechat-started', rid, '', { _id, username }, { token: guest.token });
```

apps/meteor/app/livechat/server/lib/LivechatTyped.ts
```diff
@@ -296,35 +293,26 @@ class LivechatClass {
...
-			newRoom,
-		);
+
+		const transcriptRequested =
+			!!transcriptRequest || (!settings.get('Livechat_enable_transcript') && settings.get('Livechat_transcript_send_always'));
+
+		await Message.saveSystemMessageAndNotifyUser('livechat-close', rid, comment ?? '', closeData.closedBy, {
+			groupable: false,
+			transcriptRequested,
+		});
```
sampaiodiego, line 310:
```suggestion
		await Message.saveSystemMessageAndNotifyUser('livechat-close', rid, comment ?? '', closeData.closedBy, {
			transcriptRequested,
			...(isRoomClosedByVisitorParams(params) && { token: chatCloser.token }),
		});
```

apps/meteor/app/livechat/server/lib/LivechatTyped.ts
```diff
@@ -1254,31 +1242,29 @@ class LivechatClass {
...
+			t: 'livechat_transfer_history',
+			rid: room._id,
+			ts: new Date(),
+			msg: '',
+			u: {
+				_id,
+				username,
+			},
+			...(transferData.transferredBy.userType === 'visitor' && { token: room.v.token }),
+			groupable: false,
```
sampaiodiego, line 1255: all these properties are already set inside of `saveSystemMessageAndNotifyUser` so we don't need to pass them
```suggestion
			...(transferData.transferredBy.userType === 'visitor' && { token: room.v.token }),
```

apps/meteor/server/services/messages/service.ts
```diff
@@ -121,11 +141,19 @@ export class MessageService extends ServiceClassInternal implements IMessageServ
...
+		if (!insertedId) {
+			throw new Error('Failed to save system message.');
+		}
+
+		const createdMessage = await Messages.findOneById(insertedId);
+		if (!createdMessage) {
+			throw new Error('Failed to find the created message.');
+		}
+
+		void notifyOnMessageChange({ id: createdMessage._id });
```
sampaiodiego, line 153: since you just fetched the inserted message, we can optimize by passing the data to `notifyOnMessageChange` otherwise it will fetch it from DB again
```suggestion
		void notifyOnMessageChange({ id: createdMessage._id, data: createdMessage });
```

sampaiodiego (changes requested):

PR #32843, opened 2024-07-19
refactor: move broadcastMessageFromData to notifyListener

ricardogarim:
As per [CORE-575](https://rocketchat.atlassian.net/browse/CORE-575), the `broadcastMessageFromData` function has been moved to the `notifyListener` handler file to align with existing functions that handle server-to-listener communication. This change does not affect any use cases, inputs, or outputs of the existing function.

[CORE-575]: https://rocketchat.atlassian.net/browse/CORE-575?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/meteor/tests/unit/app/lib/server/lib/notifyListener.spec.ts
```diff
@@ -0,0 +1,215 @@
...
+            sinon.resetHistory();
+        });
+
+        afterEach(() => {
+            process.env = originalEnv;
+        });
+
+        const testCases = [
+            {
+                description: 'should return undefined if message is hidden or imported',
```
MarcosSpessatto, line 80: Can we have one case for each message? This one is the hidden one, can we create a specific one for the imported one?

apps/meteor/tests/unit/app/lib/server/lib/notifyListener.spec.ts
```diff
@@ -0,0 +1,215 @@
...
+            },
+            {
+                description: 'should hide message if type is in hideSystemMessage settings',
+                message: sampleMessage,
+                hideSystemMessages: ['user-muted', 'mute_unmute'],
+                useRealName: false,
+                expectedResult: undefined,
+            },
+            {
+                description: 'should return the message with real name if useRealName is true',
```
MarcosSpessatto, line 94: Can we create a new case for mentions using a real name as well?

apps/meteor/tests/unit/app/lib/server/lib/notifyListener.spec.ts
```diff
@@ -0,0 +1,215 @@
...
+        };
+
+        const testCases = [
+            {
+                description: 'should broadcast the message if dbWatchersDisabled is true',
+                dbWatchersDisabled: true,
+                expectBroadcast: true,
+            },
+            {
+                description: 'should not broadcast the message if dbWatchersDisabled is false',
```
MarcosSpessatto, line 192: What about a new case when there is no message to broadcast? Does it make sense?

MarcosSpessatto (changes requested):

MarcosSpessatto (approved):

KevLehman (approved):

PR #32846, opened 2024-07-19
fix: System messages are counted as agents' first responses in livechat rooms

matheusbsilva137:

 - Stop considering system messages (such as "agent joined room") as agents' first responses in livechat rooms. This caused the "average first response time" and "best first respone time" metrics to be much lower than expected.

## Steps to test or reproduce

Add a new agent to the woskspace, then create a livechat conversation and make sure to serve it to the new agent. Then, go to **Omnichannel > Analytics** and set "**Productivity**" as the analytics type.

**Current behavior:** the "Best first response time" and "Average of first response time" metrics will always be lower than the actual time that the agent took to respond to the livechat room (eg possibly 1s). System messages (such as "agent joined room") are being interpreted as first responses, which causes these metrics to go much lower;
**Expected behavior:** RC should only consider a first response in a room when the agent sends a text message. System messages should be disconsidered in the "Best first response time" and "Average of first response time" metrics.

## Further comments

[CORE-562](https://rocketchat.atlassian.net/browse/CORE-562)

[CORE-562]: https://rocketchat.atlassian.net/browse/CORE-562?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/meteor/tests/end-to-end/api/livechat/04-dashboards.ts
```diff
@@ -777,6 +779,212 @@ describe('LIVECHAT - dashboards', function () {
...
+		let roomId: string;
+		let visitorToken: string;
+
+		before(async () => {
+			agent = await createAnOnlineAgent();
+		});
+
+		after(() => deleteUser(agent.user));
+
+		afterEach(() => Promise.all([deleteVisitor(visitorToken), closeOmnichannelRoom(roomId)]));
```
KevLehman, line 898: same as the other

apps/meteor/tests/end-to-end/api/livechat/20-messages.ts
```diff
@@ -39,7 +39,7 @@ describe('LIVECHAT - messages', () => {
 
 			await sendMessage(roomId, 'Hello from visitor', token);
 			const agentMsgSentence = faker.lorem.sentence();
-			const agentMsg = await sendAgentMessage(roomId, agentMsgSentence);
+			const agentMsg = await sendAgentMessage(roomId, credentials, agentMsgSentence);
```
KevLehman, line 42: Personal pref: since the function internally defaults to credentials, you can just pass undefined here.

(Actually, passing creds as the last param prevents these types of changes as most of the times you'll use the default value for this call)

apps/meteor/tests/end-to-end/api/livechat/04-dashboards.ts
```diff
@@ -777,6 +779,212 @@ describe('LIVECHAT - dashboards', function () {
 		});
 	});
 
+	(IS_EE ? describe : describe.skip)('[livechat/analytics/agent-overview] - Average first response time', () => {
```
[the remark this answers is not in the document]
matheusbsilva137: Actually, I believe these tests can be executed on community :man_facepalming: I've just removed this condition

packages/core-typings/src/IMessage/IMessage.ts
```diff
@@ -249,6 +249,8 @@ export const isEditedMessage = (message: IMessage): message is IEditedMessage =>
 	'_id' in (message as IEditedMessage).editedBy &&
 	typeof (message as IEditedMessage).editedBy._id === 'string';
 
+export const isSystemMessage = (message: IMessage): message is ISystemMessage => 't' in message;
```
MarcosSpessatto, line 252: I'm not sure if we could only consider the message having `t` as a system message. 🤔 

Maybe we should also check the type of the message (`MessageTypesValues`)... WDYT?

MarcosSpessatto (dismissed):
There are some failing tests... Can you check those?

packages/core-typings/src/IMessage/IMessage.ts
matheusbsilva137: I'm refactoring how we define message types to validate this

packages/core-typings/src/IMessage/IMessage.ts
```diff
@@ -249,6 +255,9 @@ export const isEditedMessage = (message: IMessage): message is IEditedMessage =>
 	'_id' in (message as IEditedMessage).editedBy &&
 	typeof (message as IEditedMessage).editedBy._id === 'string';
 
+export const isSystemMessage = (message: IMessage): message is ISystemMessage =>
+	't' in message && message.t !== undefined && MessageTypes.includes(message.t);
```
KevLehman, line 259:
```suggestion
	message.t !== undefined && MessageTypes.includes(message.t);
```

will work pretty much the same

KevLehman (approved):
hope ci passes lol

apps/meteor/tests/end-to-end/api/livechat/04-dashboards.ts
```diff
@@ -777,6 +778,198 @@ describe('LIVECHAT - dashboards', function () {
 		});
 	});
 
+	describe('[livechat/analytics/agent-overview] - Average first response time', () => {
+		let agent: { credentials: Credentials; user: IUser & { username: string } };
+		let originalFirstResponseTimeInSeconds: number;
+		let roomId: string;
+		const firstDelayInS = 4;
```
jessicaschelly, line 785: Just to follow the same pattern
```suggestion
		const firstDelayInSeconds = 4;
```

apps/meteor/tests/end-to-end/api/livechat/04-dashboards.ts
```diff
@@ -777,6 +778,198 @@ describe('LIVECHAT - dashboards', function () {
...
+		const secondDelayInS = 8;
```
jessicaschelly, line 786:
```suggestion
		const secondDelayInSeconds = 8;
```

apps/meteor/tests/end-to-end/api/livechat/04-dashboards.ts
```diff
@@ -777,6 +778,198 @@ describe('LIVECHAT - dashboards', function () {
...
+			expect(result.body).to.have.property('data');
+			expect(result.body.data).to.be.an('array');
+			expect(result.body.data).to.not.deep.include({ name: agent.user.username });
+		});
+
+		it('should only consider a first response has been sent when an agent sends a text message', async () => {
+			const response = await startANewLivechatRoomAndTakeIt({ agent: agent.credentials });
+			roomId = response.room._id;
+
+			const delayInS = 4;
```
jessicaschelly, line 912:
```suggestion
			const delayInSeconds = 4;
```

apps/meteor/tests/end-to-end/api/livechat/04-dashboards.ts
```diff
@@ -777,6 +778,198 @@ describe('LIVECHAT - dashboards', function () {
...
+				.expect(200);
+
+			expect(result.body).to.have.property('success', true);
+			expect(result.body).to.have.property('head');
+			expect(result.body).to.have.property('data');
+			expect(result.body.data).to.be.an('array');
+			expect(result.body.data).to.not.deep.include({ name: agent.user.username });
+		});
+
+		it('should only consider a first response has been sent when an agent sends a text message', async () => {
```
jessicaschelly, line 908: This is more of a question: does having the same name for different tests a good idea? Even if they are on different describes, maybe we could make them more unique? Line 815 and 908

matheusbsilva137: hmmm good suggestion. I'll change both descriptions and be more specific

jessicaschelly (approved):

matheusbsilva137:
Converting this PR to draft since tests will only pass after #33022 gets merged

apps/meteor/app/livechat/server/hooks/markRoomResponded.ts
```diff
@@ -12,7 +12,7 @@ export async function markRoomResponded(
 	room: IOmnichannelRoom,
 	roomUpdater: Updater<IOmnichannelRoom>,
 ): Promise<IOmnichannelRoom['responseBy'] | undefined> {
-	if (message.t || isEditedMessage(message) || isMessageFromVisitor(message)) {
+	if (message.t || isEditedMessage(message) || isMessageFromVisitor(message) || isSystemMessage(message)) {
```
ggazzo, line 15: I don't think there is any real difference in this change, we are already testing whether the `t` exists here

ggazzo (dismissed):

apps/meteor/tests/end-to-end/api/livechat/04-dashboards.ts
```diff
@@ -777,6 +778,198 @@ describe('LIVECHAT - dashboards', function () {
...
+				.query({ from: today, to: today, name: 'Avg_first_response_time' })
+				.set(credentials)
+				.expect('Content-Type', 'application/json')
+				.expect(200);
+
+			expect(result.body).to.have.property('success', true);
+			expect(result.body).to.have.property('head');
+			expect(result.body).to.have.property('data');
+			expect(result.body.data).to.be.an('array');
+			expect(result.body.data).to.not.deep.include({ name: agent.user.username });
```
KevLehman, line 812: if the test returns empty data, this check will return true all the time. Can you assert that the array has to have at least 1 element?

apps/meteor/tests/end-to-end/api/livechat/04-dashboards.ts
```diff
@@ -777,6 +778,198 @@ describe('LIVECHAT - dashboards', function () {
...
+				.query({ from: today, to: today, name: 'Best_first_response_time' })
+				.set(credentials)
+				.expect('Content-Type', 'application/json')
+				.expect(200);
+
+			expect(result.body).to.have.property('success', true);
+			expect(result.body).to.have.property('head');
+			expect(result.body).to.have.property('data');
+			expect(result.body.data).to.be.an('array');
+			expect(result.body.data).to.not.deep.include({ name: agent.user.username });
```
KevLehman, line 905: Same on this one.

apps/meteor/tests/end-to-end/api/livechat/04-dashboards.ts
KevLehman: We should do this change

KevLehman (changes requested):

apps/meteor/tests/end-to-end/api/livechat/04-dashboards.ts
matheusbsilva137: hmmm I'm thinking, is it really safe to assert that? At this point in my tests, I'm not even sure other agents exists (so that they can be returned here) -- so if I add this check here, it will depend on the result/what's done on previous tests
wdyt? I'm thinking of validating this in the **next** test instead, where I'm sure `data` shouldn't be empty

matheusbsilva137: I added this validation to the next test case. Please tell me what you think, @KevLehman

apps/meteor/tests/end-to-end/api/livechat/04-dashboards.ts
matheusbsilva137: I added this validation to the next test case. Please tell me what you think

KevLehman: Fair enough

KevLehman (dismissed):

apps/meteor/tests/end-to-end/api/livechat/04-dashboards.ts
matheusbsilva137: Oops, I skiped this, sorry
Now fixed

KevLehman (approved):

sampaiodiego (approved):

Issue #27200, opened 2022-11-05
[BUG] Slash Commands only show 50 (increase limit or delete default slash commands?)


Cauen:
I’ve created a **Rocketchat app** that includes **~45** new **Slash Commands**
But, on the frontend, the result is limited at **50**…
There are ~25 default slash commands that we dont use…
There is a way to increase the size of result or delete all default slash commands?

Note: new requests to fetch new commands are not sent as we type different commands

Server Setup Information
Version of Rocket.Chat Server: 5.1.3
Operating System: Ubuntu
Deployment Method: Docker
NodeJS Version: v14.19.3
MongoDB Version: 4.4.15
Proxy: traefik

dudanogueira:
Hi there, @Cauen !!

The [commands list comes from this endpoint](https://developer.rocket.chat/reference/api/rest-api/endpoints/core-endpoints/commands-endpoints/list), which leads to [this lines in codebase](https://github.com/RocketChat/Rocket.Chat/blob/3f5254d8a7a42f86bef9e945014d5c0dd3720c34/apps/meteor/app/api/server/v1/commands.ts#L138):

If you go down the path of searching for the getPaginationItems method, you will find this line:

https://github.com/RocketChat/Rocket.Chat/blob/f0c37c5a8c904181577b80d239255284ac6a1280/apps/meteor/app/api/server/helpers/getPaginationItems.ts#L12

with leads us to conclude that there is a setting for this default value: **API_Default_Count**

now, if you go to Admin > Settings, and search for this setting, you will get to here:
![image](https://user-images.githubusercontent.com/1761174/200412278-83d2d631-d076-47ca-8a22-9901bb8fcdfd.png)

And finally here, where you can change this setting:
![image](https://user-images.githubusercontent.com/1761174/200412563-953ae04c-c7f7-4db2-94d1-708b85424ece.png)

Let me know if this helps and if we can close this issue.

Also, check our [Community Resources](https://docs.rocket.chat/getting-support/community-resources) on how and where to ask for help.

Thanks!

dudanogueira:
The thing is: that command list API request is obeying that limiting setting.

And it shouldn't. It shouldn't iterate and get all items in all command item pages.

I think we can keep this open and maybe go for a fix on that.

Otherwise workspaces with 50+ commands will be broken with default settings.

PR #32847, opened 2024-07-19
fix: Not all Slash Commands are fetched by default

akshithere:
### What is Broken:
Workspaces with 50+ Slash Commands are broken with default settings.

### Why it is Broken:
Currently command list API request is obeying the limiting setting. _Which it shouldn't._  Instead, It should iterate and get all items in all command item pages.

### Details:
Now If I have 30 new slash commands in addition to the default 25, the total slash commands will be 55.
However the Default_Count setting on which this API depends by default allows only 50.
So the final 5 custom commands that i added in my RC app can't be used.

DefaultCount

I have to increase the Default_Count for it to temporarily work. 
_**However Never Loading the commands beyond that limit is not the expected behavior.**_

This PR aims to fetch all the slash commands regardless of the Default_Count setting of Rest API.

**Before:**

https://github.com/user-attachments/assets/b0f8db22-a32f-406b-9e73-5e42252fce31

**After:**

https://github.com/user-attachments/assets/4bdd019d-c978-4cc7-a45c-aad133d92f7f

## Issue(s)

#27200 

## Steps to test or reproduce
-Create A RC App
-Create around 30 new custom slash commands, enough to exceed the default limit.
-And you won't be able to use the last 5 slash commands (in the default case when Default_Count is set to 50)

## Further comments
NIL

### DIFF
--- a/apps/meteor/app/api/server/v1/commands.ts
+++ b/apps/meteor/app/api/server/v1/commands.ts
@@ -7,7 +7,6 @@ import { canAccessRoomIdAsync } from '../../../authorization/server/functions/ca
 import { slashCommands } from '../../../utils/server/slashCommand';
 import { API } from '../api';
 import { getLoggedInUser } from '../helpers/getLoggedInUser';
-import { getPaginationItems } from '../helpers/getPaginationItems';
 
 API.v1.addRoute(
 	'commands.get',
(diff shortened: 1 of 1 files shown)


Gustrb:
Hey @akshithere
Thanks for the Pull Request, this problem is already on our radar for quite some time, but there is a bit more nuance to it

We don't want to fetch *all* of the slash commands, because, if there are a lot of them, we might be sending too much data through the network, making it possibly too slow and heavy.
The ideal architecture is to solve this through frontend, to have some kind of infinite scroll-bar on the chat, and whenever we hit the "end" of the scroll we fetch more slash commands.

I dedicated some time to this issue, but due to more pressing concerns I did not manage to complete it, so if you want to do it together I would be happy :)

akshithere:
Hey @Gustrb 

Thanks a lot for the review and for outlining the issue in detail. The infinite scroll concept makes a lot of sense—much like how social media feeds handle large amounts of posts. I'm excited to get back to it and would love to collaborate with you on this! :D

Gustrb:
@akshithere I've sent you an e-mail so we can work on this one together :)

akshithere:
@Gustrb I couldn't find your email for some reason so i sent one by my side!

PR #32852, opened 2024-07-21
fix: disable deletion for federated users

debdutdeb:

This PR disables the possibility of delete federated users.
## Further comments

CORE-519


Gustrb:
Could you add API tests to assure the fix??

.changeset/empty-toys-smell.md
```diff
@@ -0,0 +1,5 @@
+---
+'@rocket.chat/meteor': minor
```
MarcosSpessatto, line 2:
```suggestion
'@rocket.chat/meteor': patch
```

apps/meteor/server/methods/deleteUser.ts
```diff
@@ -49,6 +49,19 @@ Meteor.methods<ServerMethods>({
 			});
 		}
 
+		if (user.federated) {
```
MarcosSpessatto, line 52: We have a helper to test this, it would be good to use this... https://github.com/RocketChat/Rocket.Chat/blob/develop/packages/core-typings/src/IUser.ts#L196

MarcosSpessatto (changes requested):

debdutdeb:
@Gustrb I don't think we can.

a federated user (`user.federated === true`) can not be created externally. This kind of record is internally created.

https://github.com/RocketChat/Rocket.Chat/blob/7b7b8a38d62a4d3932e93869b75bf120803472ae/apps/meteor/server/services/federation/domain/FederatedUser.ts#L45-L56

example event, before a dm is created - https://github.com/RocketChat/Rocket.Chat/blob/7aa0d963bb40cc794a2a6258db3f496c53a4084d/apps/meteor/ee/server/local-services/federation/application/room/sender/DirectMessageRoomServiceSender.ts#L47

If user is "external", then - https://github.com/RocketChat/Rocket.Chat/blob/c31773947ab51764ec2687482bcb872a609bdc93/apps/meteor/ee/server/local-services/federation/application/AbstractFederationApplicationServiceEE.ts#L20-L26

Right now, we *could* trigger for this user to be created. Problem is, we are blocking this in another pr until all of federation configuration is correct.

debdutdeb:
Similar is for a user who participated in federation, i.e. have a mapped user in synapse.

Gustrb:
I see, thanks for such a great explanation :)

apps/meteor/server/methods/deleteUser.ts
MarcosSpessatto, line 1: I wonder if it would not it be better to add this validation to the `deleteUser` function rather than the method, in the near future Methods will gone and the validation must still be valid, right? WDYT?

MarcosSpessatto (changes requested):

apps/meteor/server/methods/deleteUser.ts
debdutdeb: Makes sense to me :)

MarcosSpessatto (approved):

PR #32856, opened 2024-07-22
feat!: Implement Cloud communication reliability

Gustrb:

When a Workspace Registers with Cloud it gets a set of credentials it uses to generate access tokens and communicate with Cloud. Can read more about how this works in [ADR3](https://adr.rocket.chat/0003).

When a workspace owner needs to scale and deploys multiple Rocket.Chat instances to make up their deployment. Each workspace establishes a change stream request to mongo watching settings. Then updates its collection cache locally.

When a workspace goes to use an access token to talk to a cloud service it first evaluates if the token needs renewed. If it does it request a new one. Then updates a setting containing the access token. Then also another setting containing the expire.

The problem occurs when latency enters the picture. If due to either longer mongo query or network condition the other workspaces don't see both of these changes happen and update its cache. An instance can be using an expired token instead of renewing because the setting cache hasn't been updated.

We've seen many cases of this happen and typically restarting workspace can resolve.

This PR moves all credential related data to the 'workspace_credentials' collection, and stop using settings.get to fetch the configs that can be changed on multiple server instances.
This is a new PR that was created from both #32122 & #31986 in those PRs the migration was separated from the code itself, but that would end up being really hard to get merged properly, since the migration required the code, and the code assumed the migration was already ran, I'd guess we could avoid it by doing some weird PR splitting, but I am not really sure we would benefit a lot from it.

## Steps to test or reproduce

Create a new workspace, on registering, it should create a new record inside the rocketchat_workspace_credentials collection, containing an empty date as expirationDate, an empty string as accessToken and an empty string as scopes.
To test everything else, you must use the cloud functions, not sure how to call them from RC, but you could enter the Marketplace screen, go to paid apps, open one of them, and check if there is a new token in the database.
I would love some suggestions on how we can implement API tests for this feature, I am not sure at all if it is feasible (or if we should add later)

## Further comments

This implements: https://adr.rocket.chat/0061

[CONN-5]

[CONN-5]: https://rocketchat.atlassian.net/browse/CONN-5?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.changeset/plenty-hairs-camp.md
+++ b/.changeset/plenty-hairs-camp.md
@@ -1,0 +1,8 @@
+---
+"@rocket.chat/meteor": patch
+"@rocket.chat/core-typings": patch
+"@rocket.chat/model-typings": patch
+"@rocket.chat/models": patch
+---
+
+Now, we are not fetching cloud credentials from a cache, by doing so, we are avoiding the use of stale cloud credentials when comunicating with cloud services. We are, also, using a new collection to store the credentials and their scopes.
--- a/.changeset/plenty-hairs-camp.md
+++ b/.changeset/plenty-hairs-camp.md
@@ -5,4 +5,5 @@
 "@rocket.chat/models": major
 ---
 
-Now, we are not fetching cloud credentials from a cache, by doing so, we are avoiding the use of stale cloud credentials when comunicating with cloud services. We are, also, using a new collection to store the credentials and their scopes.
+Adds a new collection to store all the workspace cloud tokens to defer the race condition management to MongoDB instead of having to handle it within the settings cache.
+Removes the Cloud_Workspace_Access_Token & Cloud_Workspace_Access_Token_Expires_At settings since they are not going to be used anymore.
--- a/.changeset/plenty-hairs-camp.md
+++ b/.changeset/plenty-hairs-camp.md
@@ -1,8 +1,8 @@
 ---
-"@rocket.chat/meteor": patch
-"@rocket.chat/core-typings": patch
-"@rocket.chat/model-typings": patch
-"@rocket.chat/models": patch
+"@rocket.chat/meteor": major
+"@rocket.chat/core-typings": major
+"@rocket.chat/model-typings": major
+"@rocket.chat/models": major
 ---
 
 Now, we are not fetching cloud credentials from a cache, by doing so, we are avoiding the use of stale cloud credentials when comunicating with cloud services. We are, also, using a new collection to store the credentials and their scopes.
--- a/apps/meteor/app/cloud/server/functions/saveRegistrationData.ts
+++ b/apps/meteor/app/cloud/server/functions/saveRegistrationData.ts
@@ -5,7 +5,6 @@ import { notifyOnSettingChangedById } from '../../../lib/server/lib/notifyListen
 import { settings } from '../../../settings/server';
 import { syncCloudData } from './syncWorkspace/syncCloudData';
 
-
 type SaveRegistrationDataDTO = {
 	workspaceId: string;
 	client_name: string;
@@ -62,7 +61,7 @@ async function saveRegistrationDataBase({
 
 	const promises = [
 		...settingsData.map(({ _id, value }) => Settings.updateValueById(_id, value)),
-		WorkspaceCredentials.updateCredentialByScope('', '', new Date(0))
+		WorkspaceCredentials.updateCredentialByScope('', '', new Date(0)),
 	];
 
 	(await Promise.all(promises)).forEach((value, index) => {
--- a/apps/meteor/app/cloud/server/functions/saveRegistrationData.ts
+++ b/apps/meteor/app/cloud/server/functions/saveRegistrationData.ts
@@ -1,9 +1,22 @@
 import { applyLicense } from '@rocket.chat/license';
-import { Settings } from '@rocket.chat/models';
+import { Settings, WorkspaceCredentials } from '@rocket.chat/models';
 
 import { notifyOnSettingChangedById } from '../../../lib/server/lib/notifyListener';
 import { settings } from '../../../settings/server';
 import { syncCloudData } from './syncWorkspace/syncCloudData';
+
+
+type SaveRegistrationDataDTO = {
+	workspaceId: string;
+	client_name: string;
+	client_id: string;
+	client_secret: string;
+	client_secret_expires_at: number;
+	publicKey: string;
+	registration_client_uri: string;
+};
+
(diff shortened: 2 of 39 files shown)


apps/meteor/app/cloud/server/functions/saveRegistrationData.ts
```diff
@@ -63,7 +59,10 @@ async function saveRegistrationDataBase({
 		{ _id: 'Cloud_Workspace_Registration_Client_Uri', value: registration_client_uri },
 	];
 
-	const promises = settingsData.map(({ _id, value }) => Settings.updateValueById(_id, value));
+	const promises = [
+		...settingsData.map(({ _id, value }) => Settings.updateValueById(_id, value)),
+		WorkspaceCredentials.updateCredentialByScope('', '', new Date(0)),
```
MarcosSpessatto, line 64: Maybe change this to an object parameter, so it's easier to understand what the empty strings are.

MarcosSpessatto (changes requested):
Overall the PR looks good to me, I just have some concerns:

- Can we add some sort of automated tests here?
- We are using the model instead of the cached settings to fetch these specific values because it's causing some bugs, but there is nothing preventing this from happening in the future, or people will not know about that problem and can still get those values from the cached settings. What bothers me is that this is turning into a really manual process, depending entirely on the reviewer, if the reviewer knows about this problem... Otherwise the same will happen

MarcosSpessatto:
> > Overall the PR looks good to me, I just have some concerns:
> > 
> > * Can we add some sort of automated tests here?
> > * We are using the model instead of the cached settings to fetch these specific values because it's causing some bugs, but there is nothing preventing this from happening in the future, or people will not know about that problem and can still get those values from the cached settings. What bothers me is that this is turning into a really manual process, depending entirely on the reviewer, if the reviewer knows about this problem... Otherwise the same will happen
> 
> I am not really sure about automated tests, I don't know how could we test something like that And I just saw there are some settings that I changed that should not be changed, only the credentials should be affected since they are the ones causing this issue
> 
> I'll rollback them

Maybe some unit tests, because e2e using the cloud is impractical...

MarcosSpessatto (dismissed):

.changeset/plenty-hairs-camp.md
```diff
@@ -0,0 +1,8 @@
+---
+"@rocket.chat/meteor": major
+"@rocket.chat/core-typings": major
+"@rocket.chat/model-typings": major
+"@rocket.chat/models": major
+---
+
+Now, we are not fetching cloud credentials from a cache, by doing so, we are avoiding the use of stale cloud credentials when comunicating with cloud services. We are, also, using a new collection to store the credentials and their scopes.
```
KevLehman, line 8: Pls change the wording of changeset to follow guidelines

apps/meteor/ee/server/models/raw/WorkspaceCredentials.ts
```diff
@@ -0,0 +1,68 @@
...
+export class WorkspaceCredentialsRaw extends BaseRaw<IWorkspaceCredentials> implements IWorkspaceCredentialsModel {
+	constructor(db: Db, trash?: Collection<RocketChatRecordDeleted<IWorkspaceCredentials>>) {
+		super(db, 'workspace_credentials', trash);
+	}
+
+	protected modelIndexes(): IndexDescription[] {
+		return [{ key: { _id: 1, scopes: 1, expirationDate: 1, accessToken: 1 }, unique: true }];
+	}
+
+	getCredentialByScope(scope: string): Promise<IWorkspaceCredentials | null> {
```
KevLehman, line 16:
```suggestion
	getCredentialByScope(scope = ''): Promise<IWorkspaceCredentials | null> {
```

This way, you don't have to pass `''` on all the calls to get/unset credentials

apps/meteor/ee/server/models/raw/WorkspaceCredentials.ts
```diff
@@ -0,0 +1,68 @@
...
+
+import { BaseRaw } from '../../../../server/models/raw/BaseRaw';
+
+export class WorkspaceCredentialsRaw extends BaseRaw<IWorkspaceCredentials> implements IWorkspaceCredentialsModel {
+	constructor(db: Db, trash?: Collection<RocketChatRecordDeleted<IWorkspaceCredentials>>) {
+		super(db, 'workspace_credentials', trash);
+	}
+
+	protected modelIndexes(): IndexDescription[] {
+		return [{ key: { _id: 1, scopes: 1, expirationDate: 1, accessToken: 1 }, unique: true }];
```
KevLehman, line 13: Since _id is already unique, you would need an actual duplicated document for this index to trigger the unique warning :/

If you remove the _id from it, then it would properly validate that scopes, expiration and access token should be unique on any document

apps/meteor/ee/server/models/WorkspaceCredentials.ts
```diff
@@ -0,0 +1,7 @@
+import { registerModel } from '@rocket.chat/models';
+
+import { trashCollection } from '../../../server/database/trash';
+import { db } from '../../../server/database/utils';
+import { WorkspaceCredentialsRaw } from './raw/WorkspaceCredentials';
+
+registerModel('IWorkspaceCredentialsModel', new WorkspaceCredentialsRaw(db, trashCollection));
```
KevLehman, line 7: do we need to keep old tokens in trash? if not, we can omit the trashcollection

apps/meteor/server/startup/migrations/v315.ts
```diff
@@ -0,0 +1,31 @@
...
+
+		if (accessTokenExpiresAt) {
+			await Settings.removeById('Cloud_Workspace_Access_Token_Expires_At');
+		}
+
+		await WorkspaceCredentials.updateCredentialByScope({
+			scope: '',
+			accessToken,
+			expirationDate: accessTokenExpiresAt,
+		});
```
KevLehman, line 29:
```suggestion
		const expirationDate = ((await Settings.getValueById('Cloud_Workspace_Access_Token_Expires_At')) as Date) || new Date(0);

		if (accessToken) {
			await Settings.removeById('Cloud_Workspace_Access_Token');
		}

		if (expirationDate) {
			await Settings.removeById('Cloud_Workspace_Access_Token_Expires_At');
		}

		await WorkspaceCredentials.updateCredentialByScope({
			scope: '',
			accessToken,
			expirationDate,
		});
```

KevLehman (changes requested):

MarcosSpessatto (approved):

PR #32868, opened 2024-07-23
feat: add `sidepanel` to teams create

juliajforesti:

### DIFF
--- a/apps/meteor/app/api/server/v1/teams.ts
+++ b/apps/meteor/app/api/server/v1/teams.ts
@@ -82,10 +82,11 @@ API.v1.addRoute(
 					members: Match.Maybe([String]),
 					room: Match.Maybe(Match.Any),
 					owner: Match.Maybe(String),
+					sidepanel: Match.Maybe(Match.Any),
 				}),
 			);
 
-			const { name, type, members, room, owner } = this.bodyParams;
+			const { name, type, members, room, owner, sidepanel } = this.bodyParams;
 
 			const team = await Team.create(this.userId, {
 				team: {
@@ -95,6 +96,7 @@ API.v1.addRoute(
 				room,
 				members,
 				owner,
+				sidepanel,
 			});
 
 			return API.v1.success({ team });
--- a/packages/core-typings/src/IRoom.ts
+++ b/packages/core-typings/src/IRoom.ts
@@ -7,6 +7,7 @@ import type { IUser, Username } from './IUser';
 import type { RoomType } from './RoomType';
 
 type CallStatus = 'ringing' | 'ended' | 'declined' | 'ongoing';
+export type SidepanelItems = 'channel' | 'discussion';
 
 export type RoomID = string;
 export type ChannelName = string;
@@ -95,6 +96,10 @@ export interface IRoom extends IRocketChatRecord {
 	customFields?: Record<string, any>;
 
 	usersWaitingForE2EKeys?: { userId: IUser['_id']; ts: Date }[];
+
+	sidepanel?: {
+		items: [SidepanelItems, SidepanelItems?];
+	};
 }
 
 export const isRoomWithJoinCode = (room: Partial<IRoom>): room is IRoomWithJoinCode =>
--- a/apps/meteor/app/lib/server/functions/createRoom.ts
+++ b/apps/meteor/app/lib/server/functions/createRoom.ts
@@ -112,6 +112,7 @@ export const createRoom = async <T extends RoomType>(
 	readOnly?: boolean,
 	roomExtraData?: Partial<IRoom>,
 	options?: ICreateRoomParams['options'],
+	sidepanel?: ICreateRoomParams['sidepanel'],
 ): Promise<
 	ICreatedRoom & {
 		rid: string;
@@ -187,6 +188,7 @@ export const createRoom = async <T extends RoomType>(
 		},
 		ts: now,
 		ro: readOnly === true,
+		sidepanel,
 	};
 
 	if (teamId) {
--- a/apps/meteor/server/services/room/service.ts
+++ b/apps/meteor/server/services/room/service.ts
@@ -16,7 +16,7 @@ export class RoomService extends ServiceClassInternal implements IRoomService {
 	protected name = 'room';
 
 	async create(uid: string, params: ICreateRoomParams): Promise<IRoom> {
-		const { type, name, members = [], readOnly, extraData, options } = params;
+		const { type, name, members = [], readOnly, extraData, options, sidepanel } = params;
 
 		const hasPermission = await Authorization.hasPermission(uid, `create-${type}`);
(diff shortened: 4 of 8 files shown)


apps/meteor/app/api/server/v1/teams.ts
```diff
@@ -82,10 +82,11 @@ API.v1.addRoute(
 					members: Match.Maybe([String]),
 					room: Match.Maybe(Match.Any),
 					owner: Match.Maybe(String),
+					sidepanel: Match.Maybe(Match.Any),
```
matheusbsilva137, line 85: We're trying to avoid using `check` for new params since we'll remove it from all endpoints on 7.0. Can you validate this param using an if statement instead? It'd be great if we could check if it contains valid values in the items field

matheusbsilva137 (changes requested):
Also, can you please add more context in the PR description about how this change should work?

apps/meteor/app/api/server/v1/teams.ts
```diff
@@ -85,7 +85,17 @@ API.v1.addRoute(
...
+
+			if (
+				sidepanel?.items &&
+				(!Array.isArray(sidepanel.items) ||
+					!sidepanel.items.length ||
+					!sidepanel.items.every(isSidepanelItem) ||
+					sidepanel.items.length > 2)
+			) {
+				throw new Error('error-invalid-sidepanel');
+			}
```
matheusbsilva137, line 98: Can you please add some end-to-end tests to validate that this is working fine? And also for the case when it should be accepted

matheusbsilva137 (changes requested):

apps/meteor/app/api/server/v1/teams.ts
juliajforesti: done! thanks 😄

apps/meteor/tests/end-to-end/api/rooms.ts
```diff
@@ -2290,6 +2300,58 @@ describe('[Rooms]', () => {
...
+				.set(credentials)
+				.send({
+					rid: testTeam.roomId,
+					sidepanel: { items: sidepanelItems },
+				});
+
+			expect('Content-Type', 'application/json');
+			expect(400);
+		});
+		it('should throw error when updating team sidepanel more than 2 items', async () => {
```
matheusbsilva137, line 2342: What if duplicated values are provided? Should we accept that? :thinking:

apps/meteor/tests/end-to-end/api/teams.ts
```diff
@@ -1926,6 +1990,33 @@ describe('[Teams]', () => {
...
+				.post(api('rooms.saveRoomSettings'))
+				.set(credentials)
+				.send({
+					rid: testTeam.roomId,
+					sidepanel: { items: sidepanelItems },
+				});
+
+			expect('Content-Type', 'application/json');
+			expect(200);
+			expect(response.body).to.have.property('success', true);
```
matheusbsilva137, line 2005: I believe you meant to use `teams.update` here, didn't you? :thinking: 
It'd be good to have tests for when `teams.update` and `teams.create` endpoints should throw errors here too

matheusbsilva137 (changes requested):

apps/meteor/tests/end-to-end/api/teams.ts
juliajforesti: actually no, I moved this test to `apps/meteor/tests/end-to-end/api/rooms.ts` but forgot to delete it here.
why: teams.update only refers to `teams` properties: name and type - which is not the case. `sidepanel` is being saved on the `room` object, therefore, using the `rooms.saveRoomSettings`
and there I added more tests to the error cases

apps/meteor/tests/end-to-end/api/rooms.ts
juliajforesti: No we shouldn't accept duplicated items. I'll add a test

apps/meteor/tests/end-to-end/api/teams.ts
matheusbsilva137: Got it, so teams.update would only be another way to trigger saveRoomSettings. We should be fine by covering tests on that side, then :+1:

packages/core-typings/src/IRoom.ts
```diff
@@ -95,8 +97,29 @@ export interface IRoom extends IRocketChatRecord {
...
+export const isValidSidepanel = (sidepanel: IRoom['sidepanel']) => {
+	if (!sidepanel?.items) {
+		return false;
+	}
+	return (
+		Array.isArray(sidepanel.items) &&
+		sidepanel.items.length &&
+		sidepanel.items.every(isSidepanelItem) &&
+		sidepanel.items.length <= 2 &&
+		sidepanel.items.length === new Set(sidepanel.items).size
```
matheusbsilva137, line 119:
```suggestion
		sidepanel.items.length === new Set(sidepanel.items).size
```
I believe the second condition makes us not need the first one

apps/meteor/tests/end-to-end/api/teams.ts
```diff
@@ -172,6 +172,84 @@ describe('[Teams]', () => {
...
+		});
+		it('should throw error when creating a team with sidepanel with duplicated items', async () => {
+			await request
+				.post(api('teams.create'))
+				.set(credentials)
+				.send({
+					name: `test-team-with-sidepanel-error-${Date.now()}`,
+					type: 0,
+					sidepanel: {
+						items: ['other'],
```
matheusbsilva137, line 247:
```suggestion
						items: ['channels', 'channels'],
```
Weren't the items here supposed to be duplicated? :thinking:

matheusbsilva137 (changes requested):

packages/core-typings/src/IRoom.ts
juliajforesti: noooot really, the second is to ensure there are no duplicates on the array - if there are, the length of the `new Set` will be different, because it will remove the duplicate.
the first oder condition is to ensure it has max 2 items :)

apps/meteor/tests/end-to-end/api/teams.ts
juliajforesti: yes indeed haha thanks!

packages/core-typings/src/IRoom.ts
matheusbsilva137: hmmmm but if there are no duplicates and all items are valid, it should be safe to say the array has max 2 elements. shouldn't it? :thinking:

juliajforesti: hammm you're right!

matheusbsilva137 (approved):

ggazzo (approved):

PR #32885, opened 2024-07-23
refactor: Circular imports

tassoevan:


apps/meteor/app/2fa/server/methods/checkCodesRemaining.ts
```diff
@@ -1,7 +1,7 @@
-import type { ServerMethods } from '@rocket.chat/ui-contexts';
+import type { ServerMethods } from '@rocket.chat/ddp-client';
```
pierre-lehnen-rc, line 1: ddp-client is an EE package; let's avoid adding a ton of new references to it from non-ee code.

pierre-lehnen-rc (changes requested):

apps/meteor/app/2fa/server/methods/checkCodesRemaining.ts
pierre-lehnen-rc: dismissing my previous review as we'll be moving the ddp-client package out of the ee folder.

pierre-lehnen-rc (approved):
While reviewing I noticed some things (unaffected by this PR, but related to the same stuff that was changed):

1. We stil have a few meteor methods on the client, but they are using the types defined on the `ServerMethods` interface.
2. Some of the apps on /ee/apps are importing files from the main meteor app and those files in turn are importing other packages like `@rocket.chat/string-helpers`, which are not being copied along on the app's Dockerfile. If something on those apps tried to call one of the functions imported from those packages, it would crash.

PR #32886, opened 2024-07-24
regression: Sending message when choose emoji from ComposerPopupBox

tiagoevanp:

Fix message being sent by "Enter" key when choosing a emoji from ComposerPopupBox list

Introduced: #32861


ggazzo:
could you please add when this was introduced?

PR #32910, opened 2024-07-26
chore(client): stop replacing  `Meteor.user`

ggazzo:
After a change in Meteor (meteor 3.0.0), where `Accounts.callLoginMethod`  trusts that `Accounts.user.find`  is still there to call the login callbacks, it became clear that it is not healthy to rewrite internal framework objects.

...
...
But no one has proof that adding extra methods causes problems :)

### DIFF
--- a/apps/meteor/app/ui-cached-collection/client/models/CachedCollectionManager.ts
+++ b/apps/meteor/app/ui-cached-collection/client/models/CachedCollectionManager.ts
@@ -11,15 +11,12 @@ class CachedCollectionManager extends Emitter<{ reconnect: void; login: string |
 
 	private _syncEnabled: boolean;
 
-	private logged: boolean;
-
 	private step: number;
 
 	constructor() {
 		super();
 		this.items = [];
 		this._syncEnabled = false;
-		this.logged = false;
 
 		const { _unstoreLoginToken } = Accounts;
 		Accounts._unstoreLoginToken = (...args) => {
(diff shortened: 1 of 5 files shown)


apps/meteor/app/ui-cached-collection/client/models/CachedCollectionManager.ts
```diff
@@ -42,10 +42,10 @@ class CachedCollectionManager extends Emitter<{ reconnect: void; login: string |
 		});
 
 		Tracker.autorun(() => {
-			const uid = Meteor.userId();
-			this.logged = uid !== null;
-			if (this.logged) {
-				this.emit('login', uid);
+			const user = Meteor.user();
+			if (user?._id) {
+				this.logged = true;
```
ggazzo, line 47:
```suggestion
			this.logged = Boolean(user?._id);
			if (this.logged) {
```

apps/meteor/client/views/root/SAMLLoginRoute.tsx
```diff
@@ -38,20 +38,6 @@ const SAMLLoginRoute = () => {
...
-		}
-
-		router.navigate(
-			{
-				pathname: '/home',
-			},
-			{ replace: true },
-		);
-	}, [userId, router]);
-
```
matheusbsilva137, line 54: Just trying to understand: why are we removing this?
Shouldn't we keep redirecting the user to the homepage in case they are already logged in RC when accessing this page?

ggazzo: Pierre had done a workaround, for some reason, in Meteor 3.0.0 this callback would never be called. However, in Meteor 2.0.0 it happens. 

If you stop to think about it, it doesn't make sense to redirect the user after logging in if he doesn't have a uid, because that means it failed.

Likewise, it doesn't make sense to redirect twice if it works.

You even did a test to see if it redirects to the home page twice, that's going to be quite interesting.

dougfabris (approved):

PR #32912, opened 2024-07-26
chore: remove afterSaveMessage save-visitor-inactivity hook

ricardogarim:
As per [OPI-13](https://rocketchat.atlassian.net/browse/OPI-13), this pull request removes the `afterSaveMessage - save-visitor-inactivity` hook and transfers its only call to the `afterSaveMessage - markRoomResponded` hook. The goal is to streamline the call stack and eliminate unnecessary database calls.

[OPI-13]: https://rocketchat.atlassian.net/browse/OPI-13?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


sampaiodiego (changes requested):
we can't do exact this.. this was done with a callback because the callback is EE, so doing this way adds no dependency on EE code.. but removing the callback and calling the EE function directly creates a dependency we can't add.

apps/meteor/app/livechat/server/hooks/markRoomResponded.ts
```diff
@@ -71,6 +72,7 @@ callbacks.add(
 
 		// this unsets waitingResponse and sets responseBy object
 		await LivechatRooms.setResponseByRoomId(room._id, responseBy);
+		await setPredictedVisitorAbandonmentTime({ _id: room._id, responseBy, departmentId: room.departmentId });
```
ggazzo, line 75: this is a `ee` feature so I'm afraid we cannot move this code to here

PR #32916, opened 2024-07-26
feat(audit): New endpoint for listing members of any room

KevLehman:

- New endpoint created `/audit/rooms.members`
- Allows users with `view-members-list-all-rooms` permission to audit _any_ room (public or private) and see the members of them.
- Part of `Audit` EE module

## Issue(s)

https://rocketchat.atlassian.net/browse/CORE-565


packages/i18n/src/locales/en.i18n.json
```diff
@@ -5792,6 +5792,8 @@
   "view-all-teams_description": "Permission to view all teams",
   "view-all-team-channels": "View All Team Channels",
   "view-all-team-channels_description": "Permission to view all team's channels",
+  "view-members-list-all-rooms": "View Member's list for all rooms",
```
KevLehman, line 5795: note: checking on these. so expect strings to not be final

.changeset/funny-boats-guess.md
```diff
@@ -0,0 +1,6 @@
+---
+"@rocket.chat/meteor": patch
+"@rocket.chat/i18n": patch
```
matheusbsilva137, line 3:
```suggestion
"@rocket.chat/meteor": minor
"@rocket.chat/i18n": minor
```

apps/meteor/ee/server/lib/audit/endpoints.ts
matheusbsilva137, line 1: Shouldn't we move this file to under `apps/meteor/ee/server/api` or `apps/meteor/ee/app/api-enterprise/server`? :thinking:

apps/meteor/ee/server/lib/audit/endpoints.ts
```diff
@@ -0,0 +1,76 @@
...
+	properties: {
+		roomId: { type: 'string' },
+		filter: { type: 'string', nullable: true },
+		count: { type: 'number', nullable: true },
+		offset: { type: 'number', nullable: true },
+		sort: { type: 'string', nullable: true },
+	},
+	required: ['roomId'],
+	additionalProperties: false,
+};
```
matheusbsilva137, line 30: I'm quite confused about what is the best approach here, but I thought we could define this under the rest-typings package. I see that we do this in some EE endpoints, but for some of them we define the schema in the EE folder as you did, so I'm not sure
My personal preference would be to define it within the package since we already have ajv instantiated there and a bunch of EE endpoints iinw

@MarcosSpessatto any preference here?

apps/meteor/ee/server/lib/audit/endpoints.ts
```diff
@@ -0,0 +1,76 @@
...
+
+type AuditRoomMembersParams = PaginatedRequest<{
+	roomId: string;
+	filter: string;
+}>;
+
+const auditRoomMembersSchema = {
+	type: 'object',
+	properties: {
+		roomId: { type: 'string' },
```
matheusbsilva137, line 22:
```suggestion
		roomId: { type: 'string', minLength: 1 },
```
Using `minLength` avoids finding a random room in the DB, which could be dangerous

apps/meteor/tests/end-to-end/api/audit.ts
```diff
@@ -0,0 +1,239 @@
...
+		testPrivateChannel = (await createRoom({ type: 'p', name: `chat.api-test-${Date.now()}` })).body.group;
+		dummyUser = await createUser();
+		auditor = await createUser({ roles: ['user', 'auditor'] });
+
+		auditorCredentials = await login(auditor.username, password);
+	});
+	after(() => deleteRoom({ type: 'c', roomId: testChannel._id }));
+	after(() => deleteUser({ _id: dummyUser._id }));
+	after(() => deleteUser({ _id: auditor._id }));
+	after(() => deleteRoom({ type: 'p', roomId: testPrivateChannel._id }));
```
matheusbsilva137, line 32: This might be a personal preference, but why not do this into a single after using `Promise.all`?

apps/meteor/tests/end-to-end/api/audit.ts
```diff
@@ -0,0 +1,239 @@
...
+
+	describe('audit/rooms.members [no permissions]', () => {
+		before(async () => {
+			await updatePermission('view-members-list-all-rooms', []);
+		});
+		after(async () => {
+			await updatePermission('view-members-list-all-rooms', ['admin', 'auditor']);
+		});
```
matheusbsilva137, line 40: We don't even need `async/await` here as you've done above

apps/meteor/tests/end-to-end/api/audit.ts
```diff
@@ -0,0 +1,239 @@
...
+	describe('audit/rooms.members', () => {
+		it('should fail if user is not logged in', async () => {
+			await request
+				.get(api('audit/rooms.members'))
+				.query({
+					roomId: 'GENERAL',
+				})
+				.expect(401);
+		});
+		it('should fail if roomId is invalid', async () => {
```
matheusbsilva137, line 68: Please add a test case for when `roomId` is an empty string

apps/meteor/ee/server/lib/audit/endpoints.ts
```diff
@@ -0,0 +1,76 @@
...
+
+			const { cursor, totalCount } = findUsersOfRoom({
+				rid: room._id,
+				filter,
+				skip,
+				limit,
+				...(sort?.username && { sort: { username: sort.username } }),
+			});
+
+			const [members, total] = await Promise.all([cursor.toArray(), totalCount]);
```
matheusbsilva137, line 67: We should have logs for when this action/query is triggered :grimacing: 
Try extracting this into an `auditGetRoomMembers` function (as we do for other audit actions in `apps/meteor/ee/server/lib/audit/methods.ts`) and use the `AuditLog` model to insert the logs
It'd be good to have e2e tests to make sure the logs for this action are working fine, if possible

packages/i18n/src/locales/en.i18n.json
```diff
@@ -5792,6 +5792,8 @@
   "view-all-teams_description": "Permission to view all teams",
   "view-all-team-channels": "View All Team Channels",
   "view-all-team-channels_description": "Permission to view all team's channels",
+  "view-members-list-all-rooms": "Can view members in all rooms",
+  "view-members-list-all-rooms_description": "Gives the ability to see the members list in all rooms, even those the user is not part of",
```
matheusbsilva137, line 5796: Not requesting any specific change, but I just wanted to confirm whether this has been approved by our UX/UI team. If not, it'd be good to have their take on this too (otherwise you can resolve this :slightly_smiling_face: )

matheusbsilva137 (changes requested):

apps/meteor/ee/server/lib/audit/endpoints.ts
KevLehman: The idea is to not to have these on core typings/rest typings cause they're EE.

There was some Tasso's PR to moce the majority time ago.

apps/meteor/tests/end-to-end/api/audit.ts
KevLehman: Actually cannot decide on which style lol. I'm ok with both so 🤷🏽

apps/meteor/ee/server/lib/audit/endpoints.ts
KevLehman: Checked with Milton on thread, no need.

packages/i18n/src/locales/en.i18n.json
KevLehman: Yep, there's actually a comment by me some lines above

apps/meteor/tests/end-to-end/api/audit.ts
KevLehman: Yeah, feel personal pref. A promise.all may work better. Haven't chrcked how this multiple afters are done on mocha so they may be doing the same? Gotta check.

apps/meteor/ee/server/lib/audit/endpoints.ts
KevLehman: Good one.

apps/meteor/ee/server/lib/audit/endpoints.ts
KevLehman: This is the folder for audit. So feel things on audit should be grouped. If this folder was named methods then we probably have to move somewhere else but for this one, no need actually.

apps/meteor/ee/server/lib/audit/endpoints.ts
KevLehman: Note: in this case, this won't find a "random room". The param `roomId` will end up as `''` as it's required. And then `Rooms.findOneById` will issue `{ _id: '' }` which won't match any room and return `null`.

I'm removing the non-obvs behavior but felt good to clarify this part 👀 

If we did something like:
```
findOne({ ...(_id && { _id })
then when _id is falsy it will issue a `findOne({})` which will return the first room in the db. Since here we're passing the param to the call, it won't happen.

File: packages/core-typings/src/ee/IAuditLog.ts [hunk 10]
```diff
@@ -12,12 +12,13 @@ export interface IAuditLog extends IRocketChatRecord {
 	fields: {
 		type: string;
```
matheusbsilva137, line 13: Since we're already changing this type, can we be more specific on this one? It'd be good to have the new type you just created mapped somewhere

packages/i18n/src/locales/en.i18n.json
```diff
@@ -5792,6 +5792,9 @@
...
+  "Room_members_list": "Member's list",
```
matheusbsilva137, line 5797:
```suggestion
  "Room_members_list": "Members list",
```

apps/meteor/ee/server/lib/audit/endpoints.ts
```diff
@@ -0,0 +1,95 @@
...
+}>;
+
+const auditRoomMembersSchema = {
+	type: 'object',
+	properties: {
+		roomId: { type: 'string', minLength: 1 },
+		filter: { type: 'string', nullable: true },
+		count: { type: 'number', nullable: true },
+		offset: { type: 'number', nullable: true },
+		sort: { type: 'string', nullable: true },
```
matheusbsilva137, line 26: This is not really an issue, but do we really need these params to be nullable? I believe just not being required should be fine :thinking:

matheusbsilva137 (changes requested):

packages/core-typings/src/ee/IAuditLog.ts
KevLehman: Not sure about that. My changes makes Type a constant one but only for my case. When auditing messages Type can apparently be,any string (or room type i guess)

We should better look into this type or we'll create an unmantainable one if we add more audit types  👀

apps/meteor/ee/server/lib/audit/endpoints.ts
```diff
@@ -0,0 +1,95 @@
...
+type AuditRoomMembersParams = PaginatedRequest<{
+	roomId: string;
+	filter: string;
+}>;
+
+const auditRoomMembersSchema = {
+	type: 'object',
+	properties: {
+		roomId: { type: 'string', minLength: 1 },
+		filter: { type: 'string', nullable: true },
```
KevLehman, line 23:
```suggestion
		filter: { type: 'string' },
```

apps/meteor/ee/server/lib/audit/endpoints.ts
```diff
@@ -0,0 +1,95 @@
...
+		count: { type: 'number', nullable: true },
```
KevLehman, line 24:
```suggestion
		count: { type: 'number' },
```

apps/meteor/ee/server/lib/audit/endpoints.ts
```diff
@@ -0,0 +1,95 @@
...
+		offset: { type: 'number', nullable: true },
```
KevLehman, line 25:
```suggestion
		offset: { type: 'number' },
```

apps/meteor/ee/server/lib/audit/endpoints.ts
KevLehman, line 26:
```suggestion
		sort: { type: 'string' },
```

matheusbsilva137 (approved):

apps/meteor/ee/server/lib/audit/endpoints.ts
MarcosSpessatto: I agree with @matheusbsilva137. I know we don't have defined a pattern for that, but the majority of the endpoints are under some `api` folder, so I think we should follow that "pattern"

apps/meteor/ee/server/lib/audit/endpoints.ts
```diff
@@ -0,0 +1,95 @@
...
+		filter: { type: 'string' },
+		count: { type: 'number' },
+		offset: { type: 'number' },
+		sort: { type: 'string' },
+	},
+	required: ['roomId'],
+	additionalProperties: false,
+};
+
+export const isAuditRoomMembersProps = ajv.compile<AuditRoomMembersParams>(auditRoomMembersSchema);
```
MarcosSpessatto, line 32: This is quite confusing for me as well. We have the exact same case as discussed above (about the place where this file should live), and again, since we don't have that "pattern" defined, we do have a bunch of ajv validation for EE endpoints living in the rest-typings folder, I'd go with the "pattern" we are using until now.

apps/meteor/ee/server/lib/audit/endpoints.ts
```diff
@@ -0,0 +1,95 @@
...
+	{ authRequired: true, permissionsRequired: ['view-members-list-all-rooms'], validateParams: isAuditRoomMembersProps },
+	{
+		async get() {
+			const { roomId, filter } = this.queryParams;
+			const { count: limit, offset: skip } = await getPaginationItems(this.queryParams);
+			const { sort } = await this.parseJsonQuery();
+
+			const room = await Rooms.findOneById<Pick<IRoom, '_id' | 'name' | 'fname'>>(roomId, { projection: { _id: 1, name: 1, fname: 1 } });
+			if (!room) {
+				return API.v1.notFound();
```
MarcosSpessatto, line 56: Should we include some message here?

apps/meteor/ee/server/startup/audit.ts
```diff
@@ -4,6 +4,7 @@ import { createPermissions } from '../lib/audit/startup';
 
 await License.onLicense('auditing', async () => {
 	await import('../lib/audit/methods');
+	await import('../lib/audit/endpoints');
```
MarcosSpessatto, line 7: See previous discussion.

julio-cfa (dismissed):

apps/meteor/tests/end-to-end/api/audit.ts
```diff
@@ -0,0 +1,301 @@
...
+		it('should fail if roomId is an empty string', async () => {
+			await request
+				.get(api('audit/rooms.members'))
+				.set(credentials)
+				.query({
+					roomId: '',
+				})
+				.expect(400);
+		});
+		it('should fetch the members of a room', async () => {
```
MarcosSpessatto, line 92: Since we didn't provide any credentials to the createRoom function, I believe the default(admin) one is being used, having said that, by using `credentials` here as well, isn't this user a member of the room? Or was it by design maybe?

apps/meteor/tests/end-to-end/api/audit.ts
```diff
@@ -0,0 +1,301 @@
...
+
+	describe('audit/rooms.members [no permissions]', () => {
+		before(async () => {
+			await updatePermission('view-members-list-all-rooms', []);
+		});
+		after(async () => {
+			await updatePermission('view-members-list-all-rooms', ['admin', 'auditor']);
+		});
+		it('should fail if user does not have view-members-list-all-rooms permission', async () => {
+			await request
```
MarcosSpessatto, line 45: Let's split these into two different `it` since (correct me if I'm wrong) one user is a room member and the other is not.

MarcosSpessatto (changes requested):

apps/meteor/ee/server/lib/audit/endpoints.ts
KevLehman: I'd prefer to bring @tassoevan around since the last time i did the opposite I was told to do this way and would prefer to have an standard def rather than 2 diff opinions.

apps/meteor/ee/server/lib/audit/endpoints.ts
KevLehman: A 404 should be self explanatory.

apps/meteor/tests/end-to-end/api/audit.ts
KevLehman: Yup, there's a case for a user not part of room at the end. These are for making sure the endpoint works, since it returns the same data, who calls it is not important.

But there's a test for checking that users outside with permissions can do

apps/meteor/tests/end-to-end/api/audit.ts
KevLehman: Not really sure. Both requests are doing the same thing even if one user is not part of the room.

So, the `it` is true for both cases, it should fail if the user doesnt have permission, which is the action being tested, no matter if user is part of the room or not.

apps/meteor/ee/server/lib/audit/endpoints.ts
KevLehman: From Tasso's old discussion:

```
It's something you're trying to push more. The augmentation itself is needed because some APIs were born to be extensible (e.g. API.addRoute, MessageAction.add etc.)
Having a single place to drop all the signatures is impractical as we have been splitting the source code in packages and sometimes behind a flag (e.g. Enterprise code)
Our first attempt to enforce typings on APIs like it was with the REST API endpoints
A single interface held all the call signatures
Then Pierre started working on moving the models to new packages
It imposed a challenged the we haven't thinking at that point: how to split the interfaces and the implementations?
That's why we have models and model-typings
As we learned that type imports are a thing, it was obvious it could be only models
Therefore a module that can't import any model still can do import type { ... } from '@rocket.chat/models'
Without changing the final JS output
So we thought "what would happen if we create a package adding more endpoints to the API?"
:hmmm:
1
Then we merged two concepts: (1) augment interfaces while (2) importing implementations

Comment (tassoevan) [reply apps/meteor/ee/server/lib/audit/endpoints.ts]: @MarcosSpessatto this is kinda related to the last thread on the architecture channel (from @pierre-lehnen-rc). For now EE endpoints don't have a definitive place to be defined (types and with schema validations). There is an intent of making a single package for FOSS endpoints, Meteor methods, and streams, letting non-FOSS splitted elsewhere by domain (e.g. a `@rocket.chat/rest-typings-ee` should **NOT** exist, but those types and schemas could reside in something like `@rocket.chat/auditing`).

**Review by julio-cfa — APPROVED** [review]

**Review by matheusbsilva137 — APPROVED** [review]

File: apps/meteor/ee/server/lib/audit/endpoints.ts (hunk 16 above)
Comment (MarcosSpessatto) [reply apps/meteor/ee/server/lib/audit/endpoints.ts]: Awesome, thanks for the explanation guys!

**Review by MarcosSpessatto — APPROVED** [review]

File: apps/meteor/ee/server/api/audit.ts [hunk 21]
```diff
@@ -0,0 +1,95 @@
...
+		offset: { type: 'number' },
+		sort: { type: 'string' },
+	},
+	required: ['roomId'],
+	additionalProperties: false,
+};
+
+export const isAuditRoomMembersProps = ajv.compile<AuditRoomMembersParams>(auditRoomMembersSchema);
+
+declare module '@rocket.chat/rest-typings' {
```
sampaiodiego, line 34: should we start looking at moving these definitions to a "rest-typings-ee" package? I've noticed since we're adding more EE endpoints we're spreading these AJV typings around

sampaiodiego (approved):

apps/meteor/ee/server/api/audit.ts
KevLehman: Probably time to separate ajv validators from types 🤔 

I'd like to have a package only for ajv of all ee/ce endpoints and then 2 diff packages for rest typings...

PR #32923, opened 2024-07-29
ci: change release repository

debdutdeb:


sampaiodiego (approved):
do you think this is gonna fix it? or can it be something else?

we'll need to cherry-pick the fix to other release branches as well

debdutdeb:
@sampaiodiego yeah this should fix that issue 😬

PR #32930, opened 2024-07-29
ci: Bump `@actions/download-artifact` and `@actions/upload-artifact`

tassoevan:

It upgrades GitHub workflows to `@actions/downlad-artifact@v4` and `@actions/upload-artifact@v4`.

## Further comments

https://github.blog/changelog/2024-04-16-deprecation-notice-v3-of-the-artifact-actions/


.github/workflows/ci-test-e2e.yml
```diff
@@ -314,14 +314,46 @@ jobs:
...
         if: inputs.type == 'ui' && inputs.release == 'ee'
-        uses: actions/upload-artifact@v3
+        uses: actions/upload-artifact@v4
         with:
-          name: e2e-ee-coverage
+          name: e2e-ee-coverage-${{ matrix.mongodb-version }}-${{ matrix.shard }}
           path: ./apps/meteor/coverage*
+
+
+  merge:
```
sampaiodiego, line 330: any real benefit on "merging" the artifacts before upload? I thought it would be actually good to have them separate.

if we want to indeed merge them I'd ask to provide a name to this job since "merge" is too vague IMO

tassoevan: I agree, but I didn't know how arbitrary the artifact names were so I went with keeping the old behavior. But wonder if using the shard indices would be useful e.g. ` playwright-trace-ce-6.0-3`  is kinda random.

ggazzo (approved):

PR #32931, opened 2024-07-29
fix: validate `bhType` before registering it

KevLehman:

Check if `bhType` is valid before attempting to register it as the business hour type. This can happen (very rarely) if settings collection on UI is loaded after license (so the check for license returns true and it tries to register).
This will cause the tracker to run again 
## Issue(s)

https://rocketchat.atlassian.net/browse/CORE-535


MarcosSpessatto (dismissed):

apps/meteor/app/livechat-enterprise/client/startup.ts
```diff
@@ -15,7 +15,7 @@ const businessHours: Record<string, IBusinessHourBehavior> = {
 Meteor.startup(() => {
 	Tracker.autorun(async () => {
 		const bhType = settings.get<string>('Livechat_business_hour_type');
-		if (await hasLicense('livechat-enterprise')) {
+		if ((await hasLicense('livechat-enterprise')) && bhType) {
```
ggazzo, line 18:
```suggestion
		if (bhType && (await hasLicense('livechat-enterprise'))) {
```

MartinSchoeler (approved):

PR #32940, opened 2024-07-30
fix: File uploads should only be allowed for room members

MartinSchoeler:

Currently non channel members can drag and drop files into a channel they do not belong, this PR fixes that

## Issue(s)

[CORE-205](https://rocketchat.atlassian.net/browse/CORE-205)

[CORE-205]: https://rocketchat.atlassian.net/browse/CORE-205?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/meteor/client/views/room/body/hooks/useFileUploadDropTarget.ts
```diff
@@ -78,12 +79,20 @@ export const useFileUploadDropTarget = (): readonly [
 			} as const;
 		}
 
+		if (!subscription) {
+			return {
+				enabled: false,
+				reason: t('error-not-allowed'),
+				...overlayProps,
+			} as const;
+		}
```
dougfabris, line 88:
```suggestion
		if (!fileUploadAllowedForUser || !subscription) {
			return {
				enabled: false,
				reason: t('error-not-allowed'),
				...overlayProps,
			} as const;
		}
```

dougfabris (changes requested):

apps/meteor/tests/e2e/file-upload.spec.ts
```diff
@@ -76,3 +76,27 @@ test.describe.serial('file-upload', () => {
 		await expect(poHomeChannel.content.btnModalConfirm).not.toBeVisible();
 	});
 });
+test.describe('file-upload-not-member', () => {
```
KevLehman, line 79: Should we check how these behaves on Omni rooms?

MartinSchoeler: I don`t think so, they behave exactly the same as the other channels in this regard, the agent/manager cannot send messages in the room unless they are a part of it

dougfabris: @MartinSchoeler indeed, not necessary!

dougfabris (approved):
This is such a complete work, congrats on that!

apps/meteor/tests/e2e/file-upload.spec.ts
KevLehman: noice noice

KevLehman (approved):

PR #32943, opened 2024-07-30
chore: Add `departmentName` to the return values of `:agentId/departments` endpoint 

KevLehman:

- Adds a new `departmentName` key to the object returned from the `:agentId/departments` endpoint in Omnichannel.
- This will help to avoid some pagination problems on the UI (that will be fixed on another PR)
## Issue(s)

https://rocketchat.atlassian.net/browse/SUP-623
## Further comments

Apparently the tests for this endpoint were not running, so those were fixed too. (Tests were made EE because of the limitation of departments on CE)


.changeset/proud-bees-design.md
matheusbsilva137, line 1: If this has a changeset, isn't the PR a feat instead of chore? :thinking: 
If so, then we should use `minor` instead of `patch`

apps/meteor/server/models/raw/LivechatDepartmentAgents.ts
```diff
@@ -395,6 +396,39 @@ export class LivechatDepartmentAgentsRaw extends BaseRaw<ILivechatDepartmentAgen
 	): FindCursor<ILivechatDepartmentAgents> {
 		return this.find({ agentId: { $in: agentsIds }, departmentId }, options);
 	}
+
+	findDepartmentsOfAgent(agentId: string, enabled = false): AggregationCursor<ILivechatDepartmentAgents & { departmentName: string }> {
+		return this.col.aggregate<ILivechatDepartmentAgents & { departmentName: string }>([
+			{
+				$match: {
+					agentId,
+					...(enabled && { departmentEnabled: true }),
```
matheusbsilva137, line 405: Can we have a compound index to cover this new query? I believe it's worth it since the ones we currently have only cover each restriction individually but not them together

apps/meteor/app/livechat/server/api/lib/agents.ts
```diff
@@ -7,14 +7,8 @@ export async function findAgentDepartments({
 }: {
 	enabledDepartmentsOnly?: boolean;
 	agentId: string;
-}): Promise<{ departments: ILivechatDepartmentAgents[] }> {
-	if (enabledDepartmentsOnly) {
-		return {
-			departments: await LivechatDepartmentAgents.findActiveDepartmentsByAgentId(agentId).toArray(),
```
matheusbsilva137, line 13: We should remove this method since this was the only place where we used it

apps/meteor/server/models/raw/LivechatDepartmentAgents.ts
```diff
@@ -395,6 +396,39 @@ export class LivechatDepartmentAgentsRaw extends BaseRaw<ILivechatDepartmentAgen
...
+			{
+				$project: {
+					_id: '$_id',
+					agentId: '$agentId',
+					departmentId: '$departmentId',
+					departmentName: '$department.name',
+					username: '$username',
+					count: '$count',
+					order: '$order',
+					_updatedAt: '$_udpatedAt',
```
matheusbsilva137, line 426: Not a big deal, but this specific property is not in the `ILivechatDepartmentAgents` type. Can you please add it?

matheusbsilva137 (changes requested):

.changeset/proud-bees-design.md
KevLehman: Good question. Honestly, i feel this is too small to be a `feat`, it deserves a changeset cause you're adding to an endpoint, but it's not a fix because you're not fixing anything.

So ended up with chore + changeset since it's a smol change. Don't mind to change either.

apps/meteor/server/models/raw/LivechatDepartmentAgents.ts
KevLehman: Actually this is not being used and it's just there because of the model typing. Gonna remove this from the return.
```suggestion
```

.changeset/proud-bees-design.md
KevLehman: wdyt @MarcosSpessatto ? To have or not to have a changeset?

MarcosSpessatto: I don't this it's necessary since we will have the changeset on the PR that fixes the pagination, right? WDYT?

apps/meteor/server/models/raw/LivechatDepartmentAgents.ts
matheusbsilva137: Is it being returned, currently? If so, then removing it would be breaking :thinking:

KevLehman: not breaking because it wasn't on the types 🤔 so even if it was there, there was no guarantee it should be there lol

Guess you can see it both ways

KevLehman: But, just in case, you have it there now

matheusbsilva137 (approved):

KevLehman:
@MarcosSpessatto feel we can skip QA on this one as QA will be performed when the UI connects with it, wdyt?

MarcosSpessatto (approved):

MartinSchoeler (approved):

aleksandernsilva (approved):

sampaiodiego (approved):

PR #32945, opened 2024-07-30
feat: Option to disable 2FA for OAuth users

yash-rajpal:

Since SSO users are already verified and coming from a reliable source, we don't necessarily need to authenticate TOTP for these users.

Added a new setting which allows workspace admins to disable email two factor authentication for SSO (OAuth) users. If enabled, SSO users won't be asked for email two factor authentication.
## Further comments

[CORE-566](https://rocketchat.atlassian.net/browse/CORE-566)

[CORE-566]: https://rocketchat.atlassian.net/browse/CORE-566?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


KevLehman (changes requested):
- Add a PR description
- Is it possible to have some tests? I think 2FA is disabled on TEST_MODE so some units would suffice

apps/meteor/app/2fa/server/code/EmailCheck.ts
```diff
@@ -28,6 +28,10 @@ export class EmailCheck implements ICodeCheck {
 			return false;
 		}
 
+		if (settings.get('Accounts_TwoFactorAuthentication_Disable_Email_For_OAuth_Users')) {
```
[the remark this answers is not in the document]
yash-rajpal: That's the idea.

It will return false for OAuth users, disabling Email 2FA for OAuth users.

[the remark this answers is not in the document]
yash-rajpal: Ohh, I see what you are saying, I missed it, will correct this.

packages/core-typings/src/IUser.ts
```diff
@@ -79,27 +101,37 @@ export interface IUserServices {
...
+	'nextcloud',
+	'saml',
+	'twitter',
+] as IOAuthServices[];
+const userServiceKeys = ['emailCode', 'email2fa', 'totp', 'resume', 'password', 'passwordHistory', 'cloud', 'email'] as IUserService[];
+
+export const isUserServiceKey = (key: string): key is IUserService => key in userServiceKeys || key in defaultOAuthKeys;
+
+export const isDefaultOAuthUser = (user: IUser): boolean =>
+	!!(user.services && Object.keys(user.services).some((key) => key in defaultOAuthKeys));
```
KevLehman, line 127:
```suggestion
	!!user.services && Object.keys(user.services).some((key) => key in defaultOAuthKeys);
```

KevLehman: no need for the () as .some will return a boolean. Just need the !! for user.services

packages/core-typings/src/IUser.ts
```diff
@@ -79,27 +101,37 @@ export interface IUserServices {
...
+
+export const isCustomOAuthUser = (user: IUser): boolean =>
+	!!(user.services && Object.keys(user.services).some((key) => !isUserServiceKey(key)));
+
+export const isOAuthUser = (user: IUser): boolean =>
+	!!(user.services && Object.keys(user.services).some((key: any) => !userServiceKeys.includes(key)));
```
KevLehman, line 133: why this one requires `key: any` but the other 2 doesnt?

packages/core-typings/src/IUser.ts
```diff
@@ -45,7 +45,35 @@ export type ILoginUsername =
 	  };
 export type LoginUsername = string | ILoginUsername;
 
-export interface IUserServices {
+export interface IOAuthUserServices {
+	google?: any;
```
KevLehman, line 49:
```suggestion
	google?: Record<string, any>;
```

:its_something:

apps/meteor/app/2fa/server/code/EmailCheck.ts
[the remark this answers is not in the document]
yash-rajpal: Done!

packages/core-typings/src/IUser.ts
```diff
@@ -79,27 +101,37 @@ export interface IUserServices {
...
+	'google',
+	'ldap',
+	'linkedin',
+	'nextcloud',
+	'saml',
+	'twitter',
+] as IOAuthServices[];
+const userServiceKeys = ['emailCode', 'email2fa', 'totp', 'resume', 'password', 'passwordHistory', 'cloud', 'email'] as IUserService[];
+
+export const isUserServiceKey = (key: string): key is IUserService => key in userServiceKeys || key in defaultOAuthKeys;
```
matheusbsilva137, line 124:
```suggestion
export const isUserServiceKey = (key: string): key is IUserService => userServiceKeys.includes(key) || defaultOAuthKeys.includes(key);
```
I believe `in` doesn't work to check if an array contains an element, does it? We should use `includes` instead

packages/core-typings/src/IUser.ts
matheusbsilva137, line 127:
```suggestion
	!!(user.services && Object.keys(user.services).some((key) => defaultOAuthKeys.includes(key)));
```
I believe `in` doesn't work to check if an array contains an element, does it? We should use `includes` instead

packages/core-typings/src/IUser.ts
matheusbsilva137, line 133: We're missing something here. We should check if `isCustomOAuthUser || isDefaultOAuthUser`

packages/core-typings/src/IUser.ts
```diff
@@ -185,6 +217,7 @@ export interface IUser extends IRocketChatRecord {
 	_pendingAvatarUrl?: string;
 	requirePasswordChange?: boolean;
 	requirePasswordChangeReason?: string;
+	isOAuthUser?: boolean;
```
matheusbsilva137, line 220: Aren't we using this only in the ednpoint return? I believe this doesn't need to be in core-typings

matheusbsilva137 (changes requested):

packages/core-typings/src/IUser.ts
yash-rajpal: You are right, we can do that too.

Above code checks if the service key should be something else than the internal service keys.

packages/core-typings/src/IUser.ts
yash-rajpal: This is a client only field, on client side we will be using `user?.isOAuthUser` to check this, so it is required here.

I'll put a comment here mentioning this is a client only field.

We don't need this on rest typings since the API returns IUser.

packages/core-typings/src/IUser.ts
```diff
@@ -79,27 +101,37 @@ export interface IUserServices {
...
-		refreshToken: string;
-		serverURL: string;
-	};
-	dolphin?: {
-		NickName?: string;
-	};
 }
 
+type IUserService = keyof IUserServices;
+type IOAuthServices = keyof IOAuthUserServices;
```
matheusbsilva137, line 107:
```suggestion
type IOAuthService = keyof IOAuthUserServices;
```
I believe singular makes more sense here

packages/core-typings/src/IUser.ts
```diff
@@ -79,27 +101,37 @@ export interface IUserServices {
...
+	'facebook',
+	'github',
+	'gitlab',
+	'google',
+	'ldap',
+	'linkedin',
+	'nextcloud',
+	'saml',
+	'twitter',
+] as IOAuthServices[];
```
matheusbsilva137, line 121:
```suggestion
] as IOAuthService[];
```

packages/core-typings/src/IUser.ts
```diff
@@ -79,27 +101,37 @@ export interface IUserServices {
...
+const userServiceKeys = ['emailCode', 'email2fa', 'totp', 'resume', 'password', 'passwordHistory', 'cloud', 'email'] as IUserService[];
+
+export const isUserServiceKey = (key: string): key is IUserService =>
+	userServiceKeys.includes(key as IUserService) || defaultOAuthKeys.includes(key as IOAuthServices);
+
+export const isDefaultOAuthUser = (user: IUser): boolean =>
+	!!user.services && Object.keys(user.services).some((key) => defaultOAuthKeys.includes(key as IOAuthServices));
```
matheusbsilva137, line 128:
```suggestion
	!!user.services && Object.keys(user.services).some((key) => defaultOAuthKeys.includes(key as IOAuthService));
```

matheusbsilva137 (changes requested):

packages/core-typings/src/IUser.ts
yash-rajpal: Reverted back to `any` as it was causing some TS issues.

apps/meteor/app/api/server/v1/misc.ts
```diff
@@ -176,15 +176,19 @@ API.v1.addRoute(
 	{ authRequired: true },
 	{
 		async get() {
-			const fields = getDefaultUserFields();
-			const { services, ...user } = (await Users.findOneById(this.userId, { projection: fields })) as IUser;
+			const userFields = { ...getBaseUserFields, services: 1 };
```
KevLehman, line 179: Isn't `getBaseUserFields` a function?

KevLehman (changes requested):

apps/meteor/app/api/server/v1/misc.ts
yash-rajpal: Yep, fixed on local, have to push. 

Working on some other things.

Bad mistake 🙈

apps/meteor/app/api/server/v1/misc.ts
```diff
@@ -176,15 +207,18 @@ API.v1.addRoute(
...
 	{
 		async get() {
-			const fields = getDefaultUserFields();
-			const { services, ...user } = (await Users.findOneById(this.userId, { projection: fields })) as IUser;
+			const { services, ...user } = (await Users.findOneById(this.userId, { projection: userFields })) as IUser;
 
 			return API.v1.success(
 				await getUserInfo({
 					...user,
+					isOAuthUser: isOAuthUser({ ...user, services }),
```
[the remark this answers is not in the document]
yash-rajpal: I was thinking that, but decided not to clutter the services object, as services object contains useful info related to sessions and tokens and stuff.

Also, this is more apt in user object as this is a user property, just a personal preference.

apps/meteor/client/views/account/security/TwoFactorEmail.tsx
```diff
@@ -1,15 +1,18 @@
...
 import React, { useCallback } from 'react';
 
 import { useEndpointAction } from '../../../hooks/useEndpointAction';
 
-const TwoFactorEmail = (props: ComponentProps<typeof Box>): ReactElement => {
+const TwoFactorEmail = (props: ComponentProps<typeof Box>): JSX.Element | null => {
 	const t = useTranslation();
 	const user = useUser();
 
+	const disableEmail2FAForOAuth = useSetting('Accounts_TwoFactorAuthentication_email_available_for_oAuth_users');
```
MarcosSpessatto, line 12: This being true, is it not enabled instead of disabled?

apps/meteor/client/views/account/security/TwoFactorEmail.spec.tsx
```diff
@@ -0,0 +1,45 @@
...
+
+	it('should render the Email two factor button', () => {
+		(useSetting as jest.Mock).mockReturnValue(true);
+		(useUser as jest.Mock).mockReturnValue({ isOAuthUser: true });
+
+		render(<TwoFactorEmail />);
+		expect(screen.getByText('Enable_two-factor_authentication_email')).toBeInTheDocument();
+	});
+
+	it('should render nothing, no button for OAuth user if setting is false', () => {
```
MarcosSpessatto, line 31: What if it is true?

apps/meteor/app/2fa/server/code/EmailCheck.spec.ts
```diff
@@ -0,0 +1,79 @@
...
+	it('should return EmailCheck is not enabled for a normal user with unverified email', () => {
+		isOAuthUserMock.returns(false);
+		settingsMock.returns(true);
+
+		const isEmail2FAEnabled = emailCheck.isEnabled(normalUserWithUnverifiedEmailMock);
+
+		expect(isEmail2FAEnabled).to.be.equal(false);
+	});
+
+	it('should return EmailCheck is not enabled for a OAuth user with setting being false', () => {
```
MarcosSpessatto, line 70: Can you cover the subsequent cases as well? Not only the ones you added (for this method only).

MarcosSpessatto (changes requested):
Also, please add a description for the PR

apps/meteor/client/views/account/security/TwoFactorEmail.tsx
```diff
@@ -1,15 +1,18 @@
...
-import { useUser, useTranslation } from '@rocket.chat/ui-contexts';
-import type { ComponentProps, ReactElement } from 'react';
+import { useUser, useTranslation, useSetting } from '@rocket.chat/ui-contexts';
+import type { ComponentProps } from 'react';
 import React, { useCallback } from 'react';
 
 import { useEndpointAction } from '../../../hooks/useEndpointAction';
 
-const TwoFactorEmail = (props: ComponentProps<typeof Box>): ReactElement => {
+const TwoFactorEmail = (props: ComponentProps<typeof Box>): JSX.Element | null => {
```
tassoevan, line 8:
```suggestion
const TwoFactorEmail = (props: ComponentProps<typeof Box>) => {
```

tassoevan (changes requested):

apps/meteor/client/views/account/security/TwoFactorEmail.tsx
yash-rajpal: I had coded this earlier, later we decided to change the setting name and default value, forgot to update the variable name accordingly.

apps/meteor/client/views/account/security/TwoFactorEmail.spec.tsx
yash-rajpal: Already covered that case, in just above testcase.

apps/meteor/app/2fa/server/code/EmailCheck.spec.ts
yash-rajpal: After talking to Marcos, decided to do this in a separate task of it's own and implement tests for all these TOTP classes.

apps/meteor/app/2fa/server/code/EmailCheck.spec.ts
```diff
@@ -0,0 +1,79 @@
...
+import { describe, it } from 'mocha';
+import proxyquire from 'proxyquire';
+import sinon from 'sinon';
+
+const isOAuthUserMock = sinon.stub();
+const settingsMock = sinon.stub();
+
+const { EmailCheck } = proxyquire.noCallThru().load('./EmailCheck', {
+	'@rocket.chat/core-typings': {
+		isOAuthUser: isOAuthUserMock,
```
KevLehman, line 11: You shouldn't need to mock this func, apart from not affecting, the checks it performs are valuable for what you're testing here.

You should pass data that will make this func fail/pass instead

KevLehman (changes requested):

apps/meteor/app/2fa/server/code/EmailCheck.spec.ts
yash-rajpal: You are right, fixed and pushed.

apps/meteor/app/2fa/server/code/EmailCheck.spec.ts
```diff
@@ -0,0 +1,71 @@
...
+		settingsMock.returns(true);
+
+		const isEmail2FAEnabled = emailCheck.isEnabled(normalUserWithUnverifiedEmailMock);
+
+		expect(isEmail2FAEnabled).to.be.equal(false);
+	});
+
+	it('should return EmailCheck is not enabled for a OAuth user with setting being false', () => {
+		settingsMock.onFirstCall().returns(true);
+		settingsMock.onSecondCall().returns(true);
```
KevLehman, line 65: If both calls are returning true...
```suggestion
		settingsMock.returns(true);
```

KevLehman (dismissed):

KevLehman (dismissed):

packages/i18n/src/locales/en.i18n.json
```diff
@@ -282,6 +282,8 @@
   "Accounts_TwoFactorAuthentication_By_Email_Code_Expiration": "Time to expire the code sent via email in seconds",
   "Accounts_TwoFactorAuthentication_By_Email_Enabled": "Enable Two Factor Authentication via Email",
   "Accounts_TwoFactorAuthentication_By_Email_Enabled_Description": "Users with email verified and the option enabled in their profile page will receive an email with a temporary code to authorize certain actions like login, save the profile, etc.",
+  "Accounts_twoFactorAuthentication_email_available_for_oAuth_users": "Make two factor via email available for oAuth users",
+  "Accounts_twoFactorAuthentication_email_available_for_oAuth_users_Description": "People that use oAuth will receive an email with a temporary code to authorize actions like login, save profile, etc.",
```
matheusbsilva137, line 286:
```suggestion
  "Accounts_twoFactorAuthentication_email_available_for_OAuth_users": "Make two factor via email available for OAuth users",
  "Accounts_twoFactorAuthentication_email_available_for_OAuth_users_Description": "People that use OAuth will receive an email with a temporary code to authorize actions like login, save profile, etc.",
```
Kinda picky, but can we please update this? :grimacing: That's how we "spell" it everywhere else

apps/meteor/app/2fa/server/code/EmailCheck.ts
```diff
@@ -24,6 +24,10 @@ export class EmailCheck implements ICodeCheck {
 			return false;
 		}
 
+		if (!settings.get('Accounts_twoFactorAuthentication_email_available_for_oAuth_users') && isOAuthUser(user)) {
```
matheusbsilva137, line 27:
```suggestion
		if (!settings.get('Accounts_twoFactorAuthentication_email_available_for_OAuth_users') && isOAuthUser(user)) {
```

apps/meteor/client/views/account/security/AccountSecurityPage.tsx
```diff
@@ -15,6 +15,11 @@ const passwordDefaultValues = { password: '', confirmationPassword: '' };
 
 const AccountSecurityPage = (): ReactElement => {
 	const t = useTranslation();
+	const user = useUser();
+
+	const isEmail2FAAvailableForOAuth = useSetting('Accounts_twoFactorAuthentication_email_available_for_oAuth_users');
```
matheusbsilva137, line 20:
```suggestion
	const isEmail2FAAvailableForOAuth = useSetting('Accounts_twoFactorAuthentication_email_available_for_OAuth_users');
```

apps/meteor/server/settings/accounts.ts
```diff
@@ -31,6 +31,18 @@ export const createAccountSettings = () =>
 				public: true,
 			});
 
+			await this.add('Accounts_twoFactorAuthentication_email_available_for_oAuth_users', true, {
```
matheusbsilva137, line 34:
```suggestion
			await this.add('Accounts_twoFactorAuthentication_email_available_for_OAuth_users', true, {
```

matheusbsilva137 (changes requested):

packages/i18n/src/locales/en.i18n.json
yash-rajpal: Done! 🙂

matheusbsilva137 (approved):

KevLehman (approved):

MarcosSpessatto (approved):

tassoevan (approved):

PR #32948, opened 2024-07-31
chore: create `orm` for specific cases

ggazzo:
[OPI-20]
https://adr.rocket.chat/0099

## Further comments

[OPI-20]: https://rocketchat.atlassian.net/browse/OPI-20?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/models/src/updater.ts
```diff
@@ -0,0 +1,103 @@
...
+		}
+
+		this.dirty = true;
+
+		try {
+			await this.model.updateOne(query, {
+				...(this._set && { $set: Object.fromEntries(this._set) }),
+				...(this._unset && { $unset: Object.fromEntries([...this._unset.values()].map((k) => [k, 1])) }),
+				...(this._inc && { $inc: Object.fromEntries(this._inc) }),
+				...(this._addToSet && { $addToSet: { $each: Object.fromEntries(this._addToSet) } }),
```
sampaiodiego, line 70: this code is used multiple times, what if you move it to a private method and use it instead?

packages/models/src/updater.ts
```diff
@@ -0,0 +1,103 @@
...
+	unset<K extends keyof UnsetProps<T>>(key: K): Updater<T> {
+		this._unset = this._unset ?? new Set<keyof UnsetProps<T>>();
+		this._unset.add(key);
+		return this;
+	}
+
+	test1<K extends UpdateFilter<T>>(key: K) {
+		console.log(key);
+	}
+
```
sampaiodiego, line 37:
```suggestion
```

sampaiodiego (changes requested):

packages/models/src/updater.ts
ggazzo: image

sampaiodiego (approved):
nojo

PR #32951, opened 2024-07-31
chore: Move Apps-Engine to monorepo

d-gubert:

Move the Apps-Engine repo (https://github.com/RocketChat/Rocket.Chat.Apps-Engine) to the core monorepo (https://github.com/RocketChat/Rocket.Chat)

This will allow for a much smoother developer experience when adding features to the engine, as it won't be necessary anymore to wait for merges on one repo to then get other PR merged, and also we won't have conflicts anymore if something is merged on the engine side but not on the core side.

### Benefits

- Smoother DX - there won't be a need to keep and conciliate PRs in two repos when making changes
- Slimmer package - there won't be a need for the published package to contain internal logic, we'll be able to simply publish types. The same goes for its dependencies, as Deno, for instance, won't be required by the package when it is installed for app development.
- Better control - we'll be able to close internal pieces of the engine as "Enterprise Edition" and keep public parts open much more easily
- Core types interop - Rocket.Chat types will be more accessible to the package.

### Drawbacks

- Increased monorepo operations - CI/CD cycles will potentially get slower given that more things will need to be installed and built
- Apps-Engine release process tightly closed to the Rocket.Chat one - this will prevent the Apps-Engine package from having a release (e.g. a patch) outside of Rocket.Chat's release cycle automatically. However, this situation has been rare in the past and it can be worked around manually if needed.
- Github workflows take a long time to run on the monorepo - well... yeah.

### Considerations

- What to do with issues and PRs on the Apps-Engine repo?

- What do we do with the API reference? Currently it is hosted in github.io. Can we just port it?

- What do we need to adapt here for Codecov?

## Further comments

https://rocketchat.atlassian.net/browse/CONN-308

~As of now, the apps-engine cannot execute `deno` without a global installation in the system (e.g. the binary being available `/bin/deno`). This seemed to start after Meteor's update to their own Node patch 14.21.4~

This was because we were overriding the ENV for the process and weren't passing a `PATH`, which causes the system to be unable to find the `deno` executable. It never happened before because previously we were using the `deno-bin` package to install deno and so we knew the direct path to the executable and didn't need to rely on the `PATH` environment variable.

## Note for reviewers

Most of the files introduced by this PR already exist and are part of the `@rocket.chat/apps-engine` package - meaning they don't need further review, as they're just being moved. The important files to keep an eye out for are the ones _outside_ of the `packages/apps-engine` folder, the `packages/apps-engine/package.json` file itself and whatever scripts it references, as those integrate the new package into the rest of the monorepo ecosystem.

In order words, don't be frightened by the amount of files :see_no_evil:


packages/apps-engine/package.json
```diff
@@ -0,0 +1,132 @@
...
+        "fix-lint": "eslint . --fix",
+        "build": "run-s .:build:clean .:build:default",
+        ".:build:clean": "rimraf client definition server",
+        ".:build:default": "tsc -p tsconfig.json",
+        ".:build:watch": "yarn .:build:default --watch",
+        "typecheck": "tsc -p tsconfig.json --noEmit",
+        "bundle": "node scripts/bundle.js",
+        "gen-doc": "typedoc",
+        "prepack": "yarn bundle",
+        "postinstall": "node scripts/postinstall.js"
```
sampaiodiego, line 24: what if you move this to the "build" script? this way we can avoid having to install `deno` in ALL Dockerfiles since the majority of them just uses apps-engine for its types..

ee/apps/authorization-service/Dockerfile
```diff
@@ -1,12 +1,74 @@
-FROM node:14.21.3-alpine
+ARG DENO_VERSION="1.37.1"
+
+FROM denoland/deno:alpine-${DENO_VERSION} as deno
+
+FROM node:14.21.3-alpine3.16
 
 ARG SERVICE
 
+## Alpine 3.16 does not have a deno package, but newer versions have it
```
sampaiodiego, line 9: if moving the `postintall` script to `build` script as suggested below we probably won't need to do this, since `authorization-service` doesn't uses apps-engine and thus doesn't need `deno`.

sampaiodiego (changes requested):
we shall review the installation of `deno` runtime in all Dockerfiles since in theory only "meteor" needs `deno` to run apps, other services only uses apps-engine's types.

packages/apps-engine/package.json
d-gubert: That's a good point, I'll take a look

packages/apps-engine/LICENSE
sampaiodiego, line 1: pls remove.. let's use same repo's license

apps/meteor/.docker/Dockerfile
```diff
@@ -1,3 +1,7 @@
+ARG DENO_VERSION="1.37.1"
+
+FROM denoland/deno:${DENO_VERSION} as deno
```
sampaiodiego, line 3: I figured they have a bin only variant that suits us better
```suggestion
FROM denoland/deno:bin-${DENO_VERSION} as deno
```

apps/meteor/.docker/Dockerfile.alpine
```diff
@@ -1,7 +1,17 @@
+ARG DENO_VERSION="1.37.1"
+
+FROM denoland/deno:alpine-${DENO_VERSION} as deno
```
sampaiodiego, line 3:
```suggestion
FROM denoland/deno:bin-${DENO_VERSION} as deno
```

.tool-versions
sampaiodiego, line 1: are you using `asdf`? 🤔

sampaiodiego (changes requested):

.tool-versions
d-gubert: Oh, this was a test for how to set the version of deno, the action reads these files. But yeah, I ended up going another route. I'll delete this

apps/meteor/.docker/Dockerfile
```diff
@@ -10,6 +14,8 @@ RUN groupadd -g 65533 -r rocketchat \
     && apt-get update \
     && apt-get install -y --no-install-recommends fontconfig
 
+COPY --from=deno /bin/deno /bin/deno
```
sampaiodiego, line 17:
```suggestion
COPY --from=deno /deno /bin/deno
```

apps/meteor/.docker/Dockerfile.alpine
```diff
@@ -44,12 +54,11 @@ RUN ALPINE_GLIBC_BASE_URL="https://github.com/sgerrand/alpine-pkg-glibc/releases
     rm \
         "$ALPINE_GLIBC_BASE_PACKAGE_FILENAME" \
         "$ALPINE_GLIBC_BIN_PACKAGE_FILENAME" \
-        "$ALPINE_GLIBC_I18N_PACKAGE_FILENAME" && \
-    apk add --no-cache ttf-dejavu
+        "$ALPINE_GLIBC_I18N_PACKAGE_FILENAME"
 
-ADD . /app
+COPY --from=deno /bin/deno /bin/deno
```
sampaiodiego, line 59:
```suggestion
COPY --from=deno /deno /bin/deno
```

sampaiodiego (changes requested):

sampaiodiego (approved):

PR #32968, opened 2024-08-02
fix: `Menu` onPointerUp event behavior

juliajforesti:

When there is not enough space for the menu popover to be placed above/below its trigger, it will overlap the trigger. This causes the popover to be immediately closed and whatever item is underneath the pointer will be selected.

The issue is caused by conflicting event handlers. This PR bumps the package `@rocket.chat/fuselage` with the necessary fix.


dougfabris (approved):
LGTM!

PR #32981, opened 2024-08-05
fix: Following/unfollowing a message when the thread is closed does not update the UI

gabriellsh:

The follow/unfollow methods were not notifying a collection change.
## Issue(s)

[CORE-542](https://rocketchat.atlassian.net/browse/CORE-542)

## Further comments

[CORE-542]: https://rocketchat.atlassian.net/browse/CORE-542?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/meteor/tests/e2e/message-actions.spec.ts
```diff
@@ -38,6 +38,34 @@ test.describe.serial('message-actions', () => {
...
+		await test.step('start thread', async () => {
+			await poHomeChannel.content.sendMessage('this is a message for reply');
+			await page.locator('[data-qa-type="message"]').last().hover();
+			await page.locator('role=button[name="Reply in thread"]').click();
+			await page.locator('.rcx-vertical-bar').locator(`role=textbox[name="Message #${targetChannel}"]`).type('this is a reply message');
+			await page.keyboard.press('Enter');
+			await expect(poHomeChannel.tabs.flexTabViewThreadMessage).toHaveText('this is a reply message');
+		});
+
+		// close thread before testing because behaviour changes
```
pierre-lehnen-rc, line 52: maybe add a test with the thread open too, since the behavior can be different?

pierre-lehnen-rc (dismissed):

apps/meteor/tests/e2e/message-actions.spec.ts
gabriellsh: The difference was that with the thread open it would always work since we listen to the subscription stream in that case. In a more recent PR I have unit tests for the Thread metrics components aswell. Do you think another test is still needed for this particular PR?

gabriellsh: I'll do it just for the sake of it. I also forgot to remove the .only from the test so all other message-actions tests are being ignored

apps/meteor/app/threads/server/methods/followMessage.ts
```diff
@@ -43,6 +44,10 @@ Meteor.methods<ServerMethods>({
 
 		const followResult = await follow({ tmid: message.tmid || message._id, uid });
 
+		void notifyOnMessageChange({
+			id: mid,
```
pierre-lehnen-rc, line 48: Just realized that the method doesn't necessarily modify the same message that is passed as argument, but your code was sending the event with the id from the params directly.

```suggestion
			id: message.tmid || message._id,
```

gabriellsh: Nice Catch! Fixed.

MarcosSpessatto (approved):

pierre-lehnen-rc (approved):

dougfabris (approved):
Approving on behalf of frontend team!

PR #32987, opened 2024-08-05
chore: Renders back button through prop in admin settings

dougfabris:

This PR proposes to render the back button in `GroupPage` through `onClickBack` prop. It will allow to use this page outside the route without displaying the button that move back to the settings page


gabriellsh (approved):
Yes, I painstakingly opened each group and tested the back button. Manually.

dougfabris:
@gabriellsh me too

PR #32993, opened 2024-08-06
chore: Use updater in `saveUser` file

matheusbsilva137:

 - Use our updater in the `saveUser` file to avoid multiple DB trips;
 - Fix `IUserSettings` type to have an optional `profile`, as set through the API;
 - Support `settings` and `language` params in the `users.update` endpoint (those were already being handled by the `saveUser` function, but were not accepted by the endpoint -- so they could only be updated using the Meteor method);
 -  Support `settings` param in the `users.create` endpoint (which was already being handled by the `saveUser` function, but was not accepted by the endpoint -- so it could only be created using the Meteor method).

## Further comments

[CORE-734](https://rocketchat.atlassian.net/browse/CORE-734)

[CORE-734]: https://rocketchat.atlassian.net/browse/CORE-734?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


MarcosSpessatto (changes requested):

apps/meteor/app/lib/server/functions/saveUser.ts
```diff
@@ -158,22 +163,20 @@ async function validateUserData(userId, userData) {
...
-export async function validateUserEditing(userId, userData) {
+export async function validateUserEditing(userId: IUser['_id'], userData: IUpdateUserParams) {
 	const editingMyself = userData._id && userId === userData._id;
 
 	const canEditOtherUserInfo = await hasPermissionAsync(userId, 'edit-other-user-info');
 	const canEditOtherUserPassword = await hasPermissionAsync(userId, 'edit-other-user-password');
 	const user = await Users.findOneById(userData._id);
 
-	const isEditingUserRoles = (previousRoles, newRoles) =>
+	if (!user) {
```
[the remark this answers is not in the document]
matheusbsilva137: Hmmm I believe this is more of a missing validation than a breaking change. For example, check what we're doing [here](https://github.com/RocketChat/Rocket.Chat/pull/32993/files#diff-b8177d686748dae835de33baf6184906f9a5db359a50b41ab4c672a085ce4c2aR391) **after** `validateUserEditing` ran. We're relying on the `userData._id` field after this function is executed, so we must  assure that the `_id` points to a valid user here

So I think we're just throwing the error earlier here, it shouldn't be breaking. wdyt?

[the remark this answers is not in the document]
MarcosSpessatto: I understand your point, but my concern is more about the error message being thrown here, since before nothing was being thrown and now we are throwing... WDYT?

MarcosSpessatto (changes requested):

apps/meteor/app/lib/server/functions/saveUser.ts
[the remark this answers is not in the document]
matheusbsilva137: > since before nothing was being thrown

hmmm I think that's not exactly true (for all cases). Errors were being thrown before this change, but not **consistently** (I mean, we'd throw an error depending on which fields were being updated)

If `userData._id` is defined, but points to an invalid user, the code will skip [this condition](https://github.com/RocketChat/Rocket.Chat/blob/81629860557f051f7ed1d6d560f8885e587d9c9e/apps/meteor/app/lib/server/functions/saveUser.js#L370). So we'll try to **update** a existing invalid user, and these are the paths we may follow:
1. If `username` or `name` properties are being updated, `saveUserIdentity` [will return false](https://github.com/RocketChat/Rocket.Chat/blob/81629860557f051f7ed1d6d560f8885e587d9c9e/apps/meteor/app/lib/server/functions/saveUserIdentity.ts#L36). Thus, an [error](https://github.com/RocketChat/Rocket.Chat/blob/81629860557f051f7ed1d6d560f8885e587d9c9e/apps/meteor/app/lib/server/functions/saveUser.js#L386) will be thrown
2. If the user tries to set the `email` property, an [error](https://github.com/RocketChat/Rocket.Chat/blob/a40541b61676d23338d72d420995bbc5a211d152/apps/meteor/app/lib/server/functions/setEmail.ts#L54) will also be thrown
3. For bio, nickname, verified, language, settings, roles and status text no error is thrown, indeed. But no (valid) update is performed, also. Still, we trigger all callbacks and events for user update on this case. I believe you agree this is not the intended behavior, but I also believe that's not "worth" of being considered breaking since this is more of an improvement to a case that was being incorrectly handled

wdyt? should we consider it breaking?

apps/meteor/app/lib/server/functions/saveUser.ts
```diff
@@ -40,50 +47,48 @@ Meteor.startup(() => {
...
 			function: 'RocketChat.saveUser',
 			message: error.message,
 		});
 	}
 }
 
-async function validateUserData(userId, userData) {
+async function validateUserData(userId: IUser['_id'], userData: ISaveUserDataParams) {
 	const existingRoles = _.pluck(await getRoles(), '_id');
+	const isUpdateUserParams = '_id' in userData && userData._id;
```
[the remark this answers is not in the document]
matheusbsilva137: hmmm this does make sense, but we can't do that because it'd cause a typecheck error
`ISaveUserDataParams` may be `UserCreateParamsPOST`, which doesn't have the `_id` prop. So if we try to "access" it directly  (as you suggested) without checking if `'_id' in userData`, then we'd have an error (`the _id property does not exist in the UserCreateParamsPOST type`, even if we try `!!userData?._id`)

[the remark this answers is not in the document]
KevLehman: If `userData` can be of 2 types (union) I'd recommend doing this check into a type check func
```
isUserCreateParams = (d): d is UserCreateParams => ...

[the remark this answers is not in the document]
Comment (matheusbsilva137) [reply apps/meteor/app/lib/server/functions/saveUser.ts]: good idea! done, looks much better now (I had to keep checking for `'_id' in userData` due to the restriction I mentioned, but this still looks better)

File: apps/meteor/app/lib/server/functions/saveUser.ts [hunk 3]
```diff
@@ -342,8 +341,8 @@ const saveNewUser = async function (userData, sendPassword) {
 	return _id;
 };
 
-export const saveUser = async function (userId, userData) {
-	const oldUserData = userData._id && (await Users.findOneById(userData._id));
+export const saveUser = async function (userId: IUser['_id'], userData: ISaveUserDataParams) {
+	const oldUserData = '_id' in userData && userData._id && (await Users.findOneById(userData._id));
```
KevLehman, line 345: Should u use the same typecheck you created before for this case?

apps/meteor/app/lib/server/functions/saveUser.ts
```diff
@@ -367,8 +366,8 @@ export const saveUser = async function (userId, userData) {
 		delete userData.setRandomPassword;
 	}
 
-	if (!userData._id) {
-		return saveNewUser(userData, sendPassword);
+	if (!('_id' in userData) || !userData._id) {
```
KevLehman, line 369: or for this one

apps/meteor/app/lib/server/functions/saveUser.ts
```diff
@@ -409,10 +407,9 @@ export const saveUser = async function (userId, userData) {
 		sendPassword = false;
 	}
 
-	const updateUser = {
-		$set: {},
-		$unset: {},
-	};
+	const updateUser: DeepWritable<UpdateFilter<IUser>> = {};
```
KevLehman, line 410: If you're doing the type DeepWritable, can't you just define $set/$unset inside here? What error is ts giving? 👀

apps/meteor/app/lib/server/functions/saveUser.ts
```diff
@@ -455,8 +452,13 @@ export const saveUser = async function (userId, userData) {
 		performedBy: await safeGetMeteorUser(),
 	});
 
-	if (sendPassword) {
-		await _sendUserEmail(settings.get('Password_Changed_Email_Subject'), passwordChangedHtml, userData);
+	if (sendPassword && userUpdated?.emails && userUpdated.emails[0].address) {
+		userData.email = userUpdated.emails[0].address;
```
KevLehman, line 456: If this assignment wasn't done before, how this func worked? If we're fixing an issue, we should fix it outside of the ts conversion.

(additionally, i think this must check the first _verified_ email the user has)

apps/meteor/app/lib/server/methods/insertOrUpdateUser.ts
```diff
@@ -18,13 +19,14 @@ Meteor.methods<ServerMethods>({
...
 
-		if (!Meteor.userId()) {
+		if (!userId) {
 			throw new Meteor.Error('error-invalid-user', 'Invalid user', {
 				method: 'insertOrUpdateUser',
 			});
 		}
 
-		return saveUser(Meteor.userId(), userData);
+		return saveUser(userId, userData as ISaveUserDataParams);
```
KevLehman, line 30: I think we should add a validation this is what we're receiving, because the method seems to check only for object.

wdyt? cc @MarcosSpessatto for the wdyt too

KevLehman (changes requested):

KevLehman:
@matheusbsilva137 the task linked doesn't seem to relate that much with what you're doing here. Maybe this needs its own task?

apps/meteor/app/lib/server/functions/saveUser.ts
matheusbsilva137: It says that $set and $unset are possibly undefined. If we'd like to assign these properties here, we'd have to change the typing so as to make TS sure that they've been defined

apps/meteor/app/lib/server/methods/insertOrUpdateUser.ts
MarcosSpessatto: Agreed

matheusbsilva137: The issue is that this would be breaking, wouldn't it? If we're doing that, I believe that should be implemented in another PR
Should we have a task for that, wdyt @MarcosSpessatto ? This method is deprecated, though

apps/meteor/app/lib/server/functions/saveUser.ts
sampaiodiego, line 410: this seems a good opportunity for using the new `Updater` https://adr.rocket.chat/0099

matheusbsilva137: Done!

matheusbsilva137:
We converted `saveUser` file to TS in another PR (#33774). However, I decided to keep my PR so that we don't lose any of the typing refinements suggested here that weren't applied to the conversion PR.
Instead of converting the file to TS, this PR now aims to improve type-checking in this file, while using our new updater :)

Feel free to review it again, everyone!

MarcosSpessatto (approved):

KevLehman (approved):

PR #32994, opened 2024-08-06
refactor: Reactions set/unset

KevLehman:

- Refactors `chat.react` endpoint to remove check for message (executeSetReaction already does it)
- Refactors `executeSetReaction` to perform emoji & reactions check before all DB calls
- Refactors `executeSetReaction` to use new `countByNameOrAlias` query instead of a `find().count()`
- Refactors `executeSetReaction` to only request the required data when requesting room from DB
- Refactor `setReaction` to not await for callbacks (we don't need the return value anyways)
- Refactor `setReaction` to notify of room changed at the end of the function (instead of next to every update). This doesn't gain any perf, but looks better (and also fixes a missing notify).
- Added unit tests for all functions (except the meteor.methods call...)

## Issue(s)

https://rocketchat.atlassian.net/browse/CORE-597


apps/meteor/app/lib/server/functions/isTheLastMessage.ts
```diff
@@ -1,7 +1,7 @@
-import type { IMessage, IRoom } from '@rocket.chat/core-typings';
+import type { AtLeast, IMessage, IRoom } from '@rocket.chat/core-typings';
 
 import { settings } from '../../../settings/server';
 
 // eslint-disable-next-line @typescript-eslint/explicit-function-return-type
-export const isTheLastMessage = (room: IRoom, message: Pick<IMessage, '_id'>) =>
+export const isTheLastMessage = (room: AtLeast<IRoom, 'lastMessage'>, message: Pick<IMessage, '_id'>) =>
 	settings.get('Store_Last_Message') && (!room.lastMessage || room.lastMessage._id === message._id);
```
MarcosSpessatto, line 7: I know we didn't change this, but this looks wrong, the settings.get condition should not be here, the function is asking if this is the last message and mentions nothing about it being enabled...

apps/meteor/app/reactions/server/setReaction.ts
```diff
@@ -100,24 +78,32 @@ async function setReaction(room: IRoom, user: IUser, message: IMessage, reaction
 		await Messages.setReactions(message._id, message.reactions);
 		if (isTheLastMessage(room, message)) {
 			await Rooms.setReactionsInLastMessage(room._id, message.reactions);
-			void notifyOnRoomChangedById(room._id);
 		}
-		await callbacks.run('setReaction', message._id, reaction);
```
MarcosSpessatto, line 105: If we are getting rid of this callback, can we remove it from the callback's definition as well?

apps/meteor/app/reactions/server/setReaction.ts
```diff
@@ -100,24 +78,32 @@ async function setReaction(room: IRoom, user: IUser, message: IMessage, reaction
...
-		await callbacks.run('afterSetReaction', message, { user, reaction, shouldReact });
+		void callbacks.run('afterSetReaction', message, { user, reaction, shouldReact: true });
 
 		isReacted = true;
 	}
 
-	await Apps.self?.triggerEvent(AppEvents.IPostMessageReacted, message, user, reaction, isReacted);
+	void Apps.self?.triggerEvent(AppEvents.IPostMessageReacted, message, user, reaction, isReacted);
```
MarcosSpessatto, line 87: Any reason for that being `await`ed here before? 🤔

apps/meteor/app/api/server/v1/chat.ts
```diff
@@ -364,19 +364,13 @@ API.v1.addRoute(
 				throw new Meteor.Error('error-messageid-param-not-provided', 'The required "messageId" param is missing.');
 			}
 
-			const msg = await Messages.findOneById(this.bodyParams.messageId);
-
-			if (!msg) {
-				throw new Meteor.Error('error-message-not-found', 'The provided "messageId" does not match any existing message.');
-			}
-
```
MarcosSpessatto, line 372: I think this is a breaking change, we are changing the error messages. Since we are validating it only inside the function, the error is different.

apps/meteor/tests/unit/app/reactions/server/setReaction.spec.ts
```diff
@@ -0,0 +1,406 @@
...
+					test: {
+						usernames: ['test'],
+					},
+				},
+			};
+
+			const result = removeUserReaction(message as any, 'test', 'test');
+			expect(result.reactions.test).to.be.undefined;
+		});
+		it('should remove username only from the reaction thats passed in', () => {
```
MarcosSpessatto, line 87: Can we add a test when the user is not present in the usernames?

apps/meteor/app/api/server/v1/chat.ts
MarcosSpessatto: We should also add a e2e test for this (if it doesn't exist)

MarcosSpessatto (changes requested):

apps/meteor/app/reactions/server/setReaction.ts
KevLehman: No reason, none of the actions in here can prevent a reaction from being set (or even change something from the reaction).
Callbacks as well were only used by federation & slackbridge so probably nothing we want to be awaiting.

apps/meteor/app/api/server/v1/chat.ts
KevLehman: There should be tests for chat.react endp.

I'm a bit in conflict on whether or not this is a breaking change 🤔 tho. Error won't be the exact same but will happen on the exact same condition, and the status code will be the same too. So, didn't saw this as a breaking change.

Can leave it back but that also introduces another DB call (that we don't actually need)

apps/meteor/app/lib/server/functions/isTheLastMessage.ts
KevLehman: There's a couple of unrelated places using this func and assuming the setting check it's inside so i'll create a task for changing it later.

MarcosSpessatto (approved):

MarcosSpessatto (changes requested):
Kevin and I agreed in split this PR into multiple smaller ones.

apps/meteor/app/reactions/server/setReaction.ts
```diff
@@ -101,33 +86,62 @@ async function setReaction(room: IRoom, user: IUser, message: IMessage, reaction
...
+
+		void callbacks.run('afterSetReaction', message, { user, reaction, shouldReact: true });
 
 		isReacted = true;
 	}
 
-	await Apps.self?.triggerEvent(AppEvents.IPostMessageReacted, message, user, reaction, isReacted);
+	void Apps.self?.triggerEvent(AppEvents.IPostMessageReacted, message, user, reaction, isReacted);
 
+	void notifyOnRoomChangedById(room._id);
```
MarcosSpessatto, line 98: This was triggered only if it was the last message...

MarcosSpessatto (changes requested):

apps/meteor/app/reactions/server/setReaction.ts
KevLehman: oops

sampaiodiego (approved):

PR #33013, opened 2024-08-08
feat: new stats

hugocostadev:


apps/meteor/app/statistics/server/lib/statistics.ts
```diff
@@ -157,6 +158,19 @@ export const statistics = {
...
 
+		// Number of livechat rooms with priorities
+		statistics.totalLivechatRoomsWithPriority = {
+			lowest: await LivechatRooms.countLivechatRoomsByPriority(LivechatPriorityWeight.LOWEST),
+			low: await LivechatRooms.countLivechatRoomsByPriority(LivechatPriorityWeight.LOW),
+			medium: await LivechatRooms.countLivechatRoomsByPriority(LivechatPriorityWeight.MEDIUM),
+			high: await LivechatRooms.countLivechatRoomsByPriority(LivechatPriorityWeight.HIGH),
+			highest: await LivechatRooms.countLivechatRoomsByPriority(LivechatPriorityWeight.HIGHEST),
+			notSpecified: await LivechatRooms.countLivechatRoomsByPriority(LivechatPriorityWeight.NOT_SPECIFIED),
+		};
```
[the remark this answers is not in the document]
hugocostadev: thanks for your feedback @ricardogarim ! I just fixed

tassoevan (approved):

sampaiodiego (dismissed):
I'm afraid the new queries might be too heavy for large deployments.. I'll take a look at them more carefully..

apps/meteor/server/models/raw/Users.js
```diff
@@ -3079,4 +3079,17 @@ export class UsersRaw extends BaseRaw {
 			},
 		);
 	}
+
+	countActiveUsersWithCustomRoles() {
+		const customRoleRegex = /^[0-9a-fA-F]{24}$/;
+		const query = {
+			active: true,
```
rique223, line 3086: This query only considers the active field but for a user to be truly active it has to have both active true and the lastLoggin field, otherwise, the stats would take pending users into consideration and this could possibly skew the statistics.

rique223 (changes requested):

sampaiodiego:
@hugocostadev you can help testing that as well.. 

if you can provide some `explain()` results, running the new queries in database with like 1 million rooms and 500k users, it would be **extremely** helpful.. ([explain docs](https://www.mongodb.com/docs/manual/reference/explain-results/#mongodb-data-explain.executionStats))

maybe even use https://github.com/KevLehman/filler and expand it to rooms

apps/meteor/server/models/raw/Users.js
hugocostadev: Thanks for the feedback @rique223 , I'm checking with product team

apps/meteor/server/models/raw/Users.js
```diff
@@ -3079,4 +3079,17 @@ export class UsersRaw extends BaseRaw {
 			},
 		);
 	}
+
+	countActiveUsersWithCustomRoles() {
+		const customRoleRegex = /^[0-9a-fA-F]{24}$/;
```
KevLehman, line 3084: Can u put a sample of what this regex is supposed to match?

apps/meteor/server/models/raw/Users.js
```diff
@@ -3079,4 +3079,17 @@ export class UsersRaw extends BaseRaw {
...
+		const query = {
+			active: true,
+			roles: {
```
KevLehman, line 3087: Can you post an `explain` of this one? Just to check if the query can use a suitable index for this kind of filter, cause `active` is not a very discriminative filter.

apps/meteor/server/models/raw/LivechatRooms.ts
```diff
@@ -2543,6 +2544,54 @@ export class LivechatRoomsRaw extends BaseRaw<IOmnichannelRoom> implements ILive
...
+					low: { $sum: '$low' },
+					medium: { $sum: '$medium' },
+					high: { $sum: '$high' },
+					highest: { $sum: '$highest' },
+					notSpecified: { $sum: '$notSpecified' },
+				},
+			},
+		];
+
+		const result = await this.col.aggregate(pipeline).toArray();
```
KevLehman, line 2584:
```suggestion
		const result = await this.col.aggregate<pass a type param>(pipeline).toArray();
```

apps/meteor/server/models/raw/LivechatRooms.ts
```diff
@@ -2543,6 +2544,54 @@ export class LivechatRoomsRaw extends BaseRaw<IOmnichannelRoom> implements ILive
...
+		const stats = result[0];
+		delete stats._id;
```
KevLehman, line 2586: You don't need to manually delete `_id`, you can `project { _id: 0 }` on the agg

apps/meteor/server/models/raw/Rooms.ts
```diff
@@ -868,6 +868,27 @@ export class RoomsRaw extends BaseRaw<IRoom> implements IRoomsModel {
 		return this.col.countDocuments({ federated: true });
 	}
 
+	async countMessagesInActiveFederatedRooms(): Promise<number> {
+		const pipeline = [
+			{
+				$match: {
+					federated: true,
+					$or: [{ archived: false }, { archived: { $exists: false } }],
```
KevLehman, line 876: do we need the `exists: false` for archived? Can we change it for something like `archived: $ne: true`? would that have the same result?

KevLehman (changes requested):

apps/meteor/app/statistics/server/lib/getAppsStatistics.ts
```diff
@@ -28,12 +33,23 @@ async function _getAppsStatistics(): Promise<AppsStatistics> {
...
 		let totalFailed = 0;
+		let totalPrivateApps = 0;
+		let totalPrivateAppsEnabled = 0;
 
 		await Promise.all(
 			apps.map(async (app) => {
 				totalInstalled++;
 
 				const status = await app.getStatus();
+				const storageItem = await app.getStorageItem();
```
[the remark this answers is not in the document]
hugocostadev: I agree, would be very nice to have this, but currently we don't

apps/meteor/server/models/raw/Users.js
hugocostadev: Sure, I'll add a comment, but this regex matches object IDs from MongoDB. Default roles are created with an user friendly ID.

apps/meteor/server/models/raw/Rooms.ts
hugocostadev: Yes, it would.... much better! Thanks man

hugocostadev:
**Explains:**

cc: @diegolmello @KevLehman 

Is those acceptable times? (1M rooms and 500k users)

`totalLivechatRoomsWithPriority`
image

`activeUsersWithCustomRoles`
image

apps/meteor/server/models/raw/Users.js
hugocostadev: Just posted

apps/meteor/server/models/raw/LivechatRooms.ts
```diff
@@ -2563,6 +2563,14 @@ export class LivechatRoomsRaw extends BaseRaw<IOmnichannelRoom> implements ILive
 			.toArray();
 	}
 
+	async countLivechatRoomsByPriority(): Promise<number> {
+		return this.col.countDocuments({ priorityWeight: { $exists: true, $ne: 99 } });
+	}
+
+	countLivechatRoomsWithDepartment(): Promise<number> {
+		return this.col.countDocuments({ departmentId: { $exists: true }, t: 'l' });
```
KevLehman, line 2571: Curious: is this using the partialFilterExp index or the sparse index on `departmentId`?

apps/meteor/server/models/raw/LivechatRooms.ts
```diff
@@ -2563,6 +2563,14 @@ export class LivechatRoomsRaw extends BaseRaw<IOmnichannelRoom> implements ILive
 			.toArray();
 	}
 
+	async countLivechatRoomsByPriority(): Promise<number> {
+		return this.col.countDocuments({ priorityWeight: { $exists: true, $ne: 99 } });
```
KevLehman, line 2567: We don't have an index for priority weight, I think this should be possible as well (the query) by checking on the priorityId prop (for which we already have an index)

can u try?

apps/meteor/app/statistics/server/lib/statistics.ts
```diff
@@ -157,6 +158,12 @@ export const statistics = {
 				}),
 		);
 
+		// Number of livechat rooms with priorities
+		statistics.totalLivechatRoomsWithPriority = await LivechatRooms.countLivechatRoomsByPriority();
+
+		// Number of livechat rooms with department
+		statistics.totalLivechatRoomsWithDepartment = await LivechatRooms.countLivechatRoomsWithDepartment();
```
KevLehman, line 165: Maybe you can add these 2 like the one in L168, by pushing the promise to the array of promises instead of doing them 1 by 1

hugocostadev: Good idea, I'll do that, thanks :D

apps/meteor/server/models/raw/LivechatRooms.ts
hugocostadev: I'll check

apps/meteor/server/models/raw/LivechatRooms.ts
hugocostadev: Indeed it is, and we already have this stat  😞  the stat is called `omnichannelRoomsWithPriorities` and you made it! haha

I thought that the `notSpecified`(99) had an associated ID, that's why I didn't go to this path.

Thanks a lot @KevLehman

KevLehman: lololol

apps/meteor/server/models/raw/LivechatRooms.ts
hugocostadev: The new approach (using priorityId) uses the sparse index...

```js
{ key: { priorityId: 1 }, sparse: true },
```

Reading  about Partitioned Filter Expression (PFE), seems that make more sense for this field, WDYT?

KevLehman: You can give it a try, check times and see if it's worth it to have a new index there 👀

KevLehman (dismissed):
Despite the perf implications of adding more stats, i think code looks good.

tassoevan (approved):

apps/meteor/server/models/raw/LivechatRooms.ts
```diff
@@ -2563,6 +2563,10 @@ export class LivechatRoomsRaw extends BaseRaw<IOmnichannelRoom> implements ILive
 			.toArray();
 	}
 
+	countLivechatRoomsWithDepartment(): Promise<number> {
+		return this.col.countDocuments({ departmentId: { $exists: true }, t: 'l' });
```
sampaiodiego, line 2567: I'd recommend to drop the `t: 'l'` filter so we can make good use of the sparse index on `departmentId`.. and we're sure whenever a room has a `departmentId` it will be `t: 'l'`..
```suggestion
		return this.col.countDocuments({ departmentId: { $exists: true } });
```

JFYI I have tested this in a collection with 1.7mi rooms and removing the `t: 'l'` from the filter reduced the query time from 34s to 800ms

apps/meteor/server/models/raw/Rooms.ts
```diff
@@ -868,6 +868,27 @@ export class RoomsRaw extends BaseRaw<IRoom> implements IRoomsModel {
 		return this.col.countDocuments({ federated: true });
 	}
 
+	async countMessagesInActiveFederatedRooms(): Promise<number> {
+		const pipeline = [
+			{
+				$match: {
+					federated: true,
```
sampaiodiego, line 875: although we already have a `countFederatedRooms` which filters by this field, we're missing an index here.. I'd suggest we remove this one from this PR and add it back in another one.

apps/meteor/server/models/raw/Users.js
```diff
@@ -3121,4 +3121,20 @@ export class UsersRaw extends BaseRaw {
 			},
 		);
 	}
+
+	async countActiveUsersWithCustomRoles() {
```
sampaiodiego, line 3125: I'm also not comfortable with this one.. I'd suggest we remove it from this PR so we can go ahead with other metrics that are ok, and we visit this one in a separate PR, wdyt?

sampaiodiego (changes requested):

sampaiodiego (approved):

ggazzo:
the tag `stat: ready to merge` was applied incorrectly here, as the milestone was 7.1.0 it shouldn't be ready for merge

PR #33023, opened 2024-08-09
fix: Realtime Monitoring LineCharts not updating

MartinSchoeler:
[SUP-637](https://rocketchat.atlassian.net/browse/SUP-637)

Today, the Realtime Charts have a couple of problems, like not being able to swap data labels, poor redrawing logic, and issues refreshing existing charts, this PR tries to mitigate the issues, but we will only fully fix this with a good refactor of this code

## Further comments

[SUP-637]: https://rocketchat.atlassian.net/browse/SUP-637?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


MarcosSpessatto (changes requested):
Is there any way of adding some tests here?

MartinSchoeler:
> Is there any way of adding some tests here?

I don't think so, all of the charts are drawn into canvases, and I don't think playwright supports canvas, I do have some ideas for testing, but I'll probably add those to the refactors to TS of the Charts that I'm working on

MartinSchoeler:
> Just wondering, why do we have two separate components for ChatsPerAgent and ChatsPerDepartment if they are exactly the same component with data from different endpoints?

To avoid a refactor, this fix will be backported, and we do not backport refactors, I'm currently working on a refactor that will resolve this issue

MarcosSpessatto (approved):

KevLehman (dismissed):

KevLehman (dismissed):

KevLehman (approved):

PR #33028, opened 2024-08-12
fix: Missing retention's `ignoreThreads` param for old channels

### LINKED ISSUE #32794: Threads are not getting pruned
### Description:

Pruning of threads is not working at least on channels that have channel based separate shorter retention time.

### Steps to reproduce:

1. Set global pruning in admin settings `admin -> settings -> retention policy` for `1 year` and apply it for everything. Disable option `Do not prune threads`. (Meaning *do* prune threads.)
2. Create a test channel and set retention time on that channel separately into `1 day`.
3. Create a thread and couple of messages under that thread on the channel. Also type some individual messages on the channel for comparison later.
4. Two days later check the messages on the channel.

### Expected behavior:

No messages can be found on the channel.

### Actual behavior:

Individual messages have been pruned correctly but the thread and all messages in it remain on the channel at least for one year.

### Server Setup Information:

- Version of Rocket.Chat Server: 6.9.3
- License Type: Enterprise
- Number of Users: -
- Operating System: RHEL8
- Deployment Method: tar
- Number of Running Instances: 2
- DB Replicaset Oplog: Yes
- NodeJS Version: 14.21.3
- MongoDB Version: 6.0

### Additional information:

The thread pruning functionality works in RC 6.10.0 for newly created channels.

The bug would be in the upgrade process that does not add `ignoreThreads: false` into channel's properties for the existing channels that have already had the channel based pruning enabled in RC 6.9.x. (Which did not have such option at all.)

dougfabris:

Introduced here: #32485

While introducing this new possibility in retention policy overriding feature, old channels were not being affected, because it was missing the `ignoreThreads` parameter. So we decide to add a migration to add the missing parameter in all channels with retention policy override. 

## Issue(s)

Closes #32794

## Further comments

[CORE-613]

[CORE-613]: https://rocketchat.atlassian.net/browse/CORE-613?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/meteor/server/startup/migrations/v306.ts
+++ b/apps/meteor/server/startup/migrations/v306.ts
@@ -14,9 +14,7 @@ addMigration({
 			},
 			{
 				$set: {
-					retention: {
-						ignoreThreads: false,
-					},
+					'retention.ignoreThreads': false,
 				},
 			},
 		);
--- a/apps/meteor/server/startup/migrations/v306.ts
+++ b/apps/meteor/server/startup/migrations/v306.ts
@@ -1,0 +1,24 @@
+import { Rooms } from '@rocket.chat/models';
+
+import { addMigration } from '../../lib/migrations';
+
+addMigration({
+	version: 305,
+	name: 'Adds missing ignoreThreads parameter for old rooms with retention policy overridden',
+	async up() {
+		await Rooms.updateMany(
+			{
+				'retention.enabled': true,
+				'retention.overrideGlobal': true,
+				'retention.ignoreThreads': { $exists: false },
+			},
+			{
+				$set: {
+					retention: {
+						ignoreThreads: false,
+					},
+				},
+			},
+		);
+	},
+});
--- a/.changeset/gold-knives-sparkle.md
+++ b/.changeset/gold-knives-sparkle.md
@@ -1,0 +1,5 @@
+---
+'@rocket.chat/meteor': patch
+---
+
+Fixes an issue where ignore threads parameter were not being affected by retention policy overriding in old channels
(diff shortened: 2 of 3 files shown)


MarcosSpessatto (approved):

apps/meteor/server/startup/migrations/v305.ts
```diff
@@ -0,0 +1,24 @@
...
+			{
+				'retention.enabled': true,
+				'retention.overrideGlobal': true,
+				'retention.ignoreThreads': { $exists: false },
+			},
+			{
+				$set: {
+					retention: {
+						ignoreThreads: false,
+					},
```
sampaiodiego, line 19:
```suggestion
					'retention.ignoreThreads': false,
```

sampaiodiego (changes requested):

ggazzo (dismissed):

ggazzo (approved):

apps/meteor/server/startup/migrations/v306.ts
```diff
@@ -0,0 +1,22 @@
+import { Rooms } from '@rocket.chat/models';
+
+import { addMigration } from '../../lib/migrations';
+
+addMigration({
+	version: 305,
```
matheusbsilva137, line 6: Oops, this was supposed to be `version: 306` :grimacing: can you fix this (maybe in a new `chore` PR)?
Otherwise I believe this migration won't run
cc @dougfabris

dougfabris: @matheusbsilva137 thanks for the heads-up. @ggazzo will handle it!

PR #33036, opened 2024-08-13
fix: `processRoomAbandonment` callback not processing data correctly

KevLehman:

- Prevent callback from failing when BH configuration doesn't include all fields
- Fix internal callback logic to properly calculate data

## Issue(s)

https://rocketchat.atlassian.net/browse/CORE-619


apps/meteor/app/livechat/server/hooks/processRoomAbandonment.ts
```diff
@@ -43,65 +44,88 @@ const getSecondsSinceLastAgentResponse = async (room: IOmnichannelRoom, agentLas
...
-		inactivityDay.add(1, 'days');
-	}
-	return totalSeconds;
-};
 
-callbacks.add(
-	'livechat.closeRoom',
-	async (params) => {
-		const { room } = params;
+		const officeHourStart = officeDay.start.time.split(':') as unknown as number[];
```
lucas-a-pelegrino, line 73: What do you think of destructing the array and naming the values `officeHoursStartHour` and `officeHoursStartMinutes`? Just a tiny suggestion but It might improve readability a bit.

KevLehman: def not against it 👀

MarcosSpessatto (changes requested):
I'd suggest to split the 2 fixes you mentioned in 2 different PRs, since we are changing the way things are calculated, this could lead to some unexpected future problems.

KevLehman:
Changed this PR to another task. Both issues will be fixed anyways as this changes the same code the other PR does.

MarcosSpessatto (dismissed):

MartinSchoeler (dismissed):

lucas-a-pelegrino (approved):

MarcosSpessatto (approved):

MartinSchoeler (approved):

PR #33041, opened 2024-08-13
feat: get contact by id endpoint

tapiarafael:

Implemented a new endpoint to retrieve contact info by id.

[GET] `api/v1/omnichannel/contacts.get?contactId=123`
response example:
```json
{
	"contact": {
		"_id": "123",
		"name": "john doe",
		"emails": [
			"johndoe@example.com"
		],
		"phones": [
			"1-658-751-2622"
		],
		"contactManager": "321",
		"customFields": {},
		"unknown": false,
	},
	"success": true
}
```

## Further comments

[SCI-36](https://rocketchat.atlassian.net/browse/SCI-36)

[SCI-36]: https://rocketchat.atlassian.net/browse/SCI-36?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/meteor/app/livechat/server/lib/Contacts.ts
```diff
@@ -248,6 +248,10 @@ export async function updateContact(params: UpdateContactParams): Promise<ILivec
 	return updatedContact;
 }
 
+export async function getContactById(contactId: string): Promise<ILivechatContact | null> {
+	return LivechatContacts.findOneById(contactId);
+}
+
```
MarcosSpessatto, line 254: I wonder if it makes sense to have this function for just calling the model, we could call that directly there, what do you think? Or maybe it will have more logic in the future?

MarcosSpessatto (changes requested):

apps/meteor/app/livechat/server/lib/Contacts.ts
tapiarafael: There's no more logic. I'll get rid of this function

MarcosSpessatto (approved):

tassoevan (approved):

KevLehman (approved):

PR #33042, opened 2024-08-13
chore!: removes view-history permission

lucas-a-pelegrino:

This PR aims to completely remove the `view-history` permission, since it has never been used since its creation. It also adds a migration to remove it from the `permissions` collection, and removes the **i18n** keys for `view-history` and `view-history-description`.

## Issue(s)

[CORE-605](https://rocketchat.atlassian.net/browse/CORE-605)

## Further comments

[CORE-605]: https://rocketchat.atlassian.net/browse/CORE-605?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ggazzo:
ps: there is no need to remove all the i18n keys, if you removes the `en` lingohub will do the rest for you

also the title should be `chore!:`

lucas-a-pelegrino:
> ps: there is no need to remove all the i18n keys, if you removes the `en` lingohub will do the rest for you
> 
> 
> 
> also the title should be `chore!:`

Good to now, I wasn't sure about the i18n being auto removed if I removed only in en.

.changeset/lovely-trees-call.md
matheusbsilva137, line 1: We usually don't have changesets for chore PRs, so it should be ok to exclude this

ggazzo (approved):

apps/meteor/server/startup/migrations/v305.ts
```diff
@@ -0,0 +1,10 @@
+import { Permissions } from '@rocket.chat/models';
+
+import { addMigration } from '../../lib/migrations';
+
+addMigration({
+	version: 305,
```
matheusbsilva137, line 6:
```suggestion
	version: 305,
	name: 'Delete unused view-history permission'
```
I know this is a very straightforward one, but I think it'd be good to have a name for this migration

matheusbsilva137 (changes requested):

.changeset/lovely-trees-call.md
ggazzo: in this case, as it is a "break" change, I think it is ok to keep it, since there is still a related migration

apps/meteor/server/startup/migrations/v305.ts
lucas-a-pelegrino: Agreed, I'll make the adjustments

.changeset/lovely-trees-call.md
matheusbsilva137: Great. Maybe we should just mention that it was **unused** then, wdyt @lucas-a-pelegrino ? Just so as to not scare users :eyes:

lucas-a-pelegrino: I think it makes sense to keep the changeset, considering the change being a breaking one.

As for adding more information, maybe: "Removes `view-history` permission due to lack of usage"?

apps/meteor/server/startup/migrations/v305.ts
```diff
@@ -0,0 +1,10 @@
...
+	async up() {
+		await Permissions.deleteOne({ _id: 'view-history' });
```
KevLehman, line 8: The `permissions` collection has `trash`, wonder if we should skip the trash on this delete, like
```
Permissions.col.deleteOne({ _id: 'view-history' })
```

lucas-a-pelegrino: Hey @KevLehman , can you elaborate? Not sure I understood what you meant.

KevLehman: What i mean is that we have something called `trash` for some DB models. This copies the object before it's deleted into the trash, and then it removes from the original collection.

SO in this case, when removed like this, the permission will be in the trash. Not really sure that's a big concern but would be nice to ask if we can just skip the trash instead for this case.

lucas-a-pelegrino: Got it.

@ggazzo , @matheusbsilva137 any thoughts on this?

MarcosSpessatto (approved):

ggazzo (approved):

PR #33047, opened 2024-08-14
fix: customFields ignored in livechat room creation

abhinavkrin:

The function `createLivechatRoom` does not take into account the customFields passed in the extraData param
https://github.com/RocketChat/Rocket.Chat/blob/05db8aa2232a8f85e570933a3f1316ef011c90fc/apps/meteor/app/livechat/server/lib/Helper.ts#L60

In various places the functions is passed customFields in extraData param. This extraData is passed to `livechat.beforeRoom` callback and expects to receive extraRoomInfo. But the customField is not taken into account in this callback

https://github.com/RocketChat/Rocket.Chat/blob/05db8aa2232a8f85e570933a3f1316ef011c90fc/apps/meteor/app/livechat/server/lib/Helper.ts#L85
https://github.com/RocketChat/Rocket.Chat/blob/42a176f2dcd38a6afae33764fda021929af0fabc/apps/meteor/ee/app/livechat-enterprise/server/hooks/beforeNewRoom.ts#L6-L29

This PR fixes the callback to also include the customFields in the result from the callback `livechat.beforeRoom`

I found this bug while investigating the issue related to on premise whatsapp app. In which, a new room was created everytime a message was sent from WA to RC. Since the custom fields was not being saved in the room data (due to the reason above), the wa app was not able to find existing room related to what whatsapp number and created a new room everytime.

I believe this bug can affect other apps as well as the rc where function is used.
## Further comments

[CORE-643](https://rocketchat.atlassian.net/browse/CORE-643)

[CORE-643]: https://rocketchat.atlassian.net/browse/CORE-643?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ggazzo (changes requested):
can we provide some kind of tests here?

apps/meteor/jest.config.ts
```diff
@@ -32,11 +32,16 @@ export default {
```
KevLehman, line 32: You dont need these for server unit tests

apps/meteor/tests/unit/app/livechat/server/hooks/beforeNewRoom.spec.ts
```diff
@@ -0,0 +1,21 @@
...
+
+proxyquire.noCallThru().load('../../../../../../ee/app/livechat-enterprise/server/hooks/beforeNewRoom.ts', {
+	'meteor/meteor': {
+		Meteor: {
+			Error,
+		},
+	},
+});
+
+describe('livechat.beforeRoom', () => {
```
KevLehman, line 14: Can you add more cases? What happens when extradata override a field from room? Also, when customFields is not an obj, etc

KevLehman (changes requested):

apps/meteor/jest.config.ts
abhinavkrin: The test was not being executed without adding its path. No matching path for the file was present.
Without the adding meteor to module mapper, the test had errors of not being able to resolve meteor/meteor

apps/meteor/ee/app/livechat-enterprise/server/hooks/beforeNewRoom.ts
```diff
@@ -10,9 +10,11 @@ callbacks.add(
 			return roomInfo;
 		}
 
-		const { sla: searchTerm } = extraData;
+		const { sla: searchTerm, customFields } = extraData;
+		const roomInfoWithExtraData = { ...roomInfo, ...(!!customFields && { customFields }) };
```
MarcosSpessatto, line 14: `!!customFields` => `Boolean(customFields)` or `customFields !== undefined` maybe?

apps/meteor/jest.config.ts
MarcosSpessatto: @abhinavkrin, unfortunately, that is a Meteor limitation, but Kevin is correct, this only happens because we are importing the callbacks file directly, we need to import it using proxyquire, otherwise, it will import the whole dep tree which includes meteor/meteor.

MarcosSpessatto (changes requested):

apps/meteor/tests/unit/server/livechat/hooks/beforeNewRoom.spec.ts
```diff
@@ -0,0 +1,42 @@
...
+	});
+
+	it('should not include field in roomInfo when extraData has field other than customFields, sla', async () => {
+		const roomInfo = { name: 'test' };
+		const extraData = { customFields: { test: 'test' }, sla: 'high' };
+		const result = await callbacks.run('livechat.beforeRoom', roomInfo, extraData);
+		expect(result).to.deep.equal({ ...roomInfo, customFields: extraData.customFields, slaId: 'slaId' });
+	});
+
+	it('should return roomInfo with empty customFields when customFields is not an object', async () => {
```
KevLehman, line 36:
```suggestion
	it('should return roomInfo with no customFields when customFields is not an object', async () => {
```

KevLehman (changes requested):

apps/meteor/tests/unit/server/livechat/hooks/beforeNewRoom.spec.ts
```diff
@@ -0,0 +1,53 @@
...
+		},
+	},
+	'@rocket.chat/models': {
+		OmnichannelServiceLevelAgreements: {
+			findOneByIdOrName: sinon.stub().callsFake((id) => {
+				if (id === 'high') {
+					return Promise.resolve({ _id: 'high' });
+				}
+				return Promise.resolve(null);
+			}),
```
KevLehman, line 21: A suggestion, you can create the `sinon.stub()` outside of here, and then on tests do something like `myStub.withArgs('high').resolves({_id: 'high'})`

should look cleaner

KevLehman: If you go this way, don't forget to `reset` your stub before using them again on other tests (to avoid conflicts)

KevLehman (approved):

KevLehman:
@abhinavkrin just one small thing: this PR should have its own task related to 589 instead of being linked directly to it as this fix will require it's own QA (and it's not related to the issue described on 589)

MarcosSpessatto (approved):

ggazzo (approved):

PR #33048, opened 2024-08-14
feat: add closedBy on apps-engine

ggazzo:

### DIFF
--- a/apps/meteor/app/apps/server/converters/rooms.js
+++ b/apps/meteor/app/apps/server/converters/rooms.js
@@ -134,6 +134,7 @@ export class AppRoomsConverter {
 			_USERNAMES: '_USERNAMES',
 			description: 'description',
 			source: 'source',
+			closer: 'closer',
 			isDefault: (room) => {
 				const result = !!room.default;
 				delete room.default;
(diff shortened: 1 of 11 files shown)


apps/meteor/app/apps/server/converters/rooms.js
```diff
@@ -210,6 +211,17 @@ export class AppRoomsConverter {
...
 				return this.orch.getConverters().get('departments').convertById(departmentId);
 			},
+			closedBy: async (room) => {
+				const { closedBy } = room;
+
+				if (!closedBy) {
+					return undefined;
+				}
+
+				delete room.servedBy;
```
KevLehman, line 221: closedBy and servedBy can hold diff data (ex, when a manager closes a conversation that an agent is serving)

apps/meteor/app/apps/server/converters/rooms.js
```diff
@@ -210,6 +211,17 @@ export class AppRoomsConverter {
...
+
+				return this.orch.getConverters().get('users').convertById(closedBy._id);
```
KevLehman, line 223: The `closedBy` can also be a visitor, we need something like:
```
if (closer === 'user') return this.orch.getConverters().get('users').convertById(closedBy._id);
return return this.orch.getConverters().get('visitors').convertById(convertById(closedBy._id));

File: apps/meteor/app/apps/server/converters/rooms.js (hunk 1 above)
Comment (ggazzo) [reply apps/meteor/app/apps/server/converters/rooms.js:221]: ops

Comment (ggazzo) [inline apps/meteor/app/apps/server/converters/rooms.js:221]:
```suggestion
				delete room.closedBy;
```

KevLehman (dismissed):

PR #33054, opened 2024-08-14
fix: Livechat analytics in a given date range consider conversation data from the following day

matheusbsilva137:

 - Fixed livechat analytics/metrics that were considering one extra day beyond of the provided date range. Affected metrics are:
   - **All overview productivity metrics** (`livechat/analytics/overview` endpoint): average of response time, average of first response time, average of reaction time (among all agents);
   - **All conversations metrics for agents** (`livechat/analytics/agent-overview` endpoint): total conversations (percentage of taken conversations), average chat duration and total messages;
   - **All productivity metrics for agents** (`livechat/analytics/agent-overview` endpoint): average first response time, best first response time, average response time and average reaction time (for each specific agent).

Charts metrics and conversation overview metrics should not be affected by this change (since this specific issue was not happening to them)

## Steps to test or reproduce

Go to **Omnichannel > Analytics** and select a date range in which there is not livechat activity in the workspace, but there is activity in the following day.
**_Current behavior:_** the analytics listed above would be affected by the data from conversation from the following day (ie the day after the selected period end date).
**_Expected behavior:_** only conversations in the selected date range should be considered to build all livechat analytics.

## Further comments

[CORE-610](https://rocketchat.atlassian.net/browse/CORE-610)

[CORE-610]: https://rocketchat.atlassian.net/browse/CORE-610?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


MarcosSpessatto (dismissed):
Would it be possible and also does it make sense to add some test cases to cover (also) cases where there is some data for that specific day and we can ensure that is calculating it properly?

Just to have more cases, not ONLY with 0 conversations.

matheusbsilva137:
> Would it be possible and also does it make sense to add some test cases to cover (also) cases where there is some data for that specific day and we can ensure that is calculating it properly?
> 
> Just to have more cases, not ONLY with 0 conversations.

We can't do that on the end-to-end tests since we only have data for the "current day" there 
But that's something we could validate in the unit tests. I'll add some new unit test cases for overview metrics

KevLehman (approved):

KevLehman (approved):

sampaiodiego (approved):

MarcosSpessatto (approved):

PR #33066, opened 2024-08-15
feat: introduce new categories field for UI Action Buttons

Dnouv:

- Adds a new button in the message hover toolbar; the button has an icon ✨ where the Apps can register the actions. Any App that mentions the `category: AI` on the App buttons, those buttons will be rendered on this new context.

## Further comments

- [Task](https://rocketchat.atlassian.net/browse/SAFE-51)
- [RC-AI](https://rocketchat.atlassian.net/browse/RCAI-16)
- [Depends on PR](https://github.com/RocketChat/Rocket.Chat.Apps-engine/pull/794)
- [Depends on](https://github.com/RocketChat/Rocket.Chat.Apps-engine/pull/804)


Dnouv:
The CI is failing due to a change introduced on the #33001 for Apps Engine.

Only requires FE review since modifying the packages has introduced those additional required reviews.

Depends on the package update in #33119

rique223 (changes requested):
This isn't necessary, but you might consider subdividing the `useMessageToolbarStarsAppsAction` logic into two functions like this:

`const filterActions = useCallback((actions) => { ...filtering logic... }, [{dependencies}]);`

`const mapActions = useCallback((actions) => { ...mapping logic... }, [{dependencies}]);`

Then, you can use them like this:
`const data = useMemo(
  () => mapActions(filterActions(queryResult.data)),
  [{dependencies}]
);`
This approach better adheres to functional programming best practices, improves readability, and makes future code extensions/refactors easier. However, if this suggestion isn't in the scope of the PR, feel free to disregard it.

apps/meteor/client/hooks/useAppActionButtons.ts
```diff
@@ -219,3 +219,63 @@ export const useMessageActionAppsActionButtons = (context?: MessageActionContext
...
+	const applyButtonFilters = useApplyButtonFilters();
+	const dispatchToastMessage = useToastMessageDispatch();
+	const { t } = useTranslation();
+	const data = useMemo(
+		() =>
+			result.data
+				?.filter((action) => {
+					if (
+						context &&
+						!(action.when?.messageActionContext || ['message', 'message-mobile', 'threads', 'starred']).includes(context as any)
```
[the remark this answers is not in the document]
Dnouv: Thank you for the review. This is out of the scope, just a clone of another function for rendering the Menu, so we can revisit this by modifying both places later.

rique223 (dismissed):

packages/ui-kit/src/interactions/UserInteraction.ts
```diff
@@ -98,6 +98,19 @@ export type MesssageActionButtonUserInteraction = {
 	triggerId: string;
 };
 
+export type MessageToolbarStarsActionButtonUserInteraction = {
+	type: 'actionButton';
+	actionId: string;
+	payload: {
+		context: 'messageToolbarStarsAction';
```
d-gubert, line 105: This context has been removed, right?

d-gubert (changes requested):

packages/ui-kit/src/interactions/UserInteraction.ts
Dnouv: Thank you, nice catch!

apps/meteor/client/components/message/toolbar/MessageToolbarStarsActionMenu.tsx
```diff
@@ -0,0 +1,97 @@
...
+const MessageToolbarStarsActionMenu = ({
+	options,
+	onChangeMenuVisibility,
+	context,
+	isMessageEncrypted,
+}: MessageActionMenuProps): ReactElement => {
+	const t = useTranslation();
+	const id = useUniqueId();
+	const groupOptions = options
+		.map((option) => ({
```
debdutdeb, line 35: Why map, reduce, map while you can do the whole in a single reduce?

apps/meteor/client/hooks/useFilterActions.ts
```diff
@@ -0,0 +1,24 @@
...
+const DEFAULT_CATEGORY = 'default';
+
+export const useFilterActionsByContextAndCategory = (context: string | undefined, category = 'default') => {
+	return useCallback(
+		(action: IUIActionButton) => {
+			if (!context) {
+				return true;
+			}
+
+			const actionCategory = action?.category ?? DEFAULT_CATEGORY;
```
debdutdeb, line 13: what if `action.category == ''`?

apps/meteor/client/hooks/useFilterActions.ts
```diff
@@ -0,0 +1,24 @@
...
+			const isContextMatch = (action.when?.messageActionContext || ['message', 'message-mobile', 'threads', 'starred']).includes(
+				context as any,
```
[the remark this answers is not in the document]
Dnouv: No, I don't follow. 

`context` can be a wide variety based on the `MessageActionContext` type so we only have two options from my understanding:
```
// Option 1:
const isContextMatch = (action.when?.messageActionContext || ['message', 'message-mobile', 'threads', 'starred']).includes(
	context as any, // Casting context as any
);

// Option 2:
const isContextMatch = (action.when?.messageActionContext || ['message', 'message-mobile', 'threads', 'starred'] as MessageActionContext[] // Casting as whole MessageActionContext
).includes(
	context, 
);
```

apps/meteor/client/components/message/toolbar/MessageToolbarStarsActionMenu.tsx
Dnouv: We possibly can, but I didn't want to modify this logic here since another function used the same and would have been a good addition as a separate PR. But if you feel like it's good to have here, I don't mind.

apps/meteor/client/hooks/useFilterActions.ts
Dnouv: Why would it be in what cases?

tassoevan (approved):

d-gubert (approved):

PR #33077, opened 2024-08-16
i18n: Required fields text to "{Field} required" format

rique223:

Currently, mandatory fields display the following error message in production: 'The field %s is required.' For example: 'The field email is required.' Product has decided to change the format of this message to '%s required' For example: 'Email required' Since this is the default 'Field required' message, updating it across all areas of the system is necessary. This PR aims to address that change.

Demo gif: 
form_example
## Further comments

Jira task: [CONN-314](https://rocketchat.atlassian.net/browse/CONN-314)

[CONN-314]: https://rocketchat.atlassian.net/browse/CONN-314?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


aleksandernsilva (dismissed):

packages/i18n/src/locales/en.i18n.json
```diff
@@ -4234,7 +4234,7 @@
   "Please_fill_all_the_information": "Please fill all the information",
   "Please_fill_an_email": "Please fill an email",
   "Please_fill_name_and_email": "Please fill name and email",
-  "Please_fill_out_reason_for_report": "Please fill out the reason for the report",
+  "Please_fill_out_reason_for_report": "Reason for report required",
```
tassoevan, line 4237: Looking at `apps/meteor/client/views/room/contextualBar/UserInfo/ReportUserModal.tsx`, it feels like this string only exists because of the improper design of the form. The field label in question visually is a user name, but in terms of a11y it should be “Reason for report” or “Description“. We can have both, but the former should be decorative (i.e. marked with `role="presentation"`). This would enable us to replace `register('description', { required: t('Please_fill_out_reason_for_report') })` with `register('description', { required: t('Required_field', { field: t('Reason_for_report') }) })`.

packages/ui-client/src/components/CustomFieldsForm.tsx
```diff
@@ -52,7 +52,7 @@ const CustomField = <T extends FieldValues>({
 		(error: RHFFieldError) => {
 			switch (error?.type) {
 				case 'required':
-					return t('The_field_is_required', label || name);
+					return t('Required_field', label || name);
```
tassoevan, line 55:
```suggestion
					return t('Required_field', { field: label || name });
```

tassoevan (changes requested):

tassoevan (dismissed):

tassoevan (approved):

aleksandernsilva (approved):

PR #33101, opened 2024-08-19
fix: adds temporary fix to teams created with no room.

lucas-a-pelegrino:

### DIFF
--- a/apps/meteor/server/services/team/service.ts
+++ b/apps/meteor/server/services/team/service.ts
@@ -76,9 +76,10 @@ export class TeamService extends ServiceClassInternal implements ITeamService {
 			roomId: '', // this will be populated at the end
 		};
 
+		let teamId = '';
 		try {
 			const result = await Team.insertOne(teamData);
-			const teamId = result.insertedId;
+			teamId = result.insertedId;
 			// the same uid can be passed at 3 positions: owner, member list or via caller
 			// if the owner is present, remove it from the members list
 			// if the owner is not present, remove the caller from the members list
@@ -137,7 +138,14 @@ export class TeamService extends ServiceClassInternal implements ITeamService {
 				_id: teamId,
 				...teamData,
 			};
-		} catch (e) {
+		} catch (error) {
+			// Manually deleting team and room if an error occurs, this is a temporary fix due to urgency on version 6.12
+			// and will be refactored on next release.
+			if (teamId) {
+				await Team.deleteOneById(teamId);
+				await TeamMember.deleteByTeamId(teamId);
+			}
+
 			throw new Error('error-team-creation');
 		}
 	}
@@ -433,16 +441,18 @@ export class TeamService extends ServiceClassInternal implements ITeamService {
 		}
 
 		const room = await Rooms.findOneById<Pick<IRoom, 'name'>>(team.roomId, { projection: { name: 1 } });
-		if (!room) {
-			throw new Error('invalid-room');
-		}
+		// if (!room) {
+		// 	throw new Error('invalid-room');
+		// }
 
 		const user = await Users.findOneById<Pick<IUser, '_id' | 'username' | 'name'>>(uid, { projection: { username: 1, name: 1 } });
 		if (!user) {
 			throw new Error('invalid-user');
 		}
 
-		await Message.saveSystemMessage('user-converted-to-channel', team.roomId, room.name || '', user);
+		if (room) {
+			await Message.saveSystemMessage('user-converted-to-channel', team.roomId, room.name || '', user);
+		}
 
 		await Rooms.unsetTeamId(teamId);
 	}
--- a/apps/meteor/server/services/team/service.ts
+++ b/apps/meteor/server/services/team/service.ts
@@ -139,8 +139,8 @@ export class TeamService extends ServiceClassInternal implements ITeamService {
 				...teamData,
 			};
 		} catch (error) {
-			// Manually deleting team and room if an error occurs, this is a temporary fix due to urgency on version 6.12
-			// and will be refactored on next release.
+			// Manually deleting team if an error occurs, this is a temporary fix due to urgency
+			// to release version 6.12 and will be refactored on next release.
 			if (teamId) {
 				await Team.deleteOneById(teamId);
 				await TeamMember.deleteByTeamId(teamId);


apps/meteor/server/services/team/service.ts
```diff
@@ -137,7 +138,14 @@ export class TeamService extends ServiceClassInternal implements ITeamService {
 				_id: teamId,
 				...teamData,
 			};
-		} catch (e) {
+		} catch (error) {
+			// Manually deleting team if an error occurs, this is a temporary fix due to urgency
+			// to release version 6.12 and will be refactored on next release.
+			if (teamId) {
+				await Team.deleteOneById(teamId);
+				await TeamMember.deleteByTeamId(teamId);
```
KevLehman, line 146: If this happens while converting a room into a team, then the room will still retain the invalid `teamId` set by `Rooms.setTeamMainById`. Do you think that's possible and we should take care of? Or nah?

lucas-a-pelegrino: The team should be valid in this case.

But it's definetly worth adding a way to handle this... I'm open to suggestions, tho. haha

lucas-a-pelegrino: The team should be valid in this case.

But it's definetly worth adding a way to handle this... I'm open to suggestions, tho. haha

matheusbsilva137: IMO it feels ok not to cover this since I believe we're only trying to cover errors "thrown by RC" here, not by the DB itself (in this scenario, `Room.create` is the only part that is really prone to errors here). If we'd like to cover errors from the DB here, then there'd be much more to undo (such as deleting messages, unsetting room as team's main, deciding if we should delete the room...). wdyt @KevLehman ?
Please correct me if I'm worng, @lucas-a-pelegrino

lucas-a-pelegrino: That's correct @matheusbsilva137 

This code I've added aims to avoid teams being created with no rooms associated with it for release 6.12.

If a room is being converted to a team, it's already created, therefore valid.

This catch block where I delete the Team and TeamMembers from the db will be removed once I've designed the final solution in the next release. We will have a robust solution capable of handling errors appropriately. That will involve more time to refactor that we don't have now, though.

apps/meteor/tests/end-to-end/api/teams.ts
```diff
@@ -47,6 +47,7 @@ describe('[Teams]', () => {
 
 		before(async () => {
 			testUser = await createUser();
+			await updateSetting('UTF8_Channel_Names_Validation', '[0-9a-zA-Z-_.]+');
```
MarcosSpessatto, line 50: Let's revert this back to default `after` all (even if it's the same value).

apps/meteor/tests/end-to-end/api/teams.ts
```diff
@@ -250,6 +251,36 @@ describe('[Teams]', () => {
...
+				})
+				.expect('Content-Type', 'application/json')
+				.expect(400)
+				.expect((response) => {
+					expect(response.body).to.have.property('success', false);
+					expect(response.body).to.have.property('error');
+					expect(response.body.error).to.be.equal('error-team-creation');
+				});
+
+			await request
```
MarcosSpessatto, line 271: Does it make sense to check for members as well? 🤔

MarcosSpessatto (changes requested):

apps/meteor/tests/end-to-end/api/teams.ts
lucas-a-pelegrino: I think it makes sense.

apps/meteor/server/services/team/service.ts
KevLehman: all right, thanks guys!

KevLehman (commented):
Curious, the process will rollback and that's cool. But, is there a way to prevent the process from failing? I was thinking if the issue is the room name being invalid, maybe we can call a function that checks for that before attempting to create the team, wdyt?

lucas-a-pelegrino:
I agree with you that we should add a naming validation for this. The issue is that this is not the only thing that may fail on Room creation, if you take a look at the Room.create() service, you can see that there are a lot of things that may fail, this fix aims to fix this specific edge case that causes a Team to be created with no room associated with it because of the naming (for direct api calls, from our UI it doesn't happen).

KevLehman:
No no yeah, that's great. I was thinking in having both, cause from your description it seems you already found a consistent way for this to happen. So, was thinking it's faster to cover that way earlier (before creating teams and doing more db calls) rather than waiting for the rollback to kick in 👀 

@MarcosSpessatto @lucas-a-pelegrino wdyt? does it makes sense or do we think on these improvements on the "actual" fix on the next release?

apps/meteor/tests/end-to-end/api/teams.ts
```diff
@@ -50,7 +50,10 @@ describe('[Teams]', () => {
 			await updateSetting('UTF8_Channel_Names_Validation', '[0-9a-zA-Z-_.]+');
 		});
 
-		after(() => Promise.all([...createdTeams.map((team) => deleteTeam(credentials, team.name)), deleteUser(testUser)]));
+		after(async () => {
+			await Promise.all([...createdTeams.map((team) => deleteTeam(credentials, team.name)), deleteUser(testUser)]);
+			await updateSetting('UTF8_Channel_Names_Validation', '[0-9a-zA-Z-_.]+');
+		});
```
matheusbsilva137, line 56:
```suggestion
		after(() =>
			Promise.all([
				...createdTeams.map((team) => deleteTeam(credentials, team.name)),
				deleteUser(testUser),
				updateSetting('UTF8_Channel_Names_Validation', '[0-9a-zA-Z-_.]+'),
			]),
		);
```

lucas-a-pelegrino:
> No no yeah, that's great. I was thinking in having both, cause from your description it seems you already found a consistent way for this to happen. So, was thinking it's faster to cover that way earlier (before creating teams and doing more db calls) rather than waiting for the rollback to kick in 👀
> 
> @MarcosSpessatto @lucas-a-pelegrino wdyt? does it makes sense or do we think on these improvements on the "actual" fix on the next release?

I think we can add the name validation early on in the method start, that would help but not resolve the whole issue... :/ A lot of things could still go wrong and a Team and TeamMember records could be created with no associated room. Even the TeamMember insertion could throw an unexpected error and fall into this edge case.

apps/meteor/server/services/team/service.ts
```diff
@@ -137,7 +138,14 @@ export class TeamService extends ServiceClassInternal implements ITeamService {
...
+			}
+
```
matheusbsilva137, line 148: What if we create the team only after the rooms has been succesfully created? This way we wouldn't need any DB calls in the catch statement

matheusbsilva137 (changes requested):

apps/meteor/tests/end-to-end/api/teams.ts
lucas-a-pelegrino: Wow, that's right. Hahaha 🫠

matheusbsilva137 (changes requested):
Please add a changeset to this PR
(btw we usually describe what's the error in the PR title, not the fix. Reference: https://developer.rocket.chat/docs/pull-requests-tags)

apps/meteor/server/services/team/service.ts
KevLehman: I think the room is created with teamId iirc

apps/meteor/server/services/team/service.ts
```diff
@@ -77,6 +78,28 @@ export class TeamService extends ServiceClassInternal implements ITeamService {
 		};
 
 		try {
+			let roomId = room.id;
+			if (roomId) {
+				await Message.saveSystemMessage('user-converted-to-team', roomId, team.name, createdBy);
```
KevLehman, line 83: We shouldn't send a message before doing the actions cause if some of the actions fail or we error out later, the message will still be there.

apps/meteor/server/services/team/service.ts
```diff
@@ -77,6 +78,28 @@ export class TeamService extends ServiceClassInternal implements ITeamService {
...
+			} else {
+				await getValidRoomName(team.name.trim(), undefined);
```
KevLehman, line 85:
```suggestion
				await getValidRoomName(team.name.trim());
```

apps/meteor/server/services/team/service.ts
lucas-a-pelegrino: > I think the room is created with teamId iirc

We can update the room's `teamId` afterwards.. I've added this change as well.

apps/meteor/tests/end-to-end/api/teams.ts
```diff
@@ -250,6 +257,51 @@ describe('[Teams]', () => {
 				.expect('Content-Type', 'application/json')
 				.expect(400);
 		});
+
+		it('should delete a team if an error occurs when creating a room for it', async () => {
```
KevLehman, line 261: is this test still valid? Cause i don't see any deletion going on now 👀

KevLehman (changes requested):

apps/meteor/server/services/team/service.ts
lucas-a-pelegrino: Nice catch!

apps/meteor/tests/end-to-end/api/teams.ts
lucas-a-pelegrino: I think the test is still valid, but the description is outdated...

MarcosSpessatto (dismissed):

apps/meteor/server/services/team/service.ts
```diff
@@ -77,6 +77,25 @@ export class TeamService extends ServiceClassInternal implements ITeamService {
...
+					members: memberUsernames as string[],
+					extraData: {
+						...room.extraData,
+					},
+					sidepanel,
+				};
+
+				const createdRoom = await Room.create(owner || uid, newRoom);
+				roomId = createdRoom._id;
+			}
```
ggazzo, line 97: I know that you just moved the code, but here is the tip:
in that cases stay away from `let`, and also there is no reason to create extra `const` if they are going to be used only once.

matheusbsilva137 (approved):

KevLehman (approved):

PR #33110, opened 2024-08-20
refactor: Replace proxy functions from `LivechatTyped` class with direct calls

KevLehman:

- Removes `findGuest`, replaced with a direct model call
- Removed `enabled`, replaced with calls to `settings.get`
- Removed `getNextAgent`, replaced with a direct call to `RoutingManager`
- Removed `setLastChat`, replaced with direct model call
- Removed `sendTranscript`, replaced with direct lib call (as this func was moved to its own file some days ago)

## Issue(s)

https://rocketchat.atlassian.net/browse/CORE-648


apps/meteor/app/livechat/server/api/lib/livechat.ts
```diff
@@ -48,15 +48,15 @@ async function findDepartments(
 	}));
 }
 
-export function findGuest(token: string): Promise<ILivechatVisitor | null> {
+export function findGuest(token: string, includeActivity = true): Promise<ILivechatVisitor | null> {
```
ggazzo, line 51: create a new function

KevLehman: done, thanks!

apps/meteor/app/livechat/server/api/v1/config.ts
```diff
@@ -12,7 +13,7 @@ API.v1.addRoute(
 	{ validateParams: isGETLivechatConfigParams },
 	{
 		async get() {
-			const enabled = Livechat.enabled();
+			const enabled = serverSettings.get('Livechat_enabled');
```
matheusbsilva137, line 16:
```suggestion
			const enabled = serverSettings.get<boolean>('Livechat_enabled');
```
Not a big deal, but it'd be good to have correct typing here as we had before

matheusbsilva137 (changes requested):

matheusbsilva137 (approved):

sampaiodiego (approved):

aleksandernsilva (approved):

PR #33092, opened 2024-08-19
refactor: Realtime `Doughnout Charts` to TS

MartinSchoeler:
## Issue(s)

[CORE-635](https://rocketchat.atlassian.net/browse/CORE-635)
## Further comments

Requires https://github.com/RocketChat/Rocket.Chat/pull/33076

[CORE-635]: https://rocketchat.atlassian.net/browse/CORE-635?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #33111, opened 2024-08-20
 refactor: remaining realtime charts to TS 

MartinSchoeler:
## Issue(s)

https://rocketchat.atlassian.net/browse/CORE-633

## Further comments

Depends on #33092


apps/meteor/app/livechat/client/lib/chartHandler.ts
```diff
@@ -108,7 +109,7 @@ const doughnutChartConfiguration = (
 
 type ChartDataSet = {
 	label: string;
-	data: number;
+	data: number[];
```
MarcosSpessatto, line 112: Is this type change expected?

apps/meteor/app/livechat/client/lib/chartHandler.ts
```diff
@@ -128,12 +129,11 @@ export const drawLineChart = async (
 	chartContext: { destroy: () => void } | undefined,
 	chartLabels: string[],
 	dataLabels: string[],
-	dataSets: number[],
+	dataSets: number[][],
```
MarcosSpessatto, line 132: Same here

MarcosSpessatto (changes requested):

apps/meteor/app/livechat/client/lib/chartHandler.ts
MartinSchoeler: Yeah, it was wrong before, but since the files that consumed this type were all in js, no error was raised

apps/meteor/app/livechat/client/lib/chartHandler.ts
MartinSchoeler: Yeah, it was wrong before, but since the files that consumed this type were all in js, no error was raised

MartinSchoeler:
done in https://github.com/RocketChat/Rocket.Chat/pull/33620

PR #33138, opened 2024-08-23
chore: remove weird type assertions

MartinSchoeler:
No comments


MartinSchoeler:
Closing this, mister @dougfabris will cure us of this evil

PR #33139, opened 2024-08-23
feat: Allow admins to control if visitors can close omnichannel conversations

KevLehman:

Visitors have always had the ability of closing ongoing conversations. This, however, means a visitor can close a conversation before it has been served or before an automation process has finished on customer side, which can cause issues.

New setting `Allow visitors to finish conversations` is created with this purpose. By default, the same behavior will be kept, where visitors can close conversations. When the setting is off, visitors won't see the button to close a conversation and the API will throw an error for them. This doesn't affect agent's capabilities of closing conversations.
## Issue(s)

https://rocketchat.atlassian.net/browse/CORE-632


packages/i18n/src/locales/en.i18n.json
```diff
@@ -4035,6 +4035,8 @@
   "Omnichannel_Reports_Summary": "Gain insights into your operation and export your metrics.",
   "Omnichannel_max_fallback_forward_depth": "Maximum fallback forward departments depth",
   "Omnichannel_max_fallback_forward_depth_Description": "Maximum number of hops that a room being transfered will do when the target department has a Fallback Forward Department set up. When limit is reached, chat won't be transferred and process will stop. Depending on your configuration, setting a high number may cause performance issues.",
+  "Omnichannel_allow_visitors_to_close_conversation": "Allow visitors to close ongoing conversations",
```
KevLehman, line 4038:
```suggestion
  "Omnichannel_allow_visitors_to_close_conversation": "Allow visitors to finish conversations",
```

packages/i18n/src/locales/en.i18n.json
```diff
@@ -4035,6 +4035,8 @@
...
+  "Omnichannel_allow_visitors_to_close_conversation_Description": "When disabled, visitors won't be able to close an ongoing conversation either via UI or via API.",
```
KevLehman, line 4039:
```suggestion
  "Omnichannel_allow_visitors_to_close_conversation_Description": "When disabled, visitors won't be able to finish an ongoing conversation either via UI or via API.",
```

apps/meteor/tests/end-to-end/api/livechat/00-rooms.ts
```diff
@@ -689,6 +689,24 @@ describe('LIVECHAT - rooms', () => {
...
 		);
+
+		describe('Special case: visitors closing is disabled', () => {
+			before(async () => {
+				await updateSetting('Omnichannel_allow_visitors_to_close_conversation', false);
+			});
+			after(async () => {
+				await updateSetting('Omnichannel_allow_visitors_to_close_conversation', true);
+			});
+			it('should not allow visitor to close a conversation', async () => {
```
KevLehman, line 700:
```suggestion
			it('should not allow visitor to close a conversation', async () => {
				const {
					room,
					visitor,
				} = await startANewLivechatRoomAndTakeIt();
```

.changeset/healthy-rivers-nail.md
```diff
@@ -0,0 +1,8 @@
+---
+"@rocket.chat/meteor": patch
+"@rocket.chat/i18n": patch
+"@rocket.chat/livechat": patch
```
matheusbsilva137, line 4:
```suggestion
"@rocket.chat/meteor": minor
"@rocket.chat/i18n": minor
"@rocket.chat/livechat": minor
```

apps/meteor/app/livechat/server/api/v1/room.ts
```diff
@@ -98,6 +98,10 @@ API.v1.addRoute(
 		async post() {
 			const { rid, token } = this.bodyParams;
 
+			if (!rcSettings.get('Omnichannel_allow_visitors_to_close_conversation')) {
+				throw new Error('error-not-allowed-to-close-conversation');
```
matheusbsilva137, line 102: I'm thinking, shouldn't we return `API.v1.unauthorized()` instead? I believe this fits more the idea than a regular error (which will return code 400), but I'd like to know what you think

matheusbsilva137 (changes requested):

apps/meteor/app/livechat/server/api/v1/room.ts
KevLehman: Not sure, technically the visitor doesn't have any permission attached. And they never log in so for them there's no such thing as being authorized or not...

But since the wording says "not allowed" then it may be confusing 🤔 

So, not sure lol. I guess no other endpoint returs 401/403 for any visitor action so im inclined to leave the 400

lucas-a-pelegrino: Shouldn't we just hide/not render the close conversation button from the user if they can't close the conversation?

Also, I think 403 forbidden is a more appropriate status code.

packages/livechat/src/routes/Chat/container.js
```diff
@@ -288,8 +288,8 @@ class ChatContainer extends Component {
 	};
 
 	canFinishChat = () => {
```
lucas-a-pelegrino, line 290: What do you think of naming this `canCloseChat` for the sake of consistency?

KevLehman: We use this term close/finish indiscriminately on the codebase :/ i can for sure rename "visitorCanCloseChat" to "visitorCanFinishChat"

(I didnt modify this func tho, was already there)

apps/meteor/app/livechat/server/api/v1/room.ts
KevLehman: Yep, thats exactly on the other changes ;)

Again, I'm really not sure. We don't throw any 401/403 errors on these visitor endpoints because technically they're not authenticated, and this is not a permission of them, it's a configuration.

KevLehman: (We also have a test of the button not appearing btw)

lucas-a-pelegrino: Sorry! I made the comment before noticing those changes and forgot to update the comment.

lucas-a-pelegrino: As for the status code.

403 would be the more appropriate status for this, but... if we have a standard of using the 400 instead, we should keep to it to avoid confusing, or change all of them to use the appropriate status code... but that is another subject. haha

matheusbsilva137: @MarcosSpessatto would like to hear your opinion about this
I think Kevin has a good point, but we do use `unauthorized()` for similar use cases out of the livechat context. So I'm not sure

KevLehman: Can you share some samples? 🤗 if we have enough maybe we can start with the new pattern, otherwise I'd think it's even better to change them lol.

MarcosSpessatto: I agree with the pattern, even if 403 looks nicer and more appropriate, I agree with follow the pattern we already have

apps/meteor/tests/e2e/omnichannel/omnichannel-livechat.spec.ts
```diff
@@ -93,6 +93,58 @@ test.describe.serial('OC - Livechat', () => {
 	});
 });
 
+test.describe.serial('OC - Livechat - Visitors closing the room is disabled', () => {
+	let poLiveChat: OmnichannelLiveChat;
+	let poHomeOmnichannel: HomeOmnichannel;
+
+	test.beforeAll(async ({ api }) => {
+		const statusCode = (await api.post('/livechat/users/agent', { username: 'user1' })).status();
+		expect(statusCode).toBe(200);
```
MarcosSpessatto, line 102: We don't need to assert it here, let's call only the endpoint and that's it

MarcosSpessatto (changes requested):

apps/meteor/tests/e2e/omnichannel/omnichannel-livechat.spec.ts
KevLehman: Hmmm, i was copying this from the rest of the tests. imho there's some value on this as the test would fail when this call returns other than 200, so this is a "fail fast" check. not against removing it tho

MarcosSpessatto (dismissed):

apps/meteor/tests/e2e/omnichannel/omnichannel-livechat.spec.ts
```diff
@@ -93,6 +93,57 @@ test.describe.serial('OC - Livechat', () => {
...
+	});
+
+	test.beforeAll(async ({ browser, api }) => {
+		const { page: livechatPage } = await createAuxContext(browser, Users.user1, '/livechat', false);
+
+		poLiveChat = new OmnichannelLiveChat(livechatPage, api);
+	});
+
+	test.beforeAll(async ({ browser, api }) => {
+		await api.post('/settings/Omnichannel_allow_visitors_to_close_conversation', { value: false });
```
matheusbsilva137, line 111: I'd suggest using our `setSettingValueById` helper instead of calling the endpoint directly

apps/meteor/tests/e2e/omnichannel/omnichannel-livechat.spec.ts
```diff
@@ -93,6 +93,57 @@ test.describe.serial('OC - Livechat', () => {
...
+		const { page: omniPage } = await createAuxContext(browser, Users.user1, '/', true);
+		poHomeOmnichannel = new HomeOmnichannel(omniPage);
+	});
+
+	test.afterAll(async ({ api }) => {
+		await api.post('/settings/Omnichannel_allow_visitors_to_close_conversation', { value: true });
```
matheusbsilva137, line 117: I'd suggest using our `setSettingValueById` helper instead of calling the endpoint directly

matheusbsilva137 (changes requested):

matheusbsilva137 (approved):

lucas-a-pelegrino (approved):

MartinSchoeler (approved):

MarcosSpessatto (approved):

sampaiodiego (approved):

PR #33141, opened 2024-08-23
fix: VideoConf message not considers display avatar pref

dougfabris:
## Further comments

[CORE-656]

[CORE-656]: https://rocketchat.atlassian.net/browse/CORE-656?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/meteor/tests/e2e/video-conference.spec.ts
```diff
@@ -35,6 +35,24 @@ test.describe('video conference', () => {
...
+
+		test.beforeAll(async ({ api }) => {
+			await setUserPreferences(api, { displayAvatars: false });
+		});
+
+		test.afterAll(async ({ api }) => {
+			await setUserPreferences(api, { displayAvatars: true });
+		});
+
+		test('should not render avatars in video conference message block', async () => {
```
MarcosSpessatto, line 49: Does it make sense to have different combinations like: should render the avatars if the user preference is true, also, what should happen when the list contains user with and without avatars?

MarcosSpessatto (changes requested):

apps/meteor/tests/e2e/video-conference.spec.ts
dougfabris: I don't see why making a test to check the default assertion for this preference, we already have a test to check if the message block is there and the preference is true by default, so I added a test just to check if its ignoring the avatars properly when changing the preference. 

> what should happen when the list contains user with and without avatars
Basically what I'm testing, if the avatars are there or not

packages/fuselage-ui-kit/src/blocks/VideoConferenceBlock/VideoConferenceBlock.tsx
```diff
@@ -97,6 +101,23 @@ const VideoConferenceBlock = ({
     }
   };
 
+  const getMessageFooterText = useCallback(
```
gabriellsh, line 104: I don't think this has to be a useCallback, since it's just called below, it won't trigger rendering.

Maybe instead we could have a useMemo, or define this function outside the component?

packages/fuselage-ui-kit/src/blocks/VideoConferenceBlock/VideoConferenceBlock.tsx
```diff
@@ -105,6 +126,21 @@ const VideoConferenceBlock = ({
   const { data } = result;
   const isUserCaller = data.createdBy._id === userId;
 
+  const messageFooterText = getMessageFooterText(data.users.length);
+
+  const joinedUsernames = [...data.users]
+    .splice(0, MAX_USERS)
+    .map(({ name, username }) => (showRealName ? name : username))
```
gabriellsh, line 133: Name can be undefined here.

packages/fuselage-ui-kit/src/blocks/VideoConferenceBlock/VideoConferenceBlock.tsx
```diff
@@ -105,6 +126,21 @@ const VideoConferenceBlock = ({
...
+    .join(', ');
+
+  const iconTitle =
+    data.users.length > MAX_USERS
+      ? t('__usernames__and__count__more_joined', {
+          usernames: joinedUsernames,
+          count: data.users.length - MAX_USERS,
+        })
+      : t('__usernames__joined', { usernames: joinedUsernames });
```
gabriellsh, line 142: Shouldn't this respect the `useRealName` setting aswell?

gabriellsh (changes requested):

apps/meteor/tests/e2e/video-conference.spec.ts
gabriellsh: There's never "users without avatar" afaik. If the user has no avatar it's the default one...

packages/ui-video-conf/src/VideoConfMessage/VideoConfMessageUserStack.tsx
```diff
@@ -1,22 +1,32 @@
...
-					<Box mi={2} key={index}>
-						<Avatar size='x28' alt={username || ''} data-tooltip={username} url={getUserAvatarPath(username as string)} />
-					</Box>
-				))}
-			</Box>
+			{displayAvatars && (
+				<Box display='flex' alignItems='center' mi='neg-x2'>
+					{users.slice(0, MAX_USERS).map(({ name, username }, index) => (
+						<Box mi={2} key={index}>
+							<Avatar size='x28' alt={username || ''} title={showRealName ? name : username} url={getUserAvatarPath(username as string)} />
```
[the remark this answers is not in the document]
dougfabris: yep, let's replace it with the hook we're isolating in `ui-client`

packages/fuselage-ui-kit/src/blocks/VideoConferenceBlock/VideoConferenceBlock.tsx
dougfabris: yep, let's replace it with the hook we're isolating in `ui-client`

packages/fuselage-ui-kit/src/blocks/VideoConferenceBlock/VideoConferenceBlock.tsx
dougfabris: > Shouldn't this respect the `useRealName` setting aswell?

yep, let's replace it with the hook we're isolating in `ui-client`

gabriellsh (dismissed):

gabriellsh (approved):

MarcosSpessatto (dismissed):

PR #33163, opened 2024-08-27
feat: Adding characters limitation to file upload descriptions

csuadev:
This PR adds the characters limitation from the message settings to avoid descriptions in file uploads to exceed this.

Change copys with better error descriptions according to design.

image

## Issue(s)
[https://rocketchat.atlassian.net/browse/SUP-620](https://rocketchat.atlassian.net/browse/SUP-620)


MarcosSpessatto (changes requested):

apps/meteor/client/views/room/modals/FileUploadModal/FileUploadModal.tsx
```diff
@@ -60,6 +62,11 @@ const FileUploadModal = ({
 			});
 		}
 
+		if (description.length >= maxCharactersAllowed) {
```
MarcosSpessatto, line 65: Shouldn't it be `>` only? Shouldn't we accept it if it's equals to the limitation?

apps/meteor/tests/e2e/file-upload.spec.ts
```diff
@@ -75,7 +79,37 @@ test.describe.serial('file-upload', () => {
...
+		await expect(poHomeChannel.content.lastMessageFileName).not.toContainText('any_file1.txt');
+	});
+
+	test('Upload file exceeding message characters limit', async ({ api, page }) => {
+		const charactersLimit = 100;
+		await setSettingValueById(api, 'Message_MaxAllowedSize', charactersLimit);
+
+		await page.reload();
+
+		await test.step('expect showing an error message when description is too long', async () => {
```
MarcosSpessatto, line 98: Can we add a test when the description is equal to the limitation?

MarcosSpessatto (changes requested):

PR #33193, opened 2024-09-02
fix: resolve avatar download issue on setUsername by refining service selection logic

ricardogarim:
As per [CORE-650](https://rocketchat.atlassian.net/browse/CORE-650), this pull request addresses the issue where the avatar blob image was not being correctly set in the `setUserAvatar` method, resulting in an undefined error for the user. The root cause was identified as incorrect handling of avatar service selection, which led to errors during the avatar setting process.

Key changes:
- Corrected the service handling logic to ensure the avatar blob is properly set from the first available service.
- Enhanced the process to prioritize non-gravatar services while maintaining a fallback to gravatar if no other service is available.

[CORE-650]: https://rocketchat.atlassian.net/browse/CORE-650?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


MarcosSpessatto (changes requested):
Is there any way we could add some tests here?

apps/meteor/tests/unit/app/lib/server/functions/setUsername.spec.ts
```diff
@@ -0,0 +1,244 @@
...
+			stubs.checkUsernameAvailability.resolves(true);
+			stubs.settings.get.withArgs('Accounts_SetDefaultAvatar').returns(true);
+			stubs.getAvatarSuggestionForUser.resolves({ gravatar: { blob: 'blobData', contentType: 'image/png' } });
+
+			await _setUsername(userId, username, mockUser);
+
+			expect(stubs.setUserAvatar.calledOnceWith(mockUser, 'blobData', 'image/png', 'gravatar')).to.be.true;
+		});
+
+		it('should not set avatar if Accounts_SetDefaultAvatar is disabled', async () => {
```
MarcosSpessatto, line 205: Ca we add a case when there is no `!previousUsername`?

MarcosSpessatto (changes requested):

sampaiodiego (approved):
for a further optimization, when calling `setUsername` we don't need to call `getAvatarSuggestionForUser` which will fetch avatar for EVERY SERVICE available, we just need the first one =)

PR #33209, opened 2024-09-04
fix: `LivechatSessionTaken` webhook event called without `agent` param

KevLehman:

Basically, the call to `assignAgent` was updating the room object as expected, but when calling `afterTakeInquiry` callback we were passing the room _before_ it was updated.
## Issue(s)

https://rocketchat.atlassian.net/browse/SUP-654
## Further comments

I'm planning to spin up a `webhook.test` docker image to test this one but probably in the future 😬


MarcosSpessatto (approved):

apps/meteor/app/livechat/server/lib/RoutingManager.ts
```diff
@@ -265,11 +265,21 @@ export const RoutingManager: Routing = {
...
+			throw new Error('error-room-not-found');
+		}
+
 		callbacks.runAsync(
 			'livechat.afterTakeInquiry',
 			{
-				inquiry: await this.assignAgent(inquiry as InquiryWithAgentInfo, room, agent),
-				room,
+				inquiry: inq,
+				// @eslint-disable-next-line @typescript-eslint/no-non-null-assertion
```
ggazzo, line 281: nop

ggazzo (changes requested):

apps/meteor/app/livechat/server/lib/RoutingManager.ts
ggazzo: remove the eslint/ts disable

KevLehman: ah, remanent of first fix.

MarcosSpessatto (changes requested):
~~Is there any way of adding tests?~~ Just saw the description

MarcosSpessatto (approved):

sophia-devs:
Can we get an update on this?

MartinSchoeler (approved):

PR #33212, opened 2024-09-04
feat: Adds new admin feature preview setting management

lucas-a-pelegrino:

This PR adds the new admin feature preview pages, where workspace admins can enable feature previewing and manage what features will be enabled by default for all users.

ADR: https://adr.rocket.chat/0103

## Issue(s)

[CORE-654](https://rocketchat.atlassian.net/browse/CORE-654)

## Steps to test or reproduce

As a admin, go to Workspace management, the feature preview page link should be present in the sidebar.

## Further comments

image

This PR has a few technical debts; for instance, in order to save the default feature preview list in our settings, I had to use transform the list into a string, and cast it back to an array of objects when I'm using it. Same as when I'm modifying the default settings. This adds complexity that could be avoided if we could save direct objects in the settings. 

I understand there's `select` and `multiSelect` configuration that allows us to save arrays of objects, but that would not be appropriate for this particular setting as its purpose is to store a list of objects that builds the feature previews configuration, such as if its a feature that we made available for use, or if its enabled in the workspace, translations keys, etc. It's a logical setting that we use on the background and the admin sets as a default preference for the workspace users.

Another issue is that we currently have the new features hard coded in the front end, this was fine when there was no default setting configuration. But now we actually use this information in the backend and we would have to duplicate the list, or... as I implemented, save an empty array initially that will be populated with the list once the user access the feature preview page for the first time.

We could work on a directory that is used by both BE and FE to collect data such as this one. This way we have a single source of truth and no need to duplicate data or add workarounds in the code to accommodate the requirements.

[CORE-654]: https://rocketchat.atlassian.net/browse/CORE-654?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.changeset/quick-rings-wave.md
dougfabris, line 1: since it's a feat, should be a minor

apps/meteor/client/views/account/featurePreview/AccountFeaturePreviewPage.tsx
```diff
@@ -105,15 +105,16 @@ const AccountFeaturePreviewPage = () => {
...
-							<Box
-								className={css`
-									white-space: break-spaces;
-								`}
-								pbe={24}
-								fontScale='p1'
-							>
-								{t('Feature_preview_page_description')}
-							</Box>
+							<FieldGroup marginBlockEnd={16}>
```
dougfabris, line 108: `FieldGroup` and `Field` should be used to form fields, you can use `Box` instead

apps/meteor/client/views/admin/featurePreview/AdminFeaturePreviewPage.tsx
```diff
@@ -0,0 +1,172 @@
...
+import type { ChangeEvent } from 'react';
+import React, { useEffect, Fragment } from 'react';
+import { useForm } from 'react-hook-form';
+
+import { Page, PageHeader, PageScrollableContentWithShadow, PageFooter } from '../../../components/Page';
+import { useEditableSetting } from '../EditableSettingsContext';
+import Setting from '../settings/Setting';
+import SettingsGroupPageSkeleton from '../settings/SettingsGroupPage/SettingsGroupPageSkeleton';
+
+const handleEnableQuery = (features: FeaturePreviewProps[]) => {
```
dougfabris, line 27: since it's the same `handleEnableQuery` we use in the accounts page, lets try to abstract into a hook

apps/meteor/client/views/admin/featurePreview/AdminFeaturePreviewPage.tsx
```diff
@@ -0,0 +1,172 @@
...
+		// TODO: Implement FeaturePreviewSkeleton component
+		return <SettingsGroupPageSkeleton />;
+	}
+
+	return (
+		<Page>
+			<PageHeader title={t('Feature_preview')} />
+			<PageScrollableContentWithShadow>
+				<Box maxWidth='x600' w='full' alignSelf='center'>
+					<FieldGroup marginBlockEnd={16}>
```
dougfabris, line 115: lets avoid `FieldGroup` and `Field` when we don't have form fields

dougfabris (changes requested):
Some expected behavior I'm missed:
1. The badge with the number of not viewed features in the new admin sidebar item
Screenshot 2024-09-06 at 16 41 39

2. The feature preview list should be shown despite being enabled or not
Screenshot 2024-09-06 at 16 41 58

3. The badge with the number of not viewed features in the user account sidebar item 
Screenshot 2024-09-06 at 16 43 32

4. Setting a feature preview as default doesn't reflect on the user preferences
5. Even activating the feature on user preferences, the desired feature doesn't work

packages/i18n/src/locales/en.i18n.json
```diff
@@ -1125,8 +1126,8 @@
   "Common_Access": "Common Access",
   "Commit": "Commit",
   "Community": "Community",
-  "Contextualbar_resizable": "Contextual bar resizable",
-  "Contextualbar_resizable_description": "Allows you to adjust the size of the contextual bar by simply dragging, giving you instant customization and flexibility",
+  "Contextualbar_resizable": "Resizable contextual bar",
```
gabriellsh, line 1129: You can't change translations without changin the keys. You have to either remove this key and create a new one, or you have to delete all keys from the other languages, so that the strings get translated again.

Only cases you can edit translations is if you need to fix something like a typo that does not change it's meaning or any terms that would affect other translations.

gabriellsh (changes requested):
Other from the translations, we still have to decide the behaviour when changing the settings, as we discussed in DM

packages/i18n/src/locales/en.i18n.json
lucas-a-pelegrino: I've checked the other i18n languages, but it doesn't seem they have these keys defined yet...

tapiarafael (approved):

.changeset/quick-rings-wave.md
```diff
@@ -0,0 +1,7 @@
+---
+"@rocket.chat/meteor": patch
+"@rocket.chat/i18n": patch
+"@rocket.chat/ui-client": patch
```
KevLehman, line 4: Should be minor imho

apps/meteor/server/settings/accounts.ts
```diff
@@ -743,6 +743,11 @@ export const createAccountSettings = () =>
 				i18nLabel: 'Sidebar_Sections_Order',
 				i18nDescription: 'Sidebar_Sections_Order_Description',
 			});
+
+			await this.add('Accounts_Default_User_Preferences_featuresPreview', '[]', {
```
KevLehman, line 747: I think we have a "multiselect" type of setting that allows you to store an array of elements. That would be better suitable than this string

packages/ui-client/src/hooks/useFeaturePreviewList.ts
```diff
@@ -82,22 +87,46 @@ export const defaultFeaturesPreview: FeaturePreviewProps[] = [
...
 
 	if (!featurePreviewEnabled) {
 		return { unseenFeatures: 0, features: [] as FeaturePreviewProps[], featurePreviewEnabled };
 	}
+	return { unseenFeatures, features, featurePreviewEnabled };
+};
+
+export const useDefaultSettingFeaturePreviewList = () => {
+	const featurePreviewSettingJSON = useSetting<string>('Accounts_Default_User_Preferences_featuresPreview');
+	// TODO: Remove this logic after we have a way to store object settings.
```
KevLehman, line 110: If this is the limitation i wonder if it would be better to extend `multiselect` or create a new setting type that supports this rather than manually doing the parsing when the setting is called.

.changeset/quick-rings-wave.md
lucas-a-pelegrino: Its a new feature in the workspace settings, do we usually add things like this as a minor?

KevLehman: Yep

apps/meteor/client/hooks/useFeaturePreviewEnableQuery.ts
```diff
@@ -0,0 +1,24 @@
+import type { FeaturePreviewProps } from '@rocket.chat/ui-client';
+
+export const useFeaturePreviewEnableQuery = (features: FeaturePreviewProps[]) => {
+	return Object.entries(
+		features
+			.map((item) => {
+				if (item.enableQuery) {
```
KevLehman, line 7: nitpick:
```
if (!item.enableQuery) {
	return item
}
...
```

apps/meteor/client/views/admin/featurePreview/AdminFeaturePreviewPage.tsx
```diff
@@ -0,0 +1,130 @@
...
+	} = useForm({
+		defaultValues: { featuresPreview: features },
+	});
+	const { featuresPreview } = watch();
+	const dispatch = useSettingsDispatch();
+
+	const handleSave = async () => {
+		try {
+			if (!allowFeaturePreviewSetting) {
+				throw Error(`AdminFeaturePreviewPage-handleSave-SettingNotFound`);
```
KevLehman, line 49: Maybe use a more human readable error?
(if this is translated somewhere, ignore me)

packages/ui-client/src/hooks/useFeaturePreviewList.ts
```diff
@@ -82,22 +87,46 @@ export const defaultFeaturesPreview: FeaturePreviewProps[] = [
...
+	const settingFeaturePreview = useMemo<FeaturePreviewProps[]>(() => {
+		if (typeof featurePreviewSettingJSON === 'string') {
+			return JSON.parse(featurePreviewSettingJSON);
+		}
+		return featurePreviewSettingJSON;
+	}, [featurePreviewSettingJSON]);
+
+	return useFeaturePreviewList(settingFeaturePreview);
+};
```
KevLehman, line 119: This is (with some smol changes) the same code as L94. Can we make this a func and call it in both places? (even if its a local func, no need for new files/helpers/etc)

apps/meteor/client/hooks/useFeaturePreviewEnableQuery.ts
lucas-a-pelegrino: I don't think this makes much of a difference, we are returning the item object either way, only applying  more logic in case enableQuery is present.

apps/meteor/client/views/admin/featurePreview/AdminFeaturePreviewPage.tsx
gabriellsh: This should not be an error, we should disable the submit button in this case. The setting should always exist, and if it is set to false we already disable the other toggles, so only possible change in this case is turning the setting on.

packages/ui-client/src/hooks/useFeaturePreviewList.ts
gabriellsh: Agreed. No need to make a helper as we don't want to encourage this type of patter. Just abstract to a local function.

apps/meteor/client/views/admin/featurePreview/AdminFeaturePreviewPage.tsx
KevLehman: That's even better!

apps/meteor/client/hooks/useFeaturePreviewEnableQuery.ts
KevLehman: Yep, this is not a change on semantics but on styling. 1 less level of nesting is always a good thing to do. The early return helps with it.

wdyt?

gabriellsh: This is done just to reduce the code kamehameha (reduces indentation thus facilitating reading).

In the end of the day it is the same thing, we either enforce this pattern as a rule or resist making this kind of review. Think of all time spent discussing basic indentation (or using switch vs if) that could be invested in more relevant improvements.

TL;DR: Early return better, time coding > time discussing indentation

lucas-a-pelegrino: I still believe this won't make much of a difference... IMHO this code although nested is still simple to read. 

If the early return was beneficial for performance, I would agree with you, tho.

KevLehman: We also have this on our guides: https://developer.rocket.chat/docs/code-style-guide

gabriellsh: Even better :). Now we don't have to document it.

gabriellsh (changes requested):
I just noticed, we need to test the new screen! Let me know if you need any help ;)

apps/meteor/client/hooks/useFeaturePreviewEnableQuery.ts
```diff
@@ -0,0 +1,27 @@
+import type { FeaturePreviewProps } from '@rocket.chat/ui-client';
+
+export const useFeaturePreviewEnableQuery = (features: FeaturePreviewProps[]) => {
```
ggazzo, line 3: if you are creating a hook it should have hooks inside, otherwise is only a regular function

packages/ui-client/src/hooks/useFeaturePreviewList.ts
```diff
@@ -82,22 +87,46 @@ export const defaultFeaturesPreview: FeaturePreviewProps[] = [
...
+	}
+	return setting;
+};
+
+export const usePreferenceFeaturePreviewList = () => {
 	const featurePreviewEnabled = useSetting<boolean>('Accounts_AllowFeaturePreview');
-	const userFeaturesPreview = useUserPreference<FeaturePreviewProps[]>('featuresPreview');
+	const userFeaturesPreviewPreference = useUserPreference<FeaturePreviewProps[]>('featuresPreview');
+	const userFeaturesPreview = useMemo<FeaturePreviewProps[]>(
+		() => parseSetting(userFeaturesPreviewPreference!),
```
ggazzo, line 102: its not 100% secure assume the setting will be always there

packages/i18n/src/locales/en.i18n.json
ggazzo, line 1129: I know probably someone asked you to improve those translations, but keep in mind this could be improved in a different pr, its not 100% related with the feature itself

apps/meteor/client/hooks/useFeaturePreviewEnableQuery.ts
```diff
@@ -0,0 +1,27 @@
...
+	return Object.entries(
```
ggazzo, line 4: ok an Array is an object but there is no reason for entries

packages/ui-client/src/hooks/useFeaturePreviewList.ts
```diff
@@ -82,22 +87,46 @@ export const defaultFeaturesPreview: FeaturePreviewProps[] = [
...
-export const useFeaturePreviewList = () => {
+// TODO: Remove this logic after we have a way to store object settings.
+const parseSetting = (setting: FeaturePreviewProps[] | string) => {
+	if (typeof setting === 'string') {
+		return JSON.parse(setting);
+	}
+	return setting;
+};
+
+export const usePreferenceFeaturePreviewList = () => {
```
ggazzo, line 98: one file one hook

apps/meteor/tests/e2e/admin-feature-preview.spec.ts
```diff
@@ -0,0 +1,40 @@
+import { Users } from './fixtures/userStates';
+import { test, expect } from './utils/test';
+
+test.use({ storageState: Users.admin.state });
+
+test.describe.serial('admin-feature-preview', () => {
```
ggazzo, line 6: each `test(...)` opens a new browser tab which takes time, are you sure this is really necessary? you dont need to create one test each assertion. maybe `test.step(...)`

apps/meteor/tests/e2e/admin-feature-preview.spec.ts
```diff
@@ -0,0 +1,40 @@
+import { Users } from './fixtures/userStates';
```
ggazzo, line 1: instead of creating an e2e test, I think creating few unit tests would be much more useful. also you didn test the whole point: if the default works for the users, you just tested if the page renders, which is expected already ;)

PR #33214, opened 2024-09-05
feat: Fetch avatars using user id

gabriellsh:

Created a new route `<workspace>/avatar/uid/<userID>` which allows fetching avatars using the user's ID instead of just using username.

This is being done to help improve performance in places where we only have access to user id. This way we avoid making an API call just to get the user's username.

This route will also return the generated SVG avatars (Initial letter with colored background) but does not allow using `@` to force returning this fallback, since a userID is expected and the fallback is based on username/name.

## Issue(s)

[CORE-673](https://rocketchat.atlassian.net/browse/CORE-673)


apps/meteor/server/routes/avatar/user.ts
```diff
@@ -0,0 +1,161 @@
...
+		return;
+	}
+
+	const avatarSize = getAvatarSizeFromRequest(req);
+
+	setCacheAndDispositionHeaders(req, res);
+
+	// if request starts with @ always return the svg letters
+	if (requestUsername[0] === '@') {
+		serveSvgAvatar(requestUsername, avatarSize, req.query.format, res);
```
pierre-lehnen-rc, line 75: You forgot to remove the `@` from the requestUsername here.

pierre-lehnen-rc (changes requested):

apps/meteor/server/routes/avatar/user.ts
gabriellsh: fixed. Nice catch

apps/meteor/server/routes/avatar/user.ts
```diff
@@ -0,0 +1,161 @@
...
+
+	// if request starts with @ always return the svg letters
+	if (requestUsername[0] === '@') {
+		serveSvgAvatar(requestUsername.slice(1), avatarSize, req.query.format, res);
+		return;
+	}
+
+	const file = await Avatars.findOneByName(requestUsername);
+	if (file) {
+		return serveAvatarFile(file, req, res, next);
```
sampaiodiego, line 81:
```suggestion
	serveAvatarFile(file, req, res, next);
	return;
```

apps/meteor/server/routes/avatar/user.ts
```diff
@@ -0,0 +1,161 @@
...
+		return;
+	}
+
+	const avatarSize = getAvatarSizeFromRequest(req);
+
+	setCacheAndDispositionHeaders(req, res);
+
+	const file = await Avatars.findOne({ userId: requestUserId });
+	if (file) {
+		return serveAvatarFile(file, req, res, next);
```
sampaiodiego, line 132:
```suggestion
	serveAvatarFile(file, req, res, next);
	return;
```

apps/meteor/server/routes/avatar/user.ts
```diff
@@ -0,0 +1,161 @@
...
+	}
+
+	const file = await Avatars.findOneByName(requestUsername);
+	if (file) {
+		return serveAvatarFile(file, req, res, next);
+	}
+
+	if (await handleExternalProvider(requestUsername, res)) {
+		return;
+	}
```
sampaiodiego, line 86: this function behaves different than the others.. I'd follow the other pattern, by checking the setting here and only calling the function if needed:

```suggestion
	if (settings.get('Accounts_AvatarExternalProviderUrl')) {
		await handleExternalProvider(requestUsername, res);
		return;
	}
```

apps/meteor/server/routes/avatar/user.ts
```diff
@@ -0,0 +1,161 @@
...
+	res.setHeader('Content-Length', file.size || 0);
+
+	return FileUpload.get(file, req, res, next);
+};
+
+const handleExternalProvider = async (username: string, res: ServerResponse): Promise<boolean> => {
+	const externalProviderUrl = settings.get<string>('Accounts_AvatarExternalProviderUrl');
+	if (!externalProviderUrl) {
+		return false;
+	}
```
sampaiodiego, line 48: see comment below
```suggestion
```

sampaiodiego (commented):
very small code style suggestions

sampaiodiego (approved):
I was wondering how we could add tests for avatar endpoints. any ideas?

apps/meteor/server/routes/avatar/middlewares/index.js
```diff
@@ -5,3 +5,4 @@ import { protectAvatars } from './auth';
 import './browserVersion';
 
 WebApp.connectHandlers.use('/avatar/', protectAvatars);
+WebApp.connectHandlers.use('/avatar/uid/', protectAvatars);
```
pierre-lehnen-rc, line 8: For users without permission to access the avatar, this will treat the uid as an username and render its first digit as an avatar. I don't know if that is an OK behavior?

gabriellsh: I made another function for this specific case. Thanks for pointing this out, I took the function name at face value.

apps/meteor/server/routes/avatar/middlewares/auth.js
```diff
@@ -1,22 +0,0 @@
-import { userCanAccessAvatar, renderSVGLetters } from '../utils';
```
KevLehman, line 1: If this change is not critical to the feat, i'd recommend moving this ts conversion to another PR

apps/meteor/server/routes/avatar/user.js
KevLehman, line 1: Same about this TS conversion

apps/meteor/server/routes/avatar/user.ts
```diff
@@ -0,0 +1,169 @@
...
+import type { IUpload, IIncomingMessage } from '@rocket.chat/core-typings';
+import { Avatars, Users } from '@rocket.chat/models';
+import { serverFetch as fetch } from '@rocket.chat/server-fetch';
+import type { NextFunction } from 'connect';
+
+import { FileUpload } from '../../../app/file-upload/server';
+import { settings } from '../../../app/settings/server';
+import { renderSVGLetters, serveSvgAvatarInRequestedFormat, wasFallbackModified, setCacheAndDispositionHeaders } from './utils';
+
+const MAX_USER_SVG_AVATAR_SIZE = 1024;
```
KevLehman, line 12: This same constants are in `/room.js` file, maybe we can put them in a place both files can access?

apps/meteor/server/routes/avatar/user.ts
```diff
@@ -0,0 +1,169 @@
...
+	res.setHeader('Content-Length', file.size || 0);
+
+	return FileUpload.get(file, req, res, next);
+};
+
+const handleExternalProvider = async (externalProviderUrl: string, username: string, res: ServerResponse): Promise<boolean> => {
+	const response = await fetch(externalProviderUrl.replace('{username}', username));
+	response.headers.forEach((value, key) => res.setHeader(key, value));
+	response.body.pipe(res);
+	return true;
```
KevLehman, line 47: The 2 places where this is called don't care about the return value, we can omit the return and make the func return void

apps/meteor/server/routes/avatar/user.ts
```diff
@@ -0,0 +1,169 @@
...
+		res.writeHead(404);
+		res.end();
+		return;
+	}
+
+	setCacheAndDispositionHeaders(req, res);
+
+	const externalProviderUrl = settings.get<string>('Accounts_AvatarExternalProviderUrl');
+	if (externalProviderUrl) {
+		const user = await Users.findOneById(requestUserId, { projection: { username: 1, name: 1 } });
```
KevLehman, line 127:
```suggestion
		const user = await Users.findOneById<Pick<...>>(requestUserId, { projection: { username: 1, name: 1 } });
```

apps/meteor/server/routes/avatar/user.ts
```diff
@@ -0,0 +1,169 @@
...
+			res.writeHead(404);
+			res.end();
+			return;
+		}
+
+		void handleExternalProvider(externalProviderUrl, user.username, res);
+		return;
+	}
+
+	const file = await Avatars.findOne({ userId: requestUserId });
```
KevLehman, line 139: Can we maybe create a `findOneByUserId` func inside the model?

apps/meteor/server/routes/avatar/user.ts
```diff
@@ -0,0 +1,169 @@
...
+	if (file) {
+		void serveAvatarFile(file, req, res, next);
+		return;
+	}
+
+	const user = await Users.findOneById(requestUserId, { projection: { username: 1, name: 1 } });
```
KevLehman, line 145:
```suggestion
	const user = await Users.findOneById<Pick<...>>(requestUserId, { projection: { username: 1, name: 1 } });
```

gabriellsh:
> I was wondering how we could add tests for avatar endpoints. any ideas?

I was very unsure about how to proceed with tests, but I think I can make some end-to-end cases (I found some tests for assets so I think they should be similar). I'll do them whenever I can, reverting this to draft for now.

apps/meteor/server/routes/avatar/user.ts
```diff
@@ -72,3 +73,62 @@ export const userAvatarByUsername = async function (request: IncomingMessage, re
 
 	serveSvgAvatarInRequestedFormat({ nameOrUsername: requestUsername, req, res });
 };
+
+export const userAvatarById = async function (request: IncomingMessage, res: ServerResponse, next: NextFunction) {
```
ggazzo, line 77: Why not create a 3rd function that receives user as an argument, and not duplicate the entire function?

gabriellsh: I can't share the functions because in some cases I don't need the user, and the logic for get the avatar by ID is slightly different. It would either be less efficient, or very convoluted.

pierre-lehnen-rc (approved):

apps/meteor/server/routes/avatar/user.spec.ts
```diff
@@ -37,6 +41,138 @@ const { userAvatarByUsername } = proxyquire.noCallThru().load('./user', {
...
+		mocks.utils.wasFallbackModified.returns(false);
+
+		await userAvatarById(request, response, next);
+
+		expect(mocks.utils.setCacheAndDispositionHeaders.calledWith(request, response)).to.be.true;
+		expect(response.writeHead.calledWith(304)).to.be.true;
+		expect(response.end.calledOnce).to.be.true;
+	});
+
+	it(`should write 404 if username is not found`, async () => {
```
MarcosSpessatto, line 133:
```suggestion
	it(`should write 404 if userId is not found`, async () => {
```

pierre-lehnen-rc (dismissed):

apps/meteor/server/models/raw/Avatars.ts
```diff
@@ -1,11 +1,15 @@
...
 
 import { BaseUploadModelRaw } from './BaseUploadModel';
 
 export class AvatarsRaw extends BaseUploadModelRaw implements IAvatarsModel {
 	constructor(db: Db, trash?: Collection<RocketChatRecordDeleted<IAvatar>>) {
 		super(db, 'avatars', trash);
 	}
+
+	findOneByUserId(userId: IUser['_id'], options?: FindOptions<IAvatar>) {
+		return this.findOne({ userId }, options);
```
sampaiodiego, line 13: `userId` is not indexed.

apps/meteor/server/routes/avatar/user.ts
```diff
@@ -72,3 +73,62 @@ export const userAvatarByUsername = async function (request: IncomingMessage, re
...
+		res.writeHead(404);
+		res.end();
+		return;
+	}
+
+	setCacheAndDispositionHeaders(req, res);
+
+	const externalProviderUrl = settings.get<string>('Accounts_AvatarExternalProviderUrl');
+	if (externalProviderUrl) {
+		const user = await Users.findOneById<Pick<IUser, 'name' | 'username'>>(requestUserId, { projection: { username: 1, name: 1 } });
```
sampaiodiego, line 96: `name` is not being used, right?

sampaiodiego (changes requested):

apps/meteor/server/routes/avatar/user.ts
gabriellsh: sorry bro

sampaiodiego (approved):

PR #33218, opened 2024-09-05
fix: File Upload description bypassing message maximum character limit setting

MartinSchoeler:

Added the message character limit restriction to the file upload description field, before this, you could send a message bigger than the limit by sending a file with a large description

## Issue(s)

[SUP-620](https://rocketchat.atlassian.net/browse/SUP-620)

[SUP-620]: https://rocketchat.atlassian.net/browse/SUP-620?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


abhinavkrin (changes requested):
Hello @MartinSchoeler,

Would it be a good idea to implement the check on the backend as well? Currently, users can send long messages through the API bypassing the character limit.

abhinavkrin (dismissed):

tiagoevanp (dismissed):

gabriellsh (changes requested):
The title should describe the issue and not the solution. I'll change that for you.

gabriellsh (dismissed):

apps/meteor/tests/e2e/file-upload.spec.ts
```diff
@@ -76,6 +76,7 @@ test.describe.serial('file-upload', () => {
 		await expect(poHomeChannel.content.btnModalConfirm).not.toBeVisible();
 	});
 });
+
```
MarcosSpessatto, line 79:
```suggestion
```

apps/meteor/app/api/server/v1/rooms.ts
MarcosSpessatto, line 1: I think we need to add the same validation for the `rooms.upload` endpoint, even if it's deprecated...

MarcosSpessatto (changes requested):
Can we add some UI

apps/meteor/tests/e2e/file-upload.spec.ts
MartinSchoeler: https://github.com/RocketChat/Rocket.Chat/pull/33218#discussion_r1811521745

MarcosSpessatto (approved):

dougfabris (approved):
Great work here!

PR #33229, opened 2024-09-07
fix: livechat stuck unread counter on new messages

ricardogarim:
This pull request addresses the issue described in [CORE-674](https://rocketchat.atlassian.net/browse/CORE-674), where the unread message counter does not update when a livechat visitor, with the widget minimized, receives their first response from an agent. The problem occurs because the `lastReadMessageId` is absent from the livechat client's state, and the `getUnreadMessages` function does not handle this situation correctly.

Shoutout to @ggazzo for helping address the changes correctly within the widget code!

[CORE-674]: https://rocketchat.atlassian.net/browse/CORE-674?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


MarcosSpessatto (approved):

packages/livechat/src/lib/main.js
```diff
@@ -107,24 +107,36 @@ export const getUnreadMessages = () => {
...
+	if (!shouldMarkAsUnread()) {
+		return;
 	}
+
+	const unreadMessages = getUnreadMessages();
+	if (!unreadMessages.length) {
+		return;
+	}
+
+	const lastReadMessage = getLastReadMessage();
```
ggazzo, line 119: I know that return early looks better, but doing this during a fix it may be not the ideal.
it chan prevent/conflict during backports

ggazzo (changes requested):
Missing tests ;)

abhinavkrin (approved):

ggazzo (approved):

MarcosSpessatto (approved):

PR #33231, opened 2024-09-09
fix(federation-v2): don't include protocol in homeserver domain setting

debdutdeb:


KevLehman (commented):
Can you add a couple of tests of what this is fixing?

PR #33237, opened 2024-09-09
fix: Retention Policy cached settings not updated during upgrade procedure

gabriellsh:

During the server version upgrade procedure, internal DB events are not run before the migration procedures run. This means that other services initialized before that, that depend on updates from the DB are unaware of changes during this time.

In this specific case, the settings cached collection, which was being used to configure the retention policy cron, would have outdated data at the time of running the cron job. This caused issues in workspaces that had the Retention Policy enabled before upgrading to 6.10.

This issue was introduced here https://github.com/RocketChat/Rocket.Chat/pull/32425
And this PR https://github.com/RocketChat/Rocket.Chat/pull/33025 was aimed at fixing it, without success (the issue didn't trigger during testing with the fix, despite being tested by multiple people in different environments. Later a better process for reproducing it was found).

The fix for it was to force the cached settings to be updated manually during the migration procedure. An appropriate documentation will be written regarding directives when handling updates that affect sensitive services/features.
## Issue(s)

[CORE-681](https://rocketchat.atlassian.net/browse/CORE-681)
## Further comments

[CORE-673]: https://rocketchat.atlassian.net/browse/CORE-673?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[CORE-681]: https://rocketchat.atlassian.net/browse/CORE-681?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ggazzo (approved):

ggazzo (changes requested):
changeset

MarcosSpessatto (approved):

ggazzo (approved):

PR #33241, opened 2024-09-10
feat!: apply restrictions to air gapped environments

MarcosSpessatto:

This PR introduces some restrictions when running an air-gapped environment

[AIR-5]
[AIR-6]
[AIR-10]
## Further comments

[AIR-6]: https://rocketchat.atlassian.net/browse/AIR-6
[AIR-10]: https://rocketchat.atlassian.net/browse/AIR-10
[AIR-5]: https://rocketchat.atlassian.net/browse/AIR-5?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/meteor/ee/app/license/server/settings.ts
```diff
@@ -20,4 +20,11 @@ await settingsRegistry.addGroup('Enterprise', async function () {
 			i18nLabel: 'Status',
 		});
 	});
+	await this.add('Cloud_Workspace_AirGapped_Restrictions_Remaining_Days', -1, {
+		type: 'int',
+		hidden: true,
```
gabriellsh, line 25: Hidden settings are only available in the server level (means they cannot be fetched by the client or through any endpoint)

In order for the frontend to be able tu use this we have to remove this property. We can find a way to hide it in the admin panel if that's necessary.

apps/meteor/ee/app/license/server/settings.ts
```diff
@@ -20,4 +20,11 @@ await settingsRegistry.addGroup('Enterprise', async function () {
...
+		readonly: true,
+		secret: true,
```
gabriellsh, line 27: This secret property is only used in the frontend to use password inputs instead of regular inputs. I don't think it is necessary aswell.

gabriellsh (changes requested):

apps/meteor/ee/app/license/server/airGappedRestrictionsCheck.spec.ts
```diff
@@ -0,0 +1,45 @@
+import { License, AirGappedRestriction } from '@rocket.chat/license';
+import { Statistics } from '@rocket.chat/models';
+
+import { checkAirGappedRestrictions } from './airGappedRestrictionsCheck';
+
+jest.mock('@rocket.chat/models', () => ({
```
gabriellsh, line 6: Jest is not working/available in this context

gabriellsh (dismissed):

gabriellsh:
I had to add the path for `airgappedRestrictionsCheck.spec.ts` in the mocharc.config file, since the tests were not working.

apps/meteor/ee/app/license/server/airGappedRestrictions.ts
```diff
@@ -0,0 +1,44 @@
...
+	}
+};
+
+AirGappedRestriction.on('remainingDays', async ({ days }: { days: number }) => {
+	await updateRestrictionSetting(days);
+	await sendRocketCatWarningToAdmins(days);
+});
+
+Meteor.startup(async () => {
+	await cronJobs.add('AirGapped Restrictions Cron', '0 */12 * * *', async () => {
```
sampaiodiego, line 41: instead of creating a cronjob for this purpose, I'd suggest we use the same one for statistics (`apps/meteor/server/cron/statistics.ts`) and already use the stats token in case it runs successfully.

sampaiodiego (changes requested):

sampaiodiego (changes requested):
since this is a breaking change, the target branch should be `release-7.0.0`

ee/packages/license/src/AirGappedRestriction.ts
```diff
@@ -0,0 +1,57 @@
...
+
+	public applyRestrictions(): void {
+		this.restricted = true;
+		this.emit('remainingDays', { days: 0 });
+	}
+
+	public removeRestrictions(days = -1): void {
+		this.restricted = false;
+		this.emit('remainingDays', { days });
+	}
```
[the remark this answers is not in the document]
gabriellsh: It shouldn't, the logic for those are in the `notifyRemainingDaysUntilRestriction` method. Your suggestion seems good. Alternatively we could also make those functions private and have a public function that receives the license and testes if the correct module is available (I think that's the only case these functions are used outside this class)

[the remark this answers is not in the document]
gabriellsh: It shouldn't but could*

There is external use but there shouldn't IMO. I just respected the pattern that was being used when changing the implementation.

ee/packages/license/src/AirGappedRestriction.ts
```diff
@@ -0,0 +1,57 @@
...
+import { decryptStatsToken } from './token';
+
+const DAY_IN_MS = 24 * 60 * 60 * 1000;
+const NO_ACTION_PERIOD_IN_DAYS = 3;
+const WARNING_PERIOD_IN_DAYS = 7;
+
+class AirGappedRestrictionClass extends EventEmitter {
+	private restricted = false;
+
+	public get isRestricted(): boolean {
```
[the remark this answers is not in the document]
gabriellsh: No real reason, I just thought to respect the class usual patterns. Is there a reason why a regular method would be preferred over a getter?

apps/meteor/ee/app/license/server/airGappedRestrictions.ts
gabriellsh: Only concern I have is if this should be outside the EE folder...

apps/meteor/ee/app/license/server/airGappedRestrictionsCheck.spec.ts
KevLehman: For server tests we use mocha + proxyquire. Not sure if we are moving towards using jest but for now i'd suggest to rewrite the tests using the tooling of the other server tests

.changeset/little-gifts-do.md
```diff
@@ -0,0 +1,7 @@
+---
+"@rocket.chat/meteor": minor
+"@rocket.chat/i18n": minor
+"@rocket.chat/license": minor
+---
+
+Introduced restrictions (connect to cloud or have a license) to air-gapped environments
```
KevLehman, line 7: We should use the new changeset guidelines, so the verb here should be on present tense

apps/meteor/app/lib/server/methods/sendMessage.ts
```diff
@@ -135,9 +136,9 @@ Meteor.methods<ServerMethods>({
 		}
 
 		try {
-			return await executeSendMessage(uid, message, previewUrls);
+			return await applyAirGappedRestrictionsValidation(() => executeSendMessage(uid, message, previewUrls));
 		} catch (error: any) {
-			if ((error.error || error.message) === 'error-not-allowed') {
+			if ((error.error || error.message) === 'error-not-allowed' || (error.error || error.message) === 'restricted-workspace') {
```
KevLehman, line 141: Should we change this for something like an includes?
```suggestion
			if (['error-not-allowed', 'restricted-workspace'].includes(error.error || error.message)) {
```

apps/meteor/ee/app/license/server/airGappedRestrictions.ts
```diff
@@ -0,0 +1,44 @@
...
+import { i18n } from '../../../../server/lib/i18n';
+import { sendMessagesToAdmins } from '../../../../server/lib/sendMessagesToAdmins';
+import { checkAirGappedRestrictions } from './airGappedRestrictionsCheck';
+
+const updateRestrictionSetting = async (remainingDays: number) => {
+	const value = settings.get('Cloud_Workspace_AirGapped_Restrictions_Remaining_Days');
+	if (value === remainingDays) {
+		return;
+	}
+	await Settings.updateValueById('Cloud_Workspace_AirGapped_Restrictions_Remaining_Days', remainingDays);
```
KevLehman, line 17: Maybe, we can create a new method that does the findOneAndUpdate kind of update and that way we avoid `notifyOnSetting...` to find the updated setting again.

apps/meteor/ee/app/license/server/airGappedRestrictionsCheck.ts
```diff
@@ -0,0 +1,16 @@
+import { License, AirGappedRestriction } from '@rocket.chat/license';
+import { Statistics } from '@rocket.chat/models';
+
+export async function checkAirGappedRestrictions(): Promise<void> {
+	if (License.hasModule('unlimited-presence')) {
+		AirGappedRestriction.removeRestrictions();
+		return;
+	}
+
+	const { statsToken: encryptedStatsToken } = (await Statistics.findLast()) || {};
```
KevLehman, line 10: Do we need _everything_ from the statistics record? Bet we should create a new model method that only returns the last statsToken. wdyt?

KevLehman (changes requested):

apps/meteor/ee/app/license/server/airGappedRestrictionsCheck.spec.ts
gabriellsh: This review was outdated. I already fixed it using proxyquire

ee/packages/license/src/AirGappedRestriction.ts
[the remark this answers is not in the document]
tassoevan: Getter methods should follow the same naming convention as fields. Boolean fields have no verbs, therefore `public get restricted()` is preferred but conflicts with the `restricted: boolean` field.

apps/meteor/ee/app/license/server/airGappedRestrictionsCheck.ts
gabriellsh: This file doesn't exist anymore. I moved the check to the statistics cron and getting the token from there.

apps/meteor/ee/app/license/server/airGappedRestrictions.ts
gabriellsh: I discussed this with @tassoevan yesterday, we concluded that this might not be the best place or time to do this. Specially since this is supposed to run at max twice a day, so shouldn't be a performance issue.

apps/meteor/ee/app/license/server/airGappedRestrictions.ts
sampaiodiego: oh, too late to answer, I saw @gabriellsh already moved the code.. but my reasoning was that since the default state is "restricted", it is fine using the same cronjob on CE code, because it is required to run otherwise it will be locked up.

gabriellsh: I already moved it @MarcosSpessatto. We can discuss this in DM if you want

ee/packages/license/src/AirGappedRestriction.ts
```diff
@@ -0,0 +1,73 @@
...
+
+import { License } from '.';
+import { decryptStatsToken } from './token';
+
+const DAY_IN_MS = 24 * 60 * 60 * 1000;
+const NO_ACTION_PERIOD_IN_DAYS = 3;
+const WARNING_PERIOD_IN_DAYS = 7;
+
+class AirGappedRestrictionClass extends EventEmitter {
+	#restricted = false;
```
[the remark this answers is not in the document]
gabriellsh: it was before, I changed it after discussing with Tasso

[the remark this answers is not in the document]
MarcosSpessatto: Any reason to follow this pattern?

apps/meteor/ee/app/license/server/airGappedRestrictions.ts
gabriellsh: Yes, there's also that concern. If the Statistics cron runs after the Restriction cron, we'd be always doing the checks with outdated data, possibly locking a workspace that shouldn't be locked @MarcosSpessatto

ee/packages/license/src/AirGappedRestriction.ts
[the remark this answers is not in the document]
gabriellsh: the `private` keyword is provided by TS, so this doesn't guarantee that the property is actually private since typechecking is usually done at compile time. There was also an issue with the naming convetion. See [tassoevan's comment above](https://github.com/RocketChat/Rocket.Chat/pull/33241#discussion_r1790521663).

But as in a Linkin Park song:
> In the end, it doesn't even matter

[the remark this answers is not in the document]
gabriellsh: I just went research it. For reference https://stackoverflow.com/questions/59641564/what-are-the-differences-between-the-private-keyword-and-private-fields-in-types

apps/meteor/app/statistics/server/index.ts
```diff
@@ -7,3 +7,4 @@ export { statistics } from './lib/statistics';
 export { getLastStatistics } from './functions/getLastStatistics';
 export { getStatistics } from './functions/getStatistics';
 export { updateCounter } from './functions/updateStatsCounter';
+export { sendUsageReport } from './functions/sendUsageReport';
```
sampaiodiego, line 10: we're avoiding using barrel exports.. we should add it to our docs
```suggestion
```

ee/packages/license/src/AirGappedRestriction.ts
sampaiodiego, line 11: should the initial state be `restricted = true` ? I think for the purposes we want to achieve it should be restricted by default.

we might also need a hook when adding a license so we can remove the restriction right away

ee/packages/license/src/index.ts
```diff
@@ -2,3 +2,4 @@ export { DuplicatedLicenseError } from './errors/DuplicatedLicenseError';
 export * from './licenseImp';
 export * from './MockedLicenseBuilder';
 export * from './applyLicense';
+export * from './AirGappedRestriction';
```
sampaiodiego, line 5: from what I could see there is no "init" action that needs to be done, so there is not need to import the class/instance here.
```suggestion
```

sampaiodiego (changes requested):

ee/packages/license/src/AirGappedRestriction.ts
gabriellsh: Really nice catch. This was not described in the Project whatsoever. I already asked for definitions on this. There's the possibility that new workspaces should start in the grace period, but that will require some changes, and most likely a discussion. An update to the ADR will also be necessary.

apps/meteor/app/statistics/server/index.ts
gabriellsh: Didn't know that beforehand. Done, thanks!

ee/packages/license/src/AirGappedRestriction.ts
gabriellsh: After some discussion, this will stay like this, workspaces should be restricted by default.

sampaiodiego (dismissed):

sampaiodiego (approved):

KevLehman (approved):

ggazzo (dismissed):

jeanfbrito (dismissed):

apps/meteor/server/models/raw/Statistics.ts
```diff
@@ -26,6 +26,22 @@ export class StatisticsRaw extends BaseRaw<IStats> implements IStatisticsModel {
 		return records?.[0];
 	}
 
+	async findLastStatsToken(): Promise<IStats['statsToken']> {
+		const records = await this.find(
```
KevLehman, line 30: Can we use a `findOne` here? it's technically the same, but this is what findOne is for 👀

ee/packages/license/src/AirGappedRestriction.ts
```diff
@@ -0,0 +1,73 @@
...
+
+import { License } from '.';
+import { decryptStatsToken } from './token';
+
+const DAY_IN_MS = 24 * 60 * 60 * 1000;
+const NO_ACTION_PERIOD_IN_DAYS = 3;
+const WARNING_PERIOD_IN_DAYS = 7;
+
+class AirGappedRestrictionClass extends EventEmitter {
+	#restricted = true;
```
KevLehman, line 11: nice

apps/meteor/server/models/raw/Statistics.ts
gabriellsh: I just copied the one above, do you think it is ok if I do this another time? I'll make sure to track it so I don't forget. Just so we can merge this asap

ee/packages/license/src/AirGappedRestriction.ts
gabriellsh: This a review or a compliment? lol

KevLehman: lol, just surprise, i think this is the first time i see someone using private fields on RC

apps/meteor/server/models/raw/Statistics.ts
KevLehman: No probs with that.

KevLehman (approved):

ee/packages/license/src/token.ts
```diff
@@ -3,13 +3,36 @@ import crypto from 'crypto';
 import type { ILicenseV3 } from '@rocket.chat/core-typings';
 import { verify, sign, getPairs } from '@rocket.chat/jwt';
 
+const convertV2KeyToV3 = (key: string): string => Buffer.from(key, 'base64').toString('utf-8');
```
ggazzo, line 6:
```suggestion
const base64Decode = (key: string): string => Buffer.from(key, 'base64').toString('utf-8');
```

ee/packages/license/src/token.ts
```diff
@@ -3,13 +3,36 @@ import crypto from 'crypto';
...
+const PUBLIC_STATS_KEY_V2 =
+	'LS0tLS1CRUdJTiBQVUJMSUMgS0VZLS0tLS0KTUlJQklqQU5CZ2txaGtpRzl3MEJBUUVGQUFPQ0FROEFNSUlCQ2dLQ0FRRUFuT0IvR3lCUXg1cVJZOC83dGxhTApkd0hSOWZBWHVzQ3ZBVU9YRjFPYjExaWx4ejdqY0pqWitaRjE2UTk3bjF3UDlvQnJMUDg5M3ZXUlc1bUtDdFpDClJQNTdJU1o2YjlHOXoyNStnV0NEa3ZZUUg1djJlMGoxUnE5TnNYMktlTWViOUZXenRQVFEvdDRvUW1SdUpiTEIKV013ajRRbHZ4OStpdWpkdGdQYmx5S0VFM0I3T1RPWk8xNzNOK2RDZW8wOWlQcStyKzBiRjNGSTRVcVFiZkFrdApOdGd0OUxVbjdlalNaQXNhSTZPbVBmOVVvSzM4NUdxWUxvbk1WbXBFbmZyRDlHSHl0OXFOY2pCTmZ1MUJLZWlUClBwVit6WE1FViszLzZ2bi84OW1id28zS0ZtTnRxVTJpRk5Yak5Cb2w1ZE82N3VlM0pURXBVQmJWRHZ2V3gwTk8KWndJREFRQUIKLS0tLS1FTkQgUFVCTElDIEtFWS0tLS0tCg==';
 
-const PUBLIC_LICENSE_KEY_V3 = Buffer.from(PUBLIC_LICENSE_KEY_V2, 'base64').toString('utf-8');
+const PUBLIC_LICENSE_KEY_V3 = convertV2KeyToV3(PUBLIC_LICENSE_KEY_V2);
```
ggazzo, line 13:
```suggestion
const PUBLIC_LICENSE_KEY_V3 = base64Decode(PUBLIC_LICENSE_KEY_V2);
```

ee/packages/license/src/token.ts
```diff
@@ -3,13 +3,36 @@ import crypto from 'crypto';
...
+			throw new Error('Missing PUBLIC_STATS_KEY_V3');
+		}
+
+		const [spki] = TEST_KEYS;
+
+		const [payload] = await verify(encrypted, spki);
+		return JSON.stringify(payload);
+	}
+
+	const [payload] = await verify(encrypted, convertV2KeyToV3(PUBLIC_STATS_KEY_V2));
```
ggazzo, line 31:
```suggestion
	const [payload] = await verify(encrypted, base64Decode(PUBLIC_STATS_KEY_V2));
```

packages/mock-providers/src/MockedAppRootBuilder.tsx
```diff
@@ -239,7 +239,7 @@ export class MockedAppRootBuilder {
 		return this;
 	}
 
-	withJohnDoe(): this {
+	withJohnDoe(overrides: Partial<IUser> = {}): this {
```
ggazzo, line 242: if you changes the name its not john anymore

ggazzo (approved):

sampaiodiego (approved):

tassoevan (approved):

KevLehman (approved):

PR #33247, opened 2024-09-10
chore: move GenericMenu to ui-client package

aleksandernsilva:
Moved the component `GenericMenu` to the `@rocket.chat/ui-client` package and adjusted imports accordingly. This change will allow this component to be used by packages outside `apps/meteor`.

### DIFF
--- a/packages/ui-client/src/components/GenericMenu/GenericMenu.tsx
+++ b/packages/ui-client/src/components/GenericMenu/GenericMenu.tsx
@@ -12,6 +12,7 @@ type GenericMenuCommonProps = {
 	disabled?: boolean;
 	callbackAction?: () => void;
 };
+
 type GenericMenuConditionalProps =
 	| {
 			sections?: {
@@ -28,7 +29,7 @@ type GenericMenuConditionalProps =
 
 type GenericMenuProps = GenericMenuCommonProps & GenericMenuConditionalProps & Omit<ComponentProps<typeof MenuV2>, 'children'>;
 
-export const GenericMenu = ({ title, icon = 'menu', disabled, onAction, callbackAction, ...props }: GenericMenuProps) => {
+const GenericMenu = ({ title, icon = 'menu', disabled, onAction, callbackAction, ...props }: GenericMenuProps) => {
 	const t = useTranslation();
 
 	const sections = 'sections' in props && props.sections;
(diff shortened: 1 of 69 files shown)


packages/ui-client/src/components/GenericMenu/GenericMenuItem.tsx
```diff
@@ -16,7 +15,7 @@ export type GenericMenuItemProps = {
 	variant?: string;
 };
 
-const GenericMenuItem = ({ icon, content, addon, status, gap, tooltip }: GenericMenuItemProps) => (
+export const GenericMenuItem = ({ icon, content, addon, status, gap, tooltip }: GenericMenuItemProps) => (
```
gabriellsh, line 18: Everything seems good, but since now we're using named exports, please remove the default export in the end of the file.

packages/ui-client/src/components/GenericMenu/index.ts
```diff
@@ -0,0 +1,4 @@
+export { default } from './GenericMenu';
```
gabriellsh, line 1: You also don't need to export default here anymore then.

packages/ui-client/src/components/GenericMenu/GenericMenu.tsx
```diff
@@ -29,7 +28,7 @@ type GenericMenuConditionalProps =
 
 type GenericMenuProps = GenericMenuCommonProps & GenericMenuConditionalProps & Omit<ComponentProps<typeof MenuV2>, 'children'>;
 
-const GenericMenu = ({ title, icon = 'menu', disabled, onAction, callbackAction, ...props }: GenericMenuProps) => {
+export const GenericMenu = ({ title, icon = 'menu', disabled, onAction, callbackAction, ...props }: GenericMenuProps) => {
```
gabriellsh, line 31: Gotta remove from this file too.

gabriellsh (changes requested):

packages/ui-client/src/components/GenericMenu/GenericMenuItem.tsx
aleksandernsilva: @gabriellsh My bad, the convention _[(Item 4)](https://developer.rocket.chat/docs/conventions#components)_ is to use default exports for component modules. I adjusted the files accordingly, leaving named exports only for index files.

gabriellsh: My bad, I forgot we could export default exports as named exports using `as`.

It's better this way, thank you!

gabriellsh (dismissed):

tassoevan (dismissed):

gabriellsh (approved):

PR #33253, opened 2024-09-11
fix(federation-v2): await bridge startup

debdutdeb:
## Further comments

[SUP-584](https://rocketchat.atlassian.net/browse/SUP-584)

CORE-683

[SUP-584]: https://rocketchat.atlassian.net/browse/SUP-584?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/meteor/ee/server/local-services/federation/service.ts
```diff
@@ -213,7 +213,7 @@ export class FederationServiceEE extends AbstractBaseFederationServiceEE impleme
 		return federationService;
 	}
 
-	async created(): Promise<void> {
+	async started(): Promise<void> {
 		return super.created();
```
debdutdeb, line 217:
```suggestion
		return super.started();
```

MarcosSpessatto (changes requested):
Can we add a changeset?

.changeset/green-queens-end.md
```diff
@@ -0,0 +1,5 @@
+---
+"@rocket.chat/meteor": patch
+---
+
+Fixes an issue with Federation startup where the bridge would intermittently fail to start causing errors being thrown "Matrix Bridge isn't running yet".
```
debdutdeb, line 5:
```suggestion
Fixes an issue with Federation startup where the bridge would intermittently fail to start causing error being shown "Matrix Bridge isn't running yet".
```

MarcosSpessatto (approved):

PR #33273, opened 2024-09-12
feat: add cursor pagination on chat.syncMessages

ricardogarim:
As per [CORE-686](https://rocketchat.atlassian.net/browse/CORE-686), this PR introduces cursor pagination to the messages/get Meteor method route, which supports the chat.syncMessages route. This improvement allows users to retrieve updates from a given date more efficiently, without compromising response times.

While cursor pagination is now available, it does not affect any existing behaviors. To use cursor pagination, instead of passing lastUpdate as query parameters, users will need to pass next or previous, along with count and type.

- `next` or `previous`: These represent the cursor pointers indicating whether the query should retrieve items from a later or earlier point in time. The cursor value must be the number of milliseconds, as it follows [Date getTime()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/getTime).
- `count`: Specifies the number of records to return.
- `type`: With the introduction of cursors, both updated and deleted messages cannot be retrieved in the same response. Users must specify whether they want to retrieve `DELETED` (for deleted messages) or `UPDATED` (for updated messages, which is the default use case).

Notes:
- In the future, it might be worth considering isolating this functionality into a separate route. However, for now, the current implementation ensures no conflicts or breaking changes, and the route’s signature remains unchanged.
- This implementation has been tested to ensure that existing users relying on `lastUpdate` will continue to have the same experience, with no changes required on their end.
- Keep in mind that while performance is acceptable for now, there may be opportunities to optimize database indexes further. Pagination has been a missing feature, so addressing it was a high priority.
- This change is especially beneficial for users who frequently work with large volumes of messages, such as when a username with a high message count in a room is changed, as cursor pagination offers a more efficient way to paginate through results.

[CORE-682]: https://rocketchat.atlassian.net/browse/CORE-682?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[CORE-686]: https://rocketchat.atlassian.net/browse/CORE-686?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,7 +49,7 @@ Free for 30 days. Afterward, choose between continuing to host on our secure clo
 
 You can follow these instructions to setup a dev environment:
 
-- Install **Node 14.x (LTS)** either [manually](https://nodejs.org/dist/latest-v14.x/) or using a tool like [nvm](https://github.com/creationix/nvm) or [volta](https://volta.sh/) (recommended)
+- Install **Node 20.x (LTS)** either [manually](https://nodejs.org/dist/latest-v20.x/) or using a tool like [nvm](https://github.com/creationix/nvm) or [volta](https://volta.sh/) (recommended)
 - Install **Meteor** ([version here](apps/meteor/.meteor/release)): https://docs.meteor.com/about/install.html
 - Install **yarn**: https://yarnpkg.com/getting-started/install
 - Install **Deno 1.x**: https://docs.deno.com/runtime/fundamentals/installation/


apps/meteor/app/api/server/v1/chat.ts
```diff
@@ -79,28 +79,42 @@ API.v1.addRoute(
 	{ authRequired: true },
 	{
 		async get() {
-			const { roomId, lastUpdate } = this.queryParams;
+			const { roomId, lastUpdate, count, next, previous, type } = this.queryParams;
```
MarcosSpessatto, line 82: Shouldn't we validate the type of these new parameters as well? AJV maybe? 🤔

apps/meteor/server/publications/messages.ts
```diff
@@ -7,65 +7,230 @@ import type { FindOptions } from 'mongodb';
...
-			rid,
-			latest: latestDate,
-			oldest: oldestDate,
-			inclusive,
-			count,
-			unreads,
-		});
+		const response = lastUpdate
+			? await handleWithoutPagination(rid, lastUpdate)
+			: await handleCursorPagination(type ?? 'UPDATED', rid, count, next, previous);
```
MarcosSpessatto, line 232: Why are we using Updated as default?

apps/meteor/server/publications/messages.ts
```diff
@@ -7,65 +7,230 @@ import type { FindOptions } from 'mongodb';
...
+	// if messages length is greater than count, it means we are not at the end of the messages
+	// and we shall return a cursor
+	if (count && messages.length > count) {
+		return mountCursorFromMessage(messages[messages.length - 2], type);
+	}
+
+	return null;
+}
+
+function popMessagesFromArray(messages: IMessage[], count?: number) {
```
MarcosSpessatto, line 132: I think we could pop the message directly there, since it's not being reused, we should not create this function.

apps/meteor/server/publications/messages.ts
```diff
@@ -7,65 +7,230 @@ import type { FindOptions } from 'mongodb';
...
 			deleted: IMessage[];
+			cursor: {
+				next: string | null;
+				previous: string | null;
+			};
 		}>;
 	}
 }
 
+function extractTimestampFromCursor(cursor: string): Date {
```
MarcosSpessatto, line 39: Let's please add unit tests for this function.

apps/meteor/server/publications/messages.ts
```diff
@@ -7,65 +7,230 @@ import type { FindOptions } from 'mongodb';
...
+	return new Date(parseInt(cursor, 10));
+}
+
+function mountCursorQuery({ next, previous, count }: { next?: string; previous?: string; count: number }): {
```
MarcosSpessatto, line 43: Let's please add unit tests for this function.

apps/meteor/server/publications/messages.ts
```diff
@@ -7,65 +7,230 @@ import type { FindOptions } from 'mongodb';
...
+	}
+
+	if (previous) {
+		return { query: { $lt: extractTimestampFromCursor(previous) }, options };
+	}
+
+	return { query: { $gt: new Date(0) }, options };
+}
+
+function mountCursorFromMessage(message: IMessage & { _deletedAt?: Date }, type: 'UPDATED' | 'DELETED'): string {
```
MarcosSpessatto, line 63: Let's please add unit tests for this function.

apps/meteor/server/publications/messages.ts
```diff
@@ -7,65 +7,230 @@ import type { FindOptions } from 'mongodb';
...
+	}
+
+	if (type === 'DELETED' && message._deletedAt) {
+		return `${message._deletedAt.getTime()}`;
+	}
+
+	throw new Meteor.Error('error-cursor-not-found', 'Cursor not found', { method: 'messages/get' });
+}
+
+function mountNextCursor(messages: IMessage[], type: CursorPaginationType, next?: string, previous?: string): string | null {
```
MarcosSpessatto, line 75: Let's please add unit tests for this function.

apps/meteor/server/publications/messages.ts
```diff
@@ -7,65 +7,230 @@ import type { FindOptions } from 'mongodb';
...
+	}
+
+	if (messages.length === 0 && previous) {
+		return `${parseInt(previous, 10) - 1}`;
+	}
+
+	return null;
+}
+
+function mountPreviousCursor(messages: IMessage[], type: CursorPaginationType, count?: number, next?: string): string | null {
```
MarcosSpessatto, line 98: Let's please add unit tests for this function.

apps/meteor/tests/end-to-end/api/chat.ts
```diff
@@ -2806,6 +2806,278 @@ describe('[Chat]', () => {
...
+				.expect('Content-Type', 'application/json')
+				.expect(400)
+				.expect((res) => {
+					expect(res.body).to.have.property('success', false);
+					expect(res.body.errorType).to.be.equal('error-roomId-param-not-provided');
+				})
+				.end(done);
+		});
+
+		it('should return an error when the neither "lastUpdate", "next" or "previous" parameter is sent', (done) => {
```
MarcosSpessatto, line 2832:
```suggestion
		it('should return an error when neither "lastUpdate", "next" or "previous" parameter is sent', (done) => {
```

packages/rest-typings/src/v1/chat.ts
MarcosSpessatto, line 1: Probably this is an old problem, but we are not using this validation on the endpoint 😶

MarcosSpessatto (changes requested):
Let's add a changeset please.

PR #33277, opened 2024-09-13
fix: TypeError in recent emoji list updates.

abhinavkrin:

This PR addresses the issue where the client threw a TypeError on updating the emoji list after receiving updates from the server. The issue occurred because recentList was not always an array. A check has been added to ensure recentList is an array before iterating over it.
image

## Further comments

[CORE-690](https://rocketchat.atlassian.net/browse/CORE-690)

[CORE-690]: https://rocketchat.atlassian.net/browse/CORE-690?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.changeset/late-suits-guess.md
+++ b/.changeset/late-suits-guess.md
@@ -1,0 +1,5 @@
+---
+'@rocket.chat/meteor': patch
+---
+
+Fixed issue where recentList was not always an array, causing a TypeError on emoji recent list updates
--- a/.changeset/late-suits-guess.md
+++ b/.changeset/late-suits-guess.md
@@ -2,4 +2,4 @@
 '@rocket.chat/meteor': patch
 ---
 
-Fixed issue where recentList was not always an array, causing a TypeError on emoji recent list updates
+Fixes an issue where the recent emoji list didn’t update properly, which caused errors on the client side.
--- a/apps/meteor/app/emoji/client/helpers.ts
+++ b/apps/meteor/app/emoji/client/helpers.ts
@@ -149,7 +149,10 @@ export const removeFromRecent = (emoji: string, recentEmojis: string[], setRecen
 	setRecentEmojis(recentEmojis);
 };
 
-export const updateRecent = (recentList: string[]) => {
+export const updateRecent = (recentList: string | string[]) => {
+	if (!Array.isArray(recentList)) {
+		recentList = [recentList];
+	}
 	const recentPkgList: string[] = emoji.packages.base.emojisByCategory.recent;
 	recentList?.forEach((_emoji) => {
 		!recentPkgList.includes(_emoji) && recentPkgList.push(_emoji);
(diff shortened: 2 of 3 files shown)


MarcosSpessatto (changes requested):
Is there any way to add a test?

Also, please add a changeset.

.changeset/late-suits-guess.md
```diff
@@ -0,0 +1,5 @@
+---
+'@rocket.chat/meteor': patch
+---
+
+Fixed issue where recentList was not always an array, causing a TypeError on emoji recent list updates
```
gabriellsh, line 5: As per [our docs](https://developer.rocket.chat/docs/development-workflow)

> A changeset is information about changes made in a branch or commit that will be present in the release's changelog.

We must avoid jargon when writing changesets. In my own words, they must be as simple as possible without ommiting information.

gabriellsh (changes requested):
Please update the title as per [Our docs](https://developer.rocket.chat/v1/docs/pull-requests-tags)

> The title should describe what's being fixed, not how it was fixed.

MarcosSpessatto (dismissed):

apps/meteor/app/emoji/client/helpers.ts
```diff
@@ -149,7 +149,10 @@ export const removeFromRecent = (emoji: string, recentEmojis: string[], setRecen
 	setRecentEmojis(recentEmojis);
 };
 
-export const updateRecent = (recentList: string[]) => {
+export const updateRecent = (recentList: string | string[]) => {
+	if (!Array.isArray(recentList)) {
```
ggazzo, line 153: I'm wondering how we pass string sometimes and ts doest not complain, would you mind do dig a bit more and find the root cause? this solves the consequence but i'm afraid does not solve the real isse

ggazzo (changes requested):

apps/meteor/app/emoji/client/helpers.ts
abhinavkrin: hey @ggazzo !
This refactor might not be needed anymore because only arrays are passed now due to recent refactorings.
But i have found out potential bugs.
Here we are calling updateRecent with hard coded string value 'rocket'. Instead, we should call updateRecent with the value of the emoji which was recently deleted. :
https://github.com/RocketChat/Rocket.Chat/blob/5fddfcc99ea6d17eb63cd33772dc7388a1d7fba8/apps/meteor/app/emoji-custom/client/lib/emojiCustom.ts#L52
same here:
https://github.com/RocketChat/Rocket.Chat/blob/5fddfcc99ea6d17eb63cd33772dc7388a1d7fba8/apps/meteor/app/emoji-custom/client/lib/emojiCustom.ts#L97

abhinavkrin: https://github.com/RocketChat/Rocket.Chat/pull/33808
This pr fixes the above mentioned issue.
Closing the current PR.

PR #33278, opened 2024-09-13
fix: no spaces between emoji aliases

abhinavkrin:

This PR fixes the issue where emoji aliases were not displayed properly in the custom emoji list. There were no spaces between the aliases. The change ensures that `emojis.aliases` is checked to be an array before displaying and joined with a comma. 

Screenshot 2024-09-13 at 12 47 39 PM

## Further comments

[CORE-691](https://rocketchat.atlassian.net/browse/CORE-691)

[CORE-691]: https://rocketchat.atlassian.net/browse/CORE-691?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


MarcosSpessatto (changes requested):
Can we add test here?

Also, lets add a changeset.

gabriellsh (changes requested):
Please update the title as per [Our docs](https://developer.rocket.chat/v1/docs/pull-requests-tags)

> The title should describe what's being fixed, not how it was fixed.

Here's a suggestion: `fix: no spaces between emoji aliases`

Also both in the task and PR, the issue is not well described, since it is presented that there is an issue but it does not convey what is happening. I had to look at the code to be able to discover what the actual issue was.

apps/meteor/client/views/admin/customEmoji/CustomEmoji.spec.tsx
```diff
@@ -0,0 +1,51 @@
...
+			name: 'smile',
+			aliases: ['happy', 'joy'],
+			extension: 'webp',
+			_updatedAt: new Date().toISOString(),
+			etag: 'abcdef',
+		},
+	],
+}));
+
+describe('CustomEmoji Component', () => {
```
MarcosSpessatto, line 26: Can we add a test case for when the aliases are not an array?

MarcosSpessatto (changes requested):

MarcosSpessatto (approved):

MarcosSpessatto (approved):

yash-rajpal (approved):
LGTM

PR #33283, opened 2024-09-13
feat: new E2EE composer hint

hugocostadev:

**Problem statement**

When the setting `Unencrypted messages` in encrypted rooms is turned off, there is a warning in the UI for the cases the user has inserted their E2EE password but doesn’t have the E2EE room key. However, when the setting is ON, there is no indicator in the UI making it clear they don’t have the room key.
As a result, the user may send unencrypted messages in an encrypted room believing it is E2EE encrypted, or may believe a bug prevents the messages from being encrypted.

**Solution**

Create a composer hint to make users aware they don’t have the room key.

## Further comments

https://rocketchat.atlassian.net/browse/E2EE2-63


aleksandernsilva (dismissed):

tassoevan (changes requested):

.changeset/many-files-turn.md
```diff
@@ -0,0 +1,7 @@
+---
+"@rocket.chat/meteor": minor
+"@rocket.chat/i18n": patch
+"@rocket.chat/ui-composer": minor
+---
+
+Add a warning to let users know they are about to send unencrypted messages in E2E Encrypted room if they have the setting `Unencrypted messages in encrypted rooms` enabled
```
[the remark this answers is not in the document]
hugocostadev: Thank you @tassoevan ! I took the chance to improve the grammar as well

tassoevan (approved):

apps/meteor/client/views/room/composer/messageBox/MessageBoxHint.tsx
```diff
@@ -0,0 +1,61 @@
...
+	};
+
+	return showComposerHint ? (
+		<MessageComposerHint
+			icon={isEditing ? 'pencil' : undefined}
+			helperText={isEditing && !isMobile ? <Trans i18nKey='Editing_message_hint' /> : undefined}
+		>
+			{renderHintText()}
+		</MessageComposerHint>
+	) : null;
```
ggazzo, line 58:
```suggestion
	if (!isEditing && !isUnencryptedHintVisible && !isReadOnly) {
		return null;
	}

	const renderHintText = (): string => {
		if (isEditing) {
			return t('Editing_message');
		}

		if (isReadOnly) {
			return t('This_room_is_read_only');
		}

		if (isUnencryptedHintVisible) {
			return t('E2EE_Composer_Unencrypted_Message');
		}

		return '';
	};

	
	return <MessageComposerHint
			icon={isEditing ? 'pencil' : undefined}
			helperText={isEditing && !isMobile ? <Trans i18nKey='Editing_message_hint' /> : undefined}
		>
		{renderHintText()}
	</MessageComposerHint>
```

packages/ui-composer/src/MessageComposer/MessageComposerHint.stories.tsx
```diff
@@ -0,0 +1,29 @@
...
+			Editing message
+		</MessageComposerHint>
+	);
+};
+
+export const ReadOnly: ComponentStory<typeof MessageComposerHint> = () => <MessageComposerHint>This room is read only</MessageComposerHint>;
+
+export const UnencryptedMessage: ComponentStory<typeof MessageComposerHint> = () => (
+	<MessageComposerHint>You're sending an unencrypted message</MessageComposerHint>
+);
```
ggazzo, line 29: I'm not sure if those variations are actually variations. all of them are the same, only with a different text.

the stories could be "hint, hint with helper text and hint with icon", but I dont see a lot of value adding a story for each text. thats the idea of the component, allow the application set any text inside

ggazzo (changes requested):
can you check my review? if you believe that doesnt make sense, please, just dismiss

packages/ui-composer/src/MessageComposer/MessageComposerHint.stories.tsx
hugocostadev: Yep, make sense, thanks for your view, I'll apply the necessary changes

apps/meteor/client/views/room/composer/messageBox/MessageBoxHint.tsx
hugocostadev: This earlier return looks cleaner, thanks @ggazzo 🚀

packages/ui-composer/src/MessageComposer/MessageComposerHint.stories.tsx
hugocostadev: What do you think about the three variations? 

image

ggazzo (dismissed):

tassoevan (dismissed):

ggazzo (approved):

PR #33284, opened 2024-09-13
feat: Scroll to bottom in a click.

Barrylimarti:

There is no way to go down to the bottom of the chat if you scrolled way too much in your channel. It becomes irritating sometimes to scroll way down to reach the bottom. Although, there were buttons that popped whenever there are new messages to see that message which practically brings you to the bottom, but that's not always the case.

Things that I added includes:-
1.  New Jump to bottom button, which pops at the screen whenever user scrolls and is not at bottom.
2. This button disappears automatically after 4 seconds (This can be changed according to requirement)
3. This won't be displayed in case user has new message as it has already been implemented.

In terms of code change:- 
1. I added a new component for Jump to bottom button.
2. Created a new hook useIsScroll, to know when scroll is triggered and when it is stopped. This is a reusable hook if needed for other purposes.
3. I also added the translations for the button with key "Jump_to_bottom" in the i18n package.

https://github.com/user-attachments/assets/995c8af1-2015-4f5b-987a-da3f7ac49e1b

## Issue(s)

This is a long pending [feature request](https://github.com/RocketChat/feature-requests/issues/804) 
## Steps to test or reproduce

This is a new feature
## Further comments

I couldn't find tests for these. Let me know if its there and it needs to be added.


Barrylimarti:
Hi! Can I get a review on this? If it needs changes do let know! Thanks.

PR #33287, opened 2024-09-15
fix: wrong user role showing in channel

### LINKED ISSUE #33162: Caching issue -- Problems when owner is removed from a team or channel and then added
### Description:

There's a caching issue when a user is removed and then re-added to a room. If the user was an owner, after he is added back it shows as if he was still an owner. If you refresh the screen you will see that he is no longer an owner (which is the correct behavior)

### Steps to reproduce:

1. Add a new admin user to a channel or team.
2. Remove the user.
3. Add the user back, When adding it back, the user is still shown as an owner.

### Expected behavior:

The user should not be shown as an owner.

### Actual behavior:

The user is shown as an owner

https://github.com/user-attachments/assets/29e53091-b510-4762-acec-09babbc3b916

Kishn0109:

this PR will fix the  #33162

 

## Issue(s)

Closes #33162 

## Steps to test or reproduce

1. Add a new admin user to a channel or team.
2. Remove the user.
3. Add the user back, When adding it back, the user is still shown as an owner.

## Further comments

I need clarification on the following:

image

- I’ve resolved the issue we discussed, and the solution is now working as expected. However, I wanted to get your input on optimizing the current implementation. Specifically, I’ve noticed that multiple API calls are being made, and I’m wondering if there are any strategies or best practices you would recommend to reduce the number of API calls without impacting functionality.

- The RoomRoles state is marked as deprecated. Can I safely remove it from the codebase?


Kishn0109:
This issue is already fixed

PR #33034, opened 2024-08-13
chore: promote roomUpdater to afterSaveMessage hook

ricardogarim:
This pull request addresses the requirements of [OPI-33](https://rocketchat.atlassian.net/browse/OPI-33) by promoting updater support to the `afterSaveMessage` hook - that aims to optimize performance by reducing database calls related to rooms and omnichannel rooms.

[OPI-33]: https://rocketchat.atlassian.net/browse/OPI-33

PR #33295, opened 2024-09-16
fix: correct parameter order in afterSaveMessage to restore outgoing webhooks and related features

ricardogarim:
After upgrading from Rocket.Chat version 6.9.0 to 6.12.0, outgoing integrations stopped functioning. The root cause was traced to a change in the parameter order passed by the `afterSendMessage` callback to the `sendMessage` event from the integrations app triggers. Specifically, the room parameter was not forwarded correctly, resulting in errors when searching for outgoing webhook triggers.

This issue was introduced by [PR #33034](https://github.com/RocketChat/Rocket.Chat/pull/33034), which modified the `afterSaveMessage` parameter signature but did not update the corresponding integration logic.

As outlined in [SUP-665](https://rocketchat.atlassian.net/browse/SUP-665), this malfunction blocked the propagation of certain actions due to the callback failing to forward the correct parameters.

This PR resolves the issue by correcting the parameter order, ensuring the room parameter is passed correctly, thus restoring the functionality of outgoing webhooks.

Additionally, the same issue affected the following areas, which have now been fixed:
- **IRC Bridge:** Fixed an issue where the service failed to bind Rocket.Chat users with their correct IRC nickname or channel path.
- **Autotranslate:** Resolved an issue preventing messages from being automatically translated.
- **EngagementDashboard:** Corrected the issue where new messages were not being counted in the engagement dashboard.

[SUP-665]: https://rocketchat.atlassian.net/browse/SUP-665?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/meteor/app/integrations/server/triggers.ts
+++ b/apps/meteor/app/integrations/server/triggers.ts
@@ -8,7 +8,12 @@ const callbackHandler = function _callbackHandler(eventType: string) {
 	};
 };
 
-callbacks.add('afterSaveMessage', callbackHandler('sendMessage'), callbacks.priority.LOW, 'integrations-sendMessage');
+callbacks.add(
+	'afterSaveMessage',
+	(message, { room }) => callbackHandler('sendMessage')(message, room),
+	callbacks.priority.LOW,
+	'integrations-sendMessage',
+);
 callbacks.add('afterCreateChannel', callbackHandler('roomCreated'), callbacks.priority.LOW, 'integrations-roomCreated');
 callbacks.add('afterCreatePrivateGroup', callbackHandler('roomCreated'), callbacks.priority.LOW, 'integrations-roomCreated');
 callbacks.add('afterCreateUser', callbackHandler('userCreated'), callbacks.priority.LOW, 'integrations-userCreated');
--- a/apps/meteor/app/irc/server/irc-bridge/index.js
+++ b/apps/meteor/app/irc/server/irc-bridge/index.js
@@ -209,7 +209,7 @@ class Bridge {
 		// Chatting
 		callbacks.add(
 			'afterSaveMessage',
-			(message, { room }) => this.onMessageReceived.bind(this, 'local', 'onSaveMessage')(message, room),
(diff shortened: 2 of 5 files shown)


apps/meteor/app/integrations/server/triggers.ts
```diff
@@ -8,7 +8,12 @@ const callbackHandler = function _callbackHandler(eventType: string) {
 	};
 };
 
-callbacks.add('afterSaveMessage', callbackHandler('sendMessage'), callbacks.priority.LOW, 'integrations-sendMessage');
+callbacks.add(
+	'afterSaveMessage',
+	(message, { room }) => callbackHandler('sendMessage')(message, room),
```
MarcosSpessatto, line 13: I'm not sure if this will fix the issue, since we changed the parameters of the callback here: https://github.com/RocketChat/Rocket.Chat/pull/33034/files#diff-deed993f8ad9d2963a4a2b6a22c0d629078a2a4895f1fcf7507d2c94901c56c7L53

The `sendMessage` handler is expecting 3 parameters and since we removed/replaced uid and roomUpdater, I'm not sure if it really fixes it.
Validation here: https://github.com/RocketChat/Rocket.Chat/blob/develop/apps/meteor/app/integrations/server/lib/triggerHandler.js#L131

apps/meteor/app/irc/server/irc-bridge/index.js
```diff
@@ -209,7 +209,7 @@ class Bridge {
 		// Chatting
 		callbacks.add(
 			'afterSaveMessage',
-			this.onMessageReceived.bind(this, 'local', 'onSaveMessage'),
+			(message, { room }) => this.onMessageReceived.bind(this, 'local', 'onSaveMessage')(message, room),
```
MarcosSpessatto, line 212: This method does not return a high-order function right? 🤔

MarcosSpessatto (changes requested):

MarcosSpessatto (approved):

PR #33308, opened 2024-09-17
feat: add contact channels

tapiarafael:

Ensure that when a new room is created, it creates a channel for the existing contact if there's none with the same `visitorId` and room type.
Also, include the destination for better visualization of the channel. (e.g whatsapp number, SMS number, widget host)

## Further comments

[SCI-74](https://rocketchat.atlassian.net/browse/SCI-74)

[SCI-74]: https://rocketchat.atlassian.net/browse/SCI-74?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/meteor/app/livechat/server/api/v1/room.ts
```diff
@@ -77,6 +77,7 @@ API.v1.addRoute(
 				const roomInfo = {
 					source: {
 						type: isWidget(this.request.headers) ? OmnichannelSourceType.WIDGET : OmnichannelSourceType.API,
+						destination: isWidget(this.request.headers) ? this.request.headers.host : undefined,
```
pierre-lehnen-rc, line 80: Maybe use the conditional only once and have separate objects for Widget and API?
```ts
				const roomInfo = {
					source: {
						...(isWidget(this.request.headers)
							? { type: OmnichannelSourceType.WIDGET, destination: this.request.headers.host }
							: { type: OmnichannelSourceType.API }),
					},
				};
```

apps/meteor/app/livechat/server/lib/LivechatTyped.ts
```diff
@@ -459,6 +460,31 @@ class LivechatClass {
 			extraData,
 		});
 
+		if (process.env.TEST_MODE?.toUpperCase() === 'TRUE') {
```
pierre-lehnen-rc, line 463: use the function from that other PR.

apps/meteor/app/livechat/imports/server/rest/sms.ts
```diff
@@ -128,6 +128,7 @@ API.v1.addRoute('livechat/sms-incoming/:service', {
 			source: {
 				type: OmnichannelSourceType.SMS,
 				alias: service,
+				destination: sms.to,
```
pierre-lehnen-rc, line 131: it doesn't look like that types are being enforced here (you can send anything as roomInfo to the createRoom function). Can you check if you can add the proper types to createRoom without having to change anything else? Otherwise just set the type of this roomInfo object.

apps/meteor/app/livechat/server/lib/LivechatTyped.ts
```diff
@@ -459,6 +460,31 @@ class LivechatClass {
...
+			const visitorContact = await LivechatVisitors.findOne<Pick<ILivechatVisitor, 'contactId'>>(visitor._id, {
```
pierre-lehnen-rc, line 464: In theory you shouldn't need to run a query here since this function already expects a full ILivechatVisitor, however I noticed two problems:
1. Calls from the apps-engine do not include the contactId, as the visitor converter is dropping this information (it needs to be fixed there)
2. The `findGuest` function that is used by some of the places that call this createRoom, is reading the visitor with a projection, loading only six attributes from the mongo document.

So I ask that you either:
1. Fix the places that call `createRoom` to send a full visitor object
2. Change the param type here to `AtLeast<ILivechatVisitor, ...>` specifying the attributes that are required by this function.

apps/meteor/tests/end-to-end/api/livechat/contacts.ts
```diff
@@ -632,4 +632,54 @@ describe('LIVECHAT - contacts', () => {
 			expect(res.body.errorType).to.be.equal('invalid-params');
 		});
 	});
+
+	describe('Contact Channels', () => {
+		let visitor: ILivechatVisitor;
+
+		beforeEach(async () => {
+			visitor = await createVisitor();
```
pierre-lehnen-rc, line 640: You're creating a visitor here on the `beforeEach` but also creating inside of each test directly;
Due to that, only one of the visitors created gets deleted in the afterEach.

apps/meteor/tests/end-to-end/api/livechat/contacts.ts
```diff
@@ -632,4 +632,54 @@ describe('LIVECHAT - contacts', () => {
...
+		});
+
+		afterEach(async () => {
+			await deleteVisitor(visitor.token);
+		});
+
+		it('should add a channel to a contact when creating a new room', async () => {
```
pierre-lehnen-rc, line 647: Any chance of having tests with other entry points from which a visitor can be created? (the SMS endpoint for example).

pierre-lehnen-rc (changes requested):

apps/meteor/app/livechat/server/lib/LivechatTyped.ts
tapiarafael: I added the `contactId` to the visitor entity in some places that I could find that call the `createRoom`, but I probably missed some places. I kept a verification to ensure that we will have the contact id.

apps/meteor/app/livechat/server/lib/LivechatTyped.ts
```diff
@@ -684,6 +714,8 @@ class LivechatClass {
 			returnDocument: 'after',
 		});
 
+		console.log({ upsertedLivechatVisitor });
+
```
pierre-lehnen-rc, line 718:
```suggestion
```

pierre-lehnen-rc (changes requested):

packages/apps-engine/src/definition/livechat/ILivechatRoom.ts
```diff
@@ -21,6 +21,8 @@ interface IOmnichannelSourceApp {
     label?: string;
     sidebarIcon?: string;
     defaultIcon?: string;
+    // The destination of the message (e.g widget host, email address, whatsapp number, etc)
+    destination?: string;
```
KevLehman, line 25: Don't we have this data already? The `label` prop for Whatsapp channels holds the phone number it was directed to. The email address is on an `email` prop, sms number is on an `sms` prop on the room. Any reason for adding another prop?

apps/meteor/app/livechat/server/lib/LivechatTyped.ts
```diff
@@ -459,6 +460,35 @@ class LivechatClass {
...
+		if (isSingleContactEnabled()) {
+			let { contactId } = visitor;
+			if (contactId) {
+				const visitorContact = await LivechatVisitors.findOne<Pick<ILivechatVisitor, 'contactId'>>(visitor._id, {
+					projection: { contactId: 1 },
+				});
+				contactId = visitorContact?.contactId;
+			}
+
+			const contact = await LivechatContacts.findOne({ _id: contactId });
```
KevLehman, line 472: I think models have a `findOneById` by default

KevLehman: Also, do you need the full `contact` object? Or can you go with just the _id & channels prop?

apps/meteor/app/livechat/server/lib/LivechatTyped.ts
```diff
@@ -459,6 +460,35 @@ class LivechatClass {
...
+				if (!channel) {
+					Livechat.logger.debug(`Adding channel for contact ${contact._id}`);
+
+					await LivechatContacts.addChannel(contact._id, {
+						name: roomInfo.source?.label || roomInfo.source?.type.toString() || 'other',
+						visitorId: visitor._id,
+						blocked: false,
+						verified: false,
+						details: roomInfo.source,
+						channelIdentifier: roomInfo.source?.destination,
```
KevLehman, line 486: In this case, when a `destination` is on roomInfo, you'll have the same property twice, one inside `details.destination` and one on `channelIdentifier`. Maybe we don't need that.

apps/meteor/app/livechat/server/lib/LivechatTyped.ts
```diff
@@ -459,6 +460,35 @@ class LivechatClass {
...
+			if (contact) {
+				const channel = contact.channels?.find((channel) => channel.name === roomInfo.source?.type && channel.visitorId === visitor._id);
+
+				if (!channel) {
+					Livechat.logger.debug(`Adding channel for contact ${contact._id}`);
+
+					await LivechatContacts.addChannel(contact._id, {
+						name: roomInfo.source?.label || roomInfo.source?.type.toString() || 'other',
+						visitorId: visitor._id,
+						blocked: false,
```
KevLehman, line 483: Pretty curious why a channel can be blocked instead of just blocking the contact on all channels 👀

KevLehman (changes requested):

apps/meteor/app/livechat/server/lib/LivechatTyped.ts
tapiarafael: It is by product design. Renato can explain it to you better, but the idea is that if your phone is stolen or your email is hacked, for example, the agent can block only the channel that is no longer trustworthy, and you can still contact the workspace by other means.

apps/meteor/app/livechat/server/lib/LivechatTyped.ts
MarcosSpessatto, line 472: I think we can use findOneById instead and also add a projection as Kevin already said

apps/meteor/app/livechat/server/lib/LivechatTyped.ts
```diff
@@ -684,6 +714,8 @@ class LivechatClass {
 			returnDocument: 'after',
 		});
 
+		console.log({ upsertedLivechatVisitor });
```
MarcosSpessatto, line 717:
```suggestion
```

MarcosSpessatto (changes requested):

MarcosSpessatto (dismissed):

apps/meteor/app/livechat/server/lib/LivechatTyped.ts
```diff
@@ -459,6 +462,38 @@ class LivechatClass {
 			extraData,
 		});
 
+		if (isSingleContactEnabled()) {
+			let { contactId } = visitor;
+			if (contactId) {
```
pierre-lehnen-rc, line 467:
```suggestion
			if (!contactId) {
```
I'm assuming you meant to load the contactId from the visitor if it is not yet loaded (though the function is supposed to be receiving a full visitor document). 
If I misunderstood it, please add a comment on the code explaining why it is being loaded.

packages/apps-engine/src/definition/livechat/ILivechatRoom.ts
pierre-lehnen-rc: It looks like you're not using this destination for anything yet, right? 
In that case we can leave it out of this PR and implement it separately if it adds value.

pierre-lehnen-rc (dismissed):

packages/apps-engine/src/definition/livechat/ILivechatRoom.ts
tapiarafael: We are, it is saving some information in the `livechat/room` endpoint, on the apps.engine bridge and also on the sms endpoint. Also this make available for the apps to save the info in a appropriated field

https://github.com/RocketChat/Rocket.Chat/pull/33308/files#diff-9e26aa53a3f9ad45396b8eb88faf7ee7ca9629d0c60544f27aadae68b2ec4017R80
https://github.com/RocketChat/Rocket.Chat/pull/33308/files#diff-5174fa8b7d018b6625c680bcf2e0ef6b319a9577b4856aca8df8073d508263d2R134
https://github.com/RocketChat/Rocket.Chat/pull/33308/files#diff-5b21e788e68edd23d55d2b1ba9c7610bd20a06ae63a9689bad84625c153604cfR121

apps/meteor/app/livechat/server/lib/LivechatTyped.ts
KevLehman: thanks!

apps/meteor/app/livechat/server/lib/LivechatTyped.ts
```diff
@@ -459,6 +462,38 @@ class LivechatClass {
...
 		});
 
+		if (isSingleContactEnabled()) {
+			let { contactId } = visitor;
+			if (!contactId) {
+				const visitorContact = await LivechatVisitors.findOne<Pick<ILivechatVisitor, 'contactId'>>(visitor._id, {
+					projection: { contactId: 1 },
+				});
+				contactId = visitorContact?.contactId;
+			}
```
KevLehman, line 472: I assume this is for covering the cases where visitor is called with less data, but i wonder if it wouldn't be better to ensure visitor has always the complete data and avoid this db call 🤔 wdyt?

apps/meteor/app/livechat/server/lib/LivechatTyped.ts
```diff
@@ -459,6 +462,38 @@ class LivechatClass {
...
+
+			const contact = await LivechatContacts.findOneById<Pick<ILivechatContact, '_id' | 'channels'>>(contactId as string, {
```
KevLehman, line 474: A visitor may not have a contactId (specially for visitors created before this call) and this will make it pass `undefined` to the findOneById, which iirc will return null, but at the same time is a db call we can save by validating contactId being an string before calling. wdyt?

apps/meteor/app/livechat/server/lib/LivechatTyped.ts
```diff
@@ -459,6 +462,38 @@ class LivechatClass {
...
+			if (contact) {
+				const channel = contact.channels?.find(
+					(channel: ILivechatContactChannel) => channel.name === roomInfo.source?.type && channel.visitorId === visitor._id,
+				);
+
+				if (!channel) {
+					Livechat.logger.debug(`Adding channel for contact ${contact._id}`);
+
+					await LivechatContacts.addChannel(contact._id, {
+						name: roomInfo.source?.label || roomInfo.source?.type.toString() || 'other',
```
KevLehman, line 487: We have an `OmnichannelSourceType.OTHER` entry already
```suggestion
						name: roomInfo.source?.label || roomInfo.source?.type.toString() || OmnichannelSourceType.OTHER,
```

KevLehman (changes requested):

apps/meteor/app/livechat/server/lib/LivechatTyped.ts
tapiarafael: Good catch! I added a method to create the contact if the visitor doesn't have one. We'll need it anyway, so it's an excellent way to ensure that we migrate old visitors by demand.

apps/meteor/app/livechat/server/lib/LivechatTyped.ts
tapiarafael: Pierre suggested the same. I've added the contactId on the projection of every place I could find to ensure that we're passing the parameters. I chose to leave this verification in case we miss some spots.

MarcosSpessatto (approved):

KevLehman (approved):
No more complains than the usual complains

PR #33324, opened 2024-09-19
chore: Single Contact ID: improved typings and removed some duplicated code

pierre-lehnen-rc:


sampaiodiego (approved):

apps/meteor/app/livechat/server/lib/Contacts.ts
```diff
@@ -281,14 +264,20 @@ export function validateCustomFields(allowedCustomFields: ILivechatCustomField[]
...
 		}
+
+		validValues[cf._id] = cfValue;
 	}
 
-	const allowedCustomFieldIds = new Set(allowedCustomFields.map((cf) => cf._id));
-	for (const key in customFields) {
-		if (!allowedCustomFieldIds.has(key)) {
-			throw new Error(i18n.t('error-custom-field-not-allowed', { key }));
+	if (!options?.ignoreAdditionalFields) {
```
KevLehman, line 271: Should we actually allow this? Not sure if there was some conversation but we never allowed more custom fields than the ones in the database.

pierre-lehnen-rc: Only known fields will be saved. The difference is if it will throw an error or just ignore the unknown fields when they are received. I kept this in because the old livechat code didn't throw exceptions but the new single contact code does (and I changed both to use the same code through this function).

tapiarafael (approved):

KevLehman (approved):

PR #33329, opened 2024-09-20
refactor!: Room's Key ID generation

KevLehman:

Our current way of generating the e2eRoomKeyId is flawed. Our current steps are:
- Generate a CryptoKey (AES)
- Export CryptoKey using `exportJWKKey`, this produces an object similar to this:
```
{
    "alg": "A128CBC",
    "ext": true,
    "k": "6RkYqlDAx_-xxxxxxxxxx",
    "key_ops": [
        "encrypt",
        "decrypt"
    ],
    "kty": "oct"
}
```
- We stringify the object
- We convert the stringified to base64
- Then, we take an slice of 12 chars.

Since after stringify/encoding the object above, the first 12 chars would be the string `{"alg": "A` which encoded is `eyJhbGciOiJB`, and then this is defined as the "room key id"

We currently don't use this value for anything that would require it to be random or "secure". We just use it to "identify" the key. However, since a room only has 1 key, it didn't matter if a key was actually "different", system would think it was "the key" for the room.

This would cause the client to try to decrypt the message and failing (if the key was actually different) as all the keys produced the same identifier.

This PR introduces a change in this keyID mechanism, creating an SHA-256 hash of the exported session key, and then taking the first 12 characters of the hash. Current rooms with current keys will continue to work as they should, as we now will use the `keyID` stored in the room object instead of calculating it from the key everytime.

## Issue(s)

https://rocketchat.atlassian.net/browse/E2EE2-68


apps/meteor/app/e2e/client/rocketchat.e2e.ts
```diff
@@ -144,52 +144,56 @@ class E2E extends Emitter {
...
 
-					if (!e2eRoom.isReady()) {
-						return;
-					}
+		await e2eRoom.decryptSubscription();
+	}
 
-					await e2eRoom.decryptSubscription();
-				}, 0);
+	debouncedOnSubscriptionChanged = _.debounce(this.onSubscriptionChanged.bind(this), 200);
```
KevLehman, line 189:
```suggestion
	debouncedOnSubscriptionChanged = _.debounce(this.onSubscriptionChanged.bind(this), 100);
```

julio-cfa (dismissed):

.changeset/spicy-eggs-march.md
```diff
@@ -0,0 +1,8 @@
+---
+"@rocket.chat/meteor": patch
```
hugocostadev, line 2:
```suggestion
"@rocket.chat/meteor": major
```

rodrigok (approved):

hugocostadev (approved):
Congrats @KevLehman every problem that we discussed in not related with this code.

PR #33334, opened 2024-09-23
fix: Guests that match an existing visitor data are generating an unnamed contact

pierre-lehnen-rc:
## Issue(s)

When a visitor document is updated with new data coming from a new request that does not include a name (for example: SMS), a new contact is being created without any name.


KevLehman (commented):
would you mind adding a test? 😬

tapiarafael (approved):

pierre-lehnen-rc:
Fixed by ##33594

Issue #32974, opened 2024-08-03
Incorrect Email Validation and  Persistent display of error message on message selection in MailExportForm


smritidoneria:
### Description:

 The `validateEmail` function does not correctly validate email addresses according to the RFC standard.  The error message for the `messagesCount` field is persistently displayed whenever a message is selected.

### Steps to reproduce:

1. Go to export messages
2. Write any invalid email. 
3. click on send

### Expected behavior:

If an invalid email is entered, it should display an error. Additionally, if the message is not selected first and the send button is clicked, it should display an error. However, if the user then selects the message, the error should not be displayed.

### Actual behavior:

https://github.com/user-attachments/assets/6a3ec43a-ede5-4f29-89d9-3d8ec40111c1

### Server Setup Information:

- Version of Rocket.Chat Server: 6.12.0
- License Type:  Enterprise
- Number of Users: 2
- Operating System: MacOs
- Deployment Method: Docker
- Number of Running Instances: 2
- DB Replicaset Oplog: Enabled 
- NodeJS Version:  14.21.4
- MongoDB Version: 7.0.5 

### Client Setup Information

- Desktop App or Browser Version: Safari Version 17.5
- Operating System: MacOs

smritidoneria:
If the issue is relevant, I want to work upon this

reetp:
Thanks for reporting.

> Server: 6.12.0

I don't think so - 6.11 hasn't been released yet?

Please confirm the version.

> function does not correctly validate email addresses

It looks like you mean it does not validate addresses at all?

smritidoneria:
Rocket chat version-6.10.2

yes, the validator do not seems to render the email address properly

reetp:
Thanks for your work.

I have requested that this be reviewed.

Please note that this may take some time as it will have to go in the queue with lots of other work.

casalsgh:
FYI - Community Liaison (reetp) shared this with internal teams and we'll be working on it soon (including review to existing proposed PR)

Starkz24:
I have solved this issue if this is not resolved till now please assign it to me.

reetp:
> I have solved this issue if this is not resolved till now please assign it to me.

No one assigns issue here.

You are free to work on whatever you want.

However, please see the previous team response.

This already has a PR and is is under review by the team so look at other issues - feature requests is a good place, or look at apps.

rique223:
[...]

PR #33337, opened 2024-09-23
fix: incorrect Email Validation and Persistent display of error message on message selection in MailExportForm

Shahmaz0:

In the Export Message dialog box
1. Made some changes in validateEmail function to validate the email according to RFC standards.
2. Also did some changes in regex in emailValidator and made it more efficient

## Issue(s)

fixed: #32974


Gustrb (changes requested):
Could you add some tests showing which emails were rejected that shouldn't??
I don't think the email regex matches exactly the RFC in any of the cases 🤔

Shahmaz0:
## Actual Behaviour: 

* This is the current emailValidation that we are seeing in our codebase.

https://github.com/user-attachments/assets/1bd8ccf6-6913-42d9-a796-288e45fa47ad

## After the changes that i did:

* This is the output after changes i did which more efficient and robust.


https://github.com/user-attachments/assets/12ff6a7e-2221-49d1-a273-5229094b5c26

Gustrb:
> ## After the changes that i did:
> * This is the output after changes i did, more efficient and robust.
> 
>  Fixed.mov

Wouldn't this happen just by virtue of passing the { style: 'rfc' }?
I am skeptical of the Regex change, that is why I asked for some test cases

Gustrb:
As per @rique223 explained in the issue, emails without TLD are not invalid per se, I don't think this PR makes a lot of sense

PR #32620, opened 2024-06-18
refactor: return created record from registerGuest

ricardogarim:
Related to [OPI-5](https://rocketchat.atlassian.net/browse/OPI-5), [PR #72 from Apps.WhatsApp](https://github.com/RocketChat/Apps.WhatsApp/pull/72), and [PR #771 from Rocket.Chat.Apps-engine](https://github.com/RocketChat/Rocket.Chat.Apps-engine/pull/771).

On the Apps.WhatsApp project, the process of handling new messages and incoming message endpoints has been streamlined. Specifically, there is no need to call `getLivechatVisitors` after `createVisitor` when a new message arrives. In `endpoints/WhatsAppMessage.post`, the `defineVisitor` function within `registerVisitor` has been modified to use `modify.getCreator().getLivechatCreator().createVisitor()` only. The call to `read.getLivechatReader().getLivechatVisitors()` has been removed since `createVisitor` now returns the inserted or updated record.

Also, changes were made here at Rocket.Chat project. The `createVisitor` method in the Livechat bridge has been updated. Within `LivechatTyped.registerGuest`, a new method on model `LivechatVisitors`, called `updateOneByIdOrToken` was added to handle both insert and update operations. This eliminates the need to run an insert followed by an update, optimizing the overall process.

## Performance comparison

| Use case | Avg Duration (ms) | Min Duration (ms) | Median Duration (ms) | Max Duration (ms) | 90th Percentile (ms) | 95th Percentile (ms) |
|---|---|---|---|---|---|---|
| New visitor BEFORE  | 426.73            | 247.68            | 398.33               | 945.99            | 567.71                | 727.53                |
| New visitor AFTER   | 381.29            | 205.85            | 360.51               | 761.91            | 531.72                | 565.77                |
| Same visitor BEFORE  | 178.33            | 84.09             | 163.78               | 555.61            | 257.91                | 286.1                 |
| Same visitor AFTER   | 164.74            | 83.6              | 146.3                | 446.9             | 223.83                | 252.66                |

<details>
<summary>K6 test results before changes</summary>

### Incoming messages from WhatsApp flow with **new visitor** on every request.
```text
          /\      |‾‾| /‾‾/   /‾‾/   
     /\  /  \     |  |/  /   /  /    
    /  \/    \    |     (   /   ‾‾\  
   /          \   |  |\  \ |  (‾)  | 
  / __________ \  |__| \__\ \_____/ .io

     execution: local
        script: src/incoming-message.js
        output: -

[...]

PR #33339, opened 2024-09-23
fix: Avoid destructuring `connectionData` when value is undefined

KevLehman:
## Issue(s)

SUP-672
## Further comments

#32620 introduced the bug, mainly for apps engine, which caused that when GDPR setting is enabled, new visitors cannot be created. This is because the server was trying to destructure `connectionData`, which was undefined.

The code on `LivechatTyped` was edited in a way that it was not checking `connectionData` to be an object before destructuring, causing the visitor creation to error out before completion.


.changeset/brave-brooms-invent.md
```diff
@@ -0,0 +1,5 @@
+---
+"@rocket.chat/meteor": patch
+---
+
+Fixed a problem that caused visitor creation to fail when GDPR setting was enabled and visitor was created via Apps Engine or the deprecated `livechat:registerGuest` method.
```
matheusbsilva137, line 5:
```suggestion
Fixes a problem that caused visitor creation to fail when GDPR setting was enabled and visitor was created via Apps Engine or the deprecated `livechat:registerGuest` method.
```
Not an issue, but I found out we changed our changeset best practices docs and we should now use the present tense. I got surprised by that, just found an opportunity to spread the word :kekw:

matheusbsilva137 (dismissed):

.changeset/brave-brooms-invent.md
KevLehman: Well

aleksandernsilva (approved):

PR #33343, opened 2024-09-24
ci: use node20 for release action

trivikr:

Uses node20 for release action

## Issue(s)
https://github.blog/changelog/2024-03-07-github-actions-all-actions-will-run-on-node20-instead-of-node16-by-default/

### DIFF
--- a/packages/release-action/action.yml
+++ b/packages/release-action/action.yml
@@ -10,7 +10,7 @@ inputs:
     required: false
 
 runs:
-  using: "node16"
+  using: "node20"
   main: "dist/index.js"
 
 branding:


Gustrb:
Hey, we are currently working on getting everything to work properly with node 20, I don't think just switching the release action would work 🤔

trivikr:
> I don't think just switching the release action would be interesting while not everything is properly migrated 🤔

GitHub is aiming to do it anyway as per their [blog post](https://github.blog/changelog/2024-03-07-github-actions-all-actions-will-run-on-node20-instead-of-node16-by-default/) linked in the description.
> We plan to migrate all actions to run on Node20 by Spring 2024.

If that's true, it's safer to migrate in the configuration before the action is taken in the backend.

ggazzo:
the chat is tied to node 14 but the actions that are not related with we dont have this limitation.

PR #33345, opened 2024-09-24
feat: Open messages menu by right clicking

xonx4l:
ISSUE-: #32387

This implementation offers -:

Improved accessibility: Users now have multiple ways to access the same functionality, catering to different preferences and habits.
Faster interaction: Right-clicking is often quicker than moving the cursor to a specific icon, especially on larger messages.
Familiarity: Many users are accustomed to right-click context menus in other applications, making this a natural and expected feature.
Flexibility: The menu appearing at the cursor position can be more convenient than a fixed location, especially on larger screens or with longer messages.


apps/meteor/client/components/message/variants/RoomMessage.tsx
```diff
@@ -62,6 +62,31 @@ const RoomMessage = ({
 
 	const messageRef = useJumpToMessage(message._id);
 
+
+	const handleRightClick = async (e: React.MouseEvent<HTMLDivElement>) => {
+        e.preventDefault();
+
```
Gustrb, line 68: this function should be indented properly

apps/meteor/client/components/message/variants/RoomMessage.tsx
```diff
@@ -62,6 +62,31 @@ const RoomMessage = ({
 
 	const messageRef = useJumpToMessage(message._id);
 
+
+	const handleRightClick = async (e: React.MouseEvent<HTMLDivElement>) => {
```
Gustrb, line 66: why is this async?

apps/meteor/client/components/message/variants/RoomMessage.tsx
```diff
@@ -62,6 +62,31 @@ const RoomMessage = ({
...
+
+       // Check if the click target is an interactive element
+       const target = e.target as HTMLElement;
+       const isInteractiveElement = target.closest('a, button, [role="button"]');
+       if (isInteractiveElement) {
+        // Allow default right-click behavior for interactive elements
+        return;
+}
+		
+       const moreButton = (e.currentTarget as HTMLElement).querySelector('[aria-label^="More options"]') as HTMLElement;
```
Gustrb, line 77: I don't think this approach is good because of a few reasons, first one:
- This would brake if we change the text of the aria-label
- If we decide for some reason to change ever so slightly the HTML this feature won't work

I think the better approach would be use it the same way we use when we click in the triple dots

Gustrb (changes requested):
We are not sure if this is going to be the approach we are taking (probably not), as explained by @reetp in #32387 

I am not really fond of the idea of doing this on the web app, might be an interesting idea in the electron app, but even then I am not sure.

ThatOneCalculator:
As someone who uses the Electron app daily (and is very used to other platforms like Discord doing this), I would *love* to see this in the electron app :)

xonx4l:
So as per all the suggestions I am closing this pr and  will look  into the electron app if I can do something there . Thanks

PR #33043, opened 2024-08-13
feat: search contacts endpoint

tapiarafael:
## Proposed changes (including videos or screenshots)

Implements a new endpoint to retrieve a list of paginated contacts, allowing sorting and filtering by name, email, or phone number.

[GET] `api/v1/omnichannel/contacts.search?searchText=john`
response example:
```json
{
	"contacts": [
		{
			"_id": "123",
			"name": "john doe",
			"emails": [
				"johndoe@example.com"
			],
			"phones": [
				"1-658-751-2622"
			],
			"contactManager": "321",
			"customFields": {},
			"unknown": false,
		},
	],
	"count": 25,
	"offset": 0,
	"total": 1,
	"success": true
}
```
Also, three indexes were created for the Contact collection to improve performance when querying for contacts. A normal index was created for `name` field, and partial indexes for `emails` and `phones` fields. 
<details>
<summary>Explain output showing that the indexes are being used</summary>

```
{
  "executionSuccess": true, 
  "nReturned": new NumberInt("1"), 
  "executionTimeMillis": new NumberInt("1"),
  "totalKeysExamined": new NumberInt("158"),
  "totalDocsExamined": new NumberInt("158"),
  "executionStages": {
    "stage": "SUBPLAN",
    "nReturned": new NumberInt("1"),
    "executionTimeMillisEstimate": new NumberInt("0"),
    "works": new NumberInt("161"),
    "advanced": new NumberInt("1"),
    "needTime": new NumberInt("159"),
    "needYield": new NumberInt("0"),
    "saveState": new NumberInt("0"),
    "restoreState": new NumberInt("0"),
    "isEOF": new NumberInt("1"),
    "inputStage": {
      "stage": "OR",
      "nReturned": new NumberInt("1"),
      "executionTimeMillisEstimate": new NumberInt("0"),
      "works": new NumberInt("161"),
      "advanced": new NumberInt("1"),
      "needTime": new NumberInt("159"),
      "needYield": new NumberInt("0"),
      "saveState": new NumberInt("0"),
      "restoreState": new NumberInt("0"),
      "isEOF": new NumberInt("1"),
      "dupsTested": new NumberInt("1"),
      "dupsDropped": new NumberInt("0"),
      "inputStages": [
        {
          "stage": "FETCH", 
          "filter": {"$or": [{"name": {"$regex": "Clint", "$options": "i"}}]}, 
          "nReturned": new NumberInt("1"),
          "executionTimeMillisEstimate": new NumberInt("0"),
[...]

PR #33349, opened 2024-09-24
feat: add history endpoint

tapiarafael:

Include a new endpoint to list the chat history of a contact. It does basically the same query as the current visitor history but for multiple visitors.

[GET] `api/v1/omnichannel/contacts.history?contactId=[id]`
response example
```json
{
	"history": [
			{
				"_id": "E7DxYy9HNX4xPqhyy",
				"v": {
					"_id": "66df0ccb6f2de823455bca97",
					"username": "guest-52",
					"token": "8e5ae65364b2190c228a49368896677674d77ffa7fb00014189d37d30b553c31",
					"status": "offline",
					"activity": [
						"2024-09"
					],
					"lastMessageTs": "2024-09-09T14:57:36.144Z"
				},
				"fname": "guest-52",
				"source": {
					"type": "widget"
				},
				"ts": "2024-09-09T14:57:31.164Z",
				"msgs": 8,
				"servedBy": {
					"_id": "jnuReT9zctgoNKz7t",
					"username": "johndoe",
					"ts": "2024-09-09T14:57:31.195Z"
				},
				"closedAt": "2024-09-09T14:57:40.250Z",
				"closedBy": {
					"_id": "66df0ccb6f2de823455bca97",
					"username": "guest-52"
				},
				"closer": "visitor"
			},
	],
	"count": 25,
	"offset": 0,
	"total": 1,
	"success": true
}
```

## Further comments

[SCI-82](https://rocketchat.atlassian.net/browse/SCI-82)

[SCI-82]: https://rocketchat.atlassian.net/browse/SCI-82?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/meteor/app/livechat/server/lib/Contacts.ts
```diff
@@ -248,6 +257,57 @@ export async function updateContact(params: UpdateContactParams): Promise<ILivec
...
+): Promise<PaginatedResult<{ history: VisitorSearchChatsResult[] }>> {
+	const { contactId, source, count, offset, sort } = params;
+
+	const contact = await LivechatContacts.findOneById<Pick<ILivechatContact, 'channels'>>(contactId, { projection: { channels: 1 } });
+
+	if (!contact) {
+		throw new Error('error-contact-not-found');
+	}
+
+	const visitorsIds = contact.channels?.map((channel) => channel.visitorId);
```
KevLehman, line 271: Make this a `set` so you send less data to mongo in the cases where visitorId is repeated

apps/meteor/server/models/raw/LivechatRooms.ts
```diff
@@ -1220,6 +1220,26 @@ export class LivechatRoomsRaw extends BaseRaw<IOmnichannelRoom> implements ILive
...
+		visitorsIds: string[];
+		source?: string;
+		options: FindOptions;
+	}): FindPaginated<FindCursor<IOmnichannelRoom>> {
+		return this.findPaginated<IOmnichannelRoom>(
+			{
+				'v._id': { $in: visitorsIds },
+				...(source && {
+					$or: [{ 'source.type': new RegExp(escapeRegExp(source), 'i') }, { 'source.alias': new RegExp(escapeRegExp(source), 'i') }],
+				}),
```
KevLehman, line 1237: Do we have a proper index for covering this query?

packages/i18n/src/locales/pt-BR.i18n.json
```diff
@@ -4223,7 +4223,7 @@
   "The_application_name_is_required": "O nome do aplicativo é obrigatório",
   "The_channel_name_is_required": "O nome do canal é obrigatório",
   "The_emails_are_being_sent": "Os e-mails estão sendo enviados.",
-  "The_empty_room__roomName__will_be_removed_automatically": "A sala vazia <span style=\"font-weight: bold;\">{{roomName}}</span> será removida automaticamente.",   
+  "The_empty_room__roomName__will_be_removed_automatically": "A sala vazia <span style=\"font-weight: bold;\">{{roomName}}</span> será removida automaticamente.",
```
KevLehman, line 4226: Is this change part of the PR?

packages/model-typings/src/models/ILivechatRoomsModel.ts
```diff
@@ -260,4 +260,9 @@ export interface ILivechatRoomsModel extends IBaseModel<IOmnichannelRoom> {
 	getVisitorActiveForPeriodUpdateQuery(period: string, updater?: Updater<IOmnichannelRoom>): Updater<IOmnichannelRoom>;
 	getMACStatisticsForPeriod(period: string): Promise<MACStats[]>;
 	getMACStatisticsBetweenDates(start: Date, end: Date): Promise<MACStats[]>;
+	findPaginatedRoomsByVisitorsIdsAndSource(params: {
+		visitorsIds: string[];
+		source?: string;
+		options: FindOptions;
```
KevLehman, line 266: TS will probably complain about these params, as options is not optional

KevLehman (changes requested):

apps/meteor/server/models/raw/LivechatRooms.ts
tapiarafael: We already have an index for `v_.id`, and since the `source` will not always be present, I believe that should be enough. Also, we already have more than 40 indexes for rooms. This could start affecting our writing performance, couldn't it?

KevLehman: Yup, which makes me think: it's a requirement to search by type/alias? The query should work pretty normal in normal conditions, but if the number of visitors is big (since a contact apparently don't have a limit on how many visitors can be added to them) and visitors can have unlimited rooms, then this could be a problem. 🤔

pierre-lehnen-rc: There's no limit but the number of visitors per contact is unlikely to ever be large. Even if a contact has a large number of rooms, this is a very specific query that can take a bit longer in that case without any issue. It's better than adding yet another index to the rooms collection for something that won't be relevant to most workspaces.

Also adding the filter by type/alias will just filter the rooms that have already been searched with an index, so it won't make it any slower.

apps/meteor/server/models/raw/LivechatRooms.ts
```diff
@@ -1220,6 +1220,26 @@ export class LivechatRoomsRaw extends BaseRaw<IOmnichannelRoom> implements ILive
...
 	}
 
+	findPaginatedRoomsByVisitorsIdsAndSource({
+		visitorsIds,
+		source,
+		options = {},
+	}: {
+		visitorsIds: string[];
+		source?: string;
+		options: FindOptions;
```
MarcosSpessatto, line 1230: Are options required here? 🤔 

If so, let's reverse the params order, `source` is not required.

packages/rest-typings/src/v1/omnichannel.ts
```diff
@@ -1355,6 +1355,36 @@ const GETOmnichannelContactsSearchSchema = {
...
 export const isGETOmnichannelContactsSearchProps = ajv.compile<GETOmnichannelContactsSearchProps>(GETOmnichannelContactsSearchSchema);
 
+type GETOmnichannelContactHistoryProps = PaginatedRequest<{
+	contactId: string;
+	source?: string;
+}>;
+
+const GETOmnichannelContactHistorySchema = {
+	type: 'object',
+	properties: {
```
MarcosSpessatto, line 1365: Shouldn't we mark all the not-required props as nullable true?

MarcosSpessatto (changes requested):
We need a changeset here. Or will this be added only when the whole feature is available?

tapiarafael:
> We need a changeset here. Or will this be added only when the whole feature is available?

We'll add it when the whole feature is available

packages/rest-typings/src/v1/omnichannel.ts
tapiarafael: I also thought, but Kevin explained [here](https://github.com/RocketChat/Rocket.Chat/pull/33043/files/9e3cbf8ea1e4afb1209fcc081dd4ec16f65cf478#r1767402358) that the `nullable: true` on the prop is only when we accept them as null and not necessary that they're not required props.

MarcosSpessatto (changes requested):
Forgot to ask before 😞, can we write some tests for the endpoint?

pierre-lehnen-rc (approved):

MarcosSpessatto (approved):

KevLehman (approved):

PR #33361, opened 2024-09-25
chore!: Update default E2EE settings behavior

yash-rajpal:

We introduced a new E2EE settings with 6.x release - `E2E_Allow_Unencrypted_Messages`. Earlier their default value was kept in a way to not change the default behavior.

We decided to change this default behavior in the major 7.0.0 release and now by default users shouldn't be allowed to send un-encrypted messages in encrypted rooms.
## Further comments

[E2EE2-69](https://rocketchat.atlassian.net/browse/E2EE2-69)

[E2EE2-69]: https://rocketchat.atlassian.net/browse/E2EE2-69?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/meteor/server/startup/migrations/v314.ts
+++ b/apps/meteor/server/startup/migrations/v314.ts
@@ -4,7 +4,7 @@ import { addMigration } from '../../lib/migrations';
 
 addMigration({
 	version: 314,
-	name: 'Update default behavior of E2EE settings, to not allow un-encrypted messages by default and allow files encryption by default.',
+	name: 'Update default behavior of E2E_Allow_Unencrypted_Messages setting, to not allow un-encrypted messages by default.',
 	async up() {
 		await Settings.updateOne(
 			{
@@ -14,17 +14,6 @@ addMigration({
 				$set: {
 					value: false,
 					packageValue: false,
-				},
-			},
-		);
-		await Settings.updateOne(
-			{
-				_id: 'E2E_Enable_Encrypt_Files',
-			},
-			{
-				$set: {
-					value: true,
-					packageValue: true,
 				},
 			},
 		);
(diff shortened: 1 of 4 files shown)


hugocostadev (changes requested):

MarcosSpessatto (dismissed):
The suggestion from @hugocostadev for the changeset is valid @yash-rajpal, I would recommend the same as he did. Can we rewrite that?

hugocostadev (approved):

apps/meteor/server/startup/migrations/v314.ts
```diff
@@ -0,0 +1,32 @@
...
+	version: 314,
+	name: 'Update default behavior of E2EE settings, to not allow un-encrypted messages by default and allow files encryption by default.',
+	async up() {
+		await Settings.updateOne(
+			{
+				_id: 'E2E_Allow_Unencrypted_Messages',
+			},
+			{
+				$set: {
+					value: false,
```
KevLehman, line 15: Why not changing just the packageValue? So if someone has it as `true` and likes it being `true`, we don't break his workflow

KevLehman (changes requested):
Did we change the setting creation code in another PR? The `this.add(...` cause otherwise the code for new workspaces will retain the current behavior as migrations won't run on clean envs

apps/meteor/server/startup/migrations/v314.ts
KevLehman: This should be said to the customer support team too, so if they receive comments from customers they know what to do 👀

KevLehman (approved):

MarcosSpessatto (approved):

PR #33378, opened 2024-09-26
fix: Async E2EE key exchange not working on develop

KevLehman:

This was due to missing notifiers on development. As these notifiers are not used in production, we didn't have this problem yet.

## Issue(s)

https://rocketchat.atlassian.net/browse/E2EE2-70


ggazzo:
this title 💩 

if its only in develop, it should be a regression, and if only happens in develop its not a real problem ;)

KevLehman:
Actually was on "development environments" because of Diego's changes not in prod yet lel

The changeset had the proper text, which is a win.

PR #33381, opened 2024-09-26
fix: race condition when forwarding livechat by splitting subscription removal

ricardogarim:
As noted in [CORE-696](https://rocketchat.atlassian.net/browse/CORE-696), when a livechat agent forwards a user to another department or agent, the user's chat can sometimes remain stuck in the sidebar of the old agent. Upon investigation, it was found that the old agent (the one forwarding the user) doesn't receive the expected socket event for the subscription deletion, which is crucial for clearing the user from their sidebar.

This issue was introduced in #32540 with the implementation of the `BaseRaw.findOneAndDelete` method, which is intended to remove a record from a collection and move it to the corresponding trash collection. However, due to the behavior of the RocketChat server's DB Watcher, between the deletion of the record and its insertion into the trash collection, the MongoDB change stream triggers a notification that initiates the DB Watcher flow. The DB Watcher then receives the subscription's trash ID and attempts to find it in the trash collection. At this point, however, the record hasn’t yet been inserted, resulting in a null response. Consequently, the notification of the subscription removal is lost, leaving the user's chat stuck in the old agent's sidebar.

This PR modifies the flow of the `findOneAndDelete` method to first insert the removed document into the trash collection before deleting it from the original collection. Additionally, a rollback condition has been implemented to revert the removal if the trash insertion fails.

## Further comments
Shoutout to @matheusbsilva137 and @sampaiodiego for their help in investigating and identifying the root cause of the problem. Your contributions were invaluable!

[CORE-696]: https://rocketchat.atlassian.net/browse/CORE-696?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


matheusbsilva137 (approved):

.changeset/dry-taxis-cry.md
```diff
@@ -0,0 +1,5 @@
+---
+'@rocket.chat/meteor': patch
+---
+
+Adjusts the findOneAndDelete method to insert into the trash collection before deletion to prevent a race condition that causes livechat users to get stuck in the agent's sidebar panel after being forwarded.
```
sampaiodiego, line 5: small wording suggestion to focus more on the issue being fixed other than the technical reason
```suggestion
Fixes a race condition that causes livechats to get stuck in the agent's sidebar panel after being forwarded.
```

KevLehman: Change `livechats` for `Livechat conversations` and we good

sampaiodiego (approved):

matheusbsilva137 (approved):

KevLehman (approved):

ggazzo (approved):

PR #33386, opened 2024-09-27
fix: Blank selected department when selecting out of range

rique223:

The Omnichannel current chats departments filter uses "on scroll" pagination (it fetches a list of the 25 first departments and loads more items every 25 items that the user scrolls down). If a user selects a department that isn’t within the first 25 departments listed, the selection will appear blank after refreshing the page. This happens because the `PaginatedSelectFiltered` component cannot find the selected department in the paginated options list due to it not being part of the initially fetched list of departments. This PR adds a "normalization" function to the listing of departments that will add the missing department to the initial list so the component can correctly show the previously selected department.

Before:
oc-departments-before

After:
oc-departments-after

## Further comments

Jira task: [SUP-675](https://rocketchat.atlassian.net/browse/SUP-675)

[SUP-675]: https://rocketchat.atlassian.net/browse/SUP-675?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


KevLehman (changes requested):
Can we add a couple of e2e tests?

rique223:
> Can we add a couple of e2e tests?

Hey @KevLehman, talking to the frontend folks we concluded that it would be better to test this specific behavior through unit tests of the `useDepartmentLists` hook so I've added the new unit tests and also added a new test to verify the localStorage usage of the departments filter.

KevLehman (approved):
I'll remove the omni review 🤗

KevLehman (dismissed):

apps/meteor/client/components/Omnichannel/hooks/useDepartmentsList.spec.ts
```diff
@@ -0,0 +1,107 @@
...
+					departments: initialDepartmentsListMock,
+				}))
+				.build(),
+		},
+	);
+
+	await waitFor(() => expect(result.current.itemsList.items).toContainEqual(selectedDepartmentMappedToOption));
+	// The expected length is 26 because the hook will add the 'All' item on run time
+	await waitFor(() => expect(result.current.itemsList.items.length).toBe(26));
+});
```
aleksandernsilva, line 55: I believe this test case could be improved by mocking the endpoint `/v1/livechat/department/:_id` and making sure it is not being called. This can be achieved with a combination of `withEndpoint` and `jest.fn()`. If you need help implementing, give me a heads up.

apps/meteor/client/components/Omnichannel/normalizeDepartments.ts
```diff
@@ -0,0 +1,20 @@
+import type { EndpointFunction } from '@rocket.chat/ui-contexts';
+
+import type { DepartmentListItem } from './hooks/useDepartmentsList';
+
+export const normalizeDepartments = async (
+	departments: DepartmentListItem[],
+	selectedDepartment: string,
+	getDepartment: EndpointFunction<'GET', '/v1/livechat/department/:_id'>,
+) => {
```
aleksandernsilva, line 9: I'd recommend adding a return type in this case just to ensure the item is being formatted correctly.

apps/meteor/client/components/Omnichannel/normalizeDepartments.ts
```diff
@@ -0,0 +1,20 @@
...
+	const isSelectedDepartmentAlreadyOnList = departments.find((department) => department._id === selectedDepartment);
```
aleksandernsilva, line 10: Two things. 
1. Using `.some()` is more appropriate for this scenario
2. `isSelectedDepartmentAlreadyOnList` could be a function to prevent searching the array unnecessarily.

apps/meteor/client/components/Omnichannel/normalizeDepartments.ts
aleksandernsilva, line 1: Create a `/utils` folder for this function

aleksandernsilva (changes requested):

apps/meteor/client/components/Omnichannel/hooks/useDepartmentsList.ts
```diff
@@ -14,9 +15,10 @@ type DepartmentsListOptions = {
 	excludeDepartmentId?: string;
 	enabled?: boolean;
 	showArchived?: boolean;
+	selectedDepartment?: string;
 };
 
-type DepartmentListItem = {
+export type DepartmentListItem = {
```
aleksandernsilva, line 21: I recommend moving this to a separate module to avoid the cyclic dependency with `normalizeDepartments`

apps/meteor/client/components/Omnichannel/utils/normalizeDepartments.ts
```diff
@@ -0,0 +1,20 @@
...
+	if (!selectedDepartment || selectedDepartment === 'all' || isSelectedDepartmentAlreadyOnList()) {
+		return departments;
+	}
+
+	const { department: missingDepartment } = await getDepartment({});
+
+	return missingDepartment
+		? [...departments, { _id: missingDepartment._id, label: missingDepartment.name, value: missingDepartment._id }]
+		: departments;
+};
```
aleksandernsilva, line 20: Just to be on the safe side, might be a good idea to add a try catch for this fetch and guarantee that departments is returned even if something goes wrong.

aleksandernsilva (changes requested):

aleksandernsilva (approved):

aleksandernsilva (approved):

PR #33392, opened 2024-09-27
feat: add-ons for apps

d-gubert:

Apply add-on validation to apps that require add-ons.

- [x] Prevent apps that require an add-on from being enabled if the add-on is not available
- [x] When the license is invalidated, disable apps that require add-on - this should already happen based on previous rules, but it did not happen when a license was removed, only when it became invalid.
- [ ] When the license is updated, validate changes to add-ons and update configuration accordingly


d-gubert:
Closing in favor of #33433

Issue #33354, opened 2024-09-14
Incorrect Email Validation on various pages


smritidoneria:
### Search before asking

- [X] I had searched in the [issues](https://github.com/RocketChat/Rocket.Chat.Electron/issues?q=is%3Aissue) and found no similar issues.

### Operating System

- [X] macOS
- [ ] Windows
- [ ] Linux

### Operating System Version

MacOs

### It happens on the web browser too?

Yes, it happens on the web browser too

### Rocket.Chat Desktop App Version

4.1.0

### Rocket.Chat Server Version

6.12.1

### Describe the bug

Email Validation occur at various places.  

1. while registering , the email has incorrect validation.
2. While exporting messages

Exporting Messages-

https://github.com/user-attachments/assets/ffc0b646-7b97-4a50-b42d-30ec1a9c359a

While registering-

https://github.com/user-attachments/assets/6a1b99c7-81e0-4123-b4ab-dd189d8d871f

### How to Reproduce

While registering-
Register yourself without the domain. The interface will lead to another page without showing any error.

Export message-
Click on export messages
Input the wrong email
click send

### Describe your Expected behavior

There should be proper email validation, not allowing user to login and export messages without the proper domain.

### Anything else

This problem occurs everytime.

### Are you willing to submit a code contribution?

- [X] Yes, I am willing to submit a Pull Request!

Tanu1201:
@jeanfbrito I was exploring the Rocket.chat and I noticed that this issue is valid issue, Can I take this.

reetp:
> jeanfbrito I was exploring the Rocket.chat and I noticed that this issue is an valid issue, so raised the PR for this please check this

You don't need to @ anyone - those who need to know will see.

You should check that a fix is required before starting work.
 
I think this is likely related to this and should be closed as not a bug:

https://github.com/RocketChat/Rocket.Chat/issues/32974#issuecomment-2382605184



>As per the comment here

https://github.com/RocketChat/Rocket.Chat/issues/32974#issuecomment-2369375668

>> Dotless or no TLD emails are valid emails according to the most recent RFC regarding emails (RFC 5322), this is not a bug.

> And here:

https://github.com/RocketChat/Rocket.Chat/pull/33012#discussion_r1773484524

>> Top-level domains (.com, .org, etc) are not mandatory for an email to be considered valid, there is no need to change this regex as this was a purposeful decision on email validation done by our engineering team.

[...]

PR #33398, opened 2024-09-29
fix: For issue #33354 (email regex validation)

Tanu1201:

In the apps/meteor/lib/emailValidator file, there were issues with how email validation was being handled. Previously, the validation function only checked the part of the email before the "@" symbol. As a result, it incorrectly accepted invalid email addresses like tanu16782@gmail, which is missing the domain suffix.

In this PR, I have updated the code to improve the validation. It now validates the entire email address, including the part after the "." to ensure proper formatting.

Additionally, I addressed issues with the rfcEmailRegex. Previously, it failed to handle certain edge cases, such as accepting empty strings . These cases have now been fixed as part of the update.

## Issue(s)

#33354 

## Steps to test or reproduce

Please refer to #33354 .

### DIFF
--- a/apps/meteor/lib/emailValidator.ts
+++ b/apps/meteor/lib/emailValidator.ts
@@ -1,13 +1,18 @@
 export const validateEmail = (email: string, options: { style: string } = { style: 'basic' }): boolean => {
-	const basicEmailRegex = /^[^@]+@[^@]+$/;
-	const rfcEmailRegex =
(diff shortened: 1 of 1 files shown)


reetp:
See

https://github.com/RocketChat/Rocket.Chat/pull/33012

https://github.com/RocketChat/Rocket.Chat/issues/33354#issuecomment-2382609422

Notbale:

> Dotless or no TLD emails are valid emails according to the most recent RFC regarding emails (RFC 5322), this is not a bug.

This should be closed thanks.

PR #33405, opened 2024-09-30
fix: Don't return users with `E2ESuggestedKey` set when selecting users to provide E2EKey

KevLehman:

This should prevent an issue where a user may receive an indefinite amount of "suggested keys" from all the users on a room even when a user already suggested a key for them, duplicating the work.

Now, the endpoint that returns users waiting for keys will return only users that don't have neither `E2EKey` nor `E2ESuggestedKey`.

## Issue(s)

https://rocketchat.atlassian.net/browse/E2EE2-72


sampaiodiego (commented):
we agreed to work on API tests in a further task

sampaiodiego (approved):
we agreed to work on API tests in a further task

PR #33410, opened 2024-10-01
fix: login service configuration notifier event

ricardogarim:
As per [OPI-39](https://rocketchat.atlassian.net/browse/OPI-39), we've identified scenarios where a "no broker set to broadcast" message appears, particularly in the "watch.loginServiceConfiguration" event. After reviewing the root cause, we've determined that it's unnecessary to stream changes in real-time through the socket when the server checks for updates to OAuth provider configurations.

[OPI-39]: https://rocketchat.atlassian.net/browse/OPI-39?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/meteor/app/lib/server/lib/notifyListener.ts
+++ b/apps/meteor/app/lib/server/lib/notifyListener.ts
@@ -162,6 +162,19 @@ export const notifyOnLoginServiceConfigurationChanged = withDbWatcherCheck(
 	},
 );
 
+export const notifyOnLoginServiceConfigurationChangedById = withDbWatcherCheck(
+	async <T extends ILoginServiceConfiguration>(_id: T['_id'], clientAction: ClientAction = 'updated'): Promise<void> => {
+		const item = await LoginServiceConfiguration.findOneById<Omit<LoginServiceConfigurationData, 'secret'>>(_id, {
+			projection: { secret: 0 },
+		});
+		if (!item) {
+			return;
+		}
+
+		void notifyOnLoginServiceConfigurationChanged(item, clientAction);
+	},
+);
+
 export const notifyOnLoginServiceConfigurationChangedByService = withDbWatcherCheck(
 	async <T extends ILoginServiceConfiguration>(service: T['service'], clientAction: ClientAction = 'updated'): Promise<void> => {
 		const item = await LoginServiceConfiguration.findOneByService<Omit<LoginServiceConfigurationData, 'secret'>>(service, {
--- a/apps/meteor/server/lib/oauth/updateOAuthServices.ts
+++ b/apps/meteor/server/lib/oauth/updateOAuthServices.ts
@@ -8,10 +8,6 @@ import type {
 import { LoginServiceConfiguration } from '@rocket.chat/models';
 
 import { CustomOAuth } from '../../../app/custom-oauth/server/custom_oauth_server';
-import {
-	notifyOnLoginServiceConfigurationChanged,
-	notifyOnLoginServiceConfigurationChangedByService,
-} from '../../../app/lib/server/lib/notifyListener';
 import { settings } from '../../../app/settings/server/cached';
 import { logger } from './logger';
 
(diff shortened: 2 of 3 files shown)


apps/meteor/server/lib/oauth/updateOAuthServices.ts
```diff
@@ -117,14 +113,10 @@ export async function updateOAuthServices(): Promise<void> {
 			}
 
 			await LoginServiceConfiguration.createOrUpdateService(serviceKey, data);
-			void notifyOnLoginServiceConfigurationChangedByService(serviceKey);
 		} else {
 			const service = await LoginServiceConfiguration.findOneByService(serviceName, { projection: { _id: 1 } });
 			if (service?._id) {
-				const { deletedCount } = await LoginServiceConfiguration.removeService(service._id);
-				if (deletedCount > 0) {
-					void notifyOnLoginServiceConfigurationChanged({ _id: service._id }, 'removed');
```
matheusbsilva137, line 126: I'm wondering: couldn't we just completely remove `notifyOnLoginServiceConfigurationChanged` (its definition)? I see it's called in only one other place where we also call `LoginServiceConfiguration.removeService` ([here](https://github.com/RocketChat/Rocket.Chat/blob/bcacbb1ceec16f9d6c3ee10cd29d0f9af1156779/apps/meteor/app/meteor-accounts-saml/server/lib/settings.ts#L138)), so we may not need it there too :thinking:

matheusbsilva137 (changes requested):
Please add a changeset to this PR

apps/meteor/server/lib/oauth/updateOAuthServices.ts
pierre-lehnen-rc: Removing this event probably breaks OAuth login, since the event is used to notify meteor that the login services have changed. 
We no longer rely on the client's loginServices collection to show the login buttons but AFAIK we still need them there for the login itself to work.

pierre-lehnen-rc: The mentioned error (`No broker set to broadcast`) also seems to happen for `watch.settings` sometimes, so it looks to me this is a problem related to #32540 - with some operations that used to rely on DbWatchers and now trigger notifications being executed before the broker is ready.

pierre-lehnen-rc (changes requested):
commented under matheus' previous comment.

apps/meteor/app/lib/server/lib/notifyListener.ts
```diff
@@ -162,6 +162,19 @@ export const notifyOnLoginServiceConfigurationChanged = withDbWatcherCheck(
 	},
 );
 
+export const notifyOnLoginServiceConfigurationChangedById = withDbWatcherCheck(
+	async <T extends ILoginServiceConfiguration>(_id: T['_id'], clientAction: ClientAction = 'updated'): Promise<void> => {
+		const item = await LoginServiceConfiguration.findOneById<Omit<LoginServiceConfigurationData, 'secret'>>(_id, {
+			projection: { secret: 0 },
+		});
+		if (!item) {
```
sampaiodiego, line 170: this check will prevent this function from working for `removed` items

apps/meteor/server/lib/refreshLoginServices.ts
```diff
@@ -1,11 +1,20 @@
...
 export async function refreshLoginServices(): Promise<void> {
 	await ServiceConfiguration.configurations.removeAsync({});
 
-	await Promise.allSettled([updateOAuthServices(), loadSamlServiceProviders(), updateCasServices()]);
+	const [oAuthServices] = await Promise.allSettled([updateOAuthServices(), loadSamlServiceProviders(), updateCasServices()]);
+
+	if (oAuthServices.status === 'fulfilled') {
+		await Promise.all(
+			oAuthServices.value.map((service) =>
+				notifyOnLoginServiceConfigurationChangedById(service._id, service.deleted ? 'removed' : 'inserted'),
```
sampaiodiego, line 16: previously we always sent `updated` instead of `inserted`.. not sure which one is the correct one though

sampaiodiego (changes requested):

PR #33412, opened 2024-10-01
fix: Updating path `responseBy` would create a conflict on `responseBy`

KevLehman:

- Removes unnecessary call to `getAgentLastMessageTsUpdateQuery`
- Changes the logic for applying the update on the `responseBy` object in a way that won't create a conflict.

## Issue(s)

https://rocketchat.atlassian.net/browse/CORE-703
## Steps to test or reproduce

- Kinda hard to reproduce. It just happened in candidate with a single whatsapp conversation after having 2 agents + visitor interacting on the same conversation.
- On local, you can force it to happen by sending a message on a conversation, then manually updating the database to set `waitingResponse: true` flag on the room object and send a message again. It should fail.

When these preconditions are met, Mongo will get an update similar to:
```
{
  $set: {
		'responseBy.lastMessageTs': ISODate(),
   		 responseBy: { lastMessageTs: ISODate() }
  },
}
```

Since the same update is modifying the same property on different ways, Mongo issues a "ConflictError" and blocks the update.

### DIFF
--- a/apps/meteor/app/livechat/server/hooks/markRoomResponded.ts
+++ b/apps/meteor/app/livechat/server/hooks/markRoomResponded.ts
@@ -33,10 +33,6 @@ export async function markRoomResponded(
 		}
 	}
 
-	if (room.responseBy) {
-		LivechatRooms.getAgentLastMessageTsUpdateQuery(roomUpdater);
-	}
-
 	if (!room.waitingResponse) {
 		// case where agent sends second message or any subsequent message in a room before visitor responds to the first message
 		// in this case, we just need to update the lastMessageTs of the responseBy object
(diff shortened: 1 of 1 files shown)


aleksandernsilva (approved):
Omnichannel

aleksandernsilva (changes requested):
@KevLehman If possible, adding a unit test that verifies this scenario might be a good idea.

apps/meteor/tests/unit/server/livechat/hooks/markRoomResponded.spec.ts
```diff
@@ -0,0 +1,125 @@
...
+	LivechatRooms: { getVisitorActiveForPeriodUpdateQuery: Sinon.stub(), getAgentLastMessageTsUpdateQuery: Sinon.stub(), getResponseByRoomIdUpdateQuery: Sinon.stub() },
+};
+
+const { markRoomResponded } = proxyquire.load('../../../../../app/livechat/server/hooks/markRoomResponded.ts', {
+	'../../../../lib/callbacks': { callbacks: { add: Sinon.stub(), priority: { HIGH: 'high' } } },
+	'../../../lib/server/lib/notifyListener': { notifyOnLivechatInquiryChanged: Sinon.stub() },
+	'@rocket.chat/models': models,
+});
+
+describe.only('markRoomResponded', () => {
```
KevLehman, line 18:
```suggestion
describe('markRoomResponded', () => {
```

aleksandernsilva (approved):

cuonghuunguyen:
We need this for v6, could you create a backport?

PR #33413, opened 2024-10-01
fix: missing pt-BR transation for non-member user mentions

abhinavkrin:

Fix missing translations for mention notifications in pt-BR locale when a user is mentioned but not part of the current channel.
Screenshot 2024-10-01 at 11 03 44 PM

## Further comments

[CORE-710](https://rocketchat.atlassian.net/browse/CORE-710)


packages/i18n/src/locales/pt-BR.i18n.json
```diff
@@ -4867,6 +4870,9 @@
   "You_have_not_verified_your_email": "Você ainda não verificou o seu e-mail.",
   "You_have_successfully_unsubscribed": "A partir de agora você não está mais cadastrado em nossa lista de e-mails.",
   "You_must_join_to_view_messages_in_this_channel": "Você deve participar para ver mensagens neste canal",
+  "You_mentioned___mentions__but_theyre_not_in_this_room": "Você mencionou {{mentions}}, mas eles não estão nesta sala.",
```
tiagoevanp, line 4873: Can you use [plurals](https://www.i18next.com/translation-function/plurals) here?

tiagoevanp (changes requested):

packages/i18n/src/locales/pt-BR.i18n.json
tiagoevanp: image
Here you are mentioning a single user, but the rest of the sentence is for more than a single one

packages/i18n/src/locales/pt-BR.i18n.json
```diff
@@ -1486,6 +1487,7 @@
   "Do_not_display_unread_counter": "Não exibir nenhum contador desse canal",
   "Do_not_provide_this_code_to_anyone": "Não forneça este código para ninguém.",
   "Do_Nothing": "Não fazer nada",
+  "Do_nothing": "Não fazer nada",
```
tiagoevanp, line 1490:
```suggestion
  "Do_nothing": "Fazer nada",
```

packages/i18n/src/locales/pt-BR.i18n.json
```diff
@@ -1486,6 +1487,7 @@
   "Do_not_display_unread_counter": "Não exibir nenhum contador desse canal",
   "Do_not_provide_this_code_to_anyone": "Não forneça este código para ninguém.",
   "Do_Nothing": "Não fazer nada",
+  "Do_nothing": "Fazer nada",
```
aleksandernsilva, line 1490: Guys, as wrong as it may seem I believe in pt-BR the correct way is to have a double negation. 
If still in doubt I recommend talking with one of our ux writers.

```suggestion
  "Do_nothing": "Não fazer nada",
```
_(btw not super happy having both `Do_nothing` and `Do_Nothing` with the exact same expression 😞 )_

aleksandernsilva (changes requested):

packages/i18n/src/locales/pt-BR.i18n.json
gabriellsh: In the english translation there is a case difference between the two:

```
  "Do_Nothing": "Do Nothing",
  "Do_nothing": "Do nothing",
```

So there is a reason for both existing but they should probably be slightly different

gabriellsh (dismissed):
I don't think this is a regression. Did anyone test this in older versions to make sure?

Also, the issue reported is not a missing translation. When there is a translation missing, it should always default to the english version (english translation is the base for all other translations, so it is the only one 100% guaranteed to have all keys present, always)

In the attached Jira task the title does indeed just mentions a missing translation, but in the `Expected behaviour` section there is a mention to how missing translations should be defaulted to english. I understand the confusion.

jessicaschelly:
> Did anyone test this in older versions to make sure?

Yes, on version 6.12.1, it shows the text in English (fallback for missing translation)
image

aleksandernsilva:
@gabriellsh Your points were addressed here #33440

Since the translations itself don't categorize as regression, I am changing this PR's prefix to `i18n` and milestone to `6.14`.

MarcosSpessatto (approved):

packages/i18n/src/locales/pt-BR.i18n.json
```diff
@@ -1494,6 +1495,7 @@
   "Do_not_display_unread_counter": "Não exibir nenhum contador desse canal",
   "Do_not_provide_this_code_to_anyone": "Não forneça este código para ninguém.",
   "Do_Nothing": "Não fazer nada",
+  "Do_nothing": "Não fazer nada",
```
ggazzo, line 1498: why 2 keys with same value and almost same key?

PR #33399, opened 2024-09-30
chore!: Update private apps cap on Community Edition

matheusbsilva137:
## Proposed changes (including videos or screenshots)

- Update private apps cap on CE from 3 to 0;
- Add end-to-end tests for private apps upload on CE and EE.

## Steps to test or reproduce

Enabling new private apps should not be allowed on CE starting on 7.0. They can be installed, but their status should be "Disabled" by default, and enabling them should not be available to CE.
While on Premium workspaces (including the Starter plan), no behavior should be changed: new private apps can be installed and their status should be "Enabled" by default as long as they're under the plan's max private apps limitation.

## Further comments

[PASLR-2](https://rocketchat.atlassian.net/browse/PASLR-2)

[PASLR-2]: https://rocketchat.atlassian.net/browse/PASLR-2?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #33417, opened 2024-10-02
fix!: Private apps are always auto enabled when updated

matheusbsilva137:

 - Stop auto-enabling private apps when they are updated. Instead, only try to startup the updated app (and consequently, auto-enable it) when it was already in an enabled status before the update.

## Steps to test or reproduce

Go to **Marketplace > Private apps** and try updating a disabled private app in CE (eg by uploading its file once again).
**Current behavior:** the updated app will be enabled after the update, regardless of any condition.
**Expected behavior:** the app should only be enabled if it was already enabled before the update (so that license restriction are also followed accordingly)

## Further comments

[PASLR-25](https://rocketchat.atlassian.net/browse/PASLR-25)

[PASLR-25]: https://rocketchat.atlassian.net/browse/PASLR-25?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/apps-engine/src/server/AppManager.ts
+++ b/packages/apps-engine/src/server/AppManager.ts
@@ -720,7 +720,8 @@ export class AppManager {
         aff.setApp(app);
 
         if (updateOptions.loadApp) {
-            await this.updateLocal(stored, app);
+            const shouldEnableApp = AppStatusUtils.isEnabled(descriptor.status);
+            await this.updateLocal(stored, app, shouldEnableApp);
 
             await this.bridges
                 .getAppActivationBridge()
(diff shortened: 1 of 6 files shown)


matheusbsilva137:
Waiting for #33399 to be merged to open this PR (so that E2E tests work properly)

packages/apps-engine/src/server/AppManager.ts
```diff
@@ -742,7 +743,7 @@ export class AppManager {
      * With an instance of a ProxiedApp, start it up and replace
      * the reference in the local app collection
      */
-    public async updateLocal(stored: IAppStorageItem, appPackageOrInstance: ProxiedApp | Buffer) {
+    public async updateLocal(stored: IAppStorageItem, appPackageOrInstance: ProxiedApp | Buffer, enable: boolean) {
```
tassoevan, line 746: Instead of introducing a boolean parameter, what about creating two methods? https://understandlegacycode.com/blog/what-is-wrong-with-boolean-parameters/

tassoevan (changes requested):

tassoevan (approved):

PR #33419, opened 2024-10-02
chore!: e2ee mention setting update

hugocostadev:

Change the default value of `E2E_Enabled_Mentions` to true to always accept mentions in E2EE messages

## Further comments

https://rocketchat.atlassian.net/browse/E2EE2-71


tassoevan (approved):

KevLehman (dismissed):
Confirmed with Hugo via DM that this change gets propagated to the cached collection accordingly, 💯

ggazzo (approved):

PR #33420, opened 2024-10-02
chore!: Tag private apps as migrated (exempt from CE limitations) when moving to the next major

matheusbsilva137:

 - Tag **all** private apps from CE workspaces as `migrated` when moving to 7.0 so that they're exempt from the new CE limitations;

## Further comments

[PASLR-3](https://rocketchat.atlassian.net/browse/PASLR-3)

[PASLR-3]: https://rocketchat.atlassian.net/browse/PASLR-3?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/meteor/server/models/raw/AppLogsModel.ts
+++ b/apps/meteor/server/models/raw/AppLogsModel.ts
@@ -13,7 +13,9 @@ export class AppsLogsModel extends BaseRaw<any> implements IAppLogsModel {
 	}
 
 	async resetTTLIndex(expireAfterSeconds: number): Promise<void> {
-		await this.col.dropIndex('_updatedAt_1');
+		if (await this.col.indexExists('_updatedAt_1')) {
+			await this.col.dropIndex('_updatedAt_1');
+		}
 		await this.col.createIndex({ _updatedAt: 1 }, { expireAfterSeconds });
 	}
 }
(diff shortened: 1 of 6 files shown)


apps/meteor/server/models/raw/AppLogsModel.ts
```diff
@@ -13,7 +13,9 @@ export class AppsLogsModel extends BaseRaw<any> implements IAppLogsModel {
 	}
 
 	async resetTTLIndex(expireAfterSeconds: number): Promise<void> {
-		await this.col.dropIndex('_updatedAt_1');
+		if (await this.col.indexExists('_updatedAt_1')) {
```
matheusbsilva137, line 16: Running the new migration would trigger an error here due to the attempt of removing an unexistent `_updatedAt_1` index in AppsLogs, so I also needed to check whether this index exists to avoid throwing this error :man_shrugging:

tapiarafael (approved):

apps/meteor/server/startup/migrations/v307.ts
```diff
@@ -0,0 +1,40 @@
...
+import { License } from '@rocket.chat/license';
+
+import type { AppRealStorage } from '../../../ee/server/apps/storage';
+import { addMigration } from '../../lib/migrations';
+
+addMigration({
+	version: 307,
+	name: "Mark all installed private apps as 'migrated'",
+	async up() {
+		const isEE = License.hasValidLicense();
```
pierre-lehnen-rc, line 13: Shouldn't we flag them on EE workspaces as well, in case any of them switch to CE in the future?

matheusbsilva137: We don't need to because we're now handling this in the `onRemoveLicense` and `onInvalidateLicense` callbacks (check #33439). Private apps will always be flagged as migrated on downgrade starting on 7.0

tassoevan (approved):

PR #33422, opened 2024-10-03
feat: Add support for Omnichannel Contacts on the Importer system

pierre-lehnen-rc:

This PR is one of two parts for implementing Contact Import. What this one does is add "Contact" as one of the possible data types that can be managed on the importer classes. It does not implement any logic for actually converting imported data into Contacts. Such logic will be introduced in a separate PR.

## Issue(s)

[SCI-43](https://rocketchat.atlassian.net/browse/SCI-43)

## Further comments

[SCI-43]: https://rocketchat.atlassian.net/browse/SCI-43?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/meteor/app/importer-csv/server/addParsedContacts.ts
```diff
@@ -0,0 +1,36 @@
+import { Random } from '@rocket.chat/random';
+
+import type { ImportDataConverter } from '../../importer/server/classes/ImportDataConverter';
+
+export async function addParsedContacts(this: ImportDataConverter, parsedContacts: string[][]): Promise<number> {
```
MarcosSpessatto, line 5: It would be better to move this as the importer class method instead of providing the `this` for the function.

apps/meteor/app/importer/server/classes/ImporterSelection.ts
```diff
@@ -1,12 +1,19 @@
-import type { IImporterSelection, IImporterSelectionChannel, IImporterSelectionUser } from '@rocket.chat/core-typings';
+import type {
+	IImporterSelection,
+	IImporterSelectionChannel,
+	IImporterSelectionUser,
+	IImporterSelectionContactOrIdentifier,
+} from '@rocket.chat/core-typings';
 
-export class ImporterSelection implements IImporterSelection {
+export class ImporterSelection<WithData extends boolean = true> implements IImporterSelection<WithData> {
```
MarcosSpessatto, line 8: Since this generic type is used only for the new contacts support, wouldn't it be better to create a specialized class for it and then use the generic, I think this would create confusion especially since we are using a boolean for it. It's not explicit to me when I'm reading the importer process. WDYT?

MarcosSpessatto (changes requested):

apps/meteor/app/importer-csv/server/addParsedContacts.ts
pierre-lehnen-rc: I did it here just to keep the two PRs independent from each other; I'll still be opening a third one with the actual code to convert the contacts and I'll be moving this function on it.

apps/meteor/app/importer/server/classes/ImporterSelection.ts
pierre-lehnen-rc: I want to use it for the other data types as well, but that would be a breaking change as this type is used on an endpoint.

MarcosSpessatto (approved):

PR #33424, opened 2024-10-03
fix: Custom sounds not working when storage is set to filesystem

KevLehman:

- Use `mime` type package to infer the content type of the uploaded file from its extension
- Add logs to places where we were swallowing the errors so we have visibility of what's going on 
## Issue(s)

 https://rocketchat.atlassian.net/browse/SUP-680
## Steps to test or reproduce

- Set custom sound storage to filesystem and assign a valid path
- Restart server (for some reason, these changes only apply after restart)
- Upload new custom sound
- Try to play 💥


pierre-lehnen-rc (approved):

.changeset/seven-hotels-collect.md
```diff
@@ -0,0 +1,5 @@
+---
+"@rocket.chat/meteor": patch
+---
+
+Fixes a problem with custom sounds causing files to not be playable when using filesystem storage.
```
Dnouv, line 5: Based on the structure guidelines it should be:

> Fixes a problem with custom sounds (What) that prevents files from being playable when using filesystem storage (Why) by adjusting the file access method (How)

Dnouv (changes requested):

.changeset/seven-hotels-collect.md
KevLehman: Actually, i don't mind changing. But since the original CS has the 2 of 3 elements (what why) like all the other CSs out there, we should be good.

Thanks!

PR #33430, opened 2024-10-03
feat: Current chats department filter fallback for department deletion

rique223:

Added a fallback to the current chats departments filter in which when a department is deleted and another user has the department selected in the filter, if the page is reloaded the user will be warned about the deletion and sent back to the default option of the filter.

Demo gif:
fallback_departments_filter

## Further comments

Jira task: [CONN-352](https://rocketchat.atlassian.net/browse/CONN-352)

[CONN-352]: https://rocketchat.atlassian.net/browse/CONN-352?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


gabriellsh (changes requested):
There's some code reviews I'd have liked to do first, but I have some questions that I must ask before.

Where did this improvement come from? From what I can see, there's a related issue linked in the task, but this is not a fix to that. Was this prompted from a PM or were other people involved in the definition of this feat?

Did the design team participate in this at all? At least to define the new translation?

Also, the chageset is missing. I'll convert this PR to a draft until we iron out these kinks.

rique223:
> There's some code reviews I'd have liked to do first, but I have some questions that I must ask before.
> 
> Where did this improvement come from? From what I can see, there's a related issue linked in the task, but this is not a fix to that. Was this prompted from a PM or were other people involved in the definition of this feat?
> 
> Did the design team participate in this at all? At least to define the new translation?
> 
> Also, the chageset is missing. I'll convert this PR to a draft until we iron out these kinks.

Answering here for future reference:
Where did this improvement come from?
A: It is a related issue to the one linked in the task as the structure suggests, in this case, it causes the same problem as the linked bug but instead of being just a simple bug it also is a product gap as there was no definition for what should happen with the filter if the department is deleted.

Was this prompted from a PM or were other people involved in the definition of this feat?
A: The initial idea of fixing the problem was mine but my solution was different, I took this to Renato for product input and he suggested the current flow;

Did the design team participate in this at all? At least to define the new translation?
A: Yes, the text you see in the gif is from the POC, I've just shipped the new design approved copy;

Also, the chageset is missing.
A: Nice catch, will add it shortly.

gabriellsh (changes requested):
I forgot to request these changes (I communicated the review in a meet call), and I'll request this review just for documentation purposes.

gabriellsh (changes requested):

rique223:
> @rique223 With the introduction of SCI we're going to deprecate Current Chats, so I don't see any reason to add this feature now. Would you mind to check with the Product team if we could hold this one until we get the SCI code merged and introduce this feature under Omnichannel Directory chats directly? FYI @gabriellsh

Hey Doug, thanks for the info man. I've talked with Renato about this and he approved waiting for SCI. I'll keep this PR on hold.

PR #33433, opened 2024-10-03
feat: license add-ons (external modules)

d-gubert:

Add-ons will extend the license's modules beyond the core features that are available

- [x] Prevent apps that require an add-on from being enabled if the add-on is not available
- [x] When the license is invalidated, disable apps that require add-on - this should already happen based on previous rules, but it did not happen when a license was removed, only when it became invalid.
- [x] When the license is updated, validate changes to add-ons and update configuration accordingly
## Issue(s)

[Jira task](https://rocketchat.atlassian.net/browse/ADD-14)

## Further comments

[ADR](https://adr.rocket.chat/105)


apps/meteor/ee/server/apps/communication/rest.ts
```diff
@@ -418,9 +415,13 @@ export class AppsRestApi {
...
 					void notifyAppInstall(orchestrator.getMarketplaceUrl() as string, 'install', info);
 
-					if (await canEnableApp(aff.getApp().getStorageItem())) {
+					try {
+						await canEnableApp(aff.getApp().getStorageItem());
+
 						const success = await manager.enable(info.id);
 						info.status = success ? AppStatus.AUTO_ENABLED : info.status;
+					} catch (error) {
+						// should report the error?
```
tiagoevanp, line 424: Do we have an answer for that?

ee/packages/license/src/MockedLicenseBuilder.ts
```diff
@@ -190,13 +197,17 @@ export class MockedLicenseBuilder {
...
+		this.grantedModules.push(
+			...(modules.map((module) =>
+				CoreModules.includes(module as LicenseModule) ? { module } : { module, external: true },
+			) as GrantedModules),
+		);
 		return this;
 	}
 
-	withNoGratedModules(modules: LicenseModule[]): this {
+	withNoGratedModules(modules: string[]): this {
```
tiagoevanp, line 210: Why string removing specificity here?

ee/packages/license/src/events/emitter.ts
```diff
@@ -1,27 +1,44 @@
-import type { BehaviorWithContext, LicenseModule } from '@rocket.chat/core-typings';
+import type { BehaviorWithContext } from '@rocket.chat/core-typings';
 
 import type { LicenseManager } from '../license';
 import { logger } from '../logger';
+import { isLicenseModule } from '../modules';
 
-export function moduleValidated(this: LicenseManager, module: LicenseModule) {
+export function moduleValidated(this: LicenseManager, module: string) {
```
tiagoevanp, line 7: same

ee/packages/license/src/events/emitter.ts
```diff
@@ -1,27 +1,44 @@
...
 	try {
-		this.emit('module', { module, valid: false });
+		const external = !isLicenseModule(module);
+
+		this.emit('module', { module, external, valid: false });
 	} catch (error) {
 		logger.error({ msg: `Error running module (valid: false) event: ${module}`, error });
 	}
+
+	// Duplicated call for type-guard to work
```
tiagoevanp, line 37: ?

ee/packages/license/src/v2/bundles.ts
```diff
@@ -1,33 +1,11 @@
-import type { LicenseModule } from '@rocket.chat/core-typings';
+import { CoreModules, type LicenseModule } from '@rocket.chat/core-typings';
```
tiagoevanp, line 1: Importing a non type from a package of types? This looks weird to me...

packages/core-typings/src/license/LicenseModule.ts
```diff
@@ -1,22 +1,25 @@
...
-	| 'device-management'
-	| 'federation'
-	| 'videoconference-enterprise'
-	| 'message-read-receipt'
-	| 'outlook-calendar'
-	| 'hide-watermark'
-	| 'custom-roles'
-	| 'accessibility-certification'
-	| 'unlimited-presence';
+export const CoreModules = [
```
tiagoevanp, line 1: ?

ee/packages/license/src/events/emitter.ts
d-gubert: ?

packages/core-typings/src/license/LicenseModule.ts
d-gubert: ?

ee/packages/license/src/v2/bundles.ts
d-gubert: That array is declared as a constant, which essentially means it is a readonly tuple: not a value to be changed and manipulated but rather to be referenced.

Doing it this way allows us to use the value during runtime to validate whether any given module name is a core module or not

ee/packages/license/src/MockedLicenseBuilder.ts
d-gubert: Please check the ADR above for details on the decision for add-ons (external modules)

apps/meteor/ee/server/apps/communication/rest.ts
d-gubert: Not yet

apps/meteor/ee/server/apps/orchestrator.js
```diff
@@ -177,16 +177,13 @@ export class AppServerOrchestrator {
...
+				await canEnableApp(app.getStorageItem());
 
-			if (!canEnable) {
-				this._rocketchatLogger.warn(`App "${app.getInfo().name}" can't be enabled due to CE limits.`);
-				// We need to continue as the limits are applied depending on the app installation source
-				// i.e. if one limit is hit, we can't break the loop as the following apps might still be valid
-				continue;
+				await this.getManager().loadOne(app.getID());
+			} catch (error) {
+				this._rocketchatLogger.warn(`App "${app.getInfo().name}" could not be enabled: `, error.message);
```
pierre-lehnen-rc, line 185: Shouldn't this try catch be around the `canEnableApp` call only? If the app can be enabled but fails to load, I imagine it should throw something else

```ts
			} catch (error) {
				this._rocketchatLogger.warn(`App "${app.getInfo().name}" could not be enabled: `, error.message);
				continue;
			}
			await this.getManager().loadOne(app.getID());
```

By the way can remove that `eslint-disable no-await-in-loop` if you change the loop to be a `for await`. The warning happens because this code would not work if `apps` was as AsyncIterator instead of a simple array.

ee/packages/license/src/MockedLicenseBuilder.ts
pierre-lehnen-rc: You can make a string alias like `type ExternalModule = string`, or better yet:
```ts
type ExternalModule = `${string}.${string}`;
```

to limit the format of extermal modules to strings with at least one period (the domain separator);
Then you can make `type LicenseModule = CoreModule | ExternalModule`.

pierre-lehnen-rc: You can also use this to associate `external: false` with `InternalModule` and `external: true` with `ExternalModule`, which would remove the need of doing additional typeguards on the code below.

pierre-lehnen-rc (dismissed):

apps/meteor/ee/server/apps/orchestrator.js
d-gubert: In both cases (`canEnableApp` throws and `loadOne` throws) we should not prevent the loop from going forward and checking the next apps - so I think the try/catch should surround both calls.

> By the way can remove that `eslint-disable no-await-in-loop` if you change the loop to be a `for await`. The warning happens because this code would not work if `apps` was as AsyncIterator instead of a simple array.

I'll be changing the loop to a `for..await`, then, thanks!

ee/packages/license/src/MockedLicenseBuilder.ts
d-gubert: I'm checking internally on the naming convention (I've used it in the tests but it is not confirmed)

d-gubert: Alright, naming convention applies :)

MarcosSpessatto (changes requested):
I know it will require further work to customize the test app, but it would be good to have some API tests to ensure this new behavior, but I'm good to work on that in a subsequent initiative.

apps/meteor/ee/server/startup/apps.ts
```diff
@@ -0,0 +1,30 @@
...
+Meteor.startup(() => {
+	async function disableAppsCallback() {
+		void Apps.disableApps();
+	}
+
+	License.onInvalidateLicense(disableAppsCallback);
+	License.onRemoveLicense(disableAppsCallback);
+	// Disable apps that depend on add-ons (external modules) if they are invalidated
+	License.onModule(async ({ module, external, valid }) => {
+		if (!external || valid) return;
```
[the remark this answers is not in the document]
d-gubert: Out of scope for the project. Also, as getting a valid add-on into the license is a very deliberate process, we can leave it to the admin to re-enable apps as needed IMHO

ee/packages/license/src/events/emitter.ts
```diff
@@ -2,13 +2,17 @@ import type { BehaviorWithContext, LicenseModule } from '@rocket.chat/core-typin
 
 import type { LicenseManager } from '../license';
 import { logger } from '../logger';
+import { isInternalModuleName } from '../modules';
 
 export function moduleValidated(this: LicenseManager, module: LicenseModule) {
 	try {
-		this.emit('module', { module, valid: true });
+		const external = !isInternalModuleName(module);
```
pierre-lehnen-rc, line 9: Ideally this place would have access to the full module definition to pass on to the event instead of having to determine by itself if the module is external or not, but that might be a PITA to implement now (specially for the module removed event) so I guess we can leave it as is for now and just refactor it if and only if we ever add any other attribute to the definition.

apps/meteor/ee/app/license/server/canEnableApp.ts
```diff
@@ -1,25 +1,42 @@
 import type { IAppStorageItem } from '@rocket.chat/apps-engine/server/storage';
 import { Apps } from '@rocket.chat/core-services';
+import type { LicenseModule } from '@rocket.chat/core-typings';
 import { License } from '@rocket.chat/license';
 
 import { getInstallationSourceFromAppStorageItem } from '../../../../lib/apps/getInstallationSourceFromAppStorageItem';
 
-export const canEnableApp = async (app: IAppStorageItem): Promise<boolean> => {
```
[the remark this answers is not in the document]
d-gubert: Done

apps/meteor/ee/server/startup/apps.ts
```diff
@@ -0,0 +1,30 @@
...
+		if (!enabledApps) return;
+
+		await Promise.all(
+			enabledApps.map(async (app) => {
+				if (app.getInfo().addon !== module) return;
+
+				await Apps.getManager()?.disable(app.getID(), AppStatus.DISABLED, false);
+			}),
+		);
+	});
```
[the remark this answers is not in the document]
d-gubert: Done

apps/meteor/ee/server/apps/communication/rest.ts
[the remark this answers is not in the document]
d-gubert: Done

MarcosSpessatto (approved):

pierre-lehnen-rc (approved):

MarcosSpessatto (approved):

PR #33434, opened 2024-10-03
chore: renamed voip related setting groups

aleksandernsilva:


tassoevan (changes requested):
It has to be handled as a breaking change -- i.e. tagged as `feat!:` and heading towards `release-7.0.0` branch.

apps/meteor/server/startup/migrations/v311.ts
```diff
@@ -0,0 +1,11 @@
+import { Settings } from '@rocket.chat/models';
+
+import { addMigration } from '../../lib/migrations';
+
+addMigration({
+	version: 305,
```
KevLehman, line 6:
```suggestion
	version: 311,
```

apps/meteor/server/startup/migrations/index.ts
```diff
@@ -37,11 +37,11 @@ import './v300';
 import './v301';
 import './v303';
 import './v304';
-import './v305';
```
KevLehman, line 40: Leave the 305 migration there and just add the new one

tassoevan (dismissed):

KevLehman (approved):

PR #33435, opened 2024-10-03
fix: Server not notifying users of E2EE key suggestions

KevLehman:

- Add missing notifiers
- Changed a function to be a findOneAndUpdate instead of an update to not find twice.

## Issue(s)

https://rocketchat.atlassian.net/browse/E2EE2-74

## Further comments

This prevents users to know that they got a new key suggestion, forcing them to either refresh the page or send a message (which triggers a room notification because of the message counter increasing) in order to know they have a suggested E2EE key.


pierre-lehnen-rc (dismissed):

apps/meteor/app/e2e/server/methods/updateGroupKey.ts
```diff
@@ -34,8 +34,8 @@ Meteor.methods<ServerMethods>({
 			}
 
 			// uid also has subscription to this room
-			const { modifiedCount } = await Subscriptions.setGroupE2ESuggestedKey(uid, rid, key);
-			if (modifiedCount) {
+			const { value } = await Subscriptions.setGroupE2ESuggestedKey(uid, rid, key);
+			if (value) {
 				void notifyOnSubscriptionChangedByRoomIdAndUserId(rid, uid);
```
sampaiodiego, line 39: looks like we could use `notifyOnSubscriptionChanged(value)` here as well

sampaiodiego (changes requested):

apps/meteor/app/e2e/server/methods/updateGroupKey.ts
KevLehman: there u go

PR #33439, opened 2024-10-03
chore!: Tag all private apps as migrated when downgrading to CE

matheusbsilva137:

 - Update all private apps when downgrading to CE (by removing the license or invalidating it) in order to add the `migrated` flag to them so that they are exempt from the CE limitations.

## Steps to test or reproduce

**Demo (expected behavior):**

[screen-capture (5).webm](https://github.com/user-attachments/assets/9b600c08-4444-4635-ad6c-58601c8225d0)

## Further comments

[PASLR-24](https://rocketchat.atlassian.net/browse/PASLR-24)
[PASLR-5](https://rocketchat.atlassian.net/browse/PASLR-5)


tapiarafael (approved):

packages/apps-engine/src/server/AppManager.ts
```diff
@@ -492,6 +496,37 @@ export class AppManager {
...
+        }
+
+        await app.call(AppMethod.ONUPDATE).catch((e) => console.warn('Error while migrating:', e));
+
+        await this.purgeAppConfig(app, { keepScheduledJobs: true });
+
+        const storageItem = await this.appMetadataStorage.retrieveOne(id);
+
+        app.getStorageItem().marketplaceInfo = storageItem.marketplaceInfo;
+        await app.validateLicense().catch();
```
d-gubert, line 513: Calling `catch` and passing `undefined` still ends up triggering an unhandled rejection

packages/apps-engine/src/server/AppManager.ts
```diff
@@ -492,6 +496,37 @@ export class AppManager {
...
+
+        storageItem.migrated = true;
+        storageItem.signature = await this.getSignatureManager().signApp(storageItem);
+        // This is async, but we don't care since it only updates in the database
+        // and it should not mutate any properties we care about
+        const stored = await this.appMetadataStorage.update(storageItem).catch();
```
d-gubert, line 519: Calling catch and passing undefined still ends up triggering an unhandled rejection

packages/apps-engine/src/server/AppManager.ts
```diff
@@ -492,6 +496,37 @@ export class AppManager {
...
+
+        await this.updateLocal(stored, app);
+        await this.bridges
+            .getAppActivationBridge()
+            .doAppUpdated(app)
+            .catch(() => {});
```
d-gubert, line 525: Same with the `OnUpdate` - the app did not actually change, there is no use notifying other instances

packages/apps-engine/src/server/AppManager.ts
```diff
@@ -492,6 +496,37 @@ export class AppManager {
...
     }
 
+    public async migrate(id: string): Promise<boolean> {
+        const app = this.apps.get(id);
+
+        if (!app) {
+            throw new Error(`No App by the id "${id}" exists.`);
+        }
+
+        await app.call(AppMethod.ONUPDATE).catch((e) => console.warn('Error while migrating:', e));
```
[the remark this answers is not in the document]
matheusbsilva137: I had to trigger an update here, otherwise the migration wouldn't be "visible" in the client side after a workspace is downgrade. Not sure if there's another way to do so without having to trigger an update, but I couldn't find any other way by the time I implemented this

packages/apps-engine/src/server/AppManager.ts
matheusbsilva137: Oops, I believe there are a bunch more places to fix this, then :grimacing: :grimacing: 
We're doing the same in `enable`, `disable`, `initializeApp` and `enableApp` methods in this file -- and I followed this same wrong "pattern"

PR #33520, opened 2024-10-10
fix!: Only room creator can set the E2EE room key for the first time

KevLehman:

- Checks that `room.u._id` is the same as the user requesting. Currently anyone invited to the room can set the roomKeyID (and the E2EE key) before the owner gets the key, which can cause race conditions and leave the room in a bad state.
- 
## Issue(s)

https://rocketchat.atlassian.net/browse/E2EE2-79


MarcosSpessatto (approved):
We should change this to a fix instead

sampaiodiego (approved):

PR #33525, opened 2024-10-10
chore: move to matrix-sdk-crypto

ggazzo:


MarcosSpessatto (dismissed):
By having compiled the bindings for alpine, do we still need to dynamically import the packages? 🤔

KevLehman (approved):
Apart from Marcos' comment, everything else looks fine.

.github/workflows/ci.yml
```diff
@@ -372,7 +414,8 @@ jobs:
 
   test-api:
     name: 🔨 Test API (CE)
-    needs: [checks, build-gh-docker-coverage, release-versions]
+    # needs: [checks, build-gh-docker-coverage, release-versions]
```
KevLehman, line 417: btw, is this change intended? removing the checks requirement from test-api 👀

ggazzo:
> By having compiled the bindings for alpine, do we still need to dynamically import the packages? 🤔

probably not, but there is no need to import if we are not using, also its maintaining the matrix code is not the real objective here, just update the deps and keeping the footprint smaller as we can

.github/workflows/ci.yml
```diff
@@ -373,6 +415,7 @@ jobs:
   test-api:
     name: 🔨 Test API (CE)
     needs: [checks, build-gh-docker-coverage, release-versions]
+    needs: [build-gh-docker-coverage, release-versions]
```
KevLehman, line 418:
```suggestion
```

KevLehman (approved):

PR #33527, opened 2024-10-10
fix: Cannot send messages after E2EE keys are refreshed

hugocostadev:

After the PR https://github.com/RocketChat/Rocket.Chat/pull/31589 the composer stopped updating in every state change, and because of that caused this issue with E2EE.

When the user waits for the room key to be distributed by other online users, after the distribution the room appears with the composer but the user is unable to send any message. 

When the `RoomBody` appears the composer reference is updated but the used instance is using the old reference causing issues with the composer.

## Further comments

https://rocketchat.atlassian.net/browse/E2EE2-81


apps/meteor/client/views/room/providers/hooks/useChatMessagesInstance.spec.ts
```diff
@@ -0,0 +1,67 @@
+import { renderHook } from '@testing-library/react';
+
+import { useDepsMatch } from './useDepsMatch';
+import { useInstance } from './useInstance';
```
gabriellsh, line 4: Shouldn't you be importing `useChatMessagesInstance` here?

apps/meteor/client/views/room/providers/hooks/useInstance.spec.ts
```diff
@@ -0,0 +1,83 @@
...
+
+		const { result, rerender } = renderHook(({ deps }) => useInstance(factory, deps), {
+			initialProps: { deps: ['initial-dep'] },
+			legacyRoot: true,
+		});
+
+		expect(factory).toHaveBeenCalledTimes(1);
+		expect(result.current).toEqual({ instance: new MockChatMessages() });
+
+		rerender({ deps: ['new-dep'] });
```
gabriellsh, line 34: I noticed you're mocking `useDepsMatch` return, as well as pushing new deps to the hook. I think you can remove the mock for `useDepsMatch`, just changing the array should be enough. I know you're also testing this mocked hook elsewhere, but since it's so simple of a hook I think it's better to just change the deps array instead of mocking the return value.

I recommend mocking already tested dependencies only if it decreases the test complexity enough to be justifiable. It's really a battle of how much you need to mock, sometimes mocking a return value is faster and less complex, some other times it could be the opposite, or hinder the test quality.

apps/meteor/client/views/room/providers/hooks/useInstance.spec.ts
```diff
@@ -0,0 +1,83 @@
...
+	it('should call release function when component unmounts', () => {
+		const release = jest.fn();
+		factory.mockReturnValueOnce([{ instance: new MockChatMessages() }, release]);
+		(useDepsMatch as jest.Mock).mockReturnValue(true);
+
+		const { unmount } = renderHook(() => useInstance(factory, ['initial-dep']), { legacyRoot: true });
+
+		unmount();
+
+		expect(release).toHaveBeenCalledTimes(1);
```
gabriellsh, line 81: Wdyt of adding expects to the release funcion in the other cases aswell? Ensuring it's called when the deps change, and that it's not called when they don't?

gabriellsh (changes requested):

KevLehman (approved):
symbolic approve since i want the fix :)

gabriellsh (approved):

PR #33534, opened 2024-10-11
feat: send message to admins when app is disabled due to add-on

d-gubert:


apps/meteor/tests/unit/app/lib/server/apps/disableAppsWithAddonsCallback.spec.ts
```diff
@@ -107,13 +132,55 @@ describe('disableAppsWithAddonsCallback', () => {
...
+				{ module: 'chat.rocket.test-addon', external: true, valid: false },
+			),
+		).to.not.eventually.be.rejected;
+
+		expect(AppsMock.installedApps).to.have.been.called();
+		expect(AppsMock.getManager()?.disable).to.have.been.called.once;
+		expect(AppsMock.getManager()?.disable).to.have.been.called.with('test-app-with-addon');
+	});
+
+	it('should not send messages to admins if no app was disabled', async () => {
```
MarcosSpessatto, line 147: Does it make sense to have a test case for when messages are sent?

MarcosSpessatto (changes requested):
Will we use the same changeset for the parent PR, right?

d-gubert:
> Will we use the same changeset for the parent PR, right?

I think it makes sense for us to do so, yes

apps/meteor/tests/unit/app/lib/server/apps/disableAppsWithAddonsCallback.spec.ts
d-gubert: Well, I guess it increases coverages :P

d-gubert: Done

MarcosSpessatto (approved):

PR #33316, opened 2024-09-18
feat: New Private apps limitations

MartinSchoeler:
## Proposed changes (including videos or screenshots)

This PR adds the new limitations for Private Apps in the UI, and updates the texts accordingly.

## Issue(s)

[PASLR-10](https://rocketchat.atlassian.net/browse/PASLR-10)
[PASLR-11](https://rocketchat.atlassian.net/browse/PASLR-11)
[PASLR-23](https://rocketchat.atlassian.net/browse/PASLR-23)

## Further comments

[PASLR-6]: https://rocketchat.atlassian.net/browse/PASLR-6?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[PASLR-10]: https://rocketchat.atlassian.net/browse/PASLR-10?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #33547, opened 2024-10-11
chore: Aligns Learn More link to the left corner of the modal

lucas-a-pelegrino:


MartinSchoeler (approved):

ggazzo:
Do we have cases where it is expected to be to the right? If not, we would change this in the fuselage

lucas-a-pelegrino:
> Do we have cases where it is expected to be to the right? If not, we would change this in the fuselage

I'm not sure if we have instances of these sort of content aligned solely on the right.

ggazzo (changes requested):
if its not a break change or has no 7.0.0 point to develop

lucas-a-pelegrino:
> if its not a break change or has no 7.0.0 point to develop

It's part of the new Private Apps Restrictions project, that changes the modal's text and adds a link to the docs.

It should be targeting 7.0.0, shouldn't it?

MartinSchoeler:
> if its not a break change or has no 7.0.0 point to develop

Technically this is a regression fix for an issue added here https://github.com/RocketChat/Rocket.Chat/pull/33316

I'll change the title to better inform that

tassoevan (approved):

aleksandernsilva (approved):

PR #33549, opened 2024-10-11
feat: cloud surface rendered

ggazzo:
… surfaces

### DIFF
--- a/apps/meteor/client/views/admin/subscription/SubscriptionPage.tsx
+++ b/apps/meteor/client/views/admin/subscription/SubscriptionPage.tsx
@@ -5,7 +5,9 @@ import { t } from 'i18next';
 import React, { memo, useCallback, useEffect } from 'react';
 import tinykeys from 'tinykeys';
 
-import { Page, PageHeader, PageScrollableContentWithShadow } from '../../../components/Page';
+import { Page, PageScrollableContentWithShadow } from '../../../components/Page';
+import PageBlockWithBorder from '../../../components/Page/PageBlockWithBorder';
+import PageHeaderNoShadow from '../../../components/Page/PageHeaderNoShadow';
 import { useIsEnterprise } from '../../../hooks/useIsEnterprise';
 import { useInvalidateLicense, useLicense } from '../../../hooks/useLicense';
 import { useRegistrationStatus } from '../../../hooks/useRegistrationStatus';
@@ -25,6 +27,7 @@ import PlanCardCommunity from './components/cards/PlanCard/PlanCardCommunity';
 import SeatsCard from './components/cards/SeatsCard';
 import { useRemoveLicense } from './hooks/useRemoveLicense';
 import { useWorkspaceSync } from './hooks/useWorkspaceSync';
+import UiKitSubscriptionLicense from './surface/UiKitSubscriptionLicense';
 
 function useShowLicense() {
 	const [showLicenseTab, setShowLicenseTab] = useSessionStorage('admin:showLicenseTab', false);
@@ -99,7 +102,7 @@ const SubscriptionPage = () => {
 
 	return (
 		<Page bg='tint'>
-			<PageHeader title={t('Subscription')}>
+			<PageHeaderNoShadow title={t('Subscription')}>
 				<ButtonGroup>
 					{isRegistered && (
 						<Button loading={syncLicenseUpdate.isLoading} icon='reload' onClick={() => handleSyncLicenseUpdate()}>
@@ -110,7 +113,39 @@ const SubscriptionPage = () => {
 						{t(isEnterprise ? 'Manage_subscription' : 'Upgrade')}
 					</UpgradeButton>
 				</ButtonGroup>
-			</PageHeader>
+			</PageHeaderNoShadow>
+			<PageBlockWithBorder>
+				<UiKitSubscriptionLicense
+					key='license'
+					initialView={{
+						viewId: 'license',
+						appId: 'cloud-announcements-core',
+						blocks: [
+							{
+								type: 'callout',
+								title: {
+									type: 'plain_text',
+									text: 'Callout Title',
+								},
+								text: {
+									type: 'plain_text',
+									text: 'Callout Text',
+								},
+								accessory: {
+									type: 'button',
+									text: {
+										type: 'plain_text',
+										text: 'Callout Action',
+									},
+									actionId: 'callout-action',
+									appId: 'cloud-announcements-core',
+									blockId: 'section-button',
+								},
+							},
+						],
+					}}
+				/>
+			</PageBlockWithBorder>
 			<PageScrollableContentWithShadow p={16}>
 				{(showSubscriptionCallout || syncLicenseUpdate.isLoading) && (
 					<Callout type='info' title={t('Sync_license_update_Callout_Title')} m={8}>
--- a/apps/meteor/client/components/Page/PageBlock.tsx
+++ b/apps/meteor/client/components/Page/PageBlock.tsx
@@ -1,0 +1,11 @@
+import type { Box } from '@rocket.chat/fuselage';
+import type { ComponentProps } from 'react';
+import React, { forwardRef } from 'react';
+
+import PageContent from './PageContent';
+
+const PageBlock = forwardRef<HTMLElement, ComponentProps<typeof Box>>(function PageBlock(props, ref) {
+	return <PageContent borderBlockEndColor='transparent' {...props} pb={16} ref={ref} borderBlockEndWidth='default' height='auto' />;
+});
+
+export default PageBlock;
--- a/apps/meteor/client/components/Page/PageBlockWithBorder.tsx
+++ b/apps/meteor/client/components/Page/PageBlockWithBorder.tsx
@@ -2,21 +2,12 @@ import type { Box } from '@rocket.chat/fuselage';
 import type { ComponentProps } from 'react';
 import React, { forwardRef, useContext } from 'react';
 
-import PageContent from './PageContent';
+import PageBlock from './PageBlock';
 import PageContext from './PageContext';
 
 const PageBlockWithBorder = forwardRef<HTMLElement, ComponentProps<typeof Box>>(function PageBlockWithBorder(props, ref) {
 	const [border] = useContext(PageContext);
-	return (
-		<PageContent
-			{...props}
-			pb={16}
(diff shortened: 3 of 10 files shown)


apps/meteor/client/views/admin/subscription/SubscriptionPage.tsx
```diff
@@ -48,15 +51,15 @@ const SubscriptionPage = () => {
...
+	const { data: licensesData, isLoading: isLicenseLoading } = useLicenseWithCloudAnnouncement({ loadValues: true });
 	const syncLicenseUpdate = useWorkspaceSync();
 	const invalidateLicenseQuery = useInvalidateLicense();
 
 	const subscriptionSuccess = useSearchParameter('subscriptionSuccess');
 
 	const showSubscriptionCallout = useDebouncedValue(subscriptionSuccess || syncLicenseUpdate.isLoading, 10000);
 
-	const { license, limits, activeModules = [] } = licensesData || {};
+	const { license, limits, activeModules = [], trial } = licensesData?.license || {};
```
ggazzo, line 62: why you use change this to `trial` nut ignored `cloudSyncAnnouncement` whit is the real focus here?

lucas-a-pelegrino: We were testing some changes and this got left behind, although it didn't change the behavior, I agree this change should not be part of this PR.

apps/meteor/app/cloud/server/functions/syncWorkspace/syncCloudData.ts
```diff
@@ -50,7 +51,35 @@ const fetchWorkspaceSyncPayload = async ({
 
 	assertWorkspaceSyncPayload(payload);
 
-	return payload;
+	const cloudSyncAnnouncement = {
```
gabriellsh, line 54: Remember to remove the mocked data before merge.

gabriellsh (changes requested):

apps/meteor/app/cloud/server/functions/syncWorkspace/fetchWorkspaceSyncPayload.ts
```diff
@@ -0,0 +1,46 @@
...
+	data,
+}: {
+	token: string;
+	data: Cloud.WorkspaceSyncRequestPayload;
+}): Promise<Serialized<Cloud.WorkspaceSyncResponse>> {
+	const workspaceRegistrationClientUri = settings.get<string>('Cloud_Workspace_Registration_Client_Uri');
+	const response = await fetch(`${workspaceRegistrationClientUri}/sync`, {
+		method: 'POST',
+		headers: {
+			Authorization: `Bearer ${token}`,
```
ggazzo, line 27: @lucas-a-pelegrino 
Just to be clear, I'm not against refactors, but I'm against them during feature development as it makes it difficult to understand the diffs.

dougfabris (dismissed):
LGTM!

dougfabris (approved):

apps/meteor/app/cloud/server/functions/syncWorkspace/fetchWorkspaceSyncPayload.ts
lucas-a-pelegrino: Sorry about that! I'll keep it in mind next time.

The reason I've refactored this function into a new file was so it was easier to test it.

lucas-a-pelegrino (approved):

.changeset/wicked-socks-hide.md
```diff
@@ -0,0 +1,7 @@
+---
+"@rocket.chat/meteor": patch
+"@rocket.chat/core-typings": patch
+"@rocket.chat/rest-typings": patch
+---
+
+feat: Added new callout in the subscription page to inform users of subscription upgrade eligibility when applicable.
```
gabriellsh, line 7: Changeset is not adhering to the guidelines.

ee/packages/license/src/license.ts
```diff
@@ -474,7 +474,6 @@ export class LicenseManager extends Emitter<LicenseEvents> {
 		const activeModules = getModules.call(this);
 		const externalModules = getExternalModules.call(this);
 		const license = this.getLicense();
-
```
gabriellsh, line 477: Can you remove this edit since the file hasn't changed?

packages/core-typings/src/cloud/CloudSyncAnnouncement.ts
```diff
@@ -0,0 +1,11 @@
+/* eslint-disable @typescript-eslint/naming-convention */
```
gabriellsh, line 1: Is this really necessary? The file doesn't seem to be infringing any naming-convention rules. Can you check?

gabriellsh (changes requested):

.changeset/wicked-socks-hide.md
```diff
@@ -0,0 +1,7 @@
+---
+"@rocket.chat/meteor": minor
+"@rocket.chat/core-typings": minor
+"@rocket.chat/rest-typings": minor
+---
+
+Add new callout in the subscription page to inform users of subscription upgrade eligibility when applicable.
```
gabriellsh, line 7:
```suggestion
Adds a new callout in the subscription page to inform users of subscription upgrade eligibility when applicable.
```

gabriellsh (changes requested):

apps/meteor/app/cloud/server/functions/syncWorkspace/fetchWorkspaceSyncPayload.ts
```diff
@@ -0,0 +1,46 @@
...
+		},
+		body: data,
+	});
+
+	if (!response.ok) {
+		try {
+			const { error } = await response.json();
+			throw new CloudWorkspaceConnectionError(`Failed to connect to Rocket.Chat Cloud: ${error}`);
```
KevLehman, line 35: why do you need to throw an error in here if you're ignoring it on the catch? 👀 

maybe you forgot to use the error? or to log it?

apps/meteor/app/cloud/server/functions/syncWorkspace/syncCloudData.ts
```diff
@@ -67,11 +24,17 @@ export async function syncCloudData() {
...
+		const {
+			license,
+			removeLicense = false,
+			cloudSyncAnnouncement,
+		} = await fetchWorkspaceSyncPayload({
 			token,
 			data: workspaceRegistrationData,
 		});
 
+		await Settings.updateValueById('Cloud_Sync_Announcement_Payload', JSON.stringify(cloudSyncAnnouncement ?? null));
```
KevLehman, line 36: What happens if cloudSyncAnnouncement is an empty string? Should we save it?

(Also, not sure if you need to notify that this setting has changed so the cached settings get refreshed, can you confirm? 👀 )

apps/meteor/server/settings/setup-wizard.ts
```diff
@@ -1342,5 +1342,10 @@ export const createSetupWSettings = () =>
 				},
 				secret: true,
 			});
+			await this.add('Cloud_Sync_Announcement_Payload', 'null', {
+				type: 'string', // tODO: replace setting type string for object once is implemented.
```
KevLehman, line 1346:
```suggestion
				type: 'string', // TODO: replace setting type string for object once is implemented.
```

apps/meteor/tests/unit/app/cloud/server/functions/syncWorkspace/syncCloudData.spec.ts
```diff
@@ -0,0 +1,101 @@
...
+
+		expect(
+			models.Settings.updateValueById.calledOnceWith(
+				'Cloud_Sync_Announcement_Payload',
+				JSON.stringify(workspaceSyncPayloadResponse.cloudSyncAnnouncement),
+			),
+		).to.be.true;
+	});
+
+	it("Should save as 'null' the setting update if cloudSyncAnnouncement is not present", async () => {
```
KevLehman, line 83: Can you add a case for empty strings? also , what happens if the result from the service is not a json string? can you add a case for it as well? 🙏🏽

KevLehman (changes requested):

apps/meteor/app/cloud/server/functions/syncWorkspace/fetchWorkspaceSyncPayload.ts
lucas-a-pelegrino: Huuum im not sure as to why is it built like that, as I only moved to a new file.

But I can definitely look into it and improve it.

apps/meteor/tests/unit/app/cloud/server/functions/syncWorkspace/syncCloudData.spec.ts
lucas-a-pelegrino: We won't receive an empty string from cloud, it either present with the json payload, or not.

apps/meteor/app/cloud/server/functions/syncWorkspace/syncCloudData.ts
lucas-a-pelegrino: I'm only saving the setting as string (for now) because we don't currently support objects in a setting such as this one.

This should be changed and improved once we have support for object settings.

KevLehman: > Cloud won't send empty strings for this, it is either present with a json payload or not.

I know cloud is ours but can we 100% trust it won't change some payload in the future that could end up breaking this? 👀

apps/meteor/app/cloud/server/functions/syncWorkspace/fetchWorkspaceSyncPayload.ts
KevLehman: Pls 🙏🏽

apps/meteor/app/cloud/server/functions/syncWorkspace/syncCloudData.ts
lucas-a-pelegrino: in this case, we can... the payload will only change if we change the way ui-kit works, so we should be good on this matter.

tassoevan (approved):

KevLehman (approved):

PR #33551, opened 2024-10-11
chore: Use `rmSync` instead of `rmdirSync`

KevLehman:

`rmdirSync` is soon to be deprecated


ggazzo (changes requested):
this is not a break change and also doest require nodejs > 14 please point to develop

KevLehman:
Oh did this one here cause it started to appear after node20. Can move to develop as well, don't think it is a big deal

sampaiodiego (approved):

PR #33564, opened 2024-10-14
chore!: make channels files list stop to use query param

ricardogarim:
This pull request addresses [CORE-725](https://rocketchat.atlassian.net/browse/CORE-725) by updating the `/api/v1/channels.files` endpoint.
- The query parameter has been removed from the query string, and its contents (`name` and `typeGroup`) have been moved directly to the root of the query parameters.
- For backward compatibility, the query parameter will still be supported when the environment variable `ALLOW_UNSAFE_QUERY_AND_FIELDS_API_PARAMS` is set, but this is intended for exceptional cases only and should not be used in standard implementations.

[CORE-725]: https://rocketchat.atlassian.net/browse/CORE-725?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/rest-typings/src/v1/channels/ChannelsFilesListProps.ts
```diff
@@ -0,0 +1,57 @@
...
+
+import type { PaginatedRequest } from '../../helpers/PaginatedRequest';
+
+const ajv = new Ajv({
+	coerceTypes: true,
+});
+
+export type ChannelsFilesListProps = PaginatedRequest<{
+	roomId?: string;
+	roomName?: string;
```
MarcosSpessatto, line 11: Isn't one or the other required? 🤔

packages/rest-typings/src/v1/channels/ChannelsFilesListProps.ts
```diff
@@ -0,0 +1,57 @@
...
+}>;
+
+const channelsFilesListPropsSchema = {
+	type: 'object',
+	properties: {
+		roomId: {
+			type: 'string',
+			nullable: true,
+		},
+		roomName: {
```
MarcosSpessatto, line 24: Isn't one or the other required? 🤔

MarcosSpessatto (changes requested):

MarcosSpessatto (dismissed):

PR #33569, opened 2024-10-14
feat(sci): Restrict livechat visitors to their source type scope

matheusbsilva137:

 - Added `source` field to `ILivechatVisitor` type, which stores the channel (eg API, widget, SMS, email-inbox, app) that's been used by the visitor to send messages. Each visitor can now only be linked to a single source, and each user that has already been linked to a source can't be reused in another channel. 

## Further comments

[SCI-88](https://rocketchat.atlassian.net/browse/SCI-88)

[SCI-88]: https://rocketchat.atlassian.net/browse/SCI-88?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/meteor/server/models/raw/LivechatVisitors.ts
```diff
@@ -41,17 +41,21 @@ export class LivechatVisitorsRaw extends BaseRaw<ILivechatVisitor> implements IL
 		];
 	}
 
-	findOneVisitorByPhone(phone: string): Promise<ILivechatVisitor | null> {
+	findOneVisitorByPhone(phone: string, sourceFilter?: Filter<ILivechatVisitor>): Promise<ILivechatVisitor | null> {
```
MarcosSpessatto, line 44: I think it would be better to create a new method for this making it explicit we are filtering it by source, instead of reusing this one and making the filter optional.

apps/meteor/server/models/raw/LivechatVisitors.ts
```diff
@@ -41,17 +41,21 @@ export class LivechatVisitorsRaw extends BaseRaw<ILivechatVisitor> implements IL
...
 		const query = {
 			'phone.phoneNumber': phone,
+			...(sourceFilter ? { $or: [sourceFilter, emptySourceFilter] } : emptySourceFilter),
 		};
 
 		return this.findOne(query);
 	}
 
-	findOneGuestByEmailAddress(emailAddress: string): Promise<ILivechatVisitor | null> {
+	findOneGuestByEmailAddress(emailAddress: string, sourceFilter?: Filter<ILivechatVisitor>): Promise<ILivechatVisitor | null> {
```
MarcosSpessatto, line 54: Same here

apps/meteor/app/livechat/server/api/lib/livechat.ts
```diff
@@ -48,18 +55,22 @@ async function findDepartments(
...
-			name: 1,
-			username: 1,
-			token: 1,
-			visitorEmails: 1,
-			department: 1,
-			activity: 1,
-			contactId: 1,
-		},
-	});
+export function findGuest(token: string, sourceType?: OmnichannelSourceType): Promise<ILivechatVisitor | null> {
```
MarcosSpessatto, line 58: It would be better to create a new function for filtering it by source type instead of reusing this one and making the filter optional...

apps/meteor/tests/end-to-end/api/livechat/00-rooms.ts
```diff
@@ -1256,6 +1256,32 @@ describe('LIVECHAT - rooms', () => {
 			await deleteVisitor(visitor.token);
 		});
 
+		it("should set visitor's source as API after uploading a file", async () => {
+			await updateSetting('FileUpload_Enabled', true);
+			await updateSetting('Livechat_fileupload_enabled', true);
```
MarcosSpessatto, line 1261: Let's undo these changes `after` all...

apps/meteor/tests/end-to-end/api/livechat/00-rooms.ts
```diff
@@ -1256,6 +1256,32 @@ describe('LIVECHAT - rooms', () => {
...
+				.get(api('livechat/visitors.info'))
+				.query({ visitorId: visitor._id })
+				.set(credentials)
+				.expect('Content-Type', 'application/json')
+				.expect(200);
+			expect(body).to.have.property('visitor').and.to.be.an('object');
+			expect(body.visitor).to.have.property('source').and.to.be.an('object');
+			expect(body.visitor.source).to.have.property('type', 'api');
+
+			await deleteVisitor(visitor.token);
```
MarcosSpessatto, line 1282: Move to an after

apps/meteor/tests/end-to-end/api/livechat/06-integrations.ts
```diff
@@ -115,6 +115,46 @@ describe('LIVECHAT - Integrations', () => {
 						expect(res).to.have.property('text', '<Response></Response>');
 					});
 			});
+
+			it("should set visitor's source as SMS after sending a message", async () => {
+				await updateSetting('SMS_Default_Omnichannel_Department', '');
+				await updateSetting('SMS_Service', 'twilio');
+
```
MarcosSpessatto, line 122: Let's undo these changes `after` all...

MarcosSpessatto: Also can we delete the visitor too?

apps/meteor/tests/end-to-end/api/livechat/11-livechat.ts
```diff
@@ -558,6 +558,22 @@ describe('LIVECHAT - Utils', () => {
 			const room = await createLivechatRoom(visitor.token);
 			await request.post(api('livechat/message')).set(credentials).send({ token: visitor.token, rid: room._id, msg: 'test' }).expect(200);
 		});
+		it("should set visitor's source as API after sending a message", async () => {
+			const visitor = await createVisitor();
```
MarcosSpessatto, line 562: We need to delete it after all...

apps/meteor/tests/end-to-end/api/livechat/00-rooms.ts
matheusbsilva137: Isn't it some sort of pattern from our side not to delete visitors and livechat rooms in tests? I recall @KevLehman mentioned something about this some time ago because that's an expensive behavior
None of the tests in this `describe` delete the visitor

apps/meteor/tests/end-to-end/api/livechat/11-livechat.ts
matheusbsilva137: Isn't it some sort of pattern from our side not to delete visitors and livechat rooms in tests? I recall @KevLehman mentioned something about this some time ago because that's an expensive behavior
None of the tests in this `describe` delete the visitor

apps/meteor/tests/end-to-end/api/livechat/06-integrations.ts
matheusbsilva137: I'm now resetting both settings in the `after` statement
Regarding visitors, isn't it some sort of pattern from our side not to delete them? :thinking: I could be wrong, though

apps/meteor/tests/end-to-end/api/livechat/00-rooms.ts
matheusbsilva137: I've just moved this to an `afterEach` since in this describe specifically we're already deleting visitors in multiple tests

apps/meteor/tests/end-to-end/api/livechat/11-livechat.ts
MarcosSpessatto: I think that's about rooms, If I recall correctly there's no even a way to delete it, only archive it. But we can remove visitors...

apps/meteor/tests/end-to-end/api/livechat/06-integrations.ts
matheusbsilva137: I'm now deleting visitors here too

apps/meteor/app/livechat/server/api/lib/livechat.ts
```diff
@@ -62,6 +69,21 @@ export function findGuest(token: string): Promise<ILivechatVisitor | null> {
 	});
 }
 
+export function findGuestBySource(token: string, sourceType: OmnichannelSourceType): Promise<ILivechatVisitor | null> {
```
KevLehman, line 72: This doesn't return a `ILivechatVisitor`, it returns some fields from it, pls adapt the type

apps/meteor/server/models/raw/LivechatVisitors.ts
```diff
@@ -49,6 +49,16 @@ export class LivechatVisitorsRaw extends BaseRaw<ILivechatVisitor> implements IL
 		return this.findOne(query);
 	}
 
+	findOneVisitorByPhoneAndSource(phone: string, sourceFilter: Filter<ILivechatVisitor>): Promise<ILivechatVisitor | null> {
```
KevLehman, line 52: Should these accept `options`? I know you're not using them but that would allow us to use this with projections if we need it in the future without having to change much things, wdyt?

apps/meteor/server/models/raw/LivechatVisitors.ts
```diff
@@ -57,6 +67,16 @@ export class LivechatVisitorsRaw extends BaseRaw<ILivechatVisitor> implements IL
 		return this.findOne(query);
 	}
 
+	findOneGuestByEmailAddressAndSource(emailAddress: string, sourceFilter: Filter<ILivechatVisitor>): Promise<ILivechatVisitor | null> {
```
KevLehman, line 70: Same on this one

apps/meteor/server/models/raw/LivechatVisitors.ts
```diff
@@ -57,6 +67,16 @@ export class LivechatVisitorsRaw extends BaseRaw<ILivechatVisitor> implements IL
...
+		const emptySourceFilter = { source: { $exists: false } };
+		const query = {
+			'visitorEmails.address': String(emailAddress).toLowerCase(),
```
KevLehman, line 73: Why not trusting that emailAddress is already an string when it gets here?

apps/meteor/server/models/raw/LivechatVisitors.ts
```diff
@@ -57,6 +67,16 @@ export class LivechatVisitorsRaw extends BaseRaw<ILivechatVisitor> implements IL
 		return this.findOne(query);
 	}
 
+	findOneGuestByEmailAddressAndSource(emailAddress: string, sourceFilter: Filter<ILivechatVisitor>): Promise<ILivechatVisitor | null> {
+		const emptySourceFilter = { source: { $exists: false } };
```
KevLehman, line 71: Make this one a constant on the file since you're reusing it in multiple places

apps/meteor/server/models/raw/LivechatVisitors.ts
```diff
@@ -69,10 +89,15 @@ export class LivechatVisitorsRaw extends BaseRaw<ILivechatVisitor> implements IL
...
-	findEnabled(query: Filter<ILivechatVisitor>, options?: FindOptions<ILivechatVisitor>): FindCursor<ILivechatVisitor> {
+	findEnabledBySource(
+		sourceFilter: Filter<ILivechatVisitor>,
+		query: Filter<ILivechatVisitor>,
+		options?: FindOptions<ILivechatVisitor>,
+	): FindCursor<ILivechatVisitor> {
 		return this.find(
 			{
 				...query,
+				$or: [sourceFilter, { source: { $exists: false } }],
```
KevLehman, line 100: and you could use it here too

KevLehman (changes requested):

apps/meteor/app/livechat/server/api/lib/livechat.ts
matheusbsilva137: All methods we use for sending messages in `livechat/server/v1/messages.ts` (where `findGuest` and `findGuestBySource` are used) require a full `LivechatVisitor`. We probably don't need it (since it's currently working with the incorrect type returned by `findGuest`) , but there'd be A LOT to update if we decide to fix the type returned here
I thought about holding this for another PR, wdyt? That'd almost be a TS conversion on its own :yes-gazzo:

KevLehman: Then i'd say this should return a full ILivechatVisitor :(

@MarcosSpessatto wdyt?

matheusbsilva137: Couldn't this be dangerous? idk which source of info we hide with the projection :thinking: I see we don't return some stuff such as `ip`, `host`, `livechatData`, `lastAgent`

MarcosSpessatto: I agree with what @matheusbsilva137 said, let's follow what we have already, if this is a problem, we need to fix it in a subsequent PR.

MarcosSpessatto (approved):

KevLehman (approved):

PR #33574, opened 2024-10-14
regression: voip toggle success/error toasts being displayed in quick succession

aleksandernsilva:
This PR fixes a situation where you attempt to "Enable voice calling" but it fails and both the success and erros toasts are displayed in quick succession,

errors

## Issue(s)
[VOIP-112](https://rocketchat.atlassian.net/browse/VOIP-112)

## Steps to test or reproduce
- Access workspace
- Configure and enable voip for team collab
- Attempt to enable voice calling for the same user in two different sessions
- Only the error message should appear
-

### DIFF
--- a/packages/ui-voip/src/hooks/useVoipAPI.tsx
+++ b/packages/ui-voip/src/hooks/useVoipAPI.tsx
@@ -10,6 +10,8 @@ type VoipAPI = {
 	unregister(): Promise<void>;
 	openDialer(): void;
 	closeDialer(): void;
+	onRegistered(cb: () => void): () => void;
+	onUnregistered(cb: () => void): () => void;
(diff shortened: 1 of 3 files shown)


packages/ui-voip/src/hooks/useVoipAPI.tsx
```diff
@@ -47,6 +51,8 @@ export const useVoipAPI = (): VoipAPI => {
 			closeDialer: () => voipClient.notifyDialer({ open: false }),
 			changeAudioInputDevice,
 			changeAudioOutputDevice,
+			onRegistered: (cb: () => void) => voipClient.on('registered', cb),
```
ggazzo, line 54: why not once?

MartinSchoeler (approved):

PR #33589, opened 2024-10-15
fix: Message parser build command

jeanfbrito:


debdutdeb (approved):
idk why it isn't working with just webpack, but this definitely unblocks us. 

ftr https://github.com/yarnpkg/yarn/issues/6048

the downside of this approach would be lack of control over version. but we already are using latest so we'll be good for now.

PR #33590, opened 2024-10-15
feat: block contact channel

tapiarafael:
## Further comments

[SCI-38](https://rocketchat.atlassian.net/browse/SCI-38)

[SCI-38]: https://rocketchat.atlassian.net/browse/SCI-38?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/meteor/ee/app/livechat-enterprise/server/api/lib/contacts.ts
```diff
@@ -0,0 +1,48 @@
...
+		throw new Error('error-contact-not-found');
+	}
+
+	const channel = contact.channels?.find((channel) => channel.visitorId === visitorId);
+
+	if (!channel) {
+		throw new Error('error-channel-not-found');
+	}
+
+	channel.blocked = block;
```
MarcosSpessatto, line 23: You are using the reference here to update this, right? I'd say it would be easier to read if we update the property directly instead of extracting it into a variable and then updating it. WDYT?

apps/meteor/ee/app/livechat-enterprise/server/api/lib/contacts.ts
```diff
@@ -0,0 +1,48 @@
...
+
+	await LivechatContacts.updateOne({ _id: contactId }, { $set: { channels: contact.channels } });
+}
+
+export async function hasSingleContactLicense() {
+	if (!License.hasValidLicense()) {
```
MarcosSpessatto, line 29: Shouldn't this validate a specific sci module? 🤔

apps/meteor/ee/app/livechat-enterprise/server/api/lib/contacts.ts
```diff
@@ -0,0 +1,48 @@
...
+		throw new Meteor.Error('error-action-not-allowed', 'This is an enterprise feature');
```
MarcosSpessatto, line 30: Wouldn't it be possible to throw an Error only? Instead of Meteor' specific.

apps/meteor/tests/end-to-end/api/livechat/contacts.ts
```diff
@@ -906,4 +908,133 @@ describe('LIVECHAT - contacts', () => {
...
+	});
+
+	(IS_EE ? describe : describe.skip)('[POST] omnichannel/contacts.unblock', async () => {
+		let visitor: ILivechatVisitor;
+		let room: IOmnichannelRoom;
+
+		before(async () => {
+			visitor = await createVisitor();
+			room = await createLivechatRoom(visitor.token);
+			await closeOmnichannelRoom(room._id);
```
MarcosSpessatto, line 987: Shouldn't this be present on `after` instead?

apps/meteor/tests/end-to-end/api/livechat/contacts.ts
MarcosSpessatto, line 1: Can we add some test cases for:
- When permissions are not correct
- Invalid input params

MarcosSpessatto (changes requested):

apps/meteor/ee/app/livechat-enterprise/server/api/lib/contacts.ts
tapiarafael: Yes! I copied it from another place and forgot to replace it

apps/meteor/ee/app/livechat-enterprise/server/api/lib/contacts.ts
```diff
@@ -0,0 +1,52 @@
+import type { ILivechatContact, IUser } from '@rocket.chat/core-typings';
+import { License } from '@rocket.chat/license';
+import { LivechatContacts, LivechatRooms, LivechatVisitors } from '@rocket.chat/models';
+
+import { Livechat } from '../../../../../../app/livechat/server/lib/LivechatTyped';
+import { i18n } from '../../../../../../server/lib/i18n';
+
+export async function changeContactBlockStatus({ contactId, block, visitorId }: { contactId: string; visitorId: string; block: boolean }) {
```
KevLehman, line 8: Some units would be great for this func

apps/meteor/ee/app/livechat-enterprise/server/api/lib/contacts.ts
```diff
@@ -0,0 +1,52 @@
...
+	const contact = await LivechatContacts.findOneById<Pick<ILivechatContact, '_id' | 'channels'>>(contactId, {
+		projection: { _id: 1, channels: 1 },
```
KevLehman, line 10:
```suggestion
		projection: { channels: 1 },
```

id always comes unless you put _id: 0.

apps/meteor/ee/app/livechat-enterprise/server/api/contacts.ts
```diff
@@ -0,0 +1,91 @@
...
+			POST: (params: blockContactProps) => void;
+		};
+		'/v1/omnichannel/contacts.unblock': {
+			POST: (params: blockContactProps) => void;
+		};
+	}
+}
+
+API.v1.addRoute(
+	'omnichannel/contacts.block',
```
KevLehman, line 44: i wonder if one endpoint with an `action` parameter would've been better. but this is a personal pref so if it was decided somewhere else to have both, ignore it.

apps/meteor/ee/app/livechat-enterprise/server/api/contacts.ts
```diff
@@ -0,0 +1,91 @@
...
+	},
+	{
+		async post() {
+			await hasSingleContactLicense();
+			const { contactId, visitorId } = this.bodyParams;
+
+			await changeContactBlockStatus({
+				contactId,
+				visitorId,
+
```
KevLehman, line 84:
```suggestion
```

apps/meteor/ee/app/livechat-enterprise/server/api/lib/contacts.ts
```diff
@@ -0,0 +1,52 @@
...
+	if (channelIndex === -1) {
+		throw new Error('error-channel-not-found');
+	}
+
+	contact.channels[channelIndex].blocked = block;
+
+	await LivechatContacts.updateOne({ _id: contactId }, { $set: { channels: contact.channels } });
+}
+
+export async function hasSingleContactLicense() {
```
KevLehman, line 32: why is this function async? 👀 `License.hasModule` is not async either. Plans to add some async check here? otherwise, pls make it sync.

KevLehman (changes requested):

apps/meteor/ee/app/livechat-enterprise/server/api/lib/contacts.ts
tapiarafael: Isn't the e2e enough? I mean, there's nothing too crazy that we're doing here that isn't being validated there. Most of these 'ifs' are just to ensure since a contact will never have a channel without a visitor and vice-versa

apps/meteor/ee/app/livechat-enterprise/server/api/lib/contacts.ts
tapiarafael: noted.

apps/meteor/ee/app/livechat-enterprise/server/api/contacts.ts
tapiarafael: It's a good approach as well, but we're already aligned with FE

KevLehman: no probs then!

apps/meteor/ee/app/livechat-enterprise/server/api/lib/contacts.ts
```diff
@@ -0,0 +1,52 @@
...
+		throw new Error('error-visitor-not-found');
+	}
+
+	const room = await LivechatRooms.findOneOpenByVisitorToken(visitor.token);
+
+	if (!room) {
+		return;
+	}
+
+	await Livechat.closeRoom({ room, visitor, comment: i18n.t('close-blocked-room-comment'), user });
```
KevLehman, line 51:
```suggestion
	return Livechat.closeRoom({ room, visitor, comment: i18n.t('close-blocked-room-comment'), user });
```

KevLehman: Return promise so you don't have to await twice

apps/meteor/ee/app/livechat-enterprise/server/api/lib/contacts.ts
KevLehman: Think that's fair

KevLehman (changes requested):

KevLehman (dismissed):
hope CI passes :)

matheusbsilva137 (commented):
Aren't we missing a changeset here? :eyes:

MarcosSpessatto (changes requested):
It seems we have a test breaking here, probably due to license changes

PR #33605, opened 2024-10-16
chore!: e2ee settings update and removing beta wording

hugocostadev:

Update End-to-end encryption settings translations and removed beta wording

## Further comments

https://rocketchat.atlassian.net/browse/E2EE2-83


yash-rajpal (changes requested):
I see we are changing the name of setting group, and removing old setting group in a migration, but I guess we will also have to update the E2EE setting's group property for existing E2EE settings? 🤔

hugocostadev:
> I see we are changing the name of setting group, and removing old setting group in a migration, but I guess we will also have to update the setting's group property for existing E2EE settings? 🤔

Just to keep a history here, I'll copy what @tassoevan explained in our RC chat...

> `settingsRegistry`API match settings by ID and updates them accordingly, which includes the group. It's like an implicit migration for it

apps/meteor/server/startup/migrations/v315.ts
```diff
@@ -0,0 +1,11 @@
+import { Settings } from '@rocket.chat/models';
+
+import { addMigration } from '../../lib/migrations';
+
+addMigration({
+	version: 313,
```
ggazzo, line 6: need to pay attention during conflict resolutions 
```suggestion
	version: 315,
```

hugocostadev: Sorry! You are right

apps/meteor/server/settings/e2e.ts
```diff
@@ -16,21 +16,19 @@ export const createE2ESettings = () =>
...
 
-		await this.add('E2E_Enabled_Default_DirectRooms', false, {
+		await this.add('E2E_Enabled_Default_PrivateRooms', false, {
 			type: 'boolean',
 			public: true,
 			enableQuery: { _id: 'E2E_Enable', value: true },
 		});
 
-		await this.add('E2E_Enabled_Default_PrivateRooms', false, {
+		await this.add('E2E_Enable_Encrypt_Files', true, {
```
tassoevan, line 31: ⚠️ Here the settings' IDs are changed, therefore the olds records will remain at the database and the new ones will use default values. It's different from other situation we've discussed, where other fields would be updated without data loss.

tassoevan (changes requested):

apps/meteor/server/settings/e2e.ts
hugocostadev: Only the order changed, not the ids

KevLehman (approved):
Design wise, we good

tassoevan (approved):

PR #33615, opened 2024-10-16
fix: Try to decrypt even if keyIDs don't match

KevLehman:
## Further comments

This makes the new UI a bit more forgiving with older clients.

### DIFF
--- a/apps/meteor/app/e2e/client/rocketchat.e2e.room.js
+++ b/apps/meteor/app/e2e/client/rocketchat.e2e.room.js
@@ -653,8 +653,16 @@ export class E2ERoom extends Emitter {
 		};
 	}
 
+	async doDecrypt(vector, key, cipherText) {
+		const result = await decryptAES(vector, key, cipherText);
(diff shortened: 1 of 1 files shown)


apps/meteor/app/e2e/client/rocketchat.e2e.room.js
```diff
@@ -653,29 +653,38 @@ export class E2ERoom extends Emitter {
...
+			// In these cases, we'll give a last shot using the current session key, which may not work
+			// but will be enough to help with some mobile issues.
 			if (!oldRoomKey) {
-				this.error(`Message is indecipherable. Message KeyID ${keyID} not found in old room keys`);
-				return { msg: t('E2E_indecipherable') };
+				try {
+					return await this.doDecrypt(vector, this.groupSessionKey, cipherText);
+				} catch (error) {
+					this.error('Error decrypting message: ', error, message);
+					return { msg: t('E2E_Key_Error') };
```
KevLehman, line 680:
```suggestion
					return { msg: t('E2E_indecipherable') };
```

PR #33632, opened 2024-10-17
chore: make sure appStorage.update updates the correct record

sampaiodiego:

`appStorage.update` is used by migrations and sometimes can update the wrong records, for example when there are multiple records with same `id`


d-gubert (approved):
Got it, the code had assumed there would be only one record for each app id at any moment in time. If that is not true, this query becomes necessary

PR #33633, opened 2024-10-17
fix: stop app propagation during startup

ggazzo:


d-gubert (approved):
Cool, this should cut 2 notifications for each app during the startup :rocket: Which in scale will make a big difference on resource consumption

PR #33639, opened 2024-10-18
chore: patches mongodb to make ts task less expensive

ggazzo:


ggazzo:
merging this without approval. It will speed up development, then if someone wants to modify or revert, be 100% comfortable with it

PR #33657, opened 2024-10-18
regression: Move WorkspaceCredentials models to CE

yash-rajpal:

Migration 316 breaks community edition, the migration tries to use some models which are only present in EE, which crashes the server.

image

Fixing it by moving the required models to CE

### DIFF
--- a/apps/meteor/ee/server/models/startup.ts
+++ b/apps/meteor/ee/server/models/startup.ts
@@ -7,7 +7,6 @@ import('./LivechatPriority');
 import('./OmnichannelServiceLevelAgreements');
 import('./AuditLog');
 import('./ReadReceipts');
-import('./WorkspaceCredentials');
 
 void License.onLicense('livechat-enterprise', () => {
 	import('./CannedResponse');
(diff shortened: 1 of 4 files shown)


ggazzo:
I don't see much problem with a model being public, what we should be concerned about are the controllers/logic

yash-rajpal:
> What happens if I don't have the license but then I install it?

Good Question, I have no idea, maybe @Gustrb can help us.

KevLehman:
2nd vote for making the model CE. Nothing on the model needs to be EE as long as whatever is calling it is on EE.

yash-rajpal:
3rd vote on moving the model to CE.

I'll take care of this.

Gustrb:
Yea I think the best approach would be to make the models public and avoid having to do weird things such as migrating on license change

yash-rajpal:
@ggazzo Moved the models to CE

KevLehman:
Can you update the PR title as well? We're not skipping migration anymore

KevLehman (approved):

PR #33668, opened 2024-10-18
chore: change default value for verification

tapiarafael:


apps/meteor/tests/end-to-end/api/livechat/00-rooms.ts
```diff
@@ -74,13 +74,18 @@ describe('LIVECHAT - rooms', () => {
 
 	before(async () => {
 		await updateSetting('Livechat_enabled', true);
+		await updateSetting('Livechat_Request_Verification_On_First_Contact_Only', true);
```
KevLehman, line 77: This setting won't exist on CE environments and this call will fail. I think we have an "updateEESetting" function for this, so it updates only on EE

KevLehman (changes requested):

pierre-lehnen-rc (approved):

apps/meteor/tests/end-to-end/api/livechat/00-rooms.ts
pierre-lehnen-rc: (it does exist on CE, its just disabled on it)

PR #33671, opened 2024-10-18
chore: Replace some `cursor.count` calls with `countDocuments` or `estimatedDocumentCount`

KevLehman:
## Issue(s)

https://rocketchat.atlassian.net/browse/CORE-747
## Further comments

Replaces deprecated `cursor.count` calls with updated `countDocuments` or `estimatedDocumentCount`, for full table counts.


sampaiodiego (approved):
so sad this could not land on release candidate

PR #33684, opened 2024-10-21
fix: business hours with different timezones server/client

ggazzo:


matheusbsilva137 (approved):
CI is green :partying_face: 
What about adding a changeset to this PR? Do we need one?

ggazzo:
> CI is green :partying_face: 
> 
> What about adding a changeset to this PR? Do we need one?

It is green, but I did not run on Sunday where the CI fails so I'm waiting next weekend to make a real test

ggazzo:
> What about adding a changeset to this PR? Do we need one?

Yes sure as soon as I get confirmation it's done, also I will remove from draft after confirmation . Btw its going to dismiss your review

PR #33690, opened 2024-10-21
feat: apps-engine timeout config

d-gubert:

Make the Apps-Engine runtime accept the `APPS_ENGINE_RUNTIME_TIMEOUT` environment variable to configure the time limit for requests to the Deno subprocess.
Also, makes calls to `Pre` app event methods follow the same "global" timeout, instead of the shorter 1 second.

We've had reports of some environments having timeouts where the request was actually expected to take a longer time to resolve.

Default timeout continues being 10 seconds.

### DIFF
--- a/packages/apps-engine/src/server/ProxiedApp.ts
+++ b/packages/apps-engine/src/server/ProxiedApp.ts
@@ -53,11 +53,6 @@ export class ProxiedApp {
     public async call(method: `${AppMethod}`, ...args: Array<any>): Promise<any> {
         let options;
 
-        // Pre events need to be fast as they block the user
-        if (method.startsWith('checkPre') || method.startsWith('executePre')) {
-            options = { timeout: 1000 };
-        }
-
         try {
             return await this.appRuntime.sendRequest({ method: `app:${method}`, params: args }, options);
         } catch (e) {
--- a/packages/apps-engine/src/server/runtime/deno/AppsEngineDenoRuntime.ts
+++ b/packages/apps-engine/src/server/runtime/deno/AppsEngineDenoRuntime.ts
@@ -52,6 +52,17 @@ const COMMAND_PONG = '_zPONG';
 
 export const JSONRPC_METHOD_NOT_FOUND = -32601;
 
+export function getDefaultRuntimeTimeout() {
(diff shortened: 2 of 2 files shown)


KevLehman (approved):

packages/apps-engine/src/server/ProxiedApp.ts
```diff
@@ -53,11 +53,6 @@ export class ProxiedApp {
     public async call(method: `${AppMethod}`, ...args: Array<any>): Promise<any> {
         let options;
 
-        // Pre events need to be fast as they block the user
-        if (method.startsWith('checkPre') || method.startsWith('executePre')) {
```
KevLehman, line 57: This is probably the only part i don't like. If PRE callbacks block the use, if a user registers a `executePreMessageSent` that takes the full timeout, it would take 30s by default.

For normal handlers that's good but for these, i wonder if they should have their own env, and maybe a smaller default value.

d-gubert: That's a valid concern. The use case that prompted this change was the usage of an antivirus solution, like the ClamAV app, where the time it takes for a file to be analyzed will increase with the file size.

What we may want to do is cut short the timeout of pre-message events only, which are extremelly critical, where file uploads can be expected to take a while longer

KevLehman: I like that idea

Dnouv: It might be a dumb idea, but why not do something like:

1. Users upload the file
2. On `PreFileUpload` schedule a job, and replace the file with a custom message "File is getting analyzed"
3. Once the file is done analyzing, update the default message with the File if it passed, if not show any error/delete.

d-gubert: @Dnouv that wouldn't work. On `IPreFileUpload`, you don't get to change the text of the message that will be sent, that happens after the upload has been accepted.

d-gubert: @KevLehman I'd like to leave that idea for another time, though. Maybe it is a non-problem that we're trying to fix.

PR #33695, opened 2024-10-21
regression: not allowed to install marketplace apps

Gustrb:

The issue was introduced in #32856 we were not updating the scopes according to how they were created in the cloud services, this PR associates the scopes of fleetcommand credentials to the ones in the database, using it as a cache for our cloud workspace credentials

## Further comments

CONN-368


apps/meteor/app/cloud/server/functions/getWorkspaceAccessToken.ts
```diff
@@ -24,20 +24,18 @@ export async function getWorkspaceAccessToken(forceNew = false, scope = '', save
 		return '';
 	}
 
-	const workspaceCredentials = await WorkspaceCredentials.getCredentialByScope(scope);
-	if (!workspaceCredentials) {
-		throw new CloudWorkspaceAccessTokenError();
-	}
+	const scopes = scope === '' ? [] : [scope];
```
jcbrtl, line 27: Wouldn't this end up fetching token every time?
`WorkspaceCredentials.getCredentialByScopes([])` vs `WorkspaceCredentials.updateCredentialByScopes({scopes: <something else>, ...)`

Gustrb: Oh yea, it should be the default: `workspaceScopes `

apps/meteor/app/cloud/server/functions/getWorkspaceAccessTokenWithScope.ts
```diff
@@ -62,6 +66,7 @@ export async function getWorkspaceAccessTokenWithScope(scope = '', throwOnError
 		return {
 			token: payload.access_token,
 			expiresAt,
+			scopes: scope === '' ? [...workspaceScopes] : [scope],
```
jcbrtl, line 69: I wouldn't change this func. Just
```suggestion
			scopes: payload.scope,
```

And keep things simple (same lang as backend, string instead of []string).

jcbrtl (dismissed):
mk sense

apps/meteor/app/cloud/server/functions/getWorkspaceAccessTokenWithScope.ts
```diff
@@ -8,10 +8,16 @@ import { CloudWorkspaceAccessTokenError } from './getWorkspaceAccessToken';
...
 import { retrieveRegistrationStatus } from './retrieveRegistrationStatus';
 
-export async function getWorkspaceAccessTokenWithScope(scope = '', throwOnError = false) {
+type WorkspaceAccessTokenWithScope = {
+	token: string;
+	expiresAt: Date;
+	scope: string;
+};
+
+export async function getWorkspaceAccessTokenWithScope(scope = '', throwOnError = false): Promise<WorkspaceAccessTokenWithScope> {
```
MarcosSpessatto, line 17: Is it possible to avoid this boolean parameter? It'd be clearer to have an object with named props instead.

MarcosSpessatto (changes requested):

apps/meteor/app/cloud/server/functions/getWorkspaceAccessTokenWithScope.ts
Gustrb: It is, but I wanted to avoid touching that much code on a regression fix, wdyt?

tassoevan (approved):

apps/meteor/app/cloud/server/functions/getWorkspaceAccessTokenWithScope.ts
MarcosSpessatto: agreed

MarcosSpessatto (approved):

PR #33696, opened 2024-10-21
regression: Remove user from waiting queue when the key is set via `updateGroupKey`

KevLehman:
## Further comments

This prevents the following edge case:
- User A logs in, user forgot keys
- User A creates a room, encrypted. As no keys, room stays in waiting queue state.
- User A resets their E2E keys (cause he forgot)
- User A logs in again, saves encryption keys.
- Room previously created starts processing and user sets the group keys as expected.
- User A invites User B to room
- User B is waiting keys.
- As user A reset their E2E keys, he's on the waitingQueue as well.
- User A cannot provide keys to User B so he cannot send encrypted messages.

### DIFF
--- a/apps/meteor/app/e2e/server/methods/updateGroupKey.ts
+++ b/apps/meteor/app/e2e/server/methods/updateGroupKey.ts
@@ -1,9 +1,13 @@
 import type { ServerMethods } from '@rocket.chat/ddp-client';
-import { Subscriptions } from '@rocket.chat/models';
+import { Subscriptions, Rooms } from '@rocket.chat/models';
 import { Meteor } from 'meteor/meteor';
 
 import { methodDeprecationLogger } from '../../../lib/server/lib/deprecationWarningLogger';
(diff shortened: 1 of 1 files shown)


ggazzo:
tests ? 👀

KevLehman:
The thing with the tests is that this case shouldn't be possible. You should not be able to create rooms without your E2E keys setup in the first place (that's a diff issue).

So in reality there should be no way you hit this one. But I hit on candidate because the condition (not being able to create rooms without keys) is not working :/

MarcosSpessatto:
> The thing with the tests is that this case shouldn't be possible. You should not be able to create rooms without your E2E keys setup in the first place (that's a diff issue).
> 
> So in reality there should be no way you hit this one. But I hit on candidate because the condition (not being able to create rooms without keys) is not working :/

And what about a unit-test?

MarcosSpessatto (approved):

PR #33698, opened 2024-10-21
chore: split Contacts.ts file and its tests into separate files for every function

pierre-lehnen-rc:

I'm splitting the Contacts.ts file into a different file for each function. The only function that I actually changed was `registerContact` (renamed contactId to visitorId and replaced meteor/check with an inline assertion). Everything else is unchanged other than the function locations.

I also added a few new unit tests.

## Further comments

The failing UI test is failing in the base branch.


dougfabris (approved):
frontend perfect

tapiarafael (approved):

PR #33699, opened 2024-10-21
fix: Clear filters button not being active

dougfabris:
## Further comments

[SCI-92]


apps/meteor/client/views/omnichannel/directory/providers/ChatsProvider.tsx
```diff
@@ -0,0 +1,52 @@
...
+	);
+
+	const contextValue = useMemo(
+		() => ({
+			filtersQuery,
+			setFiltersQuery,
+			resetFiltersQuery,
+			displayFilters,
+			removeFilter,
+			hasAppliedFilters,
```
gabriellsh, line 44:
```suggestion
			hasAppliedFilters: Object.values(displayFilters).filter((value) => value !== undefined).length > 0,
```

And then remove it from the other places

gabriellsh (changes requested):

gabriellsh (approved):
Thanks for implementing the context, makes it much better.

dougfabris:
@gabriellsh with your help everything gets better

PR #33702, opened 2024-10-22
fix: undefined error message

Gustrb:
## Further comments

[SCI-117](https://rocketchat.atlassian.net/browse/SCI-117)


MarcosSpessatto (changes requested):
Should we include a changeset?

Also, it seems the setting change on the tests is not part of this fix, right? If no, please create another PR

Gustrb:
Hey @MarcosSpessatto since this one is going to the feature branch we are not adding changesets to specific PR's (this was suposed to be working with the changes we made, so it would make no sense).
And the extra test files were due to the outdated branch, it is now fixed :P

tapiarafael (approved):

Gustrb:
I guess it would be good, but I don't think it would be ideal to add such change to the scope of single contact id

PR #33713, opened 2024-10-22
chore: deprecate `from` field from Email Bridge

tapiarafael:
Deprecates the `from` field from the apps.engine Email bridge to avoid misuse. Also, cloud workspaces require to have the `from` the same as defined in the settings, so we will get the value defined in the `From_Email` setting by default.


d-gubert (approved):

apps/meteor/app/apps/server/bridges/email.ts
```diff
@@ -3,14 +3,21 @@ import type { IEmail } from '@rocket.chat/apps-engine/definition/email';
...
 	constructor(private readonly orch: IAppServerOrchestrator) {
 		super();
 	}
 
 	protected async sendEmail(email: IEmail, appId: string): Promise<void> {
+		let { from } = email;
+		if (!from) {
+			this.orch.debugLog(`The app ${appId} didn't provide a from address, using the default one.`);
+			from = String(settings.get('From_Email'));
+		}
```
d-gubert, line 18: I would have gone with the good old `email.from` validation and assignment here, but you do you :joy:

Gustrb (approved):
noice

PR #33719, opened 2024-10-23
fix(omnichannel): stop calling an object through proxy

ggazzo:
[ARCH-1307](https://rocketchat.atlassian.net/browse/ARCH-1307)
backport from https://github.com/RocketChat/Rocket.Chat/pull/33596

running microservices very often you will find `cannot call shouldStart from undefined` or something pretty close.

the reason is because we are calling `getQueueWorker` from a proxy and the return undefined.

dont ask me why its working fine on CI, but locally it fails 100% local

## Further comments

[ARCH-1307]: https://rocketchat.atlassian.net/browse/ARCH-1307?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/meteor/server/services/omnichannel/service.ts
+++ b/apps/meteor/server/services/omnichannel/service.ts
@@ -16,7 +16,9 @@ export class OmnichannelService extends ServiceClassInternal implements IOmnicha
 
 	constructor() {
 		super();
+
 		this.queueWorker = new OmnichannelQueue();
+		console.log('new OmnichannelQueue', this.queueWorker);
 	}
 
 	async created() {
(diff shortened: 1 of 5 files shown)


KevLehman (dismissed):

apps/meteor/server/services/omnichannel/service.ts
```diff
@@ -16,7 +16,9 @@ export class OmnichannelService extends ServiceClassInternal implements IOmnicha
 
 	constructor() {
 		super();
+
 		this.queueWorker = new OmnichannelQueue();
+		console.log('new OmnichannelQueue', this.queueWorker);
```
KevLehman, line 21:
```suggestion
	constructor() {
		super();
		this.queueWorker = new OmnichannelQueue();
```

KevLehman (approved):

PR #33739, opened 2024-10-24
test(ui-voip): Normalize `@rocket.chat/ui-voip` unit tests

tassoevan:

It enforces the usage of `userEvent` for simulating user interactions in unit tests. It also removes the unnecessary top `describe` block in test suites and extracts state from an `useMemo` (unstable in concurrent rendering and suspense) to a ref.


packages/ui-voip/src/components/VoipDialPad/VoipDialPad.spec.tsx
```diff
@@ -1,75 +1,79 @@
...
 
-		expect(screen.getByLabelText('Phone_number')).toHaveAttribute('readOnly');
-	});
+afterEach(() => {
+	jest.clearAllTimers();
+});
 
-	it('should enable input when editable', () => {
-		render(<DialPad editable value='' onChange={jest.fn()} />, { wrapper: mockAppRoot().build(), legacyRoot: true });
+it('should not be editable by default', async () => {
```
aleksandernsilva, line 21: Does adding `async` affect the test run in any way here _(considering no `await` is being used)_ or are you just adding it preemptively?

PR #33766, opened 2024-10-24
chore: Remove hideRoom client meteor method

aleksandernsilva:
This PR removes `hideRoom` meteor method and unifies hide room calls across the application.

## Further comments

[ARCH-1340]

[ARCH-1340]: https://rocketchat.atlassian.net/browse/ARCH-1340?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/meteor/client/hooks/useRoomCloseAction.tsx
```diff
@@ -0,0 +1,80 @@
...
+
+	const hideRoom = useMutation({
+		mutationFn: () => hideRoomEndpoint({ roomId: rid }),
+		onSuccess: () => {
+			if (redirect) {
+				router.navigate('/home');
+			}
+		},
+		onError: (error) => dispatchToastMessage({ type: 'error', message: error }),
+		onSettled: () => closeModal(),
```
tiagoevanp, line 50: How about to update this onError and onSettled since we are trying to update the tool for a version that remove those methods?

tiagoevanp (changes requested):

apps/meteor/client/hooks/useRoomCloseAction.tsx
hugocostadev: The callbacks are going to be removed only for the useQuery method, useMutation continues the same

dougfabris (changes requested):
Nice working creating the new action hook using the mutation! I have a feeling that we should keep using *hide* in the name, despite the name of the endpoint we don't use to say "we closed the room", so in terms of dev exp should be better to recognize. 

I realized we're missing the optimistic update the old client method does, we should implement it in the new mutation. [Here's](https://tanstack.com/query/v4/docs/framework/react/guides/optimistic-updates) a reference on how to do it using `react-query`

dougfabris (approved):
This is a really cool refactoring job!

tassoevan (approved):

dougfabris (approved):

PR #33772, opened 2024-10-24
fix: `Waiting Queue`  feature not queueing rooms

KevLehman:

- Add new check to `getStatus` call, to return the chat as "queued" when `waiting queue` is enabled
- Change `beforeRouteChat` callback to return an inquiry, and update caller to use the returned inq instead of finding it again.
- Tests for waiting queue when running autoselection algorithms
## Issue(s)

https://rocketchat.atlassian.net/browse/SUP-663


apps/meteor/app/livechat/server/lib/QueueManager.ts
```diff
@@ -133,18 +133,42 @@ export class QueueManager {
...
+
+		if (dbInquiry.status === 'ready') {
+			logger.debug({ msg: 'Inquiry is ready. Delegating', inquiry, defaultAgent });
+			return RoutingManager.delegateInquiry(dbInquiry, defaultAgent, undefined, room);
+		}
+
+		if (dbInquiry.status === 'queued') {
+			logger.debug(`Inquiry with id ${inquiry._id} is queued. Notifying`);
+			await this.notifyQueuedInquiry(inquiry, room, defaultAgent);
+		}
```
ggazzo, line 169: Work has been done to improve performance and to avoid finds and updates, isn't this undoing the job?
Just commenting because it doesn't seem like a mere fix, it seems like everything is wrong looking at the modifications.

ggazzo (dismissed):
what do you think about starting with testing? then we can decide the best solution to fix

apps/meteor/app/livechat/server/lib/QueueManager.ts
KevLehman: Yup, this is undoing some changes. Basically, when the perf changes were done, the call to:
```
await callbacks.run('livechat.beforeRouteChat', inquiry, defaultAgent);
```
was missing. 

This call is responsible for managing the `waiting queue` feature. It basically "queues" all incoming conversations, so the queueWorker can pick them up. Some MAC limitations were removed as well when the code started to call the new `queueInquiry`. The call for `notifyQueuedInquiry` it's because previously we used the `dispatchInquiryQueued`, which did a ton of things and it's now deprecated, now there's one doing similar stuff inside queuemanager which is good.

The bug, which i have not linked yet to the draft, is caused by this missing call, so even on EE + waiting queue enabled, no chats are being sent to the queue. Code does one attempt to assign, and then when it fails, it leaves the inquiry as "ready".

As well as with MAC limitations (Felipe is testing this) since the call to `withinLimits` was not done, a chat opened after an environment is MAC-restricted would be assigned, which shouldn't happen.

(i'm still checking if there's some way to reduce the calls, I think the callback returns an updated inquiry so that would be great, here i was just testing that with the changes everything else was working, which apparently is)

KevLehman: These calls were still present on `requeueInquiry` (what a name), but this function is not called on the first chat assignment.

So, confirming that the feature comes back when the changes are reverted, confirmed they were introduced as part of the performance changes, I can start to improve from there and see a better solution.

tiagoevanp (approved):

apps/meteor/tests/e2e/omnichannel/omnichannel-livechat-queue-management-autoselection.spec.ts
```diff
@@ -0,0 +1,120 @@
...
+		const page2 = await context.newPage();
+
+		poLiveChat = new OmnichannelLiveChat(page2, api);
+		await poLiveChat.page.goto('/livechat');
+	});
+
+	test.afterAll(async ({ api }) => {
+		await Promise.all([
+			api.post('/settings/Livechat_waiting_queue', { value: false }),
+			api.post('/settings/Livechat_waiting_queue_message', { value: '' }),
```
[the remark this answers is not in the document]
KevLehman: Yup, that's intentional. `Auto_Selection` is the default value for the Routing algo, since we set it on the before, didn't change again on the after as it would be the same.

MarcosSpessatto (approved):

apps/meteor/tests/e2e/omnichannel/omnichannel-livechat-queue-management-autoselection.spec.ts
```diff
@@ -0,0 +1,120 @@
...
+					isOffline: false,
+				});
+			});
+
+			await test.step('expect to receive Waiting Queue message on chat', async () => {
+				await expect(poLiveChat.page.locator(`div >> text=${waitingQueueMessage}`)).toBeVisible();
+			});
+
+			await test.step('expect to be on spot #1', async () => {
+				await expect(poLiveChat.page.locator(`div[role='alert'] >> text=${queuePosition1}`)).toBeVisible();
```
aleksandernsilva, line 88: Could you move this locator inside `poLivechat`? I believe there's enough reuse in this test to justify it.

aleksandernsilva (changes requested):

aleksandernsilva (approved):

MarcosSpessatto (approved):

PR #33776, opened 2024-10-24
fix: user highlights matching just part of words

sampaiodiego:

created some unit tests to illustrate what was the issue with the previous code
image

### DIFF
--- a/apps/meteor/app/lib/server/functions/notifications/messageContainsHighlight.ts
+++ b/apps/meteor/app/lib/server/functions/notifications/messageContainsHighlight.ts
@@ -15,7 +15,8 @@ export function messageContainsHighlight(message: Pick<IMessage, 'msg'>, highlig
 	}
 
 	return highlights.some((highlight: string) => {
-		const regexp = new RegExp(`\\b${escapeRegExp(highlight)}\\b`, 'i');
+		const hl = escapeRegExp(highlight);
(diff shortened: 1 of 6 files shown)


apps/meteor/app/lib/server/functions/notifications/messageContainsHighlight.ts
```diff
@@ -0,0 +1,21 @@
...
+ *
+ * @returns {boolean}
+ */
+export function messageContainsHighlight(message: Pick<IMessage, 'msg'>, highlights: string[] | undefined): boolean {
+	if (!highlights || highlights.length === 0) {
+		return false;
+	}
+
+	return highlights.some((highlight: string) => {
+		const regexp = new RegExp(`\\b${escapeRegExp(highlight)}\\b`, 'i');
```
matheusbsilva137, line 18: I'm thinking, this will still make us highlight emojis (`:highlight:`), won't it? I wonder if we should take the chance to fix this too (since they're not exactly words also)

matheusbsilva137 (changes requested):

apps/meteor/app/lib/server/functions/notifications/messageContainsHighlight.ts
sampaiodiego: hmmm.. good point.. let me create a few unit tests

sampaiodiego: updated to not match emojis.. not the prettiest regex but that's the one I could come up with

matheusbsilva137 (approved):

PR #33782, opened 2024-10-25
fix: check queue settings to take inquiry

Gustrb:

Whenever an agent tried to take a specific inquiry, we were checking the shouldTriggerVerificationApp method. That method, should only be used to avoid adding a visitor inquiry to the queue, not everytime, because of that we were having a few issues such as:

If the Livechat_Require_Contact_Verification setting was set to always, no conversation would ever be available to pick, even though the visitor might have been verified

This commit introduces the isAgentAvailableToTakeContactInquiry method, that should be used in those scenarios, also removes the check for Livechat_Block_Unknown_Contacts inside the bot call, since it does not do anything there, we now only check inside the isAgentAvailableToTakeContactInquiry method, along with a new setting to block unverified contacts (that was removed for some reason)

## Further comments

[SCI-117]

[SCI-117]: https://rocketchat.atlassian.net/browse/SCI-117?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Gustrb:
I don't really know how would we do it, the shouldTriggerVerificationApp is supposed to be used before assigning a room to an agent, while the isAgentAvailableToTakeContactInquiry is supposed to be used on conversations that already have a room (might have an agent or not).

So like, we have the queue settings and the contact verifications settings, the mistake this PR fixes is that those settings should've been separate all together, but since no one really understood that we only had the check for before assigning the room

MarcosSpessatto (changes requested):
Would it be possible to add some tests to ensure the behavior?

tapiarafael (approved):

apps/meteor/ee/tests/unit/apps/livechat-enterprise/server/lib/isAgentAvailableToTakeContactInquiry.spec.ts
```diff
@@ -0,0 +1,81 @@
...
+	it('should return true if the contact has the verified channel', async () => {
+		modelsMock.LivechatContacts.findOneById.resolves({
+			unknown: false,
+			channels: [
+				{ verified: true, visitor: { source: { type: 'channelName' }, visitorId: 'visitorId' } },
+				{ verified: false, visitor: { source: { type: 'othername' }, visitorId: 'visitorId' } },
+			],
+		});
+		settingsMock.get.withArgs('Livechat_Block_Unknown_Contacts').returns(true);
+		settingsMock.get.withArgs('Livechat_Block_Unverified_Contacts').returns(true);
```
MarcosSpessatto, line 77: Can we add more cases to cover when these settings are false?

MarcosSpessatto (changes requested):

MarcosSpessatto (approved):

apps/meteor/ee/tests/unit/apps/livechat-enterprise/server/lib/shouldTriggerVerificationApp.spec.ts
```diff
@@ -0,0 +1,80 @@
...
+		const result = await runShouldTriggerVerificationApp(() => undefined, 'visitorId', 'contactId', {});
+		expect(
+			modelsMock.LivechatContacts.findOneById.calledOnceWith('contactId', sinon.match({ projection: { _id: 1, unknown: 1, channels: 1 } })),
+		).to.be.true;
+		expect(result).to.be.false;
+	});
+
+	it('should not trigger a verification app if there is no configured app', async () => {
+		modelsMock.LivechatContacts.findOneById.resolves({ _id: 'contactId' });
+		settingsMock.get.returns('');
```
KevLehman, line 44:
```suggestion
		settingsMock.get.withArgs('Livechat_Contact_Verification_App').returns('');
```

PR #33791, opened 2024-10-25
test: `edit-room-retention-policy permission`duplicated actions

MartinSchoeler:

Removes duplicated actions in the test `edit-room-retention-policy permission` making the tests run a little bit faster
the same actions were being used in beforeEach and in the tests


dougfabris (commented):
don't you think we can use a similar strategy in `retention policy override`, adding the steps to open the pruneAccordion in `beforeEach` ?

dougfabris (approved):

PR #33809, opened 2024-10-28
refactor!: Marketplace state handling

tassoevan:


MartinSchoeler (changes requested):
It seems there is also a reactivity issue when installing an app

for example here I've just enabled the app but the buttons still show install 
image

same when disabling it

image

MartinSchoeler (changes requested):
When removing an app in the installed route, this happens

Screenshot 2024-10-31 at 10 17 24

MartinSchoeler (dismissed):
Nice

MartinSchoeler (dismissed):
Nice²

MartinSchoeler (changes requested):
When I try to install an app directly from the list the browser crashes

Screenshot 2024-11-26 at 15 28 10
Screenshot 2024-11-26 at 15 28 49

MartinSchoeler:
> When I try to install an app directly from the list the browser crashes
> 
> Screenshot 2024-11-26 at 15 28 10 Screenshot 2024-11-26 at 15 28 49

It seems to be an issue with that specific app, trying to install it via its page generates the same error

MartinSchoeler (changes requested):
https://github.com/user-attachments/assets/2be1ab39-6979-45ef-80be-dda7d798acf3

If you exit a purchase modal, the install button keeps stuck in the loading state forever, this is something we've fixed a while ago

MartinSchoeler:
> Screen.Recording.2024-11-27.at.15.41.20.mov 
> If you exit a purchase modal, the install button keeps stuck in the loading state forever, this is something we've fixed a while ago

Screenshot 2024-11-27 at 15 44 04

Same with the list view

tassoevan:
> this is something we've fixed a while ago

No, I've just seen it in `develop`.

PR #33810, opened 2024-10-28
feat: add cursor pagination support on sync messages

ricardogarim:
https://rocketchat.atlassian.net/browse/ARCH-1203

As per [CORE-686](https://rocketchat.atlassian.net/browse/CORE-686) and [CORE-682](https://rocketchat.atlassian.net/browse/CORE-682), this PR introduces cursor pagination to the messages/get Meteor method route, which supports the `chat.syncMessages` route. This improvement allows users to retrieve updates from a given date more efficiently, without compromising response times.

While cursor pagination is now available, it does not affect any existing behaviors. To use cursor pagination, instead of passing `lastUpdate` as query parameter, users will need to pass `next` or `previous`, along with `count` and `type`.

- `next` or `previous`: These represent the cursor pointers indicating whether the query should retrieve items from a later or earlier point in time. The cursor value must be the number of milliseconds, as it follows [Date getTime()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/getTime).
- `count`: Specifies the number of records to return.
- `type`: With the introduction of cursors, both updated and deleted messages cannot be retrieved in the same response. Users must specify whether they want to retrieve `DELETED` (for deleted messages) or `UPDATED` (for updated messages, which is the default use case).

Notes:
- In the future, it might be worth considering isolating this functionality into a separate route. However, for now, the current implementation ensures no conflicts or breaking changes, and the route’s signature remains unchanged.
- This implementation has been tested to ensure that existing users relying on `lastUpdate` will continue to have the same experience, with no changes required on their end.
- Keep in mind that while performance is acceptable for now, there may be opportunities to optimize database indexes further. Pagination has been a missing feature, so addressing it was a high priority.
- This change is especially beneficial for users who frequently work with large volumes of messages, such as when a username with a high message count in a room is changed, as cursor pagination offers a more efficient way to paginate through results.
---
Edit: As per [OPI-53](https://rocketchat.atlassian.net/browse/OPI-53), it adds cursor null conditions when there's no more records to fetch.

[CORE-686]: https://rocketchat.atlassian.net/browse/CORE-686?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[OPI-53]: https://rocketchat.atlassian.net/browse/OPI-53?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[CORE-682]: https://rocketchat.atlassian.net/browse/CORE-682?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/meteor/server/publications/messages.ts
```diff
@@ -7,65 +7,228 @@ import type { FindOptions } from 'mongodb';
...
+				oldest: oldestDate,
+				inclusive,
+				count,
+				unreads,
+			});
+		}
+
+		const response = lastUpdate
+			? await handleWithoutPagination(rid, lastUpdate)
+			: await handleCursorPagination(type ?? 'UPDATED', rid, count, next, previous);
```
MarcosSpessatto, line 230: I'd say it would be better to throw an error if type is undefined rather than assume it as 'UPDATED' as default.

apps/meteor/server/publications/messages.ts
```diff
@@ -7,65 +7,228 @@ import type { FindOptions } from 'mongodb';
...
+	// if messages length is greater than count, it means we are not at the end of the messages
+	// and we shall return a cursor
+	if (count && messages.length > count) {
+		return mountCursorFromMessage(messages[messages.length - 2], type);
+	}
+
+	return null;
+}
+
+export async function handleWithoutPagination(rid: IRoom['_id'], lastUpdate: Date) {
```
MarcosSpessatto, line 128: Isn't `lastUpdate` possibly null | undefined at this point? 🤔

apps/meteor/server/publications/messages.ts
```diff
@@ -7,65 +7,228 @@ import type { FindOptions } from 'mongodb';
...
+	const query = { $gt: lastUpdate };
+	const options: FindOptions<IMessage> = { sort: { ts: -1 } };
+
+	const [updatedMessages, deletedMessages] = await Promise.all([
+		Messages.findForUpdates(rid, query, options).toArray(),
+		Messages.trashFindDeletedAfter(lastUpdate, { rid }, { projection: { _id: 1, _deletedAt: 1 }, ...options }).toArray(),
```
MarcosSpessatto, line 134: options will override the options you provided, is this intended?

apps/meteor/tests/end-to-end/api/chat.ts
```diff
@@ -2945,6 +2945,283 @@ describe('[Chat]', () => {
...
+				.expect('Content-Type', 'application/json')
+				.expect(400)
+				.expect((res) => {
+					expect(res.body).to.have.property('success', false);
+					expect(res.body.errorType).to.be.equal('error-lastUpdate-param-invalid');
+				})
+				.end(done);
+		});
+
+		it('should return an error when user is not allowed to access the room', (done) => {
```
MarcosSpessatto, line 2999: I think this title doesn't match the test case, it seems you are providing an invalid room id instead...

apps/meteor/tests/unit/server/publications/messages.spec.ts
```diff
@@ -0,0 +1,282 @@
...
+		},
+		'@global': true,
+	},
+	'@rocket.chat/models': {
+		Messages: messagesMock,
+	},
+});
+
+describe('extractTimestampFromCursor', () => {
+	it('should return the correct timestamp', () => {
```
MarcosSpessatto, line 45: Can we create another test case providing a no-date compliant string?

apps/meteor/tests/unit/server/publications/messages.spec.ts
```diff
@@ -0,0 +1,282 @@
...
+		const result = mountCursorQuery({ next: mockDate.toString() });
+		expect(result.query).to.deep.equal({ $gt: new Date(mockDate) });
+	});
+
+	it('should return a query with $lt when previous is provided', () => {
+		const result = mountCursorQuery({ previous: mockDate.toString() });
+		expect(result.query).to.deep.equal({ $lt: new Date(mockDate) });
+	});
+
+	it('should return a query with $gt and sort when neither next nor previous is provided', () => {
```
MarcosSpessatto, line 66: I think this is not possible due to the higher layers validation, but what if both are provided? Maybe put a validation here as well if we want to use the same approach in other places.

apps/meteor/tests/unit/server/publications/messages.spec.ts
```diff
@@ -0,0 +1,282 @@
...
+		const result = mountCursorQuery({ count: 10 });
+		expect(result.query).to.deep.equal({ $gt: new Date(0) });
+		expect(result.options).to.deep.equal({ sort: { ts: -1 } });
+	});
+});
+
+describe('mountCursorFromMessage', () => {
+	it('should return the updated timestamp for UPDATED type', () => {
+		const message: Pick<IMessage, '_updatedAt'> = {
+			_updatedAt: new Date('2024-01-01T00:00:00Z'),
```
MarcosSpessatto, line 76: Is there any possibility of the `_updatedAt` property being undefined?

apps/meteor/tests/unit/server/publications/messages.spec.ts
```diff
@@ -0,0 +1,282 @@
...
+describe('mountNextCursor', () => {
+	it('should return a cursor for the most recent message when messages are present', () => {
+		// Messages are already sorted by descending order
+		const messages = [{ _updatedAt: new Date('2024-10-01T10:00:00Z') }, { _updatedAt: new Date('2024-10-01T09:00:00Z') }];
+		const type = 'UPDATED';
+		const result = mountNextCursor(messages, type);
+		expect(result).to.equal(`${messages[0]._updatedAt.getTime()}`);
+	});
+
+	it('should reverse messages if next is provided', () => {
```
MarcosSpessatto, line 110: We could improve this title, it seems to me the array of messages will be reversed...

MarcosSpessatto (changes requested):
Can we add a changeset?

apps/meteor/tests/unit/server/publications/messages.spec.ts
```diff
@@ -0,0 +1,281 @@
...
+	it('should return the deleted timestamp for DELETED type', () => {
+		const message: Partial<IMessage> & { _deletedAt: Date } = {
+			_deletedAt: new Date('2024-01-01T00:00:00Z'),
+		};
+
+		const result = mountCursorFromMessage(message, 'DELETED');
+		expect(result).to.equal(`${message._deletedAt.getTime()}`);
+	});
+
+	it('should throw an error if DELETED type and _deletedAt is not present', () => {
```
MarcosSpessatto, line 100: Can we add a case for UPDATED as well?

apps/meteor/tests/unit/server/publications/messages.spec.ts
```diff
@@ -0,0 +1,281 @@
...
+	});
+
+	it('should return null if no messages and no previous cursor', () => {
+		const messages: IMessage[] = [];
+		const type = 'UPDATED';
+		const result = mountNextCursor(messages, type);
+		expect(result).to.equal(null);
+	});
+
+	it('should return decremented previous cursor if no messages and previous is provided', () => {
```
MarcosSpessatto, line 125:
```suggestion
	it('should return decremented previous cursor if no messages, but previous is provided', () => {
```

MarcosSpessatto (changes requested):

.changeset/mean-cobras-sneeze.md
```diff
@@ -0,0 +1,6 @@
+---
+'@rocket.chat/rest-typings': patch
+'@rocket.chat/meteor': patch
```
MarcosSpessatto, line 3:
```suggestion
'@rocket.chat/rest-typings': minor
'@rocket.chat/meteor': minor
```

MarcosSpessatto (changes requested):

MarcosSpessatto (approved):

PR #33816, opened 2024-10-28
feat: use Livechat Contacts as a reference for the MAC limit

matheusbsilva137:

 - Replace Visitors by Contacts when calculating workspaces' MAC;
 - Store `activity` object in Contacts instead of Visitors.

## Further comments

[SCI-76]

[SCI-76]: https://rocketchat.atlassian.net/browse/SCI-76?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/meteor/app/livechat/server/lib/contacts/ContactMerger.ts
```diff
@@ -229,6 +236,15 @@ export class ContactMerger {
 			}
 		}
 
+		if (newActivities.length) {
+			const newActivity = newActivities.shift();
+			if (newActivity) {
+				const distinctActivities = new Set([...newActivity, ...(contact.activity || [])]);
+				const latestActivities = Array.from(distinctActivities).sort().slice(-12);
```
debdutdeb, line 243: what is -12? a comment maybe (I am not opinionated on naming everything but at least a comment to remove confusion 🙈)

debdutdeb: also, sort on a string list, named as latestActivites, I am confused slightly. How is a string sort fetching conceptually 'latest' activities?

debdutdeb: got this from a test
```
const period = moment().format('YYYY-MM');
```

so the strings are effectively dates and 12 my guess is months?

could we type `activity` instead of `string[]` more like
```
type YearMonthString = string;
{ activity: YearMonthString[] }
```
?

matheusbsilva137: I wanted to keep the "rule" we had [here](https://github.com/RocketChat/Rocket.Chat/blob/0300995515f6f22a62cb57f3a584d054ff00e67b/apps/meteor/server/models/raw/LivechatVisitors.ts#L472). When inserting new month activities, we always keep only the latest 12 ones
I'll keep the type as string, but use a new name to make this clear as you suggested

KevLehman: Yep, just the last 12 are saved as activity. Got curious, can't we do this on mongo? the slicing?

.changeset/slow-flies-hear.md
```diff
@@ -0,0 +1,7 @@
+---
+"@rocket.chat/meteor": minor
+"@rocket.chat/core-typings": minor
+"@rocket.chat/model-typings": minor
+---
+
+Replaces Livechat Visitors by Contacts on workspaces' MAC count
```
KevLehman, line 7: Think we should get a more product-ish description here as this may affect pricing (even if its for good)

apps/meteor/server/models/raw/LivechatContacts.ts
```diff
@@ -195,4 +199,36 @@ export class LivechatContactsRaw extends BaseRaw<ILivechatContact> implements IL
...
 		).toArray();
 	}
+
+	isContactActiveOnPeriod(visitorId: string, period: string): Promise<boolean> {
+		const query = {
+			'channels.visitorId': visitorId,
+			'activity': period,
+		};
+
+		return this.findOne(query, { projection: { _id: 1 } }).then(Boolean);
```
KevLehman, line 209: what about just counting? that way it would return 0/>0 and the function can work without having to chain the promise

apps/meteor/server/models/raw/LivechatContacts.ts
```diff
@@ -195,4 +199,36 @@ export class LivechatContactsRaw extends BaseRaw<ILivechatContact> implements IL
 			options,
 		).toArray();
 	}
+
+	isContactActiveOnPeriod(visitorId: string, period: string): Promise<boolean> {
+		const query = {
+			'channels.visitorId': visitorId,
+			'activity': period,
```
KevLehman, line 206: Is this query using the index? I kinda remember pfe indexes being a bit picky on when to cover a query

apps/meteor/app/livechat/server/lib/Helper.ts
```diff
@@ -89,7 +90,7 @@ export const createLivechatRoom = async (
 	const { _id, username, token, department: departmentId, status = 'online' } = guest;
 	const newRoomAt = new Date();
 
-	const { activity } = guest;
+	const activity = guest.activity || (await LivechatContacts.findOneByVisitorId(guest._id, { projection: { activity: 1 } }))?.activity;
```
KevLehman, line 93:
```suggestion
	const activity = guest.activity || (await LivechatContacts.findOneByVisitorId<Pick<...>>(guest._id, { projection: { activity: 1 } }))?.activity;
```

if the func doesnt support, would be nice to add support

apps/meteor/server/models/raw/LivechatVisitors.ts
```diff
@@ -37,7 +37,6 @@ export class LivechatVisitorsRaw extends BaseRaw<ILivechatVisitor> implements IL
 			{ key: { username: 1 } },
 			{ key: { 'contactMananger.username': 1 }, sparse: true },
 			{ key: { 'livechatData.$**': 1 } },
-			{ key: { activity: 1 }, partialFilterExpression: { activity: { $exists: true } } },
```
KevLehman, line 40: Can you leave this commented with a note that we have to remove it? otherwise it may end up being forgotten for next major

KevLehman (changes requested):

apps/meteor/app/livechat/server/lib/contacts/ContactMerger.ts
matheusbsilva137: > Got curious, can't we do this on mongo? the slicing?
We could, but slice can't be performed inside a `$set`. So we'd need an extra db call to perform a push with a slice instead -- so I thought it wasn't worth it since I can do the slice myself and set it with a single DB call

KevLehman: Yeah but $set and $push can coexist on the same update statement 👀 

the only reason i think justifies is the sortin 🤔 so I'll let you decide if there's a way to use push or keep what you doing here.

apps/meteor/server/models/raw/LivechatContacts.ts
matheusbsilva137: Good idea! I just repeated what we were doing before, but that's definitely better

apps/meteor/server/models/raw/LivechatContacts.ts
matheusbsilva137: Yeppp, Compass tells me it's being used :candle:

apps/meteor/app/livechat/server/lib/contacts/ContactMerger.ts
matheusbsilva137: > the only reason i think justifies is the sortin 🤔 so I'll let you decide if there's a way to use push or keep what you doing here.

That's it, I was afraid of using push because it could mess up the order :thinking: we call this function once per similar contact that should be merged, so we can't just assume that the new activities we're inserting are the latest (and make previous elements be sliced out of the array)
I'll move on with `$set`, then

packages/core-typings/src/ILivechatContact.ts
```diff
@@ -37,4 +40,5 @@ export interface ILivechatContact extends IRocketChatRecord {
 		ts: Date;
 	};
 	importIds?: string[];
+	activity?: YearMonthString[];
```
KevLehman, line 43: hmmmm... not sure this provides much value 🤔 can't we keep with just `string` and the comment above here?

apps/meteor/server/models/raw/LivechatContacts.ts
```diff
@@ -195,4 +199,37 @@ export class LivechatContactsRaw extends BaseRaw<ILivechatContact> implements IL
...
 		).toArray();
 	}
+
+	async isContactActiveOnPeriod(visitorId: string, period: string): Promise<boolean> {
+		const query = {
+			'channels.visitorId': visitorId,
+			'activity': period,
+		};
+
+		const foundActiveContact = (await this.countDocuments(query)) > 0;
```
KevLehman, line 209: Think we can just use JS help and return the number here, so people can use `if isContactActiveOnPeriod` and let js convert it to bool, avoiding the second promise. wdyt?

KevLehman (changes requested):

KevLehman (approved):

apps/meteor/server/models/raw/LivechatVisitors.ts
MarcosSpessatto: By commenting the index, aren't we already removing it?

MarcosSpessatto (changes requested):

apps/meteor/server/models/raw/LivechatVisitors.ts
matheusbsilva137: > By commenting the index, aren't we already removing it?

We're only preventing thisi ndex from being created in new workspaces, I believe
We'd still need a migration to drop it in older workspaces that had this index created

MarcosSpessatto (dismissed):

tapiarafael (approved):

MarcosSpessatto (approved):

KevLehman (approved):

KevLehman (approved):

MarcosSpessatto (approved):

matheusbsilva137:
Since this has now been merged, we might need to adjust MAC related statistics later. cc @KevLehman and @MarcosSpessatto 
(not sure if there is a task for this, so just making sure :) )

PR #33817, opened 2024-10-28
chore: Remove `starMessage` client meteor method

yash-rajpal:

### DIFF
--- a/.changeset/fifty-trees-do.md
+++ b/.changeset/fifty-trees-do.md
@@ -1,0 +1,5 @@
+---
+'@rocket.chat/meteor': patch
+---
+
+Remove client side `starMessage` meteor method and replace this meteor method call with rest API.
--- a/apps/meteor/client/lib/starMessage.ts
+++ b/apps/meteor/client/lib/starMessage.ts
@@ -1,0 +1,50 @@
+import type { IMessage } from '@rocket.chat/core-typings';
+import { Meteor } from 'meteor/meteor';
+
+import { ChatMessage, ChatSubscription } from '../../app/models/client';
+import { settings } from '../../app/settings/client';
+import { t } from '../../app/utils/lib/i18n';
+import { dispatchToastMessage } from './toast';
+
+export const starMessage = (message: IMessage, starred: boolean) => {
+	const uid = Meteor.userId();
+
+	if (!uid) {
+		throw new Error(t('error-starring-message'));
+	}
+
+	if (!ChatSubscription.findOne({ rid: message.rid })) {
+		throw new Error(t('error-starring-message'));
+	}
+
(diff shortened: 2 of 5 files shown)


apps/meteor/client/startup/actionButtons/starMessage.ts
```diff
@@ -19,7 +20,8 @@ Meteor.startup(() => {
 			const { message = messageArgs(this).msg } = props;
 
 			try {
-				await sdk.call('starMessage', { ...message, starred: true });
+				await sdk.rest.post('/v1/chat.starMessage', { messageId: message._id });
+				starMessage(message, true);
```
tassoevan, line 24: What will happen if `starMessage()` fails and throws and error? Shouldn't `['rooms', message.rid, 'starred-messages']` queries be invalidated regardless of this?

tassoevan (changes requested):

apps/meteor/client/startup/actionButtons/starMessage.ts
yash-rajpal: Maybe we should put this query invalidation in a `finally` block, since it should happen regardless of try or catch block.

.changeset/fifty-trees-do.md
```diff
@@ -0,0 +1,5 @@
+---
+'@rocket.chat/meteor': patch
+---
+
+Remove client side `starMessage` meteor method and replace this meteor method call with rest API.
```
debdutdeb, line 5:
```suggestion
Removes client side `starMessage` meteor method and replaces this meteor method call with rest API.
```

But, I kinda think these changesets are not needed.

Or replace with something much much more generic, `Reduces websocket overhead from web client`.

given these are for customers, and not considered to be breaking, they do seem out of place in terms of friendliness and customer facing-ness (if that's a word);

apps/meteor/client/lib/starMessage.ts
```diff
@@ -0,0 +1,50 @@
+import type { IMessage } from '@rocket.chat/core-typings';
+import { Meteor } from 'meteor/meteor';
+
+import { ChatMessage, ChatSubscription } from '../../app/models/client';
+import { settings } from '../../app/settings/client';
+import { t } from '../../app/utils/lib/i18n';
+import { dispatchToastMessage } from './toast';
+
+export const starMessage = (message: IMessage, starred: boolean) => {
+	const uid = Meteor.userId();
```
debdutdeb, line 10: if this is a new function, shouldn't we be able to just pass the uid instead of depending on meteor?

tassoevan: As we intent to drop minimongo collections, the whole content of this function is going to be outdated soon. Also, there is a change coming from #34017 that might stale it.

tassoevan (dismissed):

tassoevan (approved):

PR #33820, opened 2024-10-29
regression: Fix email inboxes feature not working

tiagoevanp:

Upgrade some important dependencies for the feature


apps/meteor/client/views/admin/emailInbox/EmailInboxForm.tsx
```diff
@@ -45,7 +45,7 @@ const EmailInboxForm = ({ inboxData }: { inboxData?: IEmailInboxPayload }): Reac
 		handleSubmit,
 		formState: { errors, isDirty },
 	} = useForm({
-		defaultValues: {
+		values: {
```
aleksandernsilva, line 48: Why do you need these values to be reactive?

aleksandernsilva (approved):

PR #33824, opened 2024-10-29
feat: remove visitor source restrictions

pierre-lehnen-rc:

Reverting the source restriction, as we'll follow a different path:

https://adr.rocket.chat/0114


packages/model-typings/src/models/ILivechatVisitorsModel.ts
```diff
@@ -82,25 +66,15 @@ export interface ILivechatVisitorsModel extends IBaseModel<ILivechatVisitor> {
...
+	findEnabled(query: Filter<ILivechatVisitor>, options?: FindOptions<ILivechatVisitor>): FindCursor<ILivechatVisitor>;
 
 	countVisitorsOnPeriod(period: string): Promise<number>;
 	saveGuestById(
 		_id: string,
 		data: { name?: string; username?: string; email?: string; phone?: string; livechatData: { [k: string]: any } },
 	): Promise<UpdateResult | Document | boolean>;
 	setLastChatById(_id: string, lastChat: Required<ILivechatVisitor['lastChat']>): Promise<UpdateResult>;
-	setSourceById(_id: string, source: ILivechatVisitor['source']): Promise<UpdateResult>;
 	countVisitorsBetweenDate({ start, end, department }: { start: Date; end: Date; department?: string }): Promise<number>;
```
debdutdeb, line 79: this is probably changed to `getVisitorsBetweenDate` ?

debdutdeb (changes requested):

packages/model-typings/src/models/ILivechatVisitorsModel.ts
debdutdeb: likely why build is failing

PR #33828, opened 2024-10-29
regression: Tracing times

KevLehman:

Before:
image

After:
image

## Further comments

`trace.end` not taking promises into consideration and finishing spans earlier.

if makes no sens pls close, thx


sampaiodiego (changes requested):
this is changing behavior when an exception is happening. it used to just log the exception for API requests with an error, but now it is giving an `UnHandledPromiseRejection`

sampaiodiego (approved):

PR #33833, opened 2024-10-30
chore: Remove openRoom client meteor method

aleksandernsilva:
This PR removes `openRoom` meteor method both on client and server.

## Issue(s)
[ARCH-1341](https://rocketchat.atlassian.net/browse/ARCH-1341)
## Further comments

[ARCH-1341]: https://rocketchat.atlassian.net/browse/ARCH-1341?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/meteor/server/methods/openRoom.ts
yash-rajpal, line 1: We can't remove a meteor method from server just like that.

Many customers/users use these methods as an integration (just like REST API), so removing these from server side is a breaking change and requires a major release.

Also, we issue a deprecation warning on usage of such server side meteor methods, only after releasing such warnings, we remove the methods in a major.

apps/meteor/app/api/server/v1/rooms.ts
```diff
@@ -877,3 +879,21 @@ API.v1.addRoute(
...
+API.v1.addRoute(
+	'rooms.open',
+	{ authRequired: true, isRoomsOpenProps },
+	{
+		async post() {
+			const { roomId } = this.bodyParams;
+
+			if (!roomId) {
+				return API.v1.failure('Room not found');
+			}
```
debdutdeb, line 892: shouldnt this be validated by ajv?

apps/meteor/server/lib/openRoom.ts
```diff
@@ -0,0 +1,24 @@
+import { Subscriptions } from '@rocket.chat/models';
+
+import { notifyOnSubscriptionChangedByRoomIdAndUserId } from '../../app/lib/server/lib/notifyListener';
+
+export async function openRoom(userId: string, roomId: string) {
+	check(userId, String);
+	check(roomId, String);
```
debdutdeb, line 7: aren't we moving away from meteor/check?

packages/rest-typings/src/v1/rooms.ts
```diff
@@ -598,6 +598,21 @@ const roomsCleanHistorySchema = {
...
 
+type RoomsOpenProps = {
+	roomId: string;
+};
+
+const roomsOpenSchema = {
+	type: 'object',
+	properties: {
+		roomId: {
+			type: 'string',
```
debdutdeb, line 609: i don't remember but probably something like `nullable: false` or a regex to validate non empty? we must have done similar somewhere else

apps/meteor/server/lib/openRoom.ts
```diff
@@ -0,0 +1,24 @@
...
+
+	if (!roomId) {
+		throw new Meteor.Error('error-invalid-room', 'Invalid room');
+	}
+
+	if (!userId) {
+		throw new Meteor.Error('error-invalid-user', 'Invalid user');
```
debdutdeb, line 14: shouldn't these be plain Errors? for the meteor method wrap Error in Meteor.Error?

debdutdeb: also I don't think these functions should be responsible for these validations. All callers of this should have validated then passed. Which is already happening so these checks here are unnecessary.

debdutdeb (changes requested):
comments above

packages/rest-typings/src/v1/rooms.ts
KevLehman: minLength should be enough iirc

dougfabris (approved):
LGTM!

apps/meteor/client/views/room/hooks/useOpenRoomMutation.tsx
```diff
@@ -0,0 +1,45 @@
...
+	p: '/v1/groups.open',
+	d: '/v1/im.open',
+	c: '/v1/channels.open',
+	v: '/v1/channels.open',
+	l: '/v1/channels.open',
+} as const;
+
+export const useOpenRoomMutation = ({ type }: { type: RoomType }) => {
+	const openRoom = useEndpoint('POST', openEndpoints[type]);
+	const userId = useUserId();
```
tassoevan, line 17: If the mutation depends so hardly of `userId` being not-null, consider short-cutting the hook here with an `throw` statement:

```suggestion
	const userId = useUserId();
	
	if (!userId) {
		throw new Error('error-invalid-user');
	}
```

There is no reason to give component's access to a mutation that is going to fail regardless.

apps/meteor/client/views/room/hooks/useOpenRoomMutation.tsx
```diff
@@ -0,0 +1,45 @@
...
+
+	return useMutation({
+		mutationFn: async ({ roomId }: { roomId: string }) => {
+			if (!userId) {
+				throw new Error('error-invalid-user');
```
tassoevan, line 22: Consider creating a `InvalidUserError` under `client/lib/errors`.

tassoevan (changes requested):

tassoevan (dismissed):

tassoevan (dismissed):

apps/meteor/app/api/server/v1/channels.ts
```diff
@@ -70,7 +70,7 @@ async function findChannelByIdOrName({
 		room = await Rooms.findOneByName(params.roomName || '', { projection });
 	}
 
-	if (!room || (room.t !== 'c' && room.t !== 'l')) {
+	if (!room || !['c', 'l', 'v'].includes(room.t)) {
```
MarcosSpessatto, line 73: We should not add a new value here (`v`), previously all the `channels.` endpoints were not accepting `v` as a valid value, and after this change, they will.

Can you explain why do we need to add that extra value there?

Also, if this will be used only for `.open` endpoint, it would be better to add that check there only. Otherwise we need to add tests for all the endpoints

MarcosSpessatto (changes requested):

apps/meteor/app/api/server/v1/channels.ts
aleksandernsilva: > Can you explain why do we need to add that extra value there?

Looking at the other `open` endpoints we currently have, none seem to accept `v` rooms. 
Currently we allow `v` rooms to be hidden by calling `channels.close`, it seemed logical to use `channels.open` to open the same type of room.

Initially my plan was to add a new endpoint with similar restrictions to the meteor method being replaced, but considering that we have the `.open` endpoints I was wondering if there could be a better solution using what we already have.

apps/meteor/tests/end-to-end/api/rooms.ts
```diff
@@ -3599,4 +3600,41 @@ describe('[Rooms]', () => {
...
+
+		it('should fail if roomId is not provided', async () => {
+			await request
+				.post(api('rooms.open'))
+				.set(credentials)
+				.send()
+				.expect('Content-Type', 'application/json')
+				.expect(400)
+				.expect((res: Response) => {
+					expect(res.body).to.have.property('success', false);
```
MarcosSpessatto, line 3636: Can we also call the `subscriptions.getOne` endpoint to ensure the property open is equal to true?

MarcosSpessatto (changes requested):

MarcosSpessatto (approved):

tassoevan (approved):

PR #33848, opened 2024-10-31
regression: Minimongo queries with operator matching only first found _id

gabriellsh:

There is an issue in the latest mongo release that broke some minimongo queries. When removing or updating multiple local records with a query that matches using operators ($in, $eq, etc) it would match only the first record and then bailout.

This was noticed because the `bulkMessageDelete` stream would not remove all messages provided in the query.
## Issue(s)

[CORE-786](https://rocketchat.atlassian.net/browse/CORE-786)
## Further comments

[CORE-786]: https://rocketchat.atlassian.net/browse/CORE-786?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/meteor/packages/meteor-run-as-user/lib/collection.overwrites.js
+++ b/apps/meteor/packages/meteor-run-as-user/lib/collection.overwrites.js
@@ -1,9 +1,25 @@
 // TODO: Remove this after meteor fixes the issue
 // This override fixes a recent issue introduced after the update to Meteor 3.0
 // MinimongoCollection.remove(query) where `query` has `_id: { $in: Array }` removes only the first found record.
-LocalCollection.prototype['_eachPossiblyMatchingDocSync'] = function (selector, fn) {
+LocalCollection.prototype['_eachPossiblyMatchingDocAsync'] =  async function (selector, fn) {
     const specificIds = LocalCollection._idsMatchedBySelector(selector);
-	console.log('here');
+
+    if (specificIds) {
+      for (const id of specificIds) {
+        const doc = this._docs.get(id);
+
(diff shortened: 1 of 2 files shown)


apps/meteor/packages/meteor-run-as-user/lib/collection.overwrites.js
```diff
@@ -1,3 +1,38 @@
+// TODO: Remove this after meteor fixes the issue
+// This override fixes a recent issue introduced after the update to Meteor 3.0
```
MarcosSpessatto, line 2: Was this already reported to them? If so, it would be good to add the issue here, and also the piece of code with the bug, if you were able to track it in their code...

apps/meteor/packages/meteor-run-as-user/lib/collection.overwrites.js
MarcosSpessatto, line 1: Would it be possible to add some unit tests here?

MarcosSpessatto (dismissed):

apps/meteor/packages/meteor-run-as-user/lib/collection.overwrites.js
ggazzo: https://github.com/meteor/meteor/pull/13440

apps/meteor/packages/meteor-run-as-user/lib/collection.overwrites.js
gabriellsh: It doesn't make sense for us to add unit tests for a 3rd party lib, they should have their own unit tests.

MarcosSpessatto (approved):

PR #33851, opened 2024-10-31
chore: Apply a proper logic for NotificationQueue.ts

Lao-Ax:
It's a quite minor change. But the current variant crashes the terminal in my synthetic tests.

In the current code the line `worker(counter++)` passes the previous state of the counter, I mean before the increment. So, it is literally always the same as `worker(0); counter = counter + 1`.
But it looks like that the intention was: `worker(counter + 1)` to start a new worker with next number.

## Issue(s)
[CORE-851]

[CORE-851]: https://rocketchat.atlassian.net/browse/CORE-851?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


KevLehman (approved):

sampaiodiego:
thx @Lao-Ax , good catch ;)

PR #33852, opened 2024-10-31
refactor: `pinMessage` method

KevLehman:

- Removes useless notifier call
- Removes `cloneMessage` call (this was used to keep history of the message being edited, but pinning a message shouldn't be considered a message update)
- Check if a message is pinned already (endpoint allows to pin a message infinite amount of times when called directly)
- Find user after we confirm it can do action
- Remove useless call to check for subscription (canAccessRoom already does this call with the same purpose)
## Issue(s)

Relates, but not closes: https://rocketchat.atlassian.net/browse/SUP-684
[CORE-812](https://rocketchat.atlassian.net/browse/CORE-812)
## Further comments

Before: (2.25s)
image
After: (1.68s)
image

(note: times are > 1s because DB is remote, almost 100ms ping from me)

[CORE-812]: https://rocketchat.atlassian.net/browse/CORE-812?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/meteor/app/message-pin/server/pinMessage.ts
```diff
@@ -43,108 +43,101 @@ declare module '@rocket.chat/ddp-client' {
...
+	}
 
-		let originalMessage = await Messages.findOneById(message._id);
-		if (originalMessage == null || originalMessage._id == null) {
-			throw new Meteor.Error('error-invalid-message', 'Message you are pinning was not found', {
-				method: 'pinMessage',
-				action: 'Message_pinning',
-			});
-		}
+	if (!(await hasPermissionAsync(userId, 'pin-message', originalMessage.rid))) {
```
abhinavkrin, line 58: Hey, I think we should have the check for already pinned message after this part. 
If the user does not have permission and the message is already pinned, they will get the original message which they should not.

abhinavkrin (changes requested):

abhinavkrin (approved):

MarcosSpessatto (changes requested):
Can we create a Jira task for it?

MarcosSpessatto (approved):

PR #33860, opened 2024-10-31
feat: filter unknown contacts

tapiarafael:
Return only known contacts on search endpoint and allow to search only unknown contacts

## Issue(s)

[SCI-141]

## Further comments

[SCI-141]: https://rocketchat.atlassian.net/browse/SCI-141?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


MarcosSpessatto (changes requested):
Is it possible to have a test covering it?

apps/meteor/tests/end-to-end/api/livechat/contacts.ts
```diff
@@ -788,6 +794,24 @@ describe('LIVECHAT - contacts', () => {
 			expect(res.body.offset).to.be.an('number');
 		});
 
+		it('should return only known contacts by default', async () => {
```
MarcosSpessatto, line 797: If it should return only known, wouldn't this case fail? 🤔 We are asserting it as `false`.

MarcosSpessatto (changes requested):

apps/meteor/tests/end-to-end/api/livechat/contacts.ts
tapiarafael: It's kinda confusing 😢 . We're asserting that the property `unknown` is false since we want it to be known. An `unknown: true` contact, is someone that is not known by the workspace.

MarcosSpessatto (approved):

PR #33865, opened 2024-11-01
fix: engine restart process for Deno used wrong status

d-gubert:

Whenever the Deno subprocess of an app crashes, the Apps-Engine tries to restart it so we can keep the operation going. However, the app wasn't correctly reporting its status afterwards, and stayed at the `constructed` or `initialized` status, which the engine does not recognize as enabled.

Also, we're removing the limit for the number of times the Apps-Engine would restart an app. At this point, it is not possible to show the restart or process status information to the admin, so it would just cause confusion to see the app not working.
## Issue(s)

[SUP-690]
## Steps to test or reproduce
1. Start up the server
2. Grab the `pid` of the subprocess of any enabled app (either with `ps aux | grep deno` or by starting the server with `DEBUG=appsEngine:*` and checking the logs)
3. Kill that `pid` via the command line
4. The app should still show as "enabled" in the UI and should still function as expected
## Further comments

[SUP-690]: https://rocketchat.atlassian.net/browse/SUP-690?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/apps-engine/src/server/runtime/deno/LivenessManager.ts
```diff
@@ -10,7 +10,7 @@ const defaultOptions: LivenessManager['options'] = {
     pingRequestTimeout: 10000,
     pingFrequencyInMS: 10000,
     consecutiveTimeoutLimit: 4,
-    maxRestarts: 3,
+    maxRestarts: Infinity,
```
tapiarafael, line 13: Is this safe? If an app restarted more than 10 times, should we continue trying to restart?

tapiarafael (dismissed):

packages/apps-engine/src/server/runtime/deno/LivenessManager.ts
d-gubert: Not really, but until we have a place where we can show the restart information it will just cause confusion as to why the app is not working properly. Mainly because a simple "enable" call will not make the engine create a new subprocess for the app

debdutdeb: Is there any chance of deno being "restarted" wrongly? 

I ask because we can effectively then have a tremendous number of stale processes running

debdutdeb: Maybe after a small amount of restarts the log gets pushed to all installed apps' individual logs? So we dont need a way to expose it centrally but is available through all apps

d-gubert: > Is there any chance of deno being "restarted" wrongly?

You mean like, unnecessarily? I guess it's possible, but the "restart" procedure attempts to kill the living process before spawning a new one, and if we fail killing the process we just halt. So I'd think the chances are slim

> Maybe after a small amount of restarts the log gets pushed to all installed apps'/affected app's individual log?

I can add a new warn log if the restart call happens after a threshold of 3 times

.changeset/old-coins-bow.md
```diff
@@ -0,0 +1,5 @@
+---
+'@rocket.chat/apps-engine': patch
+---
+
+Fixed an issue that would cause apps to appear disabled after a subprocess restart
```
KevLehman, line 5:
```suggestion
Fixes an issue that would cause apps to appear disabled after a subprocess restart
```

packages/apps-engine/src/server/runtime/deno/LivenessManager.ts
KevLehman: What happens if a user attempts to "enable" an app while it's being restarted? Could it cause 2 attempts of start, and 2 loops of "app restarting"?

I think we can have a special log of "app restarting" every time it happens so users can also check on them, and we as well. Cause if an app is restarting 2 times every some time, some actions may not work but since we're not passing the threshold, we won't see what's going on.

Also, i think these restart logs should (even if we don't show on the UI) have all the information needed about the process failing and a why if possible. wdyt?

KevLehman (changes requested):

.changeset/old-coins-bow.md
d-gubert: Style guide has changed :triumph:

packages/apps-engine/src/server/runtime/deno/LivenessManager.ts
d-gubert: > What happens if a user attempts to "enable" an app while it's being restarted? Could it cause 2 attempts of start, and 2 loops of "app restarting"?

If a user tries to enable an app while it's being restarted it will fail because the process is not ready to take in requests

d-gubert: I've modified the logging a bit. Please @tapiarafael @debdutdeb @KevLehman check it out again

KevLehman (approved):

debdutdeb (approved):

d-gubert:
@KevLehman @debdutdeb I'm gonna need to ask you for help with QAing this :pray:

PR #33885, opened 2024-11-05
chore: prune the token whenever we receieve a 403

Gustrb:

Whenever we try to send a push notification to a stale token, we get the SENDER_ID_MISSMATCH, and we should prune it from our database so we don't keep trying to send to those tokens.
This happens when the user has to regenerate their credentials, which had to happen in order to migrate the legacy push notification provider to FCM.
Here the approach is similar to: https://github.com/thingsboard/thingsboard/pull/10679
Related issue on that codebase: https://github.com/thingsboard/thingsboard/issues/10602#issuecomment-2085452463

## Further comments

[CONN-376](https://rocketchat.atlassian.net/browse/CONN-376)

[CONN-376]: https://rocketchat.atlassian.net/browse/CONN-376?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/meteor/app/push/server/fcm.ts
```diff
@@ -81,7 +86,9 @@ async function fetchWithRetry(url: string, _removeToken: () => void, options: Ex
 	const retryAfter = response.headers.get('retry-after');
 	const retryAfterSeconds = retryAfter ? parseInt(retryAfter, 10) : 60;
 
-	if (response.status === 404) {
+	// This could happen if the sender ID has been changed (which happens when we change from legacy to new FCM).
+	// The same approach is used here: https://github.com/thingsboard/thingsboard/pull/10679
```
diegolmello, line 90: I'd put this comment on PR description for reference, but not in code.

apps/meteor/app/push/server/fcm.ts
```diff
@@ -81,7 +86,7 @@ async function fetchWithRetry(url: string, _removeToken: () => void, options: Ex
 	const retryAfter = response.headers.get('retry-after');
 	const retryAfterSeconds = retryAfter ? parseInt(retryAfter, 10) : 60;
 
-	if (response.status === 404) {
+	if (response.status === 404 || response.status === SENDER_ID_MISMATCH_ERROR_CODE) {
```
KevLehman, line 89: Why not use an includes here and convert the 404 to a constant too?

Gustrb: Didn't do it because 404 is one of those numbers that everyone knows what means, while 403 might not make a lot of sense of why it needs to have the token removed.

But I guess there is no problem on having both as constants :)

KevLehman (approved):

PR #33892, opened 2024-11-05
test: wait for composer send button to be enabled

aleksandernsilva:


apps/meteor/tests/e2e/page-objects/fragments/home-content.ts
```diff
@@ -85,6 +85,7 @@ export class HomeContent {
 		await this.joinRoomIfNeeded();
 		await this.page.waitForSelector('[name="msg"]:not([disabled])');
 		await this.page.locator('[name="msg"]').fill(text);
+		await this.page.waitForSelector('button[aria-label="Send"]:not([disabled])');
```
ggazzo, line 88: i think click also performs `waitForSelector` inside

PR #33893, opened 2024-11-05
feat: use contact's name on as omnichannel rooms' display name

pierre-lehnen-rc:
## Issue(s)

[SCI-114](https://rocketchat.atlassian.net/browse/SCI-114)

## Further comments

[SCI-114]: https://rocketchat.atlassian.net/browse/SCI-114?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/meteor/server/models/raw/LivechatRooms.ts
+++ b/apps/meteor/server/models/raw/LivechatRooms.ts
@@ -9,6 +9,8 @@ import type {
 	ReportResult,
 	MACStats,
 	ILivechatContactVisitorAssociation,
+	ILivechatContact,
+	AtLeast,
 } from '@rocket.chat/core-typings';
 import { UserStatus } from '@rocket.chat/core-typings';
 import type { FindPaginated, ILivechatRoomsModel } from '@rocket.chat/model-typings';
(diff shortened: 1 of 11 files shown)


Gustrb (approved):

tapiarafael (approved):

apps/meteor/tests/end-to-end/api/livechat/contacts.ts
```diff
@@ -594,6 +594,65 @@ describe('LIVECHAT - contacts', () => {
...
+			await updatePermission('view-livechat-contact', ['admin']);
+		});
+
+		after(async () => {
+			await restorePermissionToRoles('view-livechat-contact');
+		});
+
+		it('should create a contact and assign it to the room', async () => {
+			const visitor = await createVisitor();
+			const room = await createLivechatRoom(visitor.token);
```
matheusbsilva137, line 608: What about using `let` (globally in this describe) and deleting the visitor and closing the room on an `afterEach` instead?
Not a big deal, though

apps/meteor/server/models/raw/LivechatRooms.ts
```diff
@@ -2761,19 +2763,47 @@ export class LivechatRoomsRaw extends BaseRaw<IOmnichannelRoom> implements ILive
...
+		contact: Partial<Pick<ILivechatContact, '_id' | 'name'>>,
+	): Promise<UpdateResult | Document> {
+		const update = {
+			...(contact._id ? { 'v.contactId': contact._id } : {}),
+			...(contact.name ? { fname: contact.name } : {}),
+		};
+
+		if (!Object.keys(update).length) {
+			throw new Error('error-invalid-operation');
+		}
```
matheusbsilva137, line 2797: It'd be nice if we had some sort of `AtLeastOne` type in `core-typings` so that this kind of error could be avoided
But also not a big deal, we can keep it like this if we need to move fast

matheusbsilva137 (changes requested):

PR #33894, opened 2024-11-05
fix: apps-engine room converter dropping contactId attribute

pierre-lehnen-rc:


Gustrb:
don't we have to add the contact to a type so I can use it on apps engine?

Gustrb (approved):

PR #33899, opened 2024-11-06
chore: improve unread badges a11y

juliajforesti:

improve accessibility of sidebar unread badges
- `role-'status'`


aleksandernsilva (dismissed):
LGTM.

One thing I noticed is that the screen reader only announces that there are unread messages (e.g "4 unread messages, 4"). Ignoring the fact the 4 is read twice, it would be nice if it told you which channel or dm the notification coming from (e.g "4 unread messages from Alek")

PR #33900, opened 2024-11-06
chore: create more page/header variations

ggazzo:


ggazzo:
> `PageBlock` doesn't mean anything to my, whats the usage proposal?

I imagined it as something more basic that should be used by others, which has margins/paddings

lucas-a-pelegrino:
> `PageBlock` doesn't mean anything to my, whats the usage proposal?

@dougfabris We kept the naming that was already in place, although I'm open to suggestions if you believe there's a name that better represents the component's purpose.

@ggazzo what do you think?

PR #33904, opened 2024-11-06
fix: missing sidebar-v2 item template rest props

juliajforesti:

- sidebar-v2 item templates were not forwarding rest props to templates


ggazzo (changes requested):
the changeset and description should contain what actually is being fixed not why/how

PR #33920, opened 2024-11-08
feat: Persistent Chat Discussion Name wildcard

MartinSchoeler:

Changes the VideoConf_Persistent_Chat_Discussion_Name to accept a wildcard to decide the position of the date in relation to the discussion name

Using the `[date]`  wildcard you can specify where the date should be in the name:
Screenshot 2024-11-08 at 14 45 31
Screenshot 2024-11-08 at 14 45 47

If no wildcard is provided the date will default to the start of the string:
Screenshot 2024-11-08 at 14 46 08
Screenshot 2024-11-08 at 14 46 16

## Issue(s)

[CORE-805](https://rocketchat.atlassian.net/browse/CORE-805)

[CORE-805]: https://rocketchat.atlassian.net/browse/CORE-805?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/meteor/tests/end-to-end/apps/video-conferences.ts
```diff
@@ -577,7 +577,8 @@ describe('Apps - Video Conferences', () => {
 							expect(res.body.room)
 								.to.have.a.property('fname')
 								.that.is.a('string')
-								.that.satisfies((msg: string) => msg.startsWith('Chat History'));
+								.that.satisfies((msg: string) => !msg.startsWith('Chat History'))
+								.that.satisfies((msg: string) => msg.includes('Chat History'));
```
abhinavkrin, line 581: Hey, if we could have a test to cover the case, when we explicitly provide the place for [date]. wdys ?

abhinavkrin (changes requested):

apps/meteor/server/services/video-conference/service.ts
```diff
@@ -1097,8 +1097,15 @@ export class VideoConfService extends ServiceClassInternal implements IVideoConf
 			return;
 		}
 
-		const name = settings.get<string>('VideoConf_Persistent_Chat_Discussion_Name') || i18n.t('Conference Call Chat History');
-		const displayName = `${name} - ${new Date().toISOString().substring(0, 10)}`;
+		const name = settings.get<string>('VideoConf_Persistent_Chat_Discussion_Name') || i18n.t('[date] Video Call Chat');
```
KevLehman, line 1100: Doesn't i18n allows to pass params? wonder if it would be better to use i18n compatible placeholders to avoid doing it manually 👀 

or was it a requirement to have it in this format?

apps/meteor/server/services/video-conference/service.ts
```diff
@@ -1097,8 +1097,15 @@ export class VideoConfService extends ServiceClassInternal implements IVideoConf
...
+		let displayName;
+		const date = new Date().toISOString().substring(0, 10);
+
+		if (name.includes('[date]')) {
```
KevLehman, line 1104: Do you need to check if included? Thought you can just do the `name.replace` which would do nothing if there's no `[date]` inside already

KevLehman (changes requested):
Add a changeset pls

apps/meteor/server/services/video-conference/service.ts
MartinSchoeler: The actual value of the setting is not a i18n string, just a normal string that the user inputs, so no i18n functions available

apps/meteor/server/services/video-conference/service.ts
MartinSchoeler: just checked to have a case to add the date to the start of the string if no date was provided

apps/meteor/server/services/video-conference/service.ts
KevLehman: Yeah, I mean, can't we treat it as an i18n string? 

if it's not possible, pls ignore this one

.changeset/silent-steaks-happen.md
```diff
@@ -0,0 +1,6 @@
+---
+"@rocket.chat/meteor": major
+"@rocket.chat/i18n": major
```
KevLehman, line 3:
```suggestion
"@rocket.chat/meteor": minor
"@rocket.chat/i18n": minor
```

KevLehman (dismissed):

abhinavkrin (dismissed):

KevLehman (approved):

PR #33923, opened 2024-11-08
chore: add more logs to verification and merging

tapiarafael:

image


apps/meteor/ee/server/patches/mergeContacts.ts
```diff
@@ -21,12 +21,15 @@ export const runMergeContacts = async (
 	if (!channel) {
 		throw new Error('error-invalid-channel');
 	}
+	logger.debug(`Getting similar contacts for contact ${contactId}`);
 	const similarContacts: ILivechatContact[] = await LivechatContacts.findSimilarVerifiedContacts(channel, contactId);
 
 	if (!similarContacts.length) {
+		logger.debug(`No similar contacts found for contact ${contactId}`);
```
KevLehman, line 28: Wdyt of using object way of logging instead?

```
{
msg: 'no similar contacts found',
contactId,
}
```
And the same for the other logs?

tapiarafael: Wouldn't this make it harder to read logs that contain multiple IDs? Like in the verification file where we show the contact id, visitor, and room

Gustrb (approved):
LGTM

PR #33939, opened 2024-11-11
regression: Update and notify subscriptions on contact name update

matheusbsilva137:

 - Update subscriptions and inquiries when a contact name is updated;
 - Notify changes whenever a contact name is updated (so that the UI is updated accordingly without having to refresh the page).

## Further comments

[SCI-160]

[SCI-160]: https://rocketchat.atlassian.net/browse/SCI-160?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/meteor/app/lib/server/lib/notifyListener.ts
```diff
@@ -87,6 +89,18 @@ export const notifyOnRoomChangedByUsernamesOrUids = withDbWatcherCheck(
 	},
 );
 
+export const notifyOnRoomChangedByContactId = withDbWatcherCheck(
+	async <T extends IOmnichannelRoom>(
+		contactId: Exclude<T['v']['contactId'], undefined>,
+		clientAction: ClientAction = 'updated',
+	): Promise<void> => {
+		const items = LivechatRooms.findByContactId(contactId);
```
KevLehman, line 97: Should this be open rooms only?

apps/meteor/app/lib/server/lib/notifyListener.ts
```diff
@@ -553,6 +567,19 @@ export const notifyOnSubscriptionChangedByUserIdAndRoomType = withDbWatcherCheck
 	},
 );
 
+export const notifyOnSubscriptionChangedByVisitorIds = withDbWatcherCheck(
+	async (
+		visitorIds: Exclude<ISubscription['v'], undefined>['_id'][],
+		clientAction: Exclude<ClientAction, 'removed'> = 'updated',
+	): Promise<void> => {
+		const cursor = Subscriptions.findByVisitorIds(visitorIds, { projection: subscriptionFields });
```
KevLehman, line 575: Same

apps/meteor/server/models/raw/LivechatRooms.ts
```diff
@@ -2806,4 +2807,8 @@ export class LivechatRoomsRaw extends BaseRaw<IOmnichannelRoom> implements ILive
 			},
 		);
 	}
+
+	findByContactId(contactId: ILivechatContact['_id'], options?: FindOptions<IOmnichannelRoom>): FindCursor<IOmnichannelRoom> {
+		return this.find({ 't': 'l', 'v.contactId': contactId }, options);
```
KevLehman, line 2812: The model addsthe `t` filter by default

KevLehman (changes requested):

apps/meteor/app/lib/server/lib/notifyListener.ts
matheusbsilva137: Hmmm right, that makes more sense. Changed it

apps/meteor/app/lib/server/lib/notifyListener.ts
```diff
@@ -87,6 +89,18 @@ export const notifyOnRoomChangedByUsernamesOrUids = withDbWatcherCheck(
 	},
 );
 
+export const notifyOnRoomChangedByContactId = withDbWatcherCheck(
+	async <T extends IOmnichannelRoom>(
+		contactId: Exclude<T['v']['contactId'], undefined>,
+		clientAction: ClientAction = 'updated',
+	): Promise<void> => {
+		const items = LivechatRooms.findOpenByContactId(contactId);
+		for await (const item of items) {
```
MarcosSpessatto, line 98: Any reason for the `for await` here? 🤔

apps/meteor/server/models/raw/Subscriptions.ts
```diff
@@ -594,6 +605,21 @@ export class SubscriptionsRaw extends BaseRaw<ISubscription> implements ISubscri
 		return this.updateMany(query, update);
 	}
 
+	updateNameAndFnameByVisitorIds(visitorIds: string[], name: string): Promise<UpdateResult | Document> {
+		const query = { 't': 'l' as const, 'v._id': { $in: visitorIds } };
+
+		const update = {
+			$set: {
+				name,
+				lowerCaseName: name.toLowerCase(),
```
MarcosSpessatto, line 614: Any reason for this new property? Didn't get it...

apps/meteor/server/models/raw/Subscriptions.ts
```diff
@@ -594,6 +605,21 @@ export class SubscriptionsRaw extends BaseRaw<ISubscription> implements ISubscri
...
+				fname: name,
+				lowerCaseFName: name.toLowerCase(),
```
MarcosSpessatto, line 616: Any reason for this new property? Didn't get it...

MarcosSpessatto (changes requested):

apps/meteor/server/models/raw/Subscriptions.ts
matheusbsilva137: Hmmm I just noticed this isn't in `core-typings`, but we actually have this prop in the DB due to [this line](https://github.com/RocketChat/Rocket.Chat/blob/92b6a50c343948b48415fe81de0b1f601aa02979/apps/meteor/app/livechat/server/lib/Helper.ts#L265) and we rely on it on the client side
@KevLehman do you recall what was the idea for this? And why we didn't include it in core-typings? :thinking: 

But I believe this prop should be updated even though it isn't in `core-typings` so that we can keep consistency (and maybe add it to `core-typings` in another PR)

apps/meteor/server/models/raw/Subscriptions.ts
matheusbsilva137: (same as in the other comment)

apps/meteor/server/models/raw/Subscriptions.ts
KevLehman: Can you create a task to remove it from the `createSubscription` call as well? 🤗 so we remove it on the future

matheusbsilva137: Done, @KevLehman ! [CORE-809](https://rocketchat.atlassian.net/browse/CORE-809)

apps/meteor/app/lib/server/lib/notifyListener.ts
```diff
@@ -87,6 +89,19 @@ export const notifyOnRoomChangedByUsernamesOrUids = withDbWatcherCheck(
 	},
 );
 
+export const notifyOnRoomChangedByContactId = withDbWatcherCheck(
+	async <T extends IOmnichannelRoom>(
+		contactId: Exclude<T['contactId'], undefined>,
```
KevLehman, line 94: Do you actually need the generic here? I see this more as an `string`, or even fancier, `ILivechatContact['_id']`

wdyt?

apps/meteor/app/lib/server/lib/notifyListener.ts
```diff
@@ -251,6 +263,20 @@ export const notifyOnLivechatInquiryChangedById = withDbWatcherCheck(
 	},
 );
 
+export const notifyOnLivechatInquiryChangedByVisitorIds = withDbWatcherCheck(
+	async (
+		visitorIds: ILivechatInquiryRecord['v']['_id'][],
+		clientAction: Exclude<ClientAction, 'removed'> = 'updated',
+		diff?: Partial<Record<keyof ILivechatInquiryRecord, unknown> & { queuedAt: unknown; takenAt: unknown }>,
```
KevLehman, line 270: both are Dates, do we need the `unknown`?

apps/meteor/server/models/raw/Subscriptions.ts
```diff
@@ -341,6 +342,16 @@ export class SubscriptionsRaw extends BaseRaw<ISubscription> implements ISubscri
 		return this.find(query, options || {});
 	}
 
+	findOpenByVisitorIds(visitorIds: string[], options?: FindOptions<ISubscription>): FindCursor<ISubscription> {
+		const query = {
+			't': 'l' as const,
```
KevLehman, line 347: For this case, i think would be better to have an idx on `open + v._id` as rooms having `v._id` are for sure type `t`. So the index may not be exclusive enough.

wanna give it a try?

apps/meteor/server/models/raw/Subscriptions.ts
```diff
@@ -594,6 +605,19 @@ export class SubscriptionsRaw extends BaseRaw<ISubscription> implements ISubscri
 		return this.updateMany(query, update);
 	}
 
+	updateNameAndFnameByVisitorIds(visitorIds: string[], name: string): Promise<UpdateResult | Document> {
+		const query = { 't': 'l' as const, 'v._id': { $in: visitorIds } };
```
KevLehman, line 609: You can remove the `t` part from here, don't think mongo gets any extra improvements with it.

apps/meteor/app/lib/server/lib/notifyListener.ts
matheusbsilva137: We're typing these 2 props as unknown everywhere in this file. Maybe @ricardogarim can tell us why :thinking: or if it's ok to provide a more specific typing

dougfabris (changes requested):

KevLehman (dismissed):

MarcosSpessatto (approved):

PR #33949, opened 2024-11-13
feat: disable uploading of exempt apps

Gustrb:

Previously community workspaces had the hard limit of 3 private apps, with the private apps restriction, it was defined that community workspaces would not have access to any private apps.
The apps that were already in the workspace, are called exempted.

This PR makes it so that exempt apps cannot be updated, since that would render all the effort of removing community workspaces private apps useless, because the user would be able to just upload the same app while changing everything about it 😢 

The 'expect updated private app in CE to be kept as disabled' test was removed since this behavior was changed and now we cannot test that in the CI, since there is no way to upgrade/downgrade it

## Further comments

[CONN-381](https://rocketchat.atlassian.net/browse/CONN-381)
[CONN-379](https://rocketchat.atlassian.net/browse/CONN-379)

[CONN-381]: https://rocketchat.atlassian.net/browse/CONN-381?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


aleksandernsilva (changes requested):
If possible I'd recommend at least implementing unit tests to cover this scenario.

apps/meteor/client/views/marketplace/AppExemptModal.tsx
```diff
@@ -0,0 +1,56 @@
...
+				<Modal.FooterControllers>
+					<Button onClick={onCancel}>{t('Cancel')}</Button>
+					<Button onClick={goToManageSubscriptionPage} primary>
+						{t('Upgrade')}
+					</Button>
+				</Modal.FooterControllers>
+			</Modal.Footer>
+		</Modal>
+	);
+};
```
hugocostadev, line 54: Could you use `<GenericModal />` here?

apps/meteor/ee/server/apps/communication/rest.ts
```diff
@@ -748,6 +749,17 @@ export class AppsRestApi {
...
+
+					// Note: exempt apps happen when a private app was uploaded to a community workspace before
+					//       the private app restriction was enforced. We still allow the users to use their
+					//       exempt apps, but they can't update them, since they could just upload a new version
+					//       containing a totally different app under the same id :(
+					const isExemptApp = isPrivateAppUpload && isCommunityWorkspace;
+					if (isExemptApp) {
+						return API.v1.failure({ error: 'Cannot_Update_Exempt_App' });
+					}
+
```
hugocostadev, line 762: add tests, please

apps/meteor/client/views/marketplace/AppExemptModal.tsx
Gustrb: Im not familiar with it, but probably, I'll check if it matches the design

apps/meteor/client/views/marketplace/AppExemptModal.tsx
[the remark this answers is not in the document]
Gustrb: do you think we have to test this component?
it barely has any logic, I was thinking about testing the hook, don't think testing this would give any confidence
since we can't write UI tests for this feature

Gustrb:
Hey guys, since this task is really important to get into the release and I have zero familiarity with testing in the frontend, I think that would end up blocking the merge of this simple task, I am thinking about creating a task that adresses this issues later, wdyt @aleksandernsilva @hugocostadev

apps/meteor/client/views/marketplace/AppExemptModal.tsx
Gustrb: I am not sure we can do that, since there is no way to add footer annotation in the generic modal

hugocostadev (dismissed):

aleksandernsilva (approved):

PR #33962, opened 2024-11-14
chore: Discussion message action meteor removal

MartinSchoeler:

Moves the discussion message action to a hook

## Further comments

[ARCH-1348]

[ARCH-1348]: https://rocketchat.atlassian.net/browse/ARCH-1348?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/meteor/client/components/message/toolbar/useNewDiscussionMessageAction.tsx
```diff
@@ -1,35 +1,35 @@
...
 
-Meteor.startup(() => {
-	Tracker.autorun(() => {
-		if (!settings.get('Discussion_enabled')) {
+export const useNewDiscussionMessageAction = () => {
+	const enabled = useSetting('Discussion_enabled', false);
+
+	const setModal = useSetModal();
+
+	useEffect(() => {
```
aleksandernsilva, line 14: Might be a good idea to also remove the button (`MessageAction.removeButton`) in this effect's clean up function.

aleksandernsilva:
Could you also add some tests to cover this scenario?

dougfabris (changes requested):
I think the proper fix for the issue should be use the `isLoading` from the `createDiscussionMutation` to set the `loading` while clicking in the create button. The code you did seems a good refactor so we could handle separately under the initiative we want to start. 

Regarding tests, we already have a e2e test for this. I think it's just missing the assertion to guarantee, that only one discussion was created

MartinSchoeler:
Split the fix and the chore, fix moved to https://github.com/RocketChat/Rocket.Chat/pull/33985

apps/meteor/client/components/message/toolbar/useNewDiscussionMessageAction.tsx
MartinSchoeler: good catch, done!

MartinSchoeler:
> Split the fix and the chore, fix moved to #33985

Tests were done in https://github.com/RocketChat/Rocket.Chat/pull/33985

dougfabris (approved):
LGTM!

aleksandernsilva (approved):

PR #33963, opened 2024-11-14
fix: Thread unread not in unread group

juliajforesti:

There's an edge case where a room with thread unread is not being added to the unread group because `alert` is false. 
This PR adds `tunread` to the condition of adding the room to `unread` list

### DIFF
--- a/.changeset/sweet-needles-melt.md
+++ b/.changeset/sweet-needles-melt.md
@@ -1,0 +1,5 @@
+---
+'@rocket.chat/meteor': patch
+---
+
+Fixes edge case of thread unread not being added to unread group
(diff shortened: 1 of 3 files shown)


mors1eska:
Could you please let me know if my request on the forum is similar to what you have addressed in your issue?
Thank you in advance for your response

https://forums.rocket.chat/t/cant-set-up-notifications/15714/10

juliajforesti:
> Could you please let me know if my request on the forum is similar to what you have addressed in your issue? Thank you in advance for your response
> 
> [forums.rocket.chat/t/cant-set-up-notifications/15714/10](https://forums.rocket.chat/t/cant-set-up-notifications/15714/10)

Hey! I replied to you on the forum :)

PR #33981, opened 2024-11-18
chore: Move `notify-user` `/notification` and `/subscriptions-changed` stream events to hook

tiagoevanp:

Move user notifications `/notification` and `subscriptions-changed` stream events to a react hook

## Further comments

[CORE-677](https://rocketchat.atlassian.net/browse/CORE-677?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ)

[CORE-677]: https://rocketchat.atlassian.net/browse/CORE-677?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


dougfabris (approved):

apps/meteor/client/hooks/useNotifyUser.ts
```diff
@@ -0,0 +1,92 @@
...
+			const hasFocus = document.hasFocus();
+			const messageIsInOpenedRoom = openedRoomId === notification.payload.rid;
+
+			fireGlobalEvent('notification', {
+				notification,
+				fromOpenedRoom: messageIsInOpenedRoom,
+				hasFocus,
+			});
+
+			if (router.getSearchParameters().layout === 'embedded') {
```
dougfabris, line 65: should we use our util function from `client/lib` ?

dougfabris (changes requested):

apps/meteor/client/hooks/useNotifyUser.ts
```diff
@@ -0,0 +1,92 @@
...
+		notifyUserStream(`${user?._id}/subscriptions-changed`, (action, sub) => {
+			if (action === 'removed') {
+				return;
+			}
+
+			void notifyNewRoom(sub);
+		});
+	}, [notifyNewMessageAudio, notifyNewRoom, notifyUserStream, router, user?._id]);
+
+	CachedChatSubscription.collection.find().observe({
```
dougfabris, line 87: previously this was called inside the `Tracker.autorun` this change is really intended?

dougfabris (changes requested):

apps/meteor/client/hooks/useNotifyUser.ts
tiagoevanp: this router is from useRouter hook 🤔

apps/meteor/client/hooks/useNotifyUser.ts
tiagoevanp: Since we are migrating Meteor code from React, yes... `Tracker.autorun` works pretty similar to a React hook with useEffect on it...

apps/meteor/client/hooks/useNotifyUser.ts
dougfabris: I don't get it. I wanted to mean, above this function we're using the [util function](https://github.com/RocketChat/Rocket.Chat/pull/33981/files/5ac85c030d1fa7ebc1fc810995e5b556b5f1544c#diff-1f3fe2aee19925b43ea380028dfaeb093aad014738ba744f820c30fb8df14873R35), why can't we use it again?

tiagoevanp: ouch, I passed by line 35 without noticed... The router used in this lib is a deprecated way of usage

apps/meteor/client/hooks/useNotifyUser.ts
tiagoevanp: you are right, fixed!

PR #33991, opened 2024-11-19
chore: improve indexes of the livechat contacts collection

pierre-lehnen-rc:

### DIFF
--- a/apps/meteor/server/models/raw/LivechatContacts.ts
+++ b/apps/meteor/server/models/raw/LivechatContacts.ts
@@ -58,12 +58,6 @@ export class LivechatContactsRaw extends BaseRaw<ILivechatContact> implements IL
 				},
 				unique: false,
 			},
-			{
-				key: {
-					channels: 1,
-				},
-				unique: false,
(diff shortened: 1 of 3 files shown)


KevLehman (dismissed):

apps/meteor/server/models/raw/LivechatContacts.ts
```diff
@@ -56,12 +56,24 @@ export class LivechatContactsRaw extends BaseRaw<ILivechatContact> implements IL
...
 				},
+				partialFilterExpression: { 'channels.verified': true },
+				name: 'verificationKey',
+				unique: false,
+			},
+			{
+				key: {
+					preRegistration: 1,
+				},
+				name: 'preRegistration',
```
sampaiodiego, line 76: I guess we can make this index `sparse`, since there will be less records with it set.. right?
```suggestion
				sparse: true,
```

sampaiodiego (changes requested):

sampaiodiego (approved):

PR #33995, opened 2024-11-19
refactor: assets methods

ggazzo:
backport from https://github.com/RocketChat/Rocket.Chat/pull/33935

### DIFF
--- a/apps/meteor/app/api/server/v1/assets.ts
+++ b/apps/meteor/app/api/server/v1/assets.ts
@@ -1,13 +1,18 @@
+import { Settings } from '@rocket.chat/models';
 import { isAssetsUnsetAssetProps } from '@rocket.chat/rest-typings';
 
-import { RocketChatAssets, setAsset, unsetAsset, refreshClients } from '../../../assets/server';
+import { RocketChatAssets, refreshClients } from '../../../assets/server';
+import { notifyOnSettingChangedById } from '../../../lib/server/lib/notifyListener';
 import { settings } from '../../../settings/server';
 import { API } from '../api';
 import { getUploadFormData } from '../lib/getUploadFormData';
 
 API.v1.addRoute(
 	'assets.setAsset',
-	{ authRequired: true },
+	{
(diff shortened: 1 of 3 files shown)


apps/meteor/app/api/server/v1/assets.ts
```diff
@@ -1,13 +1,18 @@
...
 import { settings } from '../../../settings/server';
 import { API } from '../api';
 import { getUploadFormData } from '../lib/getUploadFormData';
 
 API.v1.addRoute(
 	'assets.setAsset',
-	{ authRequired: true },
+	{
+		authRequired: true,
+		permissionsRequired: ['manage-assets'],
```
MarcosSpessatto, line 14: Would it be possible to add the (unused) validation `isAssetsSetAssetProps`?

apps/meteor/app/assets/server/assets.ts
```diff
@@ -260,29 +258,29 @@ class RocketChatAssetsClass {
...
+					});
+				}, 200);
+			});
+
+			rs.pipe(ws);
+		});
 	}
 
-	public async unsetAsset(asset: string): Promise<void> {
+	public async unsetAsset(asset: string) {
```
MarcosSpessatto, line 283: Can we have the return typed here?

MarcosSpessatto (changes requested):

apps/meteor/app/assets/server/assets.ts
ggazzo: I don't see much reason to type a return if it can be inferred automatically, unless I was working with an interface and wanted to enforce a contract(which is not the case here).

apps/meteor/app/api/server/v1/assets.ts
ggazzo: I'm afraid not, in this case the body type is not a json, and needs to be built first, with our current technology our parser is not smart enough to do this for us :/

MarcosSpessatto (approved):

PR #34004, opened 2024-11-19
feat: keep VoIP history

pierre-lehnen-rc:
## Issue(s)

[Core-775](https://rocketchat.atlassian.net/browse/CORE-775)


apps/meteor/server/models/raw/FreeSwitchChannel.ts
```diff
@@ -0,0 +1,67 @@
...
+export class FreeSwitchChannelRaw extends BaseRaw<IFreeSwitchChannel> implements IFreeSwitchChannelModel {
+	constructor(db: Db, trash?: Collection<RocketChatRecordDeleted<IFreeSwitchChannel>>) {
+		super(db, 'freeswitch_channels', trash);
+	}
+
+	protected modelIndexes(): IndexDescription[] {
+		return [
+			{ key: { uniqueId: 1 }, unique: true },
+			{ key: { otherLegUniqueId: 1 }, unique: false },
+			{ key: { callId: 1 }, unique: false },
```
sampaiodiego, line 16: from the query patterns, looks like this can be a single index:
```suggestion
			{ key: { uniqueId: 1, otherLegUniqueId: 1, callId: 1 }, unique: true },
```

sampaiodiego (changes requested):

apps/meteor/ee/server/local-services/voip-freeswitch/service.ts
```diff
@@ -1,27 +1,53 @@
...
+import { FreeSwitchCall, FreeSwitchEvent, Users } from '@rocket.chat/models';
+import { objectMap, wrapExceptions } from '@rocket.chat/tools';
+import type { WithoutId } from 'mongodb';
+
+import { settings } from '../../../../app/settings/server';
 
 export class VoipFreeSwitchService extends ServiceClassInternal implements IVoipFreeSwitchService {
 	protected name = 'voip-freeswitch';
 
-	constructor(private getSetting: <T extends SettingValue = SettingValue>(id: ISetting['_id']) => T) {
```
MarcosSpessatto, line 8: Any reason for stopping injecting the setting-getter here? 🤔 
Also, if this is not necessary for the feature itself, we could keep it in another PR.

apps/meteor/ee/server/local-services/voip-freeswitch/service.ts
```diff
@@ -31,6 +57,451 @@ export class VoipFreeSwitchService extends ServiceClassInternal implements IVoip
...
+			callState,
+			timestamp,
+			firedAt,
+			answerState,
+			hangupCause,
+			referencedIds,
+			receivedAt: new Date(),
+			channelName,
+			direction,
+
```
MarcosSpessatto, line 244:
```suggestion
```

apps/meteor/ee/server/local-services/voip-freeswitch/service.ts
```diff
@@ -31,6 +57,451 @@ export class VoipFreeSwitchService extends ServiceClassInternal implements IVoip
...
+			caller,
+			call,
+
```
MarcosSpessatto, line 247:
```suggestion
```

packages/tools/src/convertSubObjectsIntoPaths.ts
```diff
@@ -0,0 +1,16 @@
+export function convertSubObjectsIntoPaths(object: Record<string, any>, parentPath?: string): Record<string, any> {
```
MarcosSpessatto, line 1: Can we have a unit test for this?

packages/tools/src/objectMap.ts
MarcosSpessatto, line 1: Can we have a unit test for this?

MarcosSpessatto (changes requested):

MarcosSpessatto (approved):

apps/meteor/server/models/raw/FreeSwitchEvent.ts
```diff
@@ -0,0 +1,45 @@
...
+import { BaseRaw } from './BaseRaw';
+
+export class FreeSwitchEventRaw extends BaseRaw<IFreeSwitchEvent> implements IFreeSwitchEventModel {
+	constructor(db: Db, trash?: Collection<RocketChatRecordDeleted<IFreeSwitchEvent>>) {
+		super(db, 'freeswitch_events', trash);
+	}
+
+	protected modelIndexes(): IndexDescription[] {
+		return [
+			{ key: { channelUniqueId: 1, sequence: 1 }, unique: false },
```
sampaiodiego, line 14: based on the `registerEvent` method, it looks like the idea is that these fields are unique.
```suggestion
			{ key: { channelUniqueId: 1, sequence: 1 }, unique: true },
```

apps/meteor/server/models/raw/FreeSwitchEvent.ts
```diff
@@ -0,0 +1,45 @@
...
+	}
+
+	public async registerEvent(event: WithoutId<InsertionModel<IFreeSwitchEvent>>): Promise<void> {
+		const { channelUniqueId, sequence } = event;
+
+		if (channelUniqueId && sequence && (await this.exists({ channelUniqueId, sequence }))) {
+			return;
+		}
+
+		await this.insertOne(event);
```
sampaiodiego, line 27: making the index _unique_, this check can be removed.
```suggestion
		await this.insertOne(event);
```

apps/meteor/server/models/raw/FreeSwitchEvent.ts
```diff
@@ -0,0 +1,45 @@
...
+	}
+
+	public findAllByCallUUID<T extends IFreeSwitchEvent>(callUUID: string, options?: FindOptions<IFreeSwitchEvent>): FindCursor<T> {
+		return this.find<T>({ 'call.UUID': callUUID }, options);
```
sampaiodiego, line 31: missing index for this query

apps/meteor/server/models/raw/VideoConference.ts
```diff
@@ -136,6 +137,13 @@ export class VideoConferenceRaw extends BaseRaw<VideoConference> implements IVid
 		return (await this.insertOne(call)).insertedId;
 	}
 
+	public async createVoIP(call: InsertionModel<IVoIPVideoConference>): Promise<string | undefined> {
+		const { externalId, ...data } = call;
+
+		const doc = await this.findOneAndUpdate({ externalId }, { $set: data }, { upsert: true, returnDocument: 'after' });
+		return doc.value?._id;
```
sampaiodiego, line 144: if we are returning the `_id` only, why not use `insertOne` and return the `insertedId`

sampaiodiego (changes requested):
should I receive this error on console when starting up rocket.chat with nothing configured?

```bash
@rocket.chat/meteor:dsv: W20241219-14:48:59.026(-3)? (STDERR) Error: VoIP is disabled.
@rocket.chat/meteor:dsv: W20241219-14:48:59.026(-3)? (STDERR)     at VoipFreeSwitchService.getConnectionSettings (ee/server/local-services/voip-freeswitch/service.ts:40:10)
@rocket.chat/meteor:dsv: W20241219-14:48:59.026(-3)? (STDERR)     at VoipFreeSwitchService.started (ee/server/local-services/voip-freeswitch/service.ts:31:10)
@rocket.chat/meteor:dsv: W20241219-14:48:59.026(-3)? (STDERR)     at /Users/diegosampaio/dev/Rocket.Chat/packages/core-services/src/LocalBroker.ts:115:65
@rocket.chat/meteor:dsv: W20241219-14:48:59.026(-3)? (STDERR)     at Array.map (<anonymous>)
@rocket.chat/meteor:dsv: W20241219-14:48:59.026(-3)? (STDERR)     at LocalBroker.start (/Users/diegosampaio/dev/Rocket.Chat/packages/core-services/src/LocalBroker.ts:115:40)
@rocket.chat/meteor:dsv: W20241219-14:48:59.026(-3)? (STDERR)     at Api.start (/Users/diegosampaio/dev/Rocket.Chat/packages/core-services/src/lib/Api.ts:75:21)
@rocket.chat/meteor:dsv: W20241219-14:48:59.026(-3)? (STDERR)     at module.wrapAsync.self (server/startup/localServices.ts:12:10)
@rocket.chat/meteor:dsv: W20241219-14:48:59.026(-3)? (STDERR)     at Module.wrapAsync (/Users/diegosampaio/.meteor/packages/modules/.0.20.2.1rksah5.q497++os+web.browser+web.browser.legacy+web.cordova/npm/node_modules/@meteorjs/reify/lib/runtime/index.js:252:8)
@rocket.chat/meteor:dsv: W20241219-14:48:59.026(-3)? (STDERR)     at module (/Users/diegosampaio/dev/Rocket.Chat/apps/meteor/.meteor/local/build/programs/server/app/app.js:241370:9)
@rocket.chat/meteor:dsv: W20241219-14:48:59.026(-3)? (STDERR)     at fileEvaluate (packages/modules-runtime.js:335:7)
@rocket.chat/meteor:dsv: W20241219-14:48:59.026(-3)? (STDERR)     at Module.require (packages/modules-runtime.js:237:14)
@rocket.chat/meteor:dsv: W20241219-14:48:59.026(-3)? (STDERR)     at Module.mod.require (/Users/diegosampaio/.meteor/packages/modules/.0.20.2.1rksah5.q497++os+web.browser+web.browser.legacy+web.cordova/npm/node_modules/@meteorjs/reify/lib/runtime/index.js:30:33)
@rocket.chat/meteor:dsv: W20241219-14:48:59.026(-3)? (STDERR)     at require (packages/modules-runtime.js:257:21)
@rocket.chat/meteor:dsv: W20241219-14:48:59.026(-3)? (STDERR)     at startup (server/startup/index.ts:20:3)
@rocket.chat/meteor:dsv: W20241219-14:48:59.026(-3)? (STDERR)     at processTicksAndRejections (node:internal/process/task_queues:95:5)
@rocket.chat/meteor:dsv: W20241219-14:48:59.027(-3)? (STDERR)     at async Promise.all (index 3)
@rocket.chat/meteor:dsv: W20241219-14:48:59.027(-3)? (STDERR)     at module.wrapAsync.self (server/main.ts:28:1)
```

apps/meteor/server/models/raw/FreeSwitchEvent.ts
KevLehman: Since events from FS are a bit weird, think the `unique` false are there because some events may have same channel + seq and be different 🙈

KevLehman: But let's try, worst case a try catch may be enough.

sampaiodiego: but the code below is _ignoring_ events with the same `channelUniqueId` and `sequence`: 
https://github.com/RocketChat/Rocket.Chat/blob/0d685216e30d93fe10a77f7b7b0b5e5b183f8ea8/apps/meteor/server/models/raw/FreeSwitchEvent.ts#L23-L25

so that is basically an "unique index logic", but it doesn't make sense taking that responsibility if the database can take of that.. also moving that responsibility to mongo is safer.

apps/meteor/server/models/raw/BaseRaw.ts
```diff
@@ -231,6 +231,10 @@ export abstract class BaseRaw<
 		};
 	}
 
+	exists(query: Filter<T>): Promise<boolean> {
+		return this.col.find(query, { limit: 1, batchSize: 1, projection: { _id: 1 } }).hasNext();
```
sampaiodiego, line 235: just curious, what is the motivation behind this instead of using a `findOne`?

apps/meteor/server/models/raw/VideoConference.ts
KevLehman: yeah but in this case `externalId` is not unique, so i assume this was to provide some sort of uniqueness without making it actually unique (cause then other videoconf documents with no externalId would complain).

Was planning to use a partialFilterExp to make the unique only when externalId was present, but as i'm not very familiar with videoconf decided to leave that improvement for later. wdyt?

apps/meteor/server/models/raw/BaseRaw.ts
KevLehman: Perf maybe? Not sure. At least hasNext dont return any data so it should be a bit nicer on network. But since we're removing the only usage of it i'll remove this as well.

apps/meteor/ee/server/local-services/voip-freeswitch/service.ts
```diff
@@ -33,6 +83,490 @@ export class VoipFreeSwitchService extends ServiceClassInternal implements IVoip
...
+			if (event.eventName === 'CHANNEL_DESTROY' && event.call?.UUID) {
+				await this.computeCall(event.call?.UUID);
+			}
+		} catch (error) {
+			// avoid logging that an event was duplicated from mongo
+			if (error instanceof MongoError && error.code === 11000) {
+				return;
+			}
+
+			console.error(error);
```
MarcosSpessatto, line 302: Shouldn't we use a logger instead?

apps/meteor/ee/server/local-services/voip-freeswitch/service.ts
```diff
@@ -33,6 +83,490 @@ export class VoipFreeSwitchService extends ServiceClassInternal implements IVoip
...
+		if (!timestamp || timestamp === '0') {
+			return undefined;
+		}
+
+		const value = parseInt(timestamp);
+		if (Number.isNaN(value)) {
+			return undefined;
+		}
+
+		const timeValue = Math.floor(value / 1000);
```
MarcosSpessatto, line 286: Maybe use toMillis?

MarcosSpessatto (changes requested):

apps/meteor/server/services/video-conference/service.ts
```diff
@@ -454,6 +461,16 @@ export class VideoConfService extends ServiceClassInternal implements IVideoConf
 		return true;
 	}
 
+	public async createVoIP(data: InsertionModel<IVoIPVideoConference>): Promise<IVoIPVideoConference['_id'] | undefined> {
+		return wrapExceptions(async () => VideoConferenceModel.createVoIP(data)).catch((e) => {
+			logger.error({
+				name: 'Error on VideoConf.createVoIP',
+				error: e,
```
sampaiodiego, line 468:
```suggestion
				err: e,
```

apps/meteor/ee/server/local-services/voip-freeswitch/service.ts
sampaiodiego: or just throw again

apps/meteor/ee/server/local-services/voip-freeswitch/service.ts
```diff
@@ -1,14 +1,64 @@
...
 	constructor() {
 		super();
+
+		this.serviceStarter = new ServiceStarter(() => Promise.resolve(this.startEvents()));
+	}
+
+	private listening = false;
+
+	public async started(): Promise<void> {
+		this.onEvent('watch.settings', async ({ setting }): Promise<void> => {
```
sampaiodiego, line 37: `this.onEvent` should be in `constructor` otherwise it won't work as expected..

sampaiodiego (changes requested):

apps/meteor/ee/server/local-services/voip-freeswitch/service.ts
KevLehman: For my curiosity, what's the difference?

sampaiodiego: it has to do on how the `LocalBroker` reads the service instance to register all it's events and methods:

- the constructor is called when you calls `registerService`
- the `LocalBroker` calls `created`
- it gets all registered events
- it gets all methods
- and then it calls `started`

https://github.com/RocketChat/Rocket.Chat/blob/f90b1c99bdf4d89350a70c9ef4c2d12ed3784fd8/packages/core-services/src/LocalBroker.ts#L59-L83

sampaiodiego (approved):

PR #34009, opened 2024-11-20
fix(messages): support multi-parameter sorting for IM/DM messages

### LINKED ISSUE #34008: IM/DM messages sorting relies only on ts, unlike group messages
### Description:

Sorting for IM/DM messages currently uses only the ts (timestamp) field, whereas group messages support sorting by multiple parameters. This inconsistency limits flexibility and prevents custom sorting options for direct and instant messages.

### Steps to reproduce:

1. Query for IM/DM messages using the Rocket.Chat API or interface.
2. Attempt to use sorting parameters (e.g., by sender or custom fields).
3. Observe that the sorting relies solely on the ts timestamp.

### Expected behavior:

IM/DM messages should support sorting by multiple parameters, similar to group messages, to allow consistent and customizable sorting behavior.

### Actual behavior:

IM/DM messages are sorted only by the ts timestamp, regardless of the specified parameters.

### Additional context

This issue creates inconsistency between group and IM/DM sorting functionality. Fixing this will align the behavior across all message types.

MahmoudMowiena:
Previously, the sorting for IM/DM messages relied only on the `ts` timestamp. This fix updates the logic to allow sorting by multiple parameters, similar to group messages, enhancing consistency across message types.

Closes #34008
[CORE-830]

[CORE-830]: https://rocketchat.atlassian.net/browse/CORE-830?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/meteor/app/api/server/v1/im.ts
+++ b/apps/meteor/app/api/server/v1/im.ts
@@ -400,9 +400,10 @@ API.v1.addRoute(
 				...parseIds(starredIds, 'starred._id'),
 				...(pinned && pinned.toLowerCase() === 'true' ? { pinned: true } : {}),
 			};
+			const sortObj = sort?.ts ?? { ts: -1 };
 
 			const { cursor, totalCount } = Messages.findPaginated(ourQuery, {
-				sort: sort,
+				sort: sortObj,
 				skip: offset,
 				limit: count,
 				...(fields && { projection: fields }),
--- a/apps/meteor/app/api/server/v1/im.ts
+++ b/apps/meteor/app/api/server/v1/im.ts
@@ -400,10 +400,9 @@ API.v1.addRoute(
 				...parseIds(starredIds, 'starred._id'),
 				...(pinned && pinned.toLowerCase() === 'true' ? { pinned: true } : {}),
 			};
-			const sortObj = { ts: sort?.ts ?? -1 };
 
 			const { cursor, totalCount } = Messages.findPaginated(ourQuery, {
-				sort: sortObj,
+				sort: sort,
 				skip: offset,
 				limit: count,
 				...(fields && { projection: fields }),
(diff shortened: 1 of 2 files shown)


apps/meteor/app/api/server/v1/im.ts
```diff
@@ -400,10 +400,9 @@ API.v1.addRoute(
 				...parseIds(starredIds, 'starred._id'),
 				...(pinned && pinned.toLowerCase() === 'true' ? { pinned: true } : {}),
 			};
-			const sortObj = { ts: sort?.ts ?? -1 };
 
 			const { cursor, totalCount } = Messages.findPaginated(ourQuery, {
-				sort: sortObj,
+				sort: sort,
```
MarcosSpessatto, line 405: I believe it would be good to have a default (ts) in case nothing is provided.

MarcosSpessatto (changes requested):
We need to add a changeset for this change. And I think this is more a feat than a fix

apps/meteor/app/api/server/v1/im.ts
MarcosSpessatto: Also, can we add a test case to cover this new behavior?

MahmoudMowiena: Yes, it's a great idea to add a test to the suite to ensure everything works as expected. I have implemented that test, along with the default ts (timestamp) sorting behavior you mentioned. I believe your point about the default ts value is valid, and it makes the implementation more robust.

abhinavkrin:
Hey @MahmoudMowiena Thanks for bringing this issue forward and for your contribution. 
Closing this PR since it will be resolved by the pr #35260

PR #34030, opened 2024-11-21
regression: Do not throw an error on custom fields validation when migrating visitors to contacts

matheusbsilva137:

 - In case an invalid custom field is found when migrating a visitor to a contact, we should just not use it instead of failing the whole process;
 - Do not create custom field conflicts on contact merge

## Further comments

[SCI-179]

[SCI-179]: https://rocketchat.atlassian.net/browse/SCI-179?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


KevLehman (approved):

KevLehman (changes requested):
Pls fix tests

pierre-lehnen-rc (changes requested):
If an invalid value is already saved on the visitor, we shouldn't just discard it.
Let's register that value as a conflict for the custom field instead.

Also: we probably need to do the same thing on the merge process.

KevLehman (approved):

PR #34057, opened 2024-11-25
feat: Workspace/subscription ux

rique223:

1. Changed the fontScale of the WorkspaceCardSection body to `p2`, the fontScale of the WorkspaceCardSection title to `p2b` and the fontScale of the WorkspaceCardSection header titles to `h4`;
2. Centralized the information of the Seats and Monthly Active Contracts cards `vertically` and `horizontally`;
3. Centralized the information of the Active Sessions and Active Sessions peak cards `horizontally`;
4. Added a minHeight of `244px` to the cards and changed the margin between the community card content to be `8px`;
5. Componentized the WorkspaceCardSectionTitle so that it can have variations;
6. Updated the version card of the workspace page to better align with design specifications.

Demos: 

Workspace page:

image
*FontScale and spacing changes*

workspace_background_responsivity
*Change responsivity strategy*

Subscription page:

image
*Centralize content vertically and horizontally*

image
*Change spacing for community card and apps card. Add 244px "min-height" to all the cards*

image
*Centralize content horizontally*

## Further comments

Jira tasks:
[CONN-413: Workspace font size](https://rocketchat.atlassian.net/browse/CONN-413)
[CONN-414: Version card background image responsivity](https://rocketchat.atlassian.net/browse/CONN-414)
[CONN-415: Centralize seats and MAC](https://rocketchat.atlassian.net/browse/CONN-415)
[CONN-417: Subscription page spacing](https://rocketchat.atlassian.net/browse/CONN-417)
[CONN-418: Centralize active sessions and active sessions peak](https://rocketchat.atlassian.net/browse/CONN-418)


apps/meteor/client/views/admin/subscription/components/cards/PlanCard/PlanCardCommunity.tsx
```diff
@@ -12,12 +12,12 @@ const PlanCardCommunity = (): ReactElement => {
 		<Card height='full'>
 			<PlanCardHeader name={t('Community')} />
 			<CardBody flexDirection='column'>
-				<CardRow>
+				<Box display='flex' flexDirection='row' alignItems='center'>
```
ggazzo, line 15: i think we should stay with CardRow dont you think? otherwise we are going to only use box for everything, and bye bye componentization (look just repeated yourself) twice here

apps/meteor/client/views/admin/workspace/components/WorkspaceCardSection.tsx
```diff
@@ -5,12 +5,13 @@ import React from 'react';
 type WorkspaceCardSectionProps = {
 	title: string;
 	body?: ReactNode;
+	isHeader?: boolean;
 };
 
-const WorkspaceCardSection = ({ title, body }: WorkspaceCardSectionProps) => {
+const WorkspaceCardSection = ({ title, body, isHeader }: WorkspaceCardSectionProps) => {
```
ggazzo, line 11: make a new component contemplating this variation

apps/meteor/client/views/admin/subscription/components/cards/PlanCard/PlanCardCommunity.tsx
rique223: In this specific situation, I believe there is no problem with the boxes as they are part of arbitrary custom content, like the seats and MACs counter cards of the Subscription page. Also, CardRow was not made to be used with vertical Cards, and using it here caused issues with the CardBody content

ggazzo:
if reflects in ui/ux changes is not a refactor, also requires changeset

apps/meteor/client/views/admin/subscription/SubscriptionPage.tsx
```diff
@@ -130,16 +130,17 @@ const SubscriptionPage = () => {
 						)}
 						<Box marginBlock='none' marginInline='auto' width='full' color='default'>
 							<Grid m={0}>
-								<Grid.Item lg={4} xs={4} p={8}>
+								<Grid.Item lg={4} xs={4} p={8} minHeight={244}>
```
aleksandernsilva, line 133: Since the grid has a padding of 8px, you might want to consider increasing the minHeight to 260px (244 + 16). Adding the minHeight 244px directly to the `FeatureUsageCard` would also work (and wouldn't break if the padding changes in the future).

aleksandernsilva (changes requested):

apps/meteor/client/views/admin/subscription/SubscriptionPage.tsx
rique223: Adding the minHeight to FeatureUsageCard isn't really an option for now as it would require changing fuselage since the Card component doesn't have box style props. I'll change the minHeight of the grid item to 260

apps/meteor/client/views/admin/subscription/components/FeatureUsageCard.tsx
```diff
@@ -13,17 +13,26 @@ export type CardProps = {
...
 				{title} {infoText && <InfoTextIconModal title={title} infoText={infoText} />}
 			</CardTitle>
 			<CardBody>
-				<Box h='full' w='full'>
+				<Box
+					h='full'
+					w='full'
+					display='flex'
+					alignItems='center'
+					justifyContent={isHorizontallyCenteredOnly ? 'flex-start' : 'center'}
```
juliajforesti, line 33: instead of passing this `isHorizontallyCenteredOnly` as a prop, couldn't you do this conditional styling directly on the parent, on the `children` wrapper?

juliajforesti (changes requested):

apps/meteor/client/views/admin/subscription/components/FeatureUsageCard.tsx
rique223: If I understood your suggestion, this would require raising the box that is currently inside of the FeatureUsageCard to each card and then changing the justifyContent according to which cards need to be centered. I think this would require a lot of unnecessary duplication of code as all of the styling of this box besides the justifyContent is default to all cards. What I could do instead is change the boolean to a string called justifyContent and just pass it to the justifyContent of the children's parent box. Wdyt? If I didn't get the suggestion right please let me know.

juliajforesti: So maybe create a component:
`FeatureUsageCardBody` that receives the styling props, and this component would be passed as children of the `FeatureUsageCard`
eg
```
		<FeatureUsageCard card={card}>
			<FeatureUsageCardBody justifyContent='flex-start'>
					(...........)
			</FeatureUsageCardBody justifyContent='flex-start'>
		</FeatureUsageCard>
```
I just don't like styling props being inside a `card` object. This prop should only receive values that are relative to the card content (title, text, etc).

apps/meteor/client/views/admin/workspace/components/WorkspaceCardSectionTitle.tsx
```diff
@@ -0,0 +1,16 @@
+import { Box } from '@rocket.chat/fuselage';
+import React from 'react';
+
+type WorkspaceCardSectionTitleProps = {
+	variant?: 'p2b' | 'h4';
+	title: string;
+	titleToContentGap?: number;
+};
+
+const WorkspaceCardSectionTitle = ({ variant = 'p2b', title, titleToContentGap = 0 }: WorkspaceCardSectionTitleProps) => (
```
juliajforesti, line 10: If the variations using `h4` all have the `titleToContentGap={20}` they can be a different component, or you can simply add the margin if the variant is `h4`. 
Again, we want this components to be as much standardized as much as we can - so we are avoiding opening those styling props.

juliajforesti (changes requested):

dougfabris:
if it changes something to the end user it shouldn't be a chore

juliajforesti (approved):

tassoevan (approved):

aleksandernsilva (approved):

PR #32906, opened 2024-07-25
feat: Improve thread metrics

gabriellsh:
## Proposed changes (including videos or screenshots)

Updated the thread metrics. Changes include elements positioning, titles and a new avatar "list" of the users following the thread.
image
image
image

## Issue(s)

[CORE-542](https://rocketchat.atlassian.net/browse/CORE-542)
## Further comments

Depends on #33214

[CORE-542]: https://rocketchat.atlassian.net/browse/CORE-542?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #34061, opened 2024-11-26
regression: `UserAvatar` should always render `BaseAvatar`

dougfabris:

Introduced here: #32906

Previously we didn't check `userId` or `username` to render `BaseAvatar`, neither throw an Error. There are some cases that `username` can be `undefined` such as `ComposerBoxPopupUser` which will break the entire UI if we keep the current solution. 

The PR aims to not throw an Error for now, until we fix the issue causing the `undefined` value on `ComposerBoxPopupUser`

## Further comments

[CORE-834]

[CORE-834]: https://rocketchat.atlassian.net/browse/CORE-834?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/ui-contexts/src/AvatarUrlContext.ts
+++ b/packages/ui-contexts/src/AvatarUrlContext.ts
@@ -6,7 +6,7 @@ export type AvatarUrlContextValue = {
 	getUserPathAvatar: {
 		(username: string, etag?: string): string;
 		(params: { userId: string; etag?: string }): string;
-		(params: { username?: string; etag?: string }): string;
+		(params: { username: string; etag?: string }): string;
 	};
 	getRoomPathAvatar: (...args: any) => string;
 };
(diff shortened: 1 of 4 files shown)


MartinSchoeler (dismissed):

packages/ui-contexts/src/AvatarUrlContext.ts
```diff
@@ -6,7 +6,7 @@ export type AvatarUrlContextValue = {
 	getUserPathAvatar: {
 		(username: string, etag?: string): string;
 		(params: { userId: string; etag?: string }): string;
-		(params: { username: string; etag?: string }): string;
+		(params: { username?: string; etag?: string }): string;
```
ggazzo, line 9: how is this possible? a function that accepts an empty object will return something?

dougfabris: @ggazzo did you checked the function?

ggazzo: sure and still makes no sense rendering something if you have no info about thats why this params are wrong

you are trying to fix a different problem with a "naive" solution


your problem is the `/invite @something`  breaking right ?


for a small fraction the autocomplete uses one render type for a different item list thats why username is undefined there. but this is the real problema.

 even this change solves the issue, I think you are covering the wrong place

ggazzo (approved):

PR #34062, opened 2024-11-26
fix: Omnichannel queue starting multiple times due to race condition

aleksandernsilva:
This PR fixes a bug causing the Omnichannel queue to start more than once due to a race condition. In order to prevent the race condition `settings.watch` was changed to `settings.watchMultiple` which has a debounce built in.

Bug caught through this test [omnichannel-livechat-queue-management-autoselection.spec.ts](https://github.com/RocketChat/Rocket.Chat/blob/4d0ab190ec7e6a2bf9b2809697d5edb87a0ab4ab/apps/meteor/tests/e2e/omnichannel/omnichannel-livechat-queue-management-autoselection.spec.ts)

Introduced here: [#33719](https://github.com/RocketChat/Rocket.Chat/pull/33719)

## Further comments
[CONN-429](https://rocketchat.atlassian.net/browse/CONN-429)

[CONN-429]: https://rocketchat.atlassian.net/browse/CONN-429?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/meteor/client/components/UserCard/UserCard.stories.tsx
+++ b/apps/meteor/client/components/UserCard/UserCard.stories.tsx
@@ -3,6 +3,20 @@ import React from 'react';
 
 import { UserCard, UserCardRole, UserCardAction } from '.';
 
+const user = {
+	name: 'guilherme.gazzo',
+	customStatus: '🛴 currently working on User Card',
+	roles: (
+		<>
+			<UserCardRole>Admin</UserCardRole>
+			<UserCardRole>Rocket.Chat</UserCardRole>
+			<UserCardRole>Team</UserCardRole>
+		</>
+	),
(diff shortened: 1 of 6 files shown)


pierre-lehnen-rc (commented):
It will still be called multiple times as `shouldStart` is also called when the license is validated.

apps/meteor/server/services/omnichannel/queue.ts
```diff
@@ -45,6 +51,7 @@ export class OmnichannelQueue implements IOmnichannelQueue {
 		await LivechatInquiry.unlockAll();
 
 		this.running = false;
+		this.lock = Promise.resolve();
```
pierre-lehnen-rc, line 54: this will just remove the reference to the existing stack of calls to `start`, without preventing them from running. 
If a new call is made afterwards, it'll not wait for the existing stack of calls to complete.

pierre-lehnen-rc (changes requested):

pierre-lehnen-rc (dismissed):

apps/meteor/server/services/omnichannel/queue.ts
```diff
@@ -34,20 +46,34 @@ export class OmnichannelQueue implements IOmnichannelQueue {
...
+
 		queueLogger.info('Service stopped');
 	}
 
+	async start() {
+		return this.serviceStarter.start();
+	}
+
+	async stop() {
+		await this.serviceStarter.stop();
```
pierre-lehnen-rc, line 74:
```suggestion
		return this.serviceStarter.stop();
```

apps/meteor/server/services/omnichannel/queue.ts
```diff
@@ -34,20 +46,34 @@ export class OmnichannelQueue implements IOmnichannelQueue {
 		this.running = true;
 
 		queueLogger.info('Service started');
-		return this.execute();
+		await this.execute();
```
pierre-lehnen-rc, line 49:
```suggestion
		return this.execute();
```

packages/core-services/tests/ServiceStarter.test.ts
```diff
@@ -0,0 +1,105 @@
...
+			});
+		});
+
+		await instance.wait();
+
+		expect(start).toHaveBeenCalledTimes(2);
+		expect(stop).not.toHaveBeenCalled();
+	});
+
+	it('should skip the chained calls to .start if .stop is called', async () => {
```
KevLehman, line 81: Can you add cases for:
- calling multiple times the same `start` function without the setinmediate and without await?
- ensuring the chain is in the right order, like start stop start and assure the order is right.

aleksandernsilva: - Adjusted the code to not need `setImmediate`. 
- Considering the restrictions `ServiceStarter` puts on the number of actions allowed to be executed I believe the the current test cases already covers the chain order. If you have any suggestions let me know.

pierre-lehnen-rc (approved):

KevLehman (approved):

PR #34076, opened 2024-11-28
feat: Export selected room messages as JSON file

dougfabris:

Kapture 2024-12-05 at 16 54 10

Introduces a new option when exporting messages, allowing users to select and download a JSON file directly from the client

## Further comments

[CORE-807]
[CORE-808]

[CORE-807]: https://rocketchat.atlassian.net/browse/CORE-807?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[CORE-808]: https://rocketchat.atlassian.net/browse/CORE-808?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


MarcosSpessatto (changes requested):
Would it be possible to add tests?

apps/meteor/client/views/room/body/hooks/useGetMore.ts
```diff
@@ -1,17 +1,28 @@
...
+import { useToggleSelectAll } from '../../MessageList/contexts/SelectedMessagesContext';
 
 export const useGetMore = (rid: string, atBottomRef: MutableRefObject<boolean>) => {
+	const ref: MutableRefObject<HTMLElement | null> = useRef(null);
+	const handleToggleAll = useToggleSelectAll();
+
+	const handleGetMore = () => {
+		ref.current?.scrollTo({ top: 0, behavior: 'smooth' });
+		handleToggleAll();
+	};
```
gabriellsh, line 15: Instead of defining this function here in order to get the ref, I recommend making a hook for this (or, place the scroll logic in the `useToggleSelectAll`  hook) and pass the ref to the hook from the `RoomBody` component.

gabriellsh: also, `handleGetMore` is a little confusing, to be honest... Can we have a more descriptive name? Like `selectAllAndScrollToTop` or maybe something else.

apps/meteor/client/views/room/composer/ComposerSelectMessages.tsx
```diff
@@ -0,0 +1,34 @@
+import { Button, ButtonGroup } from '@rocket.chat/fuselage';
+import { MessageFooterCallout, MessageFooterCalloutContent } from '@rocket.chat/ui-composer';
+import type { ReactElement } from 'react';
+import React, { useContext } from 'react';
+import { useTranslation } from 'react-i18next';
+
+import type { ComposerMessageProps } from './ComposerMessage';
+import { useCountSelected, useClearSelection, SelectedMessageContext } from '../MessageList/contexts/SelectedMessagesContext';
+
+const ComposerSelectMessages = ({ onGetMore }: ComposerMessageProps): ReactElement => {
```
gabriellsh, line 10: Naming here could use some improvement too.

```suggestion
const ComposerSelectMessages = ({ onClickSelectAll }: ComposerMessageProps): ReactElement => {
```

Or maybe something different, just a suggestion.

apps/meteor/client/views/room/providers/SelectedMessagesProvider.tsx
```diff
@@ -14,8 +16,19 @@ export const selectedMessageStore = new (class SelectMessageStore extends Emitte
...
+
+	availableMessages = new Set<string>();
+
 	isSelecting = false;
 
+	constructor() {
+		super();
+		this.on('change', () => {
+			this.storeArray = Array.from(this.store);
+		});
```
gabriellsh, line 29: Why are you saving the set as an array in memory at all times? When selecting a large number of messages (toggle all for example) this would run many times and can be expensive depending on the ammount of messages already selected.

apps/meteor/client/views/room/providers/SelectedMessagesProvider.tsx
```diff
@@ -61,13 +84,24 @@ export const selectedMessageStore = new (class SelectMessageStore extends Emitte
 		this.isSelecting = false;
 		this.emit('toggleIsSelecting', false);
 	}
+
+	toggleAll(mids: string[]): void {
+		mids.forEach((mid) => this.select(mid));
+	}
```
gabriellsh, line 90: Instead of a foreach, is it possible to have a method in this class that does a bulk procedure? This way we avoid emmiting a ton of events from the store.

```
this.store = new Set([...this.store, ...mids])
```

I think this should work. A foreach can also work, but then you should emit the 'change' event only once after the operation has ended, instead of one time for each entry.

apps/meteor/client/views/room/providers/SelectedMessagesProvider.tsx
```diff
@@ -61,13 +84,24 @@ export const selectedMessageStore = new (class SelectMessageStore extends Emitte
...
 type SelectedMessagesProviderProps = {
 	children?: ReactNode;
 };
 
 export const SelectedMessagesProvider = ({ children }: SelectedMessagesProviderProps) => {
+	const room = useRoom();
+	const messages = useMessages({ rid: room._id });
+
+	useEffect(() => {
+		selectedMessageStore.availableMessages = new Set(messages.map((message) => message._id));
```
gabriellsh, line 102: Since the store is a class, instead of assigning a value directly to to a property, a method that receives an array of messages would be best for this use.

But I think we can improve a little on this solution so that we don't even need this useEffect. Each rendered message component calls `useIsSelectedMessage` with the messageId. You could register the available messages from inside the hook, then you don't need to keep syncing this value in the Provider every time something changes...

packages/i18n/src/locales/en.i18n.json
```diff
@@ -3773,7 +3775,8 @@
   "Message_VideoRecorderEnabledDescription": "Requires 'video/webm' files to be an accepted media type within 'File Upload' settings.",
   "messages": "messages",
   "Messages": "Messages",
-  "Messages_selected": "Messages selected",
+  "number_messages_selected": "{{count}} messages selected",
```
gabriellsh, line 3778:
```suggestion
  "__count__messages_selected": "{{count}} messages selected",
```

Not sure if this is still enforced today, but we have a pattern of adding double underscores before and after a property, in the key, like this:  `__property__`.

Also, the property referenced in the key usually has the same name as the property used inside the brackets. Ex: `__count__messages_selected` instead of `number`

packages/i18n/src/locales/en.i18n.json
```diff
@@ -6684,6 +6688,7 @@
   "Go_to_href": "Go to: {{href}}",
   "Anyone_can_send_new_messages": "Anyone can send new messages",
   "Select_messages_to_hide": "Select messages to hide",
+  "Select_number_messages": "Select {{count}} messages",
```
gabriellsh, line 6691:
```suggestion
  "Select__count__messages": "Select {{count}} messages",
```

gabriellsh (changes requested):

apps/meteor/client/views/room/providers/SelectedMessagesProvider.tsx
dougfabris: removed!

MarcosSpessatto (approved):

gabriellsh (approved):

apps/meteor/tests/e2e/export-messages.spec.ts
```diff
@@ -0,0 +1,72 @@
...
+
+		await poHomeChannel.content.getMessageByText('hello world').click();
+		await poHomeChannel.tabs.exportMessages.btnSend.click();
+
+		await expect(
+			poUtils.getAlertByText('You must select one or more users or provide one or more email addresses, separated by commas'),
+		).toBeVisible();
+	});
+
+	test('should display an error when trying to send email without selecting any email', async () => {
```
MarcosSpessatto, line 50:
```suggestion
	test('should display an error when trying to send email without selecting any message', async () => {
```

MarcosSpessatto (approved):

MarcosSpessatto (approved):

gabriellsh (approved):

PR #34079, opened 2024-11-28
fix: twitter oembed image not showing

abhinavkrin:

This pr fixes the issue where the image is not showing link for twitter article links. This pr removes the provider for twitter oembed and hence now relies on the meta tags  (which is the default behavior when a provider is not provided).
Screenshot 2024-11-28 at 11 15 26 PM

## Further comments

[SUP-685](https://rocketchat.atlassian.net/browse/SUP-685)

[SUP-685]: https://rocketchat.atlassian.net/browse/SUP-685?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


abhinavkrin:
I have not yet added tests, since it would include adding a tweet link for testing. The tests will fail if the tweet is deleted.

scuciatto (changes requested):
Can you a changeset, please?

MarcosSpessatto (approved):

scuciatto (dismissed):

apps/meteor/app/oembed/server/providers.ts
```diff
@@ -11,13 +13,20 @@ class Providers {
 		this.providers = [];
 	}
 
-	static getConsumerUrl(provider: OEmbedProvider, url: string): string {
+	static getConsumerUrl(provider: OEmbedProvider, url: string): string | undefined {
+		if (!provider.endPoint) {
```
MarcosSpessatto, line 17: Is there any possibility of endPoint being undefined here? If so, we should update that to reflect on the typings.

MarcosSpessatto (changes requested):

apps/meteor/app/oembed/server/providers.ts
abhinavkrin: Yes, 
I have modified the provider definition.
- made endpoint optional
- introduced optional headerOverrides
This will help us in cases when we want to modify the headers without adding endpoint.
Hence the server will call the default parsed url with custom headers. 
Some urls returns meta tags depending on user agents.

PS. I have modified the types already.

KevLehman (approved):

MarcosSpessatto (approved):

PR #34090, opened 2024-11-29
regression: old omnichannel rooms only being migrated after their data is loaded by the client

pierre-lehnen-rc:
## Further comments

[SCI-190]

[SCI-190]: https://rocketchat.atlassian.net/browse/SCI-190?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

We no longer use `contacts.get` using the visitor association as parameter, only using the contactId; Since this endpoint was introduced on this same release, we've removed the option of using the visitor altogether with no risk of breaking anything.


lucas-a-pelegrino (approved):

apps/meteor/app/livechat/server/startup.ts
```diff
@@ -21,7 +22,11 @@ import './roomAccessValidator.internalService';
 const logger = new Logger('LivechatStartup');
 
 Meteor.startup(async () => {
-	roomCoordinator.setRoomFind('l', (_id) => LivechatRooms.findOneById(_id));
+	roomCoordinator.setRoomFind('l', async (_id) => {
+		const room = await LivechatRooms.findOneById(_id);
+
+		return (await maybeMigrateLivechatRoom(room)) ?? undefined;
```
aleksandernsilva, line 28: I noticed that `LivechatRooms.findOneById` returns `IOmnichannelRoom | null`, same for `maybeMigrateLivechatRoom`. Are we intentionally changing the fallback to `undefined` here?

KevLehman (approved):
if ci happy me happy

apps/meteor/app/livechat/server/startup.ts
pierre-lehnen-rc: `roomFind?: (rid: string) => Promise<IRoom | undefined> | Promise<IOmnichannelRoom | null> | IRoom | undefined;`

:(

But this was the last place still calling maybeMigrateLivechatRoom, so I changed it to use IOmnichannelRoom instead of IRoom.

PR #34105, opened 2024-12-03
fix: fix username sorted in the member list

SaraswatSahoo:
hey I fixed the issue using the sort function and setting another function as parameter to sort the username alphabetically 
screenshot_04122024_000849
Please go through the code.

### DIFF
--- a/apps/meteor/client/views/room/contextualBar/RoomMembers/RoomMembers.tsx
+++ b/apps/meteor/client/views/room/contextualBar/RoomMembers/RoomMembers.tsx
@@ -86,6 +86,15 @@ const RoomMembers = ({
 
 	const useRealName = useSetting('UI_Use_Real_Name', false);
 
+	members.sort((usr1,usr2) => {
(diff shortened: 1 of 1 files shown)


apps/meteor/client/views/room/contextualBar/RoomMembers/RoomMembers.tsx
```diff
@@ -86,6 +86,15 @@ const RoomMembers = ({
...
+	members.sort((usr1,usr2) => {
+		const username1 = usr1.username ?? '';
+		const username2 = usr2.username ?? '';
+
+		if (username1 < username2) return -1;
+		if (username1 > username2) return 1;
+		return 0;
+	})
+
 	return (
```
ggazzo, line 98: I'm confident that should come from server already sorted.

also mutating an array inside a render function is not something recommended on react

ggazzo (changes requested):

PR #34106, opened 2024-12-03
fix: app subprocess restart on error and better reporting

ggazzo:


packages/apps-engine/deno-runtime/handlers/app/handler.ts
```diff
@@ -21,7 +21,23 @@ export default async function handleApp(method: string, params: unknown): Promis
 
     // We don't want the getStatus method to generate logs, so we handle it separately
     if (appMethod === 'getStatus') {
-        return handleGetStatus();
+        try {
+            return handleGetStatus();
```
ggazzo, line 25:
```suggestion
            return await handleGetStatus();
```

d-gubert (approved):

PR #34124, opened 2024-12-05
fix: Use all initials to create avatar when setting is `UI_Use_Full_Name` is true

KevLehman:
## Issue(s)

https://rocketchat.atlassian.net/browse/SUP-678

### DIFF
--- a/apps/meteor/server/routes/avatar/user.ts
+++ b/apps/meteor/server/routes/avatar/user.ts
@@ -66,7 +66,7 @@ export const userAvatarByUsername = async function (request: IncomingMessage, re
 		});
 
 		if (user?.name) {
-			serveSvgAvatarInRequestedFormat({ nameOrUsername: user.name, req, res });
+			serveSvgAvatarInRequestedFormat({ nameOrUsername: user.name, req, res, useAllInitials: true });
 			return;
 		}
 	}
@@ -126,7 +126,7 @@ export const userAvatarById = async function (request: IncomingMessage, res: Ser
 
 	// Use real name for SVG letters
 	if (settings.get('UI_Use_Name_Avatar') && user?.name) {
-		serveSvgAvatarInRequestedFormat({ nameOrUsername: user.name, req, res });
+		serveSvgAvatarInRequestedFormat({ nameOrUsername: user.name, req, res, useAllInitials: true });
 		return;
 	}
 
--- a/apps/meteor/server/routes/avatar/utils.spec.ts
+++ b/apps/meteor/server/routes/avatar/utils.spec.ts
@@ -165,6 +165,21 @@ describe('#renderSvgLetters', () => {
 		expect(renderSVGLetters('Bob', 32)).to.include('viewBox="0 0 32 32"');
 		expect(renderSVGLetters('yan', 64)).to.include('viewBox="0 0 64 64"');
 	});
+	it('should return a default size of 125 for a single letter', () => {
+		expect(renderSVGLetters('a', 200)).to.include('font-size="125"');
+	});
+	it('should render a single letter when useAllInitials is false', () => {
+		expect(renderSVGLetters('arthur', 16, false)).to.include('>\nA\n');
+	});
+	it('should render a single letter when useAllInitials is true but username has no spaces', () => {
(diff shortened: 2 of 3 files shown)


apps/meteor/server/routes/avatar/utils.spec.ts
```diff
@@ -165,6 +165,21 @@ describe('#renderSvgLetters', () => {
...
+	});
+	it('should render a single letter when useAllInitials is true but username has no spaces', () => {
+		expect(renderSVGLetters('arthur', 16, true)).to.include('>\nA\n');
+	});
+	it('should render more than one letter when useAllInitials is true', () => {
+		expect(renderSVGLetters('arthur void', 16, true)).to.include('>\nAV\n');
+	});
+	it('should decrease the font size when username has more than 1 word', () => {
+		expect(renderSVGLetters('arthur void', 200, true)).to.include('font-size="100"');
+	});
```
MarcosSpessatto, line 182: Can we have a test with a name with multiple letters? Like a name with 4 words?

MarcosSpessatto (changes requested):

MarcosSpessatto (dismissed):

apps/meteor/server/routes/avatar/utils.ts
```diff
@@ -18,6 +18,7 @@ const cookie = new Cookies();
 
 export const MAX_SVG_AVATAR_SIZE = 1024;
 export const MIN_SVG_AVATAR_SIZE = 16;
+const MAX_SVG_AVATAR_INITIALS = 3;
```
Gustrb, line 21: It seems this is used as a count of how many initials we are using, do we support larger charsets?
I don't think this would work with japanese, or chineese characters (not sure if we support them tho)

Gustrb: It looks like if you change the `getFirstLetter` to just be something like this:
```ts
const getFirstLetter = (name: string) => name.charAt(0);
```
Or even inline it, we can better handle unicode strings (I am just not 100% if they will be rendered properly in the icon)

KevLehman: > It looks like if you change the `getFirstLetter` to just be something like this:
> I just think we need to sanitize name before fetching the first char, I think we use DOMPurify to do that right?

There's another PR that touches this function to support UTF8 chars, so we can skip this for now.

KevLehman: > It seems this is used as a count of how many initials we are using, do we support larger 4-byte character points? I don't think this would work with japanese, or chinese characters (not sure if we support them tho)

Don't think we currently support them, but we can revisit that after we support at least 1 character svgs

Gustrb: For sure :)
just wanted to bring to attention

Gustrb (approved):

gabriellsh (dismissed):

MarcosSpessatto (approved):

apps/meteor/server/routes/avatar/user.ts
```diff
@@ -40,7 +40,20 @@ export const userAvatarByUsername = async function (request: IncomingMessage, re
...
+					name: 1,
+				},
+			});
+
+			if (user?.name) {
+				serveSvgAvatarInRequestedFormat({ nameOrUsername: user.name, req, res, useAllInitials: true });
+				return;
+			}
+		}
+		serveSvgAvatarInRequestedFormat({ nameOrUsername: usernameWithoutAt, req, res });
```
gabriellsh, line 56: Unsure if this is the correct place for this logic...

First, if I'm changing my name in my user profile, that won't be reflected in the preview avatar until I save the changes. Secondly, this was never suportted like this, and I don't think it is the resposibility of this resource to deal with this in specific.

I'd argue this should be fixed in the UI (we can just send the name in the place of the username), and for this to also be a separate task, since it's a different issue, also requiring it's own changelog. Can we revert this?

gabriellsh (changes requested):

MarcosSpessatto (approved):

PR #34137, opened 2024-12-06
fix: Unit's `numDepartments` prop not being updated after a department is removed

KevLehman:
## Issue(s)

https://rocketchat.atlassian.net/browse/CORE-541
## Further comments

This is not a retroactive fix. Units with an already incorrect count will continue to have it. However, new units created after the fix won't have it anymore.

### DIFF
--- a/apps/meteor/ee/app/livechat-enterprise/server/hooks/afterRemoveDepartment.ts
+++ b/apps/meteor/ee/app/livechat-enterprise/server/hooks/afterRemoveDepartment.ts
@@ -1,11 +1,11 @@
 import type { AtLeast, ILivechatAgent, ILivechatDepartment } from '@rocket.chat/core-typings';
-import { LivechatDepartment } from '@rocket.chat/models';
+import { LivechatDepartment, LivechatUnit } from '@rocket.chat/models';
 
 import { callbacks } from '../../../../../lib/callbacks';
 import { cbLogger } from '../lib/logger';
 
 const afterRemoveDepartment = async (options: {
-	department: AtLeast<ILivechatDepartment, '_id' | 'businessHourId'>;
+	department: AtLeast<ILivechatDepartment, '_id' | 'businessHourId' | 'parentId'>;
 	agentsId: ILivechatAgent['_id'][];
 }) => {
 	if (!options?.department) {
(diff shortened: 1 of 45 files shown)


apps/meteor/ee/app/livechat-enterprise/server/hooks/afterRemoveDepartment.ts
```diff
@@ -15,8 +15,14 @@ const afterRemoveDepartment = async (options: {
...
 
-	cbLogger.debug(`Removing department from forward list: ${department._id}`);
-	await LivechatDepartment.removeDepartmentFromForwardListById(department._id);
+	cbLogger.debug({
+		msg: 'Post removal actions on EE code for department',
+		department,
+	});
+	await Promise.all([
+		LivechatDepartment.removeDepartmentFromForwardListById(department._id),
+		...(department.parentId ? [LivechatUnit.removeDepartmentFromUnit(department.parentId)] : []),
```
matheusbsilva137, line 24: I'm wondering if we should really use the `LivechatUnit` model here since it adds some extra queries to assure the user who requested this action can really view/maange the unit. In this case, having this validation could cause inconsistencies, couldn't it? I mean in case the user can manage the department, but can't manage its unit

Maybe we should use the `LivechatDepartment` model instead :thinking: wdyt?

matheusbsilva137 (changes requested):

apps/meteor/ee/app/livechat-enterprise/server/hooks/afterRemoveDepartment.ts
KevLehman: - Should be on `Units` model, cause you're updating a unit.
- The restrictions on the Units model should be manual. I don't remember applying automatic restrictions to any query (we removed that behavior some time ago) so the `updateOne` should go clean. Unless something has changed that i'm not aware of 👀

Gustrb:
Do we also have a PR for the next major to fix this inconsistency on existing Units?

apps/meteor/ee/server/models/raw/LivechatUnit.ts
```diff
@@ -190,4 +190,15 @@ export class LivechatUnitRaw extends BaseRaw<IOmnichannelBusinessUnit> implement
 	countUnits(): Promise<number> {
 		return this.col.countDocuments({ type: 'u' });
 	}
+
+	removeDepartmentFromUnit(unitId: string): Promise<UpdateResult> {
```
Gustrb, line 194: Shouldn't we tie it to the BusinessUnit interface?
```suggestion
	removeDepartmentFromUnit(unitId: IOmnichannelBusinessUnit['_id']): Promise<UpdateResult> {
```

apps/meteor/ee/app/livechat-enterprise/server/hooks/afterRemoveDepartment.ts
matheusbsilva137: Just noticed the `addQueryRestrictions` function we have defined in this file is not applied to updateOne, so we should be fine :+1:

matheusbsilva137 (changes requested):

matheusbsilva137 (approved):

PR #34141, opened 2024-12-08
fix: fix sorting of usernames in the member-list

SaraswatSahoo:
Hey I fixed,
1. the sorting problem of the usernames in the member-list  for all room types \
2. by adding a sorting function at all three endpoint **channels.members, groups.members, im.members** before returning the member list.\
3. 
241209_12h22m42s_screenshot

### DIFF
--- a/apps/meteor/app/api/server/v1/channels.ts
+++ b/apps/meteor/app/api/server/v1/channels.ts
@@ -1082,6 +1082,8 @@ API.v1.addRoute(
 
 			const [members, total] = await Promise.all([cursor.toArray(), totalCount]);
 
+			members.sort((a, b) => (a.username ?? "").localeCompare(b.username ?? ""));
+
 			return API.v1.success({
 				members,
 				count: members.length,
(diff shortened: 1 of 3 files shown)


SaraswatSahoo:
Hey please check this Pull Request.

KevLehman (changes requested):
Sorting comes from MongoDB automatically. You can see on [L1080](https://github.com/RocketChat/Rocket.Chat/pull/34141/files#diff-518ba4b854cfb6791cdef873de80f16e9d967b85c6d1fb24e5d487354854fdd8L1080) for example that we're passing a `sort` property, that one is sent to MongoDB which returns the data sorted. Same happens on other files.

So we don't need to sort in memory.

SaraswatSahoo:
well the issue is it is not returning a sorted data from the database

SaraswatSahoo:
so it is better if after getting the dat from the database we return a sorted list to the call
if i am wrong please brief me

reetp:
> well the issue is it is not returning a sorted data from the database

Where is the github issue with a completed template documenting this?

Some random PR will get nowhere fast.

Also read this please.

https://gist.github.com/reetp/b0ba4e3d0ea2ff48ca9da00a5a647d42

PR #34148, opened 2024-12-09
fix: Messages not being processed for all slack servers

lucas-a-pelegrino:

This PR aims to fix an issue with the `onMessage()` function that exited before processing messages for all slack adapters, the cause was that while inside the loop, after processing the message on the first adapter the loop exited when accessing the any of the if statements present in the function.

By replacing the `return` statements for `continue`, the loop now continues iterating through the slack adapters as expected.

## Issue(s)

[SUP-700](https://rocketchat.atlassian.net/browse/SUP-700)

## Further comments

[SUP-700]: https://rocketchat.atlassian.net/browse/SUP-700?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


matheusbsilva137 (changes requested):
Can you please add a changeset and describe what's being fixed in the PR title (instead of how)?

apps/meteor/app/slackbridge/server/RocketAdapter.js
```diff
@@ -134,22 +134,23 @@ export default class RocketAdapter {
...
 				}
 				// Ignore messages originating from Slack
 				if (rocketMessage._id.indexOf('slack-') === 0) {
-					return rocketMessage;
+					continue;
 				}
 
 				if (rocketMessage.file) {
-					return this.processFileShare(rocketMessage, slack);
+					await this.processFileShare(rocketMessage, slack);
```
matheusbsilva137, line 152: This function call has a side effect since it changes the content of `rocketMessage.msg` [here](https://github.com/RocketChat/Rocket.Chat/blob/c2cf2d773adb161f331bd3736a1a94e5b21dd035/apps/meteor/app/slackbridge/server/RocketAdapter.js#L211) :grimacing: this could create issues for all next adapters
I believe we want to use the same `rocketMessage` object that is provided to this method for **all** adapters (in case a message needs to be handled by multiple ones)

matheusbsilva137 (changes requested):

apps/meteor/app/slackbridge/server/RocketAdapter.js
lucas-a-pelegrino: I've done a tiny refactor in `processFileShare()` that should do the trick to avoid this edge case from happening.

KevLehman (commented):
is there any way to add, at least, units? 😬 

if not, can you create a task for adding them later? (and to convert the file to TS, cause we have an eslint rule for ensuring we don't return inside of loops lol)

apps/meteor/app/slackbridge/server/RocketAdapter.js
matheusbsilva137: Not sure I'm confident about this refactor. Shouldn't we return the message with the side effects generated by this function?
With your refactor, we're currently returning the same `rocketMessage` that is provided as a param to this function :thinking:

matheusbsilva137 (changes requested):
I believe we should also fix `onMessageDelete` and `addChannel` methods
They also iterate on `slackAdapters` and are currently using `return` instead of `continue`

apps/meteor/app/slackbridge/server/RocketAdapter.js
lucas-a-pelegrino: > Not sure I'm confident about this refactor. Shouldn't we return the message with the side effects generated by this function? With your refactor, we're currently returning the same `rocketMessage` that is provided as a param to this function 🤔

Not sure I follow, @matheusbsilva137 , with the refactor we are not replacing the original content of `rocketMessage` anymore.

> If all adapters will end up doing the same to the message, why not doing the side effect before looping it?
> 
> So the func has 0 side effects and it's just in charge of sending the message to slack.

I really like this ideia, tho. It would avoid processing the same thing over and over again.

lucas-a-pelegrino:
> is there any way to add, at least, units? 😬
> 
> if not, can you create a task for adding them later? (and to convert the file to TS, cause we have an eslint rule for ensuring we don't return inside of loops lol)

I agree, we should create tasks for migrating this service to TS.

apps/meteor/app/slackbridge/server/RocketAdapter.js
matheusbsilva137: > with the refactor we are not replacing the original content of `rocketMessage` anymore.

I was thinking: shouldn't we? Because we currently change it, but only **once**
Since this methods is used in a callback, I thought the next ones might need it. But I've just checked again the change we apply to `msg` and I don't feel like we were doing it on purpose. I'm fine with your refactor, then :)

KevLehman (approved):
pls create the tasks we mentioned to improve the logic on these adapters 🙏🏽

apps/meteor/app/slackbridge/server/RocketAdapter.js
```diff
@@ -264,15 +265,15 @@ export default class RocketAdapter {
 
 		for await (const slack of this.slackAdapters) {
 			if (addedRoom) {
-				return;
+				continue;
```
matheusbsilva137, line 268: I believe we could use a `break` here, couldn't we?
Since once this variable is set, it won't be undefined ever again

matheusbsilva137 (changes requested):

MarcosSpessatto (approved):

apps/meteor/app/slackbridge/server/RocketAdapter.js
matheusbsilva137: I still think this could be a `break`, but we can leave this for a refactor

matheusbsilva137 (approved):

apps/meteor/app/slackbridge/server/RocketAdapter.js
KevLehman: yeah, should be a break. But since we're refactoring this soon, it may get removed there with other logic changes i hope.

PR #34153, opened 2024-12-10
[WIP] feat: add endpoints groups.membersOrderedByRole channels.membersOrderedByRole

abhinavkrin:

Adds `groups.membersOrderedByRole` and `channels.membersOrderedByRole` endpoints to retrieve members of groups and channels sorted according to their respective role in the room.
This will help in grouping of users according to their roles in the room like below. 
image

The endpoint works similar as groups.members and channels.members, but accepts an optional rolesOrder[] param which can be used for defining roles order while sorting. The index of the role will signify its priority. By default, owner > moderator > rest of members.

A sample request
```
curl --location --globoff 'http://localhost:3000/api/v1/channels.membersOrderedByRole?roomId=GENERAL&sort={%22username%22%3A%201}&offset=0&rolesOrder[]=leader&rolesOrder[]=owner' \
--header 'X-Auth-Token: Ce_2eOJ19mnW59wVO5IfSv4s650MF1w7mqY6i52uPG5' \
--header 'X-User-Id: u29m4aQNjZujdE4r5' \
--data ''
```
## Further comments

[CORE-846](https://rocketchat.atlassian.net/browse/CORE-846)


.changeset/silly-kings-approve.md
```diff
@@ -0,0 +1,6 @@
+---
+'@rocket.chat/meteor': patch
+'@rocket.chat/rest-typings': patch
```
MarcosSpessatto, line 3:
```suggestion
'@rocket.chat/meteor': minor
'@rocket.chat/rest-typings': minor
```

apps/meteor/tests/end-to-end/api/channels.ts
```diff
@@ -1479,6 +1480,196 @@ describe('[Channels]', () => {
...
+
+		let ownerCredentials: { 'X-Auth-Token': string; 'X-User-Id': string };
+
+		before(async () => {
+			ownerUser = await createUser({ username: Random.id(), roles: ['admin'] });
+			ownerCredentials = await login(ownerUser.username, password);
+
+			moderatorUser = await createUser({ username: Random.id() });
+			memberUser1 = await createUser({ username: Random.id() });
+			memberUser2 = await createUser({ username: Random.id() });
```
MarcosSpessatto, line 1498: We can use Promise.all here.

apps/meteor/tests/end-to-end/api/channels.ts
```diff
@@ -1479,6 +1480,196 @@ describe('[Channels]', () => {
...
+				})
+				.expect(200);
+		});
+
+		after(async () => {
+			await deleteRoom({ type: 'c', roomId: testChannel._id });
+			await deleteUser(ownerUser);
+			await deleteUser(moderatorUser);
+			await deleteUser(memberUser1);
+			await deleteUser(memberUser2);
```
MarcosSpessatto, line 1550: We can use Promise.all here.

packages/rest-typings/src/v1/groups/GroupsMembersByOrderedRole.ts
```diff
@@ -0,0 +1,59 @@
...
+		roomId: {
+			type: 'string',
+		},
+		roomName: {
+			type: 'string',
+		},
+		rolesOrder: {
+			type: 'array',
+			items: {
+				type: 'string',
```
MarcosSpessatto, line 27: I think we need to support an enum here OR at least validate the values where we process it and not to allow the users to send arbitrary values here.

packages/rest-typings/src/v1/channels/ChannelsMembersByOrderedRole.ts
```diff
@@ -0,0 +1,59 @@
...
+		roomId: {
+			type: 'string',
+		},
+		roomName: {
+			type: 'string',
+		},
+		rolesOrder: {
+			type: 'array',
+			items: {
+				type: 'string',
```
MarcosSpessatto, line 27: See my other comment about that.

MarcosSpessatto (changes requested):
I think we could create a `rooms.membersOrderedByRole` instead of creating one for each room type.

Also, we need to create some mock data to test the aggregation part and its performance.
@ricardogarim can you help us with that.

packages/rest-typings/src/v1/groups/GroupsMembersByOrderedRole.ts
abhinavkrin: The roles can be custom roles added by the admin. We can check the values if they exists in the db. wdys ?

apps/meteor/tests/end-to-end/api/rooms.ts
```diff
@@ -3647,4 +3648,179 @@ describe('[Rooms]', () => {
 				});
 		});
 	});
+
+	describe('[/rooms.membersOrderedByRole]', () => {
```
KevLehman, line 3652: Add test cases for:
- Trying to fetch a private room by a user not part of room
- Trying to fetch a public room by a user not part of room
- The same 2, but with users that are part of the rooms
- Trying to fetch a private channel inside a public team for someone not part of the room
- Trying to fetch a public channel inside a private team for someone not part of the room
- The same 2, but with users that are part of the rooms

Just to ensure we don't show more data than what we should 🙈

apps/meteor/server/lib/findUsersOfRoomOrderedByRole.ts
```diff
@@ -0,0 +1,188 @@
...
+		},
+		{
+			$project: {
+				members: 1,
+				totalCount: { $arrayElemAt: ['$totalCount.total', 0] },
+			},
+		},
+	];
+
+	const [result] = await Subscriptions.col.aggregate(facetPipeline, { allowDiskUse: true }).toArray();
```
KevLehman, line 178: @MarcosSpessatto , should we run this on the secondary like the other aggregations?

apps/meteor/server/lib/findUsersOfRoomOrderedByRole.ts
```diff
@@ -0,0 +1,188 @@
...
+	rolesInOrder?: IRole['_id'][];
+	exceptions?: string[];
+	extraQuery?: Document[];
+};
+
+type UserWithRoleData = IUser & {
+	roles: IRole['_id'][];
+};
+
+export async function findUsersOfRoomOrderedByRole({
```
KevLehman, line 24: Can you add some explains with some decent amount of data to check on the perf of the aggregation? :please:

apps/meteor/server/lib/findUsersOfRoomOrderedByRole.ts
```diff
@@ -0,0 +1,188 @@
...
+						...(exceptions.length > 0 && { $nin: exceptions }),
+					},
+					...(filter && orStmt.length > 0 && { $or: orStmt }),
+				},
+				...extraQuery,
+			],
+		},
+	});
+
+	userLookupPipeline.push({
```
KevLehman, line 70: You can push both objects at the same call 👀

apps/meteor/app/api/server/v1/rooms.ts
```diff
@@ -860,6 +862,45 @@ API.v1.addRoute(
...
 
+API.v1.addRoute(
+	'rooms.membersOrderedByRole',
+	{ authRequired: true, validateParams: isRoomsMembersOrderedByRoleProps },
+	{
+		async get() {
+			const findResult = await findRoomByIdOrName({
+				params: this.queryParams,
+				checkedArchived: false,
+			});
```
KevLehman, line 873: Correct me, but if this `findRoomByIdOrName` call doesn't check for the user's permissions or access, then this will leak the list of members of private rooms to non admins, right? 👀

KevLehman (changes requested):

apps/meteor/app/api/server/v1/rooms.ts
abhinavkrin: 👀  Yes you are correct. I have added the canAccessRoomAsync logic in the endpoint to check for the user permissions.

apps/meteor/server/lib/findUsersOfRoomOrderedByRole.ts
MarcosSpessatto: I think it makes sense

MarcosSpessatto (commented):
It looks good to me, we just need to do some perf tests with some fake data to ensure the query perf

MartinSchoeler:
Hey, it seems that the `total` property is missing from the endpoint result

(Old Members Endpoint)
Screenshot 2024-12-27 at 16 21 12

(New Members Endpoint)
Screenshot 2024-12-27 at 16 21 30

apps/meteor/server/lib/findUsersOfRoomOrderedByRole.ts
```diff
@@ -0,0 +1,165 @@
...
+			...(limit > 0 ? [{ $limit: limit }] : []),
+		],
+		{
+			allowDiskUse: true,
+		},
+	);
+
+	const totalResult = Subscriptions.col.aggregate([{ $match: { rid } }, ...filteredPipeline, { $count: 'total' }], { allowDiskUse: true });
+
+	const [members, [{ totalCount }]] = await Promise.all([membersResult.toArray(), totalResult.toArray()]);
```
MartinSchoeler, line 155:
```suggestion
	const [members, [{ total: totalCount }]] = await Promise.all([membersResult.toArray(), totalResult.toArray()]);
```

MartinSchoeler: This causes the total property to not be returned on the endpoint

MartinSchoeler (changes requested):

packages/rest-typings/src/v1/rooms.ts
```diff
@@ -786,4 +840,10 @@ export type RoomsEndpoints = {
 	'/v1/rooms.open': {
 		POST: (params: RoomsOpenProps) => void;
 	};
+
+	'/v1/rooms.membersOrderedByRole': {
+		GET: (params: RoomsMembersOrderedByRoleProps) => PaginatedResult<{
+			members: IUser & { roles: IRole['_id'] }[];
```
MartinSchoeler, line 846:
```suggestion
			members: IUser & { roles: IRole['_id'][] };
```

MartinSchoeler (changes requested):

apps/meteor/server/lib/findUsersOfRoomOrderedByRole.ts
abhinavkrin: Hey, the new approach of having a precalculated rolePriority score based on the role seems to solve the performance issues.
I have tested this on a room with 110,00 members. ~500ms which is same as the current channels.members endpoint.

apps/meteor/app/api/server/v1/rooms.ts
```diff
@@ -857,6 +859,48 @@ API.v1.addRoute(
...
+
+			const { status, filter } = this.queryParams;
+
+			const { members, total } = await findUsersOfRoomOrderedByRole({
+				rid: findResult._id,
+				...(status && { status: { $in: status } }),
+				skip,
+				limit,
+				filter,
+				sort: { ...(sort.rolePriority && { rolePriority: sort.rolePriority }), ...(sort.username && { username: sort.username }) },
```
KevLehman, line 891: Why do we need this? can't u just pass `sort` to the function?

apps/meteor/server/lib/findUsersOfRoomOrderedByRole.ts
```diff
@@ -0,0 +1,130 @@
...
+					...(exceptions.length > 0 && { $nin: exceptions }),
+				},
+				...(status && { status }),
+				...(filter && orStmt.length > 0 && { $or: orStmt }),
+			},
+			...extraQuery,
+		],
+	};
+
+	const membersResult = Users.col.aggregate(
```
KevLehman, line 65: Type the aggregation 👀 `Users.col.aggregate<>(`

apps/meteor/server/startup/migrations/v320.ts
```diff
@@ -0,0 +1,108 @@
...
+	(ROOM_ROLE_PRIORITY_MAP as { [key: IRole['_id']]: number })[role] ?? ROOM_ROLE_PRIORITY_MAP.default;
+
+export const calculateRoomRolePriorityFromRoles = (roles: IRole['_id'][]) => {
+	return roles.reduce((acc, roleId) => Math.min(acc, getRoomRolePriorityForRole(roleId)), ROOM_ROLE_PRIORITY_MAP.default);
+};
+
+const READ_BATCH_SIZE = 1000;
+const WRITE_BATCH_SIZE = 1000;
+
+async function assignRoomRolePrioritiesFromMap(userIdAndroomRolePrioritiesMap: Map<IUser['_id'], IUser['roomRolePriorities']>) {
```
KevLehman, line 17: Can you take a look at https://www.mongodb.com/docs/manual/reference/method/db.collection.initializeUnorderedBulkOp/#behavior and see if it works better with less code? Mongo says it kinda groups the operations https://www.mongodb.com/docs/v6.0/reference/method/Bulk/

KevLehman, line 17: Just to confirm, you got arch approval for doing the migration outside of major? 👀

apps/meteor/app/api/server/v1/rooms.ts
abhinavkrin: I took reference from the findUserOfRoom function. Maybe the intention was to give sort support only for username and rolePriority? Should be just pass the complete sort object?

apps/meteor/server/startup/migrations/v320.ts
```diff
@@ -0,0 +1,108 @@
...
+	version: 320,
+	name: 'Add "roomRolePriorities" field to users',
+	async up() {
+		let userIdAndroomRolePrioritiesMap = new Map<IUser['_id'], IUser['roomRolePriorities']>();
+
+		let skip = 0;
+
+		let hasMoreSubscriptions = true;
+		do {
+			const subscriptionsCursor = Subscriptions.find(
```
KevLehman, line 71: Can you check if you can change this iteration with something like:
```
const cursor = Subscriptions.find().batchSize(READ_BATCH_SIZE)

for await (const entry of cursor) { ... }

File: packages/models/src/models/Subscriptions.ts [hunk 17]
```diff
@@ -2045,4 +2073,20 @@ export class SubscriptionsRaw extends BaseRaw<ISubscription> implements ISubscri
...
+	async updateOneAndSyncRolePriority(query: Filter<ISubscription>, update: UpdateFilter<ISubscription>): Promise<UpdateResult> {
+		const updateResult = await this.updateOne(query, update);
+
+		const subscription = await this.findOne(query, { projection: { roles: 1, u: 1, rid: 1 } });
+
+		if (!subscription) {
+			return updateResult;
+		}
+
+		const rolePriority = subscription.roles ? calculateRoomRolePriorityFromRoles(subscription.roles) : ROOM_ROLE_PRIORITY_MAP.default;
```
KevLehman, line 2086: Ideally we shouldn't have dependencies between models, i think most of them were removed already. Can you, instead of this, just call the `addRoomRolePriority` where the other functions are called, so the call for users ends outside of the model?

KevLehman (changes requested):

packages/rest-typings/src/v1/rooms.ts
```diff
@@ -616,6 +616,53 @@ const roomsOpenSchema = {
...
+		},
+		roomName: {
+			type: 'string',
+		},
+		status: {
+			type: 'array',
+			items: {
+				type: 'string',
+			},
+			nullable: true,
```
KevLehman, line 641: You don't need the `nullable` on these as you're not making them required

apps/meteor/server/startup/migrations/v320.ts
```diff
@@ -0,0 +1,93 @@
...
+			existingPriorities[roomId] = priority;
+			userIdAndroomRolePrioritiesMap.set(userId, existingPriorities);
+
+			if (userIdAndroomRolePrioritiesMap.size >= WRITE_BATCH_SIZE) {
+				userIdAndroomRolePrioritiesMap = await assignRoomRolePrioritiesFromMap(userIdAndroomRolePrioritiesMap);
+			}
+		}
+
+		// Flush any remaining priorities in the map
+		await assignRoomRolePrioritiesFromMap(userIdAndroomRolePrioritiesMap);
```
KevLehman, line 91: Do you still need this? I thought you could initialize an unorderedop and just push things on there, and then consume the operation (since mongo says it groups it efficiently) so you don't have to do the manual batching 👀

KevLehman (changes requested):

apps/meteor/app/api/server/v1/rooms.ts
```diff
@@ -857,6 +859,48 @@ API.v1.addRoute(
 	},
 );
 
+API.v1.addRoute(
+	'rooms.membersOrderedByRole',
+	{ authRequired: true, validateParams: isRoomsMembersOrderedByRoleProps },
+	{
+		async get() {
+			const findResult = await findRoomByIdOrName({
```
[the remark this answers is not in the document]
abhinavkrin: Yes, you are right. this endpoint does not make sense for dm, livechat rooms. I have restricted this endpoint to only public and private channels. In other cases this endpoint will fail with a clear message of room type being not supported

apps/meteor/server/startup/migrations/v320.ts
abhinavkrin: It no longer includes migration. I have used a soft migration approach

packages/models/src/models/Rooms.ts
```diff
@@ -2206,4 +2206,12 @@ export class RoomsRaw extends BaseRaw<IRoom> implements IRoomsModel {
 			{ returnDocument: 'after' },
 		);
 	}
+
+	markRolePrioritesCreatedForRoom(rid: IRoom['_id']) {
+		return this.updateOne({ _id: rid }, { $set: { rolePrioritiesCreated: true } });
+	}
+
+	async hasCreatedRolePrioritiesForRoom(rid: IRoom['_id']) {
+		return (await this.findOne({ _id: rid, rolePrioritiesCreated: true })) !== null;
```
KevLehman, line 2215: Just return the `findOne` and let the caller do the condition check so you don't await and return.

KevLehman: (Or just run the `count` with the query and check it on the caller)

apps/meteor/app/lib/server/functions/syncRolePrioritiesForRoomIfRequired.ts
```diff
@@ -0,0 +1,93 @@
...
+	}
+
+	return userIdAndroomRolePrioritiesMap;
+}
+
+export const syncRolePrioritiesForRoomIfRequired = async (rid: IRoom['_id']) => {
+	let userIdAndroomRolePrioritiesMap = new Map<IUser['_id'], IUser['roomRolePriorities']>();
+
+	if (await Rooms.hasCreatedRolePrioritiesForRoom(rid)) {
+		return false;
```
KevLehman, line 59: I think you're not using the return values, you can return nothing instead.

apps/meteor/app/lib/server/functions/syncRolePrioritiesForRoomIfRequired.ts
```diff
@@ -0,0 +1,93 @@
...
+		const roomId = sub.rid;
+		const priority = calculateRoomRolePriorityFromRoles(sub.roles);
+
+		const existingPriorities = userIdAndroomRolePrioritiesMap.get(userId) || {};
+		existingPriorities[roomId] = priority;
+		userIdAndroomRolePrioritiesMap.set(userId, existingPriorities);
+
+		if (userIdAndroomRolePrioritiesMap.size >= WRITE_BATCH_SIZE) {
+			userIdAndroomRolePrioritiesMap = await assignRoomRolePrioritiesFromMap(userIdAndroomRolePrioritiesMap);
+		}
```
KevLehman, line 85: What i was saying before is you can create the bulk in here, and add stuff to the bulk, and then execute the bulkop. Mongo should be intelligent enough to batch it, similar to what you're doing here.

That should reduce a bit of code and make it more readable (worst case it doesn't batch stuff but i don't think so)

KevLehman (changes requested):

apps/meteor/app/api/server/v1/rooms.ts
```diff
@@ -857,6 +859,65 @@ API.v1.addRoute(
...
+					}).toArray(),
+				);
+
+				const rolesIds = roles.map(({ _id }) => _id);
+				rolesOrder.forEach((providedRole) => {
+					if (!rolesIds.includes(providedRole)) {
+						throw new Error(`role "${providedRole}" not found`);
+					}
+				});
+			}
```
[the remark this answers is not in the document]
abhinavkrin: Thanks Ricardo, I have used your suggestion in other places. 
No longer using this piece of code, so marking as resolved.

apps/meteor/app/api/server/v1/rooms.ts
```diff
@@ -857,6 +860,58 @@ API.v1.addRoute(
...
+			const findResult = await findRoomByIdOrName({
+				params: this.queryParams,
+				checkedArchived: false,
+			});
+
+			if (!(await canAccessRoomAsync(findResult, this.user))) {
+				return API.v1.notFound('The required "roomId" or "roomName" param provided does not match any room');
+			}
+
+			if (findResult.t !== 'c' && findResult.t !== 'p') {
```
MarcosSpessatto, line 877: We should use the helpers isPublicChannel and private

apps/meteor/app/lib/server/functions/syncRolePrioritiesForRoomIfRequired.ts
```diff
@@ -0,0 +1,74 @@
...
+	}
+
+	if (bulk.length > 0) {
+		await bulk.execute();
+	}
+}
+
+export const syncRolePrioritiesForRoomIfRequired = async (rid: IRoom['_id']) => {
+	const userIdAndroomRolePrioritiesMap = new Map<IUser['_id'], IUser['roomRolePriorities']>();
```
MarcosSpessatto, line 42:
```suggestion
	const userIdAndRoomRolePrioritiesMap = new Map<IUser['_id'], IUser['roomRolePriorities']>();
```

apps/meteor/server/lib/roles/syncRoomRolePriority.ts
```diff
@@ -0,0 +1,51 @@
+import type { IRole, IRoom, IUser } from '@rocket.chat/core-typings';
+import { ROOM_ROLE_PRIORITY_MAP } from '@rocket.chat/core-typings';
+import { Subscriptions, Users } from '@rocket.chat/models';
+
+/**
+ * Retrieves the role priority for a given role.
+ * @param role The role ID.
+ * @returns The priority of the role.
+ */
+export const getRoomRolePriorityForRole = (role: IRole['_id']): number =>
```
MarcosSpessatto, line 10: This function is duplicated, isn't it?

apps/meteor/server/lib/roles/syncRoomRolePriority.ts
```diff
@@ -0,0 +1,51 @@
...
+	ROOM_ROLE_PRIORITY_MAP[role as keyof typeof ROOM_ROLE_PRIORITY_MAP] ?? ROOM_ROLE_PRIORITY_MAP.default;
+
+/**
+ * Calculates the minimum role priority from a list of roles.
+ * @param roles The array of role IDs.
+ * @returns The minimum role priority.
+ */
+export const calculateRoomRolePriorityFromRoles = (roles: IRole['_id'][]): number =>
```
MarcosSpessatto, line 18: Same

apps/meteor/tests/end-to-end/api/rooms.ts
MarcosSpessatto, line 1: Can we create some test cases to cover when the room members don't have any roles?

MarcosSpessatto (changes requested):
We must test and cover the custom roles as well... On the admin panel, it's possible for users to create room-scoped roles, so we need to take care of it

apps/meteor/tests/end-to-end/api/rooms.ts
abhinavkrin: Hey, isn't a room need to have atleast one owner.

apps/meteor/app/lib/server/functions/syncRolePrioritiesForRoomIfRequired.ts
```diff
@@ -0,0 +1,68 @@
...
+import { Subscriptions, Users, Rooms } from '@rocket.chat/models';
+
+import { calculateRoomRolePriorityFromRoles } from '../../../../server/lib/roles/syncRoomRolePriority';
+
+const READ_BATCH_SIZE = 1000;
+
+async function assignRoomRolePrioritiesFromMap(userIdAndRoomRolePrioritiesMap: Map<IUser['_id'], IUser['roomRolePriorities']>) {
+	const bulk = Users.col.initializeUnorderedBulkOp();
+
+	for await (const [userId, roomRolePriorities] of userIdAndRoomRolePrioritiesMap.entries()) {
```
KevLehman, line 11:
```suggestion
	for (const [userId, roomRolePriorities] of userIdAndRoomRolePrioritiesMap.entries()) {
```

KevLehman: Or even better, you can https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map/forEach

apps/meteor/app/lib/server/functions/syncRolePrioritiesForRoomIfRequired.ts
```diff
@@ -0,0 +1,68 @@
...
+
+		const existingPriorities = userIdAndRoomRolePrioritiesMap.get(userId) || {};
+		existingPriorities[roomId] = priority;
+		userIdAndRoomRolePrioritiesMap.set(userId, existingPriorities);
+	}
+
+	// Flush any remaining priorities in the map
+	await assignRoomRolePrioritiesFromMap(userIdAndRoomRolePrioritiesMap);
+
+	await Rooms.markRolePrioritesCreatedForRoom(rid);
```
KevLehman, line 67: I wonder if we need the flag. Or can we do something like
```
db.users.find({ rooms: rid, roomRolePriorities.rid: { $exists: true }})
```

basically, assume that if we have a user in the room that already has role priorities, then the rest should have as well. That way we avoid storing and checking the flag. Would that work?

KevLehman (approved):

KevLehman (approved):

MartinSchoeler (approved):

apps/meteor/tests/end-to-end/api/rooms.ts
```diff
@@ -3633,4 +3636,531 @@ describe('[Rooms]', () => {
...
+			[ownerUser, moderatorUser, memberUser1, memberUser2] = await Promise.all([
+				createUser({ username: `a_${Random.id()}`, roles: ['admin'] }),
+				createUser({ username: `b_${Random.id()}` }),
+				createUser({ username: `c_${Random.id()}` }),
+				createUser({ username: `d_${Random.id()}` }),
+			]);
+
+			ownerCredentials = await login(ownerUser.username, password);
+
+			customRole = await createCustomRole({
```
MarcosSpessatto, line 3662: Should we check if enterprise here as well? I think it throws an error if not EE

abhinavkrin: It should not throw error but resolve with undefined if not EE.

MarcosSpessatto (approved):

PR #34156, opened 2024-12-10
fix: First response time livechat metrics are associated with the last agent who served the room

matheusbsilva137:

 - Fixed rooms' "first response time" metric being attached to the last agent who served a room instead of the first one -- this leads to incorrect values for the "Best first response time" and "Average first response time" metrics for each agent in case rooms are forwarded among agents/departments.

## Steps to test or reproduce

Make sure to grant the `transfer-livechat-guest` permission to agents or to a manager. 
Create a new agent and respond a livechat room with it, then transfer the room to another one. The "first response time" aggregation metrics are available under **Omnichannel > Analytics > Productivity** (or by using the `livechat/analytics/agent-overview` endpoint), and we need to make sure that both metrics related to the room that has been transferred are associated to the first agent who responded to the room (and not to the last one who served it).

## Further comments

[CORE-669]

[CORE-669]: https://rocketchat.atlassian.net/browse/CORE-669?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


KevLehman (changes requested):
Can u add unit tests cases for your fix?
if possible, similar tests to the API tests ones 👀

KevLehman (approved):

apps/meteor/tests/end-to-end/api/livechat/04-dashboards.ts
```diff
@@ -984,6 +984,60 @@ describe('LIVECHAT - dashboards', function () {
 			expect(originalFirstResponseTimeInSeconds).to.be.greaterThanOrEqual(firstDelayInSeconds);
 		});
 
+		it('should correctly associate the first response time to the first agent who responded the room', async () => {
+			const response = await startANewLivechatRoomAndTakeIt({ agent: forwardAgent.credentials });
+			roomId = response.room._id;
+
+			await sendAgentMessage(roomId, 'first response from agent', forwardAgent.credentials);
+
+			await request
```
MarcosSpessatto, line 993: Shouldn't we send a message with the new agent just to ensure it?

MarcosSpessatto (changes requested):

apps/meteor/tests/end-to-end/api/livechat/04-dashboards.ts
matheusbsilva137: Hmmm good idea, done!
This issue should happen even if the user who received the room didn't send any new message, but it's good to have this covered

KevLehman (approved):

MarcosSpessatto (approved):

PR #34158, opened 2024-12-10
fix: race condition in livechat popout mode cross origin

MartinSchoeler:

We had an issue that caused the livechat popout mode to not work when in cross domains, this fix was developed https://github.com/RocketChat/Rocket.Chat/pull/33944 but it had a race condition issue, trying to use window.postMessage() right after window.open() does not guarantee that the message will be received, and assuring that the page has loaded in cross domains can be very finicky, so I changed the approach to use the `checkPoppedOutWindow` callback to properly register the guest to the popover using something similar to the `setGuestToken` livechat api call

## Issue(s)

[SUP-708]

## Steps to test or reproduce

A good way to test this issue is to disable the registration form, that can cause the popover to have different tokens

[SUP-694]: https://rocketchat.atlassian.net/browse/SUP-694?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[SUP-708]: https://rocketchat.atlassian.net/browse/SUP-708?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/livechat/src/components/Screen/ScreenProvider.tsx
```diff
@@ -137,15 +138,26 @@ export const ScreenProvider: FunctionalComponent = ({ children }) => {
...
-	}, [dispatch]);
+
+		if (token && typeof token === 'string') {
+			if (registrationForm && !name && !email) {
+				dispatch({ token });
+			} else {
+				await evaluateChangesAndLoadConfigByFields(async () => {
+					await createOrUpdateGuest({ token });
+				});
+			}
```
MarcosSpessatto, line 158:
```suggestion
			if (registrationForm && !name && !email) {
				dispatch({ token });
				return;
			} 
			await evaluateChangesAndLoadConfigByFields(async () => {
					await createOrUpdateGuest({ token });
			});
```

packages/livechat/src/widget.ts
```diff
@@ -489,20 +489,13 @@ const api: InternalWidgetAPI = {
 		if (!config.url) {
 			throw new Error('Config.url is not set!');
 		}
+		const urlToken = token && `&token=${token}`;
+
 		api.popup = window.open(
-			`${config.url}${config.url.lastIndexOf('?') > -1 ? '&' : '?'}mode=popout`,
+			`${config.url}${config.url.lastIndexOf('?') > -1 ? '&' : '?'}mode=popout${urlToken}`,
```
MarcosSpessatto, line 495: If token is false or undefined this can cause problems like having the explicit `undefined` value in the URL.

MarcosSpessatto (changes requested):

packages/livechat/src/widget.ts
MartinSchoeler: When the widget is initialized a token is always generated, so it never is undefined

MarcosSpessatto: If its always defined, why are we checking it if its not undefined? 👀

MarcosSpessatto (changes requested):

MarcosSpessatto (approved):

dougfabris (approved):

PR #34165, opened 2024-12-11
fix: Deletion of 1st message in a thread causes notification to be up without showing why.

Gustrb:

This PR fixes the issue that would arise all the times the first (and only) message of a thread would get deleted. What was happening is that, we were not checking, on deletion, if all the messages of a thread had been deleted, which was making the UI feel unresponsive, since the 'Following' bell would show as if there were unread messages even if the message was deleted.
So, now, we do a couple checks to ensure consistency:
1. We decrease the counter of the number of thread messages in the parent message
2. We remove unread from the subscription of the deleted message, and, if some message was updated, we notify the client to refresh that data.
3. If the number of thread messages of the parent message is 0, we also notify the client to update the parent message, so we avoid showing that there are 1 thread message even if it was deleted.

## Steps to test or reproduce

You are going to need two users, let's call them user A and B, they both must be at the same room for you to be able to test.
1. As user A send a message in the room
2. As user B add a thread message in the room
3. As user B remove the thread message
4. Click on the user A browser so the websocket update takes place

### Expected:
- The thread should show the "View thread" button, and it show there is 0 replies, and there should be no visual representation of an unread message

## Further comments

[CONN-261](https://rocketchat.atlassian.net/browse/CONN-261)

[CONN-261]: https://rocketchat.atlassian.net/browse/CONN-261?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/meteor/app/lib/server/functions/deleteMessage.ts
```diff
@@ -107,3 +107,22 @@ export async function deleteMessage(message: IMessage, user: IUser): Promise<voi
...
+
+	if (room) {
+		const { modifiedCount } = await Subscriptions.removeUnreadThreadsByRoomId(room._id, [message.tmid]);
+		if (modifiedCount > 0) {
+			void notifyOnSubscriptionChangedByRoomIdAndUserId(room._id, user._id);
+		}
+	}
+
+	// Check if this was the last reply in the thread
+	const thread = await Messages.findOneById(message.tmid, { projection: { tcount: 1 } });
```
matheusbsilva137, line 122: what do you think about modifying `decreaseReplyCountById` so as to use a `findOneAndUpdate` with the `returnDocument: 'after'` option?
This way we'd avoid this extra query here :eyes:

matheusbsilva137 (changes requested):

apps/meteor/app/lib/server/functions/deleteMessage.ts
```diff
@@ -107,3 +107,22 @@ export async function deleteMessage(message: IMessage, user: IUser): Promise<voi
...
 	}
 }
+
+async function deleteThreadMessage(message: Required<Pick<IMessage, 'tmid'>>, user: IUser, room: IRoom | null): Promise<void> {
+	await Messages.decreaseReplyCountById(message.tmid, -1);
+
+	if (room) {
+		const { modifiedCount } = await Subscriptions.removeUnreadThreadsByRoomId(room._id, [message.tmid]);
+		if (modifiedCount > 0) {
+			void notifyOnSubscriptionChangedByRoomIdAndUserId(room._id, user._id);
```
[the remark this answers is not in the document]
Gustrb: Hmmm you're right

apps/meteor/tests/end-to-end/api/chat.ts
```diff
@@ -2005,6 +2005,50 @@ describe('[Chat]', () => {
...
+						roomId: testChannel._id,
+					})
+					.expect(200)
+					.expect((res) => {
+						expect(res.body).to.have.property('success', true);
+						expect(res.body.subscription).to.have.property('tunread');
+						expect(res.body.subscription.tunread).to.be.an('array');
+						expect(res.body.subscription.tunread).to.deep.equal([]);
+					});
+			});
```
matheusbsilva137, line 2050: Can we apply this same check, but using `otherUserCredentials`?

matheusbsilva137 (changes requested):

apps/meteor/tests/end-to-end/api/chat.ts
```diff
@@ -2005,6 +2005,50 @@ describe('[Chat]', () => {
...
+
+			before(async () => {
+				otherUser = await createUser();
+				otherUserCredentials = await login(otherUser.username, password);
+				parentThreadId = (await sendSimpleMessage({ roomId: testChannel._id })).body.message._id;
+			});
+
+			after(() => Promise.all([deleteUser(otherUser), deleteMessage({ msgId: parentThreadId, roomId: testChannel._id })]));
+
+			it('should reset the unread counter if the message was removed', async () => {
```
KevLehman, line 2022: can you add a check for a user following the thread?

KevLehman (changes requested):

apps/meteor/app/lib/server/functions/deleteMessage.ts
```diff
@@ -107,3 +107,23 @@ export async function deleteMessage(message: IMessage, user: IUser): Promise<voi
...
+
+	if (room) {
+		const { modifiedCount } = await Subscriptions.removeUnreadThreadsByRoomId(room._id, [message.tmid]);
+		if (modifiedCount > 0) {
+			const userIdsThatAreWatchingTheThread = [...new Set([user._id, ...(message.replies || [])])];
+			void notifyOnSubscriptionChangedByRoomIdAndUserIds(room._id, userIdsThatAreWatchingTheThread);
+		}
+	}
+
+	// If we could not find the parent message, there is no need to notify whoever is listening
```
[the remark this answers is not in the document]
Gustrb: nvm this comment is a leftover, I'll remove it

apps/meteor/app/lib/server/functions/deleteMessage.ts
```diff
@@ -107,3 +107,23 @@ export async function deleteMessage(message: IMessage, user: IUser): Promise<voi
...
 	}
 }
+
+async function deleteThreadMessage(message: IThreadMessage, user: IUser, room: IRoom | null): Promise<void> {
+	const { value: updatedParentMessage } = await Messages.decreaseReplyCountById(message.tmid, -1);
+
+	if (room) {
+		const { modifiedCount } = await Subscriptions.removeUnreadThreadsByRoomId(room._id, [message.tmid]);
+		if (modifiedCount > 0) {
+			const userIdsThatAreWatchingTheThread = [...new Set([user._id, ...(message.replies || [])])];
```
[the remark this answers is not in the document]
Gustrb: I'll have to test it, but I'm pretty sure they do appear on the lis

apps/meteor/tests/end-to-end/api/chat.ts
Gustrb: sure

MarcosSpessatto (approved):

MarcosSpessatto (approved):

KevLehman (approved):

matheusbsilva137 (approved):

PR #34180, opened 2024-12-15
feat: Add "Jump to Message" option for files

abirc8010:
[Screencast from 2024-12-15 12-27-57.webm](https://github.com/user-attachments/assets/7bdd30f5-092a-49b5-b0e2-489e5d20609f)

## Issue(s)
RocketChat/feature-requests#915

## Steps to test or reproduce

• Go to the All files option.
• Choose a file and click on its kebab menu.
• Select Jump to message.


abirc8010:
Currently, the feature works for messages sent after these changes are implemented, as I am adding a `messageId` attribute to the `Uploads`model. 
For files sent before these changes, which don’t have a `messageId`, the functionality can be achieved by searching for the file by its ID in threads and normal messages. However, I am uncertain about the performance impact of this approach.

PR #34191, opened 2024-12-16
feat: Allow granting the `mobile-upload-file` permission to guests

matheusbsilva137:

 - Allow granting the `mobile-upload-file` permission to the guest role (but not make it the default behavior)

## Steps to test or reproduce

Use the `authorization:addPermissionToRole` Meteor method to check that the `mobile-upload-file` permission can now be granted to the guest role. **Example (cURL) request:**

```
curl --request POST \
  --url http://localhost:3000/api/v1/method.call/authorization%3AaddPermissionToRole \
  --header 'X-Auth-Token: jVrwkAmobgfne3aEdL-dJh-znNXMkhg79pEAZS2rEIg' \
  --header 'X-User-Id: rMq9ovRmZabXHW4Do' \
  --data '{"message":"{\"msg\":\"method\",\"id\":\"16\",\"method\":\"authorization:addPermissionToRole\",\"params\":[\"mobile-upload-file\",\"guest\"]}"}'
```

This Meteor method should return an error (`Permission is restricted [error-action-not-allowed]`) in case an invalid permission is provided (e.g. a permission that can't be granted to the guest role -- any permission that isn't in this list: `view-d-room`, `view-joined-room`, `view-p-room`, `start-discussion`, `mobile-upload-file`).

Use the `authorization:removeRoleFromPermission` Meteor method to check that the `mobile-upload-file` permission can also be removed from the guest role. **Example (cURL) request:**

```
curl --request POST \
  --url http://localhost:3000/api/v1/method.call/authorization%3AremoveRoleFromPermission \
  --header 'X-Auth-Token: jVrwkAmobgfne3aEdL-dJh-znNXMkhg79pEAZS2rEIg' \
  --header 'X-User-Id: rMq9ovRmZabXHW4Do' \
  --data '{"message":"{\"msg\":\"method\",\"id\":\"16\",\"method\":\"authorization:addPermissionToRole\",\"params\":[\"mobile-upload-file\",\"guest\"]}"}'
```

## Further comments

[CORE-837]

[CORE-837]: https://rocketchat.atlassian.net/browse/CORE-837?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/meteor/app/authorization/server/methods/addPermissionToRole.ts
```diff
@@ -15,6 +16,10 @@ declare module '@rocket.chat/ddp-client' {
 
 Meteor.methods<ServerMethods>({
 	async 'authorization:addPermissionToRole'(permissionId, role) {
+		if (role === 'guest' && !AuthorizationUtils.hasRestrictionsToRole(role) && (await License.hasValidLicense())) {
+			AuthorizationUtils.addRolePermissionWhiteList(role, await License.getGuestPermissions());
+		}
```
matheusbsilva137, line 21: For some reason EE tests were failing on CI because we didn't have any restrictions stored for the guest role (which didn't happen locally) :man_shrugging: so I went through all of the places where we populate such restrictions and couldn't find anything off, please tell me if you do
So I decided to reinforce these restrictions in the endpoint/method itself (which I believe is harmless) and create another task to identify what could be causing this issue later

apps/meteor/app/authorization/server/constant/permissions.ts
```diff
@@ -67,6 +67,8 @@ export const permissions = [
 	{ _id: 'set-owner', roles: ['admin', 'owner'] },
 	{ _id: 'send-many-messages', roles: ['admin', 'bot', 'app'] },
 	{ _id: 'set-leader', roles: ['admin', 'owner'] },
+	{ _id: 'start-discussion', roles: ['admin', 'user', 'guest', 'app'] },
+	{ _id: 'start-discussion-other-user', roles: ['admin', 'user', 'owner', 'app'] },
```
matheusbsilva137, line 71: I had to add the `start-discussion` permission to this file so that it could be restored on my new end-to-end tests (with `restorePermissionToRoles`). We're currently inserting this permission somewhere else (on `Meteor.startup`) and it was missing here
`start-discussion-other-user` is not related to my tests, but I decided to bring this here too since it's created in the same `Meteor.startup` callback (so we avoid one more tech debt)

PR #34197, opened 2024-12-17
fix: Switch language button in login page not displaying language name

gabriellsh:

Before:
image

After:
image

## Issue(s)

[CORE-894](https://rocketchat.atlassian.net/browse/CORE-894)
## Further comments

[CORE-894]: https://rocketchat.atlassian.net/browse/CORE-894?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/meteor/tests/e2e/login.spec.ts
```diff
@@ -50,4 +50,10 @@ test.describe.parallel('Login', () => {
 			await expect(poUtils.mainContent).toBeVisible();
 		});
 	});
+
+	test('Should correctly display switch language button', async ({ page }) => {
+		test.use({ locale: 'pt-BR' });
+
+		await expect(page.getByText('Change to Default')).toBeVisible();
```
dougfabris, line 57: instead of asserting the button by text, why not testing the button behavior and checking if the language changed, and doing the query by the button name? so its a win-win

dougfabris (changes requested):

apps/meteor/tests/e2e/login.spec.ts
gabriellsh: I could have indeed tested clicking the button, but I didn't because... It didn't cross my mind :melting_face:

gabriellsh: fixed

dougfabris (approved):
LGTM!

PR #34199, opened 2024-12-17
refactor: Remove settings dependency on startup

sampaiodiego:

The idea is to remove code dependencies to "settings cached" during the startup that are spread everywhere in the codebase. This is basically but not only represented by code using `settings` in the file's top level scope.

The currently method to make sure the settings cache is built before code is run is importing the "cached settings" from the file that builds the cache, so the top level await will ensure the cache is built before everything else runs. The proposal here is that we don't rely on this mechanism anymore by having a more clear code path, so we first build the cache and after that we import everything else.

So this is the first step of many to come, the changes are basically moving code relying on `settings` to functions and then calling this functions on the appropriate moment.

## Issue(s)

[ARCH-1398]

## Further comments

[ARCH-1398]: https://rocketchat.atlassian.net/browse/ARCH-1398?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/meteor/server/main.ts
```diff
@@ -1,18 +1,28 @@
...
  * and the startup should be done in parallel
  */
 import './settings';
-import '../app/settings/server';
 
 import { configureLoginServices } from './configuration';
 import { configureLogLevel } from './configureLogLevel';
 import { registerServices } from './services/startup';
 import { startup } from './startup';
+import { configureBoilerplate } from './startup/configureBoilerplate';
```
pierre-lehnen-rc, line 14: Why not use `/configuration/` instead of `/startup/`?

sampaiodiego: @pierre-lehnen-rc I've moved everything to `/configuration`, it makes much more sense, thx

pierre-lehnen-rc (approved):

KevLehman (dismissed):

KevLehman (approved):

pierre-lehnen-rc (approved):

PR #34204, opened 2024-12-17
refactor: remove Meteor calls for message history methods

MarcosSpessatto:
https://rocketchat.atlassian.net/browse/CORE-897


apps/meteor/server/publications/messages.ts
```diff
@@ -173,11 +189,89 @@ export async function handleCursorPagination(
 	}
 
 	return {
-		[type.toLowerCase()]: response,
+		updated: type === 'UPDATED' ? response : [],
+		deleted: type === 'DELETED' ? response : [],
```
KevLehman, line 193: Do we need the refactor? Or can we just create more detailed types to cover what was there before? 👀

apps/meteor/app/api/server/v1/chat.ts
```diff
@@ -100,17 +101,20 @@ API.v1.addRoute(
...
 
-			const result = await Meteor.callAsync('messages/get', roomId, getMessagesQuery);
+			const result = await getMessageHistory(roomId, this.userId, getMessagesQuery);
 
 			if (!result) {
 				return API.v1.failure();
 			}
+			if (typeof result === 'boolean') {
+				return API.v1.success({ result: {} });
+			}
```
KevLehman, line 111: maybe instead of `boolean`, this one could return `false` so the check above can cover it and return failure instead? would that be a breaking change? would this one be a breaking change? 👀

apps/meteor/server/publications/messages.ts
MarcosSpessatto: I think we can keep as-is and refactor the types later

KevLehman (approved):

PR #34208, opened 2024-12-17
feat: `rooms.hide` endpoint

aleksandernsilva:
This PR introduces a new endpoint `rooms.hide` responsible for hiding rooms without type-specific restrictions.

## Further comments
[ARCH-1340](https://rocketchat.atlassian.net/browse/ARCH-1340)

[ARCH-1340]: https://rocketchat.atlassian.net/browse/ARCH-1340?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/meteor/app/api/server/v1/rooms.ts
```diff
@@ -926,7 +926,7 @@ API.v1.addRoute(
 			const user = await Users.findOneById(this.userId, { projections: { username: 1 } });
 
 			if (!user) {
-				throw new Meteor.Error('error-invalid-user', 'Invalid user', { method: 'rooms.close' });
+				return API.v1.failure('Invalid user');
```
KevLehman, line 929: This is a breaking change :/

(Not removing the Meteor error, but changing the error message. We should go with the `error-invalid-user` as it can be used to translate it on UI)

apps/meteor/tests/end-to-end/api/rooms.ts
```diff
@@ -3687,16 +3700,56 @@ describe('[Rooms]', () => {
 				.end(done);
 		});
 
-		it('should fail if roomId is not provided', async () => {
-			await request
+		it('should fail if roomId is not provided', (done) => {
+			void request
```
KevLehman, line 3704: write tests with promises instead :)

apps/meteor/app/api/server/v1/rooms.ts
KevLehman: nvm im dum

apps/meteor/app/api/server/v1/rooms.ts
```diff
@@ -962,3 +964,37 @@ API.v1.addRoute(
 		},
 	},
 );
+
+API.v1.addRoute(
+	'rooms.close',
```
KevLehman, line 969: wonder why this is called `.close` if we call `hideRoom` inside 🤔

apps/meteor/app/api/server/v1/rooms.ts
```diff
@@ -962,3 +964,37 @@ API.v1.addRoute(
...
+				return API.v1.unauthorized();
+			}
+
+			const user = await Users.findOneById(this.userId, { projections: { username: 1 } });
+
+			if (!user) {
+				return API.v1.failure('error-invalid-user');
+			}
+
+			const subscription = await Subscriptions.findOneByRoomIdAndUserId(roomId, this.userId);
```
KevLehman, line 985: You can pass a projection here iirc (not sure if the method accepts)

KevLehman (changes requested):

apps/meteor/app/api/server/v1/rooms.ts
aleksandernsilva: I wondered the same thing on the FE. Client-side it is "hide room", server side "close room" (`channels.close`, `dm.close`), probably as a counterpart to the "open room" (`channels.open`, `dm.open`).

Considering this is an old name convention, do you think it is worth it renaming it to `rooms.hide` and align FE and BE?

aleksandernsilva: Going with `rooms.hide` to keep the endpoint name true to its function.

apps/meteor/app/api/server/v1/rooms.ts
```diff
@@ -962,3 +964,37 @@ API.v1.addRoute(
...
+			}
+
+			const subscription = await Subscriptions.findOneByRoomIdAndUserId(roomId, this.userId, { projection: { open: 1 } });
+
+			if (!subscription) {
+				return API.v1.failure(`error-invalid-subscription`);
+			}
+
+			if (!subscription.open) {
+				return API.v1.failure('error-room-already-closed');
```
KevLehman, line 992:
```suggestion
				return API.v1.failure('error-room-already-hidden');
```

apps/meteor/app/api/server/v1/rooms.ts
```diff
@@ -962,3 +964,37 @@ API.v1.addRoute(
...
+				return API.v1.unauthorized();
+			}
+
+			const user = await Users.findOneById(this.userId, { projections: { username: 1 } });
+
+			if (!user) {
+				return API.v1.failure('error-invalid-user');
+			}
+
+			const subscription = await Subscriptions.findOneByRoomIdAndUserId(roomId, this.userId, { projection: { open: 1 } });
```
KevLehman, line 985: I'm thinking you could create a new subs method called `findOneOpenByRoomIdAndUserId` that checks if open is true, so you don't fetch the Sub if it's already closed

apps/meteor/app/api/server/v1/rooms.ts
```diff
@@ -962,3 +964,37 @@ API.v1.addRoute(
...
+	{ authRequired: true, validateParams: isRoomsHideProps },
+	{
+		async post() {
+			const { roomId } = this.bodyParams;
+
+			if (!(await canAccessRoomIdAsync(roomId, this.userId))) {
+				return API.v1.unauthorized();
+			}
+
+			const user = await Users.findOneById(this.userId, { projections: { username: 1 } });
```
KevLehman, line 979: Don't see you using the username, you can just pass `_id: 1` to the projection

KevLehman (changes requested):

KevLehman (approved):

apps/meteor/tests/end-to-end/api/rooms.ts
```diff
@@ -4204,4 +4204,93 @@ describe('[Rooms]', () => {
...
+		before(async () => {
+			memberA = await createUser();
+			memberB = await createUser();
+			nonMember = await createUser();
+			nonMemberCredentials = await login(nonMember.username, password);
+		});
+
+		before(async () => {
+			roomA = (await createRoom({ type: 'c', name: roomName, members: [memberA.username, memberB.username] })).body.channel;
+			roomB = (await createRoom({ type: 'd', username: memberB.username })).body.room;
```
MarcosSpessatto, line 4226: Shouldn't we delete this room too?

MarcosSpessatto (changes requested):

MarcosSpessatto (approved):

KevLehman (approved):

PR #34210, opened 2024-12-18
fix: New livechat conversations are not assigned to contact manager

matheusbsilva137:

 - Consider the `contactManager` property stored in **contacts** when deciding the default new agent to assign to a livechat room (instead of only considering the `contactManager` field in visitors, which is not maintained anymore).

## Steps to test or reproduce

In a **Premium** workspace, enable the "Assign new conversations to the contact manager" setting under **Settings > Omnichannel > Routing**. Then create a new contact manually (in the livechat contact center) using a livechat agent and assign yourself user as the contact manager (use the `contacts.get` endpoint to make sure the `contactManager` field contains the user id of the agent who created the contact). Then start a new livechat conversation using the info from the contact that has just been created.
**Expected result:** the contact manager should be automatically assigned to the room if it is online.

**Note:** use "Auto Selection" as the "Omnichannel Routing Method" (under **Settings > Omnichannel > Routing**). Otherwise, only bot contact managers will be auto assigned to rooms.

## Further comments

[CORE-862]

[CORE-862]: https://rocketchat.atlassian.net/browse/CORE-862?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/meteor/app/livechat/server/lib/LivechatTyped.ts
```diff
@@ -353,11 +353,12 @@ class LivechatClass {
 			throw new Meteor.Error('error-omnichannel-is-disabled');
 		}
 
-		if (await LivechatContacts.isChannelBlocked(this.makeVisitorAssociation(visitor._id, roomInfo.source))) {
+		const visitorAssociation = this.makeVisitorAssociation(visitor._id, roomInfo.source);
```
KevLehman, line 356: Is this part of the fix or is another bug? 👀

apps/meteor/lib/callbacks.ts
```diff
@@ -118,7 +119,7 @@ type ChainedCallbackSignatures = {
 	) => Promise<T>;
 
 	'livechat.beforeRouteChat': (inquiry: ILivechatInquiryRecord, agent?: { agentId: string; username: string }) => ILivechatInquiryRecord;
-	'livechat.checkDefaultAgentOnNewRoom': (agent: SelectedAgent, visitor?: ILivechatVisitor) => SelectedAgent | null;
+	'livechat.checkDefaultAgentOnNewRoom': (agent: SelectedAgent, visitor?: ILivechatContactVisitorAssociation) => SelectedAgent | null;
```
KevLehman, line 122: This one is no longer a visitor isn't it?

apps/meteor/ee/app/livechat-enterprise/server/hooks/handleNextAgentPreferredEvents.ts
```diff
@@ -111,7 +121,7 @@ callbacks.add(
 			return defaultAgent;
 		}
 
-		const guestAgent = lastAgent?.username && getDefaultAgent(lastAgent?.username);
+		const guestAgent = lastAgent?.username && getDefaultAgent({ username: lastAgent?.username });
```
KevLehman, line 124:
```suggestion
		const guestAgent = lastAgent?.username && getDefaultAgent({ username: lastAgent.username });
```

apps/meteor/ee/app/livechat-enterprise/server/hooks/handleNextAgentPreferredEvents.ts
```diff
@@ -18,12 +18,15 @@ const normalizeDefaultAgent = (agent?: Pick<IUser, '_id' | 'username'> | null):
...
-const getDefaultAgent = async (username?: string): Promise<SelectedAgent | null> => {
-	if (!username) {
+const getDefaultAgent = async ({ username, id }: { username?: string; id?: string }): Promise<SelectedAgent | null> => {
+	if (!username && !id) {
 		return null;
 	}
 
-	return normalizeDefaultAgent(await Users.findOneOnlineAgentByUserList(username, { projection: { _id: 1, username: 1 } }));
+	if (id) {
+		return normalizeDefaultAgent(await Users.findOneOnlineAgentById(id, { projection: { _id: 1, username: 1 } }));
```
KevLehman, line 27: These 2 calls should pass the setting value too. The livechat_enabled_when_agent_idle. Not sure if it should be on this PR or in another one 🤔

KevLehman (changes requested):
Tests? :please:

apps/meteor/app/livechat/server/lib/LivechatTyped.ts
matheusbsilva137: Hmm it's part of the fix. I just split this into a new var since it'll be used in the next line and reused by the callback

apps/meteor/lib/callbacks.ts
matheusbsilva137: We call the visitor association just `visitor` in a bunch of places, but I agree it's better to have a more descriptive name here. I'll change it

matheusbsilva137: Actually, I didn't need this param anymore :partying_face:

apps/meteor/ee/app/livechat-enterprise/server/hooks/handleNextAgentPreferredEvents.ts
matheusbsilva137: hmmm wdyt about leaving this to another task? This way we can focus on better refine and testing this
I'll create a task

KevLehman: Nice, pls create task 🙏🏽

apps/meteor/server/models/raw/Users.js
```diff
@@ -1657,11 +1657,11 @@ export class UsersRaw extends BaseRaw {
 		return this.findOne(query);
 	}
 
-	findOneOnlineAgentById(_id, isLivechatEnabledWhenAgentIdle) {
+	findOneOnlineAgentById(_id, options, isLivechatEnabledWhenAgentIdle) {
```
MarcosSpessatto, line 1660: Would it not be better to have options as the last parameter to follow the other cases?

apps/meteor/tests/end-to-end/api/livechat/24-routing.ts
```diff
@@ -91,9 +126,20 @@ import { IS_EE } from '../../../e2e/config/constants';
 			expect(roomInfo.servedBy).to.be.an('object');
 			expect(roomInfo.servedBy?._id).to.not.be.equal(testUser2.user._id);
 		});
+		(IS_EE ? it : it.skip)('should route to contact manager if it is online', async () => {
+			const room = await createLivechatRoom(visitor.token);
+
+			await sleep(5000);
```
MarcosSpessatto, line 132: Is this really necessary?

apps/meteor/tests/end-to-end/api/livechat/24-routing.ts
```diff
@@ -91,9 +126,20 @@ import { IS_EE } from '../../../e2e/config/constants';
 			expect(roomInfo.servedBy).to.be.an('object');
 			expect(roomInfo.servedBy?._id).to.not.be.equal(testUser2.user._id);
 		});
+		(IS_EE ? it : it.skip)('should route to contact manager if it is online', async () => {
```
MarcosSpessatto, line 129: I think we should also add the information that `Livechat_assign_new_conversation_to_bot` is true, right?

MarcosSpessatto, line 129: Can we create a test case when the contact manager is not online?

apps/meteor/tests/end-to-end/api/livechat/24-routing.ts
MarcosSpessatto, line 1: Also, can we create a test case to cover the scenario found by the QA team where:
`While visitor attempts to send a second message, it gets redirected to the agent, but on first attempt, it throws error.`

MarcosSpessatto (changes requested):

apps/meteor/tests/end-to-end/api/livechat/24-routing.ts
matheusbsilva137: We're doing this in all other routing tests in this describe :thinking: shouldn't we keep doing this, then?

apps/meteor/server/models/raw/Users.js
matheusbsilva137: You're right, not sure why I've done that

apps/meteor/tests/end-to-end/api/livechat/24-routing.ts
matheusbsilva137: This was happening on CE and is already being caught by our tests. CI failed due to this same condition by the time our QA team reported this, so we should be fine :)

apps/meteor/tests/end-to-end/api/livechat/24-routing.ts
matheusbsilva137: yeah, I'm just afraid of removing this and creating a flaky test :cry: 
but I believe we could try removing all sleeps from these tests at some point. I believe we're running all callbacks sync :eyes:

MarcosSpessatto (approved):

KevLehman (approved):

PR #34214, opened 2024-12-18
fix: validation on bot field used by js.sdk

sampaiodiego:

Change the validation of the `bot` field to make it compatible with our js.SDK: https://github.com/RocketChat/Rocket.Chat.js.SDK/blob/master/src/lib/message.ts#L17


apps/meteor/app/lib/server/methods/sendMessage.ts
```diff
@@ -132,7 +132,7 @@ Meteor.methods<ServerMethods>({
 			ts: Match.Maybe(Date),
 			t: Match.Maybe(String),
 			otrAck: Match.Maybe(String),
-			bot: Match.Maybe(Boolean),
+			bot: Match.Maybe(Object),
```
KevLehman, line 135: image


(Can we do Match.maybe(boolean | Object)?

KevLehman (approved):

PR #34215, opened 2024-12-18
fix: Start cronjobs after the database has been migrated

KevLehman:
## Issue(s)

https://rocketchat.atlassian.net/browse/CORE-743


apps/meteor/server/startup/cron.ts
```diff
@@ -12,14 +11,13 @@ import { videoConferencesCron } from '../cron/videoConferences';
 
 const logger = new Logger('SyncedCron');
 
-Meteor.defer(async () => {
+export const startCronJobs = async (): Promise<void> => {
 	await startCron();
```
lucas-a-pelegrino, line 15: Maybe we could compile this into a `Promise.all()`?

Not really an issue, just a suggestion.

lucas-a-pelegrino (approved):

MarcosSpessatto (approved):

PR #34217, opened 2024-12-18
fix: New version banner text translation

jeanfbrito:
## Steps to test or reproduce

To test it, edit `apps/meteor/app/utils/rocketchat.info` with a lower version like 3.0.0 so the banner will appear.

## Further comments

[CORE-849]

[CORE-849]: https://rocketchat.atlassian.net/browse/CORE-849?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


dougfabris (changes requested):
if its a fix it should have a changeset

dougfabris (approved):
LGTM!

Issue #914, opened 2015-09-27
Auto-scroll breaks on lazy image loading


joepie91:
When saying `@rocketbot map me The Netherlands`, the following happens:

![selection_426](https://cloud.githubusercontent.com/assets/1663259/10122790/47b4c1cc-6524-11e5-9cda-28d446305b35.png)

It only scrolls down far enough to account for the text, but not for the corresponding image(s). The images fall below the fold. This is likely because the scroll happens *before* the images are completely loaded, and it doesn't scroll again *after* they are loaded.

What it *should* look like:

![selection_427](https://cloud.githubusercontent.com/assets/1663259/10122792/742b9b68-6524-11e5-8585-18772096792d.png)

engelgabriel:
Fixed, a long time ago.

PR #34844, opened 2024-12-29
feat: line numbers in code blocks

Harshbansal8705:
This commit adds line numbers in the code when sent. Since hightlight.js doesn't natively support line numbers, this commit implements a custom solution to prepend the line numbers to each line of code manually.

Since highligh.js doesn't support line numbers, this PR implements it by adding line numbers manually to each line of code. It also ensures that line numbers are not selected when someone is trying to select and copy the code.

This is the updated screenshot of new code block:
image
image

## Issue(s)
Closes: [#914](https://github.com/RocketChat/feature-requests/issues/914)

### DIFF
--- a/packages/gazzodown/src/code/CodeBlock.tsx
+++ b/packages/gazzodown/src/code/CodeBlock.tsx
@@ -14,7 +14,7 @@ const CodeBlock = ({ lines = [], language }: CodeBlockProps): ReactElement => {
 
 	const { highlightRegex } = useContext(MarkupInteractionContext);
 
-	const code = useMemo(() => lines.map((line) => line.value.value).join('\n'), [lines]);
+	const code = useMemo(() => lines.map((line, index) => `<span>${index + 1}</span> ${line.value.value}`).join('\n'), [lines]);
 
 	const content = useMemo(() => {
 		const regex = highlightRegex?.();
@@ -41,7 +41,7 @@ const CodeBlock = ({ lines = [], language }: CodeBlockProps): ReactElement => {
 			);
 		}
 
-		return code;
+		return <div dangerouslySetInnerHTML={{ __html: code }} />;
 	}, [code, highlightRegex]);
 
 	useLayoutEffect(() => {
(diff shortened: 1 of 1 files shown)


reetp:
Can you note that there probably should be an option of other indicators on this.

Perhaps a different way of marking up - different character sequence? Eg markup if it sees code enclosed with say `@@@` or somesuch?

We frequently copy pasted sections of code and don't want to be trying to strip line numbers from it.

It break established behaviour for all users.

It would also be a massive retrograde step for me and my users.

The line numbers are largely irrelevant to us.

Case 1. We normally just paste sections of code, so line numbers may be 250-269 so numbering from say 1 would be useless/irrelevant/misleading

Case 2. If we really want line numbers we can just `cat -n` a file in a terminal and copy/paste.

Harshbansal8705:
I reviewed your comments and noted the concern regarding how it's handled in Slack. I couldn't find a similar feature in Slack either.


> We frequently copy pasted sections of code and don't want to be trying to strip line numbers from it.

In this case, it shouldn't cause any issues, as line numbers are currently not even present. Introducing line numbers would allow users to refer to specific lines, which could be beneficial.

> Perhaps a different way of marking up - different character sequence? Eg markup if it sees code enclosed with say `@@@` or somesuch?

Would it be feasible to add an option to toggle the line numbers instead?

> It break established behaviour for all users.

I'm not sure how this would disrupt the established behavior for users. Line numbers are merely displayed and do not interfere with copying, which aligns with the current behavior.

> The line numbers are largely irrelevant to us.

If line numbers are not needed, then maybe we could resolve the linked issue by adding an option to toggle them on or off?

reetp:
> I reviewed your comments and noted the concern regarding how it's handled in Slack

I never mentioned Slack. No idea how it works there.

> Introducing line numbers would allow users to refer to specific lines, which could be beneficial.

It could, but as I pointed out the line numbers could cause confusion if they do not align with the line numbers in the file.

If I paste line numbers 511-520 but your code numbers them 1-10 you can see it may be an issue.

> Line numbers are merely displayed and do not interfere with copying, which aligns with the current behavior.

Remember that current behaviour is no line numbers at all....

> Would it be feasible to add an option to toggle the line numbers instead?

I don't know. You are the one coding, not me. I suggested it should be an option.

Harshbansal8705:
> I never mentioned Slack. No idea how it works there.

Yeah, it was mentioned in the issue itself.

reetp:
> > I never mentioned Slack. No idea how it works there.
> 
> Yeah, it was mentioned in the issue itself.

So if you read that you will see that Slack has a key sequence to insert code with numbers.

How are you implementing that?

PR #34847, opened 2024-12-30
feat: add configurable retry with exponential backoff to HTTP requests

jannatkhandev:
### Add HTTP Retry Mechanism

The HTTP accessor in the Apps Engine now supports automatic retries with exponential backoff for failed requests. This feature helps apps handle transient network failures and API rate limits more gracefully.

### Configuration

The retry mechanism is disabled by default and can be enabled per request through the retry configuration option in the HTTP request options.

#### Default Configuration
```typescript
{
  enabled: false,    // Retry mechanism disabled by default
  maxAttempts: 3,    // Maximum number of attempts including the initial request
  initialDelay: 1000 // Initial delay in milliseconds before first retry
}
```

#### Configuration Options

| Option | Type | Default | Description |
|--------|------|---------|-------------|
| enabled | boolean | false | Enables/disables the retry mechanism |
| maxAttempts | number | 3 | Maximum number of attempts including the initial request |
| initialDelay | number | 1000 | Initial delay in milliseconds before first retry |

#### Basic Usage with Default Settings
```typescript
import { IHttp } from '@rocket.chat/apps-engine/definition/accessors';

export async function makeRequest({ http }: { http: IHttp }) {
    const response = await http.get('https://api.example.com/data', {
        retry: {
            enabled: true // Uses default maxAttempts and initialDelay
        }
    });
    return response;
}
```

#### Custom Configuration
```typescript
import { IHttp } from '@rocket.chat/apps-engine/definition/accessors';

export async function makeRequest({ http }: { http: IHttp }) {
    const response = await http.get('https://api.example.com/data', {
        retry: {
            enabled: true,
            maxAttempts: 5,     // Try up to 5 times
            initialDelay: 2000  // Start with 2 second delay
        }
    });
    return response;
}
```

### How It Works

When enabled, the retry mechanism will:

1. Attempt the HTTP request
2. If the request fails:
   - If retries are disabled or max attempts reached: throw the error
   - Otherwise: wait for the calculated delay and retry
3. For each retry:
   - The delay increases exponentially: initialDelay * (2 ^ attemptNumber)
   - Example with initialDelay = 1000:
     - First retry: 1000ms delay
     - Second retry: 2000ms delay
     - Third retry: 4000ms delay

### Error Handling

The final error from the last retry attempt will be thrown after all retries are exhausted. This error maintains the original structure and stack trace from the failed request.
```typescript
try {
    await http.get('https://api.example.com/data', {
        retry: {
            enabled: true,
            maxAttempts: 3
        }
    });
} catch (error) {
    // Error from the last failed attempt
    console.error('All retries failed:', error);
}
```

### Best Practices

1. *Selective Usage*: Enable retries only for operations that are safe to retry (idempotent operations)
2. *Timeout Configuration*: Consider setting appropriate request timeouts along with retry configuration
3. *Monitoring*: Monitor retry patterns to identify recurring issues with external services

### Interface Definition
```typescript
interface IHttpRetryConfig {
    enabled?: boolean;      // Enable/disable retries
    maxAttempts?: number;   // Maximum number of attempts
    initialDelay?: number;  // Initial delay in milliseconds
}

interface IHttpRequest {
    // ... existing options ...
    retry?: IHttpRetryConfig;
}
```

### Implementation Details

The retry mechanism is implemented in the Http class of the Apps Engine. It preserves all existing functionality including:
- Pre-request handlers
- Pre-response handlers
- Default headers and parameters
- Error handling

Each retry attempt is treated as a new request, running through all pre-request and pre-response handlers to maintain consistency with the original request behavior.

### DIFF
--- a/packages/apps-engine/src/server/accessors/Http.ts
+++ b/packages/apps-engine/src/server/accessors/Http.ts
@@ -1,9 +1,15 @@
-import type { IHttp, IHttpExtend, IHttpRequest, IHttpResponse } from '../../definition/accessors';
+import type { IHttp, IHttpExtend, IHttpRequest, IHttpResponse, IHttpRetryConfig } from '../../definition/accessors';
 import { RequestMethod } from '../../definition/accessors';
 import type { AppBridges } from '../bridges/AppBridges';
 import type { AppAccessorManager } from '../managers/AppAccessorManager';
 
 export class Http implements IHttp {
+    private static readonly DEFAULT_RETRY_CONFIG: IHttpRetryConfig = {
+        enabled: false,
+        maxAttempts: 3,
+        initialDelay: 1000,  // 1 second
+    };
+
     constructor(
         private readonly accessManager: AppAccessorManager,
         private readonly bridges: AppBridges,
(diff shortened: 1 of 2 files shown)


packages/apps-engine/src/server/accessors/Http.ts
```diff
@@ -31,47 +37,76 @@ export class Http implements IHttp {
...
+                return response;
+
+            } catch (error) {
+                lastError = error;
+                if (!retryConfig.enabled || attempt === (retryConfig.maxAttempts || 3) - 1) {
+                    throw error;
+                }
+
+                // Exponential backoff
+                const delay = retryConfig.initialDelay * Math.pow(2, attempt);
```
Gustrb, line 104: I think we should probably be smarter here instead of doing exponential backoff despite the return of the request.
For example, let's say the server returned a 404. In that case, it would be pointless to keep requesting the server for the same resource.

I think we should probably just re-do the calls for service unavailable status code (503) and (maybe) 500, not sure though, otherwise, I think re-doing the call is detrimental, and failure logic should be handled by the apps themselves WDYT?

Gustrb:
Thanks for the contribution!!
This seems like a really interesting approach, I have some concerns about:
- When should we retry, I think we should retry if and only if the response is a 503 or a 500
- Should we (potentially) call the pre http request handlers more than once(?) won't that be problematic?

packages/apps-engine/src/server/accessors/Http.ts
```diff
@@ -31,47 +37,76 @@ export class Http implements IHttp {
...
+        method: RequestMethod,
+        options?: IHttpRequest
+    ): Promise<IHttpResponse> {
+        const retryConfig = {
+            ...Http.DEFAULT_RETRY_CONFIG,
+            ...options?.retry
+        };
+
+        let attempt = 0;
+        let lastError: any;
```
Gustrb, line 51: Can't we type this as Error?

jannatkhandev:
@Gustrb 
Thank you for the valuable feedback! I completely agree with your points.
Yes, my initial intention was actually to be more selective with retries. The retry mechanism should primarily handle transient failures like service unavailability (503) and network errors, rather than retrying on all error types.
Here's what I propose for the retry conditions:

1. Network/connection errors (e.g., timeout, connection refused)
2. HTTP 503 (Service Unavailable) - perfect candidate for retries as it indicates temporary unavailability
3. HTTP 500 (Internal Server Error) - could be included as these might be transient server issues
4. Rate limit responses (HTTP 429) - could also be a candidate with proper delay handling
5. or more 4xx and 5xx responses?

Regarding the pre-request handlers - point to be discussed. While the current implementation maintains consistency by running them on each attempt, we should discuss if this is the desired behavior or if we should preserve the handlers' result from the initial attempt?

packages/apps-engine/src/server/accessors/Http.ts
jannatkhandev: Made changes as requested.

jannatkhandev:
@Gustrb 
I have made changes addressing your review, can you PTAL?
Thanks.

packages/apps-engine/src/server/accessors/Http.ts
```diff
@@ -1,9 +1,22 @@
-import type { IHttp, IHttpExtend, IHttpRequest, IHttpResponse } from '../../definition/accessors';
-import { RequestMethod } from '../../definition/accessors';
+import type { IHttp, IHttpExtend, IHttpRequest, IHttpResponse, IHttpRetryConfig } from '../../definition/accessors';
+import { HttpStatusCode, RequestMethod } from '../../definition/accessors';
 import type { AppBridges } from '../bridges/AppBridges';
 import type { AppAccessorManager } from '../managers/AppAccessorManager';
 
 export class Http implements IHttp {
+    private static readonly DEFAULT_RETRY_CONFIG: IHttpRetryConfig = {
```
Gustrb, line 7: I think instead of having this hardcoded in here, we should have settings so the user can configure that in their workspace, wdyt @d-gubert ??

Gustrb: Or even a retry policy inside the app's settings, telling how should it retry. But I am not sure if it would not be overkill for now though

jannatkhandev:
@Gustrb thanks for your prompt reply.
Requesting you to please add me to the group conversation with @d-gubert @rodrigok sir on our open server? (on open.rocket.chat, my username is jannat.khan)
I have several improvement ideas for the apps engine, I will be happy to create sprints for the same and work on the tasks.

packages/apps-engine/src/server/accessors/Http.ts
jannatkhandev: +1 on the app's settings, like having default retry config for the app.
For the workspace setting as these changes revolve around apps using Http API, I feel it as an overkill.
Happy to make changes as advised :)

Gustrb: I mean, I think having it as a workspace setting would be the easier way off.

Gustrb:
Hey, I don't think we need to create sprints &/or tasks to implement those.
But I'll connect with you on open so we can discuss your ideas wdyt?

jannatkhandev:
@Gustrb Moved retry logic to higher order function as advised, PTAL.

PR #34849, opened 2024-12-30
fix: Inactive Clients Not Refreshing Action Buttons After App Deployment

Gustrb:

Currently, when the app is redeployed, active clients properly fetch and display the updated action buttons (via GET /apps/actionButtons). However, any clients that were inactive during the deployment (e.g., a laptop in sleep mode) do not fetch the updated list upon becoming active again. As a result, these inactive clients continue using outdated actionIds, causing inconsistent and potentially broken interactions.

So now, everytime the client goes offline, we invalidate the cache, since the buttons might have changed in the meantime

## Steps to test or reproduce

Start with the client active and confirm it fetches updated action buttons upon deployment.
Put the client into an inactive state (e.g., close laptop lid or simulate a network disconnection), then redeploy the app.
Reactivate the client and verify that it immediately retrieves the updated action buttons list.
Repeat steps 2-3 for multiple clients to ensure consistent behavior across different endpoints and environments.

Note: You can use `Meteor.disconnect` to test socket reconnection

## Further comments

[CONN-434](https://rocketchat.atlassian.net/browse/CONN-434)

[CONN-434]: https://rocketchat.atlassian.net/browse/CONN-434?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/meteor/client/hooks/useAppActionButtons.ts
```diff
@@ -36,6 +36,26 @@ export const useAppActionButtons = <TContext extends `${UIActionButtonContext}`>
...
 	);
 
+	useEffect(() => {
+		if (!uid) {
+			return;
+		}
+
+		// Setup Tracker to listen to Meteor's status changes
+		const statusTracker = Tracker.autorun(() => {
+			const isOffline = !Meteor.status().connected;
```
Gustrb, line 46: Hey @tassoevan we were talking about trying to have some kind of less-indirect way of telling when the socket disconnected, I found that Meteor.status is what we used in similar situations so I decided to try to implement using it here, but if you thing this approach is not good I guess we can try to find something different

Gustrb:
@Harmeet221 gave us her blessing on this one.

apps/meteor/client/hooks/useAppActionButtons.ts
```diff
@@ -42,6 +43,24 @@ export const useAppActionButtons = <TContext extends `${UIActionButtonContext}`>
 		[],
 	);
 
+	useEffect(() => {
```
gabriellsh, line 46: You can use `useStatusConnection` hook for this. It's pretty much the same thing

gabriellsh (changes requested):

apps/meteor/client/hooks/useAppActionButtons.ts
```diff
@@ -42,6 +41,13 @@ export const useAppActionButtons = <TContext extends `${UIActionButtonContext}`>
 		[],
 	);
 
+	const { status } = useConnectionStatus();
```
gabriellsh, line 44: Please use `status` in the query key and disable the query if `status !== 'connected;` as we discussed in dm

gabriellsh (changes requested):

tassoevan (approved):

PR #34852, opened 2024-12-30
fix: Scoped roles are assignable through admin users page

gabriellsh:

Scoped roles only work when assigned in the context of a room, so there's no reason to show them in the admin page.

Additionally, if the user has scoped roles assigned, they'll be hidden in the select, which in turn will remove them the next time the user's roles are edited.

To reinforce, this SHOULD not impact any existing features or "unintended use", since scoped roles are ignored when checking global permissions.

### DIFF
--- a/apps/meteor/client/views/admin/users/AdminUserFormWithData.tsx
+++ b/apps/meteor/client/views/admin/users/AdminUserFormWithData.tsx
@@ -5,7 +5,9 @@ import type { ReactElement } from 'react';
 import { useTranslation } from 'react-i18next';
 
 import AdminUserForm from './AdminUserForm';
+import { isTruthy } from '../../../../lib/isTruthy';
 import { FormSkeleton } from '../../../components/Skeleton';
+import type { UserInfoQueryData } from '../../../hooks/useUserInfoQuery';
 import { useUserInfoQuery } from '../../../hooks/useUserInfoQuery';
 
 type AdminUserFormWithDataProps = {
@@ -16,9 +18,19 @@ type AdminUserFormWithDataProps = {
 	roleError: unknown;
 };
 
+const filterUnavailableRoles = (availableRoles: IRole[] | undefined) => (data: UserInfoQueryData) => {
+	if (!availableRoles) {
+		return data;
+	}
+	return {
+		...data,
+		user: { ...data.user, roles: data.user.roles.filter((role: string) => isTruthy(availableRoles.find(({ _id }) => _id === role))) },
+	};
+};
+
 const AdminUserFormWithData = ({ uid, onReload, context, roleData, roleError }: AdminUserFormWithDataProps): ReactElement => {
 	const { t } = useTranslation();
-	const { data, isLoading, isError, refetch } = useUserInfoQuery({ userId: uid });
+	const { data, isLoading, isError, refetch } = useUserInfoQuery({ userId: uid }, { select: filterUnavailableRoles(roleData?.roles) });
 
 	const handleReload = useEffectEvent(() => {
 		onReload();
--- a/apps/meteor/client/hooks/useUserInfoQuery.ts
+++ b/apps/meteor/client/hooks/useUserInfoQuery.ts
@@ -1,13 +1,15 @@
 import type { UsersInfoParamsGet } from '@rocket.chat/rest-typings';
+import type { EndpointFunction } from '@rocket.chat/ui-contexts';
 import { useEndpoint } from '@rocket.chat/ui-contexts';
 import { useQuery } from '@tanstack/react-query';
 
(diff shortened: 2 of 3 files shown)


MarcosSpessatto (changes requested):
Does it make sense to add a test to ensure this?

apps/meteor/client/views/admin/users/AdminUserFormWithData.tsx
```diff
@@ -16,9 +18,19 @@ type AdminUserFormWithDataProps = {
 	roleError: unknown;
 };
 
+const filterUnavailableRoles = (availableRoles: IRole[] | undefined) => (data: UserInfoQueryData) => {
+	if (!availableRoles) {
+		return data;
+	}
+	return {
+		...data,
+		user: { ...data.user, roles: data.user.roles.filter((role: string) => isTruthy(availableRoles.find(({ _id }) => _id === role))) },
```
aleksandernsilva, line 27: Considering that `availableRoles` is an array of objects, couldn't we use `availableRoles.some(...)` instead of `isTruthy`?

aleksandernsilva: Also, I don't know if this is a concern, but if `availableRoles` array is big enough I'd even consider creating a `Set` before. E.g `const available = new Set(availableRoles.map(({ _id }) => _id)` then using `available.has(role)`.

.changeset/fast-bananas-smell.md
```diff
@@ -0,0 +1,5 @@
+---
+"@rocket.chat/meteor": patch
+---
+
+removes scoped roles (moderator, owner, leader) from list when editing an user through the admin page
```
aleksandernsilva, line 5:
```suggestion
Removes scoped roles (moderator, owner, leader) from list when editing an user through the admin page
```

aleksandernsilva (changes requested):

apps/meteor/client/views/admin/users/AdminUserFormWithData.tsx
```diff
@@ -16,9 +18,19 @@ type AdminUserFormWithDataProps = {
 	roleError: unknown;
 };
 
+const filterUnavailableRoles = (availableRoles: IRole[] | undefined) => (data: UserInfoQueryData) => {
+	if (!availableRoles) {
```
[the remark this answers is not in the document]
gabriellsh: I thought about this, and I think the only thing we could do is throw an error...

The reason I'm doing this filter is because if the user has been assigned a scoped role before, it would show in the select as an empty item with an `X` to remove it, which stops working and you can't remove this "ghost role".

So what I had to do is, filter available roles to remove Subscription scope, and then remove the scoped roles from the user in order to hide this artifact.

In few words, there should ALWAYS be available roles here, if there's not we should be treating this issue in another place, which I don't think would fit the scope of this PR. I just checked if "availableRoles" exist in order to please typescript

apps/meteor/client/views/admin/users/AdminUserFormWithData.tsx
gabriellsh: Yep that does make sense. I don't think there would be a difference between find and some performance wise, but it's way easier to read, thanks for catching this one.

I'm not sure it's necessary to make a set. I don't think roles ever exceed 50, and I think up to 100 this wouldn't be an issue. I'll do some manual perf testing just to make sure and come back to you, but I think that won't be necessary

gabriellsh:
> Does it make sense to add a test to ensure this?

@MarcosSpessatto  I didn't do any tests yet because I'm waiting on product to green light this change, since this is a suggestion of my own. I think it'll go through but I decided to no spend effort in tests until I was sure

apps/meteor/client/views/admin/users/AdminUserFormWithData.tsx
tassoevan: @aleksandernsilva too many data transformations between array and `Set` to make it worthy, I think.

pierre-lehnen-rc (commented):
> To reinforce, this SHOULD not impact any existing features or "unintended use", since scoped roles are ignored when checking global permissions.

This premise mentioned on the PR description is not true - at least not for the backend side. If a scoped role is assigned to an user globally, its permissions **will** be granted globally.

It makes sense to block certain roles such as `owner` from being assigned globally, as they are specifically designed to give permissions to a room's owner. At the same time there are other scoped roles such as `moderator` that could work both ways, allowing an user to moderate a specific room and other users to moderate an entire workspace.

gabriellsh:
As @pierre-lehnen-rc commented, scoped roles assigned globaly have different behaviors between FE & BE. We should first normalize this and make them work similarly before we can decide if we should move on with this.

Flagged this as a breaking change for now.

PR #34853, opened 2024-12-31
fix: Modal ocasionally not focusing properly making input unusable

gabriellsh:

Use `inert` prop instead of `aria-hidden` for disabling focus on document when a modal is open.

I was only able to notice this issue when trying to save an edited user in the admin page with TOTP enabled. The TOTP modal opens but the input is unaccessible with mouse clicks nor with keyboard navigation
## Issue(s)

[CORE-908](https://rocketchat.atlassian.net/browse/CORE-908)
## Steps to test or reproduce

*Make sure TOTP is enable and `Enforce password fallback` is enabled aswell* (usually both come enabled by default in new workspaces)
- Go to `Admin -> Users`
- Refresh the page (*1)
- Select an user
- Click edit and change any property
- Click save
- TOTP modal will open but unusable (only cancel or close buttons are clickable, but not accessible through keyboard)

Observations:
*1: By clicking "cancel" and trying again the issue goes away, so refreshing the page is necessary to reproduce the issue multiple times.
You can set `Remember Two Factor for (seconds)` to `0` in `Admin -> Accounts -> Two Factor Authentication` to ensure the TOTP modal appears every time, even if you successfully inputed the code/password.

## Further comments

I do not know how this change affects screen readers, so some support here would be nice.

Edit: apparently `inert` should have the same effect as `aria-hidden` , and is supported in every major browser *BESIDES IE11*. See: https://caniuse.com/?search=inert

[CORE-908]: https://rocketchat.atlassian.net/browse/CORE-908?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.changeset/moody-foxes-sleep.md
+++ b/.changeset/moody-foxes-sleep.md
@@ -1,0 +1,5 @@
+---
+"@rocket.chat/meteor": patch
+---
+
+fixes an issue where opening a modal for the first times renders it's inputs unusable
--- a/apps/meteor/client/views/root/MainLayout/LayoutWithSidebar.tsx
+++ b/apps/meteor/client/views/root/MainLayout/LayoutWithSidebar.tsx
@@ -1,11 +1,19 @@
 import { Box } from '@rocket.chat/fuselage';
 import { useLayout, useSetting, useCurrentModal, useRoute, useCurrentRoutePath } from '@rocket.chat/ui-contexts';
 import type { ReactElement, ReactNode } from 'react';
-import { useEffect, useRef } from 'react';
+import { useEffect, useRef, version } from 'react';
 
 import AccessibilityShortcut from './AccessibilityShortcut';
 import { MainLayoutStyleTags } from './MainLayoutStyleTags';
 import Sidebar from '../../../sidebar';
+
+const inertBooleanSupported = Number(version.split('.')[0]) >= 19;
+
+/**
+ * Before version 19, react JSX treats empty string "" as truthy for inert prop.
+ * @see {@link https://stackoverflow.com/questions/72720469}
+ * */
+const isInert = inertBooleanSupported ? (x: boolean) => x : (x: boolean) => (x ? '' : undefined);
 
 const LayoutWithSidebar = ({ children }: { children: ReactNode }): ReactElement => {
 	const { isEmbedded: embeddedLayout } = useLayout();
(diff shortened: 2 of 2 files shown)


MarcosSpessatto (approved):

apps/meteor/client/views/root/MainLayout/LayoutWithSidebar.tsx
```diff
@@ -1,12 +1,20 @@
...
 import Sidebar from '../../../sidebar';
 
+const inertBooleanSupported = Number(version.split('.')[0]) >= 19;
+
+/**
+ * Before version 19, react JSX treats empty string "" as truthy for inert prop.
+ * @see {@link https://stackoverflow.com/questions/72720469}
+ * */
+const isInert = inertBooleanSupported ? (x: boolean) => x : (x: boolean) => (x ? '' : undefined);
+
```
aleksandernsilva, line 17: I believe this will need to be replicated to [LayoutSidebarV2](https://github.com/RocketChat/Rocket.Chat/blob/9af2c11d64b85abd74989777eee93f48c0684ec0/apps/meteor/client/views/root/MainLayout/LayoutWithSidebarV2.tsx). It might be a good idea to have this as an util and use it in both files.

aleksandernsilva (changes requested):
Tested, and the fix is working. Adding `inert` seems to help `FocusScope` sort itself out. Now, even though `inert` covers the `aria-hidden` functionality, Playwright seems to have some problems ignoring `inert` elements in the same way it does `aria-hidden` 🤔 .

.changeset/moody-foxes-sleep.md
```diff
@@ -0,0 +1,5 @@
+---
+"@rocket.chat/meteor": patch
+---
+
+fixes an issue where opening a modal for the first times renders it's inputs unusable
```
aleksandernsilva, line 5:
```suggestion
Fixes an issue where opening a modal for the first times renders it's inputs unusable
```

apps/meteor/client/views/root/MainLayout/LayoutWithSidebar.tsx
gabriellsh: I forgot that guy :grimacing:

gabriellsh:
> Tested, and the fix is working. Adding `inert` seems to help `FocusScope` sort itself out. Now, even though `inert` covers the `aria-hidden` functionality, Playwright seems to have some problems ignoring `inert` elements in the same way it does `aria-hidden` 🤔 .

That's a good catch btw. I was wondering which problems we might face with this. I removed "aria-hidden" because aria was throwing errors and removing the prop in some cases. Maybe with both it should be fine...

Maybe there's another fix also since this didn't happen in every screen. I'll take a look

dougfabris (commented):
I don't think ignoring the main layout elements is the proper fix, neither a good aproach. IMO we should investigate deeper modal focus scope

PR #34858, opened 2024-12-31
fix: decoder error preventing succesfull app subprocess restart

d-gubert:

If msgpack's `decodeStream` throws an error, the internal state of the decoder becomes "invalid" (the encoded malformed message stays cached) and the engine was not capable of reestabilishing communications with any other subprocesses as the decoder instance never disposes of the malformed data. This essentially rendered the whole apps-engine incapable of communication, as the decoder instance was shared among apps.

To solve this, now the engine will instantiate a new decoder every time it spawns a subprocess. This guarantees that the internal state of the decoder for a subprocess will never impact the communication of other subprocesses, and also allows the engine to reestablish communication with a new subprocess of an app after a `Decode error`.

Another side effect of this problem could be observed on the subprocess side. Since the engine stopped reading data from the subprocess' `stdout` pipe, the operation system was blocking the `write` syscall that the subprocess made and essentially made it stuck with an ever growing queue of ougoing messages.

Note: I've also tried reestablishing the communication over the same pipe that we had with the subprocess - since the state of the subprocess wasn't invalid, it made sense to try that. However, the msgpack decoder left the `stdout` stream of the ChildProcess in an unrecoverable error state, so I was left with restarting the whole subprocess.
## Issue(s)

[CONN-448]

## Further comments

[CONN-448]: https://rocketchat.atlassian.net/browse/CONN-448?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/apps-engine/src/server/runtime/deno/codec.ts
```diff
@@ -21,9 +22,24 @@ extensionCodec.register({
...
+ * stream, store the data locally and decode it from its buffer.
+ *
+ * In practice, this affects the decoder when there is decode error. After an error, the decoder
+ * keeps the malformed data in its buffer, and even if we try to decode from another source (e.g. different stream)
+ * it will fail again as there's still data in the buffer.
+ *
+ * For that reason, we can't have a singleton instance of Encoder and Decoder, but rather one
+ * instance for each time we create a new subprocess
+ */
+export const newEncoder = () => new _Encoder({ extensionCodec });
```
KevLehman, line 41: Style one: wouldn't be nicer a `getEncoder` instead of `newEncoder`?

packages/apps-engine/src/server/runtime/deno/AppsEngineDenoRuntime.ts
```diff
@@ -610,51 +611,58 @@ export class DenoRuntimeSubprocessController extends EventEmitter {
...
+                    }
+
+                    console.error(`[${this.getAppId()}] Error executing handler`, e, message);
                 }
-
-                console.error(`[${this.getAppId()}] Error executing handler`, e, message);
             }
+        } catch (e) {
+            console.error(`[${this.getAppId()}]`, e);
+            this.emit('error', new Error('DECODE_ERROR'));
```
KevLehman, line 657: maybe would be better to create a custom error? so you can `new DecoderError` and then on the listener, instead of checking for the message content, you can check if the error is instance of DecoderError. wdyt?

ggazzo (approved):

packages/apps-engine/src/server/runtime/deno/AppsEngineDenoRuntime.ts
d-gubert: I think if we were throwing the Error it would be more beneficial to follow that approach, as we would be trying to inform an unknown caller of an error that happened. Here, since we're in a more controlled scope and a more specific target, I'd say there isn't much difference.

Tbh I've even considered a simple string like "E_DECODE_ERROR" :stuck_out_tongue:

KevLehman (approved):

PR #34864, opened 2025-01-02
fix: Allow users to fetch the `packageValue` of settings on `/settings` endpoint

KevLehman:

Since `fields` param is deprecated (and it's disabled by default, behind an env var), there was no way to obtain the default value of the settings when using the `settings` endpoint.

The proposed solution is to use a new boolean flag that allows the requester to fetch the `packageValue` along with the current value when using the aforementioned endpoint.

Flag is `includeDefaults`
## Issue(s)

https://rocketchat.atlassian.net/browse/CORE-911


MarcosSpessatto (dismissed):

debdutdeb:
Shouldn't this be a `feat` instead? I understand why it could be considered a fix, however it's a new thing that's being added?

KevLehman:
Well, it's a funny story: this was working, then it stopped working, then this will make it work again. So in theory is a fix... but could be a feat as well (as a new thing is added).

But i'm inclined more to "fix" cause it was working in a way and it stopped working

packages/rest-typings/src/v1/settings.ts
```diff
@@ -57,6 +57,43 @@ const SettingsPublicWithPaginationSchema = {
...
 export const isSettingsPublicWithPaginationProps = ajv.compile<SettingsPublicWithPaginationProps>(SettingsPublicWithPaginationSchema);
 
+type SettingsGetParams = {
+	includeDefaults?: boolean;
+	count?: number;
+	offset?: number;
+	sort?: string;
+	fields?: string;
+	query?: string;
+};
```
matheusbsilva137, line 67: Why not use our `PaginatedRequest` helper?

apps/meteor/tests/end-to-end/api/settings.ts
```diff
@@ -100,6 +100,20 @@ describe('[Settings]', () => {
 				})
 				.end(done);
 		});
+		it('should return the default values of the settings when includeDefaults is true', async () => {
```
matheusbsilva137, line 103: I'm wondering if it'd be good to add a test to cover the case where this param is not provided (or adapt the old ones) so that we can assure `packageValue` is not being returned by default

matheusbsilva137 (changes requested):

apps/meteor/tests/end-to-end/api/settings.ts
KevLehman: Actually, was thinking on that, but at the end didn't feel any need for it. There was no way for that to be possible, so i ended up with just the positive test.

packages/rest-typings/src/v1/settings.ts
KevLehman, line 67:
```suggestion
type SettingsGetParams = PaginatedRequest<{ includeDefaults?: boolean; query?: string }>;
```

matheusbsilva137 (approved):

PR #34868, opened 2025-01-03
Cache optimization

user312222:
feat: Add userCache.ts for cache optimization

 This Pull Request adds a new userCache.ts file to make the application faster and more efficient. It uses a simple Map to store user data in memory, so the database is not called every time. This reduces the load on the database and makes the application respond quicker. The cache is also updated when needed to keep the data correct and up-to-date. This change helps the application handle more users without slowing down.

Changes:
- Added a new `userCache.ts` file to implement a Map-based caching mechanism.
- Integrated the caching logic into user-related service functions.
- Updated related test cases to include cache validation.

Tests:
- All existing tests pass successfully.
- Added new tests for the `userCache.ts` functionality.

Notes:
- Please review the integration with `UsersRaw` for potential optimization.
- Suggestions for further improvements are welcome.

### DIFF
--- a/apps/meteor/server/services/userCache.ts
+++ b/apps/meteor/server/services/userCache.ts
@@ -1,0 +1,30 @@
+import { UsersRaw } from '../../../../packages/models/src/models/Users.js';
+import { db } from '../../server/database/utils';
+import { trashCollection } from '../../server/database/trash'; 
+
+const userCache = new Map<string, any>();
+
+const user = new UsersRaw(db, trashCollection);
+
+export async function findUserById(userId: string) {
+
+  if (userCache.has(userId)) {
+    return userCache.get(userId);
+  }
+
+
+  const user1 = await user.findOne({ _id: userId });
+
+  if (user) {
+    userCache.set(userId, user);
+  }
+  return user;
+}
+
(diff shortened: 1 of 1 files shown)


apps/meteor/server/services/userCache.ts
```diff
@@ -0,0 +1,30 @@
+import { UsersRaw } from '../../../../packages/models/src/models/Users.js';
```
Gustrb, line 1: what happens if one instance deletes an user?

user312222: If a user deletes their account but the cache is not cleared, the system might still use old cache data, leading to "user not found" errors when a new account is created. To solve this, it's good to have a command to clear the cache when this error occurs, but an automated system to refresh the cache during account deletion or after some time would be more efficient and user-friendly.

Gustrb: but what if one instance reads the cache while the 'clear command' is being ran?
NOTE: clearing an object in JS is not an atomic operation

Gustrb: I don't think we should implement an ACID compliant cache just for this, lets defer to the database to ensure correctness.

PR #34873, opened 2025-01-03
fix: SAML login redirects to incorrect room when using an invite

matheusbsilva137:

 - Remove `id` and `redirectUrl` from SAML's global `ServiceProvideOptions` (since they're actually meant to relate to each client/request);
 - Handle invite token redirect on the client side when using SAML (with localStorage).

## Further comments

[SUP-719]

[SUP-719]: https://rocketchat.atlassian.net/browse/SUP-719?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


MarcosSpessatto (dismissed):

apps/meteor/app/meteor-accounts-saml/server/lib/generators/AuthorizeRequest.ts
```diff
@@ -53,13 +53,16 @@ export class AuthorizeRequest {
...
+		serviceProviderOptions: IServiceProviderOptions,
+		credentialToken?: string,
+	): IAuthorizeRequestVariables {
 		let id = `_${SAMLUtils.generateUniqueID()}`;
 		const instant = SAMLUtils.generateInstant();
 
 		// Post-auth destination
-		if (serviceProviderOptions.id) {
-			id = serviceProviderOptions.id;
+		if (credentialToken) {
```
pierre-lehnen-rc, line 64:
```ts
		const id = credentialToken || `_${SAMLUtils.generateUniqueID()}`;
```

apps/meteor/client/views/invite/InvitePage.tsx
```diff
@@ -28,7 +30,8 @@ const InvitePage = (): ReactElement => {
 		return <PageLoading />;
 	}
 
-	if (isValidInvite) {
+	if (isValidInvite && token) {
+		localStorage.setItem(KEY, token);
```
pierre-lehnen-rc, line 34: invite token should be removed on successful non-SAML login as well.

Possibly needs to also be removed on failed login as well?

pierre-lehnen-rc (changes requested):

apps/meteor/client/views/invite/InvitePage.tsx
matheusbsilva137: Hmmm am I not covering both cases [here](https://github.com/RocketChat/Rocket.Chat/pull/34873/files#diff-450ab68d9ab08bab95e5c2a54d94e7e30bd6be6d1939dde254842be4eca8605d) already? Real question bc I thought so, but I might me missing somethind
That mutation is always called since this is responsible for redirecting the user to the invite link

pierre-lehnen-rc (dismissed):

KevLehman (dismissed):

apps/meteor/client/views/invite/hooks/useSamlInviteToken.ts
```diff
@@ -0,0 +1,9 @@
+import { useSessionStorage } from '@rocket.chat/fuselage-hooks';
+import { useRouteParameter } from '@rocket.chat/ui-contexts';
+
+const KEY = 'saml_invite_token';
+
+export const useSamlInviteToken = () => {
+	const token = useRouteParameter('hash');
```
gabriellsh, line 7: I see you're setting the token here and inside InvitePage. I think you can choose one or the other, probably just here would be enough.

apps/meteor/client/views/home/HomePage.tsx
```diff
@@ -6,8 +6,11 @@ import CustomHomePage from './CustomHomePage';
 import DefaultHomePage from './DefaultHomePage';
 import { KonchatNotification } from '../../../app/ui/client/lib/KonchatNotification';
 
+const KEY = 'invite_token';
+
 const HomePage = (): ReactElement => {
 	useEffect(() => {
+		sessionStorage.removeItem(KEY);
```
gabriellsh, line 13: I think this has to be removed aswell, right?

gabriellsh (changes requested):

apps/meteor/client/views/invite/hooks/useSamlInviteToken.ts
matheusbsilva137: Hmmm I'm not sure. As far as I checked in our code for `useSessionStorage`, the `initialValue` param (the second one, which I'm using in this file) is only assigned to the key if it hasn't been created yet. So in case the key already exists, I need to set it in the `InvitePage` again (e.g. logging out and trying to login again with an invite in the same tab would not work)

gabriellsh: Oh okay then. Let's leave it like this

gabriellsh (approved):

PR #34890, opened 2025-01-06
refactor: remove personal access token method calls (server)

MarcosSpessatto:
https://rocketchat.atlassian.net/browse/CORE-914


apps/meteor/imports/personal-access-tokens/server/api/methods/generateToken.ts
```diff
@@ -14,6 +14,38 @@ declare module '@rocket.chat/ddp-client' {
 	}
 }
 
+export const generatePersonalAccessTokenOfUser = async ({ bypassTwoFactor, tokenName, userId }: {tokenName: string, userId: string, bypassTwoFactor: boolean}): Promise<string> => {
+	if (!(await hasPermissionAsync(userId, 'create-personal-access-tokens'))) {
```
KevLehman, line 18: The only thing i dont like is that the permission is handled here, instead of at the endpoint/method.

But its something that maybe we can change later.

KevLehman (approved):

PR #34897, opened 2025-01-06
chore: Disallow updating closing properties of livechat rooms via Apps Engine

KevLehman:

This change removes the capability of apps of changing some omnichannel props by directly modifying the Updater for the room. Updates were done in an unsafe way, that allowed apps to leave rooms in a bad state.

### DIFF
--- a/apps/meteor/app/apps/server/bridges/rooms.ts
+++ b/apps/meteor/app/apps/server/bridges/rooms.ts
@@ -5,14 +5,16 @@ import { RoomType } from '@rocket.chat/apps-engine/definition/rooms';
 import type { IUser } from '@rocket.chat/apps-engine/definition/users';
 import type { GetMessagesOptions } from '@rocket.chat/apps-engine/server/bridges/RoomBridge';
 import { RoomBridge } from '@rocket.chat/apps-engine/server/bridges/RoomBridge';
+import { isOmnichannelRoom } from '@rocket.chat/core-typings';
 import {
+	type IOmnichannelRoom,
 	type ISubscription,
 	type IUser as ICoreUser,
 	type IRoom as ICoreRoom,
 	type IMessage as ICoreMessage,
-	isOmnichannelRoom,
 } from '@rocket.chat/core-typings';
 import { Subscriptions, Users, Rooms, Messages } from '@rocket.chat/models';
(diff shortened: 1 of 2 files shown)


yash-rajpal (dismissed):
LGTM, even though I'm not a backend guy 😬

apps/meteor/app/apps/server/bridges/rooms.ts
```diff
@@ -160,11 +166,24 @@ export class AppRoomBridge extends RoomBridge {
...
+		const originalRoom = await Rooms.findOneById(room.id);
+		if (!room.id || !originalRoom) {
 			throw new Error('A room must exist to update.');
 		}
 
-		const rm = await this.orch.getConverters()?.get('rooms').convertAppRoom(room);
+		let rm = await this.orch.getConverters()?.get('rooms').convertAppRoom(room);
+
+		if (isOmnichannelRoom(originalRoom)) {
+			// We shouldn't allow to mess up with a room's closed state via apps. Only via livechat bridge
```
Gustrb, line 177: shouldn't we outright throw an error if one try to do this? silent failure can be *really* annoying

KevLehman: Thought so, but felt throwing a _new_ error on update could be a breaking change. Silent here since people shouldn't be touching any of the props listed, so we'll assume they did by mistake and correct the room state.

Changing closing status, type of livechat rooms & open/close flag is not supported (and never was)

Gustrb: Yea since we never aimed to support it I thought it would not be that big of a deal adding it (even though technically it is a breaking change). But if you think it is not that big of a deal I am fine with it.
(maybe some logging would be useful so we can have visibility, when a customer is trying
 to close a room through update and it not working)

yash-rajpal: I agree, we can also add in app's logs, that the app tried to update one of the **restricted fields** and bla bla

KevLehman: Right, good call

KevLehman:
Not needed for now, we're gonna look for a 2nd alternative

PR #34901, opened 2025-01-07
refactor: Invalid values as React nodes

tassoevan:

It fixes types for values that ended up in JSX as React nodes.

## Further comments

The types from `@types/react@17.*` are too lenient, but the next majors are tighter. Therefore this is preparation for  upgrade.


dougfabris (approved):
God bless u

ggazzo (approved):

PR #34920, opened 2025-01-09
chore: Improve Transcript service call chain

KevLehman:

Basically, previous call was `meteor` => `transcript` => `queue` => `transcript`. The first call from meteor to the transcript service could be omitted if we do the logic on meteor.

With the new changes, meteor is the one that decides if we call transcript directly (on sync calls, for tests) or if we queue the item (for regular work).

Changes should work and improve a bit whether you use microservices or not.
## Issue(s)

https://rocketchat.atlassian.net/browse/CORE-920


apps/meteor/ee/app/livechat-enterprise/server/lib/requestPdfTranscript.ts
```diff
@@ -0,0 +1,44 @@
...
+import { LivechatRooms } from '@rocket.chat/models';
+
+import { logger } from './logger';
+
+const serviceName = 'omnichannel-transcript' as const;
+export const requestPdfTranscript = async (
+	room: AtLeast<IOmnichannelRoom, '_id' | 'open' | 'v' | 'pdfTranscriptRequested'> | null,
+	requestedBy: string,
+): Promise<void> => {
+	if (!room) {
```
MarcosSpessatto, line 12: Should we really check this? 🤔

MarcosSpessatto (changes requested):

apps/meteor/ee/app/livechat-enterprise/server/lib/requestPdfTranscript.ts
KevLehman: It's to remove the `null` type from the room object. I copied it over from the service without much looking, maybe we don't need it anymore

MartinSchoeler (approved):
Symbolic OC review :)

PR #34926, opened 2025-01-09
feat: Integrate call ringer volume with video conf ringing/dialing

rique223:

### DIFF
--- a/apps/meteor/client/hooks/useUserSoundPreferences.ts
+++ b/apps/meteor/client/hooks/useUserSoundPreferences.ts
@@ -1,13 +1,11 @@
 import { useUserPreference } from '@rocket.chat/ui-contexts';
 
-const relativeVolume = (volume: number, masterVolume: number) => {
-	return (volume * masterVolume) / 100;
-};
+const relativeVolume = (volume: number, masterVolume: number) => (volume * masterVolume) / 100;
 
 export const useUserSoundPreferences = () => {
-	const masterVolume = useUserPreference<number>('masterVolume', 100) || 100;
-	const notificationsSoundVolume = useUserPreference<number>('notificationsSoundVolume', 100) || 100;
-	const callRingerVolume = useUserPreference<number>('callRingerVolume', 100) || 100;
+	const masterVolume = useUserPreference<number>('masterVolume', 100) ?? 100;
+	const notificationsSoundVolume = useUserPreference<number>('notificationsSoundVolume', 100) ?? 100;
+	const callRingerVolume = useUserPreference<number>('callRingerVolume', 100) ?? 100;
 
 	return {
 		masterVolume,
--- a/apps/meteor/client/hooks/useUserSoundPreferences.ts
+++ b/apps/meteor/client/hooks/useUserSoundPreferences.ts
@@ -7,11 +7,11 @@ const relativeVolume = (volume: number, masterVolume: number) => {
 export const useUserSoundPreferences = () => {
 	const masterVolume = useUserPreference<number>('masterVolume', 100) || 100;
 	const notificationsSoundVolume = useUserPreference<number>('notificationsSoundVolume', 100) || 100;
-	const voipRingerVolume = useUserPreference<number>('voipRingerVolume', 100) || 100;
+	const callRingerVolume = useUserPreference<number>('callRingerVolume', 100) || 100;
 
 	return {
 		masterVolume,
 		notificationsSoundVolume: relativeVolume(notificationsSoundVolume, masterVolume),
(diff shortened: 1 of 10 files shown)


apps/meteor/client/views/account/preferences/AccountPreferencesPage.tsx
```diff
@@ -94,7 +94,7 @@ const AccountPreferencesPage = (): ReactElement => {
 			<PageFooter isDirty={isDirty}>
 				<ButtonGroup>
 					<Button onClick={() => reset(preferencesValues)}>{t('Cancel')}</Button>
-					<Button form={preferencesFormId} primary type='submit'>
+					<Button form={preferencesFormId} data-qa='AccountPreferencesPageSaveButton' primary type='submit'>
```
dougfabris, line 97: Please do not use `data-qa` we've already talked about it internally a lot of times

apps/meteor/tests/e2e/page-objects/account-profile.ts
```diff
@@ -44,6 +44,18 @@ export class AccountProfile {
 		return this.page.locator('//label[contains(text(), "Email")]/..//input');
 	}
 
+	get preferencesSoundAccordionOption(): Locator {
+		return this.page.locator('h2:has-text("Sound")');
+	}
+
+	get preferencesCallRingerVolumeSlider(): Locator {
```
dougfabris, line 51: we really need the (..) locator?

dougfabris (approved):
LGTM!

apps/meteor/client/views/room/contextualBar/VideoConference/VideoConfPopups/VideoConfPopup/IncomingPopup.tsx
```diff
@@ -35,19 +36,23 @@ const IncomingPopup = ({ id, room, position, onClose, onMute, onConfirm }: Incom
...
 	const setPreferences = useVideoConfSetPreferences();
+	const roomName = useVideoConfRoomName(room);
 
 	const params = useMemo(() => ({ callId: id }), [id]);
 	const { phase, value } = useEndpointData('/v1/video-conference.info', { params });
 	const showMic = Boolean(value?.capabilities?.mic);
 	const showCam = Boolean(value?.capabilities?.cam);
 
+	const dialogLabel =
+		room.t === 'd' ? `${t('Incoming_call_from__roomName__', { roomName })}` : `${t('Incoming_call_from__roomName__', { roomName })}`;
```
aleksandernsilva, line 47: There seems to be something wrong here. The same string is being returned for both conditions.

aleksandernsilva (changes requested):

apps/meteor/client/views/room/contextualBar/VideoConference/VideoConfPopups/VideoConfPopup/IncomingPopup.tsx
dougfabris: Indeed, thanks. Fixed now

dougfabris (dismissed):
LGTM!

aleksandernsilva (dismissed):

apps/meteor/server/settings/accounts.ts
```diff
@@ -711,7 +711,7 @@ export const createAccountSettings = () =>
 				i18nLabel: 'Notification_volume',
 			});
 
-			await this.add('Accounts_Default_User_Preferences_voipRingerVolume', 100, {
+			await this.add('Accounts_Default_User_Preferences_callRingerVolume', 100, {
```
sampaiodiego, line 714: I'd ask to not change the setting `_id`, this usually has not the effect you'd expect, this doesn't actually _rename_ the setting, it actually creates a new one.. so workspaces which already have the old one will actually have two..

I understand the `_id` meaning may look wrong, but we should deal with it.. the correct way to "rename" a setting is via migration, which we cannot do during regular release cycles..

sampaiodiego (changes requested):

apps/meteor/server/settings/accounts.ts
dougfabris: can't we create this new one for now and remove the old setting in the next major?

dougfabris: in order to do that we had to create a migration to update the new value with previous one, right?

dougfabris:
@rique223 there are some places where you need to avoid using the new name you just rolled back

sampaiodiego (approved):

aleksandernsilva (approved):

PR #34934, opened 2025-01-10
fix: `chat.update` allows to set message as empty string

KevLehman:

`chat.update` allowed to pass an empty `text` field. This caused the UI to render an empty message.

Now, this is removed and API will throw an error. On UI, we don't allow this and we ask the user if the message should be deleted instead.
## Issue(s)

https://rocketchat.atlassian.net/browse/CORE-603


Aaradhya-07:
I would like to work on this issue

MarcosSpessatto (changes requested):
Should we add this validation to the logic layer as well (`executeUpdateMessage`)?

PR #34940, opened 2025-01-13
feat: Allow users to force close Omnichannel rooms

KevLehman:

The new param passed to `rooms.closeByUser` endpoint allows users to forcefully close an omnichannel conversation.
This may come handy when a room, for X or Y reason ends up in a bad state, given an error during the closing procedure, or an update after the room has been closed. Currently, calling the API again will throw an error, as the room state is not valid and as such we avoid updating it.
This new setting will allow admins to allow the `forceClose` parameter, which will skip most of the validations and run the closing process again.
## Issue(s)

https://rocketchat.atlassian.net/browse/CONN-475


.changeset/eight-humans-sip.md
```diff
@@ -0,0 +1,7 @@
+---
+"@rocket.chat/meteor": patch
+"@rocket.chat/i18n": patch
+"@rocket.chat/rest-typings": patch
```
matheusbsilva137, line 4:
```suggestion
"@rocket.chat/meteor": minor
"@rocket.chat/i18n": minor
"@rocket.chat/rest-typings": minor
```

Are these really patches? Wondering because the PR is tagged as `feat`

apps/meteor/app/livechat/server/lib/LivechatTyped.ts
```diff
@@ -240,10 +240,10 @@ class LivechatClass {
 		session: ClientSession,
 	): Promise<{ room: IOmnichannelRoom; closedBy: ChatCloser; removedInquiry: ILivechatInquiryRecord | null }> {
 		const { comment } = params;
-		const { room } = params;
+		const { room, forceClose } = params;
```
matheusbsilva137, line 243: I wonder if we should re-check the setting here :thinking: 
I know you're already checking in the endpoint level, but maybe we should have something at a lower level to prevent force closing if the setting is disabled. This is more of an opinion, so I'd like to hear what you think or if you thought about this already

apps/meteor/app/livechat/server/api/v1/room.ts
```diff
@@ -195,9 +196,22 @@ API.v1.addRoute(
 	},
 	{
 		async post() {
-			const { rid, comment, tags, generateTranscriptPdf, transcriptEmail } = this.bodyParams;
+			const { rid, comment, tags, generateTranscriptPdf, transcriptEmail, forceClose } = this.bodyParams;
 
-			await closeLivechatRoom(this.user, rid, { comment, tags, generateTranscriptPdf, transcriptEmail });
+			const allowForceClose = rcSettings.get<boolean>('Omnichannel_allow_force_close_conversations');
```
matheusbsilva137, line 201: What if someone tries to force close a room in a workspace where this is not enabled? Shouldn't we throw an error? I wonder if just silently not force closing could be misleading

matheusbsilva137 (changes requested):

apps/meteor/app/livechat/server/api/v1/room.ts
KevLehman: It will throw an error. It will be the normal room closing error as of the setting was ignored.

Didn't want to add a new error just for this case, as it could be confusing

apps/meteor/app/livechat/server/lib/LivechatTyped.ts
KevLehman: Hmmmm, not sure.

I mean, there's no other place that we expose that can use this param right now so thought checking at the caller makes sense.

Wdyt?

.changeset/eight-humans-sip.md
KevLehman: And this is what I got for using the changeset comment on here 🤣 

Good one

MarcosSpessatto (approved):

Harmeet221 (approved):

apps/meteor/app/livechat/server/lib/LivechatTyped.ts
matheusbsilva137: hmmm got your point. but can we have a third opinion? I'd like to hear @MarcosSpessatto 
I thought that having the check at this level too could help us not be caught off guard in case this new prop is used without the setting check in our codebase later.

KevLehman: Yeah, i got your point. But keep in mind we don't expose the closing mechanism to any place that a user can interact with. Apps are through a bridge we control, cronjobs are code we control. The only place that allows for user intervention are the 2 room closing endpoints, and one is for visitors and only allows to receive a roomId.

So in any possible scenario, we should be safe by letting the endpoint do what it wants to do.

MarcosSpessatto: I understand @matheusbsilva137 point, and it indeed makes sense. But I think we can go with Kevin's approach, in the future, if we need to use that somewhere else, then we could refactor, moving the validation to the lowest level.

matheusbsilva137 (approved):

PR #34958, opened 2025-01-17
feat: Make Omnichannel room opening transactional

KevLehman:

This aims to prevent orphan rooms caused by some actions timing out, or just the process timing out.

An orphan room could be a room with no inquiry, or, the incredible case (?), an inquiry without a room. This happens because there were some actions in between the room creation and the inquiry creation that could fail, leaving a room with no inquiry. There was the posibiity of the inquiry creation failing, which would leave the room orphaned as well.

Since now both are in a transaction, they should be always created altogether, making it safer and less error prone.
## Issue(s)

https://rocketchat.atlassian.net/browse/CORE-929


abhinavkrin (approved):

apps/meteor/app/livechat/server/lib/QueueManager.ts
```diff
@@ -213,6 +215,51 @@ export class QueueManager {
...
+				initialStatus: await this.getInquiryStatus({ room, agent: defaultAgent }),
+				guest,
+				message,
+				extraData: { ...extraData, source: roomInfo.source },
+				session,
+			});
+			const livechatSetting = await LivechatRooms.updateRoomCount(session);
+			if (livechatSetting) {
+				void notifyOnSettingChanged(livechatSetting);
+			}
```
pierre-lehnen-rc, line 245: Send the notify only after the update has been commited.

As it is if the commit fails it'll send a new notification with every retry.

pierre-lehnen-rc (changes requested):

apps/meteor/app/livechat/server/lib/QueueManager.ts
pierre-lehnen-rc: I'm not sure but since `updateRoomCount` runs an `$inc`, this may also end up sending a notification with the wrong value if there are other places changing that value in parallel.

KevLehman: Actually, i don't even know why we have this setting being incremented. Or propagated. Cannot see any place where it's being used and its a hidden setting.

Maybe we could use it for some metrics, but for sure doesn't need to be notified.

KevLehman: Removed notification. Will create a task to check where this can be used... or to just remove it.

MarcosSpessatto (dismissed):

pierre-lehnen-rc (approved):

MartinSchoeler (approved):

PR #34975, opened 2025-01-20
fix: stop invalid `Accounts_CustomFieldsToShowInUserInfo` value from breaking the ui

MartinSchoeler:

Since this setting is user defined, it is possible that an invalid value is provided, this pr enforces the type before doing a `map()` of the array, this should prevent the ui from breaking

also adds a better description to the setting
Screenshot 2025-01-20 at 15 23 57

## Issue(s)

[CORE-854](https://rocketchat.atlassian.net/browse/CORE-854)

## Further comments

[CORE-854]: https://rocketchat.atlassian.net/browse/CORE-854?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ](https://docs.rocket.chat/docs/custom-fields)


yash-rajpal (approved):
LGTM

apps/meteor/client/hooks/useUserCustomFields.ts
```diff
@@ -27,6 +27,11 @@ export const useUserCustomFields = (customFields: CustomField): CustomFieldDispl
 		return undefined;
 	}
 
+	if (!Array.isArray(customFieldsToShowObj)) {
```
MarcosSpessatto, line 30: Is there any possibility of this value being different than an array? Like booleans or numbers? Should we validate it too?

MarcosSpessatto (changes requested):

apps/meteor/client/hooks/useUserCustomFields.ts
MartinSchoeler: The only type of data that should be accepted is an array, so I don't think we should validate anything other that if the data is an array here, since in the following map we already have some checks for valid values

MarcosSpessatto (approved):

PR #34977, opened 2025-01-20
refactor: Search permissions with regex on a single string

gabriellsh:


juliajforesti (approved):

gabriellsh:
Closing since doing performance tests, this solutions is about 20x slower than what was being done previously.

PR #34985, opened 2025-01-21
chore: add dependency control on local and network brokers

ricardogarim:

This PR introduces the ability for Local and Network brokers to manage service dependencies, as outlined in [ARCH-1395](https://rocketchat.atlassian.net/browse/ARCH-1395).

For the Local Broker, we implemented this control manually, aligning the behavior with the Moleculer Framework's standard functionality. The start function iterates through all services, attempting to start them until a timeout is reached. Currently, the timeout is hardcoded to 10 seconds.

## Issue(s)
- [ARCH-1395](https://rocketchat.atlassian.net/browse/ARCH-1395)

## Further comments

[ARCH-1395]: https://rocketchat.atlassian.net/browse/ARCH-1395?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[ARCH-1395]: https://rocketchat.atlassian.net/browse/ARCH-1395?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Ricchkdbanks (approved):

packages/core-services/src/LocalBroker.ts
```diff
@@ -56,10 +69,17 @@ export class LocalBroker implements IBroker {
 		await instance.stopped();
 	}
 
-	createService(instance: IServiceClass): void {
-		const namespace = instance.getName();
+	/**
+	 * Creates a service and adds it to the local broker. In case of the broker is already started, it will start the service automatically.
+	 */
+	createService(instance: IServiceClass, serviceDependencies: string[] = []): void {
+		const namespace = instance.getName() ?? '';
```
MarcosSpessatto, line 76: Would this empty string be a problem? I mean, if an instance has no name, it seems an error, isn't it?

packages/core-services/src/LocalBroker.ts
```diff
@@ -8,14 +9,25 @@ import type { EventSignatures } from './events/Events';
...
 	private methods = new Map<string, (...params: any) => any>();
 
 	private events = new EventEmitter();
 
-	private services = new Set<IServiceClass>();
+	private services = new Map<string, ExtendedServiceClass>();
+
+	private flattenedDependencies: Set<string> = new Set();
+
+	private defaultDependencies = ['settings'];
```
MarcosSpessatto, line 30: Wouldn't License a default one too?

MarcosSpessatto (changes requested):

packages/core-services/src/LocalBroker.ts
sampaiodiego: no. for `LocalBroker` a license is not required (even though it will run anyways)

packages/core-services/src/types/ServiceClass.ts
```diff
@@ -55,6 +55,7 @@ export abstract class ServiceClass implements IServiceClass {
 	protected api?: IApiService;
 
 	constructor() {
+		this.name = this.getName();
```
sampaiodiego, line 58:
```suggestion
```

packages/core-services/src/types/ServiceClass.ts
```diff
@@ -46,7 +46,7 @@ export interface IServiceClass {
 }
 
 export abstract class ServiceClass implements IServiceClass {
-	protected name?: string;
+	protected name: string;
```
sampaiodiego, line 49:
```suggestion
	protected abstract name: string;
```

sampaiodiego (changes requested):

packages/core-services/src/LocalBroker.ts
```diff
@@ -111,8 +142,81 @@ export class LocalBroker implements IBroker {
...
+					logger.debug(`All services available: ${availableServices.map((s) => s.instance.getName()).join(', ')}`);
+					clearInterval(intervalId);
+					return resolve();
+				}
+
+				if (elapsed > TIMEOUT) {
+					clearInterval(intervalId);
+					const pendingServices = Array.from(this.pendingServices).join(', ');
+					logger.error(`Timeout while waiting for LocalBroker services: ${pendingServices}`);
+					return reject(new Error(`Timeout while waiting for LocalBroker services: ${pendingServices}`));
```
sampaiodiego, line 208:
```suggestion
					const error = new Error(`Timeout while waiting for LocalBroker services: ${pendingServices}`)
					logger.error(error);
					return reject(error);
```

packages/core-services/src/LocalBroker.ts
```diff
@@ -111,8 +142,81 @@ export class LocalBroker implements IBroker {
...
+			const intervalId = setInterval(async () => {
+				const elapsed = Date.now() - startTime;
+
+				const availableServices = Array.from(this.services.values()).filter((service) => service.isStarted);
+
+				if (this.pendingServices.size === 0) {
+					logger.debug(`All services available: ${availableServices.map((s) => s.instance.getName()).join(', ')}`);
+					clearInterval(intervalId);
+					return resolve();
+				}
```
sampaiodiego, line 202: there is a log already with every service name, so this looks not so relevant at this point and too noisy as well.

I think it could also be a `info` log here ;)

```suggestion
				if (this.pendingServices.size === 0) {
					const availableServices = Array.from(this.services.values()).filter((service) => service.isStarted);

					logger.info(`All ${availableServices.length} services available`);
					clearInterval(intervalId);
					return resolve();
				}
```

sampaiodiego (changes requested):

sampaiodiego (approved):

MarcosSpessatto (approved):

PR #34988, opened 2025-01-21
fix: Allow pagination params on `channels.list`

KevLehman:

The type was right, but the schema didn't include the properties, and since we had `additionalProperties` as false, AJV rejected the params

## Issue(s)

https://rocketchat.atlassian.net/browse/SUP-730


lucas-a-pelegrino (approved):

packages/rest-typings/src/v1/channels/ChannelsListProps.ts
```diff
@@ -6,18 +6,25 @@ const ajv = new Ajv({
...
 const channelsListPropsSchema = {
 	type: 'object',
 	properties: {
 		_id: {
 			type: 'string',
-			nullable: true,
 		},
 		query: {
 			type: 'string',
-			nullable: true,
```
MarcosSpessatto, line 20: Why are we removing nullable from query?

MarcosSpessatto (changes requested):

packages/rest-typings/src/v1/channels/ChannelsListProps.ts
KevLehman: Actually, it's because these are optional ones. The nullable applies to required ones to allow null. But in these cases we shouldn't allow it 🤔 

I can remove, but as this was a smol change and shouldn't affect (no test, and surely no one sending null to these params) though to keep.

Wdyt?

MarcosSpessatto (approved):

PR #34999, opened 2025-01-22
chore: remove restivus

ggazzo:
[ARCH-1450]

As part of the initiative to move rest to an external package, we must to remove dependencies related to meteor.
Restivus was a old library created to have access to webrouter, this is no longer the default, we have express exposed and its how we should create new endpoints.

## Further comments

[ARCH-1450]: https://rocketchat.atlassian.net/browse/ARCH-1450?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Ricchkdbanks (commented):
Please merge

apps/meteor/.mocharc.js
```diff
@@ -36,7 +36,7 @@ module.exports = {
 		'tests/unit/lib/**/*.spec.ts',
 		'tests/unit/server/**/*.tests.ts',
 		'tests/unit/server/**/*.spec.ts',
-		'app/api/**/*.spec.ts',
+		'app/api/lib/**/*.spec.ts',
```
sampaiodiego, line 39:
```suggestion
		'app/api/server/lib/**/*.spec.ts',
```

sampaiodiego (changes requested):

apps/meteor/app/integrations/server/api/api.js
```diff
@@ -237,6 +240,51 @@ function integrationInfoRest() {
...
+						},
+					},
+				};
+			}
+		}
+
+		request.integration = await Integrations.findOne({
+			_id: request.params.integrationId,
+			token: decodeURIComponent(request.params.token),
+		});
```
KevLehman, line 264: Should we have this one on model? Can be done later, a TODO may be enough.
But if we can do it on here :)

apps/meteor/app/integrations/server/api/api.js
```diff
@@ -237,6 +240,51 @@ function integrationInfoRest() {
...
+					},
+				},
+			};
+		}
+
+		const user = await Users.findOne({
+			_id: request.integration.userId,
+		});
+
+		return user;
```
KevLehman, line 284:
```suggestion
		return Users.findOneById(request.integration.userId);
```

apps/meteor/app/livechat/imports/server/rest/departments.ts
```diff
@@ -87,6 +87,22 @@ API.v1.addRoute(
 	},
 );
 
+API.v1.addRoute(
+	'livechat/department/isDepartmentCreationAvailable',
```
KevLehman, line 91: So we have to be more worried about path collisions now :(

This is something that would be great to be shared somewhere so no one gets bitten by it.

apps/meteor/app/api/server/router.ts
```diff
@@ -0,0 +1,256 @@
...
+		| TOperations
+		| ({
+				method: Method;
+				path: TPathPattern;
+		  } & Omit<TOptions, 'response'>)
+	> {
+		const prev = this.middleware;
+		this.middleware = (router: express.Router) => {
+			prev(router);
+			// console.log('MIDDLEWARE', method, subpath);
```
KevLehman, line 88:
```suggestion
```

KevLehman (commented):
image
Not sure if this happened before, but enabling Prometheus and going to localhost:9458 returns this "index" page. 

Should we keep? Should we remove?

apps/meteor/app/integrations/server/api/api.js
ggazzo: Not sure if I followed, do you mean create a 'getByIdAndToken' ? I dont think is  related to the objetive of the pull request, this is 100% exactly how the previous code was doing

apps/meteor/app/livechat/imports/server/rest/departments.ts
ggazzo: why we should be **more**  worried? to you have any extra explanation in mind? now at least we follow the same rules as express, it should be easy to find how it works

ggazzo:
> image Not sure if this happened before, but enabling Prometheus and going to localhost:9458 returns this "index" page.
> 
> Should we keep? Should we remove?

I think it is the exact same behavior :)

julio-rocketchat:
This PR will definitely help with Iron Bank and with our vulnerability management as a whole. We have some issues related to Restivus since it's an old package that uses old dependencies.

apps/meteor/app/livechat/imports/server/rest/departments.ts
KevLehman: Oh, because we apparently used to support things like 
path/:_id
path/something

Which express doesn't iirc.

apps/meteor/app/integrations/server/api/api.js
KevLehman: Yup, that's the idea. Agree on the oos, that's why I suggested the TODO. I'm tracking and fixing some of them.

sampaiodiego (approved):

PR #35006, opened 2025-01-22
fix: voip call history doesn't always log the correct call information

pierre-lehnen-rc:
## Issue(s)

Several problems with the data currently stored on the VoIP call history:
- It doesn't always identify the right caller / callee.
- It logs too many "empty" calls that didn't happen (Still unknown why freeswitch is sending the events that are causing those in the first place)
- If the same freeswitch server is used elsewhere, it is logging all of its calls as if the rocket.chat workspace had been involved on them.
- If the same freeswitch server is used in multiple workspaces, it will log every call in every workspace, each of them with their own users.

## Further comments

- this is technically a regression;
- wip
- moved the event parsing code from the freeswitch service to the freeswitch package as it was getting too big.
- changed the schema of the collecions used to store the events and calls (including an index)
- changed the client to send custom data to the freeswitch server, such as the workspace url - this is then used to identify if the call should be logged or not and also to identify which users are internal or external.

Caveats:
- If an incoming external call doesn't reach the rocket.chat user (either because they are offline or because they ignored the call without accepting or rejecting it), then the server will not receive any workspace information from the client and the call will not be logged. If the call was initiated from another rocket.chat workspace, it'll be logged as "internal" there.


packages/freeswitch/src/eventParser/computeChannelFromEvents.ts
```diff
@@ -0,0 +1,119 @@
...
+				...state,
+				eventName,
+				sequence,
+				...changedValues,
+			};
+		},
+		{} as Record<string, any>,
+	);
+
+	const finalState = convertPathsIntoSubObjects(flattened) as IFreeSwitchChannelEventMutable;
```
cardoso, line 129: Also avoid this typecast, perhaps flattened could have the right type and via generics `convertPathsIntoSubObjects` would return what you need.

packages/freeswitch/src/eventParser/computeChannelFromEvents.ts
```diff
@@ -0,0 +1,119 @@
...
+				firedAt,
+				receivedAt,
+				caller,
+				callee,
+
+				newValues: convertPathsIntoSubObjects(newValues),
+				modifiedValues: convertPathsIntoSubObjects(changedExistingValues),
+			};
+
+			deltas.push(filterOutMissingData(delta) as IFreeSwitchChannelEventDeltaData);
```
cardoso: With all these mutations of outer scope variables and complex logic, I think it's better to use a loop here instead of reduce. This isn't going to be fun to debug in the future.

packages/freeswitch/src/eventParser/extractChannelChangesFromEvent.ts
```diff
@@ -0,0 +1,72 @@
+// Make an object with a list of changes from the previous accumulated channel state and the state of the next event
+export function extractChannelChangesFromEvent(
+	channelState: Record<string, any>,
+	_eventName: string,
+	eventValues: Record<string, any>,
+): {
+	changedValues: Record<string, any>;
+	newValues: Record<string, any>;
+	changedExistingValues: Record<string, { oldValue: any; newValue: any; delta?: number }>;
```
cardoso, line 9: We should really not be introducing new anys if possible. I think this could be done with some generics.

packages/freeswitch/src/eventParser/filterOutMissingData.ts
```diff
@@ -0,0 +1,26 @@
...
+
+			if (typeof value === 'object' && !(value instanceof Date) && !Array.isArray(value) && !Object.keys(value).length) {
+				return;
+			}
+
+			return { key, value };
+		},
+		true,
+	) as DeepPartial<T>;
+}
```
cardoso: We should avoid introducing explicit any here and using implicit anys, you may need to tweak the typings of `objectMap`.

packages/freeswitch/src/eventParser/filterOutMissingData.ts
```diff
@@ -0,0 +1,26 @@
...
+
+export function filterOutEmptyValues<T extends Record<string, string | undefined>>(data: T): Record<keyof T, string> {
+	return filterOutMissingData(data) as Record<keyof T, string>;
+}
```
cardoso: Is exporting both of these functions needed? Maybe the filterOutMissingData should just not be exported?

packages/freeswitch/src/eventParser/parseEventData.ts
```diff
@@ -0,0 +1,168 @@
...
+				},
+			},
+		},
+
+		presenceCallDirection,
+		channelPresenceId,
+	};
+
+	const filteredEvent = {
+		...(filterOutMissingData(event) as typeof event),
```
cardoso: filterOutMissingData should just give the same type back instead of requiring this cast.

packages/tools/src/convertPathsIntoSubObjects.ts
```diff
@@ -0,0 +1,39 @@
+export function convertPathsIntoSubObjects(object: Record<string, any>): Record<string, any> {
+	const newObject: Record<string, any> = {};
```
cardoso, line 2: Avoid any if possible, unknown is better.

packages/tools/src/convertPathsIntoSubObjects.ts
```diff
@@ -0,0 +1,39 @@
...
+
+	for (const key in object) {
+		if (!(key in object)) {
+			continue;
+		}
+		const value = object[key];
```
cardoso: Is this continue reachable? Should be `Object.hasOwn` or loop over `Object.entries` instead since the value is used below.

packages/ui-voip/src/lib/VoipClient.ts
```diff
@@ -936,6 +943,33 @@ class VoipClient extends Emitter<VoipEvents> {
...
 	};
+
+	private getContactHostName(): string | undefined {
+		try {
+			const url = new URL(this.config.siteUrl);
+			return url.hostname;
+		} catch {
+			//
+		}
+	}
```
cardoso: Better to return undefined explicitly, however it's fine since this is private...

packages/freeswitch/src/eventParser/computeChannelFromEvents.ts
pierre-lehnen-rc: using generics here would be too heavy and would have zero benefits (`convertPathsIntoSubObjects` is not exported by the package)

packages/freeswitch/src/eventParser/extractChannelChangesFromEvent.ts
pierre-lehnen-rc: none of those are exposed by the package.

packages/freeswitch/src/eventParser/filterOutMissingData.ts
pierre-lehnen-rc: what are you referring to?

packages/freeswitch/src/eventParser/computeChannelFromEvents.ts
pierre-lehnen-rc: I've split it into multiple loops

cardoso (approved):

dougfabris (dismissed):
FE looks good!

MarcosSpessatto (approved):

PR #35013, opened 2025-01-23
feat: filter thread messages in room message query

Dnouv:

- Currently when making a call to the bridge API for reading room messages, or reading unread messages, all the messages including the ones inside a thread are being sent, which makes it complex to filter out and group together a single thread messages
- With introduction of a new option, `threadMsgCount` would be shown on the main message of a thread, the id of this message can then be used to fetch all the threaded messages using the `thread` read bridge.

## Further comments

[RCAI4-5](https://rocketchat.atlassian.net/browse/RCAI4-5)

[RCAI4-5]: https://rocketchat.atlassian.net/browse/RCAI4-5?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


d-gubert (changes requested):
To be honest my first reaction was to be against adding a new flag for this, but since that's how the core models expose the filter, then I'm ok with it :stuck_out_tongue: 

A few things:
- I think it's best to keep the name of the parameter the model method uses, `showThreadMessages` - it will be better for us maintining it to see a familiar name.
- When adding parameters to functions, they _cannot_ change the original default behavior of the function - that would be a breaking change. As such, the default value of the option to show thread messages MUST be `true`, and the user has to opt out of showing thread messages.

packages/apps-engine/src/definition/accessors/IRoomRead.ts
```diff
@@ -100,6 +101,7 @@ export interface IRoomRead {
      *                - limit: The maximum number of messages to retrieve. If more than 100 is passed, it defaults to 100.
      *                - skip: The number of messages to skip (for pagination).
      *                - sort: An object defining the sorting order of the messages. Each key is a field to sort by, and the value is either 'asc' for ascending order or 'desc' for descending order.
+     *               - includeThreads: Whether to include thread messages in the results. Defaults to false.
```
d-gubert, line 104:
```suggestion
     *                - includeThreads: Whether to include thread messages in the results. Defaults to false.
```

packages/apps-engine/src/definition/accessors/IRoomRead.ts
```diff
@@ -48,6 +48,7 @@ export interface IRoomRead {
      *                - limit: The maximum number of messages to retrieve. Maximum 100
      *                - skip: The number of messages to skip (for pagination).
      *                - sort: An object defining the sorting order of the messages. Each key is a field to sort by, and the value is either "asc" for ascending order or "desc" for descending order.
+     *               - includeThreads: Whether to include thread messages in the results. Defaults to false.
```
d-gubert, line 51:
```suggestion
     *                - includeThreads: Whether to include thread messages in the results. Defaults to false.
```

PR #35017, opened 2025-01-23
chore: Stop using `Livechat_Room_Count` setting

KevLehman:

Setting was not being used anywhere, we can safely remove.

Thought on using it for metrics, but it's generally easier a count on the collection with the `estimatedDocumentCount` we have right now than to keep track of this value.
## Issue(s)

https://rocketchat.atlassian.net/browse/CORE-930
## Further comments

Ok, maybe we're not removing here entirely, but the point is to stop using it. There's another task for the migration to remove it.


MarcosSpessatto (changes requested):
Do we need to create a task to create a migration to a major version? So we can remove the setting?

KevLehman:
Yup, I'm creating the other task to remove the setting. This one will just stop using it (as we're updating it for no reason everytime a new Omni room is created)

MarcosSpessatto (approved):

MartinSchoeler (approved):

PR #34113, opened 2024-12-04
fix: wrong dates on last `x` date on omnichannel reports

MartinSchoeler:
## Proposed changes (including videos or screenshots)

Fixes dates being incorrect when fetching omnichannel reports, for example  currently asking for last month will return the last 30 days, even if the month just started, this fixes

`last week`
`last month`
`last 6 months`
`last year`

## Issue(s)

[SUP-703](https://rocketchat.atlassian.net/browse/SUP-703)

[SUP-703]: https://rocketchat.atlassian.net/browse/SUP-703?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #35019, opened 2025-01-23
fix: engagement dashboard message and hourly activity charts

MartinSchoeler:
### Steps to reproduce

1. Send some messages in the server using different users
2. restart the server (or do something to re run the cron job that gets statistics)
3. Go to admin -> engagement Dashboard (while in an enterprise server) 
4. Scroll down to `When is the chat busier?` section
5. The chart should show info in the correct timezone

Fixes issue with hourly chat activity chart not accounting for utc, also restores some missing data visualizers on the messages chart

About tests, this is a similar situation as this one https://github.com/RocketChat/Rocket.Chat/pull/34113#issuecomment-2549392245

## Issue(s)

[SUP-676](https://rocketchat.atlassian.net/browse/SUP-676)

[SUP-676]: https://rocketchat.atlassian.net/browse/SUP-676?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

---

This pull request addresses issues with the engagement dashboard in the Rocket.Chat repository, specifically focusing on the message and hourly activity charts. The changes include:

1. **Timezone Support**: A new `timezone` prop has been introduced to the `MessagesTab` and `MessagesSentSection` components to ensure accurate local time display. The `useMessagesSent` hook now supports UTC timezone and has an updated query key structure.

6. **Chart Configuration**: Adjustments have been made to the `MessagesSentSection` component, including changes to the chart container's height and refinements to the `ResponsiveBar` component's configuration, such as axis and tooltip settings.

7. **Simplified Logic**: In the `NewUsersSection` component, the use of a resize observer has been removed, and the tick calculation logic for the chart has been simplified. Additionally, the chart's styling and tick formatting have been adjusted.

8. **Testing Enhancements**: A new test suite for the `useHourlyChatActivity` hook has been introduced, featuring two test cases to verify the functionality of returning UTC and local time data. The tests utilize `@testing-library/react` for rendering hooks and `jest` for mocking the `moment` library.

9. **Hourly Activity Data**: UTC time conversion logic has been added for hourly chat activity data in the `useHourlyChatActivity` component.

These changes aim to improve the accuracy and maintainability of the engagement dashboard's data visualizations.


apps/meteor/client/views/admin/engagementDashboard/users/useHourlyChatActivity.ts
```diff
@@ -21,6 +21,16 @@ export const useHourlyChatActivity = ({ displacement, utc }: UseHourlyChatActivi
...
 
+			if (utc) {
+				response.hours = response.hours.map((hours) => {
+					const newDay = moment().hour(hours.hour).subtract(displacement, 'days').toDate();
+					return {
+						hour: moment.utc(newDay, 'HH').hour(),
+						users: hours.users,
+					};
+				});
+			}
```
gabriellsh, line 32: I found this a little strange, since the data should be coming correctly from the BE.

During investigation, I think I figured out what is happening: We're sending the correct dates to the backend. When set to local timezone, we send the `start` parameter as midnight of the day in the current timezone, converted to an ISO string (ISO strings afaik are UTC). For example, GMT -3 end of day is `'2025-01-24T02:59:59.999Z'`, and UTC end of day is `'2025-01-23T23:59:59.999Z'`, meaning GMT -3 midnight is 3 hours after UTC midnight. So the parameter is correct.

The issue is that in the DB, the dates for `loginAt` are saved in UTC, and then we extract the hour (which is in UTC time), and sort it. When we show the data, we are not taking that into account, so the value for 12 am for example is actually the value for 12 am UTC, instead of GMT-3. The graph is interpreting the data as as GMT, when it's actually UTC time.

What we need to do is get those hours and convert them to GMT. I think we can do that in the `format` property of the `ResponsiveBar` component in the `ContentForHours` component (or maybe here if that doesn't work).

.changeset/witty-buttons-greet.md
```diff
@@ -0,0 +1,4 @@
+---
+"@rocket.chat/meteor": patch
+---
+Fixes issue with some charts on engagement dashboard not using utc and missing data visualizers
```
gabriellsh, line 4: Remember to update the changeset when the other requested changes are complete

gabriellsh (changes requested):

gabriellsh:
I've also noticed that we're displacing the days in the graph by 1 day. For example today (Thu, 23rd Jan), we are requesting data starting from 23 Jan 23:59, instead of 23 Jan 00:00

gabriellsh (dismissed):

apps/meteor/client/views/admin/engagementDashboard/users/useHourlyChatActivity.spec.ts
```diff
@@ -0,0 +1,101 @@
+import { mockAppRoot } from '@rocket.chat/mock-providers';
+import { renderHook, waitFor } from '@testing-library/react';
+
+import { useHourlyChatActivity } from './useHourlyChatActivity';
+
+jest.mock('moment', () => {
+	const moment = jest.requireActual('moment-timezone');
+	moment.tz.setDefault('America/Los_Angeles');
```
tassoevan, line 8: Why?

tassoevan (commented):
I've some restrictions in mocking `moment` and other `Date`-handling libraries *without* mocking the system time with Jest's fake timers.

aleksandernsilva (changes requested):
Would be a good ideia to add a task to our backlog to refactor all these places using moment to date-fns.

aleksandernsilva (approved):

PR #35022, opened 2025-01-23
refactor: Remove meteor from OTR (startup)

MartinSchoeler:
Remove meteor functions from the OTR messaging flow

Demo gif:
remove_meteor_OTR

Jira task: [ARCH-1478](https://rocketchat.atlassian.net/browse/ARCH-1478)

[ARCH-1478]: https://rocketchat.atlassian.net/browse/ARCH-1478?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Steps to test
1. Login as 2 different users on two different sessions;
2. Open DMs for each one;
3. Click on the kebab menu on the top right of the screen and select `OTR`;
4. Accept the OTR request on the other user;
5. Send messages to one another;
6. When done click `End OTR`;


aleksandernsilva (changes requested):
I noticed some reactivity issues after one of the users logs out and back in. No obvious console errors, but the app stays in the bad state until the user refreshes the page.

https://github.com/user-attachments/assets/c32ef602-a7c6-433b-94ab-af8e02491f12

it's possible that this bug was already happening, but I believe it's worth taking a quick look.

rique223:
> I noticed some reactivity issues after one of the users logs out and back in. No obvious console errors, but the app stays in the bad state until the user refreshes the page.
> 
>  Screen.Recording.2025-03-10.at.20.03.22.mov 
> it's possible that this bug was already happening, but I believe it's worth taking a quick look.

I've tested it both on develop and on this branch, and wasn't able to reproduce the issue, maybe it has something to do with your workspace database having more data, your RAM being full/closer or some OS bottle neck:

rique223 (approved):

PR #35031, opened 2025-01-24
refactor: Moves `getUserDisplayName` outside from meteor app

dougfabris:

It moves the auxiliary `getUserDisplayName` function to `core-typings` and `useUserDisplayName` hook to `ui-client` in order to give more flexibility to use them outside from meteor app


gabriellsh (approved):

packages/core-typings/src/IUser.ts
```diff
@@ -264,3 +264,6 @@ export type AvatarServiceObject = {
 };
 
 export type AvatarObject = AvatarReset | AvatarUrlObj | FormData | AvatarServiceObject;
+
+export const getUserDisplayName = (name: IUser['name'], username: IUser['username'], useRealName: boolean): string | undefined =>
```
MarcosSpessatto, line 268: I know we tested the hook already, but can we add a unit test for this?

MarcosSpessatto (changes requested):

MartinSchoeler (approved):

PR #35041, opened 2025-01-27
regression: do not update logged out sessions

sampaiodiego:

Regression for https://github.com/RocketChat/Rocket.Chat/pull/35010 which was causing an issue when the same session was logging in and out multiple times.

The problem was that if the new log in was using the same `sessionId` as a previous one, that was probably logged out already, it was not creating a new session, so when the code was trying to to log out the session it couldn't find it, causing the error to be thrown.

I also changed to not throw an error if the session is not found.

```
W20250127-13:22:58.343(-3)? (STDERR) === UnHandledPromiseRejection ===
W20250127-13:22:58.344(-3)? (STDERR) Error: Session not found
W20250127-13:22:58.344(-3)? (STDERR)     at app/statistics/server/lib/SAUMonitor.ts:151:11
W20250127-13:22:58.344(-3)? (STDERR)     at processTicksAndRejections (node:internal/process/task_queues:105:5)
W20250127-13:22:58.344(-3)? (STDERR) ---------------------------------
W20250127-13:22:58.344(-3)? (STDERR) Errors like this can cause oplog processing errors.
W20250127-13:22:58.344(-3)? (STDERR) Setting EXIT_UNHANDLEDPROMISEREJECTION will cause the process to exit allowing your service to automatically restart the process
W20250127-13:22:58.344(-3)? (STDERR) Future node.js versions will automatically exit the process
W20250127-13:22:58.344(-3)? (STDERR) =================================
```

### DIFF
--- a/apps/meteor/app/statistics/server/lib/SAUMonitor.ts
+++ b/apps/meteor/app/statistics/server/lib/SAUMonitor.ts
@@ -143,7 +143,8 @@ export class SAUMonitorClass {
 				projection: { loginToken: 1 },
 			});
 			if (!session?.loginToken) {
-				throw new Error('Session not found');
+				logger.error('Session not found during logout', { userId, sessionId });
+				return;
 			}
 
 			await Sessions.logoutBySessionIdAndUserId({ loginToken: session.loginToken, userId });
--- a/apps/meteor/app/statistics/server/lib/SAUMonitor.ts
+++ b/apps/meteor/app/statistics/server/lib/SAUMonitor.ts
@@ -30,6 +30,8 @@ const getUserRoles = mem(
 	{ maxAge: 5000 },
 );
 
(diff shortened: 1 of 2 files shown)


apps/meteor/app/statistics/server/lib/SAUMonitor.ts
```diff
@@ -143,7 +143,8 @@ export class SAUMonitorClass {
 				projection: { loginToken: 1 },
 			});
 			if (!session?.loginToken) {
-				throw new Error('Session not found');
+				logger.error('Session not found during logout', { userId, sessionId });
```
ggazzo, line 146: now that you solved the issue would be the case of keeping the throw?

thanks to `throw` we finally solved the issue

ggazzo (approved):

PR #35044, opened 2025-01-28
fix: Departments could be removed via API even with setting `Omnichannel_enable_department_removal` off

KevLehman:
## Issue(s)

https://rocketchat.atlassian.net/browse/CORE-955


apps/meteor/tests/end-to-end/api/livechat/00-rooms.ts
```diff
@@ -76,12 +76,16 @@ describe('LIVECHAT - rooms', () => {
...
 		await updateEESetting('Livechat_Require_Contact_Verification', 'never');
+		await updateSetting('Omnichannel_enable_department_removal', true);
 		await createAgent();
 		await makeAgentAvailable();
 		visitor = await createVisitor();
 
 		room = await createLivechatRoom(visitor.token);
 	});
+	after(async () => {
+		await updateSetting('Omnichannel_enable_department_removal', true);
```
abhinavkrin, line 87: Should it be false? Since at all other "after" block it was set to false.

abhinavkrin (changes requested):

apps/meteor/tests/end-to-end/api/livechat/00-rooms.ts
KevLehman, line 87:
```suggestion
		await updateSetting('Omnichannel_enable_department_removal', false);
```

KevLehman: Thx!

lucas-a-pelegrino (approved):

MartinSchoeler (approved):

abhinavkrin (approved):

PR #35048, opened 2025-01-28
regression: sidebar scrolling horizontally

juliajforesti:

Remove class from the sidebar item menu button, as it is already present on its parent element. The style rules of this class was causing a horizontal scroll on the sidebar of the new navigation.


MartinSchoeler (approved):

dougfabris:
I think there are some reason why I added this class to the menu, please give me some time to look into it, until we get it merged

tassoevan:
@dougfabris I went ahead with because, regardless of the reasoning, it isn't right. Please keep investigating; an edge case can be covered in a patch.

Issue #34840, opened 2024-12-27
7.1.0 doesn't run as non-root while 7.0.0 does


jeanfabrice:
### Description:

After upgrading to 7.1.0, Rocketchat doesn't start anymore on Kubernetes when running as non root with the following securityContext:
```
          allowPrivilegeEscalation: false
          runAsUser: 1000
          runAsGroup: 1000
          capabilities:
            drop:
            - ALL
          privileged: false
          readOnlyRootFilesystem: true
```

Error thrown:
```
error on boot.js Error: EACCES: permission denied, open '/app/bundle/programs/server/npm/node_modules/meteor/webapp/node_modules/useragent-ng/lib/regexps.js'
    at Object.readFileSync (node:fs:448:20)
    at getMaybeCachedSource (node:internal/modules/cjs/loader:1492:18)
    at Object.Module._extensions..js (node:internal/modules/cjs/loader:1504:19)
    at Module.load (node:internal/modules/cjs/loader:1288:32)
    at Module.Mp.load (/tools/static-assets/server/runtime.js:35:31)
    at Function.Module._load (node:internal/modules/cjs/loader:1104:12)
    at Module.require (node:internal/modules/cjs/loader:1311:19)
    at Module.mod.require (/app/bundle/programs/server/node_modules/@meteorjs/reify/lib/runtime/index.js:30:33)
    at require (node:internal/modules/helpers:179:18)
    at Object.<anonymous> (/app/bundle/programs/server/npm/node_modules/meteor/webapp/node_modules/useragent-ng/index.js:6:15)
    at Module._compile (node:internal/modules/cjs/loader:1469:14)
    at Module.Mp._compile (/tools/static-assets/server/runtime.js:78:21)
    at Object.Module._extensions..js (node:internal/modules/cjs/loader:1548:10)
    at Module.load (node:internal/modules/cjs/loader:1288:32)
    at Module.Mp.load (/tools/static-assets/server/runtime.js:35:31)
    at Function.Module._load (node:internal/modules/cjs/loader:1104:12) {
  errno: -13,
  code: 'EACCES',
  syscall: 'open',
  path: '/app/bundle/programs/server/npm/node_modules/meteor/webapp/node_modules/useragent-ng/lib/regexps.js'
}
```

[...]

PR #35055, opened 2025-01-29
chore: update useragent-ng to 2.4.4

debdutdeb:
## Further comments

closes https://github.com/RocketChat/Rocket.Chat/issues/34840


julio-rocketchat:
Should this change also be made to our Iron Bank image or is it unrelated, @debdutdeb ?

debdutdeb:
@julio-rocketchat I have a separate patch there that works around the problem. There's a `run chmod a+r` line there. We don't need to change anything there for now. Once meteor 3.1.2 releases we can take this npm install and that chmod line both out

sampaiodiego (approved):

PR #35059, opened 2025-01-29
chore: add validation layer to marketplace comms

Gustrb:

This stems from the fact the we currently simply call marketplace-api from our core application without ever validating the return type of the API is actually what we expected. Since marketplace-api is written in a different programming language, and it is not open source, people tend to have less visibility when something changes, so we must be able to catch that as soon as possible, and never introduce a broken return inside our database (really hard to fix, since it would require migrating the data, either continuously or preemptively).

## Further comments

[CONN-64](https://rocketchat.atlassian.net/browse/CONN-64)
[CONN-491](https://rocketchat.atlassian.net/browse/CONN-491)
[CONN-492](https://rocketchat.atlassian.net/browse/CONN-492)

[CONN-64]: https://rocketchat.atlassian.net/browse/CONN-64?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/meteor/ee/server/apps/marketplace/MarketplaceAppsError.ts
KevLehman, line 1: nitpick: Since you hold both errors on the file, would make sense to make the file name more generic

apps/meteor/ee/server/apps/marketplace/fetchMarketplaceApps.ts
```diff
@@ -0,0 +1,177 @@
...
+	}
+
+	const response = await request.json();
+	if (request.status === 400 && response.code === 200) {
+		throw new MarketplaceAppsError('Marketplace_Invalid_Apps_Engine_Version');
+	}
+
+	const INTERNAL_MARKETPLACE_ERROR_CODES = [266, 256, 166, 221, 257, 320];
+
+	if (request.status === 500 && INTERNAL_MARKETPLACE_ERROR_CODES.includes(response.code)) {
```
KevLehman, line 172: Should we at least debuglog the actual error code/message?

apps/meteor/ee/server/apps/marketplace/getMarketplaceHeaders.ts
```diff
@@ -0,0 +1,9 @@
+import { Info } from '../../../../app/utils/rocketchat.info';
+
+const appsEngineVersionForMarketplace = Info.marketplaceApiVersion.replace(/-.*/g, '');
+
+export function getMarketplaceHeaders(): Record<string, any> {
```
KevLehman, line 5: Do you need this one to be a func? you can export the object directly 👀 as it apparently won't change during the runtime.

KevLehman (changes requested):

apps/meteor/ee/server/apps/marketplace/MarketplaceAppsError.ts
Gustrb: oh yea, sure

apps/meteor/ee/server/apps/marketplace/fetchMarketplaceApps.ts
Gustrb: oh yea, sure thing

MarcosSpessatto (approved):

KevLehman (approved):

PR #35069, opened 2025-01-30
refactor: remove method calls from chat endpoints (server)

MarcosSpessatto:
https://rocketchat.atlassian.net/browse/CORE-963


apps/meteor/app/lib/server/methods/getSingleMessage.ts
```diff
@@ -13,26 +13,30 @@ declare module '@rocket.chat/ddp-client' {
...
+	if (!(await canAccessRoomIdAsync(msg.rid, userId))) {
+		throw new Meteor.Error('error-not-allowed', 'Not allowed', { method: 'getSingleMessage' });
+	}
+
+	return msg;
+};
+
 Meteor.methods<ServerMethods>({
 	async getSingleMessage(mid) {
-		check(mid, String);
```
sampaiodiego, line 18: should this `check` stay on the API layer instead of being moved to the function? I think it makes more sense on the API since the data in there is essentially unknown and needs to be verified, but for the context of the function it should be fine already.

it is also inconsistent on how dealt with this situation across the whole PR.. on `pinMessage` for example you left the `check` on the Meteor method..

sampaiodiego (changes requested):

sampaiodiego (dismissed):

PR #35093, opened 2025-02-02
chore(deps): remove PostCSS IE11 plugin

julio-rocketchat:
`postcss-ie11-supports` uses vulnerable dependencies and has been abandoned for a couple of years now. Since IE 11 has reached EOL in 2022, it doesn't seem like we should or need to support this plugin anymore - notwithstanding, this plugin may not even be used by us anymore but the dependency is still there (need the FE and ARCH teams to confirm).


dougfabris (approved):
Couldn't find any `@supports` block so I think we're good to remove it

PR #35100, opened 2025-02-03
fix: use correct start dates on omnichannel reports

MartinSchoeler:

The omnichannel reports page was getting data for one day earlier than supposed in any option that required a date that starts at the start of the month

## Issue(s)

[CORE-972](https://rocketchat.atlassian.net/browse/CORE-972)

## Further comments

[CORE-972]: https://rocketchat.atlassian.net/browse/CORE-972?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


MartinSchoeler:
Screenshot 2025-02-03 at 14 38 20
As expected the test is failing without the fix

gabriellsh (dismissed):

apps/meteor/client/components/dashboards/periods.ts
```diff
@@ -23,10 +23,22 @@ const lastNDays =
...
+		subtract = 0,
+	): ((utc: boolean) => {
+		start: Date;
+		end: Date;
+	}) =>
+	(utc): { start: Date; end: Date } => {
+		const date = utc ? new Date() : moment(new Date()).utc().toDate();
+
+		const start = moment(date).subtract(subtract, 'months').startOf(startOf).toDate();
+		const end = moment(date).endOf('day').toDate();
```
aleksandernsilva, line 38: I noticed two things:

- The ternary condition seems to be reversed. 
- Also it instantiates local time then shifts it to utc (`moment().utc()`). Not sure if this is the desired behavior, considering `lastNDays` returns utc+0 (`moment.utc()`) when `utc` flag is `true`
- It might be possible to reduce the number of conversions to `Date` and back to `moment` (not tested).

```suggestion
		const date = utc ? moment().utc().toDate() : new Date();

		const start = date.subtract(subtract, 'months').startOf(startOf).toDate();
		const end = date.endOf('day').toDate();
```

apps/meteor/client/components/dashboards/getClosedPeriod.spec.ts
```diff
@@ -0,0 +1,85 @@
...
+});
+
+it('should return the correct period range for this month', () => {
+	const monthExpectedReturn = {
+		start: new Date('5/1/2024').toISOString().split('T')[0],
+		end: new Date('5/19/2024').toISOString().split('T')[0],
+	};
+
+	expect(getClosedPeriod('month')(true).start.toISOString().split('T')[0]).toEqual(monthExpectedReturn.start);
+	expect(getClosedPeriod('month')(true).end.toISOString().split('T')[0]).toEqual(monthExpectedReturn.end);
```
aleksandernsilva, line 14:
```suggestion
        const period = getClosedPeriod('month')(true);
	expect(period.start.toISOString().split('T')[0]).toEqual(monthExpectedReturn.start);
	expect(period.end.toISOString().split('T')[0]).toEqual(monthExpectedReturn.end);
```
_The same can be applied to all other test cases._

aleksandernsilva (changes requested):
- This code is shared with the Engagement Dashboard, so I recommend taking a look there to see if everything is working as it should.
- Considering the distinction between utc+0 and local time utc, it might be a good idea to test the edge scenario were changing this flag causes the `start` and/or `end` dates to differ.

PS.: I tried running the tests locally and they were not passing, but CI seems fine 🤔

apps/meteor/client/components/dashboards/periods.ts
MartinSchoeler: > It might be possible to reduce the number of conversions to Date and back to moment (not tested).
The multiple `moment()` calls are to avoid isssues with [moment's mutability](https://momentjs.com/guides/#/lib-concepts/mutability/) , in your sugggestion the end value would be the end of the day set in the start variable

The ternary is indeed flipped, let me fix that

apps/meteor/client/components/dashboards/periods.ts
```diff
@@ -23,10 +23,22 @@ const lastNDays =
 		return { start, end };
 	};
 
-const getLast6Months = () => {
-	const last6Months = moment().subtract(6, 'months');
-	return moment().diff(last6Months, 'days');
-};
+export const getClosedPeriod =
```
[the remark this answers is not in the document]
MartinSchoeler: The `lastNDays` has already a lot of ternaries and conditions, IMO adding these changes to it would detriment the function readability quite a lot

MartinSchoeler:
> * This code is shared with the Engagement Dashboard, so I recommend taking a look there to see if everything is working as it should.
They are only using the values from `lastNDays` function, I've done some testing and everything seems to be ok

apps/meteor/client/components/dashboards/periods.ts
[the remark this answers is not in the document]
aleksandernsilva: I agree, `lastNDays` has more ternaries than it should, it could probably be simplified as well. 
We don't need to change things in this PR, but I believe this file could use some refactoring, including removing `moment` eventually.

aleksandernsilva (dismissed):

KevLehman (approved):
Removing the omni review :)

KevLehman (approved):

gabriellsh (approved):

PR #35107, opened 2025-02-03
fix(apps): room converter errors out when converting livechat room closed by the visitor

d-gubert:

If the closer of a livechat room is the visitor, the `closedBy` property of the room will contain the visitor's id, not the user's - which crashed the conversion from app format to core format as the converter was always trying to fetch a user from the DB.

The behavior now checks for the `closer` property to correctly assign the value to the `closedBy` property when converting back from the app.

## Further comments

Further changes will be done to types, however to make this a safely "backportable" PR, they will be done separately

### DIFF
--- a/.github/actions/update-version-durability/package-lock.json
+++ b/.github/actions/update-version-durability/package-lock.json
@@ -71,11 +71,12 @@
 			}
 		},
 		"node_modules/@octokit/endpoint": {
-			"version": "10.1.1",
-			"resolved": "https://registry.npmjs.org/@octokit/endpoint/-/endpoint-10.1.1.tgz",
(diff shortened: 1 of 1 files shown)


KevLehman (changes requested):
A changeset maybe? :awesome:

.changeset/popular-mugs-try.md
```diff
@@ -0,0 +1,5 @@
+---
+'@rocket.chat/meteor': patch
+---
+
+Fix an issue where the room converter would throw if the room was an omnichannel room that had been closed by the visitor
```
KevLehman, line 5:
```suggestion
Fixes an issue where the room converter would throw if the room was an omnichannel room that had been closed by the visitor
```

KevLehman (approved):

KevLehman (approved):

PR #35120, opened 2025-02-05
fix(apps): prevent greedy update via apps

d-gubert:

When apps update an entity (a room or a message) via their respective `ModifyUpdater` method, the bridge will actually update ALL fields of in the DB record. That is because it uses a `$set` instruction with the whole record sent by the app. This creates a problem where the app could have queried stale data and would inadvertedly overwrite the previously updated DB record with such stale data.

To prevent this scenario, this PR modifies the builder instances used by the updaters, `RoomBuilder` and `MessageBuilder`, to actually keep track of the changes applied by the app, and send only the changed properties to be updated in the bridge.

## Issue(s)

[CONN-480]
## Further comments

[CONN-480]: https://rocketchat.atlassian.net/browse/CONN-480?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/meteor/app/apps/server/converters/messages.js
```diff
@@ -176,14 +180,32 @@ export class AppMessagesConverter {
 
 		const attachments = this._convertAppAttachments(message.attachments);
 
+		let _id = message.id;
```
KevLehman, line 183: What about:
```
const _id = !isPartial && !message.id ? Random.id() : message.id
```

wonder if it's a bit better considering we're sacrificing a bit of read vs a bunch of {}

apps/meteor/app/apps/server/converters/messages.js
```diff
@@ -176,14 +180,32 @@ export class AppMessagesConverter {
...
+				_id = Random.id();
+			}
+
+			if (!message.createdAt) {
+				ts = new Date();
+			}
+
+			if (!message.updatedAt) {
+				_updatedAt = new Date();
+			}
```
KevLehman, line 198: I don't think we need to pass this as the updateOne will override it anyways with a new _updatedAt.

packages/apps-engine/deno-runtime/lib/accessors/builders/RoomBuilder.ts
```diff
@@ -160,4 +184,14 @@ export class RoomBuilder implements IRoomBuilder {
     public getRoom(): IRoom {
```
KevLehman, line 184: Should we use this opportunity to freeze this one? Or do you consider that to be a breaking change?

.changeset/five-cherries-hang.md
```diff
@@ -0,0 +1,20 @@
...
+'@rocket.chat/ddp-streamer': patch
+'@rocket.chat/queue-worker': patch
+'@rocket.chat/apps-engine': patch
+'@rocket.chat/ui-client': patch
+'@rocket.chat/apps': patch
+'@rocket.chat/i18n': patch
+'@rocket.chat/meteor': patch
+---
+
+Fix behavior of app updates that would save undesired field changes to documents
```
KevLehman, line 20:
```suggestion
Fixes behavior of app updates that would save undesired field changes to documents
```

apps/meteor/app/apps/server/bridges/rooms.ts
```diff
@@ -163,13 +163,13 @@ export class AppRoomBridge extends RoomBridge {
 	protected async update(room: IRoom, members: Array<string> = [], appId: string): Promise<void> {
 		this.orch.debugLog(`The App ${appId} is updating a room.`);
 
-		if (!room.id || !(await Rooms.findOneById(room.id))) {
-			throw new Error('A room must exist to update.');
-		}
+		const rm = await this.orch.getConverters()?.get('rooms').convertAppRoom(room, true);
 
-		const rm = await this.orch.getConverters()?.get('rooms').convertAppRoom(room);
+		const updateResult = await Rooms.updateOne({ _id: room.id }, { $set: rm });
```
KevLehman, line 168: btw, since we are doing manual notifications, should we notify that the room changed here? with `notifyRoomChanged`?

KevLehman (changes requested):

apps/meteor/app/apps/server/converters/messages.js
```diff
@@ -135,19 +135,23 @@ export class AppMessagesConverter {
...
-		if (!message || !message.room) {
+	async convertAppMessage(message, isPartial = false) {
+		if (!message) {
 			return undefined;
 		}
 
-		const room = await Rooms.findOneById(message.room.id);
+		let rid;
+		if (message.room?.id) {
+			const room = await Rooms.findOneById(message.room.id);
```
KevLehman, line 145:
```suggestion
			const room = await Rooms.findOneById(message.room.id, { projection: {_id: 1}});
```

apps/meteor/app/apps/server/bridges/rooms.ts
d-gubert: hmmm maybe? I think that deserves a different PR, though. To be honest, I'm unsure why we're doing the update manually

apps/meteor/app/apps/server/converters/messages.js
d-gubert: I've considered it, but my reasoning was to avoid repeated tests, as I'd need to do that same thing with the other variables (`!isPartial && ...`)

apps/meteor/app/apps/server/converters/messages.js
d-gubert: That's fair. I wanted to keep it in because the function did it previously. But the impact is minimal anyway.

KevLehman (approved):

PR #35127, opened 2025-02-05
fix: `EmojiPicker` not displaying active category properly

dougfabris:

Possibly introduced when we [upgraded virtuoso](https://github.com/RocketChat/Rocket.Chat/pull/31843)

Kapture 2025-02-06 at 16 58 52

## Further comments

[CORE-976]

[CORE-976]: https://rocketchat.atlassian.net/browse/CORE-976?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/meteor/client/views/composer/EmojiPicker/EmojiPicker.tsx
```diff
@@ -129,11 +129,16 @@ const EmojiPicker = ({ reference, onClose, onPickEmoji }: EmojiPickerProps) => {
...
+
+		if (!unmounted) {
+			setCurrentCategory('recent');
 		}
-	}, [actualTone, recentEmojis, getEmojiListsByCategory, currentCategory, setRecentEmojis]);
+
+		return () => {
+			unmounted = true;
+		};
+	}, [setCurrentCategory]);
```
aleksandernsilva, line 141: I don't think I understand this hook 🤔 

`unmounted` is declared and assigned `false`. The `if` in `L134` (`if(!unmounted)`) will always enter. 
Then on the cleanup the local variable is set to `true`, but at that point if there's a next render it will re-declare the variable.

Did you mean to set the variable outside the scope of the hook? 
The way it is, wouldn't the following serve the same purpose:
```js
useEffect(() => {
  setCurrentCategory('recent')
}, [setCurrentCategory])
```

aleksandernsilva (changes requested):

apps/meteor/client/views/composer/EmojiPicker/EmojiPicker.tsx
dougfabris: please, check [this](https://react.dev/learn/synchronizing-with-effects#fetching-data) out

aleksandernsilva (approved):

PR #35131, opened 2025-02-05
refactor: Use updater for saveCustomFields function

gabriellsh:

Avoid doing multiple DB queries when saving custom fields

This will also facilitate auditing user updates
## Issue(s)

[LOG-17](https://rocketchat.atlassian.net/browse/LOG-17)
## Further comments

[LOG-17]: https://rocketchat.atlassian.net/browse/LOG-17?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/meteor/app/lib/server/functions/saveCustomFieldsWithoutValidation.ts
```diff
@@ -1,49 +1,66 @@
...
-			customFields[key] = formData[key];
-		});
-		await Users.setCustomFields(userId, customFields);
+	const updater = _updater || Users.getUpdater();
 
-		// Update customFields of all Direct Messages' Rooms for userId
-		const setCustomFieldsResponse = await Subscriptions.setCustomFieldsDirectMessagesByUserId(userId, customFields);
-		if (setCustomFieldsResponse.modifiedCount) {
-			void notifyOnSubscriptionChangedByUserIdAndRoomType(userId, 'd');
+	updater.set('customFields', customFields);
```
pierre-lehnen-rc, line 40:
```suggestion
```
You're setting customFields as a whole here, then still iterating the object to set each field individually.

pierre-lehnen-rc (changes requested):

apps/meteor/app/api/server/v1/users.ts
```diff
@@ -104,10 +103,6 @@ API.v1.addRoute(
 
 			await saveUser(this.userId, userData);
 
-			if (this.bodyParams.data.customFields) {
```
pierre-lehnen-rc, line 107: This will cause a change in behavior: if there's any issue with the custom fields, this endpoint will no longer save anything to the user. With the old code it would save the changes to the user core data and only throw an error afterwards.

I don't know which behavior we want, but it might be a breaking change.

apps/meteor/app/lib/server/functions/saveCustomFieldsWithoutValidation.ts
```diff
@@ -1,49 +1,66 @@
...
+		} else {
+			updater.set(modifyRecordField.field, customFields[fieldName]);
 		}
+	});
+
+	if (!_updater) {
+		await Users.updateFromUpdater({ _id: userId }, updater);
+	}
+
+	// Update customFields of all Direct Messages' Rooms for userId
```
pierre-lehnen-rc, line 61: Here's a problem: if the updater is received by param, it'll only commit the changes to the user later, but the subscriptions are already being updated here regardless.

If there's any problem with the user update, the user and its subscriptions will be holding different information.

pierre-lehnen-rc (changes requested):
One comment and one change request.

pierre-lehnen-rc (approved):

PR #35140, opened 2025-02-06
fix: Notify user stream when enabling/disabling email2fa

yash-rajpal:

When a user enables/disables their email two factor, we were not sending this info on the streams, due to which the client side's `Meteor.user()` wasn't reactive to this change.

This PR propagates this change in user, making the client reactive to this change.
## Issue(s)

The email 2FA buttons on user's security page are not reactive.

## Steps to test or reproduce

- Go to Security page in user's account
- Enable/Disable user's email two factor auth
- The UI doesn't reflect the changes
## Further comments

[CORE-977](https://rocketchat.atlassian.net/browse/CORE-977)

[CORE-977]: https://rocketchat.atlassian.net/browse/CORE-977?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.changeset/new-ears-check.md
```diff
@@ -0,0 +1,5 @@
+---
+'@rocket.chat/meteor': minor
```
MarcosSpessatto, line 2:
```suggestion
'@rocket.chat/meteor': patch
```

MarcosSpessatto (changes requested):
What about adding a test to cover this?

dougfabris (approved):
Approving on behalf of frontend team

PR #35143, opened 2025-02-06
chore: remove `node-gcm` library and features due to deprecation

julio-rocketchat:
The `node-gcm` (https://www.npmjs.com/package/node-gcm) library has been deprecated since Google themselves deprecated their legacy GCM API. The API and the library are no longer working since June 2024 - which makes our `sendGCM` method no longer work.

We already had a replacement created by @Gustrb using the newer API wrapped in the `sendFCM` method. This PR simply removes the deprecated functionalities.

## Issue(s)
https://rocketchat.atlassian.net/browse/VLN-101

## Steps to test or reproduce
N/A

## Further comments
N/A


apps/meteor/server/settings/push.ts
```diff
@@ -109,6 +110,7 @@ export const createPushSettings = () =>
 			});
 			await this.add('Push_gcm_api_key', '', {
 				type: 'string',
+				hidden: true,
```
sampaiodiego, line 113: both this setting and `Push_gcm_project_number` are still being used at:
https://github.com/RocketChat/Rocket.Chat/blob/4915806200b1e89e9a24523fe623d1e9921e5ef4/apps/meteor/server/lib/pushConfig.ts#L99-L102

I suppose that part of the code should be refactored since this will be always empty from now on.

sampaiodiego (changes requested):

apps/meteor/server/settings/push.ts
```diff
@@ -21,6 +21,7 @@ export const createPushSettings = () =>
 
 		await this.add('Push_UseLegacy', false, {
 			type: 'boolean',
+			hidden: true,
```
sampaiodiego, line 24: can you also add a TODO or anything that reminds of removing this setting on the next major release?

julio-rocketchat: Added a TODO comment

apps/meteor/server/settings/push.ts
julio-rocketchat: @Gustrb seems like we're no longer using this part of the code since we don't use these settings. Should the `gcm` config be removed? Would like your insight on this one since you worked on the push notifications before

Gustrb: Hey, it looks like they are not being used and could be removed, but since we check for the key's existence to see if we should call fcm or apn, I think we could keep it here (maybe empty object?) and we can remove it later on a larger refactoring.
I think this code can be improved by a lot, but I don't think we should be doing it in this PR (since it is a chore and it is supposed to just remove the dependency)

julio-rocketchat: Added a TODO comment and removed `settings.get`, but feel free to suggest a different way of doing it. @sampaiodiego @Gustrb

Gustrb (dismissed):

sampaiodiego (dismissed):

Issue #33793, opened 2024-10-25
Username Highlight Not Working in Notification When User Language is Set to Russian


SinghaAnirban005:
### Description:
When a user’s language in Rocket.Chat is set to Russian, mentions (e.g., **@username**) do not properly highlight the username in the chat, as seen with Rocket.Cat bot notification. Specifically, the issue was observed when @anirban was mentioned, but the username was not highlighted in the message. This issue does not occur with other languages like English, where mentions properly highlight the username.

![Screenshot 2024-10-26 015208](https://github.com/user-attachments/assets/3518db21-12df-448e-8838-34b2c5e24b26)

### Steps to reproduce:

1. Open Rocket.Chat Desktop Client
2. Go to Preferences 
3. Select Russian as Language from dropdown --> Save Changes
4. Open any channel -> navigate to Members -> Click on Add
5. Choose existing members from the list and click on Add Users
6. Notice the rocket.cat message  that comes up

### Expected behavior:
The mentioned username should be highlighted regardless of the user’s language setting.

### Actual behavior:
In the Russian language setting, the username is not highlighted , but works fine in English.

### Server Setup Information:

- Version of Rocket.Chat Server:  7.1.0
- Operating System: Linux
- DB Replicaset Oplog: Yes
- NodeJS Version: 20.18.0
- MongoDB Version: 7.0.5

### Client Setup Information

- Desktop App or Browser Version: N/A
- Operating System: Windows

reetp:
Referred to team for review

PR #35144, opened 2025-02-06
fix: Username highlight issue in notifications when user language is set to Russian

KUSHAL-BANKHEDE:
**Issue**: The issue (#33793) describes a problem where usernames in notifications are not highlighted correctly when the user's language is set to Russian.

**Proposed Changes**: The PR likely modifies the translation string for the notification message to ensure proper username highlighting.

**Key Changes**
The main change in this PR is updating the translation string for the Russian language. Specifically:

The old string: "Username_is_already_in_here": "\"@%s\" уже здесь."

The new string: "Username_is_already_in_here": "\@%s` уже здесь."`

This change replaces double quotes (") with backticks (`) around the username placeholder (@%s) to ensure proper rendering and highlighting in notifications.

Why This Fix Works
The backticks (`) ensure that the username is rendered as a code block or highlighted text in notifications.

This prevents issues with escaping double quotes (") in the notification message, which can cause rendering problems.


Gustrb:
hey, I think we could also tacle a different issue in the translation, in russian they don't use quotes, they use: «Name» уже здесь

PR #35147, opened 2025-02-06
feat: New filters on `livechat/rooms`

KevLehman:

Relatively simple PR:
- Allows to filter by 1 or many `departmentId`s
- Allows to filter by 1 or many `units`

Tags & agents are already supporting multiple elements, so we just need UI for them.
## Issue(s)

https://rocketchat.atlassian.net/browse/CORE-942


apps/meteor/ee/app/livechat-enterprise/server/hooks/applyRoomRestrictions.ts
```diff
@@ -6,29 +6,42 @@ import { callbacks } from '../../../../../lib/callbacks';
...
+	if (!Array.isArray(userUnits)) {
+		if (Array.isArray(unitsFilter) && unitsFilter.length) {
+			return { ...query, departmentAncestors: { $in: unitsFilter } };
+		}
 		return query;
 	}
-	const departments = await LivechatDepartment.find({ ancestors: { $in: units } }, { projection: { _id: 1 } }).toArray();
+
+	if (Array.isArray(unitsFilter) && unitsFilter.length) {
+		const userUnit = new Set([...userUnits]);
```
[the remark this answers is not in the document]
KevLehman: The `if` above should remove the undefined. 
If `userUnits` is not an array (undef) it will return either the filter for the units or the original query.

On this point, we are certain the userUnits param is an array

MarcosSpessatto (approved):

MartinSchoeler (dismissed):

apps/meteor/ee/server/models/raw/LivechatUnit.ts
```diff
@@ -168,6 +168,11 @@ export class LivechatUnitRaw extends BaseRaw<IOmnichannelBusinessUnit> implement
 		return this.findOne(query, options);
 	}
 
+	async findUnitsByMonitorId(monitorId: string): Promise<FindCursor<IOmnichannelBusinessUnit>> {
+		const monitoredUnits = await LivechatUnitMonitors.findByMonitorId(monitorId).toArray();
+		return this.find({ _id: { $in: monitoredUnits.map((u) => u.unitId) } }, {});
```
pierre-lehnen-rc, line 173: Avoid referencing models from within other models - In this case its best to make a simple `findByIds` and pass the list of unitIds directly (so wherever you called `findUnitsByMonitorId`, you call both `findByMonitorId` to get the units and then `findByIds`)

packages/rest-typings/src/v1/omnichannel.ts
```diff
@@ -2821,11 +2821,13 @@ export type GETLivechatRoomsParams = PaginatedRequest<{
 	closedAt?: string;
 	agents?: string[];
 	roomName?: string;
-	departmentId?: string;
+	// TODO: support only arrays on next major
+	departmentId?: string | string[];
```
pierre-lehnen-rc, line 2825: maybe make a separate `departments` param that accepts an array, while keeping `departmentId` as a string and flagged as deprecated?

KevLehman: This is to reduce UI changes. UI will keep sending either a string or an array without knowing there's anything happening.

apps/meteor/ee/server/models/raw/LivechatUnit.ts
KevLehman: I'm gonna remove this one anyways since it's no longer needed (ther was a need, but after building the UI the need dissapeared)

But yeah, this is a copy/paste of the one right under it.

MarcosSpessatto (approved):

MarcosSpessatto (dismissed):

aleksandernsilva (approved):

MarcosSpessatto (approved):

PR #35155, opened 2025-02-08
None and default wasn't producing sound

Ansh-dS:
 Since both the option wasn't producing sound so I removed "none" to maintain the consistency with other dropdown menus. 

*Before*
image

  

*After*
Screenshot 2025-02-08 132820

## Steps to test or reproduce
1. choose any channel. 
2. click on three dots
3. navigate to "Notification preference". 
4. click on desktop and then sound.


julio-rocketchat:
I could be wrong, but the idea of "None" may be to not have any sound

PR #35165, opened 2025-02-10
refactor: validate required custom fields on visitor creation

ricardogarim:
As per [CORE-214](https://rocketchat.atlassian.net/browse/CORE-214), this PR introduces validation for custom fields in the `POST /api/v1/livechat/visitor` route. **The validation is only triggered if custom fields are provided in the request payload.** If no custom fields are present, the current behavior remains unchanged.

This approach avoids a breaking change since altering the default behavior could disrupt existing implementations, which is only acceptable for a major release or for security reasons.

## Issue(s)
- [CORE-214](https://rocketchat.atlassian.net/browse/CORE-214)

## Steps to test or reproduce
- Create a custom field on livechat management screen with `visitor` visibility and mark it as required (need EE license)
- Opens `localhost:3000/livechat` and see it's showing
- Make a POST request to visitor creation endpoint passing the `customFields` property as shown below

```json
{
	"visitor": {
		"email": "valid@email.com",
		"name": "New Visitor",
		"token": "valid token",
		"customFields": [
			{
				"key": "foo",
				"value": "bar",
				"overwrite": true
			}
		]
	}
}
```

## Further comments
- **Validation scope:** Custom fields are validated only if they are provided. If absent, the behavior remains unchanged.
- **Compatibility:** Default Livechat widgets and other official methods for visitor creation typically do not send custom fields directly via this route. Instead, they use the `POST /api/v1/livechat/custom.field` route to add custom fields after creating the visitor.
- **Future direction:** This change lays the groundwork for requiring custom field validation during visitor creation in a future major release. Once all services are updated to support custom fields during visitor creation, we can make custom fields mandatory and adjust the use of `POST /api/v1/livechat/custom.field` to handle specific use cases rather than all custom field sets.

[CORE-214]: https://rocketchat.atlassian.net/browse/CORE-214?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[CORE-214]: https://rocketchat.atlassian.net/browse/CORE-214?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/meteor/app/livechat/server/api/v1/visitor.ts
```diff
@@ -9,6 +9,23 @@ import { settings } from '../../../../settings/server';
 import { Livechat as LivechatTyped } from '../../lib/LivechatTyped';
 import { findGuest, normalizeHttpHeaderData } from '../lib/livechat';
 
+const validateRequiredCustomFields = async (customFields: string[], livechatCustomFields: ILivechatCustomField[]) => {
```
MarcosSpessatto, line 12: Can we create unit-tests for this function?

MarcosSpessatto (changes requested):
This looks more like a fix rather than a refactor, wdyt?

apps/meteor/app/livechat/server/api/v1/visitor.ts
```diff
@@ -79,28 +80,32 @@ API.v1.addRoute(
 			);
 
 			if (customFields && Array.isArray(customFields) && customFields.length > 0) {
-				const keys = customFields.map((field) => field.key);
 				const errors: string[] = [];
+				const keys = customFields.map((field) => field.key);
+
+				const livechatCustomFields = await LivechatCustomField.findByScope(
```
KevLehman, line 86: Can u add an api test to ensure this won't break anything? one that is: if no custom fields are received, then the required validation shouldn't apply

apps/meteor/app/livechat/server/lib/validateRequiredCustomFields.ts
```diff
@@ -0,0 +1,16 @@
+import type { ILivechatCustomField } from '@rocket.chat/core-typings';
+
+export const validateRequiredCustomFields = async (customFields: string[], livechatCustomFields: ILivechatCustomField[]) => {
```
KevLehman, line 3: This doesn't need to be async

KevLehman (changes requested):

MarcosSpessatto (approved):

KevLehman (approved):

PR #35173, opened 2025-02-10
fix: unable to identify message type via apps engine

Gustrb:

Currently, it’s not possible to identity the type of messages intercepted via IPostMessageSent event. Such an information is not available on the IMessage interface.
In this PR, we now have access to the `t` field (the type of the message). And we also start triggering the event handlers even on system messages.

## Further comments

[CONN-482](https://rocketchat.atlassian.net/browse/CONN-482)

[CONN-482]: https://rocketchat.atlassian.net/browse/CONN-482?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Gustrb:
Closed in favor of #35177

PR #35175, opened 2025-02-11
chore: replace the kadira flow router with the ostrio flow router

julio-rocketchat:
According to https://forums.meteor.com/t/steps-needed-to-switch-from-kadira-flow-router-to-ostrio-flow-router-extra/59355 and https://guide.meteor.com/routing.html#flow-router, the Ostrio package seems to be the replacement of the old Kadira Flow Router. This PR replaces Kadira with the newer Ostrio Flow Router.

## Issue(s)
Fixes https://rocketchat.atlassian.net/browse/VLN-99

## Steps to test or reproduce
N/A

## Further comments
N/A


apps/meteor/client/main.ts
```diff
@@ -1,13 +1,22 @@
...
+});
+
+Meteor.startup(() => {
+	FlowRouter.initialize({
+		hashbang: false,
+		page: {
+			click: true,
+		},
+	});
+});
```
gabriellsh, line 19: I think we can transform this into a hook and place it in `AppLayout`. We have a project with the intent of removing `Meteor.startup()` so we should avoid adding new ones.

apps/meteor/client/providers/RouterProvider.tsx
```diff
@@ -147,26 +136,15 @@ const defineRoutes = (routes: RouteObject[]) => {
...
-			FlowRouter._routes = FlowRouter._routes.filter((r) => r !== flowRoute);
-			if ('name' in flowRoute && flowRoute.name) {
-				delete FlowRouter._routesMap[flowRoute.name];
-			} else {
-				FlowRouter.notFound = {
-					action: () => appLayout.render(<></>),
-				};
-			}
-		});
-
```
gabriellsh, line 164: This callback is used to unregister routes whenever needed. I don't think we can just remove this part of the callback, we probably need an alternative.

Btw `apps/meteor/client/startup/audit.tsx` is one of the places where this callback is used. Not sure if there are others though.

gabriellsh (changes requested):

apps/meteor/client/main.ts
julio-rocketchat: Would you be able to help us do these changes, @gabriellsh? I'm afraid I may end up messing it up

apps/meteor/client/providers/RouterProvider.tsx
julio-rocketchat: Would you be able to help us with this one as well, @gabriellsh? Not sure the best way we could do it here

gabriellsh (approved):

PR #35177, opened 2025-02-11
feat(apps): add IPostSystemMessageSent event

Gustrb:

Currently, there is no way to run an app when a given system message is sent, since we only trigger the IPostMessageSent for normal messages.
We could have followed the approach of having system messages also trigger the event, but that would create a pain point, since we would not be able to add the methods to check (to block messages).
This Pull Request, introduces the  new event that is triggered on system messages, so the apps can respond to it

## Further comments

[CONN-482](https://rocketchat.atlassian.net/browse/CONN-482)

[CONN-482]: https://rocketchat.atlassian.net/browse/CONN-482?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/meteor/app/apps/server/converters/messages.js
```diff
@@ -243,6 +244,14 @@ export class AppMessagesConverter {
 		);
 	}
 
+	async convertSystemMessage(message) {
```
d-gubert, line 247: I would say that we shouldn't need a new method for system messages, considering that the values don't seem to vary that much. What was your thought on adding a new method?

apps/meteor/app/apps/server/converters/messages.js
```diff
@@ -243,6 +244,14 @@ export class AppMessagesConverter {
...
+		if (!message) {
+			return undefined;
+		}
+
+		return this.convertMessageRaw(message, { type: 't' });
```
d-gubert, line 252: Preferably use the `convertMessage` method. This one is kind of reserved for "bulk" reading of messages, because it doesn't go to the database to fill the relationships with the full data.

apps/meteor/app/apps/server/converters/messages.js
```diff
@@ -48,6 +48,7 @@ export class AppMessagesConverter {
 			attachments: getAttachments,
 			sender: 'u',
 			threadMsgCount: 'tcount',
+			...extraFields,
```
d-gubert, line 51: I think we don't need to be configurable here, you can add the `type: 't'` property directly, not all messages have it anyway.

Also be sure to include the same property to the map in the `convertMessage` method

packages/apps-engine/src/definition/messages/ISystemMessage.ts
```diff
@@ -0,0 +1,79 @@
+import type { IMessage } from '.';
+
+export type SystemMessageType =
```
d-gubert, line 3: It would be very cool if we could add some comments explaining each of the types of system message. But I'm not sure we have this information available, right? Maybe where you sourced these types there is some explanation?

packages/apps-engine/src/definition/messages/ISystemMessage.ts
```diff
@@ -0,0 +1,79 @@
...
+    | 'message_pinned_e2e'
+    | 'new-moderator'
+    | 'moderator-removed'
+    | 'new-owner'
+    | 'owner-removed'
+    | 'new-leader'
+    | 'leader-removed'
+    | 'discussion-created';
+
+export interface ISystemMessage extends IMessage {
```
d-gubert, line 77: Should we go with a different type altogether? Given the difference is only one field, I can't say I see the benefits of creating a new one instead of adding it to the existing `IMessage`. Curious as to what was your reasoning here :)

apps/meteor/app/apps/server/converters/messages.js
d-gubert: An interesting "side-effect" of using the same converter method is that we can use the same  `messageEvent` method from the listener class

packages/apps-engine/src/definition/messages/ISystemMessage.ts
Gustrb: Unfortunately there is none, but I'll add a few comments explaining some of those, I guess it would be better than nothing

packages/apps-engine/src/definition/messages/ISystemMessage.ts
Gustrb: Yea, I was thinking about having the system message be a separate thing from IMessage because I thought the data structure would be really different, but in the end, the only difference is the type field, I'll change this

apps/meteor/tests/unit/app/apps/server/messages.tests.js
```diff
@@ -101,12 +101,6 @@ describe('The AppMessagesConverter instance', () => {
 			});
 		});
 
-		it('should add an `_unmappedProperties_` field to the converted message which contains the `t` property of the message', async () => {
-			const appMessage = await messagesConverter.convertMessage(messagesMock.findOneById('SimpleMessageMock'));
-
-			expect(appMessage).to.have.property('_unmappedProperties_').which.has.property('t', 'uj');
```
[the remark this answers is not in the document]
Gustrb: I mean it kind of is, but since we don't trigger events on system messages it would be undefined all the times. So I don't believe anyone would be using

MarcosSpessatto (approved):

apps/meteor/tests/unit/app/apps/server/messages.tests.js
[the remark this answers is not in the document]
d-gubert: Yeah, just to second what @Gustrb said, the `_unmappedProperties_` is technically not part of the public API, so no one should _really_ be counting on it for stuff to work. If they do, it's an unsupported property, so we don't really need to worry about taking fields out of it

PR #35194, opened 2025-02-12
fix: Users without preview permission possibility of subscribe to public channel stream

tiagoevanp:

Remove the possibility of listening to `room-messages` stream of public channels without being subscribed into it.

related to #34922 

## Further comments

[CORE-950]

[CORE-950]: https://rocketchat.atlassian.net/browse/CORE-950?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/meteor/server/modules/notifications/notifications.module.ts
```diff
@@ -105,7 +105,8 @@ export class NotificationsModule {
 			if (!canAccess) {
 				// verify if can preview messages from public channels
 				if (room.t === 'c' && this.userId) {
-					return Authorization.hasPermission(this.userId, 'preview-c-room');
+					const subscription = await Subscriptions.findOne({ rid: room._id });
```
pierre-lehnen-rc, line 108: The `preview-c-room` is used specifically to grant access to users that are not yet subscribed. The condition here should be `subscribed || hasPermission`;

Also, if the user is subscribed to the room, `canAccess` will be true and this code won't even be executed.

pierre-lehnen-rc (changes requested):

apps/meteor/server/modules/notifications/notifications.module.ts
pierre-lehnen-rc: `canAccess` will also be true for logged-out users when Anonymous Read is enabled.

apps/meteor/server/modules/notifications/notifications.module.ts
```diff
@@ -110,6 +111,16 @@ export class NotificationsModule {
...
 			}
 
+			if (room.t === 'c' && this.userId) {
+				if (await Authorization.hasPermission(this.userId, 'preview-c-room')) {
+					const subscription = await Subscriptions.findOneByRoomIdAndUserId(room._id, this.userId);
+
+					return !!subscription;
+				}
+
+				return false;
```
pierre-lehnen-rc, line 121:
```suggestion
				if (!(await Authorization.hasPermission(this.userId, 'preview-c-room'))) {
					const subscription = await Subscriptions.findOneByRoomIdAndUserId(room._id, this.userId);

					return !!subscription;
				}
```

Permission OR subscription.

pierre-lehnen-rc (changes requested):

apps/meteor/server/modules/notifications/notifications.module.ts
```diff
@@ -102,15 +102,12 @@ export class NotificationsModule {
 			}
 
 			const canAccess = await Authorization.canAccessRoom(room, { _id: this.userId || '' }, extraData);
```
pierre-lehnen-rc, line 104: You can remove this call to canAccessRoom since it'll be called inside of the canReadRoom function. Just make sure to pass the extraData to it as well.

pierre-lehnen-rc (changes requested):

pierre-lehnen-rc (approved):

pierre-lehnen-rc (approved):

PR #35199, opened 2025-02-13
feat(apps): enable configuring which hidden settings are available for the apps

Gustrb:

Enhances the server-setting.read permission, allowing it to take an array of hiddenPermissions, that are the permissions that can be searched despite being hidden (or secret). This enables us to have a Omnigateway_Url setting. 

This setting is going to be responsible to have the omnigateway of the current env. By having this setting, we can easily stage our omni-apps, because, instead of having the value of the omnigateway url hardcoded, it will be just a matter of fetching the value of the setting through the apps-engine. Thus, centralizing the information

## Further comments

[CONN-525](https://rocketchat.atlassian.net/browse/CONN-525)

[CONN-525]: https://rocketchat.atlassian.net/browse/CONN-525?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/meteor/server/settings/setup-wizard.ts
```diff
@@ -1223,6 +1223,13 @@ export const createSetupWSettings = () =>
 				secret: true,
 			});
 
+			await this.add('Omnigateway_Url', 'https://omni-gateway.rocket.chat', {
```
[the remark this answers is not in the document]
Gustrb: Yea, for some reason it overrides it regardless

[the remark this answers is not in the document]
d-gubert: Ok, cool

KevLehman, line 1226: Do we need this setting? I know apps use omnigateway but we actually never allowed customers to view/override it. This would allow that.

Gustrb: Hey, this setting is going to be really useful when we add a new staging omnigateway env. So the apps know to which env they should point out.
I don't think apps should be able to override it, only to read so we can centralize this piece of information in one place, wdyt?

Gustrb: This is a new initiative to stage omnigateway btw

Gustrb: oh, btw, since this is a hidden setting the app creator will have to specify which permissions does it require, and he needs to manually select each "hidden" they will need access to. So we don't give complete power to app's developer to mess with our internal settings

apps/meteor/app/apps/server/bridges/settings.ts
```diff
@@ -50,6 +52,39 @@ export class AppSettingBridge extends ServerSettingBridge {
 		return Boolean(setting && !setting.secret);
 	}
 
+	protected async getReadableSettingById(id: string, appId: string): Promise<ISetting | null> {
+		this.orch.debugLog(`The app ${appId} is checking if they can read the setting ${id}`);
```
KevLehman, line 56:
```suggestion
		this.orch.debugLog(`The app ${appId} is checking if it can read the setting ${id}`);
```

app's a thing

apps/meteor/app/apps/server/bridges/settings.ts
```diff
@@ -50,6 +52,39 @@ export class AppSettingBridge extends ServerSettingBridge {
...
+		const readSettingsPermission = permissions.find((perm) => perm.name === 'server-setting.read');
+		if (!readSettingsPermission) {
+			this.orch.debugLog(`The app ${appId} has no server-setting.read permission.`);
+			return null;
+		}
+
+		const readSettings = readSettingsPermission as IReadSettingPermission;
+		// If the setting is in the hiddenSettings list (defined within the permission), then it can bypass the hidden flag.
+		// If not, then it must be a non-hidden setting. This is to allow apps to read hidden settings if they have the permission to do so.
+		const setting = readSettings.hiddenSettings?.includes(id) ? await Settings.findOneById(id) : await Settings.findOneNotHiddenById(id);
```
KevLehman, line 78: Should we log somehow an app attempted to read a setting it doesnt have access to instead of the setting "not being found"?

apps/meteor/app/apps/server/bridges/settings.ts
Gustrb: Oh yea, good catch

apps/meteor/app/apps/server/bridges/settings.ts
Gustrb: I think we could log via orch.debugLog (not sure if that valuable though), but I don't think we should return this to the end user.

I think we have enough tracing to troubleshoot most issues that might arise, wdyt?

d-gubert: I'm not sure... what would we do with this info?

Gustrb: I can see it being useful as an app developer who forgot to add that specific setting to the permisson list, but I mean, it should be clear from the logs that no setting was found (also, not sure if this should be a widely documented feature of apps engine)

KevLehman (approved):

Issue #35200, opened 2025-02-13
fix: : Keeping emojiPicker open even after an emoji selection


Ansh-dS:
## Description
When an emoji is selected, the emoji picker immediately closes. This behavior forces users to reopen the picker if they want to select more than one emoji or if they accidentally choose the wrong emoji, resulting in a poor user experience.

## Steps to Reproduce
1. In any group, channel, discussion, or team chat, click the emoji icon (the smiley) located beneath the message bar in the message toolbar.
2. The emoji picker will open.
3. Click on any emoji. Notice that the picker closes immediately.

## Expected Behavior
The emoji picker should remain open until the user performs one of the following actions:
- Clicks on the message bar,
- Clicks the send button,
- Clicks outside of the emoji picker (i.e., the focus leaves the popup).

## Actual Behavior

https://github.com/user-attachments/assets/5eb37753-c7a1-4f6f-8f88-ec6749cb29c1

## Server/Client Information

**Server:**
- **Operating System:** Windows 11 (using WSL Terminal)
- **Rocket.Chat Server Version:** 7.4.0‑develop
- **NodeJS Version:** 22.11.0
- **Yarn Version:** 4.6.0

**Client:**
- **Client Type:** Browser (Microsoft Edge)
- **Access URL:** [http://localhost:3000](http://localhost:3000)

**Additional:**
- **WSL Terminal:** Used to install dependencies

dougfabris:
Hi there, thanks for the contribution! 🚀 💯 

Sorry, this is not an issue, it was planned to be like that by our design team.

---

**Questions? Help needed? Feature Requests?** 

- Join our [Open Server](https://open.rocket.chat/) in the #support channel and feel free to raise a question
- Join our [Community Forum](https://forums.rocket.chat/) and search/create a post there

Ansh-dS:
I have asked question in the support channel about this.

reetp:
Doug, theres's an inconsistency in design here.

Picker remains open in my react app. That seems typical on mobile and expected behaviour.

Not checkded Electron but guess that's the same as a browser.

Personally I'd say this would be an improvement?

![Screenshot_20250214_102101_RocketChat Experimental.jpg](https://github.com/user-attachments/assets/81f4ce19-64db-4412-86e5-6d8b56f95974)

dougfabris:
[...]

PR #35201, opened 2025-02-13
fix: : Keeping emojiPicker open even after an emoji selection 

Ansh-dS:
The emoji picker should remain open until the user performs one of the following actions:
- Clicks on the message bar,
- Clicks the send button,
- Clicks outside of the emoji picker (i.e., the focus leaves the popup).

https://github.com/user-attachments/assets/e4ffe678-87e1-4713-a47c-e4b152c0f700

## Issue(s)
 #35200


ggazzo:
i'm 99.9% sure this was intentional

PR #35203, opened 2025-02-13
chore: Use transaction for saveUser function

gabriellsh:

Currently the `saveUser` function has a lot of side-effects and interacts with many collections multiple times. This can cause sync issues between records because references can change in one place, and fail to change in another, shortcutting the method execution without a proper rollback. For example, if you change a user's username, and the method that updates the avatar fails, the username will have been changed but an error will be thrown in the `users.update` api, causing confusion since part of the record change but the rest of the changes might not have happened.

This issue was specially prominent during the development of the user auditing mechanism, because it can cause discrepancies between what is being logged and what actually hapenned. It would require to watch the operation step by step or compare records before and after the operation, which also causes discrepancies, as the record can have changed in the mean time that this operation occurs.

This PR aims to make the `saveUser` helper function more solid with the use of transactions. The use of a transaction allows side effects to be skipped in case it fails, and automatically rolls back all operations that are executed within the transaction. It also will throw errors if by any chance there is a concurrent attempt to change the same record, avoiding unintended conflicts due to race conditions.

## Issue(s)

[CORE-1022](https://rocketchat.atlassian.net/browse/CORE-1022)
## Further comments

[CORE-1022]: https://rocketchat.atlassian.net/browse/CORE-1022?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/meteor/app/lib/server/functions/saveUser/setPasswordUpdater.ts
```diff
@@ -0,0 +1,25 @@
...
+};
+
+export async function setPasswordUpdater(
+	updater: Updater<IUser>,
+	newPlaintextPassword: string,
+	options: { logout?: boolean } = { logout: true },
+) {
+	updater.set('services.password.bcrypt', await hashPassword(newPlaintextPassword));
+
+	if (options?.logout) {
```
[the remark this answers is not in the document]
gabriellsh: This options object is actually not necessary for my use case, but it mimics meteor's Accounts.setPasswordAsync method signature, in case we decide to use in other places.

I'm aware that there's no reason to implement logic that is not used just for the sake of reusability, but, in this case, given that it is necessary to study meteor's repository in order to understand the side effects the method has, I thought it was a good idea to keep this to avoid confusion.

Also, I think you mean for me to just remove the optional property chaining, correct?
`if (options.logout)` since I didn't destructure the property.

apps/meteor/app/lib/server/functions/saveUser/saveUser.ts
```diff
@@ -44,135 +47,148 @@ export type SaveUserData = {
...
+		if (userData.customFields) {
+			await saveCustomFields(userData._id, userData.customFields, { _updater: updater, session });
+		}
+
+		await Users.updateFromUpdater({ _id: userData._id }, updater, { session });
+
+		// App IPostUserUpdated event hook
+		// We need to pass the session here to ensure this record is fetched
+		// with the uncommited transaction data.
+		const userUpdated = await Users.findOneById(userData._id, { session });
```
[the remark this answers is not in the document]
gabriellsh: I guess so. But isn't it going to be still just a find after the update?

apps/meteor/server/database/utils.ts
```diff
@@ -18,3 +18,23 @@ export const { db, client } = MongoInternals.defaultRemoteCollectionDriver().mon
...
+		} catch (error) {
+			await session.abortTransaction();
+
+			throw error;
+		} finally {
+			await session.endSession();
+		}
+
+		return result;
+	};
```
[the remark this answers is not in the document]
gabriellsh: I can move the return clause inside the try block, but I need the session outside so I can have access to it inside the catch and finally.

packages/models/src/models/VideoConference.ts
```diff
@@ -233,7 +233,12 @@ export class VideoConferenceRaw extends BaseRaw<VideoConference> implements IVid
 		// TypeScript is not smart enough to infer that `messages.${'start' | 'end'}` matches two keys of `VideoConference`
 	}
 
-	public async updateUserReferences(userId: IUser['_id'], username: IUser['username'], name: IUser['name']): Promise<void> {
+	public async updateUserReferences(
```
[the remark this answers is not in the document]
gabriellsh: 🤷‍♂️ 

I noticed that too :P

apps/meteor/app/lib/server/functions/saveUser/saveUser.ts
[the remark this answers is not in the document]
gabriellsh: As per discussion in DM we'll implement this another time

apps/meteor/app/lib/server/functions/saveCustomFieldsWithoutValidation.ts
```diff
@@ -48,10 +60,12 @@ export const saveCustomFieldsWithoutValidation = async function (userId: string,
 		}
 	});
 
-	await Users.updateFromUpdater({ _id: userId }, updater);
+	if (!_updater) {
```
MarcosSpessatto, line 63: Since you are providing `updater` to the model call, is there a necessity to do this check?

apps/meteor/app/lib/server/functions/setUsername.ts
```diff
@@ -67,78 +68,81 @@ export const setUsernameWithValidation = async (userId: string, username: string
 	void notifyOnUserChange({ clientAction: 'updated', id: user._id, diff: { username } });
 };
 
-export const _setUsername = async function (userId: string, u: string, fullUser: IUser, updater?: Updater<IUser>): Promise<unknown> {
-	const username = u.trim();
+export const _setUsernameWithSession = (session?: ClientSession) =>
```
MarcosSpessatto, line 71: I'm not sure about suffix this with `withSession` since the session is not required.

MarcosSpessatto (changes requested):

apps/meteor/app/lib/server/functions/saveCustomFieldsWithoutValidation.ts
```diff
@@ -12,7 +15,14 @@ const getCustomFieldsMeta = function (customFieldsMeta: string) {
...
 	}
 };
-export const saveCustomFieldsWithoutValidation = async function (userId: string, formData: Record<string, any>): Promise<void> {
+export const saveCustomFieldsWithoutValidation = async function (
+	userId: string,
+	formData: Record<string, any>,
+	options?: {
+		_updater?: Updater<IUser>;
+		session?: ClientSession;
+	},
```
KevLehman, line 24: To me, `options` is T | undefined, but if `options` is provided, the keys inside it shouldn't be undefined 🤔 

That would remove the need for the `if (!updater)` inside the func

apps/meteor/app/lib/server/functions/saveUser/saveUser.ts
```diff
@@ -44,135 +47,148 @@ export type SaveUserData = {
...
 export const isUpdateUserData = (params: SaveUserData): params is UpdateUserData => '_id' in params && !!params._id;
 
-export const saveUser = async function (userId: IUser['_id'], userData: SaveUserData) {
-	const oldUserData = userData._id && (await Users.findOneById(userData._id));
-	if (oldUserData && isUserFederated(oldUserData)) {
-		throw new Meteor.Error('Edit_Federated_User_Not_Allowed', 'Not possible to edit a federated user');
-	}
+const _saveUser = (session?: ClientSession) =>
+	async function (userId: IUser['_id'], userData: SaveUserData) {
+		const oldUserData = userData._id && (await Users.findOneById(userData._id));
```
KevLehman, line 58: This one should receive the session as well.

apps/meteor/app/lib/server/functions/saveUser/saveUser.ts
```diff
@@ -44,135 +47,148 @@ export type SaveUserData = {
...
+			delete userData.verified;
+		}
+		void notifyOnUserChange({
+			clientAction: 'updated',
+			id: userData._id,
+			diff: {
+				...userData,
+				emails: userUpdated?.emails,
+			},
+		});
```
KevLehman, line 189: I wonder if we could move these to happen after the transaction has commited.
For example: if the transaction aborts outside of this, we'll call the sendPasswordEmail again when the transaction happens again.

The tx should contain the minimal operations required to keep state, but not tell others about that state until after commit.

apps/meteor/app/lib/server/functions/saveUserIdentity.ts
```diff
@@ -54,7 +57,7 @@ export async function saveUserIdentity({
 			return false;
 		}
 
-		if (!(await _setUsername(_id, username, user, updater))) {
+		if (!(await _setUsernameWithSession(session)(_id, username, user, updater))) {
```
KevLehman, line 60: nitpick: the mix of ways of passing the transaction feels a bit odd. 
(feel free to ignore, it's more of a personal opinion)

apps/meteor/app/lib/server/functions/saveUserIdentity.ts
```diff
@@ -115,33 +119,36 @@ async function updateUsernameReferences({
...
+		await Messages.updateUsernameOfEditByUserId(user._id, username, { session });
 
 		const cursor = Messages.findByMention(previousUsername);
 		for await (const msg of cursor) {
 			const updatedMsg = msg.msg.replace(new RegExp(`@${previousUsername}`, 'ig'), `@${username}`);
-			await Messages.updateUsernameAndMessageOfMentionByIdAndOldUsername(msg._id, previousUsername, username, updatedMsg);
+			await Messages.updateUsernameAndMessageOfMentionByIdAndOldUsername(msg._id, previousUsername, username, updatedMsg, { session });
 		}
 
+		// Mongo does not recommend paralelizing session transactions
```
KevLehman, line 144: Can you put a link to the docs about this? Would be a nice read.

apps/meteor/app/lib/server/functions/saveUserIdentity.ts
```diff
@@ -115,33 +119,36 @@ async function updateUsernameReferences({
...
+		// This is working fine and should be as long as the first transaction is no parallel
 		const responses = await Promise.all([
-			Rooms.replaceUsername(previousUsername, username),
-			Rooms.replaceMutedUsername(previousUsername, username),
-			Rooms.replaceUsernameOfUserByUserId(user._id, username),
-			Subscriptions.setUserUsernameByUserId(user._id, username),
-			LivechatDepartmentAgents.replaceUsernameOfAgentByUserId(user._id, username),
+			Rooms.replaceUsername(previousUsername, username, { session }),
+			Rooms.replaceMutedUsername(previousUsername, username, { session }),
+			Rooms.replaceUsernameOfUserByUserId(user._id, username, { session }),
```
KevLehman, line 149: is it possible to have these 3 as one single operation? that way the tx reduces load

apps/meteor/app/lib/server/functions/setUsername.ts
```diff
@@ -67,78 +68,81 @@ export const setUsernameWithValidation = async (userId: string, username: string
...
 		}
-	}
-	// If first time setting username, send Enrollment Email
-	try {
-		if (!previousUsername && user.emails && user.emails.length > 0 && settings.get('Accounts_Enrollment_Email')) {
-			setImmediate(() => {
-				Accounts.sendEnrollmentEmail(user._id);
-			});
+
+		const user = fullUser || (await Users.findOneById(userId));
```
KevLehman, line 83: pass session

apps/meteor/app/lib/server/functions/setUsername.ts
```diff
@@ -67,78 +68,81 @@ export const setUsernameWithValidation = async (userId: string, username: string
...
-	void api.broadcast('user.nameChanged', {
-		_id: user._id,
-		name: user.name,
-		username: user.username,
-	});
+		void api.broadcast('user.nameChanged', {
+			_id: user._id,
+			name: user.name,
+			username: user.username,
+		});
```
KevLehman, line 140: This should happen after transaction commits,otherwise client may get invalid updates.

apps/meteor/server/database/utils.ts
```diff
@@ -18,3 +18,21 @@ export const { db, client } = MongoInternals.defaultRemoteCollectionDriver().mon
...
+export const wrapInSessionTransaction =
+	<T extends Array<unknown>, U>(curriedCallback: (session: ClientSession) => (...args: T) => U) =>
+	async (...args: T): Promise<Awaited<U>> => {
+		const session = client.startSession();
+		try {
+			session.startTransaction();
+			const result = await curriedCallback(session).apply(this, args);
+			await session.commitTransaction();
+			return result;
+		} catch (error) {
```
KevLehman, line 31: Mongo recommends transactions to be retried. Transient transaction errors could cause a rollback even when the tx is valid (a conflict with another op or something).

Also remember that, the longer the transaction spans, the easier it's for it to conflict with something.

KevLehman (changes requested):

apps/meteor/app/lib/server/functions/saveCustomFieldsWithoutValidation.ts
gabriellsh: This is a standalone method that is called in other places (users.create for example). This check avoids the update if the updater comes from outside the function, because in that case the caller is responsible for commiting the updates

apps/meteor/app/lib/server/functions/setUsername.ts
gabriellsh: I did currying here to avoid adding extra parameters to the function itself. I exported `_setUsername = _setUsernameWithSession()` below so that old references to this function still work as they did.

The parameter is indeed not required, but you should only call this method if you plan on using the session. Otherwise the "old" function is still exported under the same name it had before.

apps/meteor/app/lib/server/functions/saveCustomFieldsWithoutValidation.ts
gabriellsh: `if (!updater)` is used in order to "remove" the responsibility of the function to commit the updater. If you pass _updater as option, you have to commit the updater manually. Otherwise the function works as it did before.

Both parameters are independent and optional. You can pass just the updater, or just the session, and they should work despite the other being present or not.

apps/meteor/app/lib/server/functions/saveUser/saveUser.ts
gabriellsh: I was basing off of the mongodb docs. I read somewhere (could have been in another place) that if you know the document hasn't changed yet it's preferable to not pass the session. Since this is the first fetch of the user and happens before everything else, I though it was safe to not use the session here.

I can pass it anyway, don't think it'll do any harm.

apps/meteor/app/lib/server/functions/saveUser/saveUser.ts
gabriellsh: I agree, but this would require much more effort. It's a concern in other functions aswell. This function has a very deep call stack, and a bunch of those methods have side effects that are not so easy to externalize. I'll think about this some more

apps/meteor/app/lib/server/functions/saveUserIdentity.ts
gabriellsh: So, this is also one of the spots I used currying to avoid adding new parameters to the functions. Admitedly it's not the prettiest thing, but I did this mainly to functions that have other references to them, in an effort to minimize possible side effects or signature changes.

apps/meteor/app/lib/server/functions/saveUserIdentity.ts
gabriellsh: It does make sense, but I think this would be effort for another PR. WDYT?

apps/meteor/app/lib/server/functions/setUsername.ts
gabriellsh: This is another place where the user collection shouldn't have been written to yet. I can pass the session here too.

apps/meteor/server/database/utils.ts
gabriellsh: So I commented this with @sampaiodiego, and I actually had implemented retry calls for it at some point. But as you've seen there's so many side effects thrown around in inner functions, that I thought this would do more harm than good.

apps/meteor/app/lib/server/functions/setUsername.ts
gabriellsh: Yep. One of the places with side effects that are not so easy to extract.

apps/meteor/app/lib/server/functions/saveUser/saveUser.ts
Gustrb: Couldn't this lead to issues in the microsservices mode? I believe that since this is not inside the transaction it will not guarantee this record to be consistent with the other ones.

apps/meteor/server/database/utils.ts
gabriellsh: So I fixed the issue with sideEffects, but we decided to not retry this transaction and show an error asking to try again. If another operation happens to the user while this method runs for example, it should probably have been aborted anyways

MarcosSpessatto (approved):

Gustrb (approved):
Awesome work my dude 🚀

PR #35208, opened 2025-02-14
feat: display leader role group in rooms' members list

abhinavkrin:

This PR adds the Leader role to the members list (contextual bar) to ensure better visibility and role consistency. Previously, the Leader role was displayed only in the room header, which created inconsistencies in role presentation.  
Screenshot 2025-02-14 at 8 17 22 PM

## Further comments

[CORE-975](https://rocketchat.atlassian.net/browse/CORE-975)

[CORE-975]: https://rocketchat.atlassian.net/browse/CORE-975?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/core-typings/src/IRoom.ts
```diff
@@ -94,7 +94,7 @@ export interface IRoom extends IRocketChatRecord {
 		items: [SidepanelItem, SidepanelItem?];
 	};
 
-	rolePrioritiesCreated?: boolean;
+	rolePrioritiesCreated?: number | boolean;
```
MarcosSpessatto, line 97: I think we can mark boolean as deprecated since we will not add those anymore, also we should create a task for removing the boolean in the next major together with a migration.

packages/models/src/models/Rooms.ts
```diff
@@ -2207,8 +2207,8 @@ export class RoomsRaw extends BaseRaw<IRoom> implements IRoomsModel {
 		);
 	}
 
-	markRolePrioritesCreatedForRoom(rid: IRoom['_id']) {
-		return this.updateOne({ _id: rid }, { $set: { rolePrioritiesCreated: true } });
+	markRolePrioritesCreatedForRoom(rid: IRoom['_id'], version: number): Promise<UpdateResult> {
+		return this.updateOne({ _id: rid }, { $set: { rolePrioritiesCreated: version } });
```
MarcosSpessatto, line 2211: Shouldn't we need to change the method to check if it `hasCreatedRolePrioritiesForRoom`?

MarcosSpessatto (changes requested):

MarcosSpessatto (approved):

aleksandernsilva (approved):

abhinavkrin:
Merged #35209 which removes the LeaderBar from the chat room header.

MarcosSpessatto (approved):

MartinSchoeler (approved):

PR #35218, opened 2025-02-14
feat: Security Logs Page

MartinSchoeler:
Requires https://github.com/RocketChat/Rocket.Chat/pull/35258
Requires https://github.com/RocketChat/Rocket.Chat/pull/35361

Screenshot 2025-02-14 at 15 40 16
Screenshot 2025-02-14 at 15 41 05
Screenshot 2025-02-14 at 15 41 20
Screenshot 2025-02-14 at 17 25 09

## Issue(s)

[LOG-11](https://rocketchat.atlassian.net/browse/LOG-11)

## Further comments

[LOG-11]: https://rocketchat.atlassian.net/browse/LOG-11?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/meteor/client/views/audit/components/SecurityLogsTable.tsx
```diff
@@ -0,0 +1,224 @@
...
+				</Field>
+				<Field width={300}>
+					<Margins inline={6}>
+						<FieldLabel>{t('Setting')}</FieldLabel>
+						<SettingSelect value={setting} onChange={setSetting} />
+					</Margins>
+				</Field>
+				<ButtonGroup>
+					<Margins inline={6}>
+						{/* <Button secondary onClick={handleExportJson}>
```
yash-rajpal, line 137: Exporting Json implementation?

yash-rajpal (dismissed):

dougfabris (changes requested):

apps/meteor/client/views/audit/components/SecurityLogsTable.tsx
MartinSchoeler: image

apps/meteor/client/views/audit/components/SecurityLogsTable.tsx
```diff
@@ -0,0 +1,216 @@
...
```
dougfabris, line 131: so remove the comment

apps/meteor/client/views/audit/hooks/useSettingSelectOptions.spec.ts
```diff
@@ -0,0 +1,110 @@
...
+jest.mock('@rocket.chat/ui-contexts', () => ({
+	useSettings: () => getMockSettings(),
+	useSettingsCount: () => getMockSettings().length,
+}));
+
+describe('useSettingSelectOptions', () => {
+	it('should return the list of options', () => {
+		const { result } = renderHook(() => useSettingSelectOptions(), {});
+
+		expect(result.current.itemsList.items).toBeDefined();
```
dougfabris, line 108: What about trying to use our MockProvider, introduce some settings and check if your hook is retrieving them?

apps/meteor/client/views/audit/hooks/useSettingSelectOptions.spec.ts
dougfabris, line 1: do you need all these settings data in the mock?

packages/ui-contexts/src/hooks/useSettingsCount.ts
dougfabris, line 1: what is the reason we're introducing it? `querySettings().length` isn't enough?

dougfabris (changes requested):

apps/meteor/client/views/audit/components/SecurityLogsTable.tsx
```diff
@@ -0,0 +1,211 @@
...
+										changedFrom: String(item.data[1].value),
+										changedTo: String(item.data[2].value),
+									})
+								}
+							>
+								<GenericTableCell withTruncatedText>
+									<Box display='flex' alignItems='center'>
+										{item.actor.type === 'user' && (
+											<Box mie={4}>
+												<UserAvatar size='x24' username={item.actor.username} />
```
gabriellsh, line 179: The setting audit record does not update when usernames change, so you need to use the avatar with the user id instead of username.

apps/meteor/client/views/audit/components/SecurityLogDisplayModal.tsx
```diff
@@ -0,0 +1,51 @@
...
+
+	return (
+		<GenericModal icon={null} onClose={onCancel} title={t('Setting_change')}>
+			<InfoPanelLabel>{t('Timestamp')}</InfoPanelLabel>
+			<InfoPanelText>{moment(timestamp).format('MMMM Do YYYY, h:mm:ss a')}</InfoPanelText>
+
+			<InfoPanelLabel>{t('Actor')}</InfoPanelLabel>
+
+			<Box display='flex' alignItems='center' mbe={16}>
+				<UserAvatar size='x24' username={actor} />
```
gabriellsh, line 29: Right here we also need to use the avatar by user id instead of username.

apps/meteor/client/views/audit/components/SecurityLogsTable.tsx
```diff
@@ -0,0 +1,211 @@
...
+										<Box fontScale='p2m' withTruncatedText color='default'>
+											{item.actor.type === 'user' ? item.actor.username : item.actor.type}
+										</Box>
+									</Box>
+								</GenericTableCell>
+								<GenericTableCell withTruncatedText>{item.ts}</GenericTableCell>
+								<GenericTableCell withTruncatedText title={t(String(item.data[0].value))}>
+									{String(item.data[0].value)}
+								</GenericTableCell>
+								<GenericTableCell withTruncatedText>{String(item.data[1].value)}</GenericTableCell>
```
gabriellsh, line 191: You're acessing the data by directly referencing the array position. There's no guarantee that the array will have all the information in the correct order. You need to filter the data by the `key` property.

apps/meteor/client/startup/audit.tsx
```diff
@@ -57,6 +62,17 @@ onToggledFeature('auditing', {
 					</MainLayout>,
 				),
 			},
+			{
+				path: '/security-logs',
+				id: 'security-logs',
+				element: appLayout.wrap(
+					<MainLayout>
+						<PermissionGuard permission='can-audit-log'>
```
gabriellsh, line 70: In the endpoint we use the `can-audit` permission. `can-audit-log` is used for the `Audit Logs` (just `Logs` in the menu), which are the logs of the audit actions. For example, if I audit a room, a log will be created in the `Audit Logs` showing that my user audited a room.

gabriellsh:
Can you update the PR description with new screenshots since the design changed a little bit?

apps/meteor/client/views/audit/components/SecurityLogsTable.tsx
```diff
@@ -0,0 +1,218 @@
...
+						<GenericTableHeaderCell>{t('Changed_to')}</GenericTableHeaderCell>
+					</GenericTableHeader>
+					<GenericTableBody>
+						{data.events.map((item) => {
+							const setting = item.data.find((item) => item.key === 'id');
+							const previous = item.data.find((item) => item.key === 'previous');
+							const current = item.data.find((item) => item.key === 'current');
+							return (
+								<GenericTableRow
+									key={item.ts}
```
gabriellsh, line 166: Sometimes (rare, but does happen) the timestamp can be the same for multiple records. Can we use the record _id instead of ts here?

gabriellsh (changes requested):
Sometimes, the setting value can change from nothing, or to nothing. Maybe we could have an indicator for that (such as "Empty"), so that we don't leave empty fields. Maybe we should make it italic also, so that the text "Empty" is not confused as a value... We could ask the design team to join this one.
image

Setting selector placeholder is empty, until I click it. After I click it, the placeholder is not translated:
image
image

I also noticed that despite having a search text there, when there's no item selected, the text I typed doesn't show until I click it again:
image
image


The filter header is being cut if it does not fit the screen size:
image

Some settings are not appearing in the settings select (there are records for those settings)
image
I think this is related to the "private" settings, since license shouldn't be always available. Cloud workspace id does not show there aswell.

MarcosSpessatto (approved):

PR #35233, opened 2025-02-17
feat: improve color contrast in image gallery for WCAG compliance  

abhinavkrin:

This PR improves the color contrast of icon buttons in the image gallery to ensure compliance with WCAG guidelines. The issue was present in all icon buttons when an image was opened from the chat.

The changes are according to the [design](https://www.figma.com/design/CcKhUgDMecW1xvlAtV86o1/Design-system---Web-%26-desktop?node-id=24546-13583)

https://github.com/user-attachments/assets/850d1582-0717-47c8-8d14-0fa62f9b92c8

## Further comments

[CORE-816](https://rocketchat.atlassian.net/browse/CORE-816)

[CORE-816]: https://rocketchat.atlassian.net/browse/CORE-816?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/meteor/client/components/ImageGallery/ImageGallery.tsx
+++ b/apps/meteor/client/components/ImageGallery/ImageGallery.tsx
@@ -1,6 +1,6 @@
 import type { IUpload } from '@rocket.chat/core-typings';
 import { css } from '@rocket.chat/css-in-js';
-import { Box, ButtonGroup, IconButton, Palette, PaletteStyleTag, Throbber } from '@rocket.chat/fuselage';
+import { Box, ButtonGroup, IconButton, Palette, PaletteStyleTag, Throbber, padding } from '@rocket.chat/fuselage';
 import { useRef, useState } from 'react';
 import { FocusScope } from 'react-aria';
 import { createPortal } from 'react-dom';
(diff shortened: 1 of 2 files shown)


juliajforesti (changes requested):

apps/meteor/client/components/ImageGallery/ImageGallery.tsx
```diff
@@ -32,10 +32,8 @@ const swiperStyle = css`
 		background-color: var(--rcx-color-surface-overlay, rgba(0, 0, 0, 0.6));
 	}
 
-	.rcx-swiper-close-button,
-	.rcx-swiper-prev-button,
-	.rcx-swiper-next-button {
-		color: var(--rcx-color-font-pure-white, #ffffff) !important;
+	.swiper-slide {
+		padding: 9rem 3.75rem 6rem;
```
juliajforesti, line 36: these hardcoded padding values look odd... is it like this by design? couldn't we follow our pattern of x4 values?

juliajforesti (changes requested):
also, I would say this is a `feat` - it improves, not fixes... wdyt?

apps/meteor/client/components/ImageGallery/ImageGallery.tsx
abhinavkrin: I agree we should se the x pattern and applied the same.
I put those values to not let the image coincide with the buttons and the top toolbar when the window is resized.

PR #35235, opened 2025-02-17
regression: Fix issue where marketplace would not fetch the apps

Gustrb:

Our old validation library would throw on error which was causing silent failures, now since we started using zod, we were able to catch this issue quite quickly.

## Further comments

Introduced by https://rocketchat.atlassian.net/browse/CONN-492

[CONN-530](https://rocketchat.atlassian.net/browse/CONN-530)

[CONN-530]: https://rocketchat.atlassian.net/browse/CONN-530?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/meteor/ee/server/apps/marketplace/fetchMarketplaceApps.ts
```diff
@@ -66,20 +66,20 @@ const fetchMarketplaceAppsSchema = z.array(
 		addonId: z.string().optional(),
 		isEnterpriseOnly: z.boolean(),
 		isBundle: z.boolean(),
-		bundedAppIds: z.array(z.string()),
+		bundedAppIds: z.array(z.string()).optional(),
```
KevLehman, line 69: Curious: shouldn't this be `bundled`?

packages/i18n/src/locales/en.i18n.json
```diff
@@ -3631,6 +3631,7 @@
   "Marketplace_Invalid_Apps_Engine_Version": "The installed Apps Engine version is not compatible with the marketplace. Please update the Apps Engine to the latest version.",
   "Marketplace_Internal_Error": "An internal error occurred communicating with Marketplace. Please try again later.",
   "Marketplace_Failed_To_Fetch_Apps": "Failed to fetch apps from the marketplace. Please try again later.",
+  "Marketplace_Core_Schema_Missmatch": "Invalid apps were found in the marketplace. If the problem persists please contact the support.",
```
KevLehman, line 3634:
```suggestion
  "Marketplace_Core_Schema_Missmatch": "Invalid apps were found in the marketplace. If the problem persists please contact support.",
```

A customer may not know what an "invalid app" is. Feel this is really an internal bug. Can we modify the message to something ore user friendly? Like
"Error fetching apps from marketplace" or so.

KevLehman (changes requested):

packages/i18n/src/locales/en.i18n.json
Gustrb: I guess we could be more generic, but I this is not handled in the frontend yet, is just the backend saying what has happened.
Not necessarily customer faced

apps/meteor/ee/server/apps/marketplace/fetchMarketplaceApps.ts
Gustrb: I'll take a look, but yea pretty sure it should

packages/i18n/src/locales/en.i18n.json
KevLehman: Yeah but you're translating and sending the error to the customer. When they face a similar situation, this is what they'll see on the network and may be weird for them.

My suggestion is to change it to be generic to the customer and detailed on error logs (as non-tech people may access marketplace)

Gustrb: I guess, not sure many non-tech people would access the network tab though. What concerns me is that the users are going to see a generic error message that is going to give then 0 guidance about what to do.

One of the main points of the initiative is to make it so the users don't get blasted with generic, Apps_Error_*-like messages whenver something goes wrong.

I think that in the near future, we could even do something like, validate the schema of all the apps, and only render the ones that don't have an error, if one has the error, we could call marketplace-api to mark it as invalid, notify the developer and set the status as invalid or something, so we can avoid fetching it next time.

MarcosSpessatto (changes requested):
If its a regression, please add the PR which introduced the problem please.

Gustrb:
Done

packages/i18n/src/locales/en.i18n.json
KevLehman: Yeah but it makes sense to give them 0 guidance cause this issue is not something they can solve...

Gustrb: But I can't see why just popping a generic error message would be better. Feels like this way the user that already knows how to access the network tab an look for errors there should contact the support so we can go and fix it.
Also, this error will probably be always catched by us, not by end-users, since it will happen usually when we change the marketplace API and test against the core

MarcosSpessatto (approved):

Gustrb:
unsuported version is not the same error, it happens due to your workspace being in a unsupported version

thepiyush-303:
Okay, How do i change my version so that it can support deploying `rc- apps`.

Gustrb:
I would test using branch 7.3.0 or just sync your workspace. should do the trick

thepiyush-303:
> I would test using branch 7.3.0 or just sync your workspace. should do the trick

okay, it will not work on latest version, right?

Gustrb:
I assume there is something fishy with your deployment, but I believe you should open an issue if you can't get it to work

thepiyush-303:
while deploying an Rc-App I land into:
 `
rc-apps deploy --url http://localhost:3000 --username ThePiyush --password asdfasdf
(node:91986) [DEP0040] DeprecationWarning: The `punycode` module is deprecated. Please use a userland alternative instead.
(Use `node --trace-deprecation ...` to show where the warning was created)
   Starting App Deployment to Server

   Getting Server Info... ✓
2025-02-19 00:20:54.385 WARN  App has external module(s) as dependency
(node:91986) [DEP0044] DeprecationWarning: The `util.isArray` API is deprecated. Please use `Array.isArray()` instead.
(node:91986) [DEP0047] DeprecationWarning: The `util.isDate` API is deprecated.  Please use `arg instanceof Date` instead.
   Packaging the app... ✓
   Uploading App... ✖
 ›   Error:    ⟿  Error: Deployment error: [3ae38cf4-3812-4e59-ac19-e8c0968b150c] Timeout: app process not ready
`

I thought there is some problem with marketplace apps fetching.
and also getting this with a fresh clone:
image


 still don't know how i can get it fix :(

Gustrb:
Hard to say without much info. But rc-apps deploy *shouldn’t* have anything to do with this PR
Please open an issue

KevLehman:
@thepiyush-303 , please open an issue with that. But please make sure you have Deno installed on the right version before opening the issue. Thanks!

KevLehman (approved):

thepiyush-303:
Hello @KevLehman ,Just find out my deno version and it was above 2 (don't know when it changed), but now it is fixed and showing apps to me.
Thankyou.

PR #35238, opened 2025-02-17
 fix: : Keeping emojiPicker open even after an emoji selection 

Ansh-dS:
The emoji picker should remain open until the user performs one of the following actions:
- Clicks on the message bar,
- Clicks the send button,
- Clicks outside of the emoji picker (i.e., the focus leaves the popup).

https://github.com/user-attachments/assets/e4ffe678-87e1-4713-a47c-e4b152c0f700

## Issue(s)
 #35200


dougfabris:
Hey @Ansh-dS as we mentioned we're not going to proceed with this code because it's not a fix. 
Thank you!

Ansh-dS:
_sorry to do this again sir but I have deleted my old fork and as code wasn't running so I reinstall the whole code and again re-fork it due to this I lose this as my pr in my "contributions ranking". I totally understand that the emojipicker behaves that way as it commands to but in my opinion this is a way to show that I can contribute too, here i am wrong but  not every time._

PR #35249, opened 2025-02-18
chore: endpoint change in channels.members test

abhinavkrin:


KevLehman:
Can you check? Seems whole API CI is failing. Maybe this is still a draft?

MarcosSpessatto (approved):

apps/meteor/tests/end-to-end/api/channels.ts
```diff
@@ -1959,46 +1956,45 @@ describe('[Channels]', () => {
...
+					.expect(400)
 					.expect((res) => {
 						expect(res.body).to.have.property('success', false);
 					});
 			});
 
 			it('should not fetch a private channel members inside a public team by someone not part of the team ', async () => {
 				await request
-					.get(api('rooms.membersOrderedByRole'))
+					.get(api('channels.members'))
```
sampaiodiego, line 1985: @abhinavkrin I'm very confused regarding all the changes on this PR, but this one makes me even more confused..

the `rooms.*` endpoints are "typeless", they don't care which type of room you're fetching the data from.. in the other hand `channels.*` endpoints are intended to be used for *public channels* only.. 

having this context in mind, the change you introduced here seems you're getting incorrect results because you're calling the `/api/v1/channels.members` endpoint passing a private channel as a param, as you can see by the name of the variable `privateChannelInPublicTeam` .. so the status `400` you're receiving is just because of that, it is not an assertion for your test which is "should not fetch a private channel members inside a public team by someone not part of the team"

abhinavkrin: While implementing the hotfix, I copied the visibility tests from rooms.membersOrderedByRole to the channels endpoints and modified them accordingly. However, for the channels.members endpoint, I mistakenly forgot to replace the endpoint reference.

One of the tests, which passes a private channel ID and expects a 400 response, is designed to verify that the channels endpoint does not return data for a group. Since a group ID is an invalid parameter for a channels endpoint, it should correctly return an error.

sampaiodiego: thanks @abhinavkrin , I looked the whole file and its tests and it makes more sense now, I was being tricked by the diff only..

PR #35275, opened 2025-02-20
fix: Composer not working on embedded layout

yash-rajpal:

Subscription wasn't being serialized before being added to the mini-mongo collection, causing weird bugs.
## Steps to test or reproduce

 - Open a room in embedded layout
 - Composer is disabled, behaves as if user is not part of the room
## Further comments

[CORE-995](https://rocketchat.atlassian.net/browse/CORE-995)

### DIFF
--- a/apps/meteor/client/views/room/RoomOpenerEmbedded.tsx
+++ b/apps/meteor/client/views/room/RoomOpenerEmbedded.tsx
@@ -59,7 +59,13 @@ const RoomOpenerEmbedded = ({ type, reference }: RoomOpenerProps): ReactElement 
 			return;
 		}
 
-		CachedChatSubscription.upsertSubscription(subscriptionData.subscription as unknown as ISubscription);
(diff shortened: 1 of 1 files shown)


aleksandernsilva (changes requested):
I believe this issue was introduced on #35204. If that's the case this PR classifies as a regression and no changeset is needed _(PR title would need to be adjusted as well)_.

FYI: A patch was also requested here #35251

yash-rajpal:
I thought so too @aleksandernsilva but this wasn't working in 7.2.x or 7.3.x as well (tested on unstable server).

The PR mentioned by you fixes some other issue related to embedded rooms, doesn't fix the issue with composer. 

In this case, I think this isn't a regression (technically) 🤔

aleksandernsilva (approved):

yash-rajpal:
Closing in favor of #35295

PR #35286, opened 2025-02-20
chore: Bump fuselage

rique223:
## Further comments

Jira task: [CONN-547](https://rocketchat.atlassian.net/browse/CONN-547)

[CONN-547]: https://rocketchat.atlassian.net/browse/CONN-547?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


juliajforesti (approved):

.changeset/bright-snakes-punch.md
```diff
@@ -0,0 +1,17 @@
...
+"@rocket.chat/ui-client": patch
+"@rocket.chat/ui-composer": patch
+"@rocket.chat/ui-contexts": patch
+"@rocket.chat/ui-video-conf": patch
+"@rocket.chat/ui-voip": patch
+"@rocket.chat/web-ui-registration": patch
+"@rocket.chat/ui-theming": patch
+---
+
+Bumps fuselage and related packages versions to use the most recent releases of each package, especially the fix for the missing track of the fuselage slider component
```
aleksandernsilva, line 17: Since this is a chore, I don't think we need the changeset 😉

aleksandernsilva (changes requested):

.changeset/bright-snakes-punch.md
rique223: Although this is marked as a chore this PR will fix a bug on the user preferences page together with the package bump and since the actual fix PR for this is on fuselage I've added this changeset so that our end-users are aware of the fix.

PR #35309, opened 2025-02-22
fix: multiple node.js version can be used 

Kaif9999:
Now developers can use any node version in development

Updated the Dockerfile
Updated CI/CD Configuration
Updated CI Workflows

## Issue(s)
The issue is that it only runs Node version 22.13.1, which don't let the developer to use another version


KevLehman:
Hey, thanks for the contribution. This is not an issue, we only support this version and all our tests are on this version so we provide no guarantees if you use another one.

I'm gonna close this one, thx!

PR #35319, opened 2025-02-24
refactor: Make `Base.col` fully private - 1

KevLehman:

Idea is to stop using `.col` in models/app code and use just the methods exposed by the Base model.

This removes the usage of `.col.countDocuments` and `.col.estimatedDocumentCount` in favor of the methods defined in `Base` model.

### DIFF
--- a/apps/meteor/app/authorization/server/methods/removeUserFromRole.ts
+++ b/apps/meteor/app/authorization/server/methods/removeUserFromRole.ts
@@ -65,7 +65,7 @@ Meteor.methods<ServerMethods>({
 
 		// prevent removing last user from admin role
 		if (role._id === 'admin') {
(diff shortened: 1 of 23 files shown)


sampaiodiego (commented):
I don't bother too much using `this.col` inside models, I think it could be just `protected` instead of `public`.. but that's fine..

as a first step I think we're good.. but I still would like to that we don't use `countDocuments` providing queries all over the place, I think we should be doing that inside models only.. I don't think is making it worse though.

sampaiodiego (approved):

KevLehman:
Yeah, moving the count queries to their models would be nice. They're a lot and that would require a lot of changes, so i'll create a task for start moving those (and other queries that are outside them).

Thx man!

MartinSchoeler (approved):

PR #35330, opened 2025-02-25
fix: videoconf ringer not working after a temporary disconnection

pierre-lehnen-rc:

The VideoConfManager was tracking changes to the authenticated UserId to determine if an user is logged in or not, but in some cases when the connection to the server is interrupted and a new connection is started, the UserId will be available before the new Connection's authentication is completed. In some situation that we were not yet able to reproduce in development, this caused the client to attempt to subscribe to videoconf events in an unauthenticated connection, which would fail and prevent videoconf messages from being received.
This PR changes the VideoConfManager to track the connection and authentication state along with the UserId, ensuring that the subscription is refreshed whenever any of those things change.

## Issue(s)

[SUP-744](https://rocketchat.atlassian.net/browse/SUP-744)

## Further comments

[SUP-744]: https://rocketchat.atlassian.net/browse/SUP-744?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


aleksandernsilva (approved):

aleksandernsilva:
@pierre-lehnen-rc could you add a milestone and the steps to reproduce the issue in the PR description?

PR #35332, opened 2025-02-25
fix: Apps preventing join actions not showing any error

tiagoevanp:
When Preview Public Channel authorization is not in the user permissions group and an app is preventing the same user to join, the UI should dispatch a toast error message with the app reason.

## Further comments

[SUP-742]

[SUP-742]: https://rocketchat.atlassian.net/browse/SUP-742?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[CORE-957](https://rocketchat.atlassian.net/browse/CORE-957?atlOrigin=eyJpIjoiYmEwOTk4MThjMmVjNGI3MzkyNTU1NzMwYTgwMzU3ZGEiLCJwIjoiaiJ9)


tiagoevanp:
Tests will be covered in another PR

apps/meteor/client/hooks/useJoinRoom.ts
```diff
@@ -2,13 +2,17 @@ import type { IRoom } from '@rocket.chat/core-typings';
...
+import type { AppPreventedError } from '../lib/errors/AppPreventedError';
+import { dispatchToastMessage } from '../lib/toast';
 
 type UseJoinRoomMutationFunctionProps = {
 	rid: IRoom['_id'];
 	reference: string;
 	type: IRoom['t'];
 };
 
+const isAppError = (error: AppPreventedError | Error): error is AppPreventedError => (error as AppPreventedError).reason !== undefined;
```
MarcosSpessatto, line 14: Can we assume that by only not having reason is an app error? 🤔

MarcosSpessatto (changes requested):

apps/meteor/client/hooks/useJoinRoom.ts
tiagoevanp: In this case, yes... the app must have a reason in the response...

gabriellsh, line 14:
```suggestion
const isAppError = (error: AppPreventedError | Error): error is AppPreventedError => error instanceof AppPreventedError;
```

gabriellsh: I didn't test, but doesn't this work? I think a bunch of errors can have the "reason" property.

gabriellsh: Otherwise, I think you shouldn't create `AppPreventedError`. You could just cast `error: unknown` and test `'reason' in error`. Just throwing ideas

gabriellsh (dismissed):

apps/meteor/client/hooks/useJoinRoom.ts
tiagoevanp: In fact, the reason could appear for another errors, I wasn't totally happy or sure with my solution, it looks like your recommendation is really good if it works. I will test it and probably approve...

I was trying to follow some pattern that we discussed before in our frontend team about given errors a proper description and details.

MarcosSpessatto (approved):

tassoevan (approved):

PR #35339, opened 2025-02-26
fix: Client application memory leaks

gabriellsh:


apps/meteor/client/hooks/useReactiveQuery.ts
```diff
@@ -10,13 +11,18 @@ export const useReactiveQuery = <TQueryFnData, TData = TQueryFnData, TQueryKey e
 	options?: UseQueryOptions<TQueryFnData, Error, TData, TQueryKey>,
 ): UseQueryResult<TData, Error> => {
 	const queryClient = useQueryClient();
+	const computation = useRef<Tracker.Computation | null>(null);
+
+	useEffect(() => {
+		return () => computation.current?.stop();
+	}, []);
```
ggazzo, line 18: use abort signal instead

ggazzo (approved):

PR #35340, opened 2025-02-26
fix: 2FA check flashing for mandatory roles

yash-rajpal:
## Further comments

---

This pull request addresses an issue in the Rocket.Chat repository related to the flashing of the two-factor authentication (2FA) check for mandatory roles. The changes are made in the `fix/2FA-mandatory-check-flashing` branch and are intended to be merged into the `develop` branch.

The modifications include:

1. **apps/meteor/app/authentication/server/startup/index.js**: The user fields projection is updated by replacing `getBaseUserFields` with `getBaseUserFieldsWithTwoFactor`. This change ensures that two-factor authentication fields are included in the default published fields for user accounts.

2. **apps/meteor/app/utils/server/functions/getBaseUserFields.ts**: A new function, `getBaseUserFieldsWithTwoFactor`, is added. This function returns user fields that include 2FA settings, enhancing the handling of user data with respect to two-factor authentication.

3. **apps/meteor/server/startup/initialData.js**: The code now includes logic to insert an admin user using environment variables. It also manages different scenarios such as the presence of existing admins, identifying the oldest user, and operating in test mode.

These changes aim to improve the handling of two-factor authentication for users with mandatory roles, ensuring a more stable and secure authentication process.

[CORE-1000](https://rocketchat.atlassian.net/browse/CORE-1000)

[CORE-1000]: https://rocketchat.atlassian.net/browse/CORE-1000?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


yash-rajpal:
@kody start-review

lucas-a-pelegrino (dismissed):

apps/meteor/app/authentication/server/startup/index.js
```diff
@@ -40,7 +40,7 @@ Accounts.config({
  *
  * we are removing the status here because meteor send 'offline'
  */
-Object.assign(Accounts._defaultPublishFields.projection, (({ status, ...rest }) => rest)(getBaseUserFields()));
+Object.assign(Accounts._defaultPublishFields.projection, (({ status, ...rest }) => rest)(getDefaultUserFields()));
```
gabriellsh, line 43: I noticed that `getDefaultUserFields` also includes `'services.password.bcrypt'`. I don't think this field is supposed to be passed to client, did you ensure this doesn't happen?

gabriellsh (changes requested):

apps/meteor/app/authentication/server/startup/index.js
lucas-a-pelegrino: Good catch @gabriellsh !

@yash-rajpal , assuming your intentions here were to include the 2fa info. I would suggest including/extending the attire manually instead of projecting all the other attrs present in the `getDefaultUserFields()` func.

What do you guys think?

gabriellsh: Ok so I confirmed and this shouldn't really be happening. What I suggest:

Extend `getBaseUserFields` to include `'services.totp.enabled': 1, 'services.email2fa.enabled': 1,` and remove those from `getDefaultUserFields`.

This should suffice for now.

yash-rajpal: As discussed on DM, pushed the fix.

lucas-a-pelegrino (dismissed):

MarcosSpessatto (approved):

gabriellsh (approved):

apps/meteor/app/authentication/server/startup/index.js
```diff
@@ -40,7 +40,10 @@ Accounts.config({
  *
  * we are removing the status here because meteor send 'offline'
  */
-Object.assign(Accounts._defaultPublishFields.projection, (({ status, ...rest }) => rest)(getBaseUserFields()));
+Object.assign(
+	Accounts._defaultPublishFields.projection,
+	(({ status, ...rest }) => rest)({ ...getBaseUserFields(), 'services.totp.enabled': 1, 'services.email2fa.enabled': 1 }),
```
pierre-lehnen-rc, line 45: Why not simply add the two fields to `getBaseUserFields` directly?

The only other place that uses this function already adds the whole `service` to it anyway. That other place could even be using `getDefaultUserFields` instead.

pierre-lehnen-rc (changes requested):
This PR only fixes the issue for workspaces that rely on meteor's DDP streamer. Our own DDP Streamer micro service will still have the issue.

apps/meteor/app/authentication/server/startup/index.js
pierre-lehnen-rc: Ideally, we should be publishing the exact same fields as the `api/v1/me` endpoint returns;

The problem is that that place returns calculated data too (for example, it creates `services.password.exists` based on `services.password.bcrypt`), which is beyond what we can do with a simple projection list; We would need to override some meteor code to be able to calculate it for the publication too.

yash-rajpal: > Why not simply add the two fields to getBaseUserFields directly?

Because we use this helper on `me` endpoint, and there we need the entire `services` object. If we make it part of helper, the mongo query would error out due to overlapping projection fields.


> That other place could even be using getDefaultUserFields instead.

Not really, because we need the entire `services` obj to check if user is OAuth user or not.

But I do see your point, this entire thing needs some refactor and planning.

apps/meteor/app/authentication/server/startup/index.js
```diff
@@ -40,7 +40,7 @@ Accounts.config({
  *
  * we are removing the status here because meteor send 'offline'
  */
-Object.assign(Accounts._defaultPublishFields.projection, (({ status, ...rest }) => rest)(getBaseUserFields()));
+Object.assign(Accounts._defaultPublishFields.projection, (({ status, ...rest }) => rest)({ ...getBaseUserFields(true) }));
```
pierre-lehnen-rc, line 43:
```suggestion
Object.assign(Accounts._defaultPublishFields.projection, (({ status, ...rest }) => rest)(getBaseUserFields(true)));
```

pierre-lehnen-rc (changes requested):

pierre-lehnen-rc (approved):

pierre-lehnen-rc (approved):

MarcosSpessatto (approved):

gabriellsh (dismissed):

ggazzo (approved):

ggazzo (approved):

PR #35361, opened 2025-02-27
fix: Some EE features not available after changing plans

MartinSchoeler:

Some features were not being available if you changed plans without reseting your server, we probably changed the query key somewhere along the way and forgot to update this one

## Issue(s)

[CORE-1007](https://rocketchat.atlassian.net/browse/CORE-1007)
## Steps to test or reproduce

1. Start with a CE server
2. Add an enterprise key with the license to audit messages
3. Reload page (this pr still does not change that)
4. Go to audit messages or logs page
5. The page should load

## Further comments

possibly introduced here https://github.com/RocketChat/Rocket.Chat/pull/30948/files#diff-db7592984a0cf1a2169339268af6bd61cb29f62f0173c3b0ea61126613151a0dL9

[CORE-1007]: https://rocketchat.atlassian.net/browse/CORE-1007?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/meteor/client/lib/onToggledFeature.ts
+++ b/apps/meteor/client/lib/onToggledFeature.ts
@@ -15,7 +15,7 @@ export const onToggledFeature = (
 	},
 ): (() => void) => {
 	const observer = new QueryObserver(queryClient, {
-		queryKey: ['ee.features'],
+		queryKey: ['licenses'],
 		queryFn: fetchFeatures,
 		staleTime: Infinity,
 	});
--- a/.changeset/eighty-meals-rule.md
+++ b/.changeset/eighty-meals-rule.md
@@ -1,0 +1,5 @@
+---
+'@rocket.chat/meteor': minor
+---
+
+Fixes Some EE features not available after changing plans


apps/meteor/client/lib/onToggledFeature.ts
```diff
@@ -15,7 +15,7 @@ export const onToggledFeature = (
 	},
 ): (() => void) => {
 	const observer = new QueryObserver(queryClient, {
-		queryKey: ['ee.features'],
```
aleksandernsilva, line 18: A quick search shows me that this query key is also being used by [`hasLicense`](https://github.com/RocketChat/Rocket.Chat/blob/develop/apps/meteor/app/license/client/index.ts). Could you check and see if it's not relevant to change there as well?

aleksandernsilva (changes requested):
Don't forget to add a changeset 😉

apps/meteor/client/lib/onToggledFeature.ts
MartinSchoeler: @aleksandernsilva  Yeah this function is only used here https://github.com/RocketChat/Rocket.Chat/blob/fix/ee-query/apps/meteor/app/livechat-enterprise/client/index.ts#L3

This function mainly sets the omnichannel EE items on the omnichannel admin section, I've done some testing and there were no difference in the way these items are behaving. So we should be ok

aleksandernsilva:
@MartinSchoeler Could you also add the "Steps to test or reproduce"?

dougfabris (approved):

aleksandernsilva (approved):

PR #35365, opened 2025-02-28
Fix: Update media URLs to work with Gitpod

### LINKED ISSUE #35364: Bug: Media (Video, Image, Audio) URLs Use localhost Instead of Gitpod URL
### 🐞 Description:

Media files (videos, images, and audio) in Rocket.Chat fail to load in Gitpod because they are being served using `localhost`, which does not work in Gitpod’s dynamic URL environment.

### ⚠️ Steps to reproduce:

1. Open Rocket.Chat in Gitpod.
2. Try to load/play a media file (video, image, or audio).
3. The request fails because the URL is using `localhost` instead of Gitpod’s dynamic domain.

### 🎯 Expected behavior:

Media should use the correct Gitpod URL instead of `localhost`, allowing it to load properly in Gitpod environments.

### 🛑 Actual behavior:

Media files fail to load because they rely on `localhost`, which is inaccessible in Gitpod.

### 📜 Additional context:

This issue occurs only in Gitpod environments where `localhost` is not accessible. Instead, the media files should use Gitpod’s generated URLs.

### 📝 Relevant logs:

Image

Raushankumar-prog:
This PR fixes issue #35364 by updating media (video, image, audio) URLs to use the correct Gitpod URL instead of localhost.

### Changes:
- Replaced `localhost` with  appropriate dynamic Gitpod URL. 
- Ensured media loads correctly in the Gitpod environment.

### Testing:
- Verified that media loads correctly in Gitpod.
- No regressions for localhost development.

Fixes #35364
Screenshot 2025-02-28 180114


dougfabris (commented):
can u check what is the `Site_Url` setting value when running inside gitpod?

Raushankumar-prog:
@dougfabris 
  Site URL: https://3000-raushankumar-rocketchat-8z0ccz7anb0.ws-us118.gitpod.io

dougfabris (changes requested):
so we should look to make this verification in the function that return this `useMediaUrl`. I think there are some help function inside the `AttachementProvider`

Raushankumar-prog:
@dougfabris 



When we run a development server in Gitpod, it starts on localhost:3000 inside the Gitpod container. Gitpod notices this and automatically creates a public link like https://3000-<workspace>.gitpod.io/ so you can access it. The getURL function works in a similar way—it checks if the app is running locally or in production and creates the right URL for each case. The difference is that getURL switches between local and production, while Gitpod is always in the cloud, so localhost inside Gitpod is not actually your computer’s localhost.




AttachmentContext.ts

import { createContext } from 'react';

export type AttachmentContextValue = {
	getURL: (url: string) => string;
	dimensions: {
		width: number;
		height: number;
	};
	collapsedByDefault: boolean;
	autoLoadEmbedMedias: boolean;
};

export const AttachmentContext = createContext<AttachmentContextValue>({
	getURL: (url: string) => url,
	dimensions: {
		width: 368,
		height: 368,
	},
	collapsedByDefault: false,
	autoLoadEmbedMedias: true,
});




getURL



getURL(path: string) {
		const rootUrl = Meteor.absoluteUrl('', { secure: UploadFS.config.https }).replace(/\/+$/, '');
		const storeName = this.getName();
		path = String(path).replace(/\/$/, '').trim();
		return encodeURI(`${rootUrl}/${UploadFS.config.storesPath}/${storeName}/${path}`);
	}

PR #35369, opened 2025-02-28
fix: inner attachment empty when quoting multiple times

abhinavkrin:

Fixes an issue where quoting an attachment (image, video, file, audio) multiple times resulted in the inner attachment being empty. The description and preview was lost in the inner message.
**before**
Screenshot 2025-03-01 at 2 47 34 AM

**after**
Screenshot 2025-03-01 at 2 45 50 AM

## Further comments

[CORE-37](https://rocketchat.atlassian.net/browse/CORE-37)

[CORE-37]: https://rocketchat.atlassian.net/browse/CORE-37?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/meteor/tests/unit/server/services/messages/hooks/BeforeSaveJumpToMessage.tests.ts
MarcosSpessatto, line 1: Can we add a new test case for when the last message does not have any file attachment? It's just a quote.

Also, can we add a case when there are more messages than the chain, just to ensure it's cutting the last ones?

MarcosSpessatto (changes requested):

MarcosSpessatto (approved):

PR #35409, opened 2025-03-06
fix: use db instead of client in startTracing function

### LINKED ISSUE #35408: fix : client is undefined in startTracing() in service.ts
In rocket.chat/_templates/service/new/service.ejs.t

**the function startTracing() is called with a client variable, but client is not defined anywhere in the file. This causes a runtime error when the service starts.**

Image

**Steps to reproduce**

- find the file  Rocket.chat/_templates/service/new/service.ejs.t

- find this line startTracing({ service: '<%= name %>', db: client });

- Notice that client is never declared in this file.

- Try running the service—it will fail due to the missing client variable.

Expected behavior:

The function should use db instead of client because db is already defined from getConnection().

Actual behavior:

Error: ReferenceError: client is not defined

The service fails at runtime because client is never declared.

Additional Context:
This is a small fix that requires replacing client with db in startTracing(). I’d be happy to submit a PR for this.

naman22a:
# fix: use db instead of client in startTracing function

## Description
This PR updates the startTracing function to use the db variable instead of client for better consistency and code clarity.

## Changes Made

- Replaced client with db in the startTracing function.
 
Before
`startTracing({ service: '<%= name %>', db: client });`
After
`startTracing({ service: '<%= name %>', db: db });`

## Issue(s)
Fixes #35408


naman22a:
Sorry is it fine now ? , this is my first contribution

PR #35420, opened 2025-03-07
fix: VoIP calls one-way audio issues

pierre-lehnen-rc:
## Further comments

To prevent react from destroying the audio element that is playing a call's audio, a workaround was made to pass this element as reference to the voip client, but in some situations the element was not passed at all and the voipclient had nothing to play the audio on.

I've changed the VoIPProvider to replace the ref with a callback which automatically updates the element whenever it is available.


packages/ui-voip/src/providers/VoipProvider.tsx
```diff
@@ -120,11 +124,12 @@ const VoipProvider = ({ children }: { children: ReactNode }) => {
 	}, [dispatchToastMessage, setStorageRegistered, t, voipClient, voipSounds]);
 
 	const changeAudioOutputDevice = useEffectEvent(async (selectedAudioDevice: Device): Promise<void> => {
-		if (!remoteAudioMediaRef.current) {
+		const element = voipClient?.getMediaElement();
+		if (!element) {
```
aleksandernsilva, line 128: Might be worth it adding a log here in the case this scenario occurs 🤔

packages/ui-voip/src/providers/VoipProvider.tsx
```diff
@@ -39,7 +39,15 @@ const VoipProvider = ({ children }: { children: ReactNode }) => {
 	const dispatchToastMessage = useToastMessageDispatch();
 
 	// Refs
-	const remoteAudioMediaRef = useRef<HTMLAudioElement>(null);
+	const remoteAudioMediaRef = useCallback(
+		(node: HTMLMediaElement | null) => {
+			if (!node || !voipClient) {
+				return;
```
tassoevan, line 45: Consider removing the old reference for the element on `voipClient`'s internal state to avoid leaks.

tassoevan (changes requested):

aleksandernsilva (approved):

tassoevan (approved):

PR #35426, opened 2025-03-07
feat: memoize app getStatus calls

Gustrb:

Due to the previous architecture of the apps-engine, several flows are dependent on calling the getStatus method exposed by the app.

However, after Deno runtime, these add a lot more overhead compared to simple in memory access, clog the output queue of the subprocess and are prone to errors that simply didn’t happen before (30s timeouts, subprocess unavailable, etc.)

These calls should be better handled to reduce the strain on the system.

## Further comments

[CONN-451](https://rocketchat.atlassian.net/browse/CONN-451)

[CONN-451]: https://rocketchat.atlassian.net/browse/CONN-451?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/apps-engine/src/server/AppStatusCache.ts
```diff
@@ -0,0 +1,40 @@
...
+    private clearExpired(): void {
+        const now = new Date();
+
+        for (const appId in this.cache) {
+            if (this.cache[appId].expiresAt < now) {
+                delete this.cache[appId];
+            }
+        }
+    }
+}
```
debdutdeb, line 40: i think this and the timer is unnecessary.

if cached item expired, you are setting a fresh one anyway and returning undefined here.

packages/apps-engine/src/server/AppStatusCache.ts
```diff
@@ -0,0 +1,40 @@
+import type { AppStatus } from '../definition/AppStatus';
+
+export class AppStatusCache {
+    private readonly cache: Record<string, { status: AppStatus, expiresAt: Date }> = {};
+    private readonly CACHE_TTL = 1000 * 60;
+
+    // We clean up the cache every 5 minutes
+    private readonly CLEANUP_INTERVAL = 1000 * 60 * 5;
```
debdutdeb, line 8: give them environment variables so we can configure from outside?

packages/apps-engine/src/server/AppStatusCache.ts
Gustrb: yea, I forgot to take into account that a workspace cannot have that many apps such that this would be required

packages/apps-engine/src/server/AppStatusCache.ts
d-gubert: I feel like we've been adding too many environment variables to configure small parts of the apps-engine, it's getting a bit out of hand.

I was thinking about something similar to the ts-node env var that gets a JSON string as value, like `TS_NODE_COMPILER_OPTIONS='{\"module\": \"commonjs\"}'`. What do you think? @debdutdeb

debdutdeb: Oh yeah that sounds better with that context. :+1:

d-gubert (approved):

packages/apps-engine/src/server/ProxiedApp.ts
```diff
@@ -78,13 +82,11 @@ export class ProxiedApp {
         }
     }
 
-    public async getStatus(): Promise<AppStatus> {
-        return this.appRuntime.getStatus().catch(() => AppStatus.UNKNOWN);
-    }
+    public getStatus = mem(() => this.appRuntime.getStatus().catch(() => AppStatus.UNKNOWN), { maxAge: 1000 * 60 * 5 });
```
debdutdeb, line 85: Can we log this error? @d-gubert 🙈

debdutdeb (approved):

PR #35436, opened 2025-03-08
chore(meteor-methods)!: remove deprecated canAccessRoom method

### LINKED ISSUE #35432: Remove deprecated canAccessRoom method scheduled for removal after v5.0
## Description
The `canAccessRoom` Meteor method in `server/methods/canAccessRoom.ts` is marked as deprecated with a warning message indicating it will be removed after version 5.0. This method is only enabled when the environment variable `ALLOW_CANACCESSROOM_METHOD` is set to "yes" or "true".

## Expected Behavior
The deprecated method should be removed, and any code relying on it should use the `canAccessRoomAsync` function from the Authorization service directly.

## Current Behavior
The deprecated method is still available when the environment variable is set, displaying a warning message:
```console.warn('Method canAccessRoom is deprecated and will be removed after version 5.0');```

## Steps to Reproduce
1. Set environment variable `ALLOW_CANACCESSROOM_METHOD=true`
2. Start Rocket.Chat server
3. Check server logs for the deprecation warning

## Relevant Code
```typescript
// server/methods/canAccessRoom.ts
if (['yes', 'true'].includes(String(process.env.ALLOW_CANACCESSROOM_METHOD).toLowerCase())) {
  console.warn('Method canAccessRoom is deprecated and will be removed after version 5.0');
    async canAccessRoom(rid, userId, extraData) {
      // Method implementation
    }
  });
}

blackmamba1231:
## Proposed changes
This PR removes the deprecated `canAccessRoom` method that was scheduled for removal after v5.0 but is still present in v7.4.0. The method was only enabled when the environment variable `ALLOW_CANACCESSROOM_METHOD` was set to "yes" or "true".

Changes made:
- Removed the `canAccessRoom.ts` file
- Updated the [index.ts](cci:7://file:///d:/Rocket.Chat/apps/meteor/server/methods/index.ts:0:0-0:0) file to remove the import for the `canAccessRoom` method

## Issue(s)
Fixes #35432

## Steps to test or reproduce
1. Verify that the `canAccessRoom` method is no longer available
2. Ensure that applications using the async alternative `canAccessRoomAsync` continue to work correctly
3. No functionality should be affected as this method was already deprecated and only available when explicitly enabled via environment variable

## Further comments
This change is part of the ongoing effort to clean up deprecated code in the Rocket.Chat codebase. The method was marked for removal after v5.0 and we're now at v7.4.0, so it's time to complete the deprecation process.


Gustrb:
Even though this was already scheduled for removal I don't see why removing it in a minor version. We should probably wait for 8.0 so we can remove it without the concern of introducing breaking changes

blackmamba1231:
> Even though this was already scheduled for removal I don't see why removing it in a minor version. We should probably wait for 8.0 so we can remove it without the concern of introducing breaking changes

Thank you for the feedback. You make a good point about following semantic versioning principles. 

While the method was scheduled for removal after v5.0 and we're now at v7.x, I understand the concern about potential breaking changes even for deprecated functionality. Users might still be relying on this method with the environment variable enabled.

I agree that it would be better to target this removal for the v8.0 release instead. Would you prefer that I:

1. Close this PR and create a new one targeting the v8.0 milestone
2. Keep this PR open but update it to target the v8.0 milestone
3. Add a more explicit deprecation warning that specifies the method will be removed in v8.0

I'm happy to follow whatever approach the team prefers for handling deprecated functionality.

ggazzo (approved):

Issue #35437, opened 2025-03-08
fix: Improve emoji name display by removing colons


Ansh-dS:
#### **Description:**
The displayed name of emojis, when hovered over, includes colons (e.g., `:smile:`), which does not look user-friendly. The colons should be removed to improve readability.

#### **Steps to Reproduce:**
1. Open a **team**, **channel**, or **direct message** from the left pane.  
2. Click on the emoji icon and choose any emoji, then send it.  
3. Hover over the emoji.  

#### **Expected Behavior:**
- The tooltip should display a clean emoji name without colons.  
- Example: Instead of `:smile:`, it should display `smile`.  
  
 #### Actual Behavior:

https://github.com/user-attachments/assets/c4360fe7-365e-430f-a0b7-e1f30d5f849f

  
  #### Server/Client Information:
  
  #### **Server:**
  - **Operating System:** Windows 11 (using WSL Terminal)
  - **Rocket.Chat Server Version:** 7.5.0-develop
  - **Node.js Version:** 22.13.1
  - **Yarn Version:** 4.6.0
  
  #### **Client:**
  - **Client Type:** Browser (Microsoft Edge)
  - **Access URL:** [http://localhost:3000](http://localhost:3000)
  
  #### **Additional Information:**
  - **WSL Terminal:** Used for installing dependencies.
  
  ### **Are you working on this?**  
✅ Yes, I am working on this.

rxhul18:
assign me

Ansh-dS:
@rxhul18 Nobody assigns the issue. First read the developers document "how to ask questions to get attention" more and more.

HarshMethwani:
Hi Just noticed this issue! Removing the colons can definitely improve readability. Another approach could be to display both the emoji and its name together for better clarity as user hovers.

Ansh-dS:
@sidmohanty11 why have to made a pr if I already mentioned that "I am working on this". Secondly let "reetp" confirm this issue weather the team need this or not.

Ansh-dS:
@HarshMethwani why you want to mention a emoji in hover with it's name( name is enough)  when you hover on emoji (displayed ) on the screen. I never seen displaying the emoji two times anywhere before in other apps or website.

sidhhantPr:
HI, I closed the PR. I took the approach of clicking on an emoji and directly adding it to the text.

priyanshuharshbodhi1:
[...]

PR #35441, opened 2025-03-09
fix: remove colons from emoji names  #35437

sidhhantPr:
## Fixes-Improve emoji name display by removing colons #35437
Changes :
1.Send an emoji in a message.
2.Hover over the emoji.
3.Previously, the tooltip displayed names with colons (e.g., :smile:).
Now, it correctly shows names without colons.
4.This is a UI improvement with no breaking changes.

##Solved
EmojiFixed


Ansh-dS:
I already mentioned that "i am working" still @sidhhantPr made this pr. #35437

Issue #35335, opened 2025-02-26
Duplicate Toast Notification on "Invite Link" Action


connectshruti:
### Description:

When a user attempts to invite another user to a public or private channel in the Rocket.Chat web application by clicking the "Invite Link" button, the system generates the invite link correctly. However, it displays the "Invite link has been generated" message twice, showing two identical toast notifications simultaneously.

### Steps to reproduce:

1. Navigate to any public or private channel.
2. Click on the "Invite" button to add a new user.
3. Select the "Invite Link" option.
4. Observe that two identical toast notifications appear instead of one.

### Expected behavior:

Only a single toast notification should be displayed when the invite link is generated.

### Additional context

Rocket.Chat uses Fuselage Toast for notifications. The issue might be related to redundant triggers for the toast component in the codebase.

While this might seem like a small issue, it could be a good first issue for new contributors or may need attention to improve the user experience.

https://github.com/user-attachments/assets/d5c8f935-3dc6-492d-8459-df01b3e40447

shivamsaklani:
working on this

Shubhamkahar196:
Hi, I noticed there are a couple of PRs already linked to this issue (#35345 and #35445).  
If those are inactive or need additional fixes, I’d be happy to work on this issue.  
Please let me know if I can proceed. 😊

Rohit544:
Hi, I’m Naveen. I attended the Rocket.Chat community workshop and would like to work on this issue as my first contribution.

Could I be assigned to it?

PR #35445, opened 2025-03-10
fix: Remove Duplicate toast notification on "Invite link" action

dasankush:
## **What does this PR do?**  
This PR resolves the issue where the toast notification was displayed twice when clicking on "Invite Link." Now, the notification appears only once as expected.  

## Proposed changes

Before Implementation:

https://github.com/user-attachments/assets/e7a4eeb6-2a18-4828-aa29-c75c8ae2f923

After Implementation:

https://github.com/user-attachments/assets/99adcf73-2f9a-4753-b6eb-da2b9a89375f

## Issue(s)

[35335](https://github.com/RocketChat/Rocket.Chat/issues/35335)

## Steps to test or reproduce

1. Go to the **Home** screen.  
2. Create a **new channel**.  
3. Inside the channel, click on the **member icon** at the top.  
4. Click on **"Invite Link"**.  
5. Previously, this would trigger **two toast notifications** instead of one.  

### Steps to Test the Fix  
- Follow the same steps as above and verify that only **one toast notification** appears.  

## Further comments

This fix ensures that the notification logic is executed only once, preventing duplication.

### DIFF
--- a/apps/meteor/client/views/room/contextualBar/RoomMembers/InviteUsers/InviteUsersWithData.tsx
+++ b/apps/meteor/client/views/room/contextualBar/RoomMembers/InviteUsers/InviteUsersWithData.tsx
@@ -119,4 +119,4 @@ const InviteUsersWithData = ({ rid, onClickBack }: InviteUsersWithDataProps): Re
 	);
 };
 
-export default InviteUsersWithData;	
+export default InviteUsersWithData;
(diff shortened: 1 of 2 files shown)


tassoevan:
Thanks for it. However, I believe there is something fundamentally wrong in triggering this toast message on an effect i.e. depending on a component update. Maybe the definitive solution need more work, like moving this trigger to that button's click event listener.

PR #35454, opened 2025-03-10
regression: room list not using full sidebar width in new navigation

abhinavkrin:

This PR fixes an issue where the room list in the new navigation was not occupying full of the sidebar width.

**Before**
Screenshot 2025-03-11 at 2 11 45 AM

**After**
Occupies full width
## Steps to test or reproduce

1. Enable feature preview
2. Enable enhanced navigation
3. Verify that the room list now takes up the full width of the sidebar.
## Further comments

[CORE-1016](https://rocketchat.atlassian.net/browse/CORE-1016)

[CORE-1016]: https://rocketchat.atlassian.net/browse/CORE-1016?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/meteor/client/components/CustomScrollbars/VirtualizedScrollbars.tsx
+++ b/apps/meteor/client/components/CustomScrollbars/VirtualizedScrollbars.tsx
@@ -1,5 +1,6 @@
+import type { Box } from '@rocket.chat/fuselage';
 import { useOverlayScrollbars } from 'overlayscrollbars-react';
-import type { HTMLAttributes, ReactElement } from 'react';
+import type { ComponentProps, ReactElement } from 'react';
 import { useEffect, useState, useRef, cloneElement, forwardRef, memo } from 'react';
 
 import BaseScrollbars, { getScrollbarsOptions } from './BaseScrollbars';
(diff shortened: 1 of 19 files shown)


abhinavkrin:
I have not added changeset, since the issue is not present in the previous version.

apps/meteor/client/sidebarv2/RoomList/RoomList.tsx
```diff
@@ -49,7 +49,7 @@ const RoomList = () => {
 
 	return (
 		<Box position='relative' display='flex' overflow='hidden' height='full' flexGrow={1} flexShrink={1} flexBasis='auto' ref={ref}>
-			<VirtualizedScrollbars>
+			<VirtualizedScrollbars width='full'>
```
dougfabris, line 52: instead of adding this prop, we should check if its parent needs to have all these attributes, mainly `display: flex`

apps/meteor/tests/e2e/sidebar.spec.ts
dougfabris, line 1: I don't think this test is necessary and worth it, since its only a visual thing

apps/meteor/client/components/CustomScrollbars/VirtualizedScrollbars.tsx
dougfabris, line 1: I don't think these changes are necessary

dougfabris (changes requested):
Its not a fix since this bug wasn't shipped in any version. Let's use regression for that!

dougfabris (approved):
Thanks for the very quick fix 👍

PR #35455, opened 2025-03-10
chore: Livechat department incorrect param name

aleksandernsilva:


sampaiodiego (approved):

sampaiodiego (changes requested):
this is actually a bug fix right?

aleksandernsilva:
@sampaiodiego I don't think this is considered a bugfix because there was nothing breaking before this change. 
The ajv schema accepted additional properties and the code was already using the correct param, so there was no bug. 
This change is more about correcting the incorrect param name, but won't affect functionality.

sampaiodiego:
oh thx @aleksandernsilva , I didn't check all `ajv` schema so I missed the `additionalProperties: true`.. thx ❤️

sampaiodiego (approved):

Issue #35457, opened 2025-03-11
Feat: message box should have ordered list feature like other chat apps.


priyanshuharshbodhi1:
### Description:

message box should have ordered list feature like other chat apps.

### Steps to reproduce:

1. 
2. 
3. 

1. press 1.(then space) in message box
2. the ordered list functionality dont come up.

### Server Setup Information:

- Version of Rocket.Chat Server: 
- License Type:
- Number of Users: 
- Operating System: 
- Deployment Method: 
- Number of Running Instances: 
- DB Replicaset Oplog: 
- NodeJS Version: 
- MongoDB Version:

### Client Setup Information

- Desktop App or Browser Version:
- Operating System:

MartinSchoeler:
Moving this to correct repo

Issue #35292, opened 2025-02-21
bug: bold formatting breaks when using *text:*


Ansh-dS:
### Description  
When trying to make `*yourMessage:*` bold, it does not render as expected because of the colon (`:`) at the end. However, if a space is added after the colon (`*yourMessage: *`), the bold formatting works correctly.  

Since I frequently use the `yourMessage:` syntax for headings , resolving this issue would improve the experience.  

### Steps to Reproduce  

1. Open any **group, channel, or direct message**.  
2. Send the following messages to yourself:  
   - `*yourMessage:*`  
   - `*yourMessage: *`  
3. Observe the difference in formatting.  

### Expected Behavior  
The text should become **bold** regardless of whether there is a space after the colon. There should be no need to add a space for Markdown bold formatting to work properly.   

### Actual behavior:

https://github.com/user-attachments/assets/fdaa0b2b-7b68-4763-818a-5ffbe135ef76

### Server/Client Information

1. **Server:**
   1. **Operating System:** Windows 11 (using WSL Terminal)
   2. **Version of Rocket.Chat Server:** 7.4.0‑develop
   3. **NodeJS Version:** 22.11.0
   4. **Yarn Version:** 4.6.0

2. **Client:**
   1. **Client Type:** Browser (Microsoft Edge)
   2. **Access URL:** [http://localhost:3000](http://localhost:3000)

3. **Additional:**
   1. **WSL Terminal:** Used to install dependencies

reetp:
There are a lot of issues around formatting, I believe largely due to the libraries used.

Worth a search for words like bold, italic etc

I am not sure this will be accepted but have labelled it.

milton-rucks:
It's a limitation in our parser. I'm glad if somebody could solve it.

PR #35458, opened 2025-03-11
Feat/ordered list functionality in message box #35457

priyanshuharshbodhi1:
## Issue(s)

#35457


Ansh-dS (changes requested):
forget to remove console.log() in line 302

Ansh-dS:
@priyanshuharshbodhi1 I am not sure, but I think you have implemented it in a wrong way, you can check the issue #35292 and compare my pr with the pr merged. I also have make changes in the UI file but we need to change how text parser works. second thing I want to mention, whether core developers said "yes" this new feature.

PR #35477, opened 2025-03-12
test: flaky mark-unread

juliajforesti:

Trace from playwright showed the `page#2` closed before `sendMessage`'s  success.
Adding an `expect` to ensure message was sent before closing page should solve it.

Trace screenshots:
image
image


juliajforesti:
fixed here #35646

PR #35486, opened 2025-03-13
feat(apps-engine): adds executeLivechatRoomCreatePrevent event

alfredodelfabro:

An App needs to be able to prevent the creation of a room. This is only possible for DMs and channels, not Livechat rooms.

This PR adds a hook called executePreLivechatRoomCreatePrevent to prevent the creation of a Livechat room.

## Further comments

Currently, the end-to-end tests rely on a test application ([here](https://github.com/RocketChat/Apps.RocketChat.Tester)). We modified this application to run the end-to-end tests ([here](https://github.com/RocketChat/Apps.RocketChat.Tester/tree/test-new-livechat-event)). So, before merging this PR, we must ensure that the master branch of this app incorporates the changes. Then, we need to update the reference file in this PR (insert-apps.ts) to point to the app file on the master branch.

[CTZ-3]

[CTZ-3]: https://rocketchat.atlassian.net/browse/CTZ-3?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/meteor/app/livechat/server/lib/rooms.ts
```diff
@@ -92,6 +101,43 @@ export async function createRoom({
...
+			username: visitor.username,
+			name: visitor.name,
+		},
+		ts: now,
+		livechatData: visitor.livechatData,
+		customFields: extraData?.customFields,
+	};
+
+	try {
+		await Apps.self?.triggerEvent(AppEvents.IPreLivechatRoomCreatePrevent, roomProps);
```
sampaiodiego, line 132: apparently there isn't an standard way of doing this, but since we allow the event just returning a `boolean` (without throwing an exception) we should look for that scenario as well:
```suggestion
		const prevent = await Apps.self?.triggerEvent(AppEvents.IPreLivechatRoomCreatePrevent, roomProps);
		if (prevent) {
			throw new Meteor.Error('error-app-prevented', 'A Rocket.Chat App prevented the room creation.');
		}
```

packages/apps-engine/src/server/managers/AppListenerManager.ts
```diff
@@ -1058,6 +1064,22 @@ export class AppListenerManager {
...
+            const app = this.manager.getOneById(appId);
+
+            prevented = (await app.call(AppMethod.EXECUTE_PRE_LIVECHAT_ROOM_CREATE_PREVENT, data)) as boolean;
+
+            if (prevented) {
+                return prevented;
+            }
+        }
+
+        return prevented;
```
sampaiodiego, line 1080: although this is very similar to other code doing same stuff, I'd suggest simplify it to something like:

```suggestion
        for (const appId of this.listeners.get(AppInterface.IPreLivechatRoomCreatePrevent)) {
            const app = this.manager.getOneById(appId);

            if (await app.call(AppMethod.EXECUTE_PRE_LIVECHAT_ROOM_CREATE_PREVENT, data)) {
                return true;
            }
        }

        return false;
```

sampaiodiego (changes requested):

sampaiodiego (approved):

packages/apps-engine/src/definition/livechat/IPreLivechatRoomCreatePrevent.ts
```diff
@@ -0,0 +1,24 @@
+import { IRead, IHttp, IPersistence } from "../accessors";
+import { AppMethod } from "../metadata";
+import { ILivechatRoom } from "./ILivechatRoom";
+
+/**
+ * Handler called before a livechat room is created.
```
d-gubert, line 6:
```suggestion
 * Handler called before a livechat room is created.
 *
 * To prevent the room from being created, the app should throw an `AppsEngineException`
```

d-gubert (dismissed):

sampaiodiego (approved):

sampaiodiego (approved):

d-gubert (approved):

PR #35488, opened 2025-03-13
refactor: Remove meteor functions from canned responses

rique223:

Refactor canned responses initialization by replacing Meteor.startup and Tracker.autorun with React hooks. Split functionality into useFetchCannedResponses for API data fetching with TanStack Query and useCannedResponsesStream for WebSocket updates. Added both hooks to AppLayout and removed legacy imports, improving maintainability while preserving original functionality.

Demo video:
https://github.com/user-attachments/assets/6066bab2-2eb5-4801-9227-c9c84b9a9634

## Steps to test or reproduce

Pre-requisites:
- Admin user with the livechat-manager role;
- Normal user with the livechat-agent role;
- Omnichannel enabled for receiving messages;

1. Login as both your admin and normal user on different sessions;
2. Open your browser's devtools; 
3. Check the Network tab for the `/v1/canned-responses.get` endpoint;
4. Check the Network tab for the initialization of the `canned-response` websocket stream;
5. Go to Administration > Omnichannel > Canned Responses and create your first canned response (remember to set it to public);
6. Check your Network tab for the `changed` canned response websocket message;
7. Create a new livechat and send the canned response using the shortcut you've defined;
8. Go back to the canned responses page and edit your canned response content;
9. Check your Network tab for the `changed` canned response websocket message;
10. Re-send the canned response on livechat;
11. Delete the canned response;
12. Check your Network tab for the `removed` canned response websocket message.

## Further comments

Jira task: [ARCH-1481](https://rocketchat.atlassian.net/browse/ARCH-1481)

[ARCH-1481]: https://rocketchat.atlassian.net/browse/ARCH-1481?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/meteor/client/views/root/hooks/useCannedResponsesStream.ts
```diff
@@ -0,0 +1,41 @@
...
+
+import { useCanAccessCannedResponses } from './useCanAccessCannedResponses';
+import { CannedResponse } from '../../../../app/canned-responses/client/collections/CannedResponse';
+
+export const useCannedResponsesStream = () => {
+	const uid = useUserId();
+	const handleStreamCannedResponses = useStream('canned-responses');
+	const canAccess = useCanAccessCannedResponses();
+
+	useEffect(() => {
```
ggazzo, line 12: missing `canned-responses.get` call to initialize

ggazzo (changes requested):
check comments

apps/meteor/client/views/root/AppLayout.tsx
```diff
@@ -52,8 +54,9 @@ const AppLayout = () => {
 	useTokenPassAuth();
 	useCustomOAuth();
 	useCorsSSLConfig();
-
 	useOTRMessaging();
+	useFetchCannedResponses();
+	useCannedResponsesStream();
```
ggazzo, line 59: please merge both hooks I dont think we gonna use them separated

apps/meteor/client/views/root/hooks/useFetchCannedResponses.ts
```diff
@@ -0,0 +1,21 @@
...
+import { useCanAccessCannedResponses } from './useCanAccessCannedResponses';
+import { CannedResponse } from '../../../../app/canned-responses/client/collections/CannedResponse';
+
+export const useFetchCannedResponses = () => {
+	const getCannedResponses = useEndpoint('GET', '/v1/canned-responses.get');
+	const isEnabled = useCanAccessCannedResponses();
+
+	const handleCannedResponses = async () => {
+		const { responses } = await getCannedResponses();
+		responses.forEach((response) => CannedResponse.insert(response));
```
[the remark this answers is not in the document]
rique223: onSuccess was deprecated on tanstack query V5

apps/meteor/client/views/root/AppLayout.tsx
rique223: We may not use them separately, but having those two things inside the same useEffect is a separation of concerns issue. Also I can see a future in which we may use `useFetchCannedResponses` for other parts of the system.

apps/meteor/client/views/root/hooks/useFetchCannedResponses.ts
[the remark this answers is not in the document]
ggazzo: ok I'm getting old, v5 does not recommend `onSuccess` anymore, thanks for letting me know :)

 but if you want to reuse this function in near future you should not perform mutations in global collections inside the result.

you can do instead:

```

const result = useFetchCannedResponses(...);


  useEffect(() => {
    if (result.isSuccess) {
	result.responses.forEach((response) => CannedResponse.insert(response));
    }
  }, [result.isSuccess]);
```

in that way you make sure the hook is actually usable multiple times, without unexpected actions.

also pay attention if this query is invalidated somehow, CannedResponse.insert probably will try to re-add some items and it will throw an error because of duplicated key

apps/meteor/client/views/root/AppLayout.tsx
ggazzo: ok, so please create an `useCannedResponse` that uses `useFetchCannedResponse` gets the data result and adds to the Collection, meanwhile starts a stream, but the useStream alone is not useful.

I never said to put both them inside the same **useEffect**, but inside the useCannedResponsesStream something like`useCannedResponse`

notice that both hooks cause a mutation on the same collection, so both share one same concern: fulfill the CannedResponse Collection.

if somewhere people wants to use the `useFetchCannedResponses` to handle the data from the server they should not mutate the Cached collection again

apps/meteor/client/views/room/providers/ComposerPopupProvider.tsx
```diff
@@ -8,11 +8,12 @@ import type { ReactNode } from 'react';
 import { useTranslation } from 'react-i18next';
 
 import { hasAtLeastOnePermission } from '../../../../app/authorization/client';
-import { CannedResponse } from '../../../../app/canned-responses/client/collections/CannedResponse';
 import { emoji } from '../../../../app/emoji/client';
 import { Subscriptions } from '../../../../app/models/client';
 import { usersFromRoomMessages } from '../../../../app/ui-message/client/popup/messagePopupConfig';
 import { slashCommands } from '../../../../app/utils/client';
+import { queryClient } from '../../../lib/queryClient';
```
ggazzo, line 15: you should use useQueryClient

PR #35497, opened 2025-03-14
fix: Add Apps log TTL index

ggazzo:
**Refactor AppsLogsModel to remove `_updatedAtIndexOptions` and use `modelIndexes`**

This change removes the `_updatedAtIndexOptions` parameter and instead uses the `modelIndexes` method to determine the updated index options.

Without overwriting `modelIndexes` to return a value, `_updatedAtIndexOptions` is never added.

The current implementation check on `Apps.isInitialized()` which is `false` during the first attempt, but you can seethe problem  when the TTL setting is changed.

### DIFF
--- a/.changeset/fifty-moles-boil.md
+++ b/.changeset/fifty-moles-boil.md
@@ -4,4 +4,4 @@
 "@rocket.chat/models": patch
 ---
 
-Fixes an issue where the app log ttl were being ignored, also set to be always 30 days 
+Fixes an issue where the app's logs index was not being created by default sometimes, also set to be always 30 days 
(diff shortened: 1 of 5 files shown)


d-gubert (dismissed):

.changeset/fifty-moles-boil.md
```diff
@@ -0,0 +1,7 @@
+---
+"@rocket.chat/meteor": patch
+"@rocket.chat/model-typings": patch
+"@rocket.chat/models": patch
+---
+
+Fixes an issue where the app log ttl were being ignored, also set to be always 30 days
```
sampaiodiego, line 7:
```suggestion
Fixes an issue where the app's logs index was not being created by default, also set to be always 30 days 
```

sampaiodiego (changes requested):

sampaiodiego (approved):

PR #35544, opened 2025-03-18
refactor: Remove meteor functions from WebRTCClass stream

rique223:

Transformed the stream initialization function from the `WebRTCClass.ts` file into a custom hook named `useWebRTC.ts`, also removed instances of Meteor.userId for useUserId, removed Meteor.startup and Tracker.autorun in favor of a custom hook with a useEffect and changed sdk.stream for the useStream hook.
## Further comments

Jira task: [ARCH-1493](https://rocketchat.atlassian.net/browse/ARCH-1493)

[ARCH-1493]: https://rocketchat.atlassian.net/browse/ARCH-1493?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/meteor/client/views/root/hooks/useWebRTC.ts
```diff
@@ -0,0 +1,37 @@
...
+			};
+
+			(handleWebRTCTypeObject[type] || handleWebRTCTypeObject.default)();
+		};
+
+		const handleStopNotifyUser = notifyUser(`${uid}/webrtc`, handleNotifyUser);
+
+		return () => {
+			handleStopNotifyUser();
+		};
```
gabriellsh, line 35:
```suggestion
		return notifyUser(`${uid}/webrtc`, handleNotifyUser);
```

apps/meteor/client/views/root/hooks/useWebRTC.ts
```diff
@@ -0,0 +1,37 @@
...
+			if (!webrtc) return;
+
+			const handleWebRTCTypeObject = {
+				candidate: () => webrtc.onUserStream('candidate', data as CandidateData),
+				description: () => webrtc.onUserStream('description', data as DescriptionData),
+				join: () => webrtc.onUserStream('join', data as JoinData),
+				default: () => console.warn(`WebRTC: Received unexpected event type: ${type}`),
+			};
+
+			(handleWebRTCTypeObject[type] || handleWebRTCTypeObject.default)();
```
gabriellsh, line 28: Can we go back to using a switch statement here? IMO it's cleaner than declaring an object. Nice catch on the default btw!

apps/meteor/client/views/root/hooks/useWebRTC.ts
gabriellsh: Also any reason that you stopped using the constant `WEB_RTC_EVENTS.WEB_RTC` here? I think WEBRTC will probably/is deprecated, but if we ever change the event name we just have tu update the constants.

gabriellsh (changes requested):

apps/meteor/client/views/root/hooks/useWebRTC.ts
rique223: Just a small oversight, was having some problems with typing and at some point removed the constant to see if it fixed the issue but forgot to add it back when I found the actual culprit.

ggazzo (approved):

PR #35545, opened 2025-03-18
fix: Video conference call again not working properly

dougfabris:

Introduced here: #35047

Due the refactor we introduced in `7.4`, `useVideoConfLoadCapabilities` are losing the `VideoConfManager` reference so when the user tries to call again it fails

## Steps to test or reproduce

- Configure video conference feature
- Initiate a call in a direct message
- When trying to call again should work normally

## Further comments

[SUP-752]

[SUP-752]: https://rocketchat.atlassian.net/browse/SUP-752?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


yash-rajpal (approved):
LGTM!

PR #35573, opened 2025-03-20
chore: set yarn nmMode to hardlinks-local

cardoso:


cardoso:
Actually need to look into some other stuff first before trying this.

PR #35580, opened 2025-03-21
fix: Update Contact's custom fields when updating visitor custom fields

lucas-a-pelegrino:

This PR aims to increment the logic of `.setCustomFields()`, updating the `customFields` from the visitor's `Contact` entity when applicable.

When `Contact.customFields` is nullish and/or `overwrite` is `true`, it inserts the customFields accordingly. If `Contact.customFields` is not nullish and overwrite is `false`, it will save it as a conflict.
## Issue(s)

[CTZ-17](https://rocketchat.atlassian.net/browse/CTZ-17)

## Steps to test or reproduce

1. Open Livechat widget
2. Call setCustomField API method
3. Start a new chat
4. call setCustomField again with a different value

## Further comments

[CTZ-17]: https://rocketchat.atlassian.net/browse/CTZ-17?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

---

This pull request introduces enhancements and bug fixes related to the handling of custom fields in the Rocket.Chat Livechat module. The key changes include:

1. **Functionality Update**: The `setCustomFields` function is updated to correctly update the Context's `customFields` property, ensuring that custom fields are managed accurately.

2. **Livechat Bridge Enhancement**: Implementation for setting custom fields is added to the `AppLivechatBridge` class, with a note for future refactoring.

3. **Visitor Registration Enhancement**: The visitor registration endpoint (`POST livechat/visitor`) is enhanced to update custom fields for associated Livechat contacts when a visitor's custom fields are updated. This is achieved by introducing the `LivechatContacts` model and using the `updateContactsCustomFields` function to perform updates in parallel.

4. **Custom Fields Refactoring**: The `setCustomFields` function is refactored to handle updates for associated livechat contacts when the scope is not 'room'. A new helper function, `updateContactsCustomFields`, is introduced to manage update logic, including handling overwrites and conflicts. This involves fetching the visitor, identifying associated contacts, and updating them concurrently.

5. **Error Handling Removal**: Error handling for custom field operations is removed, which may potentially hide failures during these operations.

These changes aim to improve the robustness and functionality of custom field management within the Livechat module.


KevLehman (changes requested):
Can u add a changeset?

apps/meteor/app/apps/server/bridges/livechat.ts
```diff
@@ -367,6 +367,10 @@ export class AppLivechatBridge extends LivechatBridge {
 	): Promise<number> {
 		this.orch.debugLog(`The App ${appId} is setting livechat visitor's custom fields.`);
 
-		return LivechatTyped.setCustomFields(data);
+		await LivechatTyped.setCustomFields(data);
+
+		// Return a placeholder number as the base method expects a number.
+		// TODO: Refactor this logic to stop using number and handle void or errors thrown by .setCustomFields().
```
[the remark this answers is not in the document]
lucas-a-pelegrino: Yes, we should have a task created for this, I'll create this task and link it here.

apps/meteor/app/livechat/server/api/v1/customField.ts
```diff
@@ -12,13 +12,14 @@ API.v1.addRoute(
...
 			const guest = await findGuest(token);
 			if (!guest) {
 				throw new Error('invalid-token');
 			}
 
-			if (!(await Livechat.setCustomFields({ token, key, value, overwrite }))) {
+			try {
+				await Livechat.setCustomFields({ token, key, value, overwrite });
+			} catch {
 				return API.v1.failure();
```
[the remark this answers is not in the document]
lucas-a-pelegrino: My goal here was to keep it as similar to the old behavior as possible. But I think in this case, it would be ok to let api handle the errors.

apps/meteor/app/apps/server/bridges/livechat.ts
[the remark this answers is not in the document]
lucas-a-pelegrino: @KevLehman [task](https://rocketchat.atlassian.net/browse/CTZ-41) created.

apps/meteor/app/livechat/server/api/v1/visitor.ts
```diff
@@ -93,6 +93,8 @@ API.v1.addRoute(
 				).toArray();
 				validateRequiredCustomFields(keys, livechatCustomFields);
 
+				const visitorHasContacts = await LivechatContacts.findAllByVisitorId(visitor._id).toArray();
```
KevLehman, line 96:
```suggestion
				const visitorContacts = await LivechatContacts.findAllByVisitorId(visitor._id).toArray();
```

apps/meteor/app/livechat/server/api/v1/visitor.ts
```diff
@@ -107,6 +109,12 @@ API.v1.addRoute(
 							errors.push(key);
 						}
 
+						if (visitorHasContacts.length > 0) {
```
KevLehman, line 112:
```suggestion
						if (visitorContacts.length > 0) {
```

apps/meteor/app/livechat/server/api/v1/visitor.ts
```diff
@@ -107,6 +109,12 @@ API.v1.addRoute(
...
+							await Promise.all(
+								visitorHasContacts.map((contact: ILivechatContact) => updateContactsCustomFields(contact, key, value, overwrite)),
```
KevLehman, line 114:
```suggestion
								visitorContacts.map((contact: ILivechatContact) => updateContactsCustomFields(contact, key, value, overwrite)),
```

apps/meteor/app/livechat/server/lib/custom-fields.ts
```diff
@@ -19,7 +19,33 @@ export const validateRequiredCustomFields = (customFields: string[], livechatCus
...
 };
 
-export async function setCustomFields({ token, key, value, overwrite }: { key: string; value: string; overwrite: boolean; token: string }) {
+export async function updateContactsCustomFields(contact: ILivechatContact, key: string, value: string, overwrite: boolean): Promise<void> {
+	if (overwrite) {
+		contact.customFields ??= {};
+		contact.customFields[key] = value;
+	} else {
+		contact.conflictingFields ??= [];
+		contact.conflictingFields.push({ field: `customFields.${key}`, value });
```
KevLehman, line 28: To my understanding: if the function receives override:false it will assign all custom fields as conflicting fields for the visitor, even if the visitor has no customfields at all?

KevLehman: If that's true, and there's no more decision being done, think you can construct the query in a way you insert all the things without having to fetch the contact and then update 1b1, like:

```
if (overwrite) 
Object.keys(obj).map((k) => update.$set.customFields[k] = obj[k])
else
update.$addToSet = { conflictingFields: { field, value } }

and then LivechatContacts.updateByVisitorId(visitorId, updates)
```

Do you think that would work?

KevLehman: (Ideally the custom fields shouldn't be processed 1b1, but at least this way its still N and not N * X lol)

apps/meteor/app/livechat/server/lib/custom-fields.ts
```diff
@@ -34,17 +60,19 @@ export async function setCustomFields({ token, key, value, overwrite }: { key: s
...
-		return 0;
-	}
+		const visitor = await LivechatVisitors.getVisitorByToken(token, { projection: { _id: 1 } });
+		if (!visitor) {
+			throw new Error(`Visitor with token "${token}" not found.`);
+		}
 
-	return result.modifiedCount;
+		const result = await LivechatContacts.findAllByVisitorId(visitor._id).toArray();
+		if (result) {
```
KevLehman, line 74: Same as above

KevLehman (changes requested):

.changeset/young-kiwis-fly.md
```diff
@@ -0,0 +1,5 @@
+---
+"@rocket.chat/meteor": patch
```
sampaiodiego, line 2: the PR title says `feat` which means it is a new feature, meaning the changeset should be a `minor` but instead it is a `patch`.. is this a feature or a fix?

apps/meteor/app/livechat/server/api/v1/visitor.ts
```diff
@@ -107,6 +107,11 @@ API.v1.addRoute(
 							errors.push(key);
 						}
 
+						const contacts = await LivechatContacts.findAllByVisitorId(visitor._id).toArray();
```
sampaiodiego, line 110: having this logic here duplicated here is bad, we should unify the codes to prevent changes not being applied in both places in the future.. we can have a task to improve this later

it is duplicated from `setCustomFields` function:
https://github.com/RocketChat/Rocket.Chat/blob/1177d65ad979003c424b06e93af63ff4f64caa5f/apps/meteor/app/livechat/server/lib/custom-fields.ts#L36

sampaiodiego (changes requested):

sampaiodiego (approved):

PR #35594, opened 2025-03-24
fix: App update with no permission causes invalid_installation_disabled

Gustrb:

If an app is installed with permissions, but is updated with a version that requires no permissions (with no permissions in the manifest file), the app is marked as invalid_installation_disabled. This was happening because when we update the app's metadata in the database, we would do a simple update (no unset) which in term, would make that field different than the one in the manifest, which would case an error when comparing the checksums.

## Steps to test or reproduce

1. Create an app with permissions (doesn't matter which, I've used user.read for testing)
2. Build and upload the app as a private app
3. Change the app's manifest (app.json) removing the permissions field from the json
4. Build and upload the app as a private app.

Expected: The app should be upload without any big issue
Actual: The app was being marked as invalid_installation_disabled and never got enabled

## Further comments

[CONN-564](https://rocketchat.atlassian.net/browse/CONN-564)

[CONN-564]: https://rocketchat.atlassian.net/browse/CONN-564?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

---

This pull request addresses an issue in the Rocket.Chat repository where removing app permissions results in the app never being enabled, leading to an `invalid_installation_disabled` error. The changes are documented in the changeset file `.changeset/shy-moles-boil.md`, although the documentation is noted to be clear, it lacks specific technical details. The fix is implemented in the `fix/app-update-with-no-permission-error` branch and is intended to be merged into the `develop` branch.


apps/meteor/ee/server/apps/storage/AppRealStorage.ts
```diff
@@ -46,7 +47,18 @@ export class AppRealStorage extends AppMetadataStorage {
...
+		};
+
+		// Note: This is really important, since we currently store the permissionsGranted as null if none are present
+		//       in the App's manifest. So, if there was a permissionGranted and it was removed, we must see the app as having
+		//       no permissionsGranted at all (which means default permissions). So we must actively unset the field.
+		if (!item.permissionsGranted) {
+			updateQuery.$unset = { permissionsGranted: 1 };
+		}
+
+		await this.db.updateOne({ id: item.id, _id: item._id }, updateQuery);
```
d-gubert, line 61: Can we replace the `updateOne` method call with a `findOneAndUpdate`? Then we can save on DB network roundtrip, removing the query below

d-gubert (changes requested):

apps/meteor/ee/server/apps/storage/AppRealStorage.ts
Gustrb: yep

.changeset/shy-moles-boil.md
```diff
@@ -0,0 +1,5 @@
+---
+"@rocket.chat/meteor": patch
+---
+
+fixes an issue where removing the permissions of an app would cause it to never be enabled.
```
d-gubert, line 5:
```suggestion
Fixes an issue where removing the permissions of an app would make it impossible to enable it after update
```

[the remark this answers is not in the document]
d-gubert: @kody changeset files are not really meant to be comprehensive documentation, they should be short and concise.

d-gubert (approved):

PR #35602, opened 2025-03-24
regression: We were not keeping track of the blocks if through BlockBuilder

Gustrb:
## Issue(s)

Introduced in #35120. This was breaking the Poll App, this is why it was noticed.

## Further comments

[CONN-569](https://rocketchat.atlassian.net/browse/CONN-569)


ggazzo (changes requested):
I'm sorry to ask, but isn't this a fix? It seems the bug was introduced on 7.4

Gustrb:
Yea @ggazzo it is a fix, just not sure about how should we proceed with this one, since it is an issue that is affecting the apps that make use of it. Should we have it as a fix and backport it later or should we already pack it with this release?

PR #35610, opened 2025-03-25
chore: Remove meteor functions from `messagePopupConfig.ts`

rique223:

Deleted the messagePopupConfig file, created a custom hook called useRoomMessageUsers that fetches the same list of users but using react query and re-did the duplicate logic using map, refactored the ComposerPopupProvider to use the new query as a source of thruth instead of the old collection, re-did the logic using array functions instead of a mongo querry and added the new query key to the queryKeys file.

## Further comments

Jira task: [ARCH-1492](https://rocketchat.atlassian.net/browse/ARCH-1492)


apps/meteor/client/views/root/hooks/useRoomMessageUsers.ts
```diff
@@ -0,0 +1,77 @@
...
+		return Array.from(uniqueUsers.values());
+	}, [queryParams.rid, queryParams.username, rid]);
+
+	const { data: roomMessageUsers = [] } = useQuery<RoomMessageUser[]>({
+		queryKey: roomMessageUsersQueryKeys.all(rid, uid),
+		queryFn: fetchRoomMessageUsers,
+		enabled: Boolean(rid && uid),
+	});
+
+	return roomMessageUsers;
```
gabriellsh, line 76: Can we return the query result instead of the data?

apps/meteor/client/views/root/hooks/useRoomMessageUsers.ts
```diff
@@ -0,0 +1,77 @@
...
+		if (!queryParams.rid || !queryParams.username) {
+			return [];
+		}
+
+		const messages = Messages.find(
+			{
+				rid,
+				'u.username': { $ne: queryParams.username },
+				't': { $exists: false },
+				'ts': { $exists: true },
```
gabriellsh, line 40: This filter was not on the other Query. Can you explain why did you add it here?

apps/meteor/client/views/room/providers/ComposerPopupProvider.tsx
```diff
@@ -64,14 +67,16 @@ const ComposerPopupProvider = ({ children, room }: ComposerPopupProviderProps) =
 					const filterRegex = filter && new RegExp(escapeRegExp(filter), 'i');
 					const items: ComposerBoxPopupUserProps[] = [];
 
-					const users = Array.from(usersFromRoomMessages.values())
-						.filter((u) => !!u.ts && (filterRegex ? u.username.match(filterRegex) || u.name?.match(filterRegex) : true))
```
gabriellsh, line 68: where did `!!u.ts` go?

apps/meteor/client/views/room/providers/ComposerPopupProvider.tsx
```diff
@@ -64,14 +67,16 @@ const ComposerPopupProvider = ({ children, room }: ComposerPopupProviderProps) =
...
+					const roomMessageUsers = queryClient.getQueryData<RoomMessageUser[]>(roomMessageUsersQueryKeys.all(rid, uid)) ?? [];
+
+					roomMessageUsers
 						.sort((a, b) => b.ts.getTime() - a.ts.getTime())
-						.slice(0, suggestionsCount ?? 5)
-						.map((u) => {
-							u.suggestion = true;
-							return u;
-						});
```
gabriellsh, line 74: Where did this map go?

apps/meteor/client/views/room/providers/ComposerPopupProvider.tsx
```diff
@@ -64,14 +67,16 @@ const ComposerPopupProvider = ({ children, room }: ComposerPopupProviderProps) =
...
+						.filter((u) => {
+							if (!filterRegex) return true;
+							return filterRegex.test(u.username) || (u.name && filterRegex.test(u.name));
+						})
+						.slice(0, suggestionsCount ?? 5);
```
gabriellsh, line 78: This is more of an exercise to improve this. Each operation here will loop through the array list some times. Can you think of a better order for these operations so we can do less work on the array?

apps/meteor/client/views/room/providers/ComposerPopupProvider.tsx
```diff
@@ -92,17 +97,20 @@ const ComposerPopupProvider = ({ children, room }: ComposerPopupProviderProps) =
 						});
 					}
 
-					return [...users, ...items];
+					return [...roomMessageUsers, ...items];
 				},
 				getItemsFromServer: async (filter: string) => {
 					const filterRegex = filter && new RegExp(escapeRegExp(filter), 'i');
-					const usernames = Array.from(usersFromRoomMessages.values())
-						.filter((u) => !!u.ts && (filterRegex ? u.username.match(filterRegex) || u.name?.match(filterRegex) : true))
```
gabriellsh, line 100: where did `!!u.ts` go?

apps/meteor/client/views/room/providers/ComposerPopupProvider.tsx
```diff
@@ -92,17 +97,20 @@ const ComposerPopupProvider = ({ children, room }: ComposerPopupProviderProps) =
...
 						.sort((a, b) => b.ts.getTime() - a.ts.getTime())
+						.filter((u) => {
+							if (!filterRegex) return true;
+							return filterRegex.test(u.username) || (u.name && filterRegex.test(u.name));
+						})
 						.slice(0, suggestionsCount ?? 5)
-						.map((u) => {
-							return u.username;
-						});
+						.map((u) => u.username);
```
gabriellsh, line 112: Same exercise here for the operations.

gabriellsh (changes requested):

apps/meteor/client/views/root/hooks/useRoomMessageUsers.ts
rique223: I did this so I don't need to do a verification like this one on the `ComposerPopupProvider.tsx`:
```
.filter((u) => !!u.ts && (filterRegex ? u.username.match(filterRegex) || u.name?.match(filterRegex) : true))
```

This way the overall code is more readable and the array functions are shorter.

apps/meteor/client/views/room/providers/ComposerPopupProvider.tsx
rique223: Asnwered above, since the Mongo query now only returns messages with the ts field this verification is unnecessary.

apps/meteor/client/views/room/providers/ComposerPopupProvider.tsx
rique223: Answered above

apps/meteor/client/views/room/providers/ComposerPopupProvider.tsx
ggazzo: I think you didnt get the question, you are filtering/mapping/sorting, but you are not using the result

ggazzo (changes requested):

PR #35611, opened 2025-03-25
chore: Only public channels cannot be previewed by permissions

tiagoevanp:

Guarantee only public channels will face a `preview-public-channels` check.

## Further comments

[CORE-1024]

[CORE-1024]: https://rocketchat.atlassian.net/browse/CORE-1024?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


gabriellsh (changes requested):

gabriellsh (changes requested):
Can you add more tests to cover this specific case?

MarcosSpessatto (commented):
Isn't it a fix instead of a chore?

PR #35613, opened 2025-03-25
feat: Use `IconButton` in favor of `Tag` to navigate to parent room

dougfabris:


apps/meteor/client/views/room/HeaderV2/ParentRoom/ParentRoom.tsx
```diff
@@ -0,0 +1,22 @@
+import type { IRoom } from '@rocket.chat/core-typings';
+
+import ParentDiscussion from './ParentDiscussion';
+import ParentTeam from './ParentTeam';
+
+const ParentRoom = ({ room }: { room: IRoom }) => {
+	const parentRoomId = room.prid || (room.teamId && !room.teamMain);
```
juliajforesti, line 7: this looks more like a boolean, right?
seems odd for it to be either an id (`room.prid`) OR a boolean that indicates if it's a channel from a team.
i'd suggest 
```suggestion
	const hasParentRoom = Boolean(room.prid) || (room.teamId && !room.teamMain);
```

apps/meteor/client/views/room/HeaderV2/ParentRoom/ParentTeam.tsx
juliajforesti, line 1: If the user is not a member of the parent team, should the IconButton be displayed?

juliajforesti (changes requested):

apps/meteor/client/views/room/HeaderV2/ParentRoom/ParentTeam.tsx
dougfabris: Im using the exact same rule as we use to when using `Tag`

juliajforesti: Ok, not related to these changes so it won't block this PR, but I suggest discussing this later, as it seems like a "bug"

tassoevan (changes requested):
`apps/meteor/tests/e2e/utils/createTargetDiscussion.ts` is empty?

apps/meteor/client/views/room/HeaderV2/ParentRoom/ParentTeam.tsx
dougfabris: I already discussed with the design team and we won't display this if the user can't see the team, which means we should display if the team is public or if the user belongs to the team

juliajforesti (approved):

tassoevan (approved):
`apps/meteor/tests/e2e/utils/createTargetDiscussion.ts` is empty?

apps/meteor/tests/e2e/feature-preview.spec.ts
juliajforesti, line 1: I think it's missing a cleanup to delete `targetChannelWithTeam` and `targetDiscussion`

juliajforesti (changes requested):

juliajforesti (approved):

tassoevan (approved):

PR #35618, opened 2025-03-25
fix: Omnichannel making a request for room data after every message sent

sampaiodiego:

This endpoint is called every time a message is sent by an omnichannel agent. On every message we used update the inquiry's `lastMessage`, which caused an invalidation on the frontend so it mistakenly thought it would need to fetch the room data again (since something changed).

The proposal is that we update the inquiry's last message only if it has not been taken yet. Once it is taken we update only the room's last message property. If it happens to the room to be sent back to the queue, then we update the inquiry's last message so it can reflect the latest state. A test was added to cover this exact behavior.

This is chart of a production workspace that uses inquiries, showing how massive the amount of calls to the endpoint is:
image

## Issue(s)

[CTZ-28]

## Further comments

[CTZ-28]: https://rocketchat.atlassian.net/browse/CTZ-28?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.changeset/heavy-boats-mix.md
+++ b/.changeset/heavy-boats-mix.md
@@ -1,0 +1,6 @@
+---
+'@rocket.chat/models': patch
+'@rocket.chat/meteor': patch
+---
+
+Fixes an issue when sending a message on an omnichannel room would cause the web client to fetch the room data again.
--- a/apps/meteor/app/livechat/server/lib/RoutingManager.ts
+++ b/apps/meteor/app/livechat/server/lib/RoutingManager.ts
@@ -175,7 +175,7 @@ export const RoutingManager: Routing = {
 		const { servedBy } = room;
 
 		if (shouldQueue) {
-			const queuedInquiry = await LivechatInquiry.queueInquiry(inquiry._id);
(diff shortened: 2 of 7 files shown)


lucas-a-pelegrino (dismissed):

KevLehman (approved):

KevLehman (changes requested):
There's one missing place: https://github.com/RocketChat/Rocket.Chat/blob/6d571d42c9d48da291f989a3fbaf070614f5bfe5/apps/meteor/app/livechat/server/lib/Helper.ts#L636

Here, the chat is "transfered" but since there's no online agents, it ends up in the department's queue. We should do the same we do on the manual move to queue func.

KevLehman (dismissed):
There's a small change in behavior, where the last message of the inquiry after being sent to department's queue ends up being the system message instead of the actual "last message"

Think we can live with that, and fix in a separate PR (as the fix would involve some extra work on other files)

KevLehman (approved):

PR #35665, opened 2025-03-31
fix: federated users overwriting local display name with their usernames

pierre-lehnen-rc:
## Issue(s)

When we invite an user from a federated workspace, we only have their username available, not the display name. In this situation, synaptic tries to get the user's profile from the remote synaptic, who will not have the data ready if that user had never federated before. In this case synaptic falls back to using the username as display name and passes that information to us as if it were the target user's real display name.

When we pre-create the federated users on all sides, we'll already have their display name properly assigned, but this username coming from synaptic ends up overwriting it.

## Further comments

This is a hacky, temporary solution for the issue. The proper fix would be to implement responses to user queries from synaptic.


pierre-lehnen-rc:
Moved this change to a regular PR branch targetting develop:
#35716

PR #35668, opened 2025-03-31
feat: allow LDAP sync to pre-create federated users

pierre-lehnen-rc:


pierre-lehnen-rc:
moved this into two PRs:
#35717
#35718

PR #22767, opened 2021-07-26
Regression: Fix empty tag field

MartinSchoeler:

PR #35679, opened 2025-04-01
fix: GUI crash when editing canned response with tags via contextual bar

aleksandernsilva:
This PR fixes a crash that occurred when editing a canned response with tags via the contextual bar. 
The issue was caused by an incorrect tag format being passed to the form as an array of objects instead of the expected array of strings.

## Issue(s)
[SUP-756](https://rocketchat.atlassian.net/browse/SUP-756)

## Steps to test or reproduce
- Access workspace
- Get any omnichannel chat;
- Click on “Canned responses” in the room toolbox
- Click on a canned response with a tag
- Click on Edit.

## Further comments
Introduced here #22767
Partially fixed here #29694


tiagoevanp (changes requested):
All great, I'm just concerned about the specificity of some locators in test. Role label + hasText is a possible future test leak between screens and functionalities...

KevLehman (approved):

MartinSchoeler (approved):

PR #35696, opened 2025-04-03
fix: dont make all instances unavaible due to fake timeout in transcr…

debdutdeb:
…ipt queue


KevLehman (commented):
Would require some testing but codewise looks good

PR #35716, opened 2025-04-07
fix: federated users overwriting local display name with their username

pierre-lehnen-rc:
## Issue(s)

When we invite an user from a federated workspace, we only have their username available, not the display name. In this situation, synaptic tries to get the user's profile from the remote synaptic, who will not have the data ready if that user had never federated before. In this case synaptic falls back to using the username as display name and passes that information to us as if it were the target user's real display name.

When we pre-create the federated users on all sides, we'll already have their display name properly assigned, but this username coming from synaptic ends up overwriting it.

## Further comments

[CORE-1071]

[CORE-1071]: https://rocketchat.atlassian.net/browse/CORE-1071?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

---

This pull request addresses an issue in the Rocket.Chat repository where federated users' display names were being overwritten by their usernames. The changes include a more detailed implementation of the `shouldUpdateDisplayName` method in the `FederatedUser` class, which now includes specific conditions to determine when a user's display name should be updated. This fix ensures that the display names of federated users are preserved appropriately. The changes are proposed from the `fix/federation-overwrite-display-name` branch to the `develop` branch.


.changeset/tall-hornets-live.md
```diff
@@ -0,0 +1,5 @@
+---
+'@rocket.chat/meteor': patch
+---
+
+fixes an issue where a federated user's display name would be overwritten by its username
```
debdutdeb, line 5:
```suggestion
Fixes an issue where a federated user's display name would be overwritten by its username
```

debdutdeb (approved):
ok to approve, just one question mostly but don't think that'll be a big deal.

debdutdeb (approved):

MarcosSpessatto (approved):

PR #35736, opened 2025-04-08
fix: Unhandled rejection when removing an agent from a department when multiple business hours are enabled

KevLehman:
## Issue(s)

https://rocketchat.atlassian.net/browse/CTZ-43


apps/meteor/ee/app/livechat-enterprise/server/business-hour/Multiple.ts
```diff
@@ -154,9 +154,10 @@ export class MultipleBusinessHoursBehavior extends AbstractBusinessHourBehavior
 		const department = await LivechatDepartment.findOneById<Pick<ILivechatDepartment, 'businessHourId'>>(departmentId, {
 			projection: { businessHourId: 1 },
 		});
-		if (!department || !agentsId.length) {
+		if (!department?.businessHourId || !agentsId.length) {
```
sampaiodiego, line 157: should this be handled by `handleRemoveAgentsFromDepartments`? it looks like it partially does already, but more importantly it seems it does other things that doesn't depend on `businessHourId`

PR #35737, opened 2025-04-08
refactor: remove emoji-custom from meteor

juliajforesti:
[ARCH-1545](https://rocketchat.atlassian.net/browse/ARCH-1545)

## Further comments

[ARCH-1545]: https://rocketchat.atlassian.net/browse/ARCH-1545?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


gabriellsh (changes requested):
For some reason the first time I enter the page, the custom emojis are not loaded. This by itself is it's own issue, as a page refresh is necessary, and apparently fixes the issue.

The other issue can be seen in the screenshot below. Since custom emoji is not loaded, the category is not present, changing the picker size to fit only 8 emoji in a line. Each line tries to render 9 (recently changed to fix performance issues), but since the size is smaller the content is wrapped. After refresh when custom emoji loads the size is fixed and it returns to normal.

image

PR #35764, opened 2025-04-10
chore: Break down big `livechatTyped` file - 6

KevLehman:
## Issue(s)

https://rocketchat.atlassian.net/browse/CTZ-36


apps/meteor/app/livechat/server/lib/guests.ts
```diff
@@ -160,3 +165,92 @@ async function cleanGuestHistory(token: string) {
...
+}
+
+function notifyRoomVisitorChange(roomId: string, visitor: ILivechatVisitor) {
+	void api.broadcast('omnichannel.room', roomId, {
+		type: 'visitorData',
+		visitor,
+	});
+}
+
+// TODO: Remove this function in next major
```
sampaiodiego, line 237: just do it now.. internal functions don't need to wait for major releases.

sampaiodiego (changes requested):

apps/meteor/app/livechat/server/lib/guests.ts
KevLehman: Oh wasn't "major", was "next task" 🤦🏽 

anyways, removed

PR #35776, opened 2025-04-10
feat: improve voip error handling

pierre-lehnen-rc:

Determine the reason for a call having ended and if the reason is not natural, show an error message to the user.

- Replaced the "Unable to complete call" error message with "Unable to negotiate call params." when the call fails with an "Not Acceptable Here" error.
- For all other errors, replaced the "Unable to complete call" error message with "Unable to complete call. Error code [{{statusCode}}]", which will show the code for the SIP error that caused the call to fail
- When a call started by us fails for a non-natural reason, we'll now display one of the following two messages in an error toast:
  - "An Incoming Voice Call was canceled due to an unexpected error."
  - "An Incoming Voice Call was canceled suddenly."
- Changed the voip client internal code to log some information to the console when a call fails or throws an error.

## Issue(s)

[CORE-1081](https://rocketchat.atlassian.net/browse/CORE-1081)

## Further comments

[CORE-1081]: https://rocketchat.atlassian.net/browse/CORE-1081?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

---

This pull request aims to enhance the error handling capabilities for voice calls in the Rocket.Chat application. The changes include:

- Improved error handling across multiple packages, specifically for voice over IP (VoIP) scenarios.
- Addition of four new English translation strings related to error messages and notifications for voice calls and call negotiation in the `i18n` JSON file.
- Updates to the `VoipErrorView` component, including new error status code handling and improved default error messages that incorporate status codes.
- Enhancements to the `VoipClient` to parse SIP invitation cancellation reasons and emit a new `incomingcallerror` event for specific failure scenarios, such as 'USER_NOT_REGISTERED' or unexpected cancellations. This also includes error logging improvements and proper wiring of the new cancellation handler to the invitation delegate.
- Updates to the `VoipProvider` to handle incoming call errors more effectively.
- A test case update to ensure the correct error message is displayed for unknown error calls in the VoIP error view component.

These improvements are intended to provide more robust error handling and clearer communication of issues during VoIP calls.


packages/ui-voip/src/lib/VoipClient.ts
```diff
@@ -843,12 +845,59 @@ class VoipClient extends Emitter<VoipEvents> {
...
+
+			if (parsedReasons.includes('USER_NOT_REGISTERED')) {
+				// This one means an error happened
+				this.emit('incomingcallerror', 'USER_NOT_REGISTERED');
+				return;
+			}
+
+			if (invitation.state === SessionState.Initial) {
+				// Call was canceled at the initial state and it was not due to one of the natural reasons, treat it as unexpected
+				this.emit('incomingcallerror', parsedReasons[0]);
```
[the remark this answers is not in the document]
pierre-lehnen-rc: done on purpose; If we get multiple reasons on the list it's likely that it's just the same reason being identified using two different protocols. If we successfully identify any of the reasons on the list, we can treat that as **the** reasons.

packages/ui-voip/src/lib/VoipClient.ts
```diff
@@ -843,12 +845,59 @@ class VoipClient extends Emitter<VoipEvents> {
 		this.emit('unregistrationerror', error);
 	};
 
+	private onInvitationCancel(invitation: Invitation, message: SipCancel): void {
```
MarcosSpessatto, line 848: Can we unit test this?

pierre-lehnen-rc: Not yet :/

vmarinogg:
This PR is important for us to test the VoIP solution on Open continuously, and also represents important improvements in how we currently handle call errors. If approved technically, we can move forward with it.

ggazzo (approved):

MarcosSpessatto (approved):

PR #35790, opened 2025-04-11
fix: Enterprise settings disabled in `Admin > Settings` despite a license with required modules being present

gabriellsh:

Added missing modules to enteprise settings.

~Enforced `modules` property to be present in settings with `enteprise: true` by throwing an error if the property is not present at the time of the setting creation.~ This will be done separately
## Issue(s)

[CORE-1034](https://rocketchat.atlassian.net/browse/CORE-1034)
## Steps to test or reproduce

In a workspace with a valid license:
- Go to admin > settings > layout
- Enable `Show custom content to homepage`
- Save
- `Show custom content only` will be disabled despite having a valid license.
## Further comments

[CORE-1034]: https://rocketchat.atlassian.net/browse/CORE-1034?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


MarcosSpessatto (changes requested):
It would be nice to ensure this through type as well... WDYT?

PR #35797, opened 2025-04-14
fix: cleanRoomHistory with filesOnly=true removes all attachments including quotes

abhinavkrin:

When `filesOnly` is set to `true`, the current pruning logic removes all attachments—including quotes and other non file attachments This fix ensures only attachments of type `file` (e.g., images, documents, videos) are pruned while keeping non-file attachments intact.

This impacts:
- Retention Policy cleanup under `/admin/settings/RetentionPolicy`
- The `rooms.cleanHistory` endpoint
## Steps to test or reproduce

1. Send a message with both file and quote attachments.
2. Run retention cleanup or use the `rooms.cleanHistory` API with `filesOnly: true`.
3. After cleanup, confirm:
   - File attachments are removed.
   - Quote and other non-file attachments remain.
## Further comments

[CORE-1051](https://rocketchat.atlassian.net/browse/CORE-1051)

[CORE-1051]: https://rocketchat.atlassian.net/browse/CORE-1051?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

---

This pull request addresses an issue in the Rocket.Chat repository where the `cleanRoomHistory` function, when invoked with `filesOnly=true`, was incorrectly removing all attachments, including quotes. The changes ensure that only file-type attachments are removed, preserving quotes and other non-file attachments.

Key changes include:

1. **Refactoring in `cleanRoomHistory.ts`**: The method for handling file attachment removal has been optimized. Instead of updating each message individually, the process now collects all relevant message IDs and performs bulk database operations to remove attachments and clear file-related fields, improving performance.

2. **New Method in `IMessagesModel.ts`**: A method has been added to remove file attachments by message IDs, with an optional replacement feature.

3. **Enhancements in `Messages.ts`**: A new method is introduced to efficiently remove file attachments from messages using MongoDB aggregation operators, with an option to replace the attachments.

4. **Changeset Entries**: Two changeset files have been added to document the patch affecting `@rocket.chat/model-typings`, `@rocket.chat/models`, and `@rocket.chat/meteor`.

These updates ensure that the room history pruning functionality works as intended, maintaining non-file attachments while removing only the specified file-type attachments.


apps/meteor/app/lib/server/functions/cleanRoomHistory.ts
```diff
@@ -51,7 +51,22 @@ export async function cleanRoomHistory({
 
 		fileCount++;
 		if (filesOnly) {
-			await Messages.updateOne({ _id: document._id }, { $unset: { file: 1 }, $set: { attachments: [{ color: '#FD745E', text }] } });
+			await Messages.updateOne({ _id: document._id }, [
```
MarcosSpessatto, line 54: I think it would be better to have this instruction in a model method. WDYT?

apps/meteor/tests/end-to-end/api/rooms.ts
```diff
@@ -1176,6 +1176,75 @@ describe('[Rooms]', () => {
...
+			const siteUrl = await getSettingValueById('Site_Url');
+			const message1Response = await sendSimpleMessage({ roomId: publicChannel._id });
+			const mediaResponse = await request
+				.post(api(`rooms.media/${publicChannel._id}`))
+				.set(credentials)
+				.attach('file', imgURL)
+				.expect('Content-Type', 'application/json')
+				.expect(200);
+
+			const message2Response = await request
```
MarcosSpessatto, line 1190: Let's separate this in a specific case, e.g: 'should not delete quote attachments when filesOnly is set to true'

MarcosSpessatto (changes requested):

.changeset/tough-chefs-repeat.md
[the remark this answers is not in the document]
abhinavkrin: I thought i deleted it. 😮‍💨

MarcosSpessatto (approved):

PR #35809, opened 2025-04-15
chore: Improve PDF Transcript request lock mechanism

KevLehman:

The idea of the lock is to prevent the user from requesting a transcript for the same room when there's already an active request. Previously, we did this by setting a flag on the `room` object of the DB, and while this works, when the process hangs, there was no way to remove the flag without DB intervention. Additionally, this meant 2 DB calls, one for setting and one for unsetting the flag.

By using an auto expiration map (same we do for E2EE), we store the room's id in memory, allowing a new request each 15 seconds (this is a "magic" number, taken from 3 times the Queue Worker polling interval). If the process succeeds, we error out next time a transcript is requested, if the process fails, a new request can be made without any problem.
## Issue(s)

https://rocketchat.atlassian.net/browse/CTZ-45
## Further comments

---

This pull request focuses on improving the PDF transcript request mechanism in the Rocket.Chat repository. Key changes include:

1. **Statistics Collection**: The collection of statistics for PDF transcript requests has been removed, with a compatibility note added, suggesting a deprecation process.

2. **API Endpoint Update**: The `omnichannel/:rid/request-transcript` API endpoint has been updated to use a projection for fetching specific fields (`_id`, `open`, `v`, `t`) of the livechat room. This change enhances performance by reducing unnecessary data retrieval and improves type safety by clearly defining the expected structure of the `room` object.

3. **PDF Transcript Request Logic**: The logic for handling PDF transcript requests has been refactored. A database flag (`pdfTranscriptRequested`) has been replaced with an in-memory, time-based lock (`ExpiryMap`) to prevent duplicate requests within a short period. Additionally, a check is added to avoid requesting a transcript if one already exists (`pdfTranscriptFileId`). However, the use of an in-memory lock may not be suitable for horizontally scaled environments.

4. **Omnichannel Transcript Generation**: Enhancements have been made to prevent redundant PDF creation if a transcript already exists for a room. Performance and type safety are improved by using specific type projections when retrieving room and user data, ensuring only necessary fields are fetched.

5. **Database Index**: An index on the `pdfTranscriptRequested` field has been commented out with a note to remove it in the next major version, indicating a move towards deprecation.

These changes aim to optimize the PDF transcript request process, improve system performance, and prepare for future deprecations.


lucas-a-pelegrino (dismissed):
Let's not forget to create tasks to track those TODOS if they are not created already. :)

debdutdeb (dismissed):

apps/meteor/ee/app/livechat-enterprise/server/lib/requestPdfTranscript.ts
```diff
@@ -1,12 +1,15 @@
 import { OmnichannelTranscript, QueueWorker } from '@rocket.chat/core-services';
 import type { AtLeast, IOmnichannelRoom } from '@rocket.chat/core-typings';
-import { LivechatRooms } from '@rocket.chat/models';
+import ExpiryMap from 'expiry-map';
 
 import { logger } from './logger';
 
+// Allow to request a transcript again after 15 seconds, assuming the first one didn't complete
```
sampaiodiego, line 7:
```suggestion
// Allow to request a transcript again after 15 seconds, assuming the first one didn't complete
// This won't prevent multiple transcript generated for the same room in a multi-instance deployment since state is not shared, but we're ok with the drawbacks
```

sampaiodiego (approved):

PR #35828, opened 2025-04-16
fix: new user not moving from Pending to Deactivated tab in admin UI

cardoso:
## Issue(s)

[SUP-775]

## Steps to test or reproduce

1. Create a new user in RC and do not log in with the user
2. In the Pending tab on the Users admin panel, confirm that the pending action for the user is “User first log in”
3. Deactivate that user

### Expected Behavior
The user should be listed in the Deactivated tab on the Users admin panel and not the pending tab

### Actual Behavior
The user is still listed in the pending tab

## Further comments

---

This pull request addresses an issue in the Rocket.Chat admin UI where new users who have not logged in remain incorrectly listed under the "Pending" tab even after being deactivated. The changes involve updating the `useFilteredUsers` hook by adding a status field to the pending user payload. This modification ensures consistency with other user status definitions and corrects the behavior so that deactivated users are appropriately moved to the "Deactivated" tab.

[SUP-775]: https://rocketchat.atlassian.net/browse/SUP-775?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/meteor/tests/e2e/admin-users.spec.ts
```diff
@@ -0,0 +1,91 @@
...
+			await page.getByRole('menuitem', { name: 'Activate' }).click();
+			await expect(page.getByRole('link', { name: username })).not.toBeVisible();
+
+			await page.getByRole('tab', { name: 'Pending' }).click();
+			await expect(page.getByRole('link', { name: username })).toBeVisible();
+		});
+	},
+);
+
+async function createUser(poAdmin: Admin) {
```
yash-rajpal, line 68: Maybe you can use rest API to create a new user to test this, or if we want it this way, we can move this utility function in a file of its own as a better code practice.

cardoso: I'm leaning towards not using the rest api directly, so I'll follow your advice about moving the utility to its own file.

However, if you have a case for using the rest API directly, let me know.

cardoso: I'll try that first then, thanks!

yash-rajpal (dismissed):
Can we please wrap callbacks like `beforeAll` and `afterAll` inside a `test.describe` callback, to follow a set pattern? You can check other tests.

.changeset/fuzzy-gifts-move.md
```diff
@@ -0,0 +1,5 @@
+---
+"@rocket.chat/meteor": patch
+---
+
+fix: new user not moving from Pending to Deactivated tab in admin UI
```
MarcosSpessatto, line 5: Fixes an error where the user was not moving from Pending to Deactivated tab in the admin UI.

apps/meteor/tests/e2e/admin-users.spec.ts
```diff
@@ -0,0 +1,92 @@
...
+			tag: '@admin',
+			annotation: {
+				type: 'issue',
+				description: 'https://rocketchat.atlassian.net/browse/SUP-775',
+			},
+		},
+		async ({ page }) => {
+			const { username } = user;
+
+			if (!username) {
```
MarcosSpessatto, line 45: I don't think this is necessary

MarcosSpessatto (changes requested):

MarcosSpessatto (approved):

PR #35839, opened 2025-04-17
fix: upload of malformed jpegs with Message_Attachments_Strip_Exif disabled fails silently

cardoso:
## Issue(s)

[SUP-769](https://rocketchat.atlassian.net/browse/SUP-769)

## Further comments

[SUP-769]: https://rocketchat.atlassian.net/browse/SUP-769?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

---

### Description

This pull request addresses an issue in the Rocket.Chat repository where image uploads were failing for specific JPEG files. The problem only occurs when Message_Attachments_Strip_Exif is set to false (current default is true). Some of the upload logic which uses xhr also needed tweaking, and the frontend side needed to display the error.


cardoso:
May be worth looking into sharp 0.34: https://github.com/lovell/sharp/issues/3947#issuecomment-2778877582

cardoso:
@kody start-review

MarcosSpessatto (dismissed):
Please, don't forget to add a changeset

MarcosSpessatto (approved):

apps/meteor/client/lib/chats/uploads.ts
```diff
@@ -128,6 +127,9 @@ const send = async (
 						t,
 						content,
 					});
+				} else if (xhr.readyState === xhr.DONE && xhr.status === 400) {
```
dougfabris, line 130: WYT about this approach?

```tsx
xhr.onload = async () => {
	if (xhr.readyState === xhr.DONE) {
		if (xhr.status === 400) {
			const error = JSON.parse(xhr.responseText);
			updateUploads((uploads) => [...uploads, { ...upload, error: new Error(error.error) }]);
			return;
		}
	
		if (xhr.status === 200) {
			const result = JSON.parse(xhr.responseText);
			let content;
			if (getContent) {
				content = await getContent(result.file._id, result.file.url);
			}
	
			await sdk.rest.post(`/v1/rooms.mediaConfirm/${rid}/${result.file._id}`, {
				msg,
				tmid,
				description,
				t,
				content,
			});
		}
	}
};
```

apps/meteor/client/views/room/body/UploadProgress/UploadProgressIndicator.tsx
```diff
@@ -53,7 +53,7 @@ const UploadProgressIndicator = ({ id, name, percentage, error, onClose }: Uploa
 			className={customClass}
 		>
 			<Box withTruncatedText zIndex={2} borderRadius={4}>
-				[{percentage}%] {name}
+				[{percentage}%] {name} {error && ` - ${error}`}
```
dougfabris, line 56: any chance of this to be an empty string? if so, lets force boolean
```suggestion
				[{percentage}%] {name} {Boolean(error) && ` - ${error}`}
```

apps/meteor/tests/e2e/image-upload.spec.ts
dougfabris, line 1: we already have the `file-upload` test file, do you think we can take any advantage of the file or its indeed better to keep this file?

dougfabris (changes requested):

apps/meteor/tests/e2e/image-upload.spec.ts
cardoso: I think it's better in this file since images go through different code paths than what file-upload.spec.ts is handling. Ideally they'd be in a common folder, but that's a different discussion.

apps/meteor/client/lib/chats/uploads.ts
cardoso: Wanted to minimize the diff, but I committed your suggestion. Should be the same. 👍

apps/meteor/tests/e2e/image-upload.spec.ts
dougfabris: indeed, thank you!

MarcosSpessatto (approved):

apps/meteor/tests/e2e/image-upload.spec.ts
```diff
@@ -0,0 +1,58 @@
...
+			await setSettingValueById(api, 'Message_Attachments_Strip_Exif', false);
+		});
+
+		test('should show error indicator when upload fails', async ({ page }) => {
+			await poHomeChannel.content.sendFileMessage('bad-orientation.jpeg');
+			await poHomeChannel.content.fileNameInput.fill('bad-orientation.jpeg');
+			await poHomeChannel.content.descriptionInput.fill('bad-orientation_description');
+			await poHomeChannel.content.btnModalConfirm.click();
+
+			await expect(page.getByText('bad-orientation.jpeg')).toContainText('Error:');
```
dougfabris, line 40: can we query this by role? we can take the advantage of your work here and enhance the a11y of this `UploadIndicator`, I will raise a commit in order to help you out!

dougfabris (changes requested):

apps/meteor/tests/e2e/image-upload.spec.ts
cardoso: Thank you!

dougfabris: now, is just a matter of getByRole status and boom 💥

cardoso: Done :)

dougfabris (approved):
LGTM!

PR #35863, opened 2025-04-23
fix: encoding of special characters in text attachments

cardoso:

This PR fixes an issue when posting messages via the API: special characters like `>` `<` and `&` symbol in code blocks inside text attachments are displayed as `&gt;` , `&lt;`, and `&amp;`.

## Issue(s)

https://rocketchat.atlassian.net/browse/SUP-725

## Steps to test or reproduce

API Endpoint: /api/v1/chat.postMessage

Example API Body:
```json
{
  "channel": "#COMM_EMOJI_TEST",
  "text": "```text\n TEST > TEST \n```",
   "attachments": [{ "text": "```text\n TEST > TEST\n```" }]
}
```

Result: The `>` symbol becomes `&gt;` in the attachment text but not in the message body.

image-20250110-165501

## Further comments

---

**Pull Request Description**:

This pull request addresses an issue in the Rocket.Chat repository related to the encoding of special characters in attachments. The changes are made in the `DefaultAttachment.tsx` file within the `apps/meteor/client/components/message/content/attachments` directory. The update introduces the use of the Markup component from the gazzodown library to handle markdown content in attachments. Additionally, it implements conditional rendering based on the presence of the `md` property to ensure proper encoding and display of special characters. The source branch for this fix is `SUP-725-api-txt-attachment-enc`, and it targets the `develop` branch.


gabriellsh (changes requested):
Looks good, just noticed the same issue with inline code:
image

Do you think we can fix it too?

cardoso:
@gabriellsh escaping inline is trickier, so I think it's better to fix just the code block use case for now.

apps/meteor/client/components/MarkdownText.spec.tsx
```diff
@@ -31,9 +31,13 @@ const markdownText = `
   gabriel.engel@rocket.chat
   +55991999999
   \`Inline code\`
+  \`2 < 3 > 1 & 4 "Test"\`
```
ggazzo, line 34: missing the expect for this case

ggazzo (dismissed):

cardoso:
@ggazzo what do you think about using the markup component for this? The first commit:

https://github.com/RocketChat/Rocket.Chat/pull/35863/commits/cd355b947557ee413b894222a28f24cb7729c329

gabriellsh (approved):

PR #35876, opened 2025-04-24
refactor: Omnichannel's monitor autocompletes to useInfiniteQuery

aleksandernsilva:
This PR refactors the `useMonitorsList` hook to take advantage of react query's useInfiniteQuery instead of RecordList in combination with useScrollableRecordList.

## Issue(s)
[CTZ-105](https://rocketchat.atlassian.net/browse/CTZ-105)

## Steps to test or reproduce
This change can be tested by visiting pages using monitor autocompletes and confirming the expected behavior: fetching, filtering and lazy loading.
- Create/edit unit

## Further comments

---

This pull request refactors the Omnichannel's monitor autocomplete functionality to utilize the `useInfiniteQuery` approach. It introduces a new reusable component, `AutoCompleteMonitors`, which provides a paginated multi-select interface for selecting monitors. This component leverages the `useMonitorsList` hook to fetch monitor data and integrates with the `@rocket.chat/fuselage` library's `PaginatedMultiSelectFiltered` component.

Additionally, a new custom hook, `useMonitorsList`, is introduced to manage a paginated list of Livechat monitors using `@tanstack/react-query`. This hook fetches data from the `/v1/livechat/monitors` endpoint, supporting filtering and pagination, and transforms the data into a usable format.

The `UnitEdit` component within the Omnichannel units has been refactored to replace the generic `PaginatedMultiSelectFiltered` component with the new `AutoCompleteMonitors` component. This change simplifies the `UnitEdit` component by removing the previously required state and hooks for local fetching and filtering of monitors.

[CTZ-105]: https://rocketchat.atlassian.net/browse/CTZ-105?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/meteor/client/omnichannel/units/UnitEdit.tsx
```diff
@@ -228,7 +210,7 @@ const UnitEdit = ({ unitData, unitMonitors, unitDepartments }: UnitEditProps) =>
 							{errors?.visibility && <FieldError id={`${visibilityField}-error`}>{errors?.visibility.message}</FieldError>}
 						</Field>
 						<Field>
-							<FieldLabel htmlFor={departmentsField} required>
+							<FieldLabel id={departmentsField} required>
```
tassoevan, line 213: Why this change? The a11y outline looks odd:

image

aleksandernsilva: I agree it seems a bit odd, but I'd argue it is still more accessible than it was before:
Screenshot 2025-05-19 at 10 50 06
Using a screen reader this translates to "Select an option, blank, edit text" or simply "Select an option" with no mention of the label. 

The main issue is that the underlying component passes the `id` to a div wrapper instead of the input. The spec says that `<label>` `for` should point to a [labelable form-related element](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/label#attributes), so div would not be compliant. By using the `aria-labelledby` we get the minimal association and the reader mentions the label "Select an option, Departments, edit text".

My goal is to implement the definitive fix on Fuselage.

tassoevan (approved):

KevLehman (approved):

PR #35878, opened 2025-04-25
regression: Room update via apps engine failing with `Duplicated index key` error

KevLehman:
## Issue(s)

https://rocketchat.atlassian.net/browse/CTZ-86
## Steps to test or reproduce

When an app (like WhatsApp) adds custom fields to the room, it issues an update via bridge with just the necessary data (in this case, the _id and customFields props).

When the apps engine receives this, it triest to convert that into a rocketchat room object, assuming some properties are there. Since, in this case, the `name` was not in the original update, then it defaulted to `undefined`, leaving the update statement with something like:
```
{ _id: 'xxx', customFields: {}, name: undefined }
```
Previously, the driver would ignore those undefined fields, but now we're passing them into the query. This caused the room to be `name: undefined` in the database (apart from other fields, like the room's `type`!)

Then, when another update or another query to the room was made, the app wasn't able to find the room it creatd (as it lacked the type `l` for livechat) and it attempted to create a new one. Then, when it added the custom fields, the same would happen, but this time, it would conflict with the previous room as both would have `name: undefined` on the database (and name should be unique).

This would throw a Duplicated Index Key error, which the apps engine would fail to parse, causing the whole operation to fail and the gateway to reschedule the message...

## Further comments

Caused by https://github.com/RocketChat/Rocket.Chat/pull/31497
(Or at least, this one made it surface :p )


apps/meteor/app/apps/server/converters/rooms.js
```diff
@@ -170,7 +170,11 @@ export class AppRoomsConverter {
 		};
 
 		if (!isPartial) {
-			Object.assign(newRoom, room._unmappedProperties_);
+			Object.assign(
+				newRoom,
+				{ t: room.type, ts: room.createdAt, msgs: room.messageCount || 0, _updatedAt: room.updatedAt },
```
sampaiodiego, line 175: all of these are covered already with the new `typeof`s you add, I guess this is not needed?
```suggestion
```

sampaiodiego (changes requested):

apps/meteor/app/apps/server/converters/rooms.js
KevLehman: I put them here to ensure they would be always there 🤔 i can remove tho

sampaiodiego (approved):

PR #35880, opened 2025-04-25
regression: fix jump from pseudo message lists

ggazzo:


dougfabris (approved):
Since we do not have virtualization in the room, I'm afraid we add some e2e test to handle scroll in this list and as it's unstable we'll introduce some intermittence. As we discussed, we're going to add some unit in a near feature, so IMO we are ok to move forward with this now

PR #35889, opened 2025-04-28
fix: `FileUpload` initial focus order

dougfabris:

In order to be a11y full compliant we removed the forced focus on the description input to follow the natural DOM order and it **created some bad usability and frustrations** to a lot of users. 

Forcing the focus is technically compliant, but slightly deviates from the natural tab order starting at the top. Screen reader users will still encounter the elements in the DOM order, but the initial focus will be different. Since this a simple modal, with only two inputs, this will not to be a major issue.

> "When a dialog opens, focus moves to an element contained in the dialog. Generally, focus is initially set on the first focusable element. However, the most appropriate focus placement will depend on the nature and size of the content." (https://www.w3.org/WAI/ARIA/apg/patterns/dialog-modal/)

## Further comments

[CORE-1102]

[CORE-1102]: https://rocketchat.atlassian.net/browse/CORE-1102?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


dougfabris:
@kody start-review

yash-rajpal (approved):

yash-rajpal (approved):

PR #35904, opened 2025-04-29
fix: SavePreferences being called multiple times and causing issues

MartinSchoeler:
The preferences are being saved multiple times when you press save, this can cause some issues, like when selecting a new language, the page tries to reset the forms while the promise is being processed, causing the language to be saved as the previous value

## Issue(s)

[CORE-1106](https://rocketchat.atlassian.net/browse/CORE-1106)

## Steps to test or reproduce

1. Go to user preferences
2. Change the language
3. check if the language being used is the same as the one selected

This issue happened more consistently with the `Portuguese` and `Brazilan Portuguese` languages

## Further comments

---

This pull request addresses an issue where the `SavePreferences` function was being called multiple times, leading to potential problems. The changes include the introduction of a new React hook, `useSavePreferences`, which is designed to handle the saving of user account preferences. This hook uses `useMutation` from `@tanstack/react-query` to interact with the `/v1/users.setPreferences` API endpoint, processes specific fields like `highlights` and `dontAskAgainList` based on dirty fields, and provides user feedback through toast messages.

Additionally, a new test suite for the `useSavePreferences` hook has been added using React Testing Library and Jest to ensure that the hook correctly triggers the `users.setPreferences` API endpoint when invoked. The preference saving logic in the `AccountPreferencesPage` component has been refactored to use this new custom hook, simplifying the component and improving code organization by centralizing the saving mechanism.

[CORE-1106]: https://rocketchat.atlassian.net/browse/CORE-1106?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/meteor/client/views/account/preferences/AccountPreferencesPage.tsx
```diff
@@ -34,41 +31,7 @@ const AccountPreferencesPage = (): ReactElement => {
...
-			const list =
-				Array.isArray(dontAskAgainList) && dontAskAgainList.length > 0
-					? dontAskAgainList.map(([action, label]) => ({ action, label }))
-					: [];
-			Object.assign(data, { dontAskAgainList: list });
-		}
-
-		setPreferencesAction.mutateAsync({ data });
-	};
+	const handleSaveData = useSavePreferences({ dirtyFields, currentData, reset });
```
cardoso, line 34: What changed in the logic here besides moving to a hook?

apps/meteor/client/hooks/account/useSavePreferences.ts
```diff
@@ -0,0 +1,57 @@
...
+import { dispatchToastMessage } from '../../lib/toast';
+import type { AccountPreferencesData } from '../../views/account/preferences/useAccountPreferencesValues';
+
+type useSavePreferencesProps = {
+	dirtyFields: Partial<Record<keyof AccountPreferencesData, boolean | boolean[]>>;
+	reset: UseFormReset<AccountPreferencesData>;
+	currentData: AccountPreferencesData;
+};
+
+export const useSavePreferences = ({ dirtyFields, currentData, reset }: useSavePreferencesProps) => {
```
cardoso, line 16: I would advise against refactoring the logic into a separate hook in a fix, unless it would be used in another component. Makes it hard to understand what the root cause and fix was.

apps/meteor/client/hooks/account/useSavePreferences.spec.ts
```diff
@@ -0,0 +1,21 @@
...
+		const { result } = renderHook(() => useSavePreferences({ dirtyFields, reset: jest.fn(), currentData: {} }), {
+			wrapper: mockAppRoot().withEndpoint('POST', '/v1/users.setPreferences', mockSetPreferencesEndpoint).build(),
+		});
+
+		const formData: AccountPreferencesData = { language: 'en' };
+		await waitFor(() => result.current(formData));
+
+		expect(mockSetPreferencesEndpoint).toHaveBeenCalledTimes(1);
+	});
+});
```
cardoso, line 21: I think this should be an e2e test instead. I wasn't able to make this test fail, even by copying the old logic into the new hook.

cardoso (changes requested):

dougfabris (changes requested):
this is a refactor, the fix probably won't need this amount of code. I'd suggest separating it into a brand new PR in order to have a isolated fix

MartinSchoeler:
Could not reproduce the issue again, closing this for now

PR #35905, opened 2025-04-29
fix: file deletion called twice and ignoring errors with FileSystem storage type

cardoso:

Removes the call to fs.stat to check if the file existed and is a file (using isFile()) before attempting to delete it. This was done asynchronously and, if the file existed and was a file, it would then call fs.unlink to delete it.

Instead, the code now directly attempts to delete the file using unlink(path). If the file does not exist (ENOENT), the error is safely ignored.

This change also helps prevent race conditions where the file could be deleted between the stat and unlink calls.

Also fixes the delete method being called twice.

## Issue(s)

https://rocketchat.atlassian.net/browse/SUP-745


apps/meteor/tests/e2e/prune-messages.spec.ts
```diff
@@ -0,0 +1,192 @@
...
+			await settings.set('FileUpload_FileSystemPath', FILE_SYSTEM_PATHS.TEMPORARY);
+			await poHomeChannel.page.getByLabel('Close').click();
+			await poHomeChannel.content.openLastMessageMenu();
+			await poHomeChannel.page.getByRole('menuitem', { name: 'Delete' }).click();
+			const modal = poHomeChannel.page.getByLabel('Are you sure?');
+			await modal.getByRole('button', { name: 'Yes, delete' }).click();
+			await expect(modal).not.toBeVisible();
+			await expect(poHomeChannel.content.lastUserMessage).not.toBeVisible();
+		});
+	});
```
cardoso, line 151: If we delete the message when `fs.unlink` fails for any reason, then we lose track of that file and won't be able to delete it properly.

Ideally, the message gets deleted, but the reference to the file remains in the database so pruning is still attempted later.

I forced `ENOENT` in this test suite by changing `FileUpload_FileSystemPath` as just an easy way to reproduce. There are many possible errors:

https://man7.org/linux/man-pages/man2/unlink.2.html#ERRORS

cardoso:
@debdutdeb there's a [root cause analysis](https://rocketchat.atlassian.net/browse/SUP-745?focusedCommentId=50998) in the ticket and 3 other related tickets which were part of the issue and were merged (especially https://github.com/RocketChat/Rocket.Chat/pull/35797 which clears the files array).

About the improvements here: the old code would call delete twice, and there are possible race conditions between stat and unlink. There's no reason to call stat since unlink will perform the same pre-conditions and throw.

Indeed there's no way to fully guarantee deletion, but now we'll see real errors if something goes wrong (except for the file not existing, which is expected in some cases like changing `FileUpload_FileSystemPath` apart from the double call fixed here).

Maybe this PR can be renamed.

MarcosSpessatto:
> @debdutdeb there's a [root cause analysis](https://rocketchat.atlassian.net/browse/SUP-745?focusedCommentId=50998) in the ticket and 3 other related tickets which were part of the issue and were merged (especially #35797 which clears the files array).
> 
> About the improvements here: the old code would call delete twice, and there are possible race conditions between stat and unlink. There's no reason to call stat since unlink will perform the same pre-conditions and throw.
> 
> Indeed there's no way to fully guarantee deletion, but now we'll see real errors if something goes wrong (except for the file not existing, which is expected in some cases like changing `FileUpload_FileSystemPath` apart from the double call fixed here).
> 
> Maybe this PR can be renamed.

I think it makes sense to rename it @cardoso, maybe a chore instead?

cardoso:
@MarcosSpessatto I updated the name already to match what's being done:

` fix: pruning does not get rid of physical files` -> `fix: file deletion called twice and ignoring errors with FileSystem storage type`

I think it's better to keep it as a fix, but I can change to chore if you prefer.

PR #35914, opened 2025-04-30
chore: Move some translation functions to package

KevLehman:
## Further comments

Idea is to move them to a neutral place so we can use them on microservices, in the hopes of removing the translation service 😢 

---

This pull request refactors the internationalization (i18n) setup in the Rocket.Chat repository by moving certain translation functions to a new package, `@rocket.chat/i18n-utils`. The changes include:

- Refactoring the `i18n.ts` files in both the client and server directories to import utility functions like `addSprinfToI18n` from the new `@rocket.chat/i18n-utils` package, promoting code reuse and simplifying the local files.
- Updating the `TranslationProvider.tsx` to adjust import statements for i18n utilities, while maintaining the import of the `i18n` instance from the local utility file.
- Introducing a new `i18n-utils` package that provides functions for i18next integration, including sprintf support and namespace handling. The package aims to enhance type safety, performance, and maintainability.
- Adding an ESLint configuration file and a TypeScript configuration file for the `i18n-utils` package.
- Updating the `CHANGELOG.md` for the `@rocket.chat/log-format` package to document version updates and the movement of the ui-kit package to the main monorepo.
- Adding a new export for the `isObject` utility function in the `tools` package, along with a type guard function to check if a value is an object, with suggestions for improved type safety and documentation.


packages/tools/src/isObject.ts
```diff
@@ -0,0 +1,4 @@
+export function isObject(obj: unknown): obj is object {
```
sampaiodiego, line 1: pls deprecate the old function from `apps/meteor/lib/utils/isObject.ts` mentioning to use this one

packages/i18n/package.json
```diff
@@ -4,12 +4,16 @@
...
+		"@rocket.chat/tools": "workspace:~",
 		"eslint": "~8.45.0",
+		"i18next": "~23.4.9",
 		"jest": "~29.7.0",
 		"typescript": "~5.7.2"
 	},
 	"scripts": {
-		"build": "node ./src/index.mjs",
+		"build": "node ./src/scripts/index.mjs && run .:build:esm && run .:build:cjs",
+		".:build:esm": "tsc -p tsconfig.esm.json",
```
sampaiodiego, line 15: do we need to build as esm and cjs?

packages/i18n/src/scripts/index.mjs
sampaiodiego, line 1: maybe rename this file `build.ts` ? or something that makes _some_ sense 😬

packages/i18n/src/index.spec.mjs
```diff
@@ -1,94 +0,0 @@
...
-				multipleKeys: 'sentence __with__ multiple __keys__',
-			};
-			expect(normalizeI18nInterpolationsTask(fileContents, 'en')).toEqual({
-				shouldChange: 'this is the {{key}}',
-				shouldNotChange: 'normal one',
-				multipleKeys: 'sentence {{with}} multiple {{keys}}',
-			});
-		});
-
-		it('should write a txt file with the collected statistics', async () => {
```
sampaiodiego, line 43: any reason for this test to be removed?

sampaiodiego (changes requested):

packages/i18n/package.json
KevLehman: Apparently, yes 😬

packages/i18n/src/index.spec.mjs
KevLehman: This is not on a file anymore, it's now on the console 👀 on build

sampaiodiego (approved):

PR #35929, opened 2025-05-06
fix: Made the tags in aria describedby to be consistent with rendered tags.

ergot-rp:
## Issue(s)

[WA-27]

## Further comments

---

This pull request addresses accessibility and code maintainability issues in the Rocket.Chat project. Specifically, it introduces helper functions in `ChangePassword.tsx` to generate `aria-describedby` attributes, ensuring consistency between the tags and the rendered elements. Additionally, in `PasswordVerifier.tsx`, the return value is modified to include a hidden span element with the provided id when `passwordVerifications` is empty, instead of returning null. These changes aim to improve the readability and functionality of the code.

[WA-27]: https://rocketchat.atlassian.net/browse/WA-27?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/meteor/client/views/account/security/ChangePassword.tsx
```diff
@@ -43,6 +43,24 @@ const ChangePassword = (props: AllHTMLAttributes<HTMLFormElement>) => {
...
+	function getAriaDescribedbyForPassword(passwordError: boolean) {
+		let str = `${passwordVerifierId}`;
+		if (!allowPasswordChange) {
+			str += ` ${passwordId}-hint ${passwordId}-error`;
+		}
+		if (passwordError) {
+			str += ` ${passwordId}-error`;
+		}
+		return str;
+	}
```
gabriellsh, line 55: I noticed a little issue in this function, with the following properties:
`passwordError = true`
`allowPasswordChange = false`
`passwordId = 'yyy'`
`passwordVerifierId = 'xxx'`
Results in the following string: `'xxx yyy-hint yyy-error yyy-error'`. I think `${passwordId}-error` should not appear twice, correct? Maybe that case doesn't exist, but I wouldn't bet on it, specially since things can change in the future. I suggest the following:

```suggestion
	function getAriaDescribedbyForPassword(
		passwordVerifierId: string,
		passwordId: string,
		allowPasswordChange: boolean,
		passwordError: boolean,
	) {
		const error = !allowPasswordChange || passwordError;
		return [passwordVerifierId, !allowPasswordChange && `${passwordId}-hint`, error && `${passwordId}-error`].filter(Boolean).join(' ');
	}
```

Then we can remove the function declaration from the Component's body (just define the function above the component). This is not a big deal but since this function will run every time anyways theres no reason for us to not do this, as it helps with readability and separation of concerns.

apps/meteor/client/views/account/security/ChangePassword.tsx
```diff
@@ -96,7 +114,7 @@ const ChangePassword = (props: AllHTMLAttributes<HTMLFormElement>) => {
 									disabled={!allowPasswordChange || !passwordIsValid}
 									aria-required={password !== '' ? 'true' : 'false'}
 									aria-invalid={errors.confirmationPassword ? 'true' : 'false'}
-									aria-describedby={`${confirmPasswordId}-error`}
+									aria-describedby={getAriaDescribedbyForPasswordVerification(!!errors.confirmationPassword)}
```
gabriellsh, line 117: WDYT of removing this function and just doing:

```suggestion
									aria-describedby={!!errors.confirmationPassword ? `${confirmPasswordId}-error` : undefined}
```

gabriellsh (changes requested):

apps/meteor/client/views/account/security/ChangePassword.tsx
ergot-rp: ok

gabriellsh (approved):
Nice! Thank you very much!

PR #35935, opened 2025-05-06
regression: Passing `servedBy` instead of `servedBy.id` to `findOne` call when converting a room from app

KevLehman:
## Issue(s)

https://rocketchat.atlassian.net/browse/CTZ-24

## Further comments

Caused by https://github.com/RocketChat/Rocket.Chat/pull/31497

### DIFF
--- a/apps/meteor/app/apps/server/converters/rooms.js
+++ b/apps/meteor/app/apps/server/converters/rooms.js
@@ -59,12 +59,12 @@ export class AppRoomsConverter {
 		};
 	}
 
-	async __getUserIdAndUsername(uid) {
-		if (!uid) {
+	async __getUserIdAndUsername(userObj) {
+		if (!userObj?.id) {
 			return;
 		}
 
-		const user = await Users.findOneById(uid, { projection: { _id: 1, username: 1 } });
+		const user = await Users.findOneById(userObj.id, { projection: { _id: 1, username: 1 } });
 		if (!user) {
 			return;
 		}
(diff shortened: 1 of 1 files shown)


apps/meteor/app/apps/server/converters/rooms.js
```diff
@@ -129,7 +129,7 @@ export class AppRoomsConverter {
 
 		const departmentId = await this.__getDepartment(room);
 
-		const servedBy = await this.__getUserIdAndUsername(room.servedBy);
+		const servedBy = await this.__getUserIdAndUsername(room?.servedBy?.id);
```
sampaiodiego, line 132: `room` doesn't need to be checked.

```suggestion
		const servedBy = await this.__getUserIdAndUsername(room.servedBy?.id);
```

I wonder if we should keep it as `room.servedBy` and check if not `null` within `__getUserIdAndUsername`

sampaiodiego (changes requested):

apps/meteor/app/apps/server/converters/rooms.js
KevLehman: I wanted to keep the changes minimal, and since func was expecting a uid, i passed a uid.

But guess any way it's good 👀 (function before this changes it's already at 33 CC iirc)

apps/meteor/app/apps/server/converters/rooms.js
```diff
@@ -129,7 +129,7 @@ export class AppRoomsConverter {
 
 		const departmentId = await this.__getDepartment(room);
 
-		const servedBy = await this.__getUserIdAndUsername(room.servedBy);
+		const servedBy = await this.__getUserIdAndUsername(room?.servedBy);
```
KevLehman, line 132:
```suggestion
		const servedBy = await this.__getUserIdAndUsername(room.servedBy);
```

sampaiodiego (approved):

Issue #35919, opened 2025-05-02
Search Message by Custom Field


jkroepke:
Hi, I have an external system that sents messages to RocketChat. I would like to use customFields to enrich the metadata messages with the metadata from the external system, e.g. Alert ID.

But I'm missing a function to search Messages by an specific custom field within a channel.

Sauve9119:
Hey , @jkroepke 
I would like to contribute a PR to support searching messages by custom fields via the REST API.
- Add customFieldKey and customFieldValue to /api/v1/channels.messages
- It will allow filtering like:
  /channels.messages?roomId=xxx&customFieldKey=alertId&customFieldValue=ABC123

Let me know if this direction is okay, and I’ll open a PR!

jkroepke:
That sounds great!

Sauve9119:
Hey, @jkroepke I want to ask how we can connect in rocket.chat . I am a new contributor in Rocket.chat

jkroepke:
I have no idea, I'm new to rocket.chat as well.

PR #35943, opened 2025-05-07
Fix:  #35919 Implement findMessagesByCustomField API to filter messages by cu…

Rohitkumar2003:
This PR implements the findMessagesByCustomField API that allows searching for messages in a specified room based on a custom field value. The functionality includes pagination, sorting, and user validation to ensure only authorized users can access the messages.


Rohitkumar2003:
Hi! Just checking in to see if there’s anything I can improve in this PR. Thanks for reviewing!

PR #35956, opened 2025-05-09
chore: Update API to use 'route' instead of 'path'

ggazzo:
https://rocketchat.atlassian.net/browse/ARCH-1605
finishes the job started here: https://github.com/RocketChat/Rocket.Chat/pull/35948

## Further comments

---

This pull request updates the Rocket.Chat API to use 'route' instead of 'path', enhancing consistency and maintainability. In `definition.ts`, a readonly route property is added to the PartialThis type to improve type safety. In `parseJsonQuery.ts`, the code is updated to use 'route' for API endpoint handling, with added deprecation logging and maintained security checks for query parameters. Additionally, in `router.ts`, a global interface declaration is removed, and request handling in the Router class is refactored to improve type safety and request processing.


apps/meteor/app/api/server/v1/oauthapps.ts
```diff
@@ -37,7 +37,7 @@ API.v1.addRoute(
 			}
 
 			if ('appId' in this.queryParams) {
-				apiDeprecationLogger.parameter(this.request.route, 'appId', '7.0.0', this.response);
+				apiDeprecationLogger.parameter(this.route, 'appId', '7.0.0', this.response);
```
KevLehman, line 40: Since you editing these, should we put 8.0 instead of 7?

KevLehman (approved):

apps/meteor/app/api/server/v1/oauthapps.ts
ggazzo: i think it is a job for the next task :p

PR #35961, opened 2025-05-09
chore!: remove mongo 5/6 support

ggazzo:
https://rocketchat.atlassian.net/browse/ARCH-1459

### DIFF
--- a/.changeset/gold-keys-compare.md
+++ b/.changeset/gold-keys-compare.md
@@ -1,4 +1,4 @@
 ---
 "@rocket.chat/meteor": major
 ---
-Removes mongo 5/6 support
+Removes support of MongoDB versions 5.x and 6.x
--- a/.changeset/gold-keys-compare.md
+++ b/.changeset/gold-keys-compare.md
@@ -1,0 +1,4 @@
+---
+"@rocket.chat/meteor": major
+---
+Removes mongo 5/6 support
--- a/apps/meteor/server/startup/serverRunning.js
+++ b/apps/meteor/server/startup/serverRunning.js
@@ -1,17 +1,17 @@
 import fs from 'fs';
 import path from 'path';
 
-import { Users } from '@rocket.chat/models';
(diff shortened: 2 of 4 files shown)


apps/meteor/server/startup/serverRunning.js
```diff
@@ -88,42 +88,42 @@ Meteor.startup(async () => {
...
-							text,
-							textArguments: [mongoVersion],
-							modifiers: ['danger'],
-							link,
-						},
-					],
-				});
-			}
-		}
+		// if (!skipMongoDbDeprecationCheck && semver.satisfies(semver.coerce(mongoVersion), '<7.0.0')) {
```
sampaiodiego, line 91: we should keep the warning

.changeset/gold-keys-compare.md
```diff
@@ -0,0 +1,4 @@
+---
+"@rocket.chat/meteor": major
+---
+Removes mongo 5/6 support
```
sampaiodiego, line 4:
```suggestion
Removes support of MongoDB versions 5.x and 6.x
```

sampaiodiego (changes requested):

apps/meteor/server/startup/serverRunning.js
ggazzo: there is no version to be deprecated here so far, only 7.0.0

KevLehman:
Should we also support https://www.mongodb.com/docs/manual/release-notes/8.0/?

ggazzo:
> Should we also support [mongodb.com/docs/manual/release-notes/8.0](https://www.mongodb.com/docs/manual/release-notes/8.0/)?

 meteor only uses 7.0

sampaiodiego (approved):

sampaiodiego:
I think we should support MongoDB 8.0 as well, but this can be added in a separate PR

PR #35962, opened 2025-05-09
fix: Prune messages form not resetting properly

yash-rajpal:


aleksandernsilva (changes requested):
Good fix, Yash. If it's not yet the case, I believe it would be a good idea to add tests to cover this feature.

Also, could you add a changeset and steps to reproduce to the PR?

cardoso:
@aleksandernsilva I'm writing tests for this 👍

cardoso (dismissed):
Looks good to me, I did some QA and added the tests @aleksandernsilva asked for.

aleksandernsilva (approved):

PR #35963, opened 2025-05-09
fix: Parsing emojis inside code blocks in attachment content using `chat.postMessage` API call

tiagoevanp:
## Further comments

[SUP-724]

[SUP-724]: https://rocketchat.atlassian.net/browse/SUP-724?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

---

This pull request addresses an issue in the Rocket.Chat repository where emojis were not being parsed correctly inside code blocks within attachment content when using the `chat.postMessage` API call. The changes involve updating the `emojiParser.ts` file to include emoji handling within CODE elements and refactoring the code to move HTML assignment outside of the conditional block. The source branch for this fix is `fix/attachment-parse-emoji`, and it is targeted to be merged into the `develop` branch.

### DIFF
--- a/.changeset/curly-weeks-talk.md
+++ b/.changeset/curly-weeks-talk.md
@@ -1,0 +1,5 @@
+---
+"@rocket.chat/meteor": patch
+---
+
+fixes emojis being parsed inside code blocks in text message attachment using `chat.postMessage` API call
--- a/.changeset/curly-weeks-talk.md
+++ b/.changeset/curly-weeks-talk.md
@@ -2,4 +2,4 @@
 "@rocket.chat/meteor": patch
 ---
 
-fixes emojis being parsed inside code blocks in text message attachment using `chat.postMessage` API call
+Fixes emojis being parsed inside code blocks in text message attachment using `chat.postMessage` API call
--- a/apps/meteor/app/emoji/client/emojiParser.ts
+++ b/apps/meteor/app/emoji/client/emojiParser.ts
@@ -23,6 +23,14 @@ export const emojiParser = (html: string) => {
 
 	const emojis = Array.from(checkEmojiOnly.querySelectorAll('.emoji:not(:empty), .emojione:not(:empty)'));
 
+	emojis.forEach((emojiElement) => {
+		const htmlElement = emojiElement.parentElement;
+
+		if (htmlElement && htmlElement.nodeName === 'CODE') {
+			emojiElement.replaceWith(emojiElement.getAttribute('title') ?? '');
+		}
+	});
+
 	let hasText = false;
 
 	if (!isIE11) {
@@ -52,9 +60,10 @@ export const emojiParser = (html: string) => {
 				const { classList } = emojis[i];
 				classList.add('big');
 			}
-			html = checkEmojiOnly.innerHTML;
 		}
 	}
+
+	html = checkEmojiOnly.innerHTML;
 
 	// apostrophe (') back to &#39;
 	html = html.replace(/\'/g, '&#39;');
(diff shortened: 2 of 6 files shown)


apps/meteor/tests/e2e/page-objects/fragments/home-content.ts
```diff
@@ -47,6 +47,10 @@ export class HomeContent {
 		return this.lastUserMessage.locator('[data-qa-type="message-body"]');
 	}
 
+	get lastUserMessageAttachment(): Locator {
+		return this.page.locator('[data-qa-type="message-attachment"]').last();
```
MartinSchoeler, line 51: Is there any way to not use `data-qa` here? Maybe try an accessible locator.

apps/meteor/app/emoji/client/emojiParser.ts
```diff
@@ -52,10 +60,11 @@ export const emojiParser = (html: string) => {
 				const { classList } = emojis[i];
 				classList.add('big');
 			}
-			html = checkEmojiOnly.innerHTML;
 		}
 	}
 
+	html = checkEmojiOnly.innerHTML;
```
MartinSchoeler, line 66: Could you elaborate a bit why this was moved, I couldn't understand the reason only looking at the code

.changeset/curly-weeks-talk.md
```diff
@@ -0,0 +1,5 @@
+---
+"@rocket.chat/meteor": patch
+---
+
+fixes emojis being parsed inside code blocks in text message attachment using `chat.postMessage` API call
```
MartinSchoeler, line 5: The issue also happened on the threads title, as the emojiParser is used there too, maybe we should add this info on the changeset


Screenshot 2025-05-13 at 15 40 50

apps/meteor/tests/e2e/messaging.spec.ts
```diff
@@ -22,121 +22,184 @@ test.describe.serial('Messaging', () => {
...
+			await page.keyboard.press('Tab');
+			await page.keyboard.press('Tab');
+			await expect(page.locator('[data-qa-type="message"]').last()).toBeFocused();
+		});
 	});
 
-	test('should focus the latest message when moving the focus on the list and theres no previous focus', async ({ page }) => {
-		await poHomeChannel.sidenav.openChat(targetChannel);
-		await page.getByRole('button', { name: targetChannel }).first().focus();
+	test.describe.serial('Message by "chat.postMessage" API method', () => {
```
MartinSchoeler, line 144: I think this test could be flaky, since between the post message and the expect, it can take some time, the expect could in some cases get the previous message, as it is using the last message locator. Maybe await for a confirmation that the message you want to test is already on the screen?

MartinSchoeler (dismissed):

apps/meteor/tests/e2e/page-objects/fragments/home-content.ts
tiagoevanp: I don't think so... since all the similar tests use the same data-qa-type to select internal <Box> components

apps/meteor/app/emoji/client/emojiParser.ts
tiagoevanp: Because in the end, all HTML should be put onto this one var... No need to have only inside of a conditional case

.changeset/curly-weeks-talk.md
tiagoevanp: This is not related to the issue itself... this is related to threads title not recognizing code blocks on it... must have a separated issue for it

tassoevan: Unrelated, since sanitization for thread title is more deep.

apps/meteor/app/emoji/client/emojiParser.ts
tassoevan: Originally this was used only to handle the "big emoji" case (one line containing only emojis). With one additional case, better generalize it.

tassoevan (approved):

PR #35970, opened 2025-05-12
refactor: Remove `afterTakeInquiry` callback

KevLehman:
## Issue(s)

https://rocketchat.atlassian.net/browse/CTZ-117
## Further comments

---

This pull request refactors the handling of the `afterTakeInquiry` logic in the Rocket.Chat repository. The changes involve removing the generic `callbacks.runAsync` mechanism and replacing it with a direct function import and call. Specifically, the `afterTakeInquiry` function is now imported from `./hooks` and called directly after an agent is assigned in the `RoutingManager.ts` file. Additionally, a new `afterTakeInquiry` function is added in `hooks.ts` to handle post-processing tasks, such as triggering a CRM webhook if configured. In the `afterTakeInquiry.ts` file within the livechat-enterprise module, the hook is refactored to use the `.patch` method instead of the legacy `callbacks.add`. This refactoring introduces new logic to manage last chatted agent routing, schedule auto-transfer for chats, and stop queue inactivity monitoring when an inquiry is taken, based on the respective feature settings.


apps/meteor/ee/app/livechat-enterprise/server/hooks/afterTakeInquiry.ts
```diff
@@ -1,25 +1,54 @@
...
+		agent: { agentId: string; username: string },
+	) => {
+		await originalFn({ inquiry, room }, agent);
 
 		if (!inquiry) {
-			return null;
+			return;
+		}
+
+		if (settings.get('Livechat_waiting_queue')) {
```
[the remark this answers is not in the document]
KevLehman: @kody this is not needed when it's done inside an if, we're just checking the value is truthy. Update your memory for future reviews

apps/meteor/ee/app/livechat-enterprise/server/hooks/afterTakeInquiry.ts
```diff
@@ -1,25 +1,55 @@
...
+				cbLogger.info({
+					msg: 'Updated last agent of visitor',
+					token,
+					newAgent: { _id: agent.agentId, username: agent.username },
+				});
+			}
+		}
+
+		const autoTransferTimeout = settings.get<number>('Livechat_auto_transfer_chat_timeout');
+		if (autoTransferTimeout && !(room?.autoTransferredAt || room?.autoTransferOngoing)) {
```
sampaiodiego, line 46: let's get rid of `!(conditions)` 😬 
```suggestion
		if (autoTransferTimeout && !room?.autoTransferredAt && !room?.autoTransferOngoing)) {
```

apps/meteor/app/livechat/server/lib/hooks.ts
```diff
@@ -109,3 +110,11 @@ export const onNewRoom = makeFunction(async (room: IOmnichannelRoom) => {
 		await sendToCRM('LivechatSessionStart', room);
 	}
 });
+
+export const afterTakeInquiry = makeFunction(
+	async (params: { inquiry: InquiryWithAgentInfo; room: IOmnichannelRoom }, _agent: { agentId: string; username: string }) => {
+		if (settings.get('Livechat_webhook_on_chat_taken')) {
+			await sendToCRM('LivechatSessionTaken', params.room);
+		}
```
sampaiodiego, line 118: not sure if I like the old code style more 🤔  maybe we should agree in a definitive style?
```suggestion
		if (!settings.get('Livechat_webhook_on_chat_taken')) {
			return;
		}
		await sendToCRM('LivechatSessionTaken', params.room);
```

apps/meteor/ee/app/livechat-enterprise/server/hooks/afterTakeInquiry.ts
```diff
@@ -1,25 +1,55 @@
...
+			await AutoTransferChatScheduler.scheduleRoom(inquiry.rid, autoTransferTimeout);
+		}
 
-		cbLogger.debug(`Statuses for queue ${department || 'Public'} updated successfully`);
-		return inquiry;
+		const maxQueueWaitTime = settings.get<number>('Livechat_max_queue_wait_time');
+		if (!maxQueueWaitTime || maxQueueWaitTime < 0) {
```
sampaiodiego, line 51: https://github.com/RocketChat/Rocket.Chat/blob/c209ea761ab859fc9166b2fafb3c36906149a0b3/apps/meteor/ee/app/livechat-enterprise/server/hooks/autoCloseQueued.ts#L6-L9

based on the old code I think this condition is inverted, and if it is indeed inverted since no test failed it seems we don't this feature covered, pls create a task to create a test if that's the case:
```suggestion
		if (maxQueueWaitTime > 0) {
```

side note: it is unclear why `Livechat_max_queue_wait_time` is used to call `stopInquiry` or not

apps/meteor/ee/app/livechat-enterprise/server/hooks/afterTakeInquiry.ts
```diff
@@ -1,25 +1,55 @@
...
+			});
 		}
 
-		const { department } = inquiry;
-		debouncedDispatchWaitingQueueStatus(department);
+		if (settings.get<boolean>('Livechat_last_chatted_agent_routing') && agent && RoutingManager.getConfig()?.autoAssignAgent) {
+			const { v: { token } = {} } = inquiry;
+			if (token) {
+				await LivechatVisitors.updateLastAgentByToken(token, { ...agent, ts: new Date() });
+				cbLogger.info({
```
sampaiodiego, line 37: hmm.. this is a new log, is it really useful?

apps/meteor/ee/app/livechat-enterprise/server/lib/AutoTransferChatScheduler.ts
```diff
@@ -94,7 +94,7 @@ export class AutoTransferChatSchedulerClass {
 			await returnRoomAsInquiry(room, departmentId, {
 				scope: 'autoTransferUnansweredChatsToQueue',
 				comment: timeoutDuration,
-				transferredBy: await this.getSchedulerUser(),
+				transferredBy: { ...(await this.getSchedulerUser()), userType: 'user' },
```
sampaiodiego, line 97: unrelated change?

apps/meteor/app/livechat/server/lib/hooks.ts
```diff
@@ -109,3 +110,11 @@ export const onNewRoom = makeFunction(async (room: IOmnichannelRoom) => {
 		await sendToCRM('LivechatSessionStart', room);
 	}
 });
+
+export const afterTakeInquiry = makeFunction(
+	async (params: { inquiry: InquiryWithAgentInfo; room: IOmnichannelRoom }, _agent: { agentId: string; username: string }) => {
```
sampaiodiego, line 115: since this is not a callback anymore, I'd suggest we make it a single param with three fields, it doesn't make much sense for a function to a have a second param if the first one is already a "collection" (like an object with very distinct fields)
```suggestion
	async (params: { inquiry: InquiryWithAgentInfo; room: IOmnichannelRoom, agent?: { agentId: string; username: string }) => {
```

sampaiodiego (changes requested):

apps/meteor/ee/app/livechat-enterprise/server/hooks/afterTakeInquiry.ts
KevLehman: I actually like these 🤔 
Do u think it's more readable that way?

sampaiodiego: yes.. I have a really hard time wrapping my head around those inverted multiple conditions lol

sampaiodiego (approved):

PR #35973, opened 2025-05-12
refactor: Remove `afterInquiryQueued` callback

KevLehman:
## Issue(s)

https://rocketchat.atlassian.net/browse/CTZ-118
## Further comments

---

This pull request refactors the inquiry queuing process in the Rocket.Chat live chat application. The main change involves removing the `afterInquiryQueued` callback and replacing it with direct function calls to improve code structure and maintainability. Specifically, the `QueueManager` class has been updated to use explicit function imports instead of generic callbacks. Additionally, a new `afterInquiryQueued` function with an empty implementation has been added. The post-inquiry queuing logic has been extracted into a dedicated function, `afterInquiryQueuedFunc`, which is integrated by patching the `afterInquiryQueued` hook. This refactoring enhances modularity and simplifies the previous implementation that involved a settings watcher.


apps/meteor/app/livechat/server/lib/hooks.ts
```diff
@@ -119,3 +120,6 @@ export const afterTakeInquiry = makeFunction(
 		await sendToCRM('LivechatSessionTaken', room);
 	},
 );
+
+// eslint-disable-next-line @typescript-eslint/no-empty-function
+export const afterInquiryQueued = makeFunction(async (_inquiry: ILivechatInquiryRecord) => {});
```
sampaiodiego, line 125: you gotta be smart
```suggestion
export const afterInquiryQueued = makeFunction(async (_inquiry: ILivechatInquiryRecord) => {
	// no op
});
```

this or either `return null` or even `return void 0` why not? I think any of these are better than messing up with eslint rules

apps/meteor/ee/app/livechat-enterprise/server/hooks/afterInquiryQueued.ts
```diff
@@ -1,32 +1,21 @@
...
 };
 
-settings.watch<number>('Livechat_max_queue_wait_time', (value) => {
-	if (value <= 0) {
-		callbacks.remove('livechat.afterInquiryQueued', 'livechat-inquiry-queued-set-queue-timer');
-		return;
-	}
-	callbacks.add('livechat.afterInquiryQueued', afterInquiryQueued, callbacks.priority.HIGH, 'livechat-inquiry-queued-set-queue-timer');
+afterInquiryQueued.patch((_originalFn: any, inquiry: ILivechatInquiryRecord) => {
+	return afterInquiryQueuedFunc(inquiry);
```
sampaiodiego, line 20: although we know the `originalFn` is empty, I think it is a good practice if we call it anyways.
```suggestion
afterInquiryQueued.patch((originalFn: any, inquiry: ILivechatInquiryRecord) => {
	await originalFn(inquiry);

	return afterInquiryQueuedFunc(inquiry);
```

sampaiodiego (changes requested):

sampaiodiego (approved):

PR #35975, opened 2025-05-12
feat: Add experimental contenteditable message composer

ishanmitra:

This PR introduces an **experimental real-time message composer** powered by a `<div contenteditable>` instead of the classic `<textarea>`. The goal is to enable richer in-place text formatting and interaction using modern DOM APIs, while maintaining compatibility with the current composer functionality.

**Key changes include:**
- Replaced the classic `<textarea>` input with a `<div contenteditable>` in `MessageComposerInputNew`.
- Implemented selection and cursor management using the *Selection API* to support real-time formatting features.
- Fixed the Enter key behavior to send messages as expected.
- Adapted input handling logic (e.g., `onInput`, `innerText.trim()`) to support correct state management (e.g., Send button enable/disable).
- Restored basic formatting actions (bold, italic, strikethrough).
- Added a feature flag `featurePreviewComposer` to toggle between the new input and the current one safely.
- Added an in-UI experimental label (`MessageComposerHint`) to indicate this is a preview feature.

image
image

## Steps to test 

Set the `featurePreviewComposer` flag to `true` (or `false`) in `apps/meteor/client/views/room/composer/ComposerMessage.tsx`

## Known Issues & Next Steps

- Slash commands `/` and emoji autocomplete open as expected, but *pressing Enter inserts the selection and also sends the message*, which is unintended.
- Placeholder text is not yet implemented for `<div contenteditable>`.

Additional work is needed to:
- Prevent sending when the slash command menu or emoji popups are active.
- Implement placeholder text and fix component UI breakage when the message is over 6 lines tall.

### DIFF
--- a/apps/meteor/app/ui-message/client/messageBox/newCreateComposerAPI.ts
+++ b/apps/meteor/app/ui-message/client/messageBox/newCreateComposerAPI.ts
@@ -58,7 +58,6 @@ export const newCreateComposerAPI = (input: HTMLDivElement, storageID: string): 
 		sel?.addRange(range);
 	};
 
-
 	let _quotedMessages: IMessage[] = [];
 
 	const persist = withDebouncing({ wait: 300 })(() => {
--- a/apps/meteor/app/ui-message/client/messageBox/newCreateComposerAPI.ts
+++ b/apps/meteor/app/ui-message/client/messageBox/newCreateComposerAPI.ts
@@ -1,0 +1,412 @@
+import type { IMessage } from '@rocket.chat/core-typings';
+import { Emitter } from '@rocket.chat/emitter';
+import { Accounts } from 'meteor/accounts-base';
+
+import type { FormattingButton } from './messageBoxFormatting';
+import { formattingButtons } from './messageBoxFormatting';
+import type { ComposerAPI } from '../../../../client/lib/chats/ChatAPI';
+import { withDebouncing } from '../../../../lib/utils/highOrderFunctions';
+
+export const newCreateComposerAPI = (input: HTMLDivElement, storageID: string): ComposerAPI => {
+	const triggerEvent = (input: HTMLDivElement, evt: string): void => {
+		const event = new Event(evt, { bubbles: true });
+		// TODO: Remove this hack for react to trigger onChange
+		const tracker = (input as any)._valueTracker;
+		if (tracker) {
+			tracker.setValue(new Date().toString());
+		}
+		input.dispatchEvent(event);
+	};
+
+	const emitter = new Emitter<{
+		quotedMessagesUpdate: void;
+		editing: void;
+		recording: void;
+		recordingVideo: void;
+		formatting: void;
+		mircophoneDenied: void;
+	}>();
+
+	/* Use Selection API to get the selectionStart and selectionEnd from contenteditable div */
+	const getSelectionRange = (input: HTMLDivElement): { selectionStart: number; selectionEnd: number } => {
+		const selection = window.getSelection();
+		if (!selection || selection.rangeCount === 0) {
+			return { selectionStart: 0, selectionEnd: input.innerText.length };
+		}
+
+		const range = selection.getRangeAt(0);
+		const preCaretRange = range.cloneRange();
+		preCaretRange.selectNodeContents(input);
+		preCaretRange.setEnd(range.startContainer, range.startOffset);
+		const selectionStart = preCaretRange.toString().length;
+		const selectionEnd = selectionStart + range.toString().length;
+
+		return { selectionStart, selectionEnd };
+	};
+
+	/* Use Selection API to set a selection range in contenteditable div */
+	const setSelectionRange = (input: HTMLDivElement, start: number, end: number): void => {
+		const range = document.createRange();
+		const sel = window.getSelection();
+
+		if (input.firstChild) {
+			range.setStart(input.firstChild, start);
+			range.setEnd(input.firstChild, end);
+		}
+
+		sel?.removeAllRanges();
+		sel?.addRange(range);
+	};
+
+
+	let _quotedMessages: IMessage[] = [];
+
+	const persist = withDebouncing({ wait: 300 })(() => {
+		if (input.innerText) {
+			Accounts.storageLocation.setItem(storageID, input.innerText);
+			return;
+		}
+
+		Accounts.storageLocation.removeItem(storageID);
+	});
+
+	const notifyQuotedMessagesUpdate = (): void => {
+		emitter.emit('quotedMessagesUpdate');
+	};
+
+	input.addEventListener('input', persist);
+
+	const setText = (
+		text: string,
+		{
+			selection,
+			skipFocus,
+		}: {
+			selection?:
+				| { readonly start?: number; readonly end?: number }
+				| ((previous: { readonly start: number; readonly end: number }) => { readonly start?: number; readonly end?: number });
+			skipFocus?: boolean;
+		} = {},
+	): void => {
+		!skipFocus && focus();
+
+		const { selectionStart, selectionEnd } = getSelectionRange(input);
+		const textAreaTxt = input.innerText;
+
+		if (typeof selection === 'function') {
+			selection = selection({ start: selectionStart, end: selectionEnd });
+		}
+
+		if (selection) {
+			if (!document.execCommand?.('insertText', false, text)) {
+				input.innerText = textAreaTxt.substring(0, selectionStart) + text + textAreaTxt.substring(selectionStart);
+				!skipFocus && focus();
+			}
+			setSelectionRange(input, selection.start ?? 0, selection.end ?? text.length);
+		}
+
+		if (!selection) {
+			input.innerText = text;
+		}
+
+		triggerEvent(input, 'input');
+		triggerEvent(input, 'change');
+
+		!skipFocus && focus();
+	};
+
+	const insertText = (text: string): void => {
+		setText(text, {
+			selection: ({ start, end }) => ({
+				start: start + text.length,
+				end: end + text.length,
+			}),
+		});
+	};
+
+	const clear = (): void => {
+		setText('');
+	};
+
+	const focus = (): void => {
+		input.focus();
+	};
+
+	const replyWith = async (text: string): Promise<void> => {
+		if (input) {
+			input.innerText = text;
+			input.focus();
+		}
+	};
+
+	const quoteMessage = async (message: IMessage): Promise<void> => {
+		_quotedMessages = [..._quotedMessages.filter((_message) => _message._id !== message._id), message];
+		notifyQuotedMessagesUpdate();
+		input.focus();
+	};
+
+	const dismissQuotedMessage = async (mid: IMessage['_id']): Promise<void> => {
+		_quotedMessages = _quotedMessages.filter((message) => message._id !== mid);
+		notifyQuotedMessagesUpdate();
+	};
+
+	const dismissAllQuotedMessages = async (): Promise<void> => {
+		_quotedMessages = [];
+		notifyQuotedMessagesUpdate();
+	};
+
+	const quotedMessages = {
+		get: () => _quotedMessages,
+		subscribe: (callback: () => void) => emitter.on('quotedMessagesUpdate', callback),
+	};
+
+	const [editing, setEditing] = (() => {
+		let editing = false;
+
+		return [
+			{
+				get: () => editing,
+				subscribe: (callback: () => void) => emitter.on('editing', callback),
+			},
+			(value: boolean) => {
+				editing = value;
+				emitter.emit('editing');
+			},
+		];
+	})();
+
+	const [recording, setRecordingMode] = (() => {
+		let recording = false;
+
+		return [
+			{
+				get: () => recording,
+				subscribe: (callback: () => void) => emitter.on('recording', callback),
+			},
+			(value: boolean) => {
+				recording = value;
+				emitter.emit('recording');
+			},
+		];
+	})();
+
+	const [recordingVideo, setRecordingVideo] = (() => {
+		let recordingVideo = false;
+
+		return [
+			{
+				get: () => recordingVideo,
+				subscribe: (callback: () => void) => emitter.on('recordingVideo', callback),
+			},
+			(value: boolean) => {
+				recordingVideo = value;
+				emitter.emit('recordingVideo');
+			},
+		];
+	})();
+
+	const [isMicrophoneDenied, setIsMicrophoneDenied] = (() => {
+		let isMicrophoneDenied = false;
+
+		return [
+			{
+				get: () => isMicrophoneDenied,
+				subscribe: (callback: () => void) => emitter.on('mircophoneDenied', callback),
+			},
+			(value: boolean) => {
+				isMicrophoneDenied = value;
+				emitter.emit('mircophoneDenied');
+			},
+		];
+	})();
+
+	const setEditingMode = (editing: boolean): void => {
+		setEditing(editing);
+	};
+
+	const [formatters, stopFormatterTracker] = (() => {
+		let actions: FormattingButton[] = [];
+
+		const c = Tracker.autorun(() => {
+			actions = formattingButtons.filter(({ condition }) => !condition || condition());
+			emitter.emit('formatting');
+		});
+
+		return [
+			{
+				get: () => actions,
+				subscribe: (callback: () => void) => emitter.on('formatting', callback),
+			},
+			c,
+		];
+	})();
+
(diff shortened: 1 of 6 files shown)


packages/ui-composer/src/MessageComposer/MessageComposerInputNew.tsx
```diff
@@ -0,0 +1,41 @@
...
+interface MessageComposerInputNewProps extends ComponentProps<typeof Box> {}
+
+const MessageComposerInputNew = forwardRef<HTMLTextAreaElement, MessageComposerInputNewProps>(function MessageComposerInputNew(props, ref) {
+	return (
+		<Box is='label' width='full' fontSize={0}>
+			<Box
+				className={[messageComposerInputNewStyle, 'rc-message-box__textarea js-input-message']}
+				color='default'
+				width='full'
+				minHeight='20px'
```
AyushKumar123456789, line 24: why changed this from {52} to '20px' similarly , I dont think there is need to  changed this from {155} to 155?

apps/meteor/client/views/room/composer/messageBox/MessageBoxNew.tsx
```diff
@@ -0,0 +1,482 @@
...
+
+			chat.currentEditing.reset().then((reset) => {
+				if (!reset) {
+					chat.currentEditing?.cancel();
+				}
+			});
+		}
+	};
+
+	const keyboardEventHandler = useEffectEvent((event: KeyboardEvent) => {
```
AyushKumar123456789, line 195: HTMLTextAreaElement is not being since you are using the contenteditable div, not a textarea and  More importantly, HTMLDivElement does not have a .value property. So input.value.trim() will always be "" and cause an error .

AyushKumar123456789 (commented):
I know this is just the initial version of code, but there some anomaly that I found is, when you make a text bold and then you remove the astrisk around it , still show the text in bold in text editor and also there is some issue with the autogrow (I know its commented) but when the input size increase there is some error with the toolbox.

apps/meteor/client/views/room/composer/messageBox/MessageBoxNew.tsx
ishanmitra: Yes, thank you for pointing this out. This is an oversight and could lead to any potential issues later on if you had not caught this. I would first have to investigate why this did not crash anything on my end in the first place.

packages/ui-composer/src/MessageComposer/MessageComposerInputNew.tsx
ishanmitra: Yes, you are correct. The reason I did this was due to errors showing up on my VSCode. This has long been resolved. The next commit will restore this to its original values unless any layout issues arise.

ishanmitra: When `minHeight={20}`, it resembles the height closer to the original textarea
image
When I use the original value, `minHeight={52}`, the height is noticeable taller
image
Both of these editors are blank and the 52px height can be confusing in my opinion.

ishanmitra:
> I know this is just the initial version of code, but there some anomaly that I found that might can help, when you make a text bold and then you remove the astrisk around it , still show the text in bold in text editor and also there is some issue with the autogrow (I know its commented) but when the input size increase there is some error with the toolbox.

I tested the hotkey which is `Ctrl+B` and these are my observations:

When the selection is empty, the asterisks are added. And any text within that asterisk will naturally get bolded after submitting the text. However, if you bold a text with the same shortcut, it shows an actual bold text because that is the behaviour of the `div contenteditable` which I did not anticipate. I believe that some kind of a preventDefault will help solve this issue as we want only the markdown format to update the style.

So a **bold text** on submission does not remain bold since Rocket.Chat composer cannot see the asterisk formatters. It only sees plain text.

This is a really good find. Thank you for pointing that out. The next commit will be focusing on eliminating this issue.

packages/ui-composer/src/MessageComposer/MessageComposerInputNew.tsx
```diff
@@ -0,0 +1,41 @@
...
+const messageComposerInputNewStyle = css`
+	resize: none;
+
+	&::placeholder {
+		color: ${Palette.text['font-annotation']};
+	}
+`;
+
+// eslint-disable-next-line @typescript-eslint/no-empty-interface
+interface MessageComposerInputNewProps extends ComponentProps<typeof Box> {}
```
MartinSchoeler, line 15:
```suggestion
type MessageComposerInputProps = ComponentProps<typeof Box>;
```

apps/meteor/app/ui-message/client/messageBox/newCreateComposerAPI.ts
```diff
@@ -0,0 +1,411 @@
...
+				input.selectionEnd = selectionEnd + endPattern.length; */
+
+				if (!document.execCommand?.('insertText', false, selectedText)) {
+					input.innerText = initText.slice(0, initText.length - startPattern.length) + selectedText + finalText.slice(endPattern.length);
+				}
+
+				/* input.selectionStart = selectionStart - startPattern.length;
+				input.selectionEnd = input.selectionStart + selectedText.length; */
+
+				focus();
```
MartinSchoeler, line 282: Any specific reason to move the focus to happen before the `input` and `change` events here?

apps/meteor/client/views/room/composer/messageBox/MessageBoxNew.tsx
```diff
@@ -0,0 +1,482 @@
+/* eslint-disable complexity */
```
MartinSchoeler, line 1: Let's create at least a TODO for us to not forget to address the complexity issue

MartinSchoeler (changes requested):

apps/meteor/app/ui-message/client/messageBox/newCreateComposerAPI.ts
ishanmitra: Thank you for pointing that out! Honestly, I think there is no specific reason.

I think it happened while copying and pasting some code, which caused some ordering issues. Should have been more careful in my opinion. I will check if shifting `focus();` to L292 (prior to `return;`) causes any issue or not. Otherwise will resolve this quickly.

Additionally I believe it would be best to remove the comments since it would be cluttered correct?

ishanmitra: I have restored the focus to its original position. It has not caused any issues. I am not currently resolving this conversation since I would require confirmation whether I should keep comments or not unless absolutely not required.

apps/meteor/client/views/room/composer/ComposerMessage.tsx
```diff
@@ -89,7 +93,11 @@ const ComposerMessage = ({ tmid, onSend, ...props }: ComposerMessageProps): Reac
 		return <ComposerSkeleton />;
 	}
 
-	return <MessageBox key={room._id} tmid={tmid} {...composerProps} showFormattingTips={true} {...props} />;
+	return featurePreviewComposer ? (
+		<MessageBoxNew key={room._id} tmid={tmid} {...composerProps} showFormattingTips={true} {...props} />
```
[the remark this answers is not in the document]
ishanmitra: Would you suggest that I refactor the following?

`newCreateComposerAPI.ts` ⇒ `createRichTextComposerAPI.tsx`
`MessageBoxNew.tsx` ⇒ `RichTextMessageBox.tsx`
`MessageComposerInputNew.tsx` ⇒ `RichTextComposerInput.tsx`

Additionally, all the native `MessageComposer` exists in `packages/ui-composer/src/MessageComposer`.
Does that mean I need to create a `packages/ui-composer/src/RichTextComposer`?

packages/ui-composer/src/MessageComposer/MessageComposerInputNew.tsx
ishanmitra: Marking this as resolved.

packages/ui-composer/src/MessageComposer/MessageComposerInputNew.tsx
ishanmitra: Currently set to `MessageComposerInputNewProps`, which I will refactor in the refactor commit. Marking this as resolved.

apps/meteor/client/views/room/composer/ComposerMessage.tsx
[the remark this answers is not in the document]
ishanmitra: This is semi-resolved for now. I will fully resolve this when I add the TODO.

apps/meteor/client/views/room/composer/messageBox/MessageBoxNew.tsx
ishanmitra: Your issue is currently being worked on and it has been subdivided into several tasks:
- Modularize the `SelectionRange` API for `HTMLDivElement`
- Change the erroneous `HTMLTextAreaElement` to `HTMLDivElement`, and modify the API calls
-  Reduce redundancy as suggested by Martin.

ishanmitra: Marking this as resolved.

apps/meteor/client/views/room/composer/messageBox/RichTextMessageBox.tsx
```diff
@@ -0,0 +1,563 @@
...
+
+	const newSizes = useContentBoxSize(contentEditableRef);
+
+	const format = useFormatDateAndTime();
+
+	const joinMutation = useMutation({
+		mutationFn: async () => onJoin?.(),
+	});
+
+	const handlePaste = useEffectEvent((event: ClipboardEvent<HTMLDivElement>) => {
```
MartinSchoeler, line 423: We might need to strip the pasted text from styles, as it is today, some styles from the origin of the paste carry over

Screenshot 2025-07-14 at 10 28 57

Screenshot 2025-07-14 at 10 40 07

Screenshot 2025-07-14 at 10 55 10

This was not necessary before, since the MessageBox only supported text, with rich content that is an issue. 

I've tried some rich text that had some xss exploits in it and it didn't work, so we know the content is being sanitized at least

MartinSchoeler (changes requested):

apps/meteor/client/views/room/composer/messageBox/hooks/useMessageBoxAutoFocus.ts
```diff
@@ -29,6 +29,11 @@ export const useMessageBoxAutoFocus = (enabled: boolean): Ref<HTMLElement> => {
 				return;
 			}
 
+			// This prevents focus from being stolen between the two RichText Composers
+			if ((target as HTMLElement).tagName === 'DIV' && (target as HTMLElement).isContentEditable) {
```
MartinSchoeler: Instead of checking for the element type, we could check if the experimental composer is active, using the feature preview flag, this way it is clearer to anyone reading the code why we do this

MartinSchoeler (changes requested):

apps/meteor/app/ui-message/client/messageBox/createRichTextComposerAPI.ts
```diff
@@ -0,0 +1,392 @@
+import type { IMessage } from '@rocket.chat/core-typings';
```
MartinSchoeler, line 1: Would you mind testing an approach like this? 

https://gist.github.com/MartinSchoeler/a4fb74162176ce0f9d1bd32af149cba6

I think we can experiment with extending the existing composerAPI, so we do not have to worry about unrelated code

MartinSchoeler (changes requested):

apps/meteor/app/ui-message/client/messageBox/selectionRange.ts
```diff
@@ -0,0 +1,232 @@
...
+
+	const walker = document.createTreeWalker(input, NodeFilter.SHOW_ALL, {
+		acceptNode: (node) => {
+			if (node.nodeType === Node.TEXT_NODE) return NodeFilter.FILTER_ACCEPT;
+
+			if (node.nodeType === Node.ELEMENT_NODE) {
+				const tag = node.nodeName.toLowerCase();
+
+				// Skip inline tags that don't cause linebreaks
+				if (['b', 'i', 'u', 'span', 'strong', 'em', 'small', 'abbr', 'sub', 'sup', 'mark', 'code', 'del'].includes(tag) && node !== input) {
```
MartinSchoeler, line 19: Use an enum instead of hardcoding the array, it is better for reusability and typechecking

apps/meteor/app/ui-message/client/messageBox/selectionRange.ts
```diff
@@ -0,0 +1,232 @@
...
+				endNode = node;
+				endOffset = selectionEnd - runningOffset;
+			}
+
+			runningOffset += textLength;
+		} else if (node.nodeType === Node.ELEMENT_NODE) {
+			const tag = (node as HTMLElement).tagName.toLowerCase();
+
+			const isInline =
+				['b', 'i', 'u', 'span', 'strong', 'em', 'small', 'abbr', 'sub', 'sup', 'mark', 'code', 'del'].includes(tag) && node !== input;
```
MartinSchoeler, line 120: same here, use an enum

apps/meteor/client/views/room/composer/messageBox/hooks/useMessageBoxAutoFocus.ts
```diff
@@ -29,6 +29,11 @@ export const useMessageBoxAutoFocus = (enabled: boolean): Ref<HTMLElement> => {
 				return;
 			}
 
+			// This prevents focus from being stolen between the two RichText Composers
+			if ((target as HTMLElement).tagName === 'SPAN' && (target as HTMLElement).isContentEditable) {
```
MartinSchoeler, line 33: You can use this hook https://github.com/RocketChat/Rocket.Chat/blob/develop/packages/ui-client/src/hooks/useFeaturePreview.ts 
To verify that the target will be contenteditable 
just pass the parameter `realtimeMessageComposer` to it and from the result, you will know if the composer is content editable or not

package.json
```diff
@@ -70,6 +70,7 @@
 	},
 	"dependencies": {
 		"@types/stream-buffers": "^3.0.7",
+		"hono": "^4.7.9",
```
MartinSchoeler, line 73: I'm pretty sure this change is unrelated to the PR

apps/meteor/app/ui-message/client/messageBox/messageStateHandler.ts
```diff
@@ -0,0 +1,70 @@
...
+
+// Resolve state of the composer during beforeInput
+export const resolveBeforeInput = (
+	event: InputEvent,
+	setMdLines: Dispatch<SetStateAction<string[]>>,
+	setCursorHistory: Dispatch<SetStateAction<CursorHistory>>,
+	parseOptions: Options,
+): void => {
+	const input = event.target as HTMLDivElement;
+
```
MartinSchoeler: If these operations are not ready yet, let's add a well written TODO here

apps/meteor/client/views/room/composer/messageBox/hooks/useMessageBoxAutoFocus.ts
ishanmitra: Got it! I will test and resolve them! I am just in the brink of finishing up some of the code and I would be more than happy to resolve and make the code more readable and logically more stable.

package.json
ishanmitra: Yes, I think it happened due to some package installation way back in May. I do remember this change happening, and committing it to the branch and PR. Although this was the commit it is associated with: a6387918b84d65d25cbddb155673122fae5b1c1c

apps/meteor/app/ui-message/client/messageBox/messageStateHandler.ts
ishanmitra: This is exactly what I have been working on for the past few days! I pushed some commits, so this code should be well-updated by now. There are a few remaining changes I plan to finish as soon as possible.

apps/meteor/app/ui-message/client/messageBox/createRichTextComposerAPI.ts
ishanmitra: Yes, I do agree with you since it is easier to override functions where the RichTextComposer demands it. For now, an explicit copy of the ComposerAPI (as RichTextComposerAPI) gives me a greater degree of freedom to mess around, as the entire code is explicitly visible to me.
Perhaps a comparison between the two files could be done to see what could be best for developers in the long run?

PR #35554, opened 2025-03-19
chore: improve-mocked-provider-settings

ggazzo:

PR #35987, opened 2025-05-13
chore: move tsconfig bases from root to a package

cardoso:
Migrates projects to extend from `@rocket.chat/tsconfig` for consistent TypeScript configurations across the workspace.

This change also updates the Dockerfiles to include the tsconfig package, and adds `@rocket.chat/tsconfig` as a devDependency in multiple package.json files. Furthermore, it updates the TypeScript versions and removes the now redundant tsconfig base files.

## Issue(s)

https://rocketchat.atlassian.net/browse/CORE-1150

## Further comments

I noticed tsserver gets confused with monorepo dependencies attempting to extend the tsconfig bases in the root folder when installed. This is due to the relative paths.

Similarly to eslint-config and jest-presets, this PR moves the tsconfig base files from the root into a separate package, as recommended by turborepo docs:

[TypeScript | Turborepo - Sharing tsconfig.json](https://turbo.build/repo/docs/guides/tools/typescript#sharing-tsconfigjson)

I took inspiration from [turborepo's own tsconfig package](https://github.com/vercel/turborepo/tree/main/packages/tsconfig).

I also updated the hygen templates for completeness, although they seem unused and [hygen is unmaintained](https://github.com/jondot/hygen/issues/455).


packages/tools/tsconfig.json
```diff
@@ -1,9 +1,10 @@
...
-	"extends": "../../tsconfig.base.server.json",
+	"extends": "@rocket.chat/tsconfig/server.json",
 	"compilerOptions": {
 		"declaration": true,
 		"rootDir": "./src",
 		"outDir": "./dist"
 	},
-	"include": ["./src/**/*"]
+	"include": ["./src/**/*"],
+	"exclude": ["./src/**/*.spec.ts"]
```
pierre-lehnen-rc, line 9: why are you adding this exclusion rule to this specific package only?

pierre-lehnen-rc: What I meant was: why are the unit tests of this specific package excluded, while other packages still include their own tests?

cardoso: They don't include their own tests, if they did the tests would be transpiled and `*.spec.js`, `*.spec.d.ts` files would be created.

When the tsconfig files were in the root, extending them was excluding all `*.spec.ts` files.

Some packages already were doing this instead of inheriting from base:
https://github.com/RocketChat/Rocket.Chat/pull/35554/files#diff-4ee07dcf11b8207c9823d05ac6ec91bcf152377cadeb557c4d3ac107eb521f78

pierre-lehnen-rc: They are transpiling to *.spec.js! See the models package for example.
I'd say to either add the `exclude` line to every package in this PR, or don't add it to any here and do it in a separate PR instead.

cardoso: Got it. I went with the `${configDir}` route since I guess it preserves current behavior.

pierre-lehnen-rc (approved):

_templates/package/new/package.json.ejs.t
```diff
@@ -8,17 +8,18 @@ to: packages/<%= name %>/package.json
 	"private": true,
 	"devDependencies": {
 		"@rocket.chat/jest-presets": "workspace:~",
+		"@rocket.chat/tsconfig": "workspace:*",
 		"@types/jest": "~29.5.12",
 		"eslint": "~8.45.0",
 		"jest": "~30.0.0",
-		"typescript": "~5.3.3"
+		"typescript": "~5.7.2"
```
pierre-lehnen-rc:
```suggestion
		"typescript": "~5.8.3"
```
version changed since the PR was opened

_templates/service/new/package.json.ejs.t
```diff
@@ -38,11 +38,12 @@ to: ee/apps/<%= name %>/package.json
 	},
 	"devDependencies": {
 		"@rocket.chat/eslint-config": "workspace:^",
+		"@rocket.chat/tsconfig": "workspace:*",
 		"@types/eslint": "~8.44.0",
 		"@types/polka": "^0.5.4",
 		"eslint": "~8.45.0",
 		"ts-node": "^10.9.1",
-		"typescript": "~5.1.6"
+		"typescript": "~5.7.2"
```
pierre-lehnen-rc:
```suggestion
		"typescript": "~5.8.3"
```
version changed since the PR was opened

pierre-lehnen-rc (changes requested):

PR #35992, opened 2025-05-14
chore: new omni-lib package

sampaiodiego:


ee/packages/omni-lib-ee/src/isDepartmentCreationAvailable.ts
```diff
@@ -0,0 +1,13 @@
+import { License } from '@rocket.chat/license';
+import { isDepartmentCreationAvailable } from '@rocket.chat/omni-lib';
+
+export function isDepartmentCreationAvailablePatch(): void {
+	isDepartmentCreationAvailable.patch(async (next) => {
+		// Skip the standard check when Livechat Enterprise is enabled, as it allows unlimited departments
+		if (License.hasModule('livechat-enterprise')) {
```
KevLehman, line 7: I vote for keeping the license checks out of the patches 👀

ee/packages/omni-lib-ee/src/index.ts
```diff
@@ -0,0 +1,5 @@
+import { isDepartmentCreationAvailablePatch } from './isDepartmentCreationAvailable';
+
+export function patchOmniLib(): void {
```
KevLehman, line 3: Should we expose a function to unpatch as well?

ee/packages/omni-lib-ee/package.json
```diff
@@ -0,0 +1,31 @@
...
+	},
+	"main": "./dist/index.js",
+	"typings": "./dist/index.d.ts",
+	"files": [
+		"/dist"
+	],
+	"dependencies": {
+		"@rocket.chat/models": "workspace:^",
+		"@rocket.chat/omni-lib": "workspace:^",
+		"@rocket.chat/patch-injection": "workspace:^"
```
KevLehman, line 29: do u need this one here?

apps/meteor/ee/server/patches/index.ts
```diff
@@ -1,7 +1,10 @@
...
+
 import './closeBusinessHour';
 import './getInstanceList';
-import './isDepartmentCreationAvailable';
 import './verifyContactChannel';
 import './mergeContacts';
 import './isAgentAvailableToTakeContactInquiry';
 import './airGappedRestrictionsWrapper';
+
+patchOmniLib();
```
KevLehman, line 10: This file is always called, I'd say we should call the Omni patches on the ee/livechat-enteprise folder

wdyt?

ee/packages/omni-lib-ee/package.json
sampaiodiego: maybe not

ee/packages/omni-lib-ee/src/isDepartmentCreationAvailable.ts
sampaiodiego: yeah, need to define a proper way for that.. this one specifically was just moving the file to the package though

apps/meteor/ee/server/patches/index.ts
sampaiodiego: yep, it looks more organized that way

ee/packages/omni-lib-ee/src/index.ts
sampaiodiego: I don't see that feature in the `makeFunction`:

https://github.com/RocketChat/Rocket.Chat/blob/abe83209363e8ea4336afde0c8d6801d2526b5b2/packages/patch-injection/src/makeFunction.ts#L5

KevLehman: The `patch` one has it
```
export const addPatch = <T extends BaseFunction>(baseFunction: T, patch: PatchFunction<T>, condition?: () => boolean) => {
	const patches = getFunctionPatches(baseFunction);

	const patchData: PatchData<T> = {
		patchFunction: patch,
		condition,
	};

	patches.add(patchData);

	return () => {
		patches.delete(patchData);
	};
};
```

KevLehman (approved):

KevLehman (approved):

PR #36007, opened 2025-05-15
fix: remove apps-engine dependency on internal package

d-gubert:

Apps-engine version 1.51.0 was published referencing an internal package that we don't publish

This PR removes that dependency from the apps-engine and applies it in the meteor app

## Further comments

This needs to be released as a patch for version 7.6

[SUP-797]

[SUP-797]: https://rocketchat.atlassian.net/browse/SUP-797?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


d-gubert:
> no changes to `yarn.lock`? 🤔

Rocket.Chat is likely just installing the version in the monorepo, right? Since there was no change to the dependency on the meteor side, I don't think anything would change

sampaiodiego (approved):

PR #36012, opened 2025-05-16
fix: 1.1.1 Added is-tag to ToggleSwitch. Related to change in ToggleSwitch-compo…

ergot-rp:
…nent in fuselage


juliajforesti (changes requested):
Why change a toggle-switch from `input` to a `div`?

PR #36017, opened 2025-05-16
refactor: Remove dependency on translation service from transcript service

KevLehman:
## Issue(s)

https://rocketchat.atlassian.net/browse/CTZ-164


ee/apps/omnichannel-transcript/src/initTranslation.ts
```diff
@@ -0,0 +1,22 @@
+import i18nDict, { availableTranslationNamespaces, defaultTranslationNamespace, extractTranslationNamespaces } from '@rocket.chat/i18n';
+import i18next from 'i18next';
+import sprintf from 'i18next-sprintf-postprocessor';
+
+const i18n = i18next.use(sprintf);
+
+void i18n.init({
```
sampaiodiego, line 7: this will be third place we call `i18n.init` .. I understand the frontend will have different needs, but I was wondering if at least he backend usage could be shared (maybe moving the init to `@rocket.chat/i18n` package).. wdyt?

ee/apps/omnichannel-transcript/src/initTranslation.ts
sampaiodiego, line 1: maybe the file could named just as `i18n.ts`?

ee/packages/omnichannel-services/src/OmnichannelTranscript.ts
```diff
@@ -525,7 +470,7 @@ export class OmnichannelTranscript extends ServiceClass implements IOmnichannelT
 					file: transcriptFile,
 					message: {
 						// Translate from service
-						msg: await translationService.translateToServerLanguage('pdf_success_message'),
+						msg: this.translator.t('pdf_success_message', { lng: serverLanguage, fallbackLng: 'en' }),
```
sampaiodiego, line 473: do we need to provide `fallbackLng` here even though we've already provided at `i18n.init`?

sampaiodiego (changes requested):
a more broad question:

the whole implementation changed just a little bit.. now `omnichannel-transcript` is using `i18n` to return an object of all "known" translation keys so when `pdf-worker` renders the templates it uses this object to find the translated key.

so my question, should `pdf-worker` actually use `i18n` ? so `omnichannel-transcript` doesn't need to pre-translate anything?

ee/apps/omnichannel-transcript/src/initTranslation.ts
KevLehman: Good question. To me, the PDF worker shouldn't know about translations, it should just do what it should do: pdf things. That's subjective however, so, wdyt?

ee/packages/omnichannel-services/src/OmnichannelTranscript.ts
KevLehman, line 473:
```suggestion
						msg: this.translator.t('pdf_success_message', { lng: serverLanguage }),
```

KevLehman: No need tho, i was testing and forgot this one

ee/apps/omnichannel-transcript/src/initTranslation.ts
sampaiodiego: (you might be replying to my global comment, but I'll reply here anyways 😂 )
the point is that it actually knows about it https://github.com/RocketChat/Rocket.Chat/blob/0ff7a54f2e65406e5aef2081622307c0bb324975/ee/packages/pdf-worker/src/strategies/ChatTranscript.ts#L45

KevLehman: > (you might be replying to my global comment, but I'll reply here anyways 😂 ) the point is that it actually knows about it
> 
> https://github.com/RocketChat/Rocket.Chat/blob/0ff7a54f2e65406e5aef2081622307c0bb324975/ee/packages/pdf-worker/src/strategies/ChatTranscript.ts#L45

About this one: that getTranslations func is weird... is not translating anything, its just finding the key on the translations we provide.

We could maybe look at if the pdf worker should handle all the translations, but i'd say it's more of a responsibility of whoever calls it to translate stuff. Or we could also take a look at that "translation" logic inside the pdfworker, cause it's not translating lol, its just finding.

sampaiodiego:
(let move the conversation to "main thread" - [from here](https://github.com/RocketChat/Rocket.Chat/pull/36017#discussion_r2093651947))

> Or we could also take a look at that "translation" logic inside the pdfworker, cause it's not translating lol, its just finding.

right, it is not translating but I wonder if it actually **should** translate things.. it owns the PDF template which *are* multi-language, so it knows about it, it should just know to which language it should translate the things to. if you think about it the responsibilities are very mixed with current implementation:

- `omnichannel-transcript` service:
  - depends on `i18n` (previously on translation service)
  - "knows" about a bunch of strings it *does not use* but translates them
  - sends the translated string as "data" to `pdf-worker` service

- `pdf-worker` service:
  - has the React template with labels that are multi-language
  - receives a big object with a bunch of translated strings
  - uses the mapped translated string on the React templates to render translated strings

it looks weird to me.. imo `pdf-worker` should have the `i18n` dependency and use it to translate things.. the `omnichannel-transcript` service should only send the target language (which would be either the user language or default server language)

ee/apps/omnichannel-transcript/src/initTranslation.ts
sampaiodiego: replied at https://github.com/RocketChat/Rocket.Chat/pull/36017#issuecomment-2887691430

sampaiodiego (approved):
we can discuss about service's responsibilities in a later PR

PR #36026, opened 2025-05-19
test: Automated tests for display mode settings.

Harmeet221:
JIRA Task: https://rocketchat.atlassian.net/browse/QA-75

Automated following test cases:-

1. expect to select display mode as Extendedand verify if sidebar item contains a message.
2. expect to select display mode as Medium' and verify if sidebar item does not contain a message
3. expect to select display mode as condensed and verify if sidebar item is condensed and its size is reduced to 28
4. expect to display Avatars 
5. expect to hide Avatars.


gabriellsh (changes requested):
Hey nice work! It's always a joy to see more tests be implemented.

I have a small suggestion to make though. I see that for each step, there's a `test` clause, and a `test.beforeEach` that navigates to home. That navigation is necessary (and you might now this already) because each `test` clause opens a new browser window and has to load the application from the beggining.

I see that you added this tests as `parallel` which is already a very good improvement regarding the time these tests will take to run. Very good catch!

What I was going to suggest though, is that instead of making a `test` for each test, I think you could get away with a single `test` clause for ALL the steps, and you can separate them by using `test.step`. The tests will be less "parallelized" because they'll run all in the same browser window, but it'll also not require the application to load from scratch every time (which for simple tests like these is what takes the most time from the CI).

There are benefits and disadvantages for both, but for this case, if we use `test.step` we'll be better testing the UI reactivity, since there are differences from waiting on a UI change (regular use) and reopening the page, because everything will be recalculated.

apps/meteor/tests/e2e/page-objects/fragments/home-sidenav.ts
dougfabris, line 1: please lets use accessible locators and we do already have page objects for some of these locators in other files

dougfabris (changes requested):

apps/meteor/tests/e2e/display-mode.spec.ts
```diff
@@ -0,0 +1,64 @@
...
+	});
+
+	test.beforeEach(async ({ page }) => {
+		poHomeChannel = new HomeChannel(page);
+
+		await page.goto('/home');
+	});
+
+	test.afterAll(async ({ api }) => {
+		await deleteChannel(api, channelName);
```
ggazzo, line 23: I think the preference should be reseted,  even though  one step resets partially the state, but if fails next steps would inherit non default settings

ggazzo (changes requested):

PR #36027, opened 2025-05-19
fix: `audit.settings` endpoint filtering by actor returning empty results

gabriellsh:

The filtering by actor was not working, and 2 issues were identified:
1. The tests for the endpoint were not properly testing the endpoint. Firstly, neither the `end` test function was being called at the end of the test, nor were the promises being awaited. This means every possible error was being ommited. Also in the filtering tests, the result was only being asserted by verifying if the result was an array, and not ensuring that the results match the passed query.
2. The `actor` filter arrives in the endpoint function as an object, and being directly passed to the query. When passing objects in a mongo query, it only matches if there's a full match between the objects. This means that passing `actor: { type: 'user' }` returns no records, because `actor` has many more properties, thus not an exact match.

Firstly the tests were improved by awaiting the promises, and adding extra assertions, guaranteeing that they are failing with the current implementation, and that the filters are working properly.

Then the issue was fixed by mapping the actor object properties to a proper mongo query, transforming `actor: { type: 'user', _id: 'xxx' }` to `{ 'actor.type': 'user', 'actor._id': 'xxx' }`.

This PR should fix the issue, and ensure it doesn't happen again due to fixed/improved tests.
## Issue(s)

[CORE-1159](https://rocketchat.atlassian.net/browse/CORE-1159)
## Further comments

[CORE-1159]: https://rocketchat.atlassian.net/browse/CORE-1159?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/meteor/tests/end-to-end/api/settings.ts
```diff
@@ -276,26 +276,34 @@ describe('[Settings]', () => {
...
 	(IS_EE ? describe : describe.skip)('/audit.settings', () => {
-		const formatDate = (date: Date) => date.toISOString().slice(0, 10).replace(/-/g, '/');
+		const formatDate = (date: Date) => date.toISOString();
+		let startDate: Date;
+		let endDate: Date;
+
+		before(async () => {
+			// Change a few settings so we can test filtering
+			startDate = new Date();
+			await Promise.all([updateSetting('Accounts_AllowAnonymousRead', true), updateSetting('Accounts_AllowDeleteOwnAccount', true)]);
```
pierre-lehnen-rc, line 286: changing the setting and then changing it back right afterwards?

gabriellsh: This is just to trigger logs so that I can properly test the filters. That's why I revert them immediately, I don't need these settings, it's just to add log records to the server-events collection.

pierre-lehnen-rc (approved):

PR #36035, opened 2025-05-20
fix: Support file filtering across all room types endpoints

gabriellsh:

Currently only the endpoint `channels.files` supports file filtering by passing either `name` or `typeGroup`. Previously this behaviour was supported with the `query` parameter which is currently deprecated.

In this PR https://github.com/RocketChat/Rocket.Chat/pull/33564, `name` and `typeGroup` were added as query parameters for the `channels.files` endpoint, and the FE logic changed to filter by these parameters. The issue is that other types of rooms, such as private channels and DMs, have separate endpoints, which were not changed to support the new query parameters. This makes the filtering in the files contextual bar unusable for any room type that is not a public channel.

This PR introduces the support on the aforementioned endpoints, and also implements tests for the filtering to ensure the behaviour is consistent.
## Issue(s)

[SUP-799](https://rocketchat.atlassian.net/browse/SUP-799)
## Further comments

[SUP-799]: https://rocketchat.atlassian.net/browse/SUP-799?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/meteor/app/api/server/v1/groups.ts
```diff
@@ -389,21 +389,28 @@ API.v1.addRoute(
...
 				checkedArchived: false,
 			});
 
 			const { offset, count } = await getPaginationItems(this.queryParams);
 			const { sort, fields, query } = await this.parseJsonQuery();
 
-			const ourQuery = Object.assign({}, query, { rid: findResult.rid });
+			const filter = {
+				rid: findResult.rid,
+				...query,
```
abhinavkrin, line 408: The query object is spread after setting rid. If the query also has an rid property, it will override the previously set rid. Hence, I am concerned it can be used to fetch files for arbitrary rooms irrespective of the passed roomId query param.

apps/meteor/app/api/server/v1/im.ts
```diff
@@ -234,19 +225,26 @@ API.v1.addRoute(
...
 
 			const canAccess = await canAccessRoomIdAsync(room._id, this.userId);
 			if (!canAccess) {
 				return API.v1.forbidden();
 			}
 
-			const ourQuery = query ? { rid: room._id, ...query } : { rid: room._id };
+			const filter = {
+				rid: room._id,
+				...query,
```
abhinavkrin, line 242: same.

abhinavkrin (changes requested):

apps/meteor/app/api/server/v1/groups.ts
gabriellsh: That's true, and it was also copied from another endpoint (channels.files) which the same thing happens. Issue is that we also need to find by the queryparams in this case. I'll see what I can do. Thanks!

gabriellsh: I just confused myself a bit. Just moved the spread, thanks for the review man!

MarcosSpessatto (approved):

PR #36046, opened 2025-05-21
chore: Detach idle detection from component state

tassoevan:

It doesn't store a boolean identifying the idle state in a component state to avoid spurious re-renders of `UserProvider`.


yarn.lock
```diff
@@ -8407,7 +8407,7 @@ __metadata:
     storybook-dark-mode: "npm:^4.0.2"
     typescript: "npm:~5.7.2"
   peerDependencies:
-    "@rocket.chat/apps-engine": 1.51.0
+    "@rocket.chat/apps-engine": 1.52.0-rc.0
```
ggazzo, line 8410: non related I believe

ggazzo (changes requested):

ggazzo (approved):

PR #36059, opened 2025-05-22
regression: broken incoming webhook validation

abhinavkrin:

This PR fixes an issue which made the validation to not work for incoming webhooks in public rooms.
## Steps to test or reproduce

1. Create an incoming webhook with a public room.
2. Call the webhook endpoint with the appropriate message payload. The webhook url last part should be modified to simulate invalid token.
3. Check that the message is not sent to the room
## Further comments

[CORE-1161](https://rocketchat.atlassian.net/browse/CORE-1161)

### DIFF
--- a/packages/model-typings/src/models/IIntegrationsModel.ts
+++ b/packages/model-typings/src/models/IIntegrationsModel.ts
@@ -1,5 +1,5 @@
 import type { IIntegration, IUser } from '@rocket.chat/core-typings';
-import type { FindCursor } from 'mongodb';
+import type { FindCursor, WithId } from 'mongodb';
 
 import type { IBaseModel } from './IBaseModel';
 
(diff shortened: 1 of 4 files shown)


packages/model-typings/src/models/IIntegrationsModel.ts
```diff
@@ -10,4 +10,6 @@ export interface IIntegrationsModel extends IBaseModel<IIntegration> {
 	findOneByIdAndCreatedByIfExists(params: { _id: IIntegration['_id']; createdBy?: IUser['_id'] }): Promise<IIntegration | null>;
 	findOneByUrl(url: string): Promise<IIntegration | null>;
 	updateRoomName(oldRoomName: string, newRoomName: string): ReturnType<IBaseModel<IIntegration>['updateMany']>;
+	findOneByIdAndToken(id: IIntegration['_id'], token: string): Promise<WithId<IIntegration> | null>;
```
julio-rocketchat, line 13: Is it duplicated? I see two "findOneByIdAndToken" here

abhinavkrin: Thats just typing. Only one function exists.

abhinavkrin: If we call findOneByIdAndValue, the type return is of IIntegration. We can explicitly mention if the return type should be IIncomingIntegration or IOutgoingIntegration.

julio-rocketchat (approved):
LGTM

MarcosSpessatto (approved):

PR #36079, opened 2025-05-26
test: menu disappears once page finishes loading

cardoso:
## Issue(s)

https://rocketchat.atlassian.net/browse/CORE-1164


apps/meteor/client/views/root/DocumentTitleWrapper.tsx
```diff
@@ -13,7 +13,14 @@ const useRouteTitleFocus = () => {
...
 
-		node.focus();
+		// If the parent element has aria-hidden set to true, it means the element is not visible to screen readers.
+		// This is a common practice to hide elements that are not currently in use or are not relevant to the user.
+		// For example, a modal or popover might be open, and the main content is hidden to keep the user's focus on the modal.
+		// Since the title is marked with `tabIndex={-1}`, it is not reachable by keyboard navigation.
+		// https://accessibilityinsights.io/info-examples/web/aria-hidden-focus/
+		if (node.parentElement?.ariaHidden !== 'true') {
+			node.focus();
+		}
```
cardoso, line 23: @gabriellsh does this make sense?

I'll improve the test once this solution is validated.

cardoso: For reference, the parentElement is the react root, and aria-hidden is set to true when a popover or modal is open.
image

gabriellsh:
> Please don't forget to add a changeset

Maybe we should change this pr to `test:`, because the issue and changes made are not noticeable by the end-user. There might be some edge-case scenario where the issue happens, but it's very hard to bump into it, even when actively trying to do so. @cardoso wdyt? Maybe let's change the PR to "chore" or "test", I don't think we need to include this in the changelog.

gabriellsh (approved):
LGTM!

PR #36088, opened 2025-05-27
fix: 1.1.1: Changed <FieldLabel> to <Box is='legend'> to avoid 'Orphaned form label'

ergot-rp:


ergot-rp:
handeled in PR #36525

PR #36095, opened 2025-05-27
fix: avoid font-size tag being created multiple times

juliajforesti:

The hook to create the font-size tag was being called multiple times inside UserProvider:
- remove `useCreateFontStyleElement` call from `UserProvider`
- create `useFontStylePreference` and call it in `LoggedInArea`

### DIFF
--- a/apps/meteor/client/hooks/useFontStylePreference.tsx
+++ b/apps/meteor/client/hooks/useFontStylePreference.tsx
@@ -1,0 +1,16 @@
+import { useUserPreference } from '@rocket.chat/ui-contexts';
+import { useLayoutEffect } from 'react';
+
+import { useCreateFontStyleElement } from '../views/account/accessibility/hooks/useCreateFontStyleElement';
+
+export const useFontStylePreference = () => {
+	const fontSize = useUserPreference('fontSize', '100%');
+	const createFontStyleElement = useCreateFontStyleElement();
+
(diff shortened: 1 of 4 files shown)


dougfabris (approved):
LGTM!

apps/meteor/client/hooks/useFontStylePreference.tsx
```diff
@@ -0,0 +1,16 @@
...
+
+export const useFontStylePreference = () => {
+	const fontSize = useUserPreference('fontSize', '100%');
+	const createFontStyleElement = useCreateFontStyleElement();
+
+	useLayoutEffect(() => {
+		if (!fontSize) {
+			return;
+		}
+		createFontStyleElement(fontSize);
```
ggazzo, line 14: What happens if the preference changes?

ggazzo (changes requested):

apps/meteor/client/hooks/useFontStylePreference.tsx
juliajforesti: this is handled on the Accessibility page, it overwrites the css value of `font-size`.
it's not creating duplicates, but I could add the condition to query the element by id `rcx-font-size` before creating it:
```
const styleElement = document.getElementById('rcx-font-size') || createStyleElement('rcx-font-size');
```
wdyt?

PR #35981, opened 2025-05-13
chore: deprecates dbWatchers  and set default value to false

ggazzo:
https://rocketchat.atlassian.net/browse/ARCH-1602

## Further comments

---

This pull request updates the Rocket.Chat repository by modifying the default behavior of database watchers. The changes are made in the `packages/models/src/index.ts` file. The logic is adjusted so that database watchers are now disabled by default in more scenarios, particularly in production environments where the `DISABLE_DB_WATCHERS` environment variable is not set or is not explicitly set to 'no' or 'false'. Additionally, a console warning is introduced to alert users when watchers are active, highlighting their deprecation. The source branch for this change is `chore/deprecate-streamhub-oplog-changestreams`, and it targets the `develop` branch.

PR #36099, opened 2025-05-28
fix: ui does not update after pruning only files

abhinavkrin:

This PR resolves a regression where the UI fails to update reactively after pruning only files using the “Prune messages” feature. Users must currently reload the page manually to see the deleted files disappear, a behavior that deviates from version 7.6.0, where the UI responded automatically.

The issue is caused from the lack of an explicit broadcast when `filesOnly` is true—unlike full message pruning, which uses the `deleteMessageBulk` event to notify clients. To address this, a new `deleteFilesBulk` event has been added to handle these cases and ensure client-side updates.

A contributing factor to this regression was a recent [change](https://github.com/RocketChat/Rocket.Chat/pull/35981) that disables DB watchers by default in production environments. This change, most probably, prevented sending `stream-room-messages` event automatically (due to absense of explicit broadcast)

Together, these adjustments restore reactive file pruning functionality and improve consistency between environments and versions.
## Steps to test or reproduce

1. Go to a channel.
2. Send an image to the channel.
3. Open “Prune messages”.
4. Set “Newer than” to the current date.
5. Select “Only remove the attached files, keep messages”.
6. Confirm and execute the action.
7. Observe the UI now updates without requiring a reload.
## Further comments

[CORE-1168](https://rocketchat.atlassian.net/browse/CORE-1168)

[CORE-1168]: https://rocketchat.atlassian.net/browse/CORE-1168?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

**UPDATE**
- Instead of a new event, extended the existing `deleteMessageBulk` event to cover `filesOnly` pruning cases.

### DIFF
--- a/apps/meteor/app/ui-utils/client/lib/LegacyRoomManager.ts
+++ b/apps/meteor/app/ui-utils/client/lib/LegacyRoomManager.ts
@@ -8,6 +8,7 @@ import { RoomManager } from '../../../../client/lib/RoomManager';
 import { roomCoordinator } from '../../../../client/lib/rooms/roomCoordinator';
 import { fireGlobalEvent } from '../../../../client/lib/utils/fireGlobalEvent';
 import { getConfig } from '../../../../client/lib/utils/getConfig';
+import { modifyMessageOnFilesDelete } from '../../../../client/lib/utils/modifyMessageOnFilesDelete';
 import { callbacks } from '../../../../lib/callbacks';
 import { Messages, Subscriptions } from '../../../models/client';
 import { sdk } from '../../../utils/client/lib/SDKClient';
@@ -110,6 +111,40 @@ function getOpenedRoomByRid(rid: IRoom['_id']) {
 		.find((openedRoom) => openedRoom.rid === rid);
 }
 
+function createDeleteQuery({
+	excludePinned,
+	ignoreDiscussion,
+	rid,
+	ts,
+	users,
+	ids,
+}: {
+	rid: IMessage['rid'];
+	excludePinned: boolean;
+	ignoreDiscussion: boolean;
+	ts: Record<string, Date>;
+	users: string[];
+	ids?: string[];
+}) {
+	const query: Filter<IMessage> = { rid };
+
+	if (ids) {
+		query._id = { $in: ids };
+	} else {
+		query.ts = ts;
+	}
+	if (excludePinned) {
+		query.pinned = { $ne: true };
+	}
+	if (ignoreDiscussion) {
+		query.drid = { $exists: false };
+	}
+	if (users?.length) {
+		query['u.username'] = { $in: users };
+	}
+
+	return query;
+}
 const openRoom = (typeName: string, record: OpenedRoom) => {
 	if (record.ready === true && record.streamActive === true) {
 		return;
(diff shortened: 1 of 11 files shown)


gabriellsh:
> I noticed that this behavior changes based on `dbWatchersDisabled`. If set to false, the file pruning is reactive, set to true, it stops working. Recently this flag was changed from default false to default true (which could be the reason for the regression). For the CI, the flag is still set to false, this means that the implemented e2e tests would pass even without the fix (not really testing the fix).

Retification: The flag was also inverted in the CI, you can just disregard this comment

abhinavkrin:
> do you know when this regression was introduced? I'm finding a lot of changes for a regression and too much to solve this problem

Hey, @ggazzo , After debugging this, we identified a two-step root cause:

1. The `filesOnly` path in `cleanRoomHistory` was never designed to broadcast any events for reactivity. It silently relied on DB watchers to pick up the changes. So technically, this wasn't a regression in logic—it was a missing feature all along.

2. This gap was only surfaced recently when DB watchers were disabled by default in production via PR [#35981](https://github.com/RocketChat/Rocket.Chat/pull/35981). Prior to that, the watchers filled in the gap and made the UI appear reactive, masking the problem.

So while this behavior has likely existed for a long time, the regression (as experienced by users) was triggered when we disabled DB watchers by default. We’re fixing it now by explicitly adding a `deleteFilesBulk` event broadcast to ensure proper reactivity regardless of watcher settings.

Hope this clarifies the context!

apps/meteor/client/lib/utils/modifyMessageOnFilesDelete.ts
```diff
@@ -0,0 +1,35 @@
...
+ * - If `replaceFileAttachmentsWith` is given, swaps out any file-type
+ *   attachments; otherwise filters them out entirely.
+ *
+ * @param message  The original `IMessage` to copy.
+ * @param replaceFileAttachmentsWith
+ *   Optional attachment to substitute for each file attachment.
+ * @returns A new message object with `file`, `files`, and `attachments` updated.
+ */
+
+export const modifyMessageOnFilesDelete = <T extends IMessage = IMessage>(message: T, replaceFileAttachmentsWith?: MessageAttachment) => {
```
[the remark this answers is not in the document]
abhinavkrin: Added!

MarcosSpessatto (approved):

MarcosSpessatto (dismissed):

apps/meteor/client/hooks/lists/useStreamUpdatesForMessageList.ts
```diff
@@ -59,6 +67,13 @@ export const useStreamUpdatesForMessageList = (messageList: MessageList, uid: IU
 
 		const unsubscribeFromDeleteMessageBulk = subscribeToNotifyRoom(`${rid}/deleteMessageBulk`, (params) => {
 			const matchDeleteCriteria = createDeleteCriteria(params);
+			if (params.filesOnly) {
+				const items = messageList.items.filter(matchDeleteCriteria).map((message) => {
+					return modifyMessageOnFilesDelete(message, params.replaceFileAttachmentsWith);
+				});
+
+				return messageList.batchHandle(() => Promise.resolve({ items }));
+			}
```
gabriellsh: The MessageList.batchHandle method does some work in the background to synchronize updates. By leaving the map of the messageList items outside the callback, there could be a race condition, causing already modified data to be reintroduced into the list.
```suggestion
			if (params.filesOnly) {
				return messageList.batchHandle(async () => {
					const items = messageList.items.filter(matchDeleteCriteria).map((message) => {
						return modifyMessageOnFilesDelete(message, params.replaceFileAttachmentsWith);
					});
					return { items };
				});
			}
```

apps/meteor/client/lib/utils/modifyMessageOnFilesDelete.ts
```diff
@@ -0,0 +1,35 @@
...
+	const updated = structuredClone(message);
```
gabriellsh: I'm not sure if we should be using structuredClone here. I think just returning a spread shoul be sufficient.
```ts
if (attachments && attachments.length > 0) {
  return { ...message, attachments }
}
return message;
```

I also added that `if` statement there, because this method is also being used to modify messages in the `zustand` store. Zustand itself provides a way through `useShallow` to test if the record/object has changed when using the store through react. By returning the same object if no changes are made, we ensure that no unnecessary re-renders will happen.

You could also do an early return of the message in case it doesn't have any attachments, it shouldn't happen since messages are already filtered when they reach this function, but would be a small improvement.

apps/meteor/app/ui-utils/client/lib/LegacyRoomManager.ts
```diff
@@ -148,44 +184,40 @@ const computation = Tracker.autorun(() => {
...
+									(record) => record._id === msg._id,
+									({ file: _, ...record }) => ({
+										...record,
+										files,
+										attachments,
+									}),
+								);
 							}
+							return;
+						}
```
gabriellsh:
```suggestion
						if (params.filesOnly) {
							return Messages.state.update(predicate, (record) => (modifyMessageOnFilesDelete(record, params.replaceFileAttachmentsWith)));
						}
```

wouldn't this have the same effect? It would prevent us from searching the messages twice.

gabriellsh (approved):
Front-end LGTM!

gabriellsh (approved):

MarcosSpessatto (approved):

PR #36105, opened 2025-05-28
fix: improve error handling no ddp-streamer

sampaiodiego:
https://rocketchat.atlassian.net/browse/ARCH-1693

### DIFF
--- a/ee/apps/ddp-streamer/src/service.ts
+++ b/ee/apps/ddp-streamer/src/service.ts
@@ -32,3 +32,39 @@ import { startTracing } from '@rocket.chat/tracing';
 
 	await api.start();
 })();
+
+/**
+ * If some promise is rejected and doesn't have a catch (unhandledRejection) it may cause the process to exit.
+ *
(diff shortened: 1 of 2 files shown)


ggazzo (changes requested):
what about adding a global handler as we do for the main process?

MarcosSpessatto (approved):

ee/apps/ddp-streamer/src/service.ts
```diff
@@ -32,3 +32,39 @@ import { startTracing } from '@rocket.chat/tracing';
...
+	);
+	console.error('Future node.js versions will automatically exit the process');
+	console.error('=================================');
+
+	if (process.env.TEST_MODE || process.env.NODE_ENV === 'development' || process.env.EXIT_UNHANDLEDPROMISEREJECTION) {
+		process.exit(1);
+	}
+});
+
+process.on('uncaughtException', async (error) => {
```
cardoso, line 60: Why the async here? 🤔

PR #36097, opened 2025-05-27
fix: `contact.conflictingFields` being updated with nullish values

lucas-a-pelegrino:
## Proposed changes (including videos or screenshots)

7.6 introduced a bug where updating  `contact.customFields` would also update `contact.conflictingFields` with `null`. This PR fixes that bug by adding a validation that ensures that `conflictingFields` are updated only when appropriate.

## Issue(s)

[CTZ-179](https://rocketchat.atlassian.net/browse/CTZ-179)

## Further comments

[CTZ-179]: https://rocketchat.atlassian.net/browse/CTZ-179?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #36116, opened 2025-05-30
fix: contact being marked as known when updating custom fields or conflicts

lucas-a-pelegrino:

This PR adds a new model function (`.updateContactCustomFields`) that allows updating contact custom fields or adding conflicting fields without overlapping other business rules.

## Issue(s)

[CTZ-179](https://rocketchat.atlassian.net/browse/CTZ-179)

## Further comments

This PR is a additional improvement to #36097 

[CTZ-179]: https://rocketchat.atlassian.net/browse/CTZ-179?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/models/src/models/LivechatContacts.ts
```diff
@@ -126,6 +127,26 @@ export class LivechatContactsRaw extends BaseRaw<ILivechatContact> implements IL
...
+
+		const updatedResult = await this.findOneAndUpdate(
+			{ _id: contactId },
+			{
+				$set: { ...dataToUpdate },
+			},
+			{ returnDocument: 'after', ...options },
+		);
+
+		return updatedResult as ILivechatContact;
```
ggazzo, line 147:
```suggestion
		return this.findOneAndUpdate(
			{ _id: contactId },
			{
				$set: { ...dataToUpdate },
			},
			{ returnDocument: 'after', ...options },
		);
```

ggazzo (dismissed):
What would be the regression in this case?

lucas-a-pelegrino:
> What would be the regression in this case?

Hello, this regression was added as an additional improvement for original PR fixing a bug with contacts conflicting fields being saved with nullish values. 

You can check the full fix here:  #36097

packages/models/src/models/LivechatContacts.ts
sampaiodiego, line 147: pls remove the type cast and fix the return type. 

typecasts should be avoided as a general rule.. in this case for example, what happens if `contactId` is not found? it **will** return a `null`, but since you typed the function as `Promise<ILivechatContact>` and used the type casting to adhere to the return type you're creating a potential bug in your code.

sampaiodiego (changes requested):

packages/models/src/models/LivechatContacts.ts
sampaiodiego: > Small note: I followed the file's current pattern, though.

if you've found other places doing the same, pls go ahead and create a tech debt tasks so we can get rid of them.. we should always be looking into improving the code and not replicating bad code.

sampaiodiego (approved):

PR #36118, opened 2025-06-02
refactor: Replace the deprecated `substr` function with `slice` in the `processWebhookMessage` function.

ahmed-n-abdeltwab:

Replace the deprecated `substr` function with the more modern and preferred `slice` function in the `processWebhookMessage` function to ensure better compatibility and maintainability of the code. Adjust the parameters accordingly to achieve the same substring extraction logic as the original implementation.


apps/meteor/app/lib/server/functions/processWebhookMessage.ts
```diff
@@ -48,7 +48,7 @@ export const processWebhookMessage = async function (
 	for await (const channel of channels) {
 		const channelType = channel[0];
 
-		let channelValue = channel.substr(1);
+		let channelValue = channel.slice(1);
```
cardoso, line 51: substr and slice with a single argument have the same behavior. ✅

cardoso (approved):
Looks good to me 👍

Issue #35661, opened 2025-03-31
Leave button for channels is available even if the user isn't subscribed to the channel


ppl-call-me-tima:
### Description:
The room / channel information page that allows the user to see the channel avatar, hide or leave the channel also provides the Leave option visible to the user, even when they haven't joined / subscribed to the channel.

### Steps to reproduce:
1. Navigate to a channel that you aren't a part of:
    - go to Directory
    - from the list, select a channel you aren't a part of
2. click on the channel name / Room Information button
3. the Leave button is available for a click

### Expected behavior:
The Leave button shouldn't be available as it is not useful to the user who hasn't subscribed to the room in the first place.

### Actual behavior:
Leave button is visible along with the Join button as well:

![Image](https://github.com/user-attachments/assets/9be1241d-7de3-44d7-a8e1-bb47a5d7acae)

### Server Setup Information:

- Version of Rocket.Chat Server: 7.5.0-develop
- License Type: Starter
- Number of Users: 2
- Operating System: Linux x64  [5.15.167.4-microsoft-standard-WSL2]
- Deployment Method: tar
- Number of Running Instances:  1
- DB Replicaset Oplog: true
- NodeJS Version: v22.13.1
- MongoDB Version: 7.0.5

### Client Setup Information

- Browser Version: Chrome 134.0.6998.178
- Operating System: Windows 11 x64

### Further Discussion

Fix proposed: #35670

Swatibharti46:
@ppl-call-me-tima I would like to work on this issue

The-Yearly:
Hey If This issue is still open i would like to work on this

PR #36128, opened 2025-06-02
fix: Leave button for channels is available even if the user isn't subscribed to the channel #35661

The-Yearly:
[Screencast_20250603_034616.webm](https://github.com/user-attachments/assets/06f7cd0a-808f-4c4e-8bd9-5908b28b5f2d)

## Issue(s)

This PR addresses issue [#35661](https://github.com/RocketChat/Rocket.Chat/issues/35661), where the "Leave" button was incorrectly shown for users who are not subscribed to a channel.

The fix ensures that the "Leave" button only appears when the user is actually a member of the channel, preventing confusion and unintended UI behavior.

## Steps to test or reproduce

Open any public or private channel that the user is not subscribed to.

Observe that the "Leave" button does not appear anymore.

Join the channel and verify that the "Leave" button becomes visible after joining.
## Further comments

This fix ensures that the "Leave" button is only shown when the user is actually a member of the channel. By doing so, it prevents confusion and unintended behavior, and it maintains consistency between the user's state and available actions.


The-Yearly:
@ggazzo could pls review this and inform me if changes are required

Issue #36121, opened 2025-06-02
Display name in User info panel


juzser:
### Description:

I see there should be an option to display the full name of user in the info panel, even though the setup `Use Real name` in User interfaces setting is disabled. 

Because normally a workspace will use username to display the users in channels, teams,... but in the user information panel, still should there is Real name field displayed. 

### Steps to reproduce:

1. Layout > User Interfaces > Disable `Use Real Name` setting.
![Image](https://github.com/user-attachments/assets/fe518214-5bdf-44f0-af93-758b765fc020)

2. No `Name` field displayed
![Image](https://github.com/user-attachments/assets/5551a457-8b8a-4ecd-b52e-dde0f1239e06)

### Expected behavior:

All channels, groups, teams,... display username as the setting. 
Display `Name` field in User info panel. 

### Server Setup Information:

- Version of Rocket.Chat Server: 7.5.1
- License Type: Enterprise
- Number of Users: 60
- Operating System: CentOS
- Deployment Method: Docker
- Number of Running Instances: 1
- NodeJS Version: v22.13.1
- MongoDB Version: 6.0.19 / wiredTiger (oplog Enabled)

Harshith-2208:
Hi can you assign me this issue, I want  to fix it.

reetp:
> Hi can you assign me this issue, I want  to fix it.
> 

No one assigns issues.

Just do your PR.

PR #36133, opened 2025-06-03
feat: displays name in User info

Harshith-2208:

I have added two files UserInfoName.tsx and UserCardName.tsx to set props and create a new card for Name of the user and added it in the user info panel in UserInfo.tsx and changed index.ts in UserInfo to export the newly created UserInfoName.tsx and also index.ts in UserCard to export the newly created file UserCardName.tsx . Now the user info panel shows the name in channels, groups and teams.

Screenshot
Screenshot

## Issue(s)

This Closes the issue #36121 

## Steps to test or reproduce

The screen which is going to change if the changes are applied is User Info Panel

## Further comments

This is a small change and doesn't need further comments.


Harshith-2208:
Will this be reviewed?

PR #36143, opened 2025-06-03
fix: Quote chaining in E2EE room not limiting quote depth

gabriellsh:

This issue occurs because we only enforce the quote chaining limit on the server, but the server cannot evaluate the message or its content due to the message being encrypted.

Due to this constraint, the limit can only be applied client-side before the message is sent, and that is what was done. This also improves UX since previously the quote preview did not evaluate the chain limit, causing a mismatch between what showed in the preview and what was actually experienced once the message was sent.

Caveats:
- as said before there's no way to prevent exceeding the quote chain limit for encrypted messages in the backend, because that would require encryption to be broken, and no longer end-to-end. Due to this it is still possible to send messages through the API that exceed said limit. There's no way around this.
- Since we currently only apply this limit server-side, messages sent before the limit is decreased still display all quotes as they were when the message was sent (change to the setting does not modify messages retroactively). Due to this, no limit to quotes was applied to the message layout, as to not break expected behavior. This means that encrypted messages sent through the APIs can display quote chains bigger thant the limit (see previous caveat).

## Issue(s)

[SUP-782](https://rocketchat.atlassian.net/browse/SUP-782)
## Further comments

It's not clear whether the message layout should also respect the limit, but historically it has not, so no attempt was or will made to change the message layout at this point. If this changes at any point, a new PR is warranted.

### DIFF
--- a/apps/meteor/client/components/message/toolbar/items/actions/QuoteMessageAction.tsx
+++ b/apps/meteor/client/components/message/toolbar/items/actions/QuoteMessageAction.tsx
@@ -1,4 +1,5 @@
 import type { ITranslatedMessage, IMessage, ISubscription } from '@rocket.chat/core-typings';
+import { useSetting } from '@rocket.chat/ui-contexts';
 import { useTranslation } from 'react-i18next';
 
 import { useChat } from '../../../../../views/room/contexts/ChatContext';
@@ -13,6 +14,7 @@ type QuoteMessageActionProps = {
 const QuoteMessageAction = ({ message, subscription }: QuoteMessageActionProps) => {
 	const chat = useChat();
 	const autoTranslateOptions = useMessageListAutoTranslate();
+	const quoteChainLimit = Number(useSetting('Message_QuoteChainLimit'));
 	const { t } = useTranslation();
 
 	if (!chat || !subscription) {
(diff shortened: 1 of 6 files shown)


apps/meteor/client/components/message/toolbar/items/actions/QuoteMessageAction.tsx
```diff
@@ -13,6 +14,7 @@ type QuoteMessageActionProps = {
 const QuoteMessageAction = ({ message, subscription }: QuoteMessageActionProps) => {
 	const chat = useChat();
 	const autoTranslateOptions = useMessageListAutoTranslate();
+	const quoteChainLimit = Number(useSetting('Message_QuoteChainLimit'));
```
tassoevan:
```suggestion
	const quoteChainLimit = useSetting('Message_QuoteChainLimit', 2);
```

apps/meteor/client/lib/chats/ChatAPI.ts
```diff
@@ -39,7 +39,7 @@ export type ComposerAPI = {
 	setCursorToEnd(): void;
 	setCursorToStart(): void;
 	replyWith(text: string): Promise<void>;
-	quoteMessage(message: IMessage): Promise<void>;
+	quoteMessage(message: IMessage, quoteChainLimit?: number): Promise<void>;
```
tassoevan: Not sure this is a good idea because I don't know why can't we encapsulate this rule under the `quoteMessage` method, without exposing `quoteChainLimit` as a dependency. The relevant setting can be fetched at `ChatProvider`.

tassoevan (approved):

PR #36146, opened 2025-06-04
chore: deprecate `getUserRoles` `saveCustomFields` `setReaction` `setUsername` `saveUserProfile` `setUserPassword` methods

ggazzo:
https://rocketchat.atlassian.net/browse/ARCH-1702
https://rocketchat.atlassian.net/browse/ARCH-1703
https://rocketchat.atlassian.net/browse/ARCH-1709
https://rocketchat.atlassian.net/browse/ARCH-1710
https://rocketchat.atlassian.net/browse/ARCH-1711
https://rocketchat.atlassian.net/browse/ARCH-1712
https://rocketchat.atlassian.net/browse/ARCH-1713

https://rocketchat.atlassian.net/browse/SUP-804

### DIFF
--- a/apps/meteor/app/api/server/v1/channels.ts
+++ b/apps/meteor/app/api/server/v1/channels.ts
@@ -466,11 +466,11 @@ API.v1.addRoute(
 				return API.v1.forbidden();
 			}
 
-			const moderators = (
-				await Subscriptions.findByRoomIdAndRoles(findResult._id, ['moderator'], {
-					projection: { u: 1 },
-				}).toArray()
-			).map((sub: ISubscription) => sub.u);
+			const moderators = await Subscriptions.findByRoomIdAndRoles(findResult._id, ['moderator'], {
+				projection: { u: 1, _id: 0 },
+			})
+				.map((sub) => sub.u)
+				.toArray();
 
 			return API.v1.success({
 				moderators,
--- a/apps/meteor/app/api/server/v1/groups.ts
+++ b/apps/meteor/app/api/server/v1/groups.ts
@@ -1215,11 +1215,11 @@ API.v1.addRoute(
 				userId: this.userId,
 			});
 
-			const moderators = (
-				await Subscriptions.findByRoomIdAndRoles(findResult.rid, ['moderator'], {
-					projection: { u: 1 },
-				}).toArray()
-			).map((sub: any) => sub.u);
+			const moderators = await Subscriptions.findByRoomIdAndRoles(findResult.rid, ['moderator'], {
+				projection: { u: 1, _id: 0 },
+			})
+				.map((sub) => sub.u)
+				.toArray();
 
 			return API.v1.success({
 				moderators,
--- a/apps/meteor/app/api/server/v1/rooms.ts
+++ b/apps/meteor/app/api/server/v1/rooms.ts
@@ -4,6 +4,7 @@ import { isPrivateRoom, isPublicRoom } from '@rocket.chat/core-typings';
 import { Messages, Rooms, Users, Uploads, Subscriptions } from '@rocket.chat/models';
 import type { Notifications } from '@rocket.chat/rest-typings';
 import {
+	ajv,
 	isGETRoomsNameExists,
 	isRoomsImagesProps,
 	isRoomsMuteUnmuteUserProps,
@@ -23,6 +24,7 @@ import * as dataExport from '../../../../server/lib/dataExport';
 import { eraseRoom } from '../../../../server/lib/eraseRoom';
 import { findUsersOfRoomOrderedByRole } from '../../../../server/lib/findUsersOfRoomOrderedByRole';
 import { openRoom } from '../../../../server/lib/openRoom';
+import type { RoomRoles } from '../../../../server/lib/roles/getRoomRoles';
 import { hideRoomMethod } from '../../../../server/methods/hideRoom';
 import { muteUserInRoom } from '../../../../server/methods/muteUserInRoom';
 import { toggleFavoriteMethod } from '../../../../server/methods/toggleFavorite';
@@ -37,12 +39,14 @@ import { sendFileMessage } from '../../../file-upload/server/methods/sendFileMes
 import { syncRolePrioritiesForRoomIfRequired } from '../../../lib/server/functions/syncRolePrioritiesForRoomIfRequired';
 import { executeArchiveRoom } from '../../../lib/server/methods/archiveRoom';
 import { cleanRoomHistoryMethod } from '../../../lib/server/methods/cleanRoomHistory';
+import { executeGetRoomRoles } from '../../../lib/server/methods/getRoomRoles';
 import { leaveRoomMethod } from '../../../lib/server/methods/leaveRoom';
 import { executeUnarchiveRoom } from '../../../lib/server/methods/unarchiveRoom';
 import { applyAirGappedRestrictionsValidation } from '../../../license/server/airGappedRestrictionsWrapper';
 import type { NotificationFieldType } from '../../../push-notifications/server/methods/saveNotificationSettings';
 import { saveNotificationSettingsMethod } from '../../../push-notifications/server/methods/saveNotificationSettings';
 import { settings } from '../../../settings/server';
+import type { ExtractRoutesFromAPI } from '../ApiClass';
 import { API } from '../api';
 import { composeRoomWithLastMessage } from '../helpers/composeRoomWithLastMessage';
 import { getPaginationItems } from '../helpers/getPaginationItems';
@@ -1005,3 +1009,62 @@ API.v1.addRoute(
 		},
 	},
 );
+
+const isRoomGetRolesPropsSchema = {
+	type: 'object',
+	properties: {
+		rid: { type: 'string' },
+	},
+	additionalProperties: false,
+	required: ['rid'],
+};
+export const roomEndpoints = API.v1.get(
+	'rooms.roles',
+	{
+		authRequired: true,
+		query: ajv.compile<{
+			rid: string;
+		}>(isRoomGetRolesPropsSchema),
+		response: {
+			200: ajv.compile<{
+				roles: RoomRoles[];
+			}>({
+				type: 'object',
+				properties: {
+					roles: {
+						type: 'array',
+						items: {
+							type: 'object',
+							properties: {
+								rid: { type: 'string' },
+								u: {
+									type: 'object',
+									properties: { _id: { type: 'string' }, username: { type: 'string' } },
+									required: ['_id', 'username'],
+								},
+								roles: { type: 'array', items: { type: 'string' } },
+							},
+							required: ['rid', 'u', 'roles'],
+						},
+					},
+				},
+				required: ['roles'],
+			}),
+		},
+	},
+	async function () {
+		const { rid } = this.queryParams;
+		const roles = await executeGetRoomRoles(rid, this.userId);
+
+		return API.v1.success({
+			roles,
+		});
+	},
+);
+
+type RoomEndpoints = ExtractRoutesFromAPI<typeof roomEndpoints>;
+
+declare module '@rocket.chat/rest-typings' {
+	// eslint-disable-next-line @typescript-eslint/naming-convention, @typescript-eslint/no-empty-interface
+	interface Endpoints extends RoomEndpoints {}
+}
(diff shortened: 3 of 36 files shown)


packages/web-ui-registration/src/ResetPassword/ResetPasswordPage.tsx
```diff
@@ -15,10 +24,20 @@ const getChangePasswordReason = ({
...
+declare module '@rocket.chat/rest-typings' {
+	// eslint-disable-next-line @typescript-eslint/naming-convention, @typescript-eslint/no-empty-interface
+	interface Endpoints {
+		// check file: apps/meteor/app/api/server/v1/users.ts
+		'/v1/users.setPassword': {
+			POST: ({ password }: { password: string }) => Promise<void>;
+		};
+	}
+}
+
```
tassoevan: ?

ggazzo: as we are declaring the endpoint in the meteor project, this package does not have access to the augmented interface, so I had to do this

tassoevan (dismissed):

apps/meteor/client/components/message/toolbar/items/actions/ReactionMessageAction.tsx
```diff
@@ -20,7 +20,7 @@ type ReactionMessageActionProps = {
 const ReactionMessageAction = ({ message, room, subscription }: ReactionMessageActionProps) => {
 	const chat = useChat();
 	const user = useUser();
-	const setReaction = useMethod('setReaction');
+	const setReaction = useEndpoint('POST', '/v1/chat.react');
```
pierre-lehnen-rc, line 23: the `setReaction` method had a client implementation too - is that no longer relevant?

apps/meteor/client/views/root/MainLayout/RegisterUsername.tsx
```diff
@@ -38,8 +37,8 @@ const RegisterUsername = () => {
 		throw new Error('Invalid user');
 	}
 
-	const setUsername = useMethod('setUsername');
-	const saveCustomFields = useMethod('saveCustomFields');
+	const setBasicInfo = useEndpoint('POST', '/v1/users.updateOwnBasicInfo');
```
pierre-lehnen-rc, line 40: This change will break the first login of users created with no username when "Allow username change" is false.

apps/meteor/client/components/message/toolbar/items/actions/ReactionMessageAction.tsx
ggazzo: I dont think this is real deal

apps/meteor/client/views/root/MainLayout/RegisterUsername.tsx
ggazzo: to be honest I did not find any mention to `Accounts_AllowUsernameChange` inside `updateOwnBasicInfo`

I added to the rules btw

apps/meteor/app/api/server/v1/rooms.ts
```diff
@@ -1005,3 +1009,62 @@ API.v1.addRoute(
...
+const isRoomGetRolesPropsSchema = {
+	type: 'object',
+	properties: {
+		rid: { type: 'string' },
+	},
+	additionalProperties: false,
+	required: ['rid'],
+};
+export const roomEndpoints = API.v1.get(
+	'rooms.getRoles',
```
sampaiodiego: we have `channels.roles` and `groups.roles` endpoints already, could we use them instead of creating a new one?

or should we deprecate them both in favor of this one?

if we decide to keep this one, maybe it could be called `rooms.roles`

apps/meteor/server/lib/roles/getRoomRoles.ts
```diff
@@ -22,13 +32,14 @@ export async function getRoomRoles(rid: IRoom['_id']): Promise<ISubscription[]>
 	const subscriptions = await Subscriptions.findByRoomIdAndRoles(rid, _.pluck(roles, '_id'), options).toArray();
 
 	if (!useRealName) {
-		return subscriptions;
+		return subscriptions as unknown as RoomRoles[];
```
sampaiodiego: 🙈

apps/meteor/server/services/authorization/service.ts
```diff
@@ -106,40 +106,49 @@ export class Authorization extends ServiceClass implements IAuthorization {
...
-				roles: user.roles.filter((roleId: string) => roleIds.includes(roleId)),
-			}));
+			});
+
+			return users
+				.map((user) => ({
+					...user,
+					roles: user.roles.filter((roleId: string) => roleIds.includes(roleId)),
+				}))
+				.toArray() as unknown as {
```
sampaiodiego: we should provide a generic type to `findUsersInRoles` to properly define the return type of the query

apps/meteor/app/api/server/v1/users.ts
```diff
@@ -1394,6 +1402,46 @@ API.v1.addRoute(
 	},
 );
 
+const userRoutes = API.v1.get(
+	'users.getPublicRoles',
```
sampaiodiego: I was thinking if this is the best "namespace" for this information, I was thinking if it could be under "permissions.*" namespace or somewhere else..

if we keep it here under "users.*" namespace maybe we could call it something like "users.listByRole" like we have "users.listByStatus" and "users.listTeams" .. wdyt?

sampaiodiego (changes requested):

apps/meteor/app/api/server/v1/rooms.ts
ggazzo: It is kind of known that those endpoints were a copy of other tools... but when it comes to using it becomes a little  boring when needs to know that type of the room before making the call.

apps/meteor/app/api/server/v1/rooms.ts
```diff
@@ -1005,3 +1009,62 @@ API.v1.addRoute(
 		},
 	},
 );
+
+const isRoomGetRolesPropsSchema = {
+	type: 'object',
+	properties: {
+		rid: { type: 'string' },
+	},
+	additionalProperties: false,
```
pierre-lehnen-rc, line 1018:
```suggestion
	additionalProperties: true,
```

Avoid blocking additional properties on endpoints unless you have a specific reason to (like passing the whole params object along to something else).
with `additionalProperties: false`, if we need to add an additional param in the future, it would need to be flagged as a breaking change, while with `additionalProperties: true` we can still add new params without necessarily breaking compatibility.

apps/meteor/server/lib/roles/getRoomRoles.ts
pierre-lehnen-rc: you could change the RoomRoles typing to:

```ts
export type RoomRoles = Pick<Required<ISubscription>, 'rid' | 'u' | 'roles'>;
```

apps/meteor/server/methods/saveUserProfile.ts
```diff
@@ -215,7 +230,10 @@ export function executeSaveUserProfile(
 	check(settings, Object);
 	check(customFields, Match.Maybe(Object));
 
-	if (settings.email || settings.newPassword) {
+	if (
+		(settings.email && user.emails?.length) ||
+		(settings.newPassword && Object.keys(user.services || {}).length && !user.requirePasswordChange)
```
pierre-lehnen-rc, line 235: Not sure I get the requirePasswordChange check here; Is it to avoid asking for 2fa if the password is being changed in some particular situation? In that case shouldn't it also avoid the 2fa when the email is being changed?

apps/meteor/app/api/server/v1/rooms.ts
ggazzo: additional parameters will never be considered breaking change, removing or saying that extra is no longer supported may be the case.

the default is to not accept extra fields, and it wouldn't be the first case where people pass extra fields believing that something different happens works while it doesn't

apps/meteor/app/api/server/v1/channels.ts
```diff
@@ -466,11 +466,9 @@ API.v1.addRoute(
 				return API.v1.forbidden();
 			}
 
-			const moderators = (
-				await Subscriptions.findByRoomIdAndRoles(findResult._id, ['moderator'], {
-					projection: { u: 1 },
-				}).toArray()
-			).map((sub: ISubscription) => sub.u);
+			const moderators = await Subscriptions.findByRoomIdAndRoles(findResult._id, ['moderator'], {
+				projection: { u: 1 },
```
sampaiodiego:
```suggestion
				projection: { u: 1, _id: 0 },
```

apps/meteor/app/api/server/v1/groups.ts
```diff
@@ -1215,11 +1215,9 @@ API.v1.addRoute(
 				userId: this.userId,
 			});
 
-			const moderators = (
-				await Subscriptions.findByRoomIdAndRoles(findResult.rid, ['moderator'], {
-					projection: { u: 1 },
-				}).toArray()
-			).map((sub: any) => sub.u);
+			const moderators = await Subscriptions.findByRoomIdAndRoles(findResult.rid, ['moderator'], {
+				projection: { u: 1 },
```
sampaiodiego:
```suggestion
				projection: { u: 1, _id: 0 },
```

apps/meteor/server/methods/saveUserProfile.ts
ggazzo: It seems that by the method the validations are more permissive if I don't have any saved password, I basically brought the same logic here, I'm not sure about the email

sampaiodiego (dismissed):

apps/meteor/app/api/server/v1/rooms.ts
pierre-lehnen-rc: I'm just following what was decided on the backend chapter meeting. If we don't accept extra fields and then add a new one on a minor release, clients will need to pass a different set of params for different server versions, even though they are in the same major.

ggazzo: if a new field is added as **required**, yes, that is a breaking change, if you add an optional one, thats fine.
but how  accepting infinite fields with no use because the client may already be passing it is a bit pointless, how will the client already be passing a field with coherent data of something that does not exist and not need to modify it? How would the client already be passing something in advanced without knowing what should be passed?

ggazzo: I think what happened in the past was our client (Rocket.Chat frontend) was passing extra information, pure laziness ours

'We have an object here, send it all something will be used the rest must be discarded-or no', but for me this is a lazy and not recommended reason, this is legacy of the lack of type and `any`usage, but it should not happen anymore

tassoevan: If a client application uses an extra parameter introduced in a minor version, that means _the client_ is not backward-compatible with versions prior that minor. If it aims to be compatible across a whole set under a major `X.*` version, it has to use the interface/contract of the initial minor (`X.0`), without extra parameters. It's normal to _assume_ an extra parameter will be consumed and impact the response; breaking this expectation by swallowing the extra parameter is bad.

apps/meteor/server/services/authorization/service.ts
```diff
@@ -106,40 +106,46 @@ export class Authorization extends ServiceClass implements IAuthorization {
...
 
 	private getUserFromRoles = mem(
 		async (roleIds: string[]) => {
-			const users = await Users.findUsersInRoles(roleIds, null, {
+			const users = Users.findUsersInRoles<Pick<Required<IUser>, '_id' | 'username' | 'roles'>>(roleIds, null, {
 				sort: {
 					username: 1,
 				},
 				projection: {
+					_id: 0,
```
sampaiodiego:
```suggestion
```

apps/meteor/app/api/server/v1/rooms.ts
sampaiodiego: I'm ok by having `additionalProperties: false` for this particular endpoint since it is a new one, but we need to be careful if we ever change it.. I agree being more permissive (by `additionalProperties: true`) is something that makes the whole development cycle easier since it doesn't bring any concerns regarding breaking changes and legacy clients, I also see value on being more restrictive as this ensures a tighter contract and removes the risk of having code using a non-documented field.

sampaiodiego: and I think we should all agree with whatever definition we come up with and document it so people understand the reasoning behind the decision and also provides a guide for us all when defining the schemas for our endpoints

sampaiodiego (approved):

PR #36147, opened 2025-06-04
chore: WCAG(1.1.1) Added is='span' on fieldlabels to avoid orphaned form labels

ergot-rp:


dougfabris:
Hello @ergot-rp thanks for the contribution but can you tell me what is the goal of these changes? Basically we associate these labels with the inputs using `for` and `id`, so on my understanding they are not "orphaned"

apps/meteor/client/views/account/preferences/PreferencesHighlightsSection.tsx
```diff
@@ -13,9 +13,11 @@ const PreferencesHighlightsSection = () => {
 		<AccordionItem title={t('Highlights')}>
 			<FieldGroup>
 				<Field>
-					<FieldLabel htmlFor={highlightsId}>{t('Highlights_List')}</FieldLabel>
+					<FieldLabel is='span' id={highlightsId}>
+						{t('Highlights_List')}
+					</FieldLabel>
 					<FieldRow>
-						<TextAreaInput id={highlightsId} {...register('highlights')} rows={4} />
+						<TextAreaInput aria-labelledby={highlightsId} {...register('highlights')} rows={4} />
```
gabriellsh: Was this one supposed to be changed? Using the `Wave` chrome plugin, this section seemed to have no violations regarding this label.

apps/meteor/client/views/account/preferences/PreferencesMessagesSection.tsx
```diff
@@ -96,7 +98,7 @@ const PreferencesMessagesSection = () => {
 					</FieldHint>
 				</Field>
 				<Field>
-					<FieldLabel>{t('Message_TimeFormat')}</FieldLabel>
+					<FieldLabel is='span'>{t('Message_TimeFormat')}</FieldLabel>
 					<FieldLink href='/account/accessibility-and-appearance'>{t('Go_to_accessibility_and_appearance')}</FieldLink>
```
gabriellsh:
```suggestion
					<FieldLabel is='span' id={messageTimeFormatLabelId}>{t('Message_TimeFormat')}</FieldLabel>
					<FieldLink id={messageTimeFormatLinkId} aria-labelledby={`${messageTimeFormatLabelId} ${messageTimeFormatLinkId}`} href='/account/accessibility-and-appearance'>{t('Go_to_accessibility_and_appearance')}</FieldLink>
```

Just a small suggestion from something I read here: https://www.w3.org/WAI/WCAG21/Techniques/aria/ARIA7

Do you think this makes sense? If so, do you think we can apply this for all affected FieldLinks in this PR?

gabriellsh: PS: Not sure how these links are currently described by screen readers, and if this will make it better or worse. If you think navigation is better without adding this, let's keep like it is.

apps/meteor/client/views/account/preferences/PreferencesNotificationsSection.tsx
```diff
@@ -95,22 +98,18 @@ const PreferencesNotificationsSection = () => {
...
 						)}
 						{notificationsPermission !== 'denied' && notificationsPermission !== 'granted' && (
-							<>
-								<Button primary onClick={onAskNotificationPermission}>
-									{t('Enable_Desktop_Notifications')}
-								</Button>
-							</>
+							<Button primary onClick={onAskNotificationPermission}>
+								{t('Enable_Desktop_Notifications')}
+							</Button>
```
gabriellsh: Would it make sense to add an ID to this label, and aria-labelledby to the buttons? My concern is that the label as a span without any reference might be confusing to screen reader users. But adding it might also make the button label text be duplicated, being read like "Desktop notifications: enable desktop notifications". If this is not an issue please disregard.

apps/meteor/client/views/account/preferences/PreferencesNotificationsSection.tsx
```diff
@@ -218,7 +223,9 @@ const PreferencesNotificationsSection = () => {
 				{showMobileRinging && (
 					<Field>
 						<FieldRow>
-							<FieldLabel htmlFor={enableMobileRingingId}>{t('VideoConf_Mobile_Ringing')}</FieldLabel>
+							<FieldLabel htmlFor={enableMobileRingingId}>
+								{t('VideoConf_Mobile_Ringing')}
+							</FieldLabel>
```
gabriellsh: The linter is complaining about this section, and nothing has changed in it other than the line breaks. Can you revert it?

apps/meteor/client/views/account/preferences/PreferencesHighlightsSection.tsx
ergot-rp: No, my mistake. I reverted it

PR #36152, opened 2025-06-04
chore: Remove display modes and options from new navigation

yash-rajpal:

We are removing display modes and options from new navigation feature preview.

This PR only removes the UI elements from new navigation, doesn't remove the underlying settings and preferences as they will still be used by the old navigation. 
## Further comments

[SIDE2-43](https://rocketchat.atlassian.net/browse/SIDE2-43)

[SIDE2-43]: https://rocketchat.atlassian.net/browse/SIDE2-43?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


juliajforesti (dismissed):
I would say this is a feat not a chore and add a changeset, emphasizing it affects only feature preview.

other than that I have a few concerns:
- why are we removing this?
- what happens if the user had a configuration and wants to change it? is the user going to have their current configuration forever?
or is this going to go somewhere else?

apps/meteor/client/sidebarv2/Sidebar.tsx
```diff
@@ -7,15 +7,12 @@ import SidebarFooter from './footer';
 import BannerSection from './sections/BannerSection';
 
 const Sidebar = () => {
-	const sidebarViewMode = useUserPreference('sidebarViewMode');
 	const sidebarHideAvatar = !useUserPreference('sidebarDisplayAvatar');
```
dougfabris: why are we keeping this?

dougfabris (changes requested):

apps/meteor/client/sidebarv2/Sidebar.tsx
yash-rajpal: Removed.

dougfabris (approved):

PR #36153, opened 2025-06-04
feat: support mapping users to exchange servers by domain

cardoso:
## Issue(s)

https://rocketchat.atlassian.net/browse/CONN-674

### DIFF
--- a/.changeset/new-doors-smash.md
+++ b/.changeset/new-doors-smash.md
@@ -1,0 +1,6 @@
+---
+'@rocket.chat/i18n': minor
+'@rocket.chat/meteor': minor
+---
+
+Prevents creation of unnamed Personal Access Tokens by requiring the form's `name` field fullfilment.
--- a/apps/meteor/client/providers/AuthenticationProvider/AuthenticationProvider.tsx
+++ b/apps/meteor/client/providers/AuthenticationProvider/AuthenticationProvider.tsx
@@ -114,7 +114,6 @@ const AuthenticationProvider = ({ children }: AuthenticationProviderProps): Reac
 					Accounts._unstoreLoginToken = _unstoreLoginToken;
 				};
 			},
-
 			queryLoginServices: {
 				getCurrentValue: () => loginServices.getLoginServiceButtons(),
 				subscribe: (onStoreChange: () => void) => loginServices.on('changed', onStoreChange),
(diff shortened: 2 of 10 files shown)


apps/meteor/app/api/server/helpers/getUserInfo.ts
```diff
@@ -25,13 +25,49 @@ const getUserPreferences = async (me: IUser): Promise<Record<string, unknown>> =
...
+ * @param email - The user's email object, which may contain the address and verification status.
+ * @returns The calendar settings for the user, including Outlook calendar settings if enabled.
+ */
+const getUserCalendar = (email: false | IUserEmail | undefined): IUserCalendar => {
+	const calendarSettings: IUserCalendar = {};
+
+	const outlook = {
+		Enabled: settings.get<boolean>('Outlook_Calendar_Enabled'),
+		Exchange_Url: settings.get<string>('Outlook_Calendar_Exchange_Url'),
+		Outlook_Url: settings.get<string>('Outlook_Calendar_Outlook_Url'),
```
aleksandernsilva, line 41: Could this be `enabled`, `exchangeUrl` and `outlookUrl`?

apps/meteor/client/hooks/roomActions/useOutlookCalenderRoomAction.ts
```diff
@@ -1,15 +1,15 @@
...
 
 const OutlookEventsRoute = lazy(() => import('../../views/outlookCalendar/OutlookEventsRoute'));
 
 export const useOutlookCalenderRoomAction = () => {
-	const enabled = useSetting('Outlook_Calendar_Enabled', false);
+	const user = useUser();
 
 	return useMemo((): RoomToolboxActionConfig | undefined => {
-		if (!enabled) {
+		if (!user?.settings?.calendar?.outlook?.Enabled) {
```
aleksandernsilva, line 12: This is a pretty long chain, it might worth moving this to it's own hook similar to `useUserId` or at least storing it in a variable to avoid repeating the chain. Same applies to the other files doing the same in this PR.

aleksandernsilva (dismissed):

apps/meteor/app/api/server/helpers/getUserInfo.ts
cardoso: I considered that, but it's better to keep consistency with the setting names.

apps/meteor/client/hooks/roomActions/useOutlookCalenderRoomAction.ts
cardoso: I think it's better not to hide what's happening here in a hook or variable, unless there's a concern other than style. 

The idea here is to keep the change minimal on the frontend, basically replacing usages of the `settings.get` with the new mapping on the backend.

PR #36156, opened 2025-06-04
fix: desktop notifications not respecting privacy settings

abhinavkrin:

This PR resolves an issue where the `Show Message in Notification` and `Show Channel/Group/Username in Notification` settings were not respected in desktop notifications. 

Key updates:
- Notification content now dynamically adheres to these privacy settings.
- Even when titles and message content are omitted, users are correctly redirected to the intended room using the room ID.
- Added support across all room types (channels, private groups, DMs, VoIP, and Livechat).
- Added comprehensive tests to validate different combinations of privacy settings.

https://github.com/user-attachments/assets/6be36dc9-dae2-4f38-8c05-7c073af0e7e1

## Steps to test or reproduce

1. Go to **Administration > Workspace > Settings > Push > Privacy**.
2. Disable one or both of:
   - *Show Message in Notification*
   - *Show Channel/Group/Username in Notification*
3. Trigger a message to a user.
4. Confirm the notification respects the settings (no content shown if disabled).
5. Click the notification and confirm the app navigates to the correct room.

## Further comments

[SUP-781](https://rocketchat.atlassian.net/browse/SUP-781)

[SUP-781]: https://rocketchat.atlassian.net/browse/SUP-781?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/meteor/definition/IRoomTypeConfig.ts
```diff
@@ -105,6 +105,7 @@ export interface IRoomTypeServerDirectives {
 		sender: AtLeast<IUser, '_id' | 'name' | 'username'>,
 		notificationMessage: string,
 		userId: string,
+		langauge?: string,
```
MarcosSpessatto:
```suggestion
		language?: string,
```

apps/meteor/server/lib/rooms/roomCoordinator.ts
```diff
@@ -41,14 +42,29 @@ class RoomCoordinatorServer extends RoomCoordinator {
...
+				let text;
+				let title;
+				let name;
+				if (showPushMessage) {
+					const useRealName = settings.get<boolean>('UI_Use_Real_Name');
+					const senderName = getUserDisplayName(sender.name, sender.username, useRealName);
+					text = `${senderName}: ${notificationMessage}`;
+				} else {
+					text = i18n.t('You_have_a_new_message', { lng });
+				}
```
MarcosSpessatto: This piece of code seems to be repeated over the room types, do you think we could extract it to a function and have it unit tested?

MarcosSpessatto (changes requested):

apps/meteor/server/lib/rooms/roomCoordinator.ts
abhinavkrin: Hey, I have move the logic to a separate function buildNotificationDetails. It makes sure about rights settings being applied. I have also added unit tests for the function.

apps/meteor/client/hooks/notification/useNotification.ts
```diff
@@ -63,56 +63,54 @@ export const useNotification = () => {
 			n.close();
 			window.focus();
 
-			if (!notification.payload._id || !notification.payload.rid || !notification.payload.name) {
-				return;
-			}
+			const jump = msgId && { jump: msgId };
```
gabriellsh, line 67: My only concern is the removal of these lines. I'm aware the type probably specifies these values are always present, but since it was there before, I wonder if there's a case where they might not be present, or be empty.

I'm open to having my mind changed though, would you like to share your opinion regarding my comment?

abhinavkrin: Only a room’s rid is ever required to navigate—you’ll always have it. Nevertheless, we still defensively check for it. The room’s display name may be omitted (for example, when “hide room name in notifications” is enabled), but routing falls back to the rid regardless. And if a specific message ID is supplied, we deep-link straight to that message; if not, we simply open the room identified by the rid.

gabriellsh: Cool, if that's the case this is ok. Thanks man!

apps/meteor/definition/IRoomTypeConfig.ts
gabriellsh: does the lan-gauge measure the lan? :trollface:

PR #36163, opened 2025-06-05
fix: incoming webhook message not respecting Message_MaxAllowedSize setting

abhinavkrin:

This PR fixes an issue where incoming webhook messages could exceed the allowed characters per message setting.
## Further comments

Closes [SUP-796](https://rocketchat.atlassian.net/browse/SUP-796)

[SUP-796]: https://rocketchat.atlassian.net/browse/SUP-796?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


abhinavkrin:
A potentially more appropriate place for this validation could be within the `sendMessage` function (`apps/meteor/app/lib/server/functions/sendMessage.ts`). However, applying the check there might introduce regressions in scenarios where bypassing this limit is intentional or required by existing implementations. Need a second opinion on this.

Note: This refers to the `sendMessage` function, not the `sendMessage` method (`apps/meteor/app/lib/server/methods/sendMessage.ts`), which already enforces this setting.

PR #36164, opened 2025-06-05
chore: apps-engine tabs to spaces

d-gubert:

Change style definitions to better align with the rest of the project


packages/apps-engine/.prettierrc
sampaiodiego, line 1: is this file required at all? should it just be removed in favor of the main one?
https://github.com/RocketChat/Rocket.Chat/blob/846627bb8787d6e97ff69d451c65e9903e02e8df/.prettierrc#L2-L11

d-gubert: did

sampaiodiego (approved):

Issue #36087, opened 2025-05-27
[API Bug] `chat.getPinnedMessages` does not support `sort` parameter despite documentation


Jacksonzhang0316:
## Description

The official documentation for the `GET /api/v1/chat.getPinnedMessages` endpoint states that a `sort` parameter can be passed to sort pinned messages. However, after reviewing the actual source code of the endpoint, it appears that the `sort` parameter is neither parsed nor passed into the internal message query function, rendering it ineffective.

## Steps to Reproduce

1. Call the `chat.getPinnedMessages` endpoint with a `sort` query parameter, e.g.:
GET /api/v1/chat.getPinnedMessages?roomId=GENERAL&sort={"ts":-1}

(Properly URL-encoded: `sort=%7B%22ts%22%3A-1%7D`)

2. The response always returns messages in the default order (likely unsorted or backend-determined), and the sort order has no effect.

## Source Code Analysis

The current implementation of `chat.getPinnedMessages` does not parse or forward the `sort` parameter:

```ts
const { cursor, totalCount } = Messages.findPaginatedPinnedByRoom(roomId, {
skip: offset,
limit: count,
// missing: sort
});
```

Even though the internal Messages.findPaginatedPinnedByRoom method does accept a FindOptions<IMessage> object that includes a sort field:

```
findPaginatedPinnedByRoom(roomId: IMessage['rid'], options?: FindOptions<IMessage>): FindPaginated<FindCursor<IMessage>> {
  const query: Filter<IMessage> = {
    t: { $ne: 'rm' },
    _hidden: { $ne: true },
    pinned: true,
    rid: roomId,
  };

  return this.findPaginated(query, options);
}
```

## Expected Behavior
The API should properly accept and parse the `sort` parameter (e.g., `{"ts": -1}`) from the query string and pass it to `Messages.findPaginatedPinnedByRoom()`.

PR #36172, opened 2025-06-06
fix: adds support for the `sort` parameter in the `getPinnedMessages` API

zscontributor:
## Summary
This pull request adds support for the `sort` parameter in the `getPinnedMessages` API, which was previously missing.

## Details

After reviewing the current Rocket.Chat source code, the following points were confirmed:
1. The `getPinnedMessages` API uses `findPaginatedPinnedByRoom`, which takes `FindOptions<IMessage>` as the `options` parameter.
2. `FindOptions` includes a `sort` property that can be used to control the result ordering.
3. `findPaginated` (used internally by `findPaginatedPinnedByRoom`) also supports sorting via the `options` parameter.

##  Conclusion
Therefore, adding support for the `sort` parameter in the `getPinnedMessages` API is appropriate and consistent with the current implementation.

## Issue(s)
https://github.com/RocketChat/Rocket.Chat/issues/36087


apps/meteor/tests/end-to-end/api/chat.ts
```diff
@@ -2869,6 +2869,52 @@ describe('[Chat]', () => {
 					})
 					.end(done);
 			});
+
+			it('should return a list of pinned messages sorted by timestamp in descending order by default', (done) => {
```
KevLehman, line 2927: I know other tests in this file are with callbacks but could you change these new ones to use async/await pls?

apps/meteor/tests/end-to-end/api/chat.ts
```diff
@@ -2869,6 +2869,52 @@ describe('[Chat]', () => {
...
+					})
+					.expect('Content-Type', 'application/json')
+					.expect(200)
+					.expect((res) => {
+						expect(res.body).to.have.property('success', true);
+						expect(res.body).to.have.property('messages').and.to.be.an('array');
+						expect(res.body.messages.length).to.be.equal(1);
+						// Verify messages are sorted by ts in ascending order
+						const { messages } = res.body;
+						for (let i = 0; i < messages.length - 1; i++) {
```
KevLehman, line 2965: maybe a for..of or a map? (or even a foreach?)

apps/meteor/app/api/server/v1/chat.ts
```diff
@@ -506,6 +507,7 @@ API.v1.addRoute(
 			const { cursor, totalCount } = Messages.findPaginatedPinnedByRoom(roomId, {
 				skip: offset,
 				limit: count,
+				sort: sort || { ts: -1 },
```
KevLehman, line 548: This endpoint uses an AJV schema for validating the params
Would you mind checking if the schema for it accepts the `sort` param already? You can find it really close to the `isChatGetPinnedMessagesProps` function, on rest-typings package

KevLehman (changes requested):

PR #36189, opened 2025-06-10
test: Automate the functionality for quoting an attachment with description. 

Harmeet221:
JIRA TASK: 

Test 1: should show file preview and description when quoting a message with attachment in the channel
Steps:
- Send message with attachment in the channel
- Quote the message with attachment
- Verify the quoted message appears correctly
Test 2: Should quote attachment file within a thread
- Send initial message in channel
- Create thread and send message with attachment
- Quote the message with attachment in thread
- Verify the quoted message appears correctly in thread


Harmeet221:
not required

PR #36195, opened 2025-06-10
fix: Handle errors properly when loading Moment Locale

nmagedman:
1. BUGFIX: Use `await` when calling async function Assets.getTextAsync()

    Commit df82bccb86fdab509c9de324ad4adfda93f0ddc1 "chore: Use Assets’ async API (#29311)"
replaced several `Assets.getText()` calls with `await Assets.getTextAsync()`.
One of those getTextAsync calls was not awaited.

    That one was especially problematic in that it is in a try/catch block,
however you cannot try/catch a Promise, so we missed handling that error.
You must either try/catch the awaited Promise or chain on a Promise#catch handler.

2. Meteor method loadLocale: Allow getMomentLocale() exceptions to pass-through

    Do not try/catch `getMomentLocale()`:

    - It is an async function.  You cannot try/catch Promises, so this try/catch has been a NOOP for years already.  We apparently don't need it.
    - getMomentLocale() already handles exceptions and raises an (almost?) identical exception already.

## Steps to test or reproduce

- Configure `Log_Exceptions_to_Channel` with a channel name
- In user preferences, change the Language to "Norwegian Bokmål (Norway)"
- In the exceptions channel, it will say:
    ```
    Exception while invoking method loadLocale
    > Error: Unknown asset: moment-locales/nb-no.js
    ```
- Change the language to Default
- Apply the PR
- Change the language back again to "Norwegian Bokmål (Norway)"
- `loadLocale` will successfully load the `nb.js` Moment Locale file (sans `-no` country)


nmagedman:
The main bugfix (the missing `await` keywords in `apps/meteor/server/lib/getMomentLocale.ts`) was fixed incidentally in #36651.  The superfluous NOOP try/catch in `apps/meteor/server/methods/loadLocale.ts` is still there but it’s not worth pursuing it.   Closing out this PR.

PR #36202, opened 2025-06-11
fix: Use `Livechat_enabled_when_agent_idle` settting where it's accepted

KevLehman:
## Issue(s)

https://rocketchat.atlassian.net/browse/CTZ-210
## Further comments

This should help with some inconsistencies when evaluating this behavior as some places were using it and some others were not.

### DIFF
--- a/apps/meteor/app/livechat/server/lib/QueueManager.ts
+++ b/apps/meteor/app/livechat/server/lib/QueueManager.ts
@@ -422,13 +422,14 @@ export class QueueManager {
 		};
 
 		let defaultAgent: SelectedAgent | undefined;
-		if (
-			servedBy?.username &&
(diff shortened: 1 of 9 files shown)


apps/meteor/app/livechat/server/lib/QueueManager.ts
```diff
@@ -422,7 +422,10 @@ export class QueueManager {
 		};
 
 		let defaultAgent: SelectedAgent | undefined;
-		if (servedBy?.username && (await Users.findOneOnlineAgentByUserList(servedBy.username))) {
+		if (
+			servedBy?.username &&
+			(await Users.findOneOnlineAgentByUserList(servedBy.username, {}, settings.get<boolean>('Livechat_enabled_when_agent_idle')))
```
lucas-a-pelegrino: I'm thinking if we could label this in a var just to improve readability. 🤔

lucas-a-pelegrino (approved):

lucas-a-pelegrino (dismissed):

PR #36214, opened 2025-06-12
chore: move hono implementation to external package

MarcosSpessatto:
This will enable other packages to declare their HTTP API using the Router abstraction and export it.

This is especially important for the federation project.


apps/meteor/app/api/server/ApiClass.ts
```diff
@@ -171,7 +171,7 @@ export class APIClass<
 		inviteToken: number;
 	};
 
-	readonly router: Router<any>;
+	readonly router: RocketChatAPIRouter<any>;
```
ggazzo: I think we can keep `Router` interface usega around... no need to replace all the places. unless in some specific case you really need some method that is only implemented  at `RocketChatAPIRouter`

apps/meteor/app/api/server/router.ts
```diff
@@ -19,214 +14,38 @@ function splitArray<T, U>(arr: [...T[], U]): [T[], U] {
...
-		| TOperations
-		| ({
-				method: Method;
-				path: TPathPattern;
-		  } & Omit<TOptions, 'response'>)
-	> {
-		const [middlewares, action] = splitArray(actions);
-
-		this.innerRouter[method.toLowerCase() as Lowercase<Method>](`/${subpath}`.replace('//', '/'), ...middlewares, async (c) => {
+	protected _convertToHonoAction(action: TypedAction<any, any>): (c: HonoContext) => Promise<any> {
```
ggazzo: move the whole class to the external package, only implements this method (maybe find a more generic name), also, make the context generic on the interface context

ggazzo (dismissed):

ggazzo (approved):

PR #36215, opened 2025-06-12
chore: upgrade jest to v30

cardoso:
Changelog: https://github.com/jestjs/jest/blob/main/CHANGELOG.md#3000
Upgrade guide: https://jestjs.io/docs/upgrading-to-jest30

### DIFF
--- a/apps/meteor/client/views/account/profile/ActionConfirmModal.tsx
+++ b/apps/meteor/client/views/account/profile/ActionConfirmModal.tsx
@@ -1,6 +1,6 @@
 import { Box, PasswordInput, TextInput, FieldGroup, Field, FieldRow, FieldError } from '@rocket.chat/fuselage';
 import type { ChangeEvent } from 'react';
-import { useState, useCallback } from 'react';
+import { useState, useCallback, useId } from 'react';
 import { useTranslation } from 'react-i18next';
 
 import GenericModal from '../../../components/GenericModal';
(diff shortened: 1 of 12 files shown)


tassoevan (changes requested):
There is a subtle issue. Looks like `jest` now delivers its own types, not depending on `@types/jest` anymore. *However*, `@types/jest-axe` still depends of `@types/jest@^29`. Using the wrong types means that removed matchers like `expect(...).toThrowError(...)` are mistakenly identified as valid in compile-time, while on runtime:

```
TypeError: expect(...).toThrowError is not a function
```

cardoso:
@tassoevan those are at least marked deprecated 

Now I see `@rocket.chat/models` , `@rocket.chat/mock-providers`, and `@rocket.chat/cas-validate` were using "^29.7.0" instead of "~29.7.0" which is why they weren't upgraded via search and replace, and weren't failing CI AFAIK. Not sure how I didn't catch that. 🤡 

There's also `@rocket.chat/presence` and `@rocket.chat/core-services` using bebel-jest "^29.7.0". I'll see what to do about that.

-- edit --

The issue with `@types/jest` might be a bit more complicated, because it's used for the global typings (which I'm not a fan of, but our codebase uses it pervasively). If it's easy to swap with `@jest/globals`, I'll try that, otherwise it should be obvious to spot the deprecated methods.

PR #36228, opened 2025-06-16
fix: GDPR for contacts

KevLehman:
## Issue(s)

https://rocketchat.atlassian.net/browse/CTZ-172

### DIFF
--- a/apps/meteor/app/livechat/server/lib/guests.ts
+++ b/apps/meteor/app/livechat/server/lib/guests.ts
@@ -80,7 +80,7 @@ export async function saveGuest(
 
 export async function removeGuest({ _id, token }: { _id: string; token: string }) {
 	await cleanGuestHistory(token);
-	return LivechatVisitors.disableById(_id);
+	return Promise.all([LivechatVisitors.disableById(_id), LivechatContacts.disableByVisitorId(_id)]);
 }
 
 export async function registerGuest(newData: RegisterGuestType): Promise<ILivechatVisitor | null> {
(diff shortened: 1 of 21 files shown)


lucas-a-pelegrino (dismissed):

d-gubert (dismissed):
Seems straightforward enough in the context of the Apps-Engine :+1:

apps/meteor/app/livechat/server/lib/guests.ts
```diff
@@ -78,11 +78,32 @@ export async function saveGuest(
...
+	livechatLogger.debug({ msg: 'Removing channels for contacts', visitorId: _id, contacts: contacts.map(({ _id }) => _id) });
+	for await (const contact of contacts) {
+		for await (const { visitor } of contact.channels) {
+			await removeGuest({ _id: visitor.visitorId });
+		}
+
+		await LivechatContacts.disableByVisitorId(_id);
+	}
+
+	return true;
```
MarcosSpessatto: Any reason for returning a boolean?

MarcosSpessatto (changes requested):

apps/meteor/app/livechat/server/lib/guests.ts
KevLehman: It's actually to keep this error on the endpoint:
```
		const { _id } = visitor;
		const result = await removeContactsByVisitorId({ _id });
		if (!result) {
			throw new Meteor.Error('error-removing-visitor', 'An error ocurred while deleting visitor');
		}

d-gubert (approved):

lucas-a-pelegrino (approved):

MarcosSpessatto (approved):

PR #36245, opened 2025-06-18
feat: endpoint to fetch distinct instanceId for logs filter

d-gubert:

Add the endpoint `/api/apps/logs/instanceIds` that returns an array of the distinct values of `instanceId` found in the logs collection.

This will assist filtering UI capabilities

## Further comments

[ALT-21]


d-gubert:
I'm pondering whether to add a changeset for this one.  The main purpose of the endpoint is to assist the UI with pre-populated filtering data, there is not much use beyond that - thus my doubt if it deserves to be in the changeset by itself, or if we leave the changeset to be added by the filters PR

packages/models/src/models/AppLogsModel.ts
```diff
@@ -24,22 +24,35 @@ export class AppsLogsModel extends BaseRaw<any> implements IAppLogsModel {
...
 			// Index for queries on general logs endpoint
 			{
 				key: {
 					'_updatedAt': -1,
 					'entries.severity': 1,
+					'instanceId': 1,
 					'method': 1,
 				},
-				name: 'general_logs_index',
+				name: 'general_logs_index_v2',
```
d-gubert, line 39: For these changes, I'm considering adding a migration file that deletes the old indexes as housekeeping. It would not cause errors if we don't do it, but leaves the indexes in the collection in a "dirty" state

d-gubert: There is a discussion on how to proceed with this approach, should be solved outside of this PR scope

apps/meteor/client/views/marketplace/AppDetailsPage/tabs/AppLogs/Filters/AppLogsFilter.tsx
```diff
@@ -1,15 +1,16 @@
...
 
+import { EventFilterSelect } from './EventFilterSelect';
 import { InstanceFilterSelect } from './InstanceFilterSelect';
 import { SeverityFilterSelect } from './SeverityFilterSelect';
 import { TimeFilterSelect } from './TimeFilterSelect';
 import { useCompactMode } from '../../../useCompactMode';
 import { useAppLogsFilterFormContext } from '../useAppLogsFilterForm';
 
-export const AppLogsFilter = () => {
+export const AppLogsFilter = ({ appId }: { appId: string }) => {
```
MartinSchoeler:
```suggestion
type AppLogsFilterProps = { appId: string }
export const AppLogsFilter = ({ appId }: AppLogsFilterProps) => {
```

MarcosSpessatto (approved):

MartinSchoeler (approved):

ggazzo (approved):
approving in respect to the approvals of my friends who are enjoying their well-deserved rest

PR #36256, opened 2025-06-19
fix: Allow queue to recover from some errors

KevLehman:
## Issue(s)

https://rocketchat.atlassian.net/browse/CTZ-229
## Further comments

There was exactly one function that was not covered in a try/catch, cause the only way for it to fail was for Mongo to actually fail... guess what happened


apps/meteor/server/services/omnichannel/queue.ts
```diff
@@ -79,28 +81,38 @@ export class OmnichannelQueue implements IOmnichannelQueue {
...
-				() => this.checkQueue(queue),
-			);
+			this.scheduleExecution();
+		} catch (e) {
+			queueLogger.error({
+				msg: 'Queue Worker Error. Rescheduling with extra delay',
+				extraDelay: this.errorDelay,
+				err: e,
+			});
+			this.scheduleExecution(this.errorDelay);
```
marceloschmidt, line 114: Is it expected that this will only run inside the catch? 
Before it would run every time and now only if an error is thrown?

KevLehman: There's the same call with no extra delay on L107

marceloschmidt: oh perfect, I did not see that :) thanks!

PR #36258, opened 2025-06-19
fix: AJV schema for _removing_ an agent from a department

KevLehman:
## Further comments

Not so common to remove agents this way, but it was not working. Caused by a refactor on 7.7.

### DIFF
--- a/apps/meteor/tests/end-to-end/api/livechat/10-departments.ts
+++ b/apps/meteor/tests/end-to-end/api/livechat/10-departments.ts
@@ -911,6 +911,30 @@ import { IS_EE } from '../../../e2e/config/constants';
 				.send({ upsert: [{ agentId: agent._id, username: agent.username, name: agent.name }], remove: [] })
 				.expect(200);
 			expect(res.body).to.have.property('success', true);
+		});
+		it('should successfully remove an agent from a department', async () => {
+			const [dep, agent] = await Promise.all([createDepartment(), createAgent()]);
+			const res = await request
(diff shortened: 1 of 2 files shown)


pierre-lehnen-rc (approved):

KevLehman:
Totally agree. However the endpoint accepted them before the ajv validator so removing them was a breaking change. I am writing a task to remove them on v8

ggazzo:
btw not fix the frontend already?
also, are we going to backport this? If not, we're almost saying that this "break change" isn't important.



> Totally my opinion and open to being convinced: from the bottom of my heart, removing these fields from the API would be a fix and not a breaking change, whoever needs to adjust it just needs to remove the fields, almost zero difficulties

KevLehman:
Yup, backport is planned. Is one of the items I mentioned for 7.7x patch 😬 and ideally it should be on 7.8 but I just noticed the candidate is out already. Gonna check on Monday this.


On the fe, since it was meant to be backported, I tried to not include more things.

ggazzo:
so a good mid ground solution is to allow the schema in the ajv, but remove it from the type and fix it in the frontend.

Also, open a PR removing it from version 8.0.0 already

When you do the backport, you can leave out the frontend if (you want) and type and just bring the schema (as was your first suggestion), but to consider this done all the  mention points should be solved

KevLehman:
@ggazzo you can check it now :)

I moved this one to 7.8 branch as we have to patch it to 7.7 anyways

ggazzo (approved):

PR #36265, opened 2025-06-20
fix: parseUrls parameter not working in chat.postMessage

abhinavkrin:

Restores support for the `parseUrls` boolean parameter in `chat.postMessage`.  Currently passing this param caused the endpoint to return error.
- Reintroduced `parseUrls` into the `chat.postMessage` payload validation schema.

## Further comments

[SUP-809](https://rocketchat.atlassian.net/browse/SUP-809)

### DIFF
--- a/apps/meteor/tests/end-to-end/api/chat.ts
+++ b/apps/meteor/tests/end-to-end/api/chat.ts
@@ -5,7 +5,7 @@ import { expect } from 'chai';
 import { after, before, beforeEach, describe, it } from 'mocha';
 import type { Response } from 'supertest';
 
-import { getCredentials, api, request, credentials } from '../../data/api-data';
+import { getCredentials, api, request, credentials, apiUrl } from '../../data/api-data';
 import { followMessage, sendSimpleMessage, deleteMessage } from '../../data/chat.helper';
 import { imgURL } from '../../data/interactions';
 import { updatePermission, updateSetting } from '../../data/permissions.helper';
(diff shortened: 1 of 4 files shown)


packages/rest-typings/src/v1/chat.ts
```diff
@@ -830,6 +830,10 @@ const ChatPostMessageSchema = {
 					type: 'object',
 					nullable: true,
 				},
+				parseUrls: {
+					type: 'boolean',
+					nullable: true,
```
KevLehman:
```suggestion
```

If param won't accept `null` directly, we can drop this flag

apps/meteor/tests/end-to-end/api/chat.ts
```diff
@@ -515,6 +515,60 @@ describe('[Chat]', () => {
...
+				.expect('Content-Type', 'application/json')
+				.expect(200)
+				.expect((res) => {
+					expect(res.body).to.have.property('success', true);
+					expect(res.body).to.have.nested.property('message.msg', apiUrl);
+					expect(res.body.message).to.have.property('urls');
+				});
+		});
+
+		it('should parse urls when parseUrls is not provided', async () => {
```
MarcosSpessatto, line 554: I think it should NOT parse if not provided correctly?

abhinavkrin: Wont it be a breaking change? The default behaviour seems to parse urls when the field is not provided.
```typescript
messageObj.parseUrls !== undefined ? messageObj.parseUrls : !messageObj.attachments,
```
https://github.com/RocketChat/Rocket.Chat/blob/c04e2f480280ec7c6f251da47ec18c1fa8be5934/apps/meteor/app/lib/server/functions/processWebhookMessage.ts#L111

MarcosSpessatto (approved):

PR #36267, opened 2025-06-20
chore: upgrade eslint from v8 to v9

cardoso:
## Issue(s)

https://rocketchat.atlassian.net/browse/CORE-1212


cardoso:
Will take a different approach.

PR #36277, opened 2025-06-23
chore: Use `useButtonPattern` hook instead of manual key check

dougfabris:

### DIFF
--- a/apps/meteor/client/components/message/variants/SystemMessage.tsx
+++ b/apps/meteor/client/components/message/variants/SystemMessage.tsx
@@ -11,11 +11,12 @@ import {
 	MessageUsername,
 	MessageNameContainer,
 } from '@rocket.chat/fuselage';
+import { useButtonPattern } from '@rocket.chat/fuselage-hooks';
 import { UserAvatar } from '@rocket.chat/ui-avatar';
 import { useUserDisplayName } from '@rocket.chat/ui-client';
 import type { TranslationKey } from '@rocket.chat/ui-contexts';
 import { useUserPresence } from '@rocket.chat/ui-contexts';
-import type { ComponentProps, ReactElement, KeyboardEvent } from 'react';
+import type { ComponentProps, ReactElement } from 'react';
 import { memo } from 'react';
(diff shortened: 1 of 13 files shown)


apps/meteor/client/components/message/variants/SystemMessage.tsx
```diff
@@ -78,16 +80,7 @@ const SystemMessage = ({ message, showUserAvatar, ...props }: SystemMessageProps
 			</MessageSystemLeftContainer>
 			<MessageSystemContainer>
 				<MessageSystemBlock>
-					<MessageNameContainer
-						tabIndex={0}
-						role='button'
-						onClick={(e) => user.username && openUserCard(e, user.username)}
```
gabriellsh, line 84: We lost the check `user.username &&` here. I checked the `openUserCard` function and that does not happen there. You sure we can take it out?

apps/meteor/client/components/message/variants/SystemMessage.tsx
```diff
@@ -78,16 +80,7 @@ const SystemMessage = ({ message, showUserAvatar, ...props }: SystemMessageProps
...
-						onKeyDown={(e: KeyboardEvent<HTMLSpanElement>) => {
-							(e.code === 'Enter' || e.code === 'Space') && openUserCard(e, message.u.username);
-						}}
-						style={{ cursor: 'pointer' }}
```
gabriellsh, line 88: This style was also removed. I'm wondering if we still need it...

gabriellsh (changes requested):

apps/meteor/client/components/message/variants/SystemMessage.tsx
dougfabris: the style still there 👀

apps/meteor/client/components/message/variants/SystemMessage.tsx
dougfabris: no issues, `username` won't be `undefined` here

gabriellsh (approved):

PR #36285, opened 2025-06-23
chore: Mock Omnichannel provider

yash-rajpal:


tassoevan (commented):
I'm not 100% sure we want the `OmnichannelContext` (and other Omnichannel-related contexts) to live in `@rocket.chat/ui-contexts`. Despite the fact its provider is placed inside `MeteorProvider`, I guess it should be independent. I want to open a discussion for it.

PR #36292, opened 2025-06-24
feat(a11y): Add anchor-based navigation with router-level scroll support

ergot-rp:
## Summary

- Added `getLocationHash()` to the router context (`RouterContextValue`), alongside the existing `getLocationPathname()` and `getLocationSearch()`
- Created `useLocationHash` hook in `@rocket.chat/ui-contexts` for reactive hash access
- Scrolling to hash targets is handled automatically at the router level via `useRouterScrollToHash`, called by `RouterProvider` — no per-page scroll setup needed
- Pages can use `useLocationHash()` directly to react to hash presence (e.g., auto-expanding accordion sections)
- Links from the Preferences page navigate directly to the target field in the Accessibility page, with the relevant accordion section auto-expanding

## Test plan

- [ ] Navigate to `/account/preferences` and click \"Go to accessibility and appearance\" next to Message TimeFormat — should scroll to the clock mode field
- [ ] Repeat for Hide usernames and Hide roles links
- [ ] Verify the \"Adjustable layout\" accordion expands automatically when navigating via anchor
- [ ] Navigate to `/account/accessibility-and-appearance` directly (no hash) — accordion should behave as default (collapsed)
- [ ] Refresh the page with a hash in the URL (e.g., \`#clockMode\`) — should scroll to the correct field on load

## Issue(s)

[WA-50]

[WA-50]: https://rocketchat.atlassian.net/browse/WA-50?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/meteor/client/views/account/accessibility/AccessibilityPage.tsx
```diff
@@ -47,6 +49,17 @@ const AccessibilityPage = () => {
...
+	useEffect(() => {
+		if (location.hash) {
+			lastHash.current = location.hash.slice(1);
+		}
+
+		if (lastHash.current && document.getElementById(lastHash.current)) {
+			document.getElementById(lastHash.current)?.scrollIntoView({ behavior: 'smooth', block: 'start' });
+			lastHash.current = '';
+		}
+	}, []);
```
aleksandernsilva: The solution looks good, but I couldn't understand the purpose of the `lastHash` ref. Considering this effect will only run on mount, you could work directly with `window.hash` and subsequent scrolls would be handled by the browser default behavior for anchors. Let me know if I'm missing something.
```ts
	useEffect(() => {
		if (!location.hash) {
			return;
		}

		const anchorId = location.hash.slice(1);
		const anchorElement = document.getElementById(anchorId);
		anchorElement?.scrollIntoView({ behavior: 'smooth', block: 'start' });
	}, []);
```
This could be a good candidate for an external hook, allowing this code to be reused in other pages.

aleksandernsilva (changes requested):

apps/meteor/client/hooks/useScrollToHash.ts
```diff
@@ -0,0 +1,37 @@
+import { usePrefersReducedMotion } from '@rocket.chat/fuselage-hooks';
+import { useEffect } from 'react';
+
+export const FIELD_ANCHORS = {
```
aleksandernsilva: I agree we should have a single source of truth for the anchors, but considering this hook is generic, wouldn't it be better to keep their domains separate? 

Something similar to what we do for `queryKeys`. Using the hook would look something like:
```ts
const { ... } = useScrollToHash(fieldAnchors.accessibilityPage);
// OR
const { ... } = useScrollToHash([
  fieldAnchors.clockMode, 
  fieldAnchors.hideUsernames,
]);
```

apps/meteor/client/hooks/useScrollToHash.ts
```diff
@@ -0,0 +1,37 @@
...
+		}
+
+		const frame = requestAnimationFrame(() => {
+			document.getElementById(hash)?.scrollIntoView({ behavior: reducedMotion ? 'auto' : 'smooth', block: 'start' });
+		});
+
+		return () => cancelAnimationFrame(frame);
+	}, [hash, reducedMotion]);
+
+	return { shouldExpand };
```
aleksandernsilva: In the same line, `shouldExpand` seems implementation specific, targeted for Accordions. 
Although this is the main use currently, I'd recommend a generic return as well. (e.g `hasKnownTarget` or `hasTarget `). Nothing prevents us from renaming the variable where it's being used:

```ts
const { hasKnownTarget: shouldExpand } = useScrollToHash(...);
```

aleksandernsilva (changes requested):

apps/meteor/client/hooks/useHasValidLocationHash.ts
```diff
@@ -0,0 +1,17 @@
...
+import { availableHashes } from './useRouterScrollToHash';
+
+export { TARGET_ANCHORS } from './useRouterScrollToHash';
+
+/**
+ * Returns whether the current URL hash matches a known target anchor.
+ * Useful for auto-expanding sections that contain the target element.
+ *
+ * @see docs/anchor-navigation.md
+ */
```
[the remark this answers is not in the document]
dougfabris: its a good pattern to add such comments in hooks

apps/meteor/client/hooks/useRouterScrollToHash.ts
```diff
@@ -0,0 +1,44 @@
...
+
+		const frame = requestAnimationFrame(() => {
+			document.getElementById(id)?.scrollIntoView({
+				behavior: reducedMotion ? 'auto' : 'smooth',
+				block: 'start',
+			});
+		});
+
+		return () => cancelAnimationFrame(frame);
+	}, [hash, id, hasValidHash, reducedMotion]);
```
aleksandernsilva: Nitpick: Both `id` and `hasValidHash` are derived from `hash`, could you move them within the effect and simplify de dependency list?


```ts
export const useRouterScrollToHash = (router: RouterContextValue) => {
	const hash = useSyncExternalStore(router.subscribeToRouteChange, router.getLocationHash);
	const reducedMotion = usePrefersReducedMotion();
	const previousHash = useRef(hash);

	useEffect(() => {
		const id = hash.slice(1);
		
		if (!availableHashes.has(id) || hash === previousHash.current) {
			previousHash.current = hash;
			return;
		}

         // [...]
         
	}, [hash, reducedMotion]);
```

aleksandernsilva (changes requested):

aleksandernsilva (approved):

PR #36293, opened 2025-06-24
fix(ux): improve delete account modal with inline errors

sandranymark:

This PR improves the Delete My Account modal (`ActionConfirmModal`) by adding inline error handling using `react-hook-form`:

- Replaces `useState` logic with `react-hook-form`
- Displays inline validation message when the input field is empty
- Shows error message `("Invalid password")` under the input field.
- Ensures the modal closes correctly after successful account deletion
- Improves accessibility via aria-* attributes

## Issue(s)
image

## Steps to test or reproduce

Navigate to **My Account** → **Profile**

1. **Click "Delete my account"**
2. **Test the following scenarios:**
3.  **Leave the input empty**
 Inline validation error should appear below the field
4.  **Enter an incorrect password**
"Invalid password" message should display under the field
5. **Enter the correct password**
Account is deleted, modal closes, user is logged out, and a success toast is shown

## Further comments

[WA-58]

## Summary by CodeRabbit

* **Bug Fixes**
  * Shows a clear “Invalid password” message on the field instead of a generic alert.
  * Focuses the input when there’s an error for quicker correction.
  * Closes the delete-account modal immediately after successful deletion.

* **Refactor**
  * Migrated the confirmation input to a form-based flow with async submission for smoother handling.
  * Improved validation and error display within the modal.
  * Enhanced accessibility with proper labels, error descriptions, and live announcements.

[WA-58]: https://rocketchat.atlassian.net/browse/WA-58?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/meteor/tests/e2e/delete-account.spec.ts
+++ b/apps/meteor/tests/e2e/delete-account.spec.ts
@@ -1,5 +1,6 @@
 import { DEFAULT_USER_CREDENTIALS } from './config/constants';
 import { AccountProfile, Registration, Authenticated } from './page-objects';
+import { setSettingValueById } from './utils';
 import { test, expect } from './utils/test';
 import { createTestUser, type ITestUser } from './utils/user-helpers';
 
@@ -12,8 +13,7 @@ test.describe('Delete Own Account', () => {
 	let userWithoutPermissions: ITestUser;
 
 	test.beforeAll(async ({ api }) => {
-		const response = await api.post('/settings/Accounts_AllowDeleteOwnAccount', { value: true });
-		expect(response.status()).toBe(200);
+		expect((await setSettingValueById(api, 'Accounts_AllowDeleteOwnAccount', true)).status()).toBe(200);
 		userToDelete = await createTestUser(api, { username: 'user-to-delete' });
 		userWithInvalidPassword = await createTestUser(api, { username: 'user-with-invalid-password' });
 		userWithoutPermissions = await createTestUser(api, { username: 'user-without-permissions' });
@@ -27,12 +27,9 @@ test.describe('Delete Own Account', () => {
 	});
 
 	test.afterAll(async ({ api }) => {
-		await Promise.all([
-			api.post('/settings/Accounts_AllowDeleteOwnAccount', { value: false }).then((res) => expect(res.status()).toBe(200)),
-			userToDelete.delete(),
-			userWithInvalidPassword.delete(),
-			userWithoutPermissions.delete(),
-		]);
+		expect((await setSettingValueById(api, 'Accounts_AllowDeleteOwnAccount', false)).status()).toBe(200);
+		await userWithInvalidPassword.delete();
+		await userWithoutPermissions.delete();
 	});
 
 	test('should not delete account when invalid password is provided', async ({ page }) => {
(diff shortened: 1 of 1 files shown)


apps/meteor/client/views/account/profile/AccountProfilePage.tsx
```diff
@@ -92,18 +92,24 @@ const AccountProfilePage = (): ReactElement => {
...
 				dispatchToastMessage({ type: 'success', message: t('User_has_been_deleted') });
+				setModal(null);
 				logout();
 			} catch (error: any) {
 				if (error.error === 'user-last-owner') {
 					const { shouldChangeOwner, shouldBeRemoved } = error.details;
 					return handleConfirmOwnerChange(passwordOrUsername, shouldChangeOwner, shouldBeRemoved);
 				}
 
-				dispatchToastMessage({ type: 'error', message: error });
```
gabriellsh, line 106: I noticed you removed this toast dispatch, but what if there are other errors? Such as a network error or something else... I think we should probably keep this as the "last resort" error handling.

gabriellsh (changes requested):
There's some linting errors, would it be possible to fix those?

About the overall solution: It's a nice approach, but I think we could make it a little more declarative/generic if instead of using this "dependency inversion" pattern (accepting a callback to set the error in the confirmation fn), we could throw a custom error with the correct message, and wrap the the fn invocation in a try catch.

```ts
export class InputError extends Error {
	constructor(message: string) {
		super(message);
	}
}
```

```ts
if (error.errorType === 'error-invalid-password') {
	throw new InputError(t('Invalid_password'));
}
```

```ts
try {
	await onConfirm(inputText);
	onCancel();
} catch (error) {
	if (error instanceof InputError) {
		setError('credential', { message: error.message });
	}
}
```

This way we decouple the modal use from it's action, and we can also emit other types of errors in the future if we wish to.

apps/meteor/client/views/account/profile/ActionConfirmModal.tsx
```diff
@@ -1,44 +1,48 @@
...
+				setFocus('credential');
 				return;
 			}
-			onConfirm(inputText);
-			onCancel();
-		},
-		[inputText, onConfirm, onCancel, t],
-	);
+		}
+	});
```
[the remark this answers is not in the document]
sandranymark: okey, this is awesome! !
I’ll fix the bug :D

apps/meteor/client/views/account/profile/ActionConfirmModal.tsx
```diff
@@ -1,43 +1,46 @@
...
+import { InputError } from './DeleteAccountError';
 
 type ActionConfirmModalProps = {
 	isPassword: boolean;
-	onConfirm: (input: string) => void;
+	onConfirm: (input: string) => Promise<void>;
 	onCancel: () => void;
 };
 
 // TODO: Use react-hook-form
```
aleksandernsilva:
```suggestion
```

apps/meteor/client/views/account/profile/ActionConfirmModal.tsx
```diff
@@ -1,43 +1,46 @@
...
-			e.preventDefault();
-			if (inputText === '') {
-				setInputError(t('Invalid_field'));
+	const handleSave = handleSubmit(async ({ credential }) => {
+		try {
+			await onConfirm(credential);
+		} catch (error) {
+			if (error instanceof InputError) {
+				setError('credential', { message: error.message });
+				setFocus('credential');
```
aleksandernsilva: Nitpick: Looking at the `react-hook-forms` [documentation](https://react-hook-form.com/docs/useform/seterror) it is possible to pass the config `shouldFocus: true` to the `setError` method. This could eliminate the need of calling `setFocus` on L38.

dougfabris:
why u did this to me 😢 ?

PR #36305, opened 2025-06-25
chore: Remove deprecated `pdfTranscriptRequested` index

KevLehman:
## Issue(s)

https://rocketchat.atlassian.net/browse/CTZ-104


apps/meteor/server/startup/migrations/index.ts
```diff
@@ -25,4 +25,5 @@ import './v316';
 import './v317';
 import './v318';
```
KevLehman:
```suggestion
```

ggazzo (approved):

PR #36317, opened 2025-06-27
fix: change sanitizeUrl to handle URLs without a protocol schema

julio-rocketchat:
After adding a new function that sanitizes URLs and blocks the usage of dangerous protocol schemas such as `javascript:`, a bug was introduced. URLs without a predefined protocol will return `#`: 

image

This PR changes that behavior to ensure that URLs without a predefined protocol will return a protocol-relative URL:

image

URLs with protocols are returned normally:

image

While dangerous protocol schemas are blocked:

image

## Issue(s)
TBD

## Steps to test or reproduce
N/A

## Further comments
N/A

### DIFF
--- a/packages/gazzodown/src/elements/sanitizeUrl.spec.ts
+++ b/packages/gazzodown/src/elements/sanitizeUrl.spec.ts
@@ -95,10 +95,6 @@ describe('sanitizeUrl', () => {
 		});
 	});
 
-	it('sanitizes malformed URLs', () => {
-		expect(sanitizeUrl('ht^tp://broken')).toBe('#');
-	});
-
 	it('sanitizes empty string', () => {
 		expect(sanitizeUrl('')).toBe('#');
 	});
@@ -107,7 +103,7 @@ describe('sanitizeUrl', () => {
 		expect(sanitizeUrl('JAVASCRIPT:alert(1)')).toBe('#');
 	});
 
-	it('sanitizes nonsense input', () => {
-		expect(sanitizeUrl('💣💥🤯')).toBe('#');
+	it('allows bare domain names', () => {
+		expect(sanitizeUrl('example.com/page')).toBe('//example.com/page');
 	});
 });
--- a/packages/gazzodown/src/elements/sanitizeUrl.ts
+++ b/packages/gazzodown/src/elements/sanitizeUrl.ts
@@ -1,8 +1,18 @@
 export const sanitizeUrl = (href: string) => {
+	if (!href) {
+		return '#';
+	}
+
 	try {
-		const url = new URL(href);
-		const dangerousProtocols = ['javascript:', 'data:', 'vbscript:'];
-		return dangerousProtocols.includes(url.protocol.toLowerCase()) ? '#' : url.href;
+		const hasProtocol = /^[a-zA-Z][a-zA-Z\d+\-.]*:/.test(href);
+
+		if (hasProtocol) {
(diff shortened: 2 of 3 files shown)


packages/gazzodown/src/elements/sanitizeUrl.ts
```diff
@@ -1,8 +1,18 @@
...
-		return dangerousProtocols.includes(url.protocol.toLowerCase()) ? '#' : url.href;
+		const hasProtocol = /^[a-zA-Z][a-zA-Z\d+\-.]*:/.test(href);
+
+		if (hasProtocol) {
+			const url = new URL(href);
+			const dangerousProtocols = ['javascript:', 'data:', 'vbscript:'];
+			return dangerousProtocols.includes(url.protocol.toLowerCase()) ? '#' : url.href;
+		}
+
+		return `//${href}`;
```
abhinavkrin, line 15: The current logic turns #foo, ?x=y, and /path into protocol-relative URLs (//#foo, //?x=y, ///path), which breaks links.

packages/gazzodown/src/elements/sanitizeUrl.ts
```diff
@@ -1,8 +1,18 @@
 export const sanitizeUrl = (href: string) => {
+	if (!href) {
+		return '#';
+	}
+
```
abhinavkrin, line 5:
```suggestion
	if (s.startsWith('#') || s.startsWith('?') || s.startsWith('/') || s.startsWith('//')) {
		return s;
	}
```
Maybe?

abhinavkrin: Maybe we could add tests for them.

julio-rocketchat: That check is just to ensure that an empty string would return `#`. In theory, anything starting with `?`, `/` or `//` won't work since the parser won't parse it:

image

(Well, even an empty string won't work though since the parser won't know what to do with it tbh)

abhinavkrin: Thanks, It makes sense in context of the parser to ignore these.

packages/gazzodown/src/elements/sanitizeUrl.ts
julio-rocketchat: Seems to be the intended behavior. This is how it behaved before 7.7.0 (where `sanitizeUrl` was introduced):

image

packages/gazzodown/src/elements/sanitizeUrl.spec.ts
```diff
@@ -107,7 +103,7 @@ describe('sanitizeUrl', () => {
 		expect(sanitizeUrl('JAVASCRIPT:alert(1)')).toBe('#');
 	});
 
-	it('sanitizes nonsense input', () => {
```
pierre-lehnen-rc, line 110: I think the old behavior was ok here, when we look at sanitizeUrl as a unit. Maybe we shouldn't change sanitizeUrl itself, but the input that we pass to it? If `example.com` gets turned into `https://example.com` at some point, then perhaps we should be passing `https://example.com` to sanitizeUrl.

julio-rocketchat: I don't think the parser nowadays turns example.com into https://example.com or http://example.com. Maybe if we were to change that it should likely be in a different PR. idk

PR #35884, opened 2025-04-27
feat: Add OpenAPI Support to Webdav API 

ahmed-n-abdeltwab:
**Description:**  
This PR integrates OpenAPI support into the `Rocket.Chat API`, migrate of `Rocket.Chat API` endpoints to the new OpenAPI pattern. The update includes improved API documentation, enhanced type safety, and response validation using AJV.

**Key Changes:**  
- **Implemented the new pattern** and added AJV-based JSON schema validation for API.  
- **(legacy)** Improved the code structure using chained route methods. 
- **(New)** Uses the ExtractRoutesFromAPI utility from the TypeScript definitions to dynamically derive the routes from the endpoint specifications.
- **Enabled Swagger UI** integration for this API.
- **This does not introduce any breaking changes** to the endpoint logic.

**Issue Reference:**  
Relates to #34983, part of the ongoing OpenAPI integration effort.

**Testing:**  
- Verified that the API response schemas are correctly documented in Swagger UI.  
- All tests passed without any breaking changes

**Endpoints:**
- [x] Webdav:
   - [x] [Get WebDAV Accounts](https://developer.rocket.chat/apidocs/get-webdav-accounts)

     Get WebDAV Accounts Screenshot
     
   - [x] [Remove WebDAV Account](https://developer.rocket.chat/apidocs/remove-webdav-account)

     Remove WebDAV Account

Looking forward to your feedback! 🚀

---

This pull request introduces OpenAPI support to the WebDAV API in the Rocket.Chat repository. The changes are made in the `apps/meteor/app/api/server/v1/webdav.ts` file. The update refactors two WebDAV API endpoints, `getMyAccounts` and `removeWebdavAccount`, to utilize a modern chained route definition syntax (`API.v1.get`, `API.v1.post`). Additionally, it replaces a local AJV instance with a shared one from `@rocket.chat/rest-typings`, enhancing the API's robustness. Comprehensive AJV-based validation schemas for request bodies and responses are added, improving the overall validation process. The source branch for this feature is `feat/openapi-webdav`, and it targets the `develop` branch.
https://rocketchat.atlassian.net/browse/ARCH-1695

PR #36318, opened 2025-06-27
refactor: update ExtractValidation type to use 'unknown'

ahmed-n-abdeltwab:
**description:**

Change the ExtractValidation type to return 'unknown' instead of 'never' for the case where the params return never'. Instead of never, unknown should be good.

**Tests:**

```javascript

```


apps/meteor/app/api/server/ApiClass.ts
```diff
@@ -64,7 +64,7 @@ export type Prettify<T> = {
 	[K in keyof T]: T[K];
 } & unknown;
 
-type ExtractValidation<T> = T extends ValidateFunction<infer TSchema> ? TSchema : never;
+type ExtractValidation<T> = T extends ValidateFunction<infer TSchema> ? TSchema : unknown;
```
cardoso: Can you add a type test ensuring this fixes the issue you have in #35884 ?

cardoso (changes requested):

apps/meteor/app/api/server/ApiClass.ts
ahmed-n-abdeltwab: Ok, sure

apps/meteor/app/api/server/ApiClass.ts
```diff
@@ -64,7 +64,7 @@ export type Prettify<T> = {
 	[K in keyof T]: T[K];
 } & unknown;
 
-type ExtractValidation<T> = T extends ValidateFunction<infer TSchema> ? TSchema : never;
+export type ExtractValidation<T> = T extends ValidateFunction<infer TSchema> ? TSchema : unknown;
```
cardoso: The type `ExtractValidation` is an implementation detail and shouldn't be exported.

Instead, you should test `ExtractRoutesFromAPI`.

Your test must ensure that the change you're making here will fix the type error in #35884 .


`client/hooks/webdav/useWebDAVAccountIntegrationsQuery.ts(22,31): error TS2554: Expected 1 arguments, but got 0.`

`useEndpoint('GET', '/v1/webdav.getMyAccounts')` must return a function that can be called with zero params.

ahmed-n-abdeltwab: Got it. Thank you for the information I learned. Okay, I will do that. In progress...

apps/meteor/app/api/server/ApiClass.spec.ts
```diff
@@ -39,3 +39,239 @@ it('Should return the expected type', () => {
...
+					200: ValidateFunction<unknown>;
+				};
+				authRequired: true;
+			}
+		>;
+		type ExpectedFunctionSignature = Expect<
+			ShallowEqual<ExtractRoutesFromAPI<APIWithNeverQuery>['/v1/endpoint.test']['GET'], (params: unknown) => unknown>
+		>;
+		true as ExpectedFunctionSignature;
+	});
```
ahmed-n-abdeltwab: This testing ensures that the WebDAV's `getMyAccounts` GET method with a query, if not present, tests return parameters unknown.

apps/meteor/app/api/server/ApiClass.spec.ts
```diff
@@ -39,3 +39,239 @@ it('Should return the expected type', () => {
...
+				body: ValidateFunction<null>;
+				authRequired: true;
+			}
+		>;
+		type ExpectedFunctionSignature = Expect<
+			ShallowEqual<ExtractRoutesFromAPI<APIWithNullBody>['/v1/endpoint.test']['POST'], (params: null) => unknown>
+		>;
+		true as ExpectedFunctionSignature;
+	});
+});
```
ahmed-n-abdeltwab: As for all of that, it's not related to the problem, but it won't do any harm if we check for other cases. Let me know if you still want this PR specific for the webdav.getMyAccounts or what. I'm excited for your insights.

apps/meteor/app/api/server/ApiClass.spec.ts
cardoso: Doesn't unknown still mean the function will require 1 parameter? Except instead of never it will be unknown?

ahmed-n-abdeltwab: I’m not sure I fully understand what you mean, but from what I know, unknown is like a parent type for everything – objects, strings, numbers, and so on. A function can also be assigned to unknown. On the other hand, never is like the end point; it represents something that can never exist. For example, an object of type never is like having no object at all – similar to how 0 represents nothing in numbers. You could say it’s "not an object" in this sense. Regarding getMyAccounts, it returns the lexically scoped this as a recursive function within the prototype object, but since never represents something impossible, it wouldn’t allow that behavior.

ahmed-n-abdeltwab: > client/hooks/webdav/useWebDAVAccountIntegrationsQuery.ts(22,31): error TS2554: Expected 1 arguments, but got 0.

Oh, I see what you mean; I will look at it and tell you why it failed.

apps/meteor/app/api/server/ApiClass.spec.ts
cardoso: It’s better to test specifically your issue, adding tests once new issues are found and fixed. A test acts like a specification (hence the .spec.ext convention), which means covering more cases than what the fix proposes asserts that the extra scenarios are correct. Further fixes would need to go and fix the specification, which defeats its purpose.

apps/meteor/app/api/server/ApiClass.spec.ts
ahmed-n-abdeltwab: I was wrong. We should use `void` instead of never or unknown. void is used when there’s no specific data type, such as for a function return type. never is for cases where a value can’t exist at all, and unknown is like a safer version of any. tell me if you agree, and I’ll make the changes

ahmed-n-abdeltwab: @cardoso just giving you a quick ping

apps/meteor/app/api/server/ApiClass.spec.ts
ahmed-n-abdeltwab: Sure, I will go ahead and do that.

apps/meteor/app/api/server/ApiClass.spec.ts
cardoso: The most correct here would be () instead of `(params: unknown)`. Although `(params?: undefined)` or even `(params?: never)` might be good enough.

See this typescript playground: https://www.typescriptlang.org/play/?#code/FAEwpgxgNghgTmABBA9gOwM4BdEDMUoBciAFAA7wwC2GxaYAbmHAJSIC8AfIgK5oDWaFAHc0AbmD4UJFhKkkA3gF9ZkgiT7hcAS3ohV8gNoBdQwAZjq0JFgJk6bIgBG8YuUo1ifQSLRsuvAJCohIucDKh8IoqkeGaYDp6qmEkJuaWEtbQ8EiomDguAF5uFHDUGAD8XmhaumD6HNzeweLARRFtMIXRyV0aNQl1+pHdaRZW4Nl2eY4AjjwAHiUelXSMzP5NQb4S8wsdez27i-21Scf7YxnAWba5Djjzi26bgT4hwE-7ql9HnzwneKJeo-AGXUzjMRAA

ahmed-n-abdeltwab: the update in progress...

ahmed-n-abdeltwab:
This is also an issue with POST; perhaps changes could also be made to POST, PUT, etc.

apps/meteor/app/api/server/ApiClass.ts
```diff
@@ -73,7 +73,9 @@ type ConvertToRoute<TRoute extends MinimalRoute> = {
 	[K in TRoute['path']]: {
 		[K2 in Extract<TRoute, { path: K }>['method']]: K2 extends 'GET'
 			? (
-					params: ExtractValidation<Extract<TRoute, { path: K; method: K2 }>['query']>,
+					...args: [ExtractValidation<Extract<TRoute, { path: K; method: K2 }>['query']>] extends [never]
+						? [params?: never]
+						: [params: ExtractValidation<Extract<TRoute, { path: K; method: K2 }>['query']>]
 				) => ExtractValidation<Extract<TRoute, { path: K; method: K2 }>['response'][200]>
```
ahmed-n-abdeltwab, line 79: Here, I'm checking if ExtractValidation returns never; if it does, it returns params?: never. Otherwise, it returns ExtractValidation.

PR #36319, opened 2025-06-27
regression: Remove pdf option in export messages as file

tiagoevanp:
## Issue(s)

https://rocketchat.atlassian.net/browse/CORE-1228


apps/meteor/tests/e2e/export-messages.spec.ts
```diff
@@ -25,12 +25,36 @@ test.describe.serial('export-messages', () => {
...
+		await expect(poHomeChannel.tabs.exportMessages.getMethodOptionByName('Send file via email')).toBeVisible();
+		await expect(poHomeChannel.tabs.exportMessages.getMethodOptionByName('Download file')).toBeVisible();
+	});
+
+	test('should display export output format correctly depending on the selected method', async () => {
+		await poHomeChannel.sidenav.openChat(targetChannel);
+		await poHomeChannel.tabs.kebab.click({ force: true });
+		await poHomeChannel.tabs.btnExportMessages.click();
+
+		await expect(poHomeChannel.tabs.exportMessages.getOutputFormat('HTML')).toHaveClass(new RegExp('disabled'));
```
MartinSchoeler: Why not use [toBeDisabled](https://playwright.dev/docs/api/class-locatorassertions#locator-assertions-to-be-disabled)?

MartinSchoeler (changes requested):

apps/meteor/tests/e2e/export-messages.spec.ts
tiagoevanp: I will create a TODO for it, the component has a button that is not disabled, only the class is added... the base component should be reviewed

MartinSchoeler (approved):

PR #36329, opened 2025-06-30
chore: enable inspecting of deno subprocess in apps-engine

d-gubert:

Add an option to the AppsEngineDenoRuntime that makes it spawn Deno subprocesses with the `--inspect-wait` flag, so that you can attach a debugger to it.
Also disables all timeouts controlling subprocess communications, as those would make the engine attempt to restart the Deno subprocess after a while


d-gubert:
Deno runtime has been removed and conflicts are deep. This will require rework from the ground up

PR #36331, opened 2025-06-30
regression: unread banner appears for thread messages after channel creation

abhinavkrin:

When you create a brand-new channel and immediately reply to your own message in a thread, the unread-messages banner was showing up even though you hadn’t missed anything. This happened because the baseline timestamp (lastMessageDate) stayed undefined until an off-screen message was detected, so every reply looked like “new” content. Now, on first load—when the message count goes from zero to some number—we set lastMessageDate to the newest message’s timestamp, giving the unread-count logic a proper anchor and preventing self-sent thread replies from ever triggering the banner.

## Further comments

[CORE-1214](https://rocketchat.atlassian.net/browse/CORE-1214)

[CORE-1214]: https://rocketchat.atlassian.net/browse/CORE-1214?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/meteor/client/views/room/body/hooks/useUnreadMessages.ts
+++ b/apps/meteor/client/views/room/body/hooks/useUnreadMessages.ts
@@ -55,6 +55,7 @@ export const useHandleUnread = (
 	const getMessage = Messages.use((state) => state.get);
 	const findFirstMessage = Messages.use((state) => state.findFirst);
 	const filterMessages = Messages.use((state) => state.filter);
+	const messagesCount = Messages.use((state) => state.records.size);
 
 	if (!chat) {
 		throw new Error('No ChatContext provided');
@@ -73,8 +74,9 @@ export const useHandleUnread = (
 			return;
 		}
 		setMessageJumpQueryStringParameter(message?._id);
+		chat.readStateManager.markAsRead();
 		setUnreadCount(0);
-	}, [room._id, setUnreadCount, findFirstMessage, unread?.since]);
+	}, [room._id, setUnreadCount, findFirstMessage, unread?.since, chat.readStateManager]);
 
 	const handleMarkAsReadButtonClick = useCallback(() => {
 		chat.readStateManager.markAsRead();
@@ -180,6 +182,22 @@ export const useHandleUnread = (
 		},
 		[getMessage, setUnreadCount],
 	);
+
+	useEffect(() => {
+		// When a room is first opened, there’s no “off-screen” message yet, so we set
+		// an initial baseline timestamp (the newest message’s ts) as our anchor.
+		// This ensures our unread-count logic only starts counting messages that arrive
+		// after the initial load, rather than everything already in the DOM.
+		if (lastMessageDate === undefined && messagesCount > 0) {
+			const newest = findFirstMessage(
+				(r) => r.rid === room._id,
+				(a, b) => b.ts.getTime() - a.ts.getTime(),
+			);
+			if (newest) {
+				setLastMessageDate(newest.ts);
+			}
+		}
+	}, [messagesCount, lastMessageDate, findFirstMessage, room._id]);
 
 	return {
 		innerRef: ref,
--- a/apps/meteor/client/views/room/body/hooks/useUnreadMessages.ts
+++ b/apps/meteor/client/views/room/body/hooks/useUnreadMessages.ts
@@ -55,7 +55,6 @@ export const useHandleUnread = (
 	const getMessage = Messages.use((state) => state.get);
 	const findFirstMessage = Messages.use((state) => state.findFirst);
 	const filterMessages = Messages.use((state) => state.filter);
-	const messagesCount = Messages.use((state) => state.records.size);
 
 	if (!chat) {
 		throw new Error('No ChatContext provided');
@@ -92,7 +91,8 @@ export const useHandleUnread = (
 		const count = filterMessages(
 			(record) =>
 				record.rid === room._id &&
-				record.ts.getTime() <= (lastMessageDate?.getTime() ?? Infinity) &&
+				!!lastMessageDate &&
+				record.ts.getTime() <= lastMessageDate?.getTime() &&
 				record.ts.getTime() > (subscription?.ls?.getTime() ?? -Infinity),
 		).length;
 
@@ -182,22 +182,6 @@ export const useHandleUnread = (
 		},
 		[getMessage, setUnreadCount],
 	);
-
-	useEffect(() => {
-		// When a room is first opened, there’s no “off-screen” message yet, so we set
-		// an initial baseline timestamp (the newest message’s ts) as our anchor.
-		// This ensures our unread-count logic only starts counting messages that arrive
-		// after the initial load, rather than everything already in the DOM.
-		if (lastMessageDate === undefined && messagesCount > 0) {
-			const newest = findFirstMessage(
-				(r) => r.rid === room._id,
-				(a, b) => b.ts.getTime() - a.ts.getTime(),
-			);
-			if (newest) {
-				setLastMessageDate(newest.ts);
-			}
-		}
-	}, [messagesCount, lastMessageDate, findFirstMessage, room._id]);
 
 	return {
 		innerRef: ref,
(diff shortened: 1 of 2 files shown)


apps/meteor/client/views/room/body/hooks/useUnreadMessages.ts
```diff
@@ -73,8 +74,9 @@ export const useHandleUnread = (
 			return;
 		}
 		setMessageJumpQueryStringParameter(message?._id);
+		chat.readStateManager.markAsRead();
```
abhinavkrin, line 77: This seemed to be a fix for me. While clicking on the banner, It would take me to message. But when page is refreshed, the banner would reapper.

apps/meteor/tests/e2e/threads.spec.ts
```diff
@@ -15,6 +15,19 @@ test.describe.serial('Threads', () => {
 		await page.goto('/home');
 		await poHomeChannel.sidenav.openChat(targetChannel);
 	});
+
+	test('expect no unread banner when replying to a thread in a fresh channel', async ({ page }) => {
+		await poHomeChannel.content.sendMessage('parent for unread-banner test');
+		await page.locator('[data-qa-type="message"]').last().hover();
+		await page.getByRole('button', { name: 'Reply in thread' }).click();
+		const threadDialog = page.getByRole('dialog');
+		await threadDialog.locator('[name="msg"]').last().fill('first thread reply');
```
jessicaschelly: Maybe `openReplyInThread` and `sendMessageInThread` would fit here?

apps/meteor/tests/e2e/threads.spec.ts
```diff
@@ -15,6 +15,19 @@ test.describe.serial('Threads', () => {
...
+		await page.keyboard.press('Enter');
+
+		await page.waitForTimeout(200);
```
jessicaschelly, line 26: Instead of the timeout, could we wait for the message to be sent/visible?

jessicaschelly (changes requested):

apps/meteor/tests/e2e/threads.spec.ts
abhinavkrin: I tried that. But didnt work. It takes some time for the banner to appear. Since they are initially not present for some time, the tests doesnt find it.

jessicaschelly (dismissed):

apps/meteor/client/views/room/body/hooks/useUnreadMessages.ts
```diff
@@ -181,6 +183,22 @@ export const useHandleUnread = (
 		[getMessage, setUnreadCount],
 	);
 
+	useEffect(() => {
```
MartinSchoeler: Could we find an alternative to this use effect? As it is, this is going to run every time the messagesCount prop changes (basically on every message). It seems costly for an operation that only needs to happen when the room is initialized.

apps/meteor/client/views/room/body/hooks/useUnreadMessages.ts
```diff
@@ -181,6 +183,22 @@ export const useHandleUnread = (
...
+		// an initial baseline timestamp (the newest message’s ts) as our anchor.
+		// This ensures our unread-count logic only starts counting messages that arrive
+		// after the initial load, rather than everything already in the DOM.
+		if (lastMessageDate === undefined && messagesCount > 0) {
+			const newest = findFirstMessage(
+				(r) => r.rid === room._id,
+				(a, b) => b.ts.getTime() - a.ts.getTime(),
+			);
+			if (newest) {
+				setLastMessageDate(newest.ts);
```
MartinSchoeler: We should avoid changing the value of a state that is in the dependency list

MartinSchoeler (changes requested):
Do we know the root cause of the issue (in what PR the bug was introduced)? Here it seems we are only patching the symptom, not actually fixing the root cause of the issue.
I think we can dive a little deeper and come up with a more elegant solution.

apps/meteor/client/views/room/body/hooks/useUnreadMessages.ts
abhinavkrin: Actually this will run only once. In first run, the lastMessageDate will be set and there is a check to return early when lastMessageDate is set.

apps/meteor/client/views/room/body/hooks/useUnreadMessages.ts
abhinavkrin: same.

abhinavkrin:
> Do we know the root cause of the issue (in what PR the bug was introduced)? Here it seems we are only patching the symptom, not actually fixing the root cause of the issue. I think we can dive a little deeper and come up with a more elegant solution.

1. The root cause is as explained, we are using the last message which is hidden behind the scroll or near the scroll to calculate the lastMessageDate. But when a room is new, there arent a lot of messages so that a message is behind the scroll or near to it. This made lastMessageDate undefined.
2. Earlier the count code was: 
```
const count = Messages.find({
			rid: room._id,
			ts: { $lte: lastMessageDate, $gt: subscription?.ls },
		}).count();
```
This will always return nothing sinc lastMessageDate is undefined.
But now the code is
```
const count = filterMessages(
			(record) =>
				record.rid === room._id &&
				record.ts.getTime() <= (lastMessageDate?.getTime() ?? Infinity) &&
				record.ts.getTime() > (subscription?.ls?.getTime() ?? -Infinity),
		).length;
```
This will always return a number > 0 if lastMessageDate is undefined.
So this code change made the already existing bug to appear now.

apps/meteor/client/views/room/body/hooks/useUnreadMessages.ts
abhinavkrin: Also, this is not expensive tasks. It is used only at one or two places all over the client. Plus it only runs once and loops around few messages. Next runs, it returns early. I dont think it is expensive.

MartinSchoeler:
> 
> ```
> const count = filterMessages(
> 			(record) =>
> 				record.rid === room._id &&
> 				record.ts.getTime() <= (lastMessageDate?.getTime() ?? Infinity) &&
> 				record.ts.getTime() > (subscription?.ls?.getTime() ?? -Infinity),
> 		).length;
> ```


I think we could fix the issue here, add an exception before we filter the messages since when the lastMessage is undefined, the filter does nothing (since infinity will always be larger). 

just by doing 

```
const count = filterMessages(
			(record) =>
				record.rid === room._id && !!lastMessageDate &&
				record.ts.getTime() <= lastMessageDate?.getTime() &&
				record.ts.getTime() > (subscription?.ls?.getTime() ?? -Infinity),
		).length;
```

the issue seems to go away (just an example, need testing), so I think we don't need another useEffect, just to adjust the existing one 

Also it seems that this issue was introduced in https://github.com/RocketChat/Rocket.Chat/pull/36158

abhinavkrin:
> > ```
> > const count = filterMessages(
> > 			(record) =>
> > 				record.rid === room._id &&
> > 				record.ts.getTime() <= (lastMessageDate?.getTime() ?? Infinity) &&
> > 				record.ts.getTime() > (subscription?.ls?.getTime() ?? -Infinity),
> > 		).length;
> > ```
> 
> I think we could fix the issue here, add an exception before we filter the messages since when the lastMessage is undefined, the filter does nothing (since infinity will always be larger).
> 
> just by doing
> 
> ```
> const count = filterMessages(
> 			(record) =>
> 				record.rid === room._id && !!lastMessageDate &&
> 				record.ts.getTime() <= lastMessageDate?.getTime() &&
> 				record.ts.getTime() > (subscription?.ls?.getTime() ?? -Infinity),
> 		).length;
> ```
> 
> the issue seems to go away (just an example, need testing), so I think we don't need another useEffect, just to adjust the existing one
> 
> Also it seems that this issue was introduced in #36158

Yes it will go back to the previous behavior. But still the issue with how we calculate lastMessageDate will remain same. It will be undefined for new rooms. Since there wont be enough messages which will cause a message to go invisible. And hence unread message banner might not show for new rooms.

MartinSchoeler (approved):

MartinSchoeler:
Behavior matches version 7.7.0

Issue #35485, opened 2025-03-13
Livechat Widget tries to connect to localhost


billynoah:
### Description:

Recently upgraded to V 7.4 and afterward the livechat widget began attempting to connect to localhost. The widget renders the icon and loads but somewhere along the way after it loads all the chunks it tries to connect with localhost:3000: It appears this is the js chunk responsible for the wrong connection with localhost: [92839.chunk.bf9fa.js](http://45.56.109.221/shared/92839.chunk.bf9fa.js)

Once the widget is loaded, I can see requests like this happening:
```text
http://localhost:3000/api/v1/livechat/config?token=5ed70de03b98b461651ccd2d3e661a931ef6ff2de270b5788755edd9dc60842c
```

I have reviewed the suggestion here: https://docs.rocket.chat/docs/livechat-widget-installation

The reverse proxy config does point to the same url as the widget install script.

In Administration > Workspace > Settings > General the Site URL is "https://chat.example.com/"

Here is my apache config file:

```text
ServerName chat.example.com
ErrorLog /var/log/apache2/chat-error.log
CustomLog /var/log/apache2/chat-access.log combined
<Location />
    Require all granted
</Location>
RewriteEngine On
RewriteCond %{HTTP:CONNECTION} Upgrade [NC]
RewriteCond %{HTTP:Upgrade} =websocket [NC]
RewriteRule /(.*)           ws://localhost:3000/$1 [P,L]
RewriteCond %{HTTP:Upgrade} !=websocket [NC]
RewriteRule /(.*)           http://localhost:3000/$1 [P,L]
ProxyPassReverse /          http://localhost:3000/
SSLCertificateFile /etc/letsencrypt/live/chat.example.com/fullchain.pem
SSLCertificateKeyFile /etc/letsencrypt/live/chat.example.com/privkey.pem
Include /etc/letsencrypt/options-ssl-apache.conf
```

And here is my livechat widget script:

```javascript
(function(w, d, s, u) {
    w.RocketChat = function(c) { w.RocketChat._.push(c) }; w.RocketChat._ = []; w.RocketChat.url = u;
    var h = d.getElementsByTagName(s)[0], j = d.createElement(s);
    j.async = true; j.src = 'https://chat.example.com/livechat/rocketchat-livechat.min.js?_=202503120000';
    h.parentNode.insertBefore(j, h);
})(window, document, 'script', 'https://chat.example.com/livechat')
```

This was all working perfectly before the update.

[...]

PR #36332, opened 2025-06-30
fix: Livechat widget connecting to localhost instead of actual site url

KevLehman:
## Issue(s)

Basically, since we changed how settings are initialized on https://github.com/RocketChat/Rocket.Chat/pull/34199, the file that "obtained" the server url to preppend to the livechat widget chunks was using the `localhost` url (default) instead of using the url the server was changed to.

This fix should allow it to be more reactive and change everytime the server url changes by just refreshing the widget on browser.

## Further comments

Fixes https://github.com/RocketChat/Rocket.Chat/issues/35485

### DIFF
--- a/apps/meteor/app/livechat/server/livechat.ts
+++ b/apps/meteor/app/livechat/server/livechat.ts
@@ -7,7 +7,7 @@ import { WebApp } from 'meteor/webapp';
 import { settings } from '../../settings/server';
 import { addServerUrlToIndex } from '../lib/Assets';
 
-const indexHtmlWithServerURL = addServerUrlToIndex((await Assets.getTextAsync('livechat/index.html')) || '');
+const indexHtmlWithServerURL = await Assets.getTextAsync('livechat/index.html');
 
 function parseExtraAttributes(widgetData: string): string {
 	const liveChatAdditionalScripts = settings.get<string>('Livechat_AdditionalWidgetScripts');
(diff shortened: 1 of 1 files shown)


lucas-a-pelegrino (approved):

aleksandernsilva (dismissed):

scuciatto:
Can we have a test for this?

KevLehman:
It would be nice, but actually apart from some units i dont see much value 🤔 and even units, i dont see much value either cause the function is just 2 lines...

but if u want, i can try to add some.

scuciatto:
> It would be nice, but actually apart from some units i dont see much value 🤔 and even units, i dont see much value either cause the function is just 2 lines...
> 
> but if u want, i can try to add some.

It’s not something I want to insist on, but since this is a fix, we should include a test to ensure it doesn’t break again in the future.

aleksandernsilva (approved):

pierre-lehnen-rc (approved):
Note: the unit tests on this PR simply check if the function that replaces URLs is replacing with the right value, assuming a valid input.

It doesn't check if the actual file gets replaced properly - but if that were to fail, then the existing tests would catch the problem already.

The existing tests didn't catch this bug because all of our CI and development environments run in the localhost URL which is used by default before the settings are loaded. Maybe we should have at least some testing environments run on a custom URL...

PR #36377, opened 2025-07-07
feat: Outbound Comms endpoints

lucas-a-pelegrino:

Implement new API endpoints to allow for listing currently listed providers and fetch provider metadata.

`GET /api/v1/omnichannel/outbound/providers:`

Endpoint should be filterable by provider type → phone | email

The endpoint should return all providers currently registered for the given type.

`GET /api/v1/omnichannel/outbound/providers/:id/metadata:` (Needs Apps bridge to work)

Endpoint should return the specified provider’s metadata

:id → appId
## Issue(s)

https://rocketchat.atlassian.net/browse/CTZ-182


apps/meteor/ee/app/livechat-enterprise/server/api/lib/outbound.ts
```diff
@@ -0,0 +1,33 @@
...
+
+export class OutboundMessageProviderService {
+	private readonly provider: OutboundMessageProvider;
+
+	constructor() {
+		this.provider = new OutboundMessageProvider();
+	}
+
+	private isProviderValid(type?: string): type is 'phone' | 'email' {
+		return type === 'phone' || type === 'email';
```
KevLehman:
```suggestion
		return ['phone' | 'email'].includes(type);
```

apps/meteor/ee/app/livechat-enterprise/server/api/lib/outbound.ts
```diff
@@ -0,0 +1,33 @@
...
+import { OutboundMessageProvider } from '../../../../../../server/lib/OutboundMessageProvider';
+
+export class OutboundMessageProviderService {
+	private readonly provider: OutboundMessageProvider;
+
+	constructor() {
+		this.provider = new OutboundMessageProvider();
+	}
+
+	private isProviderValid(type?: string): type is 'phone' | 'email' {
```
KevLehman:
```suggestion
	private isProviderValid(type?: string): type is ValidProviderType {
```

And then we do a 
```
type ValidProviderType = 'phone' | 'email'
```

in case we don't have this type somewhere already. wdyt?

apps/meteor/ee/app/livechat-enterprise/server/api/outbound.ts
```diff
@@ -0,0 +1,35 @@
...
+
+			const providers = outboundMessageProvider.listOutboundProviders(type);
+
+			try {
+				return API.v1.success({
+					providers,
+				});
+			} catch (error) {
+				return API.v1.failure(error);
+			}
```
KevLehman:
```suggestion
			return API.v1.success({
					providers,
			});
```

apps/meteor/ee/app/livechat-enterprise/server/api/lib/outbound.ts
```diff
@@ -0,0 +1,33 @@
...
+
+		const providers = this.provider.getOutboundMessageProviders(type);
+
+		return providers.map<IOutboundProvider>((provider: IOutboundMessageProviders) => {
+			return {
+				providerId: provider.appId,
+				providerName: provider.name,
+				supportsTemplates: true,
+				providerType: provider.type,
+			};
```
KevLehman: Instead of remapping here, think we can just return what the provider returns? wdyt?

apps/meteor/ee/app/livechat-enterprise/server/api/outbound.ts
```diff
@@ -0,0 +1,35 @@
+import { API } from '../../../../../app/api/server';
+import { isGETOutboundProviderParams } from '../outboundcomms/rest';
+import { OutboundMessageProviderService } from './lib/outbound';
+
+const outboundMessageProvider = new OutboundMessageProviderService();
```
KevLehman: Instead of instantiating here, let the lib/outbound file to export this instantiation so apps engine can reuse it

KevLehman (changes requested):

apps/meteor/ee/app/livechat-enterprise/server/api/lib/outbound.ts
```diff
@@ -0,0 +1,26 @@
...
+import { OutboundMessageProvider } from '../../../../../../server/lib/OutboundMessageProvider';
+
+export class OutboundMessageProviderService {
+	private readonly provider: OutboundMessageProvider;
+
+	constructor() {
+		this.provider = new OutboundMessageProvider();
+	}
+
+	private isProviderValid(type?: string): type is ValidProviderTypes {
```
KevLehman: You don't need type to be undefined here. `isProviderValid` is only called after checking type is not undefined.

packages/core-typings/src/omnichannel/outbound.ts
```diff
@@ -107,10 +107,17 @@ type TemplateParameter =
...
-	supportsTemplates: boolean;
+	supportsTemplates?: boolean;
 	providerType: 'phone' | 'email';
 };
 
 export type IOutboundProviderMetadata = IOutboundProvider & {
 	templates: Record<string, IOutboundProviderTemplate[]>;
 };
+
+export const ProviderTypeEnum = {
```
KevLehman: maybe just an array is enough?

KevLehman (changes requested):

KevLehman (approved):

aleksandernsilva (dismissed):

apps/meteor/ee/app/livechat-enterprise/server/api/lib/outbound.ts
```diff
@@ -0,0 +1,26 @@
...
+import { OutboundMessageProvider } from '../../../../../../server/lib/OutboundMessageProvider';
+
+export class OutboundMessageProviderService {
+	private readonly provider: OutboundMessageProvider;
+
+	constructor() {
+		this.provider = new OutboundMessageProvider();
+	}
+
+	private isProviderValid(type: any): type is ValidOutboundProvider {
```
MarcosSpessatto, line 13:
```suggestion
	private isProviderValid(type: unknown): type is ValidOutboundProvider {
```

apps/meteor/ee/app/livechat-enterprise/server/api/outbound.ts
```diff
@@ -0,0 +1,28 @@
...
+
+API.v1.addRoute(
+	'omnichannel/outbound/providers/:id/metadata',
+	{ authRequired: true },
+	{
+		async get() {
+			return API.v1.success();
+		},
+	},
+);
```
MarcosSpessatto: Would it be possible to use the new way of registering endpoints? Example: https://github.com/RocketChat/Rocket.Chat/blob/develop/apps/meteor/app/api/server/v1/custom-sounds.ts#L43

This way the endpoints will be already present in the Swagger docs 😃

MarcosSpessatto: Also, we need to add e2e tests here, correct? 🤔

MarcosSpessatto (changes requested):

apps/meteor/ee/app/livechat-enterprise/server/api/outbound.ts
KevLehman: Not for now, as the endpoints don't fully work (there's no way to use the providers yet)

apps/meteor/ee/app/livechat-enterprise/server/api/lib/outbound.ts
```diff
@@ -0,0 +1,26 @@
...
+import { OutboundMessageProvider } from '../../../../../../server/lib/OutboundMessageProvider';
+
+export class OutboundMessageProviderService {
+	private readonly provider: OutboundMessageProvider;
+
+	constructor() {
+		this.provider = new OutboundMessageProvider();
+	}
+
+	private isProviderValid(type: unknown): type is ValidOutboundProvider {
```
KevLehman:
```suggestion
	private isProviderValid(type: any): type is ValidOutboundProvider {
```

@MarcosSpessatto , the problem with unknown is that you have to do extra typing shenanigans to make the `includes` work, and at the end, if the value is included we assure it's the one we want, so for this case, i think `any` is fair.

KevLehman (approved):

PR #36386, opened 2025-07-09
feat: add `separateResponse` param to incoming webhooks for optional per-channel responses markdown

abhinavkrin:

Adds a new boolean parameter `separateResponse` to the incoming webhook configuration. When enabled, the API will return an array of result objects—one per channel—indicating individual success or error. If `separateResponse` is disabled (the default), the server will first validate all target channels; on validation failure it returns a single `success: false` without sending any messages, or on success it sends to all channels and returns `success: true`.

This PR also fixes an issue where in an outgoing integrations, if the message returned from the webhook had multiple channels, the messages would be sent but it would show only one channel in history.

## Steps to test or reproduce

1. Create an incoming integration.
2. Add post as user
3. Add two channels, one the user has access to and the another the user does not has access to.
4. Execute the integration with different values of `separateResponse` in the payload.  When true -> separate response will be returned. In case of falsy -> message wont be sent to either of the channels since it would have failed for the channel which the user can't access.
## Further comments

[SUP-779](https://rocketchat.atlassian.net/browse/SUP-779)

[SUP-779]: https://rocketchat.atlassian.net/browse/SUP-779?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/meteor/app/integrations/server/api/api.ts
```diff
@@ -217,16 +223,23 @@ async function executeIntegrationRest(
...
-
-		return API.v1.success(scriptResponse);
+		if (bodyParams.separateResponse) {
+			const allFailed = messageResponse.every((response) => 'error' in response && response.error);
+			if (allFailed) {
+				return API.v1.failure({ responses: messageResponse });
+			}
+			return API.v1.success({ responses: messageResponse });
+		}
+		return API.v1.success();
```
MarcosSpessatto, line 248: I think we must keep the scriptResponse, correct?

apps/meteor/app/lib/server/functions/processWebhookMessage.ts
```diff
@@ -1,10 +1,10 @@
-import type { IMessage, IUser, RequiredField, MessageAttachment } from '@rocket.chat/core-typings';
+import type { IMessage, IUser, RequiredField, MessageAttachment, IRoom } from '@rocket.chat/core-typings';
```
MarcosSpessatto, line 1: Please, do not refactor the function while we are creating a feature. If the refactor is needed, we need to do it in a separate PR, it's really hard to understand what was indeed changed.

MarcosSpessatto (changes requested):

apps/meteor/app/integrations/server/api/api.ts
abhinavkrin: Yes, when the script returns an object which contains `resoonse`  field, the `response` will be returned by the webhook. Working as expected. Also added support for the separateResponse flag for script.

apps/meteor/app/lib/server/functions/processWebhookMessage.ts
abhinavkrin: I had to refactor due to couple of reasons
1. I ran into function size limit eslint error. So I separated the logic to build message and getRoomWithOptionToJoin into separate functions. 
2. I had to loop channels twice -  1. checking whether a message could be sent to the channels or not. 2. loop again to actually send messages.

MarcosSpessatto (approved):

PR #36399, opened 2025-07-11
Implementation and Tests for Sorting Pinned Messages

CaioPacheco:

  Description of the changes made:
  - Added functionality to sort pinned messages in the `GET /api/v1/chat.getPinnedMessages` endpoint.
  - Implemented the `sort` parameter to sort messages based on the `timestamp (ts)` field.
  - Added automated tests to ensure that messages are correctly sorted based on the `sort` parameter.
  - Refactored code for better clarity and readability.

  This change allows the `/api/v1/chat.getPinnedMessages` query to sort pinned messages according to the `sort` parameter value. Messages can be sorted by timestamp either in ascending or descending order depending on the parameter's value.

## Issue(s)
  This PR resolves the following issue:
  - Adds the ability to sort pinned messages in the `chat.getPinnedMessages` endpoint based on the `sort` parameter.

  If this PR addresses a specific issue or bug, reference the issue number here:
  - Fixes the bug related to sorting pinned messages (#36087 ).

## Steps to test or reproduce
  To test the functionality, follow these steps:
  1. Send a `GET` request to the endpoint `/api/v1/chat.getPinnedMessages?roomId=room123&sort={"ts":-1}`.
  2. Check that the response contains messages sorted by the `ts` (timestamp) field in descending order (most recent first).
  3. Also verify that if the `sort` parameter is invalid, the `error-invalid-sort` error is returned.
  4. Automated tests can also be run with the command `yarn mocha --require ts-node/register apps/meteor/app/api/server/test/**/*.test.ts`.

  Make sure sorting works correctly with other `sort` parameter values (e.g., `{"ts":1}` for ascending order).

## Further comments
  The solution was chosen based on the need to allow sorting of pinned messages, which will improve the user experience when querying pinned messages in channels.
  Considerations:
  - The `sort` parameter was validated to ensure only valid JSON values are accepted.
  - Other values may be added in the future to sort by other fields.
  Alternatives considered:
  - Using more complex database aggregations for sorting, but for simplicity and flexibility, the current implementation meets the use case.


apps/meteor/app/api/server/test/chatPinnedMessages.test.js
```diff
@@ -0,0 +1,18 @@
+const request = require('supertest');
+const { expect } = require('chai');
+const path = require('path');
+const app = require(path.resolve(__dirname, '../../../../app'));  
+
+describe('GET /api/v1/chat.getPinnedMessages', function() {
+  it('deve retornar as mensagens fixadas ordenadas corretamente pelo parâmetro sort', async function() {
```
jessicaschelly, line 7: every code on the codebase should be written in english, including comments

apps/meteor/app/api/server/test/chatPinnedMessages.test.js
jessicaschelly, line 1: this test could be inside apps/meteor/tests/end-to-end/api/chat.ts

jessicaschelly, line 1: some ideas of tests to add:
should accept empty sort parameter
should return error for invalid value in sort parameter
should work without sort parameter - default sorting

apps/meteor/app/api/server/v1/chat.ts
```diff
@@ -492,33 +492,51 @@ API.v1.addRoute(
...
+      if (!(await canAccessRoomIdAsync(roomId, this.userId))) {
+        throw new Meteor.Error('error-not-allowed', 'Not allowed');
+      }
+
+      let sortParam: Record<string, 1 | -1> = {};  // Inicializa o parâmetro de ordenação
+
+      // Agora, a ordenação está sendo aplicada
+      try {
+        if (sort) {
+          sortParam = JSON.parse(sort);  // Converte o parâmetro 'sort' da string JSON para um objeto
```
jessicaschelly, line 511: just a suggestion - I noticed you're using manual JSON.parse() for the sort parameter. The codebase has a built-in parseJsonQuery() method that handles this securely, maybe it could be used here too?

package.json
jessicaschelly, line 1: try resetting the changes of this file, we already have this dependency on `apps/meteor/package.json`

jessicaschelly (changes requested):

jessicaschelly:
Note: This review was focused on code quality, consistency, and adherence to existing codebase standards rather than technical implementation details.

PR #36424, opened 2025-07-15
feat: adds new endpoint to fetch outbound provider metadata

lucas-a-pelegrino:

### DIFF
--- a/apps/meteor/ee/app/livechat-enterprise/server/api/outbound.ts
+++ b/apps/meteor/ee/app/livechat-enterprise/server/api/outbound.ts
@@ -1,4 +1,4 @@
-import type { IOutboundProvider } from '@rocket.chat/core-typings';
+import type { IOutboundProvider, IOutboundProviderMetadata } from '@rocket.chat/core-typings';
 import { ajv } from '@rocket.chat/rest-typings/src/v1/Ajv';
 
 import { API } from '../../../../../app/api/server';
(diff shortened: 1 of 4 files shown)


apps/meteor/server/lib/OutboundMessageProvider.ts
```diff
@@ -50,6 +50,12 @@ export class OutboundMessageProvider implements IOutboundMessageProvider {
 			}));
 	}
 
+	public getProviderById(appId: string): IOutboundPhoneMessageProvider | undefined {
+		const providers = this.outboundMessageProviders.get('phone') as IOutboundPhoneMessageProvider[] | undefined;
```
sampaiodiego: why is `phone` hardcoded here? 

I think you should either make it clear on the method it returns only phone provider by calling it something like `getPhoneProviderById` or you should get all providers and not only `phone` ones

packages/apps-engine/src/definition/outboundComunication/IOutboundCommsProvider.ts
```diff
@@ -1,9 +1,9 @@
 import type { IOutboundMessage } from './IOutboundMessage';
 import type { IOutboundProviderTemplate } from './IOutboundProviderTemplate';
 
-type ProviderMetadata = {
-	appId: string;
-	appName: string;
+export type ProviderMetadata = {
+	providerId: string;
+	providerName: string;
```
sampaiodiego: I think using the prefix `provider` here very redundant.. the type is already called `ProviderMetadata`, there is no need to use a `provider` prefix for each property:

```suggestion
	id: string;
	name: string;
```

same applies to `providerType` that should be called only `type`

.changeset/giant-toes-decide.md
```diff
@@ -0,0 +1,6 @@
+---
+"@rocket.chat/meteor": patch
+"@rocket.chat/apps-engine": patch
```
sampaiodiego: since this is something, we consider it a `minor` change (same reason as you chose the `feat:` as the PR title).. also I think all changes from apps-engine were removed, so no need to mention it on the changeset
```suggestion
"@rocket.chat/meteor": minor
```

sampaiodiego (changes requested):

apps/meteor/ee/app/livechat-enterprise/server/api/outbound.ts
```diff
@@ -1,55 +1,70 @@
...
-									type: 'string',
-								},
-								documentationUrl: {
-									type: 'string',
+const outboundCommsEndpoints = API.v1
+	.get(
+		'omnichannel/outbound/providers',
+		{
+			response: {
+				200: ajv.compile<{ providers: IOutboundProvider[] }>({
```
KevLehman: You're defining the other 2 in a separate file, wanna do the same for this one?

apps/meteor/ee/app/livechat-enterprise/server/api/outbound.ts
```diff
@@ -1,55 +1,70 @@
...
+			return API.v1.success({
+				providers,
+			});
+		},
+	)
+	.get(
+		'omnichannel/outbound/providers/:id/metadata',
+		{
+			response: {
+				200: isOutboundProviderMetadataSchema,
```
KevLehman: The endpoint could return a 400 response, in those cases, this will fail as there's no definition for a 400. Can you add one?

KevLehman (changes requested):

apps/meteor/ee/app/livechat-enterprise/server/api/outbound.ts
```diff
@@ -1,55 +1,50 @@
...
-					},
-				},
-			}),
+const outboundCommsEndpoints = API.v1
+	.get(
+		'omnichannel/outbound/providers',
+		{
+			response: {
+				200: GETOutboundProvidersResponseSchema,
+				400: IsOutboundProviderBadRequestErrorSchema,
```
KevLehman: nitpick, feel free to ignore if not valid: why the naming of these 2 are different? some are upper case, some are lowercase, one starts with GET and the rest with Is or is 🤔

lucas-a-pelegrino: Yeah, I'm actually unsure on how to name these Error responses...

The GET prefix for the endpoint request params feels right to me.

But the error responses... do you have any suggestions?

KevLehman: what about
200: GETxxxxxxResponse
400: GETxxxxxErrorResponse

sampaiodiego (approved):

PR #36433, opened 2025-07-15
fix: await subscription updates in notifyUsersOnMessage function

ggazzo:
~https://rocketchat.atlassian.net/browse/CORE-1239~
closes https://github.com/RocketChat/Rocket.Chat/pull/36433

### DIFF
--- a/apps/meteor/app/lib/server/lib/notifyUsersOnMessage.ts
+++ b/apps/meteor/app/lib/server/lib/notifyUsersOnMessage.ts
@@ -94,18 +94,40 @@ async function updateUsersSubscriptions(message: IMessage, room: IRoom): Promise
 	const { toAll, toHere, mentionIds } = mentions;
 	const userIds = [...new Set([...mentionIds, ...highlightIds])];
 	const unreadCount = getUnreadSettingCount(room.t);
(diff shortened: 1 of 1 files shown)


apps/meteor/app/lib/server/lib/notifyUsersOnMessage.ts
```diff
@@ -137,7 +137,6 @@ async function updateUsersSubscriptions(message: IMessage, room: IRoom): Promise
 	]);
 
 	// update subscription of the message sender
-	await Subscriptions.setAsReadByRoomIdAndUserId(message.rid, message.u._id);
 	const setAsReadResponse = await Subscriptions.setAsReadByRoomIdAndUserId(message.rid, message.u._id);
 	if (setAsReadResponse.modifiedCount) {
```
pierre-lehnen-rc, line 146: btw, this `modifiedCount` will always be truthy for existing subscriptions, because `setasReadyByRoomIdAndUserId` will always set `ls: new Date()`

ggazzo: i think for this case this is ok but yeah the query could be improved somehow

sampaiodiego (approved):

PR #36452, opened 2025-07-16
chore: Implement base signaling system for media calls

pierre-lehnen-rc:

This PR implements the base signaling system for media calls, along with a basic webrtc implementation. It doesn't integrate this new system with the rocket.chat UI so it's not  usable by itself. Changes are still expected on this codebase as we implement more features and add SIP integration.

## Issue(s)

[VAI-14](https://rocketchat.atlassian.net/browse/VAI-14)
[VAI-83](https://rocketchat.atlassian.net/browse/VAI-83)

## Further comments

This PR introduces two new packages, and modifies a few others:
### media-signaling: 
This package includes all of the business logic that clients need to implement in order to receive calls. It also exports all the types that are shared between client and server. It is meant to be published to npm, so it doesn't have any dependency on other packages from the workspace. That way any client that wants to implement media calls can import this package to get the full functionality while having only to implement the UI for it.

### media-calls:
This package implements the server logic for the whole media call system, without integrating it with rocket.chat itself (aka this package simply being there won't do anything at all). It imports things from other packages when it made sense to implement them on existing packages (like models and some core typings), but otherwise is mostly self-contained.

### other modified packages:
Some existing packages such as core-services and ddp-client (as well as the main meteor app) were modified to actually connect the rocket.chat core with the media-calls package. In general those do not include any business logic for the media call system.

[VAI-14]: https://rocketchat.atlassian.net/browse/VAI-14?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

* **New Features**
  * 1:1 WebRTC calling: ringing/accept/active/hangup, renegotiation, transfer, per-session client API and real-time per-user media signaling.

* **Bug Fixes**
  * Expired/stale calls are automatically hung up during startup/license validation to prevent lingering calls.

* **Chores**
  * Added media-call service, typings, models, client libraries, notifications integration, service registration, packaging, and Docker build updates to enable media calling.


apps/meteor/ee/server/configuration/voip.ts
```diff
@@ -1,10 +1,11 @@
...
 import { addSettings } from '../settings/voip';
 
-Meteor.startup(async () => {
-	License.onValidateLicense(async () => {
-		await addSettings();
-	});
+License.onValidateLicense(async () => {
+	await addSettings();
+
+	await cronJobs.add('expire-media-calls', '* * * * *', () => MediaCall.hangupExpiredCalls());
```
debdutdeb: Crons will have a default delay for querying what needs to be run - that's fine. But this expression won't this cause higher db load, considerably? Because of updating the job documents with locks on every interval, updating finish metadata?

apps/meteor/server/services/media-call/service.ts
```diff
@@ -0,0 +1,98 @@
...
+				type: 'user',
+				id: caller.uid,
+				contractId: caller.contractId,
+			},
+			callee: { type: 'user', id: user._id },
+		});
+	}
+
+	public async callUser(caller: { uid: IUser['_id']; contractId: string }, userId: IUser['_id']): Promise<IMediaCall> {
+		// For now every call is between two users
```
cardoso: Can you add "TODO: " to comments like this? Makes it explicit and easy to search.

ee/packages/media-calls/src/agents/Manager.ts
```diff
@@ -0,0 +1,302 @@
...
+		const endedBy = agent.actor;
+
+		const stateResult = await MediaCalls.hangupCallById(agent.callId, { endedBy, reason });
+		if (!stateResult.modifiedCount) {
+			return;
+		}
+
+		agent.notify(agent.callId, 'hangup');
+		const otherAgent = await this.getAnyOppositeAgent(agent);
+		// #ToDo: log something if the otherAgent is not found
```
cardoso:
```suggestion
		// TODO: log something if the otherAgent is not found
```

ee/packages/media-calls/src/agents/Manager.ts
```diff
@@ -0,0 +1,302 @@
...
+	}
+
+	public async expireCall(call: Pick<IMediaCall, '_id' | 'caller' | 'callee'>): Promise<void> {
+		logger.debug({ msg: 'AgentManager.expireCall', call });
+
+		if (!(await this.hangupByServer(call._id, 'timeout'))) {
+			return;
+		}
+
+		// We send a signal to rocket.chat users involved on the call without instantiating a full agent
```
cardoso: Same

ee/packages/media-calls/src/agents/Manager.ts
```diff
@@ -0,0 +1,302 @@
...
+	public async acceptCall(agent: IMediaCallAgent): Promise<void> {
+		logger.debug({ msg: 'AgentManager.acceptCall', actor: agent.actor });
+		if (agent.role !== 'callee') {
+			return;
+		}
+
+		const stateResult = await MediaCalls.acceptCallById(agent.callId, agent.contractId, MediaCallMonitor.getNewExpirationTime());
+		// If nothing changed, the call was no longer ringing
+		if (!stateResult.modifiedCount) {
+			// Notify something?
```
cardoso: Same

ee/packages/media-calls/src/agents/Manager.ts
```diff
@@ -0,0 +1,302 @@
...
+			return;
+		}
+
+		agent.sign();
+
+		// #ToDo: notify client if this throws any error
```
cardoso:
```suggestion
		// TODO: notify client if this throws any error
```

ee/packages/media-calls/src/agents/Manager.ts
```diff
@@ -0,0 +1,302 @@
...
+			case 'caller':
+				return this.getAgentFactoryForActor(call.callee);
+			case 'callee':
+				return this.getAgentFactoryForActor(call.caller);
+			default:
+				return null;
+		}
+	}
+
+	// This will only return an agent if a contract for the other role has already been signed
```
cardoso:
```suggestion
	/** This will only return an agent if a contract for the other role has already been signed */
```

For visibility when hovering

ee/packages/media-calls/src/agents/Manager.ts
```diff
@@ -0,0 +1,302 @@
...
+		// We send the accepted notification immediately to report on the call state
+		await this.notifyAll([calleeAgent, callerAgent], callId, 'accepted');
+
+		const offer = await callerAgent.getLocalDescription();
+		if (!offer) {
+			await callerAgent.requestOffer({});
+			return;
+		}
+
+		// #ToDo: handle cases where an actor's service role doesn't match the call role
```
cardoso:
```suggestion
		// TODO: handle cases where an actor's service role doesn't match the call role
```

ee/packages/media-calls/src/agents/users/Agent.ts
```diff
@@ -0,0 +1,89 @@
...
+		logger.debug({ msg: 'UserMediaCallAgent.getRemoteDescription' });
+		const channel = await agentManager.getOrCreateContract(this.callId, this);
+
+		return channel?.localDescription ?? null;
+	}
+
+	public async requestOffer(params: { iceRestart?: boolean }): Promise<void> {
+		logger.debug({ msg: 'UserMediaCallAgent.requestOffer', params, actor: this.actor, contractState: this.contractState });
+
+		// #ToDo: this function may be called multiple times for the same call until an offer is provided; maybe consider the channel state before sending the offer
```
cardoso:
```suggestion
		// TODO: this function may be called multiple times for the same call until an offer is provided; maybe consider the channel state before sending the offer
```

ee/packages/media-calls/src/agents/users/BasicAgent.ts
```diff
@@ -0,0 +1,50 @@
...
+		const { _id: id, username, name: displayName, freeSwitchExtension: sipExtension } = this.user;
+
+		return {
+			type: 'user',
+			id,
+			username,
+			displayName,
+			sipExtension,
+			// avatarUrl,
+			// host
```
cardoso: Why is it commented out? Maybe add a TODO?

ee/packages/media-calls/src/agents/users/SignalProcessor.ts
```diff
@@ -0,0 +1,128 @@
...
+		// 2. the call has not been accepted yet and the signal came from a valid sesison from the callee
+
+		const params = { channel, call };
+
+		switch (signal.type) {
+			case 'local-sdp':
+				await this.saveLocalDescription(params, signal.sdp);
+				break;
+			case 'error':
+				// #ToDo
```
cardoso:
```suggestion
				// TODO: Implement error handling
```

ee/packages/media-calls/src/agents/users/SignalProcessor.ts
```diff
@@ -0,0 +1,128 @@
...
+		logger.debug({ msg: 'UserAgentSignalProcessor.processAnswer', params, answer });
+		const { call } = params;
+
+		switch (answer) {
+			case 'ack':
+				return this.processACK(params);
+			case 'accept':
+				return agentManager.acceptCall(this.agent);
+			case 'unavailable':
+				// return processUnavailable(call, channel);
```
cardoso:
```suggestion
				// TODO: return processUnavailable(call, channel);
```

ee/packages/media-calls/src/agents/users/SignalProcessor.ts
```diff
@@ -0,0 +1,128 @@
...
+
+		await this.agent.requestOffer(params);
+	}
+
+	private async reviewLocalState({ channel }: ChannelFunctionParams, signal: ClientMediaSignalLocalState): Promise<void> {
+		if (!this.agent.signed) {
+			return;
+		}
+
+		// #ToDo: Monitor other states as well
```
cardoso:
```suggestion
		// TODO: Monitor other states as well
```

ee/packages/media-calls/src/global/SignalGateway.ts
```diff
@@ -0,0 +1,52 @@
+import type { IUser } from '@rocket.chat/core-typings';
+import { isClientMediaSignal } from '@rocket.chat/media-signaling';
+import type { ClientMediaSignal, ServerMediaSignal } from '@rocket.chat/media-signaling';
+
+import { logger } from '../logger';
+import type { ISignalGateway, ServerSignalTransport } from './ISignalGateway';
+import { GlobalSignalProcessor } from './SignalProcessor';
+
+// Class used as gateway to send and receive signals to/from clients
+// The actual function used to send the signals needs to be set by the server
```
cardoso:
```suggestion
/**
 * Class used as gateway to send and receive signals to/from clients
 * The actual function used to send the signals needs to be set by the server
 */
```

ee/packages/media-calls/src/global/SignalProcessor.ts
```diff
@@ -0,0 +1,259 @@
...
+		}
+
+		const calleeAgent = await agentManager.getNewAgentForActor(callee, 'callee');
+		if (!calleeAgent) {
+			throw new Error('invalid-callee');
+		}
+
+		const service = requestedService || 'webrtc';
+
+		// webrtc is our only known service right now, but if the call was requested by a client that doesn't also implement it, we don't need to even create a call
```
cardoso:
```suggestion
		// TODO: webrtc is our only known service right now, but if the call was requested by a client that doesn't also implement it, we don't need to even create a call
```

ee/packages/media-calls/src/global/SignalProcessor.ts
```diff
@@ -0,0 +1,259 @@
...
+
+			await agent.processSignal(signal, call);
+		} catch (e) {
+			logger.error(e);
+			throw e;
+		}
+	}
+
+	private async processRegisterSignal(_uid: IUser['_id'], _signal: ClientMediaSignalRegister): Promise<void> {
+		// #ToDo: client registration
```
cardoso:
```suggestion
		// TODO: client registration
```

packages/core-typings/src/mediaCalls/IMediaCall.ts
```diff
@@ -0,0 +1,51 @@
...
+	id: 'server';
+};
+
+export type MediaCallContactInformation = {
+	displayName?: string;
+	username?: string;
+	sipExtension?: string;
+
+	// avatarUrl?: string;
+	// host?: string;
```
cardoso: TODO?

packages/media-signaling/src/definition/call/CallEvents.ts
```diff
@@ -0,0 +1,22 @@
+import type { ClientState } from '../client';
+import type { CallState } from './IClientMediaCall';
+
+export type CallEvents = {
+	// triggered when the state of a call changes (only for states known by the server); receives the old state as a param
```
cardoso:
```suggestion
	/** triggered when the state of a call changes (only for states known by the server); receives the old state as a param */
```

packages/media-signaling/src/definition/call/CallEvents.ts
```diff
@@ -0,0 +1,22 @@
...
+	stateChange: CallState;
+
+	// the client state is more detailed than the general call state, it includes sub-states that might not be relevant to the user or the UI
```
cardoso:
```suggestion
	/** the client state is more detailed than the general call state, it includes sub-states that might not be relevant to the user or the UI */
```

packages/media-signaling/src/definition/call/CallEvents.ts
```diff
@@ -0,0 +1,22 @@
...
+	clientStateChange: ClientState;
+
+	// contactUpdate may be triggered even if nothing actually changed
```
cardoso:
```suggestion
	/** contactUpdate may be triggered even if nothing actually changed */
```

packages/media-signaling/src/definition/signals/client/answer.ts
```diff
@@ -0,0 +1,36 @@
+import type { JSONSchemaType } from 'ajv';
+
+import type { CallAnswer } from '../../call';
+
+// Client is saying that the user accepted or rejected a call, or simply reporting that the user can or can't be reached
```
cardoso:
```suggestion
/** Client is saying that the user accepted or rejected a call, or simply reporting that the user can or can't be reached */
```

packages/media-signaling/src/definition/signals/client/error.ts
```diff
@@ -0,0 +1,34 @@
+import type { JSONSchemaType } from 'ajv';
+
+// Client is reporting an error
```
cardoso:
```suggestion
/** Client is reporting an error */
```

packages/media-signaling/src/definition/signals/client/hangup.ts
```diff
@@ -0,0 +1,36 @@
+import type { JSONSchemaType } from 'ajv';
+
+import type { CallHangupReason } from '../../call';
+
+// Client is saying they hanged up from a call. The reason specifies if its a clean hangup or an error
```
cardoso:
```suggestion
/** Client is saying they hanged up from a call. The reason specifies if its a clean hangup or an error */
```

packages/media-signaling/src/definition/signals/client/local-sdp.ts
```diff
@@ -0,0 +1,46 @@
+import type { JSONSchemaType } from 'ajv';
+
+// Client is sending the local session description to the server
```
cardoso:
```suggestion
/** Client is sending the local session description to the server */
```

packages/media-signaling/src/definition/signals/client/local-state.ts
```diff
@@ -0,0 +1,63 @@
+import type { JSONSchemaType } from 'ajv';
+
+import type { CallState } from '../../call';
+import type { ClientContractState, ClientState } from '../../client';
+
+// Client is sending their local call state
```
cardoso:
```suggestion
/** Client is sending their local call state */
```

packages/media-signaling/src/definition/signals/client/register.ts
```diff
@@ -0,0 +1,29 @@
+import type { JSONSchemaType } from 'ajv';
+
+// Client is reporting a brand new session
```
cardoso:
```suggestion
/** Client is reporting a brand new session */
```

packages/media-signaling/src/definition/signals/client/request-call.ts
```diff
@@ -0,0 +1,57 @@
+import type { JSONSchemaType } from 'ajv';
+
+import type { CallService } from '../../call';
+
+export type ClientMediaSignalRequestCall = {
+	// the callId on this signal is temporary and is never propagated to other agents
```
cardoso:
```suggestion
	/** the callId on this signal is temporary and is never propagated to other agents */
```

packages/media-signaling/src/definition/signals/server/new.ts
```diff
@@ -0,0 +1,15 @@
+import type { CallContact, CallRole, CallService } from '../../call';
+
+// Sent by the server to notify an agent that there's a new call for their actor
```
cardoso:
```suggestion
/** Sent by the server to notify an agent that there's a new call for their actor */
```

packages/media-signaling/src/definition/signals/server/new.ts
```diff
@@ -0,0 +1,15 @@
...
+export type ServerMediaSignalNewCall = {
+	callId: string;
+	type: 'new';
+
+	service: CallService;
+	kind: 'direct';
+	role: CallRole;
+	contact: CallContact;
+
+	// This will only be sent to the caller, with the id it used to request this call
```
cardoso:
```suggestion
	/** This will only be sent to the caller, with the id it used to request this call */
```

packages/media-signaling/src/definition/signals/server/notification.ts
```diff
@@ -0,0 +1,12 @@
+import type { CallNotification } from '../../call';
+
+// Server is sending a notification about the call state
```
cardoso:
```suggestion
/** Server is sending a notification about the call state */
```

packages/media-signaling/src/definition/signals/server/notification.ts
```diff
@@ -0,0 +1,12 @@
...
+export type ServerMediaSignalNotification = {
+	callId: string;
+	type: 'notification';
+
+	notification: CallNotification;
+	// Used to inform which contract was signed by the server, without targeting the signal to that contract alone
+	// Optional in general, but at least one notification must be sent with it before the callee can join the call
```
cardoso:
```suggestion
	/**
    * Used to inform which contract was signed by the server, without targeting the signal to that contract alone
	 * Optional in general, but at least one notification must be sent with it before the callee can join the call
    */
```

packages/media-signaling/src/definition/signals/server/remote-sdp.ts
```diff
@@ -0,0 +1,8 @@
+// Server is sending the other actor's sdp
```
cardoso:
```suggestion
/** Server is sending the other actor's sdp */
```

packages/media-signaling/src/definition/signals/server/request-offer.ts
```diff
@@ -0,0 +1,8 @@
+// Server is requesting a webrtc offer
```
cardoso:
```suggestion
/** Server is requesting a webrtc offer */
```

packages/media-signaling/src/lib/services/webrtc/Processor.ts
```diff
@@ -0,0 +1,325 @@
...
+
+	public async createOffer({ iceRestart }: { iceRestart?: boolean }): Promise<{ sdp: RTCSessionDescriptionInit }> {
+		this.config.logger?.debug('MediaCallWebRTCProcessor.createOffer');
+		await this.initializeLocalMediaStream();
+
+		if (iceRestart) {
+			this.restartIce();
+		}
+
+		// #ToDo: direction changes
```
cardoso:
```suggestion
		// TODO: direction changes
```

packages/media-signaling/src/lib/services/webrtc/Processor.ts
```diff
@@ -0,0 +1,325 @@
...
+
+	public async createAnswer({ sdp }: { sdp: RTCSessionDescriptionInit }): Promise<{ sdp: RTCSessionDescriptionInit }> {
+		this.config.logger?.debug('MediaCallWebRTCProcessor.createAnswer');
+		if (sdp.type !== 'offer') {
+			throw new Error('invalid-webrtc-offer');
+		}
+
+		await this.initializeLocalMediaStream();
+
+		// #ToDo: direction changes
```
cardoso:
```suggestion
		// TODO: direction changes
```

packages/media-signaling/src/lib/services/webrtc/Processor.ts
```diff
@@ -0,0 +1,325 @@
...
+		await this.peer.setLocalDescription(answer);
+
+		return this.getLocalDescription();
+	}
+
+	public async setRemoteDescription({ sdp }: { sdp: RTCSessionDescriptionInit }): Promise<void> {
+		this.config.logger?.debug('MediaCallWebRTCProcessor.setRemoteDescription');
+		await this.initializeLocalMediaStream();
+
+		// #ToDo: validate this.peer.signalingState ?
```
cardoso:
```suggestion
		// TODO: validate this.peer.signalingState ?
```

packages/media-signaling/src/lib/services/webrtc/Processor.ts
```diff
@@ -0,0 +1,325 @@
...
+		this.config.logger?.debug('MediaCallWebRTCProcessor.changeInternalState', stateName);
+		this.emitter.emit('internalStateChange', stateName);
+	}
+
+	private async getLocalDescription(): Promise<{ sdp: RTCSessionDescriptionInit }> {
+		this.config.logger?.debug('MediaCallWebRTCProcessor.getLocalDescription');
+		await this.waitForIceGathering();
+
+		// always wait a little extra to ensure all relevant events have been fired
+		await new Promise((resolve) => setTimeout(resolve, 30));
```
cardoso: Why 30? Seems arbitrary, maybe add more explanation or TODO.

packages/media-signaling/src/lib/services/webrtc/Processor.ts
```diff
@@ -0,0 +1,325 @@
...
+		this.config.logger?.error(event);
+		this.emitter.emit('internalError', { critical: false, error: 'ice-candidate-error' });
+	}
+
+	private onNegotiationNeeded(peer: RTCPeerConnection) {
+		if (peer !== this.peer) {
+			return;
+		}
+		this.config.logger?.debug('MediaCallWebRTCProcessor.onNegotiationNeeded');
+		// #ToDo negotiation needed
```
cardoso:
```suggestion
		// TODO: negotiation needed
```

packages/media-signaling/src/lib/utils/createRandomToken.ts
```diff
@@ -0,0 +1,8 @@
+export function createRandomToken(size: number, base = 32): string {
+	let token = '';
+	for (let i = 0; i < size; i++) {
+		const r: number = Math.floor(Math.random() * base);
```
cardoso: Why not crypto random?

cardoso (changes requested):
Mostly comment styling, but a couple questions as well

ee/packages/media-calls/src/global/SignalProcessor.ts
pierre-lehnen-rc: This one is not a todo item.

apps/meteor/ee/server/configuration/voip.ts
pierre-lehnen-rc: ~~it'll query often, but rarely find anything. Calls will only expire if all clients involved stop communicating with the server, or if the server itself go down.~~
oh you mean the document for the job itself. I hadn't considered that.

pierre-lehnen-rc: But I can also take this part off of this first PR and review it separately later (I'm not very happy about it either)

packages/media-signaling/src/lib/utils/createRandomToken.ts
pierre-lehnen-rc: I saw no need for it;

Also haven't checked if crypto is available on react-native (where this package will also be used)

cardoso: It's available/will be available through `expo-crypto` (cc @diegolmello) . You can also abstract away in "sans-io" style, providing an adapter for web. I'm doing the same for another project.

pierre-lehnen-rc: ok, but should I? there's no security context or anything like that to consider here.

cardoso: I think you should, but feel free to resolve here, since it's WIP anyway. Just make sure to add a TODO comment if you aren't 100% sure.

diegolmello: Not sure if we should, but `expo-crypto` should be able to handle that (or I can write whatever needed natively to do so, like I did for e2ee).

Github is complaining about this
image

pierre-lehnen-rc: I'll just rename it to `createWeakToken` instead and it should stop triggering the copilot warning. If that doesn't work I'll use a timestamp instead of a random string. There's really no point in using crypto for this.

sampaiodiego: do we need another token creation function out there? why not use `uuid` or `Random.id()`? if we need a new one, pls consider adding to `@rocket.chat/tools`

cardoso: Yep... Even if it's not security critical, if AI is detecting this, it may come up in security audits, and cause us some headaches.

So that's why at least a comment explaining why this isn't security-sensitive is warranted here:

```suggestion
		// SAFETY: This doesn't require a cryptographically strong RNG because [...]
		const r: number = Math.floor(Math.random() * base);
```

Although I'll tag @julio-rocketchat as this is his area.

apps/meteor/ee/server/configuration/voip.ts
sampaiodiego: it might be better to just use a `setTimeout` to have this routine running over and over..

packages/media-signaling/src/lib/utils/createRandomToken.ts
pierre-lehnen-rc: `@rocket.chat/tools` and `@rocket.chat/random` are not published to npm.

pierre-lehnen-rc: I'll just use `Date.now().toString()` instead

julio-rocketchat: Crypto random is generally the way to go when we need a cryptographically strong RNG, as pointed out by Cardoso. If it is not needed, that is, if the randomness of the value isn't something important, I'm okay with using `Math.random()`. If definitely depends on what we're using the token for. If this token will be used to identify and/or authenticate/authorize something important, I'd go with crypto random. I'd advise against timestamps most of the time as they're somewhat predictable.

(And Cardoso is right that it may come up in automated security audits, even if ends up being a false positive)

pierre-lehnen-rc: predictability is also not an issue here since this string is broadcasted publicly.

cardoso: I'll resolve this and approve since it seems to be addressed by [`5ea0a46` (#36452)](https://github.com/RocketChat/Rocket.Chat/pull/36452/commits/5ea0a46779d9e79d7eb0baa04b73e49cd1ed6f88)

cardoso (approved):

apps/meteor/ee/server/configuration/voip.ts
pierre-lehnen-rc: I have changed it to a bunch of setTimeouts

pierre-lehnen-rc:
Switched back to draft as I realized I'll need to change the indexes.

ee/packages/media-calls/src/server/CallDirector.ts
```diff
@@ -0,0 +1,432 @@
...
+	callerAgent: IMediaCallAgent;
+	calleeAgent: IMediaCallAgent;
+
+	webrtcOffer?: RTCSessionDescriptionInit;
+};
+
+// expiration checks by call id
+const scheduledExpirationChecks = new Map<string, ReturnType<typeof setTimeout>>();
+
+export class MediaCallDirector {
```
tassoevan, line 25: [nitpick] Consider refactoring it to not use a class as a namespace for functions.

tassoevan (approved):

apps/meteor/server/services/media-call/service.ts
```diff
@@ -0,0 +1,100 @@
...
+	private async sendSignal(toUid: IUser['_id'], signal: ServerMediaSignal): Promise<void> {
+		void api.broadcast('user.media-signal', { userId: toUid, signal });
+	}
+
+	private configureMediaCallServer(): void {
+		callServer.configure(this.getMediaServerSettings());
+	}
+
+	private getMediaServerSettings(): IMediaCallServerSettings {
+		const enabled = settings.get<boolean>('VoIP_TeamCollab_Enabled') ?? false;
```
KevLehman, line 72: Shouldn't we get the `??` part for free when using settings.get?

packages/media-signaling/src/definition/signals/client/error.ts
```diff
@@ -0,0 +1,47 @@
...
+			const: 'error',
+		},
+		errorType: {
+			type: 'string',
+			enum: ['signaling', 'service', 'other'],
+			nullable: true,
+		},
+		errorCode: {
+			type: 'string',
+			nullable: true,
```
KevLehman, line 38: If you're not accepting explicit `null`, you can just omit this prop as it's optional already.

packages/media-signaling/src/definition/signals/client/hangup.ts
```diff
@@ -0,0 +1,50 @@
...
+				'rejected',
+				'unavailable',
+				'transfer',
+				'timeout',
+				'signaling-error',
+				'service-error',
+				'media-error',
+				'error',
+			],
+			nullable: false,
```
KevLehman, line 45: iirc, if prop is required, `null` and `undefined` fail automatically

packages/media-signaling/src/definition/signals/client/error.ts
pierre-lehnen-rc: functionally, you're correct; but the type doesn't match without the nullable flag.

packages/media-signaling/src/definition/signals/client/hangup.ts
pierre-lehnen-rc: this one actually doesn't need the nullable, but it's thanks to the enum.

apps/meteor/server/services/media-call/service.ts
pierre-lehnen-rc: just a safeguard to ensure the data is an actual boolean even if the setting doesn't exist (the setting is created on license validation).

KevLehman (approved):

ee/packages/media-calls/src/server/CallDirector.ts
pierre-lehnen-rc: Will do it on another PR.

PR #36472, opened 2025-07-18
fix: room's "read" state getting desynced between client and server

pierre-lehnen-rc:

### DIFF
--- a/apps/meteor/app/lib/server/lib/notifyUsersOnMessage.ts
+++ b/apps/meteor/app/lib/server/lib/notifyUsersOnMessage.ts
@@ -100,7 +100,7 @@ async function updateUsersSubscriptions(message: IMessage, room: IRoom): Promise
 	const groupMentionInc = getGroupMentions(room.t, unreadCount as Exclude<UnreadCountType, 'user_mentions_only'>);
 
 	// find all subscriptions that will need to be notified after the update.
-	// we need to use toArray() here and keep results in memory because we'll update the them later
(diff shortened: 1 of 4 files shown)


apps/meteor/app/lib/server/lib/notifyUsersOnMessage.ts
```diff
@@ -119,9 +119,11 @@ async function updateUsersSubscriptions(message: IMessage, room: IRoom): Promise
 		await Subscriptions.incUnreadForRoomIdExcludingUserIds(room._id, [...userIds, message.u._id], 1);
 	}
 
+	const uids = subs.map((sub) => sub.u._id);
+
 	// update subscriptions of other members of the room
 	await Promise.all([
-		Subscriptions.setAlertForRoomIdExcludingUserId(message.rid, message.u._id),
+		Subscriptions.setAlertForRoomIdAndUserIds(message.rid, uids),
```
sampaiodiego, line 126: I'm a bit worried on how big `uids` can be.. it can have a few thousands of entries depending on the room size.

per [mongodb docs](https://www.mongodb.com/docs/manual/reference/operator/query/in/#syntax):

> It is recommended that you limit the number of parameters passed to the [$in](https://www.mongodb.com/docs/manual/reference/operator/query/in/#mongodb-query-op.-in) operator to tens of values. Using hundreds of parameters or more can negatively impact query performance.

so what if we use something time based? like we the current time at the beginning of the function and only update subscription records that haven't been updates up to that date.. this way we can prevent overwriting new data.

PR #36493, opened 2025-07-22
regression: missing font color in App Status Tag

MartinSchoeler:

Restores one of the font colors on the App Status Tag

## Issue(s)

[ALT-27](https://rocketchat.atlassian.net/browse/ALT-27)
## Further comments

I've not added tests as the status component needs some serious refactoring before we could do snapshot tests for visual regressions. 

[ALT-27]: https://rocketchat.atlassian.net/browse/ALT-27?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/meteor/client/views/marketplace/AppDetailsPage/tabs/AppStatus/AppStatus.tsx
+++ b/apps/meteor/client/views/marketplace/AppDetailsPage/tabs/AppStatus/AppStatus.tsx
@@ -111,6 +111,10 @@ const AppStatus = ({ app, showStatus = true, isAppDetailsPage, installed, ...pro
 			return Palette.statusColor['status-font-on-warning'].toString();
 		}
 
+		if (status.type === 'danger') {
+			return Palette.statusColor['status-font-on-danger'].toString();
+		}
+
 		return Palette.text['font-default'].toString();
 	};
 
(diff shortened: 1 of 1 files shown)


gabriellsh (dismissed):

apps/meteor/client/views/marketplace/AppDetailsPage/tabs/AppStatus/AppStatus.tsx
```diff
@@ -111,6 +111,10 @@ const AppStatus = ({ app, showStatus = true, isAppDetailsPage, installed, ...pro
 			return Palette.statusColor['status-font-on-warning'].toString();
 		}
 
+		if (status.type === 'danger') {
+			return Palette.statusColor['status-font-on-danger'].toString();
```
juliajforesti: You don't need the `Palette` for this (and the other variants on this function).
Passing directly the string works and is more straight forward.
```
color='status-font-on-warning'
```

I would suggest rewriting it to:
```
color={status.type === 'warning' ? 'status-font-on-warning' : 'default'}
```

juliajforesti (changes requested):

apps/meteor/client/views/marketplace/AppDetailsPage/tabs/AppStatus/AppStatus.tsx
MartinSchoeler: > You don't need the `Palette` for this (and the other variants on this function). Passing directly the string works and is more straight forward.
> 
> ```
> color='status-font-on-warning'
> ```
> 
> I would suggest rewriting it to:
> 
> ```
> color={status.type === 'warning' ? 'status-font-on-warning' : 'default'}
> ```

I think we will not be able to escape using ifs, since there are 3 possible variations: danger, warning and default

But I'll swap the palette access with the strings

PR #31549, opened 2024-01-26
feat: Message keyboard navigability

dougfabris:
## Proposed changes (including videos or screenshots)

- [x] Message focus style
- [x] Message arrow key navigation
- [x] Message toolbar arrow key navigation
- [x] Usercard opens with keys
- [x] Message focus on Threads

## Further comments

[DSAR-428]
[DSAR-450]

[DSAR-428]: https://rocketchat.atlassian.net/browse/DSAR-428?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[DSAR-450]: https://rocketchat.atlassian.net/browse/DSAR-450?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #36531, opened 2025-07-25
fix: Prevent auto-scroll to thread parent message on click

### LINKED ISSUE #32710: Prevent auto-scroll to thread start message
### Description:

When a thread-related message is clicked in a channel the application scrolls back to the main message where the thread started.

The problem is that having long-running threads this scroll-up ruins the usability of the application because users are forced to scroll down to previous part (which can be tedious amount of paging) over and over again if they need to participate in multiple threads.

We haven't found any global and/or user settings to control this behaviour and that's why it's marked as bug (otherwise a feature a request).

Any workaround is appreciated!

### Steps to reproduce:

1. Navigate to channel
2. Click on threaded message that was started lots of messages earlier
3. The window scrolls back to the start message of the thread

### Expected behavior:

The window remains at the same place and only the thread sidebar should be used to scroll back (if needed). Even if we need to jump back to the original message RocketChat already has a functionality to that (Jump to message).

### Actual behavior:

The window scrolls back every time a threaded message is clicked.

### Server Setup Information:

- Version of Rocket.Chat Server: 6.7.5 (stable)
- Operating System: Debian stable
- Deployment Method: docker
- DB Replicaset Oplog: no
- NodeJS Version: docker
- MongoDB Version: 6.0

### Client Setup Information

- Desktop App or Browser Version: 4.0.0 / Chrome latest
- Operating System: Windows 10 / Web

onozaty:
Remove unintended navigation to parent message when clicking thread message preview. Always navigate to the clicked message instead of parent message.

Fixed the click behavior in `ThreadMessagePreview.tsx` to always navigate to the clicked reply message instead of conditionally navigating to the parent message.
This restores the intuitive behavior where clicking a message takes you to that specific message.

**Root Cause:**
The keyboard navigation feature added in #31549 unified the click handling, causing all clicks on thread message previews with parent message display (`!sequential`) to navigate to the parent message instead of the actual clicked message.

**Solution:**
- Simplified `handleThreadClick` to always use `message._id` (the reply message ID)
- Removed the conditional logic that was causing navigation to parent message based on `sequential` flag

**Changes:**
- Modified `apps/meteor/client/components/message/variants/ThreadMessagePreview.tsx`
- Updated click handler to consistently navigate to the clicked message

## Issue(s)

Closes #32710

## Steps to test or reproduce

1. Create a thread with multiple replies
2. Navigate to the main room view where thread message previews are displayed
3. Click on a thread message preview that shows the parent message (non-sequential messages)
4. **Before fix:** Navigation jumps to the parent message
5. **After fix:** Navigation jumps to the clicked reply message

**Before fix**
32710_before

**After fix**
32710_after

**Test both scenarios:**
- Sequential messages (parent message hidden): Should navigate to reply message ✓
- Non-sequential messages (parent message shown): Should navigate to reply message ✓

## Further comments

This change restores the expected user behavior from v6.6.0 and earlier versions. The modification maintains keyboard navigation functionality while fixing the unintended click behavior. No breaking changes are introduced as this simply corrects a regression in user experience.

The fix ensures consistent behavior regardless of the `sequential` flag state, making the interface more predictable for users.


reetp:
Please fox the outstanding issues above.

.changeset/slimy-actors-report.md
```diff
@@ -0,0 +1,5 @@
+---
+'@rocket.chat/meteor': patch
+---
+
+Fixed thread message preview click behavior to navigate to the clicked reply message instead of unintentionally jumping to the parent message
```
MartinSchoeler:
```suggestion
Fixes thread message preview click behavior, navigate to the clicked reply message instead of unintentionally jumping to the parent message
```

MartinSchoeler (dismissed):
Hey thanks for the contribution!

I think the old behavior of the non-sequential thread preview was mistakenly changed, can you restore the first `ThreadMessageRow` onclick? 

Let's try to restore the old behavior, that each row had it's own onclick.

You can take a look on how the component used to look like on this diff
https://github.com/RocketChat/Rocket.Chat/pull/31549/files#diff-302dadc0469566a97b3b28cfaa3f87208b9565ba5b5e192b99705423e3d95d6cR82

onozaty:
Thank you for working on this.
Is there anything I should do on my side?

PR #36540, opened 2025-07-25
regression: logs export endpoint limited to 100 records

d-gubert:

Workaround the `API_Upper_Count_Limit` setting in the logs export endpoint to increase the max count of records to 2000

## Further comments

[ALT-28]

[ALT-28]: https://rocketchat.atlassian.net/browse/ALT-28?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/meteor/ee/server/apps/communication/endpoints/appLogsExportHandler.ts
+++ b/apps/meteor/ee/server/apps/communication/endpoints/appLogsExportHandler.ts
@@ -4,7 +4,6 @@ import { ajv } from '@rocket.chat/rest-typings/src/v1/Ajv';
 import { parse } from 'cookie';
 import { json2csv } from 'json-2-csv';
 
-import { getPaginationItems } from '../../../../../app/api/server/helpers/getPaginationItems';
 import type { AppsRestApi } from '../rest';
 import { makeAppLogsQuery } from './lib/makeAppLogsQuery';
 import { APIClass } from '../../../../../app/api/server/ApiClass';
@@ -81,7 +80,6 @@ export const registerAppLogsExportHandler = ({ api, _manager, _orch }: AppsRestA
 			}
 
 			const { sort } = await this.parseJsonQuery();
-
 
 			const options = {
 				sort: sort || { _updatedAt: -1 },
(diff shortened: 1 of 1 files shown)


apps/meteor/ee/server/apps/communication/endpoints/appLogsExportHandler.ts
```diff
@@ -74,13 +74,19 @@ export const registerAppLogsExportHandler = ({ api, _manager, _orch }: AppsRestA
 				return api.notFound(`No App found by the id of: ${this.urlParams.id}`);
 			}
 
-			const { count } = await getPaginationItems(this.queryParams);
+			let count = 100;
+
+			if (this.queryParams.count !== undefined && this.queryParams.count !== null) {
+				count = parseInt(String(this.queryParams.count || 100));
```
MartinSchoeler, line 79: Is there a possibility of a non number string be passed as count? If so the Math.min operation will return an NaN as well. Wouldn't make more sense do a 
```suggestion
				count = parseInt(String(this.queryParams.count)) || 100;
```

d-gubert: Good point. I actually copied this from https://github.com/RocketChat/Rocket.Chat/blob/7dd8689188b75db8b2cb74e9d8a1eab8c78e6883/apps/meteor/app/api/server/helpers/getPaginationItems.ts#L20 might be cool to open up a PR to solve it there too :P

d-gubert: Actually, reviewing it a bit more, the validation function for the parameters will not allow a non-numeric value https://github.com/RocketChat/Rocket.Chat/blob/7dd8689188b75db8b2cb74e9d8a1eab8c78e6883/packages/rest-typings/src/apps/appLogsExportProps.ts#L15

So for the endpoint we should be good as it would get a 400 error before ever being handled

MartinSchoeler (approved):

PR #36541, opened 2025-07-25
feat: adds new model function for aggregated agents by department

lucas-a-pelegrino:

This PR expands the agent's information to include its name by using the `.aggregate` function from mongo.
## Issue(s)

[CTZ-220](https://rocketchat.atlassian.net/browse/CTZ-220)

## Further comments

[CTZ-220]: https://rocketchat.atlassian.net/browse/CTZ-220?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/meteor/app/livechat/server/api/lib/departments.ts
```diff
@@ -169,19 +169,18 @@ export async function findDepartmentAgents({
...
 }: FindDepartmentAgentsParams): Promise<PaginatedResult<{ agents: ILivechatDepartmentAgents[] }>> {
-	const { cursor, totalCount } = LivechatDepartmentAgents.findAgentsByDepartmentId<ILivechatDepartmentAgents>(departmentId, {
+	const cursor = LivechatDepartmentAgents.findAgentsByDepartmentIdAggregated(departmentId, {
 		sort: sort || { username: 1 },
 		skip: offset,
 		limit: count,
 	});
-
-	const [agents, total] = await Promise.all([cursor.toArray(), totalCount]);
+	const { result, totalCount } = (await cursor.toArray())[0];
```
sampaiodiego:
```suggestion
	const [{ result, totalCount }] = await cursor.toArray();
```

packages/models/src/models/LivechatDepartmentAgents.ts
```diff
@@ -112,6 +113,68 @@ export class LivechatDepartmentAgentsRaw extends BaseRaw<ILivechatDepartmentAgen
 		return this.findPaginated(query, options);
 	}
 
+	findAgentsByDepartmentIdAggregated(
+		departmentId: string,
+		options?: undefined | FindOptions<ILivechatDepartmentAgents>,
```
sampaiodiego:
```suggestion
		options?: FindOptions<ILivechatDepartmentAgents>,
```

packages/model-typings/src/models/ILivechatDepartmentAgentsModel.ts
```diff
@@ -39,6 +39,11 @@ export interface ILivechatDepartmentAgentsModel extends IBaseModel<ILivechatDepa
 		options?: undefined | FindOptions<ILivechatDepartmentAgents>,
 	): FindPaginated<FindCursor<ILivechatDepartmentAgents>>;
 
+	findAgentsByDepartmentIdAggregated(
+		departmentId: string,
+		options?: undefined | FindOptions<ILivechatDepartmentAgents>,
```
sampaiodiego:
```suggestion
		options?: FindOptions<ILivechatDepartmentAgents>,
```

apps/meteor/app/livechat/server/api/lib/departments.ts
```diff
@@ -169,19 +169,18 @@ export async function findDepartmentAgents({
 	departmentId,
 	pagination: { offset, count, sort },
 }: FindDepartmentAgentsParams): Promise<PaginatedResult<{ agents: ILivechatDepartmentAgents[] }>> {
-	const { cursor, totalCount } = LivechatDepartmentAgents.findAgentsByDepartmentId<ILivechatDepartmentAgents>(departmentId, {
```
KevLehman, line 172: Would u mind checking if after using this new function the old one is still used somewhere? if not, think we can remove it

apps/meteor/app/livechat/server/api/lib/departments.ts
```diff
@@ -169,19 +169,18 @@ export async function findDepartmentAgents({
...
+	const [{ result, totalCount }] = await cursor.toArray();
 
 	return {
-		agents,
-		count: agents.length,
+		agents: result,
+		count: result.length,
 		offset,
-		total,
+		total: totalCount[0]?.total || 0,
```
KevLehman: Will totalCount be always an array? or there's a posibility of it being `null` or `undefined` and thus `totalCount[0]` would error out?

lucas-a-pelegrino: This is the return type for the new `findAgentsByDepartmentId` func:

```typescript
AggregationCursor<{
		result: (ILivechatDepartmentAgents & { user: { _id: string; username: string; name: string } })[];
		totalCount: { _id: null; total: number }[];
	}>
```

It will return an `array`, but the way the code was previously written, it would error out with `total = undefined` when the query matched no departments.

I've fixed this by defaulting the value of total to 0.

apps/meteor/app/livechat/server/api/lib/departments.ts
```diff
@@ -169,17 +169,21 @@ export async function findDepartmentAgents({
 	departmentId,
 	pagination: { offset, count, sort },
 }: FindDepartmentAgentsParams): Promise<PaginatedResult<{ agents: ILivechatDepartmentAgents[] }>> {
```
KevLehman: The result of the function is `ILivechatDepartmentAgents & { user: { _id: string; username: string; name: string } })[]` but the type says it still just returns `ILivechatDepartmentAgents[]` which is not true.

packages/models/src/models/LivechatDepartmentAgents.ts
```diff
@@ -87,29 +88,69 @@ export class LivechatDepartmentAgentsRaw extends BaseRaw<ILivechatDepartmentAgen
...
 
-		return this.findPaginated(query, options);
+		const facet = {
+			$facet: {
+				result: pagination,
+				totalCount: [{ $group: { _id: null, total: { $sum: 1 } } }],
+			},
+		};
+
+		return this.col.aggregate(
```
KevLehman: Add the type param to the aggregation, that helps better with definition & typechecks 👀

KevLehman (changes requested):
Also, didn't we need to change the endpoint type definition to tell that it will return the user now?

PR #36559, opened 2025-07-29
feat: add users.setUsername endpoint

cardoso:


cardoso:
Already covered in #36146

PR #36562, opened 2025-07-29
refactor: query `ParentTeam` data from reusable hooks

dougfabris:


apps/meteor/client/hooks/useTeamInfoEndpoint.ts
```diff
@@ -0,0 +1,17 @@
+import { useEndpoint } from '@rocket.chat/ui-contexts';
+import type { UseQueryOptions } from '@tanstack/react-query';
+import { keepPreviousData, useQuery } from '@tanstack/react-query';
+
+import { teamsQueryKeys } from '../lib/queryKeys';
+
+export const useTeamInfoEndpoint = (teamId: string, retry: UseQueryOptions['retry'] = false) => {
```
tassoevan: What about replacing `retry` with an `options` parameters exposing more `UseQueryOptions` fields? It's a common pattern we use.

dougfabris: Indeed, my bad here. Fixing!

tassoevan (approved):

PR #36576, opened 2025-07-30
test: api test update custom fields

jessicaschelly:

### DIFF
--- a/apps/meteor/tests/end-to-end/api/users.ts
+++ b/apps/meteor/tests/end-to-end/api/users.ts
@@ -2096,14 +2096,10 @@ describe('[Users]', () => {
 					customFieldText1: {
 						type: 'text',
 						required: false,
-						minLength: 2,
-						maxLength: 10,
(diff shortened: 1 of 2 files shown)


apps/meteor/tests/end-to-end/api/users.ts
```diff
@@ -2087,6 +2087,110 @@ describe('[Users]', () => {
...
+					.expect(200);
+
+				const updateResponse = await request
+					.post(api('users.update'))
+					.set(credentials)
+					.send({
+						userId: testUser._id,
+						data: {
+							customFields: {
+								customFieldText2: 'additional2',
```
cardoso, line 2179: `additional2` has 11 characters, it will fail even with the bug fix due to `maxLength: 10`.

jessicaschelly: great catch! removed the length restrictions since it was not mandatory

PR #36578, opened 2025-07-30
fix: users.update API call erases other customFields

cardoso:
## Issue(s)

https://rocketchat.atlassian.net/browse/CORE-1260

### DIFF
--- a/apps/meteor/app/lib/server/functions/saveCustomFieldsWithoutValidation.ts
+++ b/apps/meteor/app/lib/server/functions/saveCustomFieldsWithoutValidation.ts
@@ -32,22 +32,20 @@ export const saveCustomFieldsWithoutValidation = async function (
 	// configured custom fields in setting
 	const customFieldsMeta = getCustomFieldsMeta(customFieldsSetting);
 
-	const customFields: Record<string, any> = Object.keys(customFieldsMeta).reduce(
-		(acc, currentValue) => ({
-			...acc,
-			[currentValue]: formData[currentValue],
-		}),
-		{},
+	const customFields = Object.fromEntries(
+		Object.keys(customFieldsMeta)
+			.filter((key) => formData[key])
+			.map((key) => [key, formData[key]]),
 	);
 
 	const { _updater, session } = options || {};
 
 	const updater = _updater || Users.getUpdater();
-
-	updater.set('customFields', customFields);
-
 	// add modified records to updater
 	Object.keys(customFields).forEach((fieldName) => {
+		// @ts-expect-error `Updater.set` does not support `customFields.${fieldName}` syntax
+		updater.set(`customFields.${fieldName}`, customFields[fieldName]);
+
 		if (!customFieldsMeta[fieldName].modifyRecordField) {
 			return;
 		}


pierre-lehnen-rc (approved):
not sure if this might affect custom fields on DMs (for good or for bad), but for the user itself :+1:

apps/meteor/tests/e2e/admin-users-custom-fields.spec.ts
```diff
@@ -0,0 +1,139 @@
...
+			await expect(poAdmin.tabs.users.getCustomField('customFieldText2')).toBeVisible();
+
+			await poAdmin.tabs.users.getCustomField('customFieldText1').fill(adminCustomFieldValue1);
+			await poAdmin.tabs.users.getCustomField('customFieldText2').fill(adminCustomFieldValue2);
+		});
+
+		await test.step('should save user custom fields', async () => {
+			await poAdmin.tabs.users.btnSaveUser.waitFor();
+			await poAdmin.tabs.users.btnSaveUser.click();
+			await expect(poHomeChannel.toastSuccess).toBeVisible();
```
dougfabris: this type of assertion can cause intermittences, I would suggest not doing this. And they're kind unnecessary since you're doing the proper check below, verifying the new fields values

apps/meteor/tests/e2e/admin-users-custom-fields.spec.ts
```diff
@@ -0,0 +1,139 @@
...
+	test('should allow admin to add user custom fields', async () => {
+		await test.step('should find and click on add test user', async () => {
+			await poAdmin.inputSearchUsers.fill(addTestUser.data.username);
+
+			await expect(poAdmin.getUserRowByUsername(addTestUser.data.username)).toBeVisible();
+			await poAdmin.getUserRowByUsername(addTestUser.data.username).click();
+		});
+
+		await test.step('should navigate to edit user form', async () => {
+			await poAdmin.btnEdit.waitFor();
```
dougfabris: Im not sure if this is necessary since the click already waits for the element

apps/meteor/tests/e2e/admin-users-custom-fields.spec.ts
```diff
@@ -0,0 +1,139 @@
...
+	test('should allow admin to update existing user custom fields', async () => {
+		await test.step('should find and click on update test user', async () => {
+			await poAdmin.inputSearchUsers.fill(updateTestUser.data.username);
+
+			await expect(poAdmin.getUserRowByUsername(updateTestUser.data.username)).toBeVisible();
+			await poAdmin.getUserRowByUsername(updateTestUser.data.username).click();
+		});
+
+		await test.step('should navigate to edit user form', async () => {
+			await poAdmin.btnEdit.waitFor();
```
dougfabris: same comment I did previously

apps/meteor/tests/e2e/admin-users-custom-fields.spec.ts
```diff
@@ -0,0 +1,139 @@
...
+			await expect(poAdmin.tabs.users.getCustomField('customFieldText2')).toHaveValue(adminCustomFieldValue2);
+
+			await poAdmin.tabs.users.getCustomField('customFieldText1').clear();
+			await poAdmin.tabs.users.getCustomField('customFieldText1').fill(adminCustomFieldUpdated1);
+		});
+
+		await test.step('should save and verify partial update', async () => {
+			await poAdmin.tabs.users.btnSaveUser.waitFor();
+			await poAdmin.tabs.users.btnSaveUser.click();
+			await expect(poHomeChannel.toastSuccess).toBeVisible();
```
dougfabris: IMO we should avoid this, unfortunately something the team discourage but it's not written anywhere

dougfabris (changes requested):

dougfabris (approved):

PR #36589, opened 2025-07-31
feat: adds new model func to disable Contacts by id

lucas-a-pelegrino:

### DIFF
--- a/packages/model-typings/src/models/ILivechatContactsModel.ts
+++ b/packages/model-typings/src/models/ILivechatContactsModel.ts
@@ -71,6 +71,7 @@ export interface ILivechatContactsModel extends IBaseModel<ILivechatContact> {
 	getStatistics(): AggregationCursor<{ totalConflicts: number; avgChannelsPerContact: number }>;
 	updateByVisitorId(visitorId: string, update: UpdateFilter<ILivechatContact>, options?: UpdateOptions): Promise<UpdateResult>;
 	disableByVisitorId(visitorId: string): Promise<UpdateResult | Document>;
+	disableByContactId(contactId: string): Promise<UpdateResult | Document>;
 	findOneEnabledById(_id: ILivechatContact['_id'], options?: FindOptions<ILivechatContact>): Promise<ILivechatContact | null>;
 	findOneEnabledById<P extends Document = ILivechatContact>(_id: P['_id'], options?: FindOptions<P>): Promise<P | null>;
 	findOneEnabledById(_id: ILivechatContact['_id'], options?: any): Promise<ILivechatContact | null>;
--- a/packages/models/src/models/LivechatContacts.ts
+++ b/packages/models/src/models/LivechatContacts.ts
@@ -331,6 +331,23 @@ export class LivechatContactsRaw extends BaseRaw<ILivechatContact> implements IL
 		);
 	}
 
+	disableByContactId(contactId: string): Promise<UpdateResult | Document> {
+		return this.updateOne(
+			{ _id: contactId },
+			{
+				$set: { enabled: false },
+				$unset: {
+					emails: 1,
+					customFields: 1,
+					lastChat: 1,
+					channels: 1,
+					name: 1,
+					phones: 1,
+				},
+			},
+		);
+	}
+
 	async addEmail(contactId: string, email: string): Promise<ILivechatContact | null> {
 		const updatedContact = await this.findOneAndUpdate({ _id: contactId }, { $addToSet: { emails: { address: email } } });
 


apps/meteor/app/livechat/server/api/v1/contact.ts
```diff
@@ -131,6 +132,37 @@ API.v1.addRoute(
...
+
+				switch (error.message) {
+					case 'error-contact-not-found':
+						return API.v1.notFound(error.message);
+					case 'error-contact-removal-disabled':
+						return API.v1.unauthorized(error.message);
+					default:
+						return API.v1.failure(error);
+				}
+			}
```
lucas-a-pelegrino: Could use some suggestions on a better approach for this... not liking it.

sampaiodiego: I think this is fine.. the only other option I can think of is doing the checks here on the endpoint instead of on the `disableContactById` function.. but I think it is better doing there, so 👍

apps/meteor/app/livechat/server/api/v1/contact.ts
```diff
@@ -131,6 +132,37 @@ API.v1.addRoute(
 	},
 );
 
+API.v1.addRoute(
```
sampaiodiego: this is deprecated.. you should use `API.v1.post` instead.

apps/meteor/app/livechat/server/api/v1/contact.ts
```diff
@@ -131,6 +132,37 @@ API.v1.addRoute(
...
+	'omnichannel/contacts/:id',
```
sampaiodiego: this is not *wrong* but it is not how we do.. the standard for `v1` is not RESTful, if you look the other endpoints defined on this file you'll see the standard is like `contacts.get`, `contacts.update`, etc.

So I'll ask you to change to `POST` and make the path like `omnichannel/contacts.delete`..

apps/meteor/server/settings/omnichannel.ts
```diff
@@ -454,6 +454,14 @@ export const createOmniSettings = () =>
 			enableQuery: omnichannelEnabledQuery,
 		});
 
+		await this.add('Omnichannel_enable_contact_removal', false, {
```
sampaiodiego: it seems weird to have a setting for this.. I'll ask on the task

packages/models/src/models/LivechatContacts.ts
```diff
@@ -331,6 +331,24 @@ export class LivechatContactsRaw extends BaseRaw<ILivechatContact> implements IL
 		);
 	}
 
+	disableByContactId(contactId: string): Promise<UpdateResult | Document> {
+		return this.updateOne(
+			{ _id: contactId },
+			{
+				$set: { unknown: true },
```
sampaiodiego: just like the `disableByVisitorId` I think this should be actually:
```suggestion
				$set: { enabled: false },
```

apps/meteor/app/livechat/server/lib/contacts/disableContact.ts
```diff
@@ -0,0 +1,16 @@
+import { LivechatContacts } from '@rocket.chat/models';
+
+import { settings } from '../../../../settings/server';
+
+export async function disableContactById(contactId: string): Promise<void> {
+	if (!settings.get('Omnichannel_enable_contact_removal')) {
+		throw new Error('error-contact-removal-disabled');
+	}
+
+	const contact = await LivechatContacts.findOneById(contactId);
```
sampaiodiego: you don't need to fetch the whole contact object just to validate it exists.. instead you can used the return of the `update` you're doing later (when you call `LivechatContacts.disableByContactId`) for that.. please check: https://mongodb.github.io/node-mongodb-native/6.18/interfaces/UpdateResult.html

you can use `matchedCount` for that.

apps/meteor/app/livechat/server/lib/contacts/disableContact.ts
```diff
@@ -0,0 +1,8 @@
+import { LivechatContacts } from '@rocket.chat/models';
```
KevLehman: Just curious: don't we have a contacts lib already (a file) where this code can be?

apps/meteor/app/livechat/server/lib/contacts/disableContact.spec.ts
```diff
@@ -0,0 +1,48 @@
...
+const settingsMock = {
+	get: sinon.stub(),
+};
+
+const { disableContactById } = proxyquire.noCallThru().load('./disableContact', {
+	'@rocket.chat/models': modelsMock,
+	'../../../../settings/server': settingsMock,
+});
+
+describe('disableContact', () => {
```
KevLehman: curious: are these tests running? (asking cause sometimes we don't have the paths added to mocharc file and so it ignores them)

apps/meteor/tests/end-to-end/api/livechat/contacts.ts
```diff
@@ -1073,6 +1073,61 @@ describe('LIVECHAT - contacts', () => {
...
+		});
+
+		it('should return 404 if contact is not found', async () => {
+			const response = await request.post(api(`omnichannel/contacts.delete`)).set(credentials).send({ contactId: 'invalid' });
+
+			expect(response.status).to.be.equal(404);
+			expect(response.body).to.have.property('success', false);
+		});
+
+		it("should return an error if user doesn't have 'delete-livechat-contact' permission", async () => {
```
KevLehman: do this on its own describe so you can use before/after to edit the permissions

apps/meteor/tests/end-to-end/api/livechat/contacts.ts
```diff
@@ -1073,6 +1073,61 @@ describe('LIVECHAT - contacts', () => {
...
+
+		const contact = {
+			name: faker.person.fullName(),
+			emails: [email],
+			phones: [phone],
+			contactManager: agentUser?._id,
+		};
+
+		before(async () => {
+			await updatePermission('delete-livechat-contact', ['admin']);
```
KevLehman: do you think this permission may have been modified before this test? if not, think we can remove this and use the "default" value which should include admin.

apps/meteor/tests/end-to-end/api/livechat/contacts.ts
```diff
@@ -1073,6 +1073,61 @@ describe('LIVECHAT - contacts', () => {
...
+			const { body } = await request
+				.post(api('omnichannel/contacts'))
+				.set(credentials)
+				.send({ ...contact });
+			contactId = body.contactId;
+
+			const visitor = await createVisitor(undefined, contact.name, email, phone);
+
+			await createLivechatRoom(visitor.token);
```
KevLehman: Close this room on the `after` hook

apps/meteor/tests/end-to-end/api/livechat/contacts.ts
lucas-a-pelegrino: What would be the benefit for that approach? Genuinely curious.

apps/meteor/tests/end-to-end/api/livechat/contacts.ts
lucas-a-pelegrino: Not to my knowledge, if the default is true for admin users, we can safely remove it. I just added the helper for assurance.

apps/meteor/app/api/server/definition.ts
```diff
@@ -71,11 +71,13 @@ export type UnavailableResult<T, StatusCode = 503> = {
 	};
 };
 
-export type NotFoundResult<T = string> = {
+export type NotFoundResult = {
```
ggazzo: open a new pr to "fix" this if its really necessary

apps/meteor/app/api/server/definition.ts
```diff
@@ -71,11 +71,13 @@ export type UnavailableResult<T, StatusCode = 503> = {
...
 	statusCode: 404;
 	body: {
 		success: false;
-		error: T;
+		error?: string;
+		errorType?: string;
+		message?: string;
```
ggazzo: if all the props are not required, it It may happen that we are not required to pass anything, this can cause errors

apps/meteor/app/livechat/server/api/v1/contact.ts
```diff
@@ -224,3 +232,40 @@ API.v1.addRoute(
...
+
+		try {
+			await disableContactById(contactId);
+			return API.v1.success();
+		} catch (error) {
+			if (!(error instanceof Error) || error.message !== 'error-contact-not-found') {
+				return API.v1.failure('error-invalid-contact');
+			}
+
+			return API.v1.notFound(error.message);
```
sampaiodiego: I'm afraid we cannot return a 404 here.. 404 means the path/endpoint doesn't exists, which is not the case.. this only works if the `_id` of the contact was part of the URL for example.

so in our case, since the contact id is sent via payload, let's return 400 if we it was not found.

.changeset/five-carpets-perform.md
```diff
@@ -0,0 +1,9 @@
+---
+"@rocket.chat/meteor": minor
+"@rocket.chat/i18n": minor
+"@rocket.chat/model-typings": minor
+"@rocket.chat/models": minor
+"@rocket.chat/rest-typings": minor
+---
+
+Adds new endpoint to disable Livechat Contacts by its id, with a new permission `delete-livechat-contact`, and setting `Omnichannel_enable_contact_removal`.
```
sampaiodiego: we dont have the setting anymore

apps/meteor/tests/unit/app/livechat/server/lib/disableContact.spec.ts
```diff
@@ -0,0 +1,44 @@
...
+});
+
+describe('disableContact', () => {
+	beforeEach(() => {
+		settingsMock.get.reset();
+		modelsMock.LivechatContacts.disableByContactId.reset();
+	});
+
+	it('should disable the contact', async () => {
+		settingsMock.get.withArgs('Omnichannel_enable_contact_removal').returns(true);
```
sampaiodiego: we don't have the setting anymore

packages/rest-typings/src/v1/omnichannel.ts
```diff
@@ -1389,6 +1389,35 @@ const POSTUpdateOmnichannelContactsSchema = {
...
+	additionalProperties: false,
+};
+
+export const isPOSTOmnichannelContactDeleteProps = ajv.compile<POSTOmnichannelContactDeleteProps>(POSTOmnichannelContactDeleteSchema);
+
+const POSTOmnichannelContactDeleteSuccess = {
+	type: 'object',
+	properties: {
+		success: {
+			type: 'boolean',
```
sampaiodiego, line 1413: this is a good addition:
```suggestion
			type: 'boolean',
			enum: [true],
```

at some point I think we'll need to come up with a "standard success schema" that can be re-used everywhere.. having to define this for each endpoint seems too much :/

apps/meteor/tests/end-to-end/api/livechat/contacts.ts
```diff
@@ -1073,6 +1073,61 @@ describe('LIVECHAT - contacts', () => {
 		});
 	});
 
+	describe('[DELETE] omnichannel/contacts.delete', () => {
```
sampaiodiego:
```suggestion
	describe('[POST] omnichannel/contacts.delete', () => {
```

packages/models/src/models/LivechatContacts.ts
```diff
@@ -331,6 +331,24 @@ export class LivechatContactsRaw extends BaseRaw<ILivechatContact> implements IL
 		);
 	}
 
+	disableByContactId(contactId: string): Promise<UpdateResult | Document> {
```
sampaiodiego: I understand you get this return type from `disableByVisitorId`, but this is not correct.. ref: https://mongodb.github.io/node-mongodb-native/6.18/classes/Collection.html#updateOne
```suggestion
	disableByContactId(contactId: string): Promise<UpdateResult> {
```

apps/meteor/app/livechat/server/api/v1/contact.ts
lucas-a-pelegrino: Actually, the `404 Not Found` means a **resource** could not be found. It may be the URL is not valid, but it the case of an API it means the resource, in this case; the contact, is non-existent.

So I think the `404` still applies in this case.

packages/rest-typings/src/v1/omnichannel.ts
lucas-a-pelegrino: Definitely agree with you on that. We might want to create some sort of `SuccessResult<T>` or something similar.

apps/meteor/tests/end-to-end/api/livechat/contacts.ts
KevLehman: so if it fails (when running locally mostly) it will always come back to the original state instead of leaving the permissions messed up 👀

apps/meteor/app/livechat/server/lib/contacts/disableContact.ts
```diff
@@ -0,0 +1,8 @@
...
+
+export async function disableContactById(contactId: string): Promise<void> {
```
KevLehman, line 7: You need to also delete/disable the associated visitors 👀 so it's GDPR compliant.

You can see at something similar on the endpoint that deletes visitors.

packages/model-typings/src/models/ILivechatContactsModel.ts
```diff
@@ -71,6 +71,7 @@ export interface ILivechatContactsModel extends IBaseModel<ILivechatContact> {
 	getStatistics(): AggregationCursor<{ totalConflicts: number; avgChannelsPerContact: number }>;
 	updateByVisitorId(visitorId: string, update: UpdateFilter<ILivechatContact>, options?: UpdateOptions): Promise<UpdateResult>;
 	disableByVisitorId(visitorId: string): Promise<UpdateResult | Document>;
+	disableByContactId(contactId: string): Promise<UpdateResult | Document>;
```
KevLehman:
```suggestion
	disableByContactId(contactId: string): Promise<UpdateResult>;
```

apps/meteor/app/livechat/server/lib/contacts/disableContact.ts
```diff
@@ -0,0 +1,15 @@
+import type { ILivechatContact } from '@rocket.chat/core-typings';
+import { LivechatContacts, LivechatVisitors } from '@rocket.chat/models';
+
+export async function disableContactById(contactId: string): Promise<void> {
```
KevLehman, line 7: We have this behavior on the visitor removal.
```
// if gdpr is enabled, bypass rooms check
		if (rooms?.length && !settings.get('Livechat_Allow_collect_and_store_HTTP_header_informations')) {
			throw new Meteor.Error('visitor-has-open-rooms', 'Cannot remove visitors with opened rooms');
		}

```

Would you mind checking with Renato if this is something we have to do here? Check if the contact has any open rooms before disabling?

apps/meteor/app/livechat/server/lib/contacts/disableContact.ts
```diff
@@ -0,0 +1,23 @@
...
+import { removeGuest } from '../guests';
+
+export async function disableContactById(contactId: string): Promise<void> {
+	const contact = await LivechatContacts.findOneEnabledById<Pick<ILivechatContact, '_id' | 'channels'>>(contactId);
+	if (!contact) {
+		throw new Error('error-contact-not-found');
+	}
+
+	// Checking if the contact has any open channel/room before removing its data.
+	const contactOpenRooms = await LivechatRooms.findOpenByContactId(contactId).toArray();
```
KevLehman: You're not using the rooms for other things. You can create a new method to just count the rooms 👀

sampaiodiego: I'd prefer using a `findOne` with `{ projection: { _id: 1 } }` though, since you don't need the total count of rooms, you just need to know if there are any

KevLehman: That's better 👀 thx

apps/meteor/app/livechat/server/api/v1/contact.ts
```diff
@@ -224,3 +231,46 @@ API.v1.addRoute(
...
+			}
+
+			switch (error.message) {
+				case 'error-contact-not-found':
+					return API.v1.failure(error.message);
+				case 'error-contact-has-open-rooms':
+					return API.v1.failure(error.message);
+			}
+
+			return API.v1.failure(error.message);
```
KevLehman: if on all cases you're returning error.message, why do you need the switch? 🙈

apps/meteor/app/livechat/server/lib/contacts/disableContact.ts
```diff
@@ -0,0 +1,23 @@
...
+import { removeGuest } from '../guests';
+
+export async function disableContactById(contactId: string): Promise<void> {
+	const contact = await LivechatContacts.findOneEnabledById<Pick<ILivechatContact, '_id' | 'channels'>>(contactId);
+	if (!contact) {
+		throw new Error('error-contact-not-found');
+	}
+
+	// Checking if the contact has any open channel/room before removing its data.
+	const contactOpenRooms = await LivechatRooms.findOne({ contactId, open: true }, { projection: { _id: 1 } });
```
KevLehman: would u mind moving this to a model func? 🙈 something like `checkContactOpenRooms` or so

apps/meteor/app/livechat/server/api/v1/contact.ts
lucas-a-pelegrino: That makes sense, it's probably an oversight from when the API was sending back different status codes.

PR #36592, opened 2025-07-31
fix: create-c and create-p not being applied to Team Creation

MartinSchoeler:

When the "Create Private/Public Channels" permission is turned off, the team creation UI still allows selecting a private team. However, attempting to create such a team results in an API error ("error-team-creation"). This is inconsistent with the channel creation UI, which disables the public/private toggle when the corresponding permission is off.

Since the Team has a main "channel", the user needs both create-team and create-c or create-p to be able to create a team

## Issue(s)

[SUP-747](https://rocketchat.atlassian.net/browse/SUP-747)

## Steps to test or reproduce

Disable the "Create Private Channels" permission for the user role.

Attempt to create a new team with the payload:

For public team: {"name": "newest_team8", "type": 0} → Succeeds

For private team: {"name": "newest_team9", "type": 1} → Fails with "error-team-creation"


dougfabris (changes requested):
But since its the same rule we have in CreateChannelModal, can we create a shared hook to handle the permission for them?

dougfabris (approved):
thankyoumartinho

PR #36597, opened 2025-08-01
test: Check User Registration Status and Modify Admin Rights

Harmeet221:
JIRA TASK: https://rocketchat.atlassian.net/browse/QA-89

This PR aims to validate various admin functionalities and improve test structure:
- Refactored existing code to use defined locators instead of hardcoding them directly.
- Create a new user via API.
- Logged in as the newly created user.
- Verified that the user's registration status is marked as "Active" after the initial login.
- Assigned the admin role to the user.
- Revoked the admin role from the user.


apps/meteor/tests/e2e/admin-users.spec.ts
```diff
@@ -2,34 +2,39 @@ import { faker } from '@faker-js/faker';
...
 import { Users } from './fixtures/userStates';
-import { expect, test } from './utils/test';
+import { Admin, Registration, Utils } from './page-objects';
+import { test, expect } from './utils/test';
+
+let user: IUser & { username: string };
+let password: string;
+let admin: Admin;
+
+test.beforeAll('Create a new user', async ({ api }) => {
```
jessicaschelly: You can use the interface `ITestUser`

example:
```
user: ITestUser
user = await createTestUser(api, { username: 'username-you-want' });

File: apps/meteor/tests/e2e/admin-users.spec.ts [hunk 2]
```diff
@@ -41,48 +46,148 @@ test.describe('Admin > Users', () => {
...
+			await poRegistration.username.fill(user.username);
+			await poRegistration.inputPassword.fill(password);
+			await poRegistration.btnLogin.click();
+		});
+
+		await test.step('Assert user is logged in', async () => {
+			await expect(poUtils.mainContent).toBeVisible();
+		});
+
+		await test.step('close browser', async () => {
```
jessicaschelly: Could this be on a before() call?

apps/meteor/tests/e2e/admin-users.spec.ts
```diff
@@ -41,48 +46,148 @@ test.describe('Admin > Users', () => {
...
+			await context.close();
+		});
+	});
+});
+
+test.describe('Login as a admin and verify registration status', () => {
```
jessicaschelly: `test.describe('User Status Management', () => {`

apps/meteor/tests/e2e/admin-users.spec.ts
```diff
@@ -41,48 +46,148 @@ test.describe('Admin > Users', () => {
...
+				await admin.tabs.users.menuItemActivate.click();
+				await expect(admin.getUserRowByUsername(username)).not.toBeVisible();
+				await admin.tabs.users.tabPending.click();
+				await expect(admin.getUserRowByUsername(username)).toBeVisible();
 			});
 		},
 	);
 });
+
+test.describe('Login', () => {
```
jessicaschelly: the describe below is dependent on this test, right? If it runs in parallel, it could be an issue

apps/meteor/tests/e2e/admin-users.spec.ts
```diff
@@ -41,48 +46,148 @@ test.describe('Admin > Users', () => {
...
+			await expect(admin.getUserRowByUsername(username)).toBeVisible();
+		});
+
+		await test.step('is not visible in the Deactivated tab', async () => {
+			await admin.tabs.users.tabDeactivated.click();
+			await expect(admin.getUserRowByUsername(username)).not.toBeVisible();
+		});
+	});
+
+	test('Make a newly created user as admin', async () => {
```
jessicaschelly: This could be another describe, e.g: `Admin Role Management`

apps/meteor/tests/e2e/admin-users.spec.ts
```diff
@@ -2,34 +2,39 @@ import { faker } from '@faker-js/faker';
...
+		username: faker.internet.userName(),
+	});
+	expect(response.status()).toBe(200);
+	const json = await response.json();
+	user = json.user;
+});
 
-test.use({ storageState: Users.admin.state });
+test.afterAll('Delete the new user', async ({ api }) => {
+	const response = await api.post('/users.delete', { userId: user._id });
```
jessicaschelly: If you use the interface, it could be
`user.delete()`

apps/meteor/tests/e2e/admin-users.spec.ts
```diff
@@ -41,48 +46,148 @@ test.describe('Admin > Users', () => {
...
+		});
+
+		await test.step('verify user is admin', async () => {
+			await admin.tabs.users.getUserRowByUsername(username).click();
+			await expect(admin.tabs.users.openDialog).toBeVisible();
+			await expect(admin.tabs.users.openDialog).toContainText('Admin');
+		});
+	});
+
+	test('Remove role as admin', async () => {
```
jessicaschelly: same dependency here, it requires the previous test to run and be successful

jessicaschelly (changes requested):

apps/meteor/tests/e2e/admin-users.spec.ts
Harmeet221: You mean afterAll();

apps/meteor/tests/e2e/admin-users.spec.ts
```diff
@@ -1,88 +1,184 @@
...
+				await admin.tabs.users.tabDeactivated.click();
+				await admin.tabs.users.btnMoreActionsMenu.click();
+				await admin.tabs.users.menuItemActivate.click();
+				await expect(admin.getUserRowByUsername(user.data.username)).not.toBeVisible();
+				await admin.tabs.users.tabPending.click();
+				await expect(admin.getUserRowByUsername(user.data.username)).toBeVisible();
+			});
+		});
+	});
+	test.describe.serial('Login', () => {
```
jessicaschelly: .serial is not needed here, it will be serial by default

apps/meteor/tests/e2e/admin-users.spec.ts
```diff
@@ -1,88 +1,184 @@
...
+
+		test.beforeEach('Go to /admin/users', async ({ page }) => {
+			admin = new Admin(page);
+			await page.goto('/admin/users');
+		});
+
+		test('After the first login, the user gets listed under the Active tab', async () => {
+			await admin.tabs.users.inputSearch.fill(user.data.username);
+
+			await test.step('is visible in the All tab', async () => {
```
jessicaschelly: maybe we could use the pattern of "should"
should be visible in the All tab
should not..

apps/meteor/tests/e2e/admin-users.spec.ts
```diff
@@ -1,88 +1,184 @@
...
+			});
+
+			await test.step('verify user is admin', async () => {
+				await admin.tabs.users.getUserRowByUsername(user.data.username).click();
+				await expect(admin.tabs.users.openDialog).toBeVisible();
+				await expect(admin.tabs.users.openDialog).toContainText('Admin');
+			});
+		});
+
+		test('Remove role as admin', async () => {
```
jessicaschelly: Will this be dependent on the previous test? Maybe you could have a separate user who is already admin to remove it

jessicaschelly (changes requested):

jessicaschelly (dismissed):

apps/meteor/tests/e2e/page-objects/fragments/admin-flextab-users.ts
```diff
@@ -71,6 +71,62 @@ export class AdminFlextabUsers {
...
+		return this.page.getByRole('menuitem', { name: 'Remove Admin' });
+	}
+
+	get openDialog(): Locator {
+		return this.page.getByRole('dialog');
+	}
+
+	get toastMessage(): Locator {
+		return this.page.getByRole('alert');
+	}
```
aleksandernsilva: We have a fragment for the toast messages [here](https://github.com/RocketChat/Rocket.Chat/blob/develop/apps/meteor/tests/e2e/page-objects/fragments/toast-messages.ts), you can reuse it. Let me know if you have any questions.

apps/meteor/tests/e2e/page-objects/fragments/admin-flextab-users.ts
```diff
@@ -71,6 +71,62 @@ export class AdminFlextabUsers {
...
+		return this.page.getByRole('menuitem', { name: 'Make Admin' });
+	}
+
+	get menuItemRemoveAdmin(): Locator {
+		return this.page.getByRole('menuitem', { name: 'Remove Admin' });
+	}
+
+	get openDialog(): Locator {
+		return this.page.getByRole('dialog');
+	}
```
aleksandernsilva: This locator seem pretty broad and the name imo implies an action "openDialog". Can we make it more specific to the dialog you are interested?

apps/meteor/tests/e2e/admin-users.spec.ts
```diff
@@ -1,88 +1,186 @@
...
 
-test.use({ storageState: Users.admin.state });
-
-test.describe('Admin > Users', () => {
-	let user: IUser & { username: string };
+let user: ITestUser;
+let userWithAdminAccess: ITestUser;
+let admin: Admin;
 
+test.describe.serial('Admin > Users', () => {
```
aleksandernsilva: I'm noticing a lot of interdependence between these tests (serial). I believe there's an opportunity here to break these test cases into multiple files and make them independent from one another. They can still be associated by name, but they don't need to be in the same file. Their setup and teardown are not huge and it would help us isolate what exactly is not working in case of a failures.

aleksandernsilva (approved):

PR #36600, opened 2025-08-01
test: add best practices to README

juliajforesti:

### DIFF
--- a/apps/meteor/tests/e2e/README.md
+++ b/apps/meteor/tests/e2e/README.md
@@ -92,6 +92,10 @@ page.locator('#modal-root .rcx-button-group--align-end .rcx-button--primary');
 page.getByRole('dialog', name: 'Modal name example').getByRole('button', { name: 'Confirm', exact: true });
 ```
 
+### 3. By position:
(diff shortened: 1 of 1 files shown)


apps/meteor/tests/e2e/README.md
```diff
@@ -35,3 +35,142 @@ Checking if a element is visible
...
+```
+## Use our page-objects:
+> apps/meteor/tests/e2e/page-objects
+- page-objects are a great way to reuse locators across tests using `getters` and `methods`.
+- They make it easier to write tests that are more readable and maintainable.
+- Always make sure to use the most restricted scope possible - not the whole page to avoid multiple matches.
+
+If you are writing a new test, make sure to look at the existing page-objects to see if there is a suitable one for your use case.
+
+If not, create the ones you need in the respective context.
```
MartinSchoeler: I think not all locators are created equal. Some cases are so specific that will never be reutilized, I'm not sure we should have a "hard" best practice for page object getters creation

tassoevan (approved):

Issue #36560, opened 2025-07-29
User avatar doesn't update in real-time


Khizarshah01:
### Description:

When a user updates their avatar, the change is not reflected consistently across the UI. The avatar in the chat feed remains the old version, even for new messages, indicating a severe client-side caching issue caused by a failed real-time update.

### Steps to reproduce:

1. Navigate to Account -> Profile.
2. Upload a new user avatar and save the changes.
3. Observe that the avatar changes correctly on the profile page.
4. Switch to a different rocket.chat tab and then switch back to the profile tab. The avatar in the top-left corner reverts to the old one.
5. Perform a full page refresh `(Ctrl+R or Cmd+R)`. The avatar now appears correctly in most of the UI.
6. Navigate to a channel or DM where you have sent messages. The avatar next to your existing messages in the chat feed is still the old one.
7. Send a new message. The new message also renders with the old avatar.

### Expected behavior:
After a user successfully updates their avatar, the new avatar should be displayed immediately and consistently across all parts of the application without requiring a page refresh.

### Actual behavior:
The user avatar update is inconsistent. It requires a hard refresh to appear in most places, but it remains stale (stuck on the old version) indefinitely within the channel and DM

https://github.com/user-attachments/assets/12f9754a-b116-47a8-9e63-e01a0d00ddbb

### Server Setup Information:

- Version of Rocket.Chat Server:  7.9.0-develop
- License Type:
- Number of Users: 1
- Operating System: Ubuntu (linux)
- Deployment Method: 
- Number of Running Instances: 
- DB Replicaset Oplog:  Enabled
- NodeJS Version:  22.16.0
- MongoDB Version: 7.0.16

### Client Setup Information

- Desktop App or Browser Version:
- Operating System:

### Additional context
Proof of Caching Issue:

Inspecting the image URLs confirms this analysis. The correctly updated avatar uses a URL with a new etag parameter, while the stale avatar in the chat feed is rendered using the base URL without the etag, causing the browser to serve the old, cached version.

Stale URL (in chat feed): `http://localhost:3000/avatar/khizarshah01`

Correct URL (in profile): `http://localhost:3000/avatar/khizarshah01?etag=Me9zRvq2qQmMhx`

Suspected Cause: The server-side process successfully saves the new avatar but fails to trigger the client-facing `updateAvatar` real-time event, leaving the client's cache stale.

Naman1223:
I would like to work on this issuse. Please assing me this

[...]

PR #36605, opened 2025-08-03
Fix: user avatar now updates everywhere in the UI with simple refresh

SAYAN02-DEV:

Fixes avatar caching issue where user avatars don't update consistently across the UI after profile picture changes. Previously, old messages and some UI elements would continue showing stale avatars until a hard refresh.

### Root Cause
- Avatar URLs lacked cache-busting parameters (`/avatar/username` vs `/avatar/username?etag=avatarETag`)
- Message components weren't using latest `avatarETag` from user data
- Browser caching prevented new avatar images from loading

### Solution
- Enhanced message components to lookup latest `avatarETag` from `Users` collection
- Construct cache-busting URLs: `/avatar/username?etag=avatarETag`
- Leverage existing real-time `user.avatarUpdate` events

### Files Modified
- `apps/meteor/client/components/message/variants/RoomMessage.tsx`
- `apps/meteor/client/components/message/variants/ThreadMessage.tsx`
- `apps/meteor/client/components/message/variants/SystemMessage.tsx`

## Issue(s)
User avatar doesn't update in real-time #36560

## Before Fix:

https://github.com/user-attachments/assets/51a3bee3-c2dd-4ac8-94a8-7f21d8e751d8

## After Fix

https://github.com/user-attachments/assets/b8647f5d-93e7-4cac-a80c-4c45e874a706

## Further comments

Minimal fix that ensures all avatar URLs include cache-busting `etag` parameter, forcing browsers to load fresh images. Maintains existing functionality and performance.

### DIFF
--- a/apps/meteor/client/components/message/variants/RoomMessage.tsx
+++ b/apps/meteor/client/components/message/variants/RoomMessage.tsx
@@ -22,6 +22,7 @@ import MessageHeader from '../MessageHeader';
 import MessageToolbarHolder from '../MessageToolbarHolder';
 import StatusIndicators from '../StatusIndicators';
 import RoomMessageContent from './room/RoomMessageContent';
+import { Users } from '../../../stores/Users';
 
 type RoomMessageProps = {
 	message: IMessage & { ignored?: boolean };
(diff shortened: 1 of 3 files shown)


Khizarshah01:
@SAYAN02-DEV  The main goal is to get the new `etag` to the client in real-time, so the avatar updates instantly without needing a manual refresh.
It looks like this PR still requires a refresh for the update to appear.

PR #36617, opened 2025-08-04
feat: Send to queue while supporting default agent

KevLehman:
## Issue(s)

https://rocketchat.atlassian.net/browse/CTZ-293


apps/meteor/app/livechat/server/lib/RoutingManager.ts
```diff
@@ -48,7 +48,12 @@ type Routing = {
 		options?: { clientAction?: boolean; forwardingToDepartment?: { oldDepartmentId?: string; transferData?: any } },
 		room?: IOmnichannelRoom,
 	): Promise<(IOmnichannelRoom & { chatQueued?: boolean }) | null | void>;
-	unassignAgent(inquiry: ILivechatInquiryRecord, departmentId?: string, shouldQueue?: boolean): Promise<boolean>;
+	unassignAgent(
+		inquiry: ILivechatInquiryRecord,
+		departmentId?: string,
+		shouldQueue?: boolean,
+		agent?: SelectedAgent | null,
```
sampaiodiego, line 55: is this enough? not sure I get it why could pass `null` if it was already optional
```suggestion
		agent?: SelectedAgent,
```

KevLehman: This is a long story 🙈 but it's something we never fixed. Routing algos can return `SelectedAgent | null`... but also the `agent` variables could be undefined as it's optional in most places...

...

sampaiodiego: can we map this out so we can fix it soon?

KevLehman: https://rocketchat.atlassian.net/browse/CTZ-303

Right here

apps/meteor/app/livechat/server/lib/RoutingManager.ts
```diff
@@ -157,11 +162,19 @@ export const RoutingManager: Routing = {
...
 		const room = await LivechatRooms.findOneById(rid);
 
-		logger.debug(`Removing assignations of inquiry ${inquiry._id}`);
+		logger.debug({
+			msg: 'Removing assignations of inquiry',
+			inquiryId: inquiry._id,
+			departmentId,
+			room: { _id: room?._id, open: room?.open, servedBy: room?.servedBy },
+			shouldQueue,
+			...(defaultAgent && { defaultAgent }),
```
sampaiodiego:
```suggestion
			defaultAgent,
```

sampaiodiego (changes requested):

sampaiodiego (approved):

PR #36622, opened 2025-08-04
fix: media playback failing due to expired urls

abhinavkrin:
### Background
Audio/video files served from S3 use presigned URLs that expire after a configurable timespan. When a user leaves a room open past that window, the media element’s subsequent `Range` requests hit a stale URL, leading to silent playback failures.

---

### Changes
1. **`useReloadOnError` hook**  
   - Listens to `play`, `stalled`, and `error` events on media elements.  
   - Tracks `expiresAt` timestamp returned by the uploads endpoint.  
   - On trigger, checks if the URL has expired before attempting recovery.  
   - If expired, fetches fresh `redirectUrl` and `expires` via an `?info=true` query to the existing `/uploads/:fileId` endpoint.  
   - Swaps the media’s `src`, calls `.load()`, seeks to last known `currentTime`, and resumes playback.

2. **Uploads endpoint enhancement**  
   - Supports an `info=true` query param.  
   - When present, returns JSON `{ redirectUrl, expires }` instead of streaming or redirecting.  
   - No changes to existing redirect or proxy behavior.

3. **UI fallback styling**  
   - On recovery for video, preserves the video dimensions by copying element dimensions to have minimal UI lag.

4. **These were the solutions considered.**  

   | Strategy                 | Chosen | Rationale                                                              |
   |--------------------------|:------:|------------------------------------------------------------------------|
   | Reactive error events    |   ❌    | Too delayed; Safari compatibility issues.                              |
   | Watchdog timer to detect stalls         |   ❌    | False positives on slow networks.                                      |
   | HEAD validation  on src url on stalls       |   ❌    | Fails without CORS headers.                                            |
   | Media Source Extensions  |   ❌    | Overly complex; CORS/storage overhead.                                 |
   | Service Worker proxy     |   ❌    | HTTPS/PWA constraints; added complexity.                               |
   | **Expiry-aware recovery**| **✅** | Minimal complexity; leverages existing metadata; no CORS constraints.      |

## Steps to test or reproduce

1. Set the upload provider to amazon s3 bucket and configure it. 
2. Set the url expiry time to a lower value (10s) under the s3 configuration.
3. Send an audio/video message to a channel and wait for more than 10 seconds.
4. Play the media and check if the playback recovers after stalling due to url expiry.
## Further comments

[SUP-771](https://rocketchat.atlassian.net/browse/SUP-771)

### DIFF
--- a/apps/meteor/app/file-upload/server/lib/FileUpload.ts
+++ b/apps/meteor/app/file-upload/server/lib/FileUpload.ts
@@ -589,6 +589,27 @@ export const FileUpload = {
 		res.end();
 	},
 
+	respondWithRedirectUrlInfo(
+		redirectUrl: string | false,
+		file: IUpload,
+		_req: http.IncomingMessage,
+		res: http.ServerResponse,
+		expiresInSeconds?: number | null,
+	) {
+		res.setHeader('Content-Type', 'application/json');
+		res.setHeader('Cache-Control', 'no-cache, no-store, must-revalidate');
+		res.writeHead(200);
+		res.end(
+			JSON.stringify({
+				redirectUrl,
+				name: file.name,
+				type: file.type,
+				size: file.size,
+				...(expiresInSeconds && { expires: Date.now() / 1000 + expiresInSeconds }),
+			}),
+		);
+	},
+
 	proxyFile(
 		fileName: string,
 		fileUrl: string,
--- a/apps/meteor/app/file-upload/server/lib/requests.ts
+++ b/apps/meteor/app/file-upload/server/lib/requests.ts
@@ -1,7 +1,21 @@
+import type { IncomingMessage } from 'http';
+
 import { Uploads } from '@rocket.chat/models';
 import { WebApp } from 'meteor/webapp';
 
 import { FileUpload } from './FileUpload';
+
+const hasInfoParam = (req: IncomingMessage) => {
+	if (!req.url) {
+		return false;
+	}
+	const [, params] = req.url.split('?');
+	if (!params) {
+		return false;
+	}
+	const searchParams = new URLSearchParams(params);
+	return searchParams.has('info');
+};
 
 WebApp.connectHandlers.use(FileUpload.getPath(), async (req, res, next) => {
 	const match = /^\/([^\/]+)\/(.*)/.exec(req.url || '');
@@ -14,6 +28,23 @@ WebApp.connectHandlers.use(FileUpload.getPath(), async (req, res, next) => {
 				res.writeHead(403);
 				res.end();
 				return;
+			}
+
+			if (hasInfoParam(req)) {
+				if (!file.store) {
+					res.writeHead(404);
+					res.end();
+					return;
+				}
+				const store = FileUpload.getStoreByName(file.store);
+				let url: string | false;
+				try {
+					url = await store.getStore().getRedirectURL(file, false);
+				} catch {
+					url = false;
+				}
+				const expiryTimespan = await store.getStore().getUrlExpiryTimeSpan();
+				return FileUpload.respondWithRedirectUrlInfo(url, file, req, res, expiryTimespan);
 			}
 
 			res.setHeader('Content-Security-Policy', "default-src 'none'");
--- a/apps/meteor/app/file-upload/server/lib/requests.ts
+++ b/apps/meteor/app/file-upload/server/lib/requests.ts
@@ -14,7 +14,8 @@ const hasInfoParam = (req: IncomingMessage) => {
 		return false;
 	}
 	const searchParams = new URLSearchParams(params);
-	return searchParams.has('info');
+	const infoParam = searchParams.get('info');
+	return infoParam === 'true' || infoParam === '1';
 };
 
 WebApp.connectHandlers.use(FileUpload.getPath(), async (req, res, next) => {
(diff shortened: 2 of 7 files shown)


apps/meteor/client/components/message/content/attachments/file/hooks/useReloadOnError.tsx
```diff
@@ -0,0 +1,144 @@
...
+					return;
+				}
+
+				events.forEach((event) => {
+					node.addEventListener(event, handleMediaURLRecovery);
+				});
+				return () => {
+					if (!node) {
+						return;
+					}
```
gabriellsh: I don't think you need to check if node exists again in the cleanup. You already asserted it's existance before, so this function will hold a reference to it until it (the function) doesn't exist anymore.

apps/meteor/app/file-upload/server/lib/requests.ts
```diff
@@ -1,7 +1,23 @@
...
+		return false;
+	}
+	const [, params] = req.url.split('?');
+	if (!params) {
+		return false;
+	}
+	const searchParams = new URLSearchParams(params);
+	const replyWithRedirectUrl = searchParams.get('replyWithRedirectUrl');
+	return replyWithRedirectUrl === 'true' || replyWithRedirectUrl === '1';
+};
```
gabriellsh, line 20: I recommend using the `URL` object from node, instead of parsing the URL (partially) manually. If you do `new URL(req.url)`, the `URLSearchParams` object will already be available inside the URL obj instance. It's also a safer way to parse the URL correctly, and one less possible point of failure.

apps/meteor/client/components/message/content/attachments/file/hooks/useReloadOnError.tsx
```diff
@@ -0,0 +1,144 @@
...
+		};
+		Object.assign(vidEl.style, {
+			width: computed.width,
+			height: computed.height,
+		});
+
+		return () => {
+			Object.assign(vidEl.style, videoTempStyles);
+		};
+	}, []);
```
gabriellsh: Both these functions can be defined outside the hook, since they do not reference anything else, and have an empty dependency array.

apps/meteor/client/components/message/content/attachments/file/hooks/useReloadOnError.tsx
```diff
@@ -0,0 +1,144 @@
...
+	const isRecovering = useRef(false);
+	const firstRecoveryAttempted = useRef(false);
+
+	const getRedirectURLInfo = useCallback(async (url: string): Promise<{ redirectUrl: string | false; expires: number | null }> => {
+		const [path, query] = url.split('?');
+		const params = new URLSearchParams(query);
+		params.set('replyWithRedirectUrl', 'true');
+		const response = await fetch(`${path}?${params.toString()}`, {
+			credentials: 'same-origin',
+		});
```
gabriellsh: Here too we should use the native `URL` API, instead of manually parsing/building the URL.

apps/meteor/app/file-upload/server/lib/FileUpload.ts
```diff
@@ -589,6 +589,27 @@ export const FileUpload = {
...
+		res.setHeader('Content-Type', 'application/json');
+		res.setHeader('Cache-Control', 'no-cache, no-store, must-revalidate');
+		res.writeHead(200);
+		res.end(
+			JSON.stringify({
+				redirectUrl,
+				name: file.name,
+				type: file.type,
+				size: file.size,
+				...(expiresInSeconds && { expires: Date.now() / 1000 + expiresInSeconds }),
```
gabriellsh: Why send this number? I noticed that in the client part you're getting this number and multiplying it by 1000 again... Is there a reason for this?

I think it would make more sense for this to be an ISO string.

apps/meteor/app/file-upload/server/lib/requests.ts
abhinavkrin: The url received in this case was just a path, so when I created using new URL(), I got invalid url error. That is why I had to switch to this way.

apps/meteor/app/file-upload/server/lib/FileUpload.ts
abhinavkrin: Yeah, ISO string would be better

apps/meteor/client/components/message/content/attachments/file/hooks/useReloadOnError.tsx
abhinavkrin: You are right. Missed that.

abhinavkrin:
> During testing, I noticed that the first time I hit play, the video reloads, no matter if the URL is not expired yet.
> I was thinking, since we're adding a way to get the URL anyway, why not do it always, instead of waiting for it to fail before we attempt a recovery? I think it would simplify the logic a lot. We just need to save the last expiry date, if any of the events are fired we just check if `expiryDate < currentDate`, and if that's true we fetch again. WDYT?

Agreed, that will make the logic simpler. But as I looke at it again, I think we can prevent this initial fetch. 
Since, this fetch is useless in case of Google Storage, Filesystem, and gridfs.
I have modified the code a little and prevented url recovery when the media is played for the first time.
Initially we will wait for the stalled or error events. Once we have an expiry value with us, then we start checking for expiry for the play event as well.

```typescript
		if (firstRecoveryAttempted.current) {
			if (!expiresAt) {
				console.debug('No expiration time set, skipping recovery');
				isRecovering.current = false;
				return;
			}
		} else if (event.type === 'play') {
			// The user has initiated a playback for the first time, probably we should wait for the stalled or error event
			// the url may still be valid since we dont know the expiration time yet
			isRecovering.current = false;
			return;
		}
```

abhinavkrin:
> I noticed there are no tests, neither unit nor end-to-end, for both BE and FE. Do you plan on adding some?

I have added unit tests for the hook.

MarcosSpessatto (approved):

PR #36625, opened 2025-08-05
fix: prevent unexpected token errors from newer syntax in apps

d-gubert:

In some cases, the deno runtime will need to parse the app code to modify some method calls into async calls.

However, the ecmaVersion used by our parser (acorn) was too outdated, and could cause some `Unexpected token` errors sometimes.

This PR updates the ecmaVersion.

## Further comments

[CORE-1248]

[CORE-1248]: https://rocketchat.atlassian.net/browse/CORE-1248?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.changeset/twelve-plums-turn.md
+++ b/.changeset/twelve-plums-turn.md
@@ -1,0 +1,6 @@
+---
+'@rocket.chat/meteor': patch
+'@rocket.chat/apps-engine': patch
+---
+
+Fixes an issue that would cause the chat server to crash with an unhandled rejection in some cases
--- a/packages/apps-engine/src/server/runtime/deno/AppsEngineDenoRuntime.ts
+++ b/packages/apps-engine/src/server/runtime/deno/AppsEngineDenoRuntime.ts
@@ -85,6 +85,8 @@ export type DenoRuntimeOptions = {
 	timeout: number;
 };
 
+type AbortFunction = (reason?: any) => void;
+
 export class DenoRuntimeSubprocessController extends EventEmitter {
 	private deno: child_process.ChildProcess | undefined;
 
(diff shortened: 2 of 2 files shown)


.changeset/late-toes-remain.md
```diff
@@ -0,0 +1,5 @@
+---
+'@rocket.chat/apps-engine': patch
```
sampaiodiego, line 3:
```suggestion
'@rocket.chat/meteor': patch
'@rocket.chat/apps-engine': patch
```

packages/apps-engine/deno-runtime/lib/ast/mod.ts
```diff
@@ -50,7 +50,8 @@ function fixAst(ast: Program): boolean {
 
 export function fixBrokenSynchronousAPICalls(appSource: string): string {
 	const astRootNode = parse(appSource, {
-		ecmaVersion: 2017,
+		// Latest ecma version supported by this version of acorn.
+		ecmaVersion: "latest",
```
sampaiodiego, line 54: do you see any risks in some specific scenarios like when for example we need to backport a fix, so we'll need to publish a patch for an older release where at the time the ecma version was an "old one" but since then the "latest" version is now a newer, so for example, 7.10.0 the ecma version was 2025 and then let's say we need a patch later 2026 so on 7.10.1 the ecma version will be 2026 because this is the current latest (in the future).. 

this and any other "rebuild" that needs to be done for an old version can potentially use a newer version than when the code was done..

d-gubert: "latest" here is going to be bound to the version of `acorn` that we're using, so it goes only as far as 2023.

When we update acorn's version, we can review the ecmaVersion if that is the case. However, that should only be needed if we indeed need to support even newer syntax.

Note that we'll eventually drop acorn, we're only using it here to workaround a breaking change in app's code. At some point in the future we'll be able to adopt the breaking change in the apps-engine itself, and stop handling this case.

PR #36633, opened 2025-08-05
fix: messages redaction echo

ricardogarim:
this PR fixes a bug in the redaction flow from Rocket.Chat to Matrix, where `isEchoMessage` was mistakenly dropping redaction events before they were processed.

with this fix, redactions now work correctly in both directions (Rocket.Chat ↔ Matrix), including for thread messages.


ricardogarim:
lgtm! tested and merged into the feature branch :)

debdutdeb:
I've been just passing a skipMatrix optional arg to each function to avoid echo. But should have a better way to both reuse internal functions and evade the echo trap.

PR #36634, opened 2025-08-05
fix: pdf upload with exif data in images

cardoso:

This can only be reproduced with AWS S3 storage type due to recently introduced integrity checks:
https://aws.amazon.com/blogs/aws/introducing-default-data-integrity-protections-for-new-objects-in-amazon-s3/

The bug is caused because the Exif stripping logic in exif-be-gone’s ExifTransformer._scrubOther is applied to all "other" file types, including PDFs. However, PDF files do not contain EXIF, XMP, or FLIR markers in the same way as images, and stripping bytes from them can corrupt their structure, especially the xref table.

To strip Exif from images embedded in a PDF, we would need to:

1. Parse the PDF structure.
2. Extract each embedded image.
3. Strip Exif from each image individually.
4. Re-embed the cleaned images back into the PDF.

This is a much more complex task and requires a dedicated PDF library to manipulate PDF internals. The current stream-based approach is not sufficient for this.

The fix for now is preventing pdfs from going through this flow altogether.

## Issue(s)

https://rocketchat.atlassian.net/browse/SUP-833

## Steps to test or reproduce

I used the following python script in Google Colab to generate a minimal pdf file that triggers the issue:

```python
!pip install pillow piexif pymupdf
import fitz  # PyMuPDF
import piexif
from PIL import Image

# Step 1: Create a JPEG with EXIF (or use your own)
img = Image.new('RGB', (400, 300), color='blue')
exif_dict = {"0th": {piexif.ImageIFD.Artist: u"Test Artist"}}
exif_bytes = piexif.dump(exif_dict)
img.save("exif_test.jpg", exif=exif_bytes)
jpeg_path = "exif_test.jpg"  # Or use your own JPEG file with EXIF

# Step 2: Create a PDF and embed the JPEG as an image XObject
doc = fitz.open()
page = doc.new_page(width=600, height=800)
rect = fitz.Rect(100, 100, 500, 400)

with open(jpeg_path, "rb") as f:
    img_bytes = f.read()

# Embed the JPEG as-is (preserves EXIF)
page.insert_image(rect, stream=img_bytes)

doc.save("test_exif_embed.pdf")
print("PDF with embedded EXIF image generated: test_exif_embed.pdf")
```


cardoso:
Superseded by #36676

PR #36645, opened 2025-08-06
chore(deps-dev): bump typescript to 5.9.2

cardoso:

All changes are type only due to lib.d.ts changes:

https://www.typescriptlang.org/docs/handbook/release-notes/typescript-5-9.html#libdts-changes

## Issue(s)

https://rocketchat.atlassian.net/browse/CORE-1289


apps/meteor/app/e2e/client/rocketchat.e2e.room.ts
```diff
@@ -622,7 +622,7 @@ export class E2ERoom extends Emitter {
 	async encryptMessageContent(
 		contentToBeEncrypted: Pick<IMessage, 'attachments' | 'files' | 'file'> & Optional<Pick<IMessage, 'msg'>, 'msg'>,
 	) {
-		const data = new TextEncoder().encode(EJSON.stringify(contentToBeEncrypted));
+		const data = new TextEncoder().encode(EJSON.stringify(contentToBeEncrypted)) as Uint8Array<ArrayBuffer>;
```
tassoevan: This assertion feels expendable, as the definition from lib "dom" ensures the return type as `Uint8Array<ArrayBuffer>`.

cardoso: Yeah, I'm not sure what is going on here, but TextEncoder.encode still returns `Uint8Array` without type parameters which is equivalent to `Uint8Array<ArrayBufferLike>` instead of `Uint8Array<ArrayBuffer>`. I think the whole meteor folder is typechecked with node types which augment the global scope and might not be up to date. I'll see if I can fix this somehow.

cardoso: Weird, it was failing typechecking in the way I described above, but now it isn't. Pushed the change, hopefully CI isn't different for some reason.

d-gubert (approved):

tassoevan (approved):

PR #36660, opened 2025-08-07
refactor: fix ConvertToRoute for POST requests missing body and test to prove it

ahmed-n-abdeltwab:
**description:**

Change the ExtractValidation type to return 'unknown' instead of 'never' for the case where the post method doesnt have body


ahmed-n-abdeltwab:
@cardoso 👍

cardoso (dismissed):
@ggazzo 👍

apps/meteor/app/api/server/ApiClass.spec.ts
```diff
@@ -58,4 +58,21 @@ describe('ExtractRoutesFromAPI', () => {
...
+		type APIWithNeverQuery = APIClass<
+			'/v1',
+			{
+				method: 'POST';
+				path: '/v1/endpoint.test';
+				response: {
+					200: ValidateFunction<unknown>;
+				};
+				authRequired: true;
+			}
```
ggazzo, line 73: can we make required the  definition of `body: undefined`? I'm afraid people will skit the body definition

also  instead of unknown should not be the case of `undefined` also?

ahmed-n-abdeltwab: ok i will work on that

ahmed-n-abdeltwab: if we make the body field required in the type definition, it’ll also end up being required for the GET method; because GET and POST are both using the same TypedOptions interface:

```ts
export type TypedOptions = {
	response: {
		[K in HTTPStatusCodes]?: ValidateFunction;
	};
	query?: ValidateFunction;
	body: ValidateFunction;
	tags?: string[];
	typed?: boolean;
	license?: LicenseModule[];
} & SharedOptions<'GET' | 'POST' | 'PUT' | 'DELETE' | 'PATCH'>;
```
To fix this, we could make body optional in the base type and then create a separate type or utility that enforces it only for the methods that actually need it; like POST or PUT. This way, GET stays clean without an unnecessary body

ahmed-n-abdeltwab:
I forgot what this PR was for

cardoso:
No worries, that happens

